PR #7823, opened 2019-02-12
[DO NOT MERGE] [BEAM-4775] Second take on portable metrics over the job-server API

ryan-williams:
Umbrella PR tracking metrics over the job API, cf. https://s.apache.org/get-metrics-api.

### [Here is a direct link to what's in this PR but not yet rebased on top of component PRs below](https://github.com/ryan-williams/beam/pull/3).

[](https://github.com/ryan-williams/beam/pull/3/commits)

Progress!

### Approved / Ready to merge:
- #7890: consolidate MetricResult implementations
- #7883: Add MetricQueryResults.allMetrics() helper

### Iterating / Discussing
- #7868: MonitoringInfo URN tweaks
- #7876: Clean up metric protos; support integer distributions, gauges

### Holding pattern: java-core dependency-graph questions
- #7867: make `MetricResult`'s "key" explicitly a `MetricKey`, and route call-sites through that API (in preparation for expanding a `MetricKey` to encompass pcollection-scoped "system" metrics)

### Merged
- #7866: move function helpers from fn-harness to sdks/java/core

cc @robertwb @ajamato the factored-out PRs above are ready for review

## Likely pieces still to be factored out:
- fully supporting MonitoringInfo data types ({URN, labels}) inside `MetricKey` (comprised of `MetricName`,`MetricLabels`)
- adding the job-API metrics RPC
- python support

## Summary

- implements counter, distribution, and gauge metrics over the Job API
  - "int" versions only; removed "double" versions
  - works in Java; I think some of them may still be TODO in Python 
  - main RPC looks similar to that in the design doc, but adds an extra `MetricResults` proto that wraps "attempted" and "committed" lists of MonitoringInfos
- bypasses a lot of proto-additions and moves from #7641 that caused some `import`-thrashing
  - moving metrics-related protos to their own file/package may still make sense
  - this PR is a net remover of metric protos, so on balance I decided saving such a move for later is OK
- creates Java SDK counterparts of various MonitoringInfo components:
  - MetricKey ≈ MonitoringInfo ({URN, labels map})
  - MetricName ≈ URN
  - MetricLabels ≈ labels map
- encodes that metrics today must either be PTRANSFORM- or PCOLLECTION-scoped (not both)
- consolidates `MetricResult`s, and attempted/committed metrics logic, across all runners

## TODO
- Make sure all the metric types are being surfaced in Python.

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

[...]

PR #7866, opened 2019-02-17
[BEAM-4775] mv function pkg from fn-harness to java-core

ryan-williams:
also: parameterize ThrowingConsumer by Exception type

This is factored out of #7823, where a use-case for a `ThrowingConsumer<IOException, …>` exists, in Jackson JSON serde code.

R: @robertwb, @mxm 

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

Lang | SDK | Apex | Dataflow | Flink | Gearpump | Samza | Spark
--- | --- | --- | --- | --- | --- | --- | ---
Go | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go/lastCompletedBuild/) | --- | --- | --- | --- | --- | ---
Java | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Apex/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Dataflow/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Flink/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Flink_Batch/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Flink_Streaming/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Gearpump/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Samza/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Spark/lastCompletedBuild/)
[...]

PR #7867, opened 2019-02-17
[BEAM-4775] key MetricResult by a MetricKey

ryan-williams:
(factored out of #7823)

`MetricResult` previously had `MetricKey`'s fields inlined, possibly due to living in java/core while `MetricKey` is in runners/core. 

Having callers use a `MetricKey` simplifies call-sites and prepares for expanding Java SDK metrics to support (ptransform- and pcollection-scoped) system metrics in addition to (ptransform-scoped) user metrics, which #7823 implements.

This change also moves `MetricKey` and `MetricFiltering` into java/core, which is basically necessary / logical based on the intra-metrics-class dependencies.

R: @robertwb, @ajamato 


ryan-williams:
Run Java PreCommit

ryan-williams:
Run Java PreCommit

ajamato:
One thing I don't like about MetricKey, is that its only applicable to user metrics. Which are identified with the following fields:
- Step Name
- Namespace
- User provided Metric Name

MonitoringInfos are defined by
- URN
- Arbitrary set of Key value pair labels. 
- (Doesn't necessarily have a step name at all).

MetricKey is insufficient today to describe any metric which isn't defined by (step, namespace, metric name).

Take a look at MonitoringInfoMetricName (extends MetricName), which I added.
https://github.com/apache/beam/blob/5b1700db834cf949547a20a30f3ee1270710585a/runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MonitoringInfoMetricName.java

This can contain a URN and an arbitrary set of key value pair labels. I think this is the more generalized way to key a Metric and I would like to land on this in the future. I think we don't really need two separte classed MetricName and MetricKey either.

We could just have one class in the end similar to which represents the MetricKey.

The challenge I see with this though, is that querying the metrics and storing the metrics internally use the same classes  today (MetricKey). It would be better to keep the user facing API more similar to how the user defines a metric (explicitly being able to access the name and namespace fields). In the end we might have an internal MonitoringInfoMetricKey separate from a (User)MetricKey the user can us in the querying API.

If this change unblocks your work now, then I think its fine to proceed. But I wanted to give a bit more context here on how I think this could be simplified in the end. I think we have too many classes right now, for identifying metrics.

sdks/java/core/src/test/java/org/apache/beam/sdk/metrics/MetricResultsMatchers.java
```diff
@@ -194,28 +192,13 @@ protected void describeMismatchSafely(
...
       String name,
       String step) {
-    if (!Objects.equals(namespace, item.getName().getNamespace())) {
+    MetricKey key = MetricKey.create(step, MetricName.named(namespace, name));
+    if (!Objects.equals(key, item.getKey())) {
       mismatchDescription
-          .appendText("inNamespace: ")
-          .appendValue(namespace)
+          .appendText("inKey: ")
+          .appendValue(key)
```
ajamato: Please check if tostring is implemented in MonitoringInfoMetricKey and MetricKey

runners/spark/src/main/java/org/apache/beam/runners/spark/metrics/SparkBeamMetric.java
```diff
@@ -17,27 +17,30 @@
...
 import org.apache.beam.sdk.metrics.MetricResult;
 import org.apache.beam.sdk.metrics.MetricResults;
 import org.apache.beam.sdk.metrics.MetricsFilter;
 import org.apache.beam.vendor.guava.v20_0.com.google.common.annotations.VisibleForTesting;
+import org.apache.beam.vendor.guava.v20_0.com.google.common.collect.ImmutableList;
 
 /**
  * An adapter between the {@link MetricsContainerStepMap} and Codahale's {@link Metric} interface.
  */
 class SparkBeamMetric implements Metric {
```
ajamato: I also notice the Metric class will only handle User Metrics (Name, Namespace). If spark only cares about UserMetrics that's okay I suppose. but won't handle more generlized MonitoringInfos

runners/extensions-java/metrics/src/main/java/org/apache/beam/runners/extensions/metrics/MetricsHttpSink.java
```diff
@@ -17,20 +17,28 @@
...
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import javax.xml.ws.http.HTTPException;
 import org.apache.beam.sdk.annotations.Experimental;
+import org.apache.beam.sdk.metrics.MetricKey;
+import org.apache.beam.sdk.metrics.MetricName;
 import org.apache.beam.sdk.metrics.MetricQueryResults;
+import org.apache.beam.sdk.metrics.MetricResult;
 import org.apache.beam.sdk.metrics.MetricsOptions;
 import org.apache.beam.sdk.metrics.MetricsSink;
```
ajamato: Do you have any more details for this class? Link to doc? Explaination of how to use it. Example code of handling the Post request, etc.? What is the wire format used? Can it be included in this header?

ryan-williams:
Thanks Alex. The good news is I think we agree about everything you've said.

In #7823 I basically turn the existing Java SDK metrics classes into wrappers for corresponding MonitoringInfo structures:
- `MetricKey`: {URN, labels}
- `MetricName`: URN
  - I basically folded `MonitoringInfoMetricName` into `MetricName`, which becomes the only implementation of that concept
- `MetricLabels`: labels map
  - thin wrapper over the raw map, that also exposes some APIs that encode the fact that today we expect MonitoringInfos to have exactly one of {`PTRANSFORM`,`PCOLLECTION`}.

So that is the larger vision, which this PR is a small step toward. LMK if that sounds off to you. I'm in a constant process of pulling stuff out of #7823 so that the real work will be in clearer relief there.

A different question that [came up with @mxm on #7866](https://github.com/apache/beam/pull/7866#issuecomment-465241925) is around what packages things should live in. I don't think it's too bad if `MetricKey` goes into java/core here, but where it gets complicated is that inevitably, MetricResult → MetricKey → {fn-api protos}, which adds [java/core → model-fn-execution], which gives me some pause.

Hopefully we can push on other parts of this (and related PRs) while I think more about that question. Ofc any thoughts are welcome.

runners/spark/src/main/java/org/apache/beam/runners/spark/metrics/SparkBeamMetric.java
ryan-williams: Agreed. Most of what I've done so far (in #7823) is [expanding the Java SDK metrics classes to support the wider MonitoringInfo data model] while [preserving semantics of existing code that assumes / requires user-metrics].

In this case, [I end up adapting this](https://github.com/apache/beam/pull/7823/files#diff-2b144bc5b4cec16779fa6126425c6fe3R71) to use [a `MetricLabels.value()` API](https://github.com/apache/beam/pull/7823/files#diff-245a1fd7e743b2f382c978a12028f8c2R46) that assumes there is exactly one label set on the MonitoringInfo (`PTRANSFORM` xor `PCOLLECTION`) and uses that, so it should be reasonably "future-proofed"

runners/extensions-java/metrics/src/main/java/org/apache/beam/runners/extensions/metrics/MetricsHttpSink.java
ryan-williams: Yea, those would be nice to have 😀 I don't really know anything about it, and just made these changes to preserve the existing format. 

I'll do some digging and try to formalize / document it better, though I assume that [not rocking the wire-format boat with whatever is going on here] is going to be what we want to do, in the short term at least.

ryan-williams:
hm, I see a bunch of 
```
BeamIOError: src and dst files do not exist. src: /tmp/beam-temp-py-wordcount-direct-42fabb3a34cb11e984ac42010a80000d/1c504dc3-20b6-4f8c-88ef-15912e953404.py-wordcount-direct, dst: /tmp/py-wordcount-direct-00000-of-00002 with exceptions None [while running 'write/Write/WriteImpl/FinalizeWrite
--
'] with exceptions None
```
[in the scan](https://scans.gradle.com/s/5qa6sgdele2rs/console-log?task=:beam-sdks-python:portableWordCountBatch#L54) for `Portable_Python` (batch; streaming seems to have passed).

ryan-williams:
Run Portable_Python PreCommit

PR #7868, opened 2019-02-17
[BEAM-4775] MonitoringInfo URN tweaks

ryan-williams:
rename a couple of proto URNs and specs, and perform a few other improvements to the Java SDK's handling of MonitoringInfos

R: @robertwb, @ajamato 

### DIFF
--- a/runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MetricUrns.java
+++ b/runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MetricUrns.java
@@ -19,6 +19,7 @@ package org.apache.beam.runners.core.metrics;
 
 import static org.apache.beam.runners.core.metrics.SimpleMonitoringInfoBuilder.USER_COUNTER_URN_PREFIX;
 
+import javax.annotation.Nullable;
 import org.apache.beam.sdk.metrics.MetricName;
 
 /** Utility for parsing a URN to a {@link MetricName}. */
@@ -29,14 +30,17 @@ public class MetricUrns {
    *
    * <p>Should be consistent with {@code parse_namespace_and_name} in monitoring_infos.py.
    */
-  public static MetricName parseUrn(String urn) {
+  @Nullable
+  public static MetricName parseUserMetricUrn(String urn) {
     if (urn.startsWith(USER_COUNTER_URN_PREFIX)) {
       urn = urn.substring(USER_COUNTER_URN_PREFIX.length());
+    } else {
+      return null;
     }
     // If it is not a user counter, just use the first part of the URN, i.e. 'beam'
     String[] pieces = urn.split(":", 2);
     if (pieces.length != 2) {
-      throw new IllegalArgumentException("Invalid metric URN: " + urn);
+      throw new IllegalArgumentException("Invalid user-metric URN: " + urn);
     }
     return MetricName.named(pieces[0], pieces[1]);
   }
--- a/runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MetricUrns.java
+++ b/runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MetricUrns.java
@@ -17,7 +17,7 @@
  */
 package org.apache.beam.runners.core.metrics;
 
-import static org.apache.beam.runners.core.metrics.SimpleMonitoringInfoBuilder.USER_COUNTER_URN_PREFIX;
+import static org.apache.beam.runners.core.metrics.SimpleMonitoringInfoBuilder.USER_METRIC_URN_PREFIX;
 
 import javax.annotation.Nullable;
 import org.apache.beam.sdk.metrics.MetricName;
@@ -32,8 +32,8 @@ public class MetricUrns {
    */
   @Nullable
   public static MetricName parseUserMetricUrn(String urn) {
-    if (urn.startsWith(USER_COUNTER_URN_PREFIX)) {
-      urn = urn.substring(USER_COUNTER_URN_PREFIX.length());
+    if (urn.startsWith(USER_METRIC_URN_PREFIX)) {
+      urn = urn.substring(USER_METRIC_URN_PREFIX.length());
     } else {
       return null;
     }
--- a/model/pipeline/src/main/proto/metrics.proto
+++ b/model/pipeline/src/main/proto/metrics.proto
@@ -33,13 +33,12 @@ import "beam_runner_api.proto";
 import "google/protobuf/descriptor.proto";
 import "google/protobuf/timestamp.proto";
 
-// A specification containing required set of fields and labels required
-// to be set on a MonitoringInfo for the specific URN for SDK->RunnerHarness
-// ProcessBundleResponse reporting.
+// Specifications containing required fields and labels for a given
+// MonitoringInfo URN, for SDK->RunnerHarness ProcessBundleResponse reporting.
 message MonitoringInfoSpec {
   string urn = 1;
   string type_urn = 2;
-  // The list of required
+  // The list of required labels for this URN
   repeated string required_labels = 3;
   // Extra non functional parts of the spec for descriptive purposes.
   // i.e. description, units, etc.
@@ -61,7 +60,7 @@ message MonitoringInfoSpecs {
   enum Enum {
     // TODO(ajamato): Add the PTRANSFORM name as a required label after
     // upgrading the python SDK.
-    USER_COUNTER = 0 [(monitoring_info_spec) = {
+    USER_METRIC = 0 [(monitoring_info_spec) = {
       urn: "beam:metric:user:",
       type_urn: "beam:metrics:sum_int_64",
     }];
@@ -183,8 +182,8 @@ message MonitoringInfo {
 
 message MonitoringInfoUrns {
   enum Enum {
-    // User counter have this format: 'beam:metric:user:<namespace>:<name>'.
-    USER_COUNTER_URN_PREFIX = 0
+    // User-metric URNs have this format: 'beam:metric:user:<namespace>:<name>'
+    USER_METRIC_URN_PREFIX = 0
         [(org.apache.beam.model.pipeline.v1.beam_urn) = "beam:metric:user:"];
 
     ELEMENT_COUNT = 1 [(org.apache.beam.model.pipeline.v1.beam_urn) =
--- a/model/pipeline/src/main/proto/metrics.proto
+++ b/model/pipeline/src/main/proto/metrics.proto
@@ -69,6 +69,8 @@ message MonitoringInfoSpecs {
     ELEMENT_COUNT = 1 [(monitoring_info_spec) = {
       urn: "beam:metric:element_count:v1",
       type_urn: "beam:metrics:sum_int_64",
+      // TODO(ryan): we currently also generate this metric with ["PTRANSFORM"] labels, but it fails validation in
+      // Java's SimpleMonitoringInfoBuilder (and emits a WARN); should ["PTRANSFORM"] and ["PCOLLEcTION"] both work?
       required_labels: [ "PCOLLECTION" ],
       annotations: [ {
         key: "description",
--- a/model/pipeline/src/main/proto/metrics.proto
+++ b/model/pipeline/src/main/proto/metrics.proto
@@ -158,8 +158,7 @@ message MonitoringInfo {
   }
 
   enum MonitoringInfoLabels {
-    // TODO(ajamato): Rename all references to TRANSFORM to PTRANSFORM
-    TRANSFORM = 0 [(label_props) = { name: "PTRANSFORM" }];
+    PTRANSFORM = 0 [(label_props) = { name: "PTRANSFORM" }];
     PCOLLECTION = 1 [(label_props) = { name: "PCOLLECTION" }];
     WINDOWING_STRATEGY = 2 [(label_props) = { name: "WINDOWING_STRATEGY" }];
     CODER = 3 [(label_props) = { name: "CODER" }];
--- a/runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MonitoringInfoMetricName.java
+++ b/runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MonitoringInfoMetricName.java
@@ -23,6 +23,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
 import javax.annotation.Nullable;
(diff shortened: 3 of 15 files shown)


ryan-williams:
Run Java PreCommit

model/fn-execution/src/main/proto/beam_fn_api.proto
```diff
@@ -316,7 +316,7 @@ message ProcessBundleProgressRequest {
 // ProcessBundleResponse reporting.
 message MonitoringInfoSpec {
   string urn = 1;
-  string type_urn = 2;
+  repeated string type_urns = 2;
```
robertwb: What does it mean for this to be repeated? A metric only has a single type.

model/fn-execution/src/main/proto/beam_fn_api.proto
```diff
@@ -339,14 +339,16 @@ message MonitoringInfoSpecs {
   enum Enum {
     // TODO(ajamato): Add the PTRANSFORM name as a required label after
     // upgrading the python SDK.
-    USER_COUNTER = 0 [(monitoring_info_spec) = {
+    USER_METRIC = 0 [(monitoring_info_spec) = {
```
robertwb: beam:metric:user: is a prefix, not a full URN, and probably shouldn't be in this table.

ryan-williams: Currently, metrics are either of "user" type (defined by {ptransform, namespace, name}), or they're one of the 5 "system"-metric types below. 

It's true that the "user" category is not a "specific URN", so there is tension with e.g. [this docstring above](https://github.com/apache/beam/pull/7868/files#diff-4d226ebf3f70cf18449c29e82511a67eR333).

I think of this table as enumerating the possible metrics-URN *categories* (indeed, that's what it does, and what our design calls for, as opposed to e.g. enumerating all allowed "user"-metric URNs here). The 5 "system" categories are each inhabited by exactly one URN, so they are particularly easy to deal with in SDKs, while the user-metric category requires special treatment (which, conveniently, the SDKs already implement).

In this PR, I am laying some groundwork toward supporting more "types" of metrics (both "user" and "system"). "User" metrics should be able to take on the three types listed here (counter, distribution, gauge).

With that context, there's a few directions I can see taking your comment:
- split the "user" category into 3 type-specific user-categories
- leave as-is, but make the docs more clear about the special status of `USER_METRIC`'s "spec"
- some larger overhaul involving removing user metrics from this table.

Let me know what you think!

model/fn-execution/src/main/proto/beam_fn_api.proto
ryan-williams: I'll add a comment, but this is a list of *allowed* types for a URN conforming to a given `MonitoringInfoSpec`.

As part of expanding the `USER_COUNTER` category to handle distributions and gauges, I need to either expand validation to support those types (done in this PR [here](https://github.com/apache/beam/pull/7868/files#diff-90a5bbb6f36b2c93562e3cffa14350fdR64)), or introduce new user-metric URN-categories for those new types (as discussed below).

model/fn-execution/src/main/proto/beam_fn_api.proto
robertwb: The "user metric" prefix should not be in this table. (We also shouldn't be restricting user metrics here, any valid metric type is a valid user metric type.)

@ajamato

ryan-williams: OK. Should [`USER_METRIC_URN_PREFIX`](https://github.com/apache/beam/pull/7868/files#diff-4d226ebf3f70cf18449c29e82511a67eR466) (aka `USER_COUNTER_URN_PREFIX`) also not be in `MonitoringInfoUrns.Enum`, or is the issue just with this `MonitoringInfoSpecs` table?

robertwb: Just an issue with the table, which should be listing specific metrics.

ryan-williams:
Run Java PreCommit

model/fn-execution/src/main/proto/beam_fn_api.proto
```diff
@@ -311,13 +311,13 @@ message ProcessBundleProgressRequest {
   string instruction_reference = 1;
 }
 
-// A specification containing required set of fields and labels required
-// to be set on a MonitoringInfo for the specific URN for SDK->RunnerHarness
+// A specification containing required set of fields and labels required to be
+// set on a MonitoringInfo for a given URN-type, for SDK->RunnerHarness
```
ajamato: That's not correct. The specific URN has a set of required field.
ElementCount has a PCollection.
ExecutionTimes have a PTransform.

URN-type refers to the value type. counter int64, counter double, gauge string, gauge int64, etc.

model/fn-execution/src/main/proto/beam_fn_api.proto
```diff
@@ -311,13 +311,13 @@ message ProcessBundleProgressRequest {
...
 // ProcessBundleResponse reporting.
 message MonitoringInfoSpec {
   string urn = 1;
   string type_urn = 2;
-  // The list of required
+  // The list of required labels for this URN-type
```
ajamato: ditto, should be URN

model/fn-execution/src/main/proto/beam_fn_api.proto
```diff
@@ -337,24 +337,19 @@ message Annotation {
...
-    USER_COUNTER = 0 [(monitoring_info_spec) = {
-      urn: "beam:metric:user:",
-      type_urn: "beam:metrics:sum_int_64",
-    }];
-
-    ELEMENT_COUNT = 1 [(monitoring_info_spec) = {
+    ELEMENT_COUNT = 0 [(monitoring_info_spec) = {
       urn: "beam:metric:element_count:v1",
       type_urn: "beam:metrics:sum_int_64",
+      // TODO(ryan): we currently also generate this metric with ["PTRANSFORM"] labels, but it fails validation in
```
ajamato: That's in python @Ardagan is fixing that implementation to emit the PCollection name. This is the one spec runners should support.

model/fn-execution/src/main/proto/beam_fn_api.proto
```diff
@@ -337,24 +337,19 @@ message Annotation {
 // MonitoringInfo protos.
 message MonitoringInfoSpecs {
   enum Enum {
-    // TODO(ajamato): Add the PTRANSFORM name as a required label after
```
ajamato: Let's keep this one in please. @Ardagan's validation is going to pull in these strings soon. We are hoping to remove the MonitoringInfoUrns section and just have MonitoringInfoSpecs

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MetricUrns.java
```diff
@@ -29,14 +30,17 @@
...
    * <p>Should be consistent with {@code parse_namespace_and_name} in monitoring_infos.py.
    */
+  @Nullable
   public static MetricName parseUrn(String urn) {
-    if (urn.startsWith(USER_COUNTER_URN_PREFIX)) {
-      urn = urn.substring(USER_COUNTER_URN_PREFIX.length());
+    if (urn.startsWith(USER_METRIC_URN_PREFIX)) {
+      urn = urn.substring(USER_METRIC_URN_PREFIX.length());
+    } else {
+      return null;
```
ajamato: Did you encounter this somewhere? Hopefully this isn't called if the metric isn't actually starting with USER_METRIC_URN_PREFIX

model/fn-execution/src/main/proto/beam_fn_api.proto
ryan-williams: hm, @robertwb [wanted it removed](https://github.com/apache/beam/pull/7868#discussion_r257840224), IIUC.

model/fn-execution/src/main/proto/beam_fn_api.proto
ryan-williams: Yea, "type" was not what I meant here; context was [discussion w/ Robert](https://github.com/apache/beam/pull/7868#discussion_r257795396) about whether these specs describe "specific URNs" or "URN categories" (e.g. "user metrics").

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MetricUrns.java
ryan-williams: As of this PR, this helper is called in two places:
- [`MonitoringInfoMetricName.parseUrn`](https://github.com/apache/beam/pull/7868/files#diff-ba1b936a7d4c8d43789e81a816b9bb90R55):
  - use the parsed user-`MetricName`, if possible
  - `throw` if it's not a user-metric (signaled by `null` return here)
- [`FlinkMetricContainer.updateMetrics`](https://github.com/apache/beam/pull/7868/files#diff-bd906385f3e3ff74094985af80da5a41R108):
  - use the parsed user-`MetricName`, if possible
  - skip the metric if it's not a user-metric (again, signaled by `null` return)
    - in later PRs I make Flink support system-metrics as well.

This was the most direct way to support those uses. The Flink metrics do actually see system metrics, and should gracefully skip them instead of `throw`ing (for now).

ryan-williams:
> Just curious, what were you hoping to get out of this change in particular? This is largely a bunch of renamings, was it too confusing before?

There are 2 renamings:
- https://github.com/apache/beam/pull/7868/commits/737c9e25ebbbf2654d2294705d5d3f14e17a841d: `MonitoringInfoLabels.{TRANSFORM → PTRANSFORM}`, per [this `TODO(ajamato)`](https://github.com/apache/beam/pull/7868/commits/737c9e25ebbbf2654d2294705d5d3f14e17a841d#diff-4d226ebf3f70cf18449c29e82511a67eL439) 😄  
- https://github.com/apache/beam/pull/7868/commits/c17694b3486e1e0f03f5dfd63b6c108dc68caf02:  `USER_COUNTER_URN_PREFIX ` → `USER_METRIC_URN_PREFIX `
  - (in `SimpleMonitoringInfoBuilder` and `MonitoringInfoUrns.Enum`)
  - we need either this change, or to add new user-URN-prefixes for distributions/gauges, afaict.

There are also 6 other commits here that each constitute minor forward progress, I think:
- https://github.com/apache/beam/pull/7868/commits/08108ccdfad500bb68241e588bb525e36e0efa3d: `UserMonitoringInfoToCounterUpdateTransformer` had [a redundant definition of  `USER_COUNTER_URN_PREFIX`](https://github.com/apache/beam/pull/7868/commits/08108ccdfad500bb68241e588bb525e36e0efa3d#diff-a2189a7c97dace5f44f04e16a71b6468L56)
- https://github.com/apache/beam/pull/7868/commits/937a9c68d7d81af28d503ad9b3fc4056bb0253ab: consolidate two (different!) implementations of [parsing a MetricName from a URN]
- etc.

Sorry if it just feels like dust. I thought these were all thematically similar and a reasonable chunk to pull out of #7823 and reckon with independently.

ryan-williams:
Run Java PreCommit

ryan-williams:
```
:beam-sdks-java-io-cassandra:test FAILED
--
 
org.apache.beam.sdk.io.cassandra.CassandraIOTest > classMethod FAILED
java.lang.AssertionError at CassandraIOTest.java:100
 
org.apache.beam.sdk.io.cassandra.CassandraIOTest > classMethod FAILED
java.lang.NullPointerException at CassandraIOTest.java:125
 
6 tests completed, 2 failed
```

[scan](https://scans.gradle.com/s/btppkeky63a5g/console-log?task=:beam-sdks-java-io-cassandra:test)

model/fn-execution/src/main/proto/beam_fn_api.proto
ryan-williams: OK. I'll leave this for now, perhaps I can remove it before this gets merged, if @Ardagan fixes it first, or lmk what you suggest!

ryan-williams:
Run Java PreCommit

model/fn-execution/src/main/proto/beam_fn_api.proto
robertwb: The problem is that "beam:metric:user:" isn't actually a counter, nor does it have a specific type, so putting it in this list seems wrong. It's just a prefix that we assert will not be used for any runner-generated counters. 

[If we do remove it, no need to renumber everything else.]

model/fn-execution/src/main/proto/beam_fn_api.proto
ryan-williams: I removed this change, 👍🏻

model/fn-execution/src/main/proto/beam_fn_api.proto
ryan-williams: > [If we do remove it, no need to renumber everything else.]

`generateProto` fails without the renumbering:

```
Execution failed for task ':beam-model-fn-execution:generateProto'.
> protoc: stdout: . stderr: beam_fn_api.proto: The first enum value must be zero in proto3.
```

ryan-williams: (update: it seems if I only renumber the first entry to `0`, I can skip `1` and leave the others untouched. Given that these aren't really used yet, it feels cleaner to just renumber them all now)

ryan-williams:
I think the big outstanding question here is around removing the spec for user metrics [here](https://github.com/apache/beam/pull/7868/files#r258386456) (with context [here](https://github.com/apache/beam/pull/7868#discussion_r257846463)).

model/fn-execution/src/main/proto/beam_fn_api.proto
Ardagan: My take on this:
1. It is better ot keep this spec present. It defines what fields user counter needs.
2. We can rename this element to something like USER_COUNTER_PREFIX for clarity.
3. It is ok to move it away from this list, but we should still keep it somewhere for generic validation and specification. However in this case we might want to move it away from MonitoringInfoUrns as well.
4. Currently, we have urns defined in this list and in MonitoringInfoUrns. I was thinking of removing MonitoringInfoUrns completely and keep only MonitoringInfoSpecs as explicit source of truth.

model/fn-execution/src/main/proto/beam_fn_api.proto
ajamato: @Ardagan and I have discussed that its not really the best idea to make the URN a prefix for the user metric. As this has led to writing a lot of code to parse the URN to obtain those fields out. And being this weird exception makes all the code have to special case it.

It would be better to package the namespace and name as a label on the MonitoringInfo.
Then we could use the same URN everywhere, no parsing, no special casing.

Given this, then there are two choices i.e.
(1) Use a different URN for user gauge, counter and distribution
- "beam:metric:user:counter:v1"
- "beam:metric:user:gauge:v1"
- "beam:metric:user:distribution:v1"
Then change the MonitoringInfoSpec:
      required_labels: [ "PTRANSFORM", "NAMESPACE", "NAME" ],
and each will have a separate  type_urn in their spec.
      type_urn:


(2) Use a single URN "beam:metric:user:v1" for user gauge, counter, distribution
Then change the MonitoringInfoSpec as well:
      required_labels: [ "PTRANSFORM", "NAMESPACE", "NAME" ],
and we will NOT enforce the type_urn in the spec.


I prefer #1, as this will better describe each one, and if we add new user metric styles we can add a spec for them

PR #7875, opened 2019-02-18
[BEAM-6730] Support configuring transforms externally in Java SDK / Expose Java's GenerateSequence in Python

mxm:
This leverages the `ExpansionService` to be able to use Java SDK's `GenerateSequence` from a Python pipeline.

I've implemented this to see how complicated it would be to configure and expose an external transform.

CC @robertwb @tweise 

------------------------

### DIFF
--- a/model/pipeline/src/main/proto/external_transforms.proto
+++ b/model/pipeline/src/main/proto/external_transforms.proto
@@ -1,0 +1,52 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * Protocol Buffers describing the external transforms available.
+ */
+
+syntax = "proto3";
+
+package org.apache.beam.model.pipeline.v1;
+
+option go_package = "pipeline_v1";
+option java_package = "org.apache.beam.model.pipeline.v1";
+option java_outer_classname = "ExternalTransforms";
+
+import "beam_runner_api.proto";
+import "google/protobuf/duration.proto";
+
+// Payload for Java's GenerateSequence transform
+message GenerateSequencePayload {
+  enum Enum {
+    PROPERTIES = 0 [(beam_urn) = "beam:external:transform:generate_sequence:v1"];
+  }
+  int64 start = 1;
+  oneof stop {
+    bool stop_provided = 2;
+    int64 stop_value = 3;
+  }
+  oneof elements_per_period {
+    bool elements_per_period_provided = 4;
+    int64 elements_per_period_value = 5;
+  }
+  google.protobuf.Duration period = 6;
+  google.protobuf.Duration max_read_time = 7;
+  // TODO How are foreign functions treated?
+  //SdkFunctionSpec timestamp_fn = 8;
+}
--- a/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/expansion/ExpansionService.java
+++ b/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/expansion/ExpansionService.java
@@ -180,11 +180,17 @@ public class ExpansionService extends ExpansionServiceGrpc.ExpansionServiceImplB
                 .filter(id -> !existingTransformIds.contains(id))
                 .collect(Collectors.toList()));
     RunnerApi.Components components = pipelineProto.getComponents();
-    LOG.debug("Expanded to {}", components.getTransformsOrThrow(expandedTransformId));
+    RunnerApi.PTransform expandedTransform =
+        components
+            .getTransformsOrThrow(expandedTransformId)
+            .toBuilder()
+            .setUniqueName(expandedTransformId)
+            .build();
+    LOG.debug("Expanded to {}", expandedTransform);
 
     return ExpansionApi.ExpansionResponse.newBuilder()
         .setComponents(components.toBuilder().removeTransforms(expandedTransformId))
-        .setTransform(components.getTransformsOrThrow(expandedTransformId))
+        .setTransform(expandedTransform)
         .build();
   }
 
--- a/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/expansion/ExpansionService.java
+++ b/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/expansion/ExpansionService.java
@@ -19,6 +19,8 @@ package org.apache.beam.runners.core.construction.expansion;
 
 import com.google.auto.service.AutoService;
 import java.io.IOException;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Method;
 import java.util.Collections;
 import java.util.Map;
 import java.util.ServiceLoader;
@@ -27,12 +29,18 @@ import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import org.apache.beam.model.expansion.v1.ExpansionApi;
 import org.apache.beam.model.expansion.v1.ExpansionServiceGrpc;
+import org.apache.beam.model.pipeline.v1.ExternalTransforms;
 import org.apache.beam.model.pipeline.v1.RunnerApi;
+import org.apache.beam.runners.core.construction.BeamUrns;
 import org.apache.beam.runners.core.construction.Environments;
 import org.apache.beam.runners.core.construction.PipelineTranslation;
 import org.apache.beam.runners.core.construction.RehydratedComponents;
 import org.apache.beam.runners.core.construction.SdkComponents;
 import org.apache.beam.sdk.Pipeline;
+import org.apache.beam.sdk.coders.Coder;
+import org.apache.beam.sdk.coders.VarLongCoder;
+import org.apache.beam.sdk.expansion.ExternalTransformRegistrar;
+import org.apache.beam.sdk.transforms.ExternalTransformBuilder;
 import org.apache.beam.sdk.transforms.PTransform;
 import org.apache.beam.sdk.values.PCollection;
 import org.apache.beam.sdk.values.PCollectionList;
@@ -45,6 +53,9 @@ import org.apache.beam.vendor.grpc.v1p13p1.io.grpc.Server;
 import org.apache.beam.vendor.grpc.v1p13p1.io.grpc.ServerBuilder;
 import org.apache.beam.vendor.grpc.v1p13p1.io.grpc.stub.StreamObserver;
 import org.apache.beam.vendor.guava.v20_0.com.google.common.annotations.VisibleForTesting;
+import org.apache.beam.vendor.guava.v20_0.com.google.common.base.CaseFormat;
+import org.apache.beam.vendor.guava.v20_0.com.google.common.base.Converter;
+import org.apache.beam.vendor.guava.v20_0.com.google.common.base.Preconditions;
 import org.apache.beam.vendor.guava.v20_0.com.google.common.collect.ImmutableMap;
 import org.apache.beam.vendor.guava.v20_0.com.google.common.collect.Iterables;
 import org.slf4j.Logger;
@@ -68,6 +79,121 @@ public class ExpansionService extends ExpansionServiceGrpc.ExpansionServiceImplB
    */
   public interface ExpansionServiceRegistrar {
     Map<String, TransformProvider> knownTransforms();
+  }
+
+  /**
+   * Exposes Java transforms via {@link org.apache.beam.sdk.expansion.ExternalTransformRegistrar}.
+   */
+  @AutoService(ExpansionService.ExpansionServiceRegistrar.class)
+  public static class ExternalTransformRegistrarLoader<ConfigT>
+      implements ExpansionService.ExpansionServiceRegistrar {
+
+    @Override
+    public Map<String, ExpansionService.TransformProvider> knownTransforms() {
+      ImmutableMap.Builder<String, ExpansionService.TransformProvider> builder =
+          ImmutableMap.builder();
+      for (ExternalTransformRegistrar registrar :
+          ServiceLoader.load(ExternalTransformRegistrar.class)) {
+        for (Map.Entry<String, Class<? extends ExternalTransformBuilder<?, ?, ?>>> entry :
+            registrar.knownBuilders().entrySet()) {
+          String urn = entry.getKey();
+          Class<? extends ExternalTransformBuilder<?, ?, ?>> builderClass = entry.getValue();
+          builder.put(
+              urn,
+              spec -> {
+                try {
+                  ExternalTransforms.ExternalConfigurationPayload payload =
+                      ExternalTransforms.ExternalConfigurationPayload.parseFrom(spec.getPayload());
+                  return translate(payload, builderClass);
+                } catch (Exception e) {
+                  throw new RuntimeException(
+                      String.format("Failed to build transform %s from spec %s", urn, spec), e);
+                }
+              });
+        }
+      }
+      return builder.build();
+    }
+
+    private static PTransform translate(
+        ExternalTransforms.ExternalConfigurationPayload payload,
+        Class<? extends ExternalTransformBuilder<?, ?, ?>> builderClass)
+        throws Exception {
+      Preconditions.checkState(
+          ExternalTransformBuilder.class.isAssignableFrom(builderClass),
+          "Provided identifier %s is not an ExternalTransformBuilder.",
+          builderClass.getName());
+
+      Object configObject = initConfiguration(builderClass);
+      populateConfiguration(configObject, payload);
+      return buildTransform(builderClass, configObject);
+    }
+
+    private static Object initConfiguration(
+        Class<? extends ExternalTransformBuilder<?, ?, ?>> builderClass) throws Exception {
+      for (Method method : builderClass.getMethods()) {
+        if (method.getName().equals("buildExternal")) {
+          Preconditions.checkState(
+              method.getParameterCount() == 1,
+              "Build method for ExternalTransformBuilder %s must have exactly one parameter, but had %s parameters.",
+              builderClass.getSimpleName(),
+              method.getParameterCount());
+          Class<?> configurationClass = method.getParameterTypes()[0];
+          if (Object.class.equals(configurationClass)) {
+            continue;
+          }
+          return configurationClass.getDeclaredConstructor().newInstance();
+        }
+      }
+      throw new RuntimeException("Couldn't find build method on ExternalTransformBuilder.");
+    }
+
+    private static void populateConfiguration(
+        Object config, ExternalTransforms.ExternalConfigurationPayload payload) throws Exception {
+      Converter<String, String> camelCaseConverter =
+          CaseFormat.LOWER_UNDERSCORE.converterTo(CaseFormat.LOWER_CAMEL);
+      for (Map.Entry<String, ExternalTransforms.ConfigValue> entry :
+          payload.getConfigurationMap().entrySet()) {
+        String fieldName = camelCaseConverter.convert(entry.getKey());
+        String coderUrn = entry.getValue().getCoderUrn();
+
+        final Coder coder;
+        final Class type;
+        if (BeamUrns.getUrn(RunnerApi.StandardCoders.Enum.VARINT).equals(coderUrn)) {
+          coder = VarLongCoder.of();
+          type = Long.class;
+        } else {
+          // TODO Use RehydratedComponents with coder ids instead
+          throw new RuntimeException("Unsupported coder urn " + coderUrn);
+        }
+        String setterName =
+            "set" + Character.toUpperCase(fieldName.charAt(0)) + fieldName.substring(1);
+        Method method;
+        try {
+          // retrieve the setter for this field
+          method = config.getClass().getMethod(setterName, type);
+        } catch (NoSuchMethodException e) {
+          throw new RuntimeException(
+              String.format(
+                  "The configuration class %s is missing a setter %s for %s",
+                  config.getClass(), setterName, fieldName),
+              e);
+        }
+        method.invoke(config, coder.decode(entry.getValue().getPayload().newInput()));
+      }
+    }
+
+    private static PTransform buildTransform(
+        Class<? extends ExternalTransformBuilder<?, ?, ?>> builderClass, Object configObject)
+        throws Exception {
+      Constructor<? extends ExternalTransformBuilder<?, ?, ?>> constructor =
+          builderClass.getDeclaredConstructor();
+      constructor.setAccessible(true);
+      ExternalTransformBuilder<?, ?, ?> externalTransformBuilder = constructor.newInstance();
+      Method buildMethod = builderClass.getMethod("buildExternal", configObject.getClass());
+      buildMethod.setAccessible(true);
+      return (PTransform) buildMethod.invoke(externalTransformBuilder, configObject);
+    }
   }
 
   /**
--- a/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/expansion/ExpansionService.java
+++ b/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/expansion/ExpansionService.java
@@ -38,6 +38,7 @@ import org.apache.beam.sdk.values.PCollection;
 import org.apache.beam.sdk.values.PCollectionList;
 import org.apache.beam.sdk.values.PCollectionTuple;
 import org.apache.beam.sdk.values.PDone;
+import org.apache.beam.sdk.values.PInput;
 import org.apache.beam.sdk.values.PValue;
 import org.apache.beam.sdk.values.TupleTag;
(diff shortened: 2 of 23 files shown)


mxm:
retest this please

mxm:
Could you have a look @robertwb?

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/graph/GreedyPipelineFuser.java
```diff
@@ -186,14 +185,6 @@ private DescendantConsumers getRootConsumers(PTransformNode rootNode) {
         "Transform %s is not at the root of the graph (consumes %s)",
         rootNode.getId(),
         rootNode.getTransform().getInputsMap());
-    checkArgument(
```
tweise: Why remove this? Isn't READ_TRANSFORM_URN a runner implemented primitive that should be seen as such?

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkStreamingPortablePipelineTranslator.java
```diff
@@ -192,18 +204,28 @@ public StreamExecutionEnvironment getExecutionEnvironment() {
...
+    // Remove once Reads can be wrapped in SDFs
+    translatorMap.put(PTransformTranslation.READ_TRANSFORM_URN, this::translateRead);
+
     this.urnToTransformTranslator = translatorMap.build();
   }
 
   @Override
   public Set<String> knownUrns() {
-    return urnToTransformTranslator.keySet();
+    // Do not expose Read as a known URN because we only want to support Read
```
tweise: Why this special case? Probably related to the change in GreedyPipelineFuser?

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/io/StreamingImpulseSource.java
```diff
@@ -26,7 +26,10 @@
 /**
  * A streaming source that periodically produces an empty byte array. This is mostly useful for
  * debugging, or for triggering periodic behavior in a portable pipeline.
+ *
+ * @deprecated Legacy non-portable source which can be replaced by a DoFn with timers.
```
tweise: Once timers are verified to be working in real world at scale :)

model/pipeline/src/main/proto/external_transforms.proto
```diff
@@ -0,0 +1,52 @@
...
+ * Protocol Buffers describing the external transforms available.
+ */
+
+syntax = "proto3";
+
+package org.apache.beam.model.pipeline.v1;
+
+option go_package = "pipeline_v1";
+option java_package = "org.apache.beam.model.pipeline.v1";
+option java_outer_classname = "ExternalTransforms";
```
tweise: From a Python SDK perspective, these may be "external" transforms. But in general that depends on which SDK and expansion service / runner is used. Conceivably the same could be implemented in Py, Go or any other language. Maybe "Default Transforms" or "Standard Transforms" would be more appropriate?

model/pipeline/src/main/proto/external_transforms.proto
```diff
@@ -0,0 +1,52 @@
...
+  }
+  oneof elements_per_period {
+    bool elements_per_period_provided = 4;
+    int64 elements_per_period_value = 5;
+  }
+  google.protobuf.Duration period = 6;
+  google.protobuf.Duration max_read_time = 7;
+  // TODO How are foreign functions treated?
+  //SdkFunctionSpec timestamp_fn = 8;
+}
```
tweise: Any other transform like KafkaIO will be added here?

tweise: In the ML discussion, there was the idea of a generic `ExternalTransform` that specific external transforms can expand to. This would reduce the protobuf / expansion service surface.

model/pipeline/src/main/proto/external_transforms.proto
mxm: The name can be refined but we would only use these messages for SDKs which are not written in the transform's language. If the language is the same, we do not have to use the `ExpansionService` to send an `ExpansionRequest`. We can just directly instantiate it.

model/pipeline/src/main/proto/external_transforms.proto
mxm: Yes, the idea is that we would add more messages like this here.

sdks/java/core/src/main/java/org/apache/beam/sdk/io/GenerateSequence.java
```diff
@@ -98,6 +101,24 @@
     abstract GenerateSequence build();
   }
 
+  public static GenerateSequence fromExternal(ExternalTransforms.GenerateSequencePayload payload) {
```
chamikaramj: As mentioned in the doc, I'm bit worried about this extra work on the transform author to make a given transform cross-language compatible. I think we should aim for a model where the transform author simply describe the interface for initiating the transform using a schema. Extra work for building the transform should be generated and performed by the framework.

This is probably OK as a first version though.

model/pipeline/src/main/proto/external_transforms.proto
```diff
@@ -0,0 +1,52 @@
...
+
+option go_package = "pipeline_v1";
+option java_package = "org.apache.beam.model.pipeline.v1";
+option java_outer_classname = "ExternalTransforms";
+
+import "beam_runner_api.proto";
+import "google/protobuf/duration.proto";
+
+// Payload for Java's GenerateSequence transform
+message GenerateSequencePayload {
```
chamikaramj: Can this be some sort of a registration mechanism instead of having to modify the proto for every transform that we want to make cross-language compatible ?

mxm: As long as transforms are bundled with Beam, this should be fine. However, that's a fairly static model and I agree that we want to make them pluggable via some sort of registry.

sdks/java/core/src/main/java/org/apache/beam/sdk/io/GenerateSequence.java
mxm: I share your concerns. Building IO and exposing it to the SDKs should be as simple as possible. Perhaps we keep the Proto-based approach for advanced users, but there should be a easy to use schema-based approach which automatically plugs in the transform and generates Proto messages.

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/graph/GreedyPipelineFuser.java
tweise: That should not be necessary if the transform is declared as native transform: https://github.com/apache/beam/blob/4adb728535719fa93a775f03dc9023556f008e0f/runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkStreamingPortablePipelineTranslator.java#L439

mxm: We can't declare it as native transform because we cannot natively translate Reads containing Python code. I suppose we could check if contains a Java payload in `IsNativeTransform`.

tweise: That or should we use a different URN altogether?

mxm: The problem with that is that it breaks the expansion logic. We return a `Read` from `GenerateSequence` which during pipeline proto creation gets mapped to the Read URN. See `ReadTranslation`. We would have to change something there to generate a different URN.

mxm: I've re-added and extended the check.

mxm:
Rebased to latest master

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/graph/GreedyPipelineFuser.java
```diff
@@ -187,11 +187,19 @@ private DescendantConsumers getRootConsumers(PTransformNode rootNode) {
         rootNode.getId(),
         rootNode.getTransform().getInputsMap());
     checkArgument(
-        !pipeline.getEnvironment(rootNode).isPresent(),
-        "%s requires all root nodes to be runner-implemented %s primitives, "
+        !pipeline.getEnvironment(rootNode).isPresent()
+            // We allow Read transforms as root transforms. The Runner can choose whether it
```
angoenka: I am not super familiar with this piece of code so let me know if this is trivial.
Will this bypass the cross language read as this will execute the transform on Flink natively and not on Java SDKHarness?

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkStreamingPortablePipelineTranslator.java
angoenka: Cross language transforms can be more than just read so giving read special treatment might not be very useful.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/expansion/ExpansionService.java
```diff
@@ -77,7 +78,7 @@
    * @param <InputT> input {@link PValue} type of the transform
    * @param <OutputT> output {@link PValue} type of the transform
    */
-  public interface TransformProvider<InputT extends PValue, OutputT extends PValue> {
+  public interface TransformProvider<InputT extends PInput, OutputT extends PValue> {
```
angoenka: This limit cross language pipelines to sources only. Shall we keep them general?

mxm: `PValue` also is a `PInput`. So this does not limit in any way, it just allows root transforms to be translated.

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/graph/GreedyPipelineFuser.java
mxm: Root transforms are enforced to be environment-free and input-free transforms, i.e. Impulse. Read executes in an environment and is executed by the harness, but in order to support bounded and unbounded reads, we need to translate it natively. Otherwise we are limited to bounded sources.

model/pipeline/src/main/proto/external_transforms.proto
mxm: Initially I had something like this in mind. We discussed pro and cons and I implemented this Proto approach here. We will probably support both ways.

chamikaramj:
Hey Max,

Could you update on the next steps here ?

Are we updating the API to be more generic ? (https://lists.apache.org/thread.html/b99ba8527422e31ec7bb7ad9dc3a6583551ea392ebdc5527b5fb4a67@%3Cdev.beam.apache.org%3E)

mxm:
Hey Cham, I was out of office the past week. Here's what I proposed on the mailing list: 

Changes to the PR:

  1) Remove GenerateSequencePayload
  2) Add generic ExternalTransformPayload, for example:

```
message ExternalTransformPayload {
  enum Enum {
    PROPERTIES = 0
        [(beam_urn) = "beam:external:transform:external_transform:v1"];
  }
  // the format may change to map<string, bytes> if types are supported
  map<string, string> parameters = 1;
}
```

I think we haven't fully figured out the wire format, but we have agreed that it is a list of parameters which use a predefined set of data types. For the scope of the PR, I'll stick with strings.

  3) Define a Pojo in GenerateSequence with the parameters
  4) Add a schema-based Pojo <=> ExternalTransformPayload conversion
  5) Let Python wrappers use ExternalTransformPayload for the request to
     the ExpansionService

chamikaramj:
Thanks for the update!

mxm:
I've pushed an update with the proposed changes. 

Instead of having a dedicated payload for each exposed transform, we use a generic payload (`ExternalConfigurationPayload`). The payload contains a flat map of encoded configuration entries.

Transforms may register a corresponding `ExternalConfigurationBuilder`. The builder is registered together with a URN via a `ExternalTransformRegistrar`. This allows a transform to expose multiple versions of a builder.

The builder is instantiated by the `ExpansionService` when the provided URN is used in an ExternalTransform.

mxm:
Can't currently run the tests on Jenkins due to https://repo.spring.io being down.

```
16:07:02 FAILURE: Build failed with an exception.
16:07:02
16:07:02 * What went wrong:
16:07:02 Execution failed for task ':beam-model-pipeline:compileJava'.
16:07:02 > Could not resolve all files for configuration ':beam-model-pipeline:errorprone'.
16:07:02    > Could not resolve com.google.errorprone:error_prone_core:latest.release.
16:07:02      Required by:
16:07:02          project :beam-model-pipeline
16:07:02       > Failed to list versions for com.google.errorprone:error_prone_core.
16:07:02          > Unable to load Maven meta-data from https://repo.spring.io/plugins-release/com/google/errorprone/error_prone_core/maven-metadata.xml.
16:07:02             > Could not HEAD 'https://repo.spring.io/plugins-release/com/google/errorprone/error_prone_core/maven-metadata.xml'.
16:07:02                > Read timed out
16:07:02
```
https://builds.apache.org/job/beam_PreCommit_Java_Commit/4722/console

mxm:
Retest this please

mxm:
Run Portable_Python PreCommit

mxm:

asfgit:
SUCCESS 
 
--none--

mxm:
@chamikaramj @tweise Let me know what you think of the changes (last commit).

mxm:
Run Python PreCommit

mxm:

sdks/java/core/src/main/java/org/apache/beam/sdk/io/ExternalConfigBuilder.java
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+package org.apache.beam.sdk.io;
+
+import org.apache.beam.sdk.transforms.PTransform;
+import org.apache.beam.sdk.values.PInput;
+import org.apache.beam.sdk.values.POutput;
+
+/** An interface for building a transform from an externally provided configuration. */
+public interface ExternalConfigBuilder<InputT extends PInput, OutputT extends POutput> {
```
chamikaramj: Probably this should be in "sdks/java/core/src/main/java/org/apache/beam/sdk/transforms" given that this solution works for any transform (not just IO).

sdks/java/core/src/main/java/org/apache/beam/sdk/io/GenerateSequence.java
```diff
@@ -98,6 +103,46 @@
...
+
+    @Override
+    public Map<String, Class<? extends ExternalConfigBuilder>> knownBuilders() {
+      return ImmutableMap.of(
+          "beam:external:java:generate_sequence:v1", ExternalConfiguration.class);
+    }
+
+    /** Parameters class to expose the transform to an external SDK. */
+    public static class ExternalConfiguration
+        implements ExternalConfigBuilder<PBegin, PCollection<Long>> {
```
chamikaramj: ExternalConfiguration extending ExternalConfigBuilder looks a bit weird and looks like we'll have to put some effort into keeping transform builder and external configuration in builder for each transform in sync. Instead how about transforms that support cross-language having to directly implement  ExternalConfigBuilder and making ExternalConfiguration POJO a parameter of the build() method ? Hopefully this will encourage developer to keep builders in sync when transforms get updated. Another option will be to use the same builder as the transform when possible.

sdks/java/core/src/main/java/org/apache/beam/sdk/expansion/ExternalTransformRegistrar.java
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+package org.apache.beam.sdk.expansion;
+
+import java.util.Map;
+import org.apache.beam.sdk.io.ExternalConfigBuilder;
+
+/**
+ * A registrar which contains a mapping from URNs to available {@link ExternalConfigBuilder}s.
+ * Should be used with {@link com.google.auto.service.AutoService}.
```
chamikaramj: Just for my knowledge, does AutoService collect builders registered from different classes into the same repository ?

model/pipeline/src/main/proto/external_transforms.proto
```diff
@@ -0,0 +1,42 @@
...
+
+option go_package = "pipeline_v1";
+option java_package = "org.apache.beam.model.pipeline.v1";
+option java_outer_classname = "ExternalTransforms";
+
+import "beam_runner_api.proto";
+
+message ConfigValue {
+  string coder_urn = 1;
+  bytes payload = 2;
```
chamikaramj: Do we also need the name of the parameter (so that we do not have to rely on the ordering of elements of the serialized proto) ?

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkStreamingPortablePipelineTranslator.java
chamikaramj: Read directly from Python or a cross-language read for Java from a Python pipeline will map to the same native read transform, right ? So not sure why we'd want to explicitly disallow this. (also true for any other native transform, for example, GBK)

sdks/python/apache_beam/transforms/external_test.py
```diff
@@ -193,6 +194,20 @@ def test_java_expansion(self):
 
         assert_that(res, equal_to(['a: 3', 'b: 1', 'c: 2']))
 
+      # Test GenerateSequence Java transform
+      options._all_options['shutdown_sources_on_final_watermark'] = True
+      options._all_options['parallelism'] = 1
+      options._all_options['streaming'] = True
+
+      with beam.Pipeline(options=options) as p:
```
chamikaramj: Can we also add an integration test for GenerateSequence (Java transform from a Python pipeline) for Flink ?

chamikaramj:
Also prob. update the title of the PR to reflect that this adds the generalized API for using Java transforms from Python.

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkStreamingPortablePipelineTranslator.java
mxm: In portable pipelines Reads are executed via an environment, i.e. in the SDK harness via ExecutableStage. So they are not native. If we expose the Read URN here, we will have to deal with translating Python and Java Reads natively. However, we can only translate Java Reads natively.

Besides the code here being an optimization to execute without the Java SDK harness, we also want to support unbounded reads which are not supported via the SDK harness.

sdks/java/core/src/main/java/org/apache/beam/sdk/expansion/ExternalTransformRegistrar.java
mxm: AutoService will create META-INF/services files which the `ServiceLoader` can load. This works across modules. When creating a Jar, we need to take care of merging the service files, e.g. `shadowJar` has `mergeServiceFiles()`.

sdks/java/core/src/main/java/org/apache/beam/sdk/io/ExternalConfigBuilder.java
mxm: Good point, will move the class.

sdks/python/apache_beam/transforms/external_test.py
mxm: Yes, that make sense. We have `ReadSourcePortableTest` but that one is not a full integration test.

sdks/java/core/src/main/java/org/apache/beam/sdk/io/GenerateSequence.java
mxm: I'll look into that. We cannot implement `ExternalConfigBuilder` directly on the transform but ideally we would implement it on the existing (non-portable) builder.

mxm:
@chamikaramj Please have another look. I've made the changes you proposed. See the last three commits.

sdks/python/apache_beam/runners/portability/flink_runner_test.py
```diff
@@ -153,6 +156,23 @@ def test_read(self):
     def test_no_subtransform_composite(self):
       raise unittest.SkipTest("BEAM-4781")
 
+    def test_external_transform(self):
+      # options = super(FlinkRunnerTest, self).create_options()
+      # options._all_options['shutdown_sources_on_final_watermark'] = True
+      # options._all_options['parallelism'] = 1
```
chamikaramj: Delete Temporary code ?

sdks/python/apache_beam/runners/portability/flink_runner_test.py
```diff
@@ -153,6 +156,23 @@ def test_read(self):
...
+      # options._all_options['streaming'] = True
+
+      self.create_options()
+      expansion_address = "localhost:" + str(FlinkRunnerTest.expansion_port)
+
+      with self.create_pipeline() as p:
+        res = (
```
chamikaramj: Probably we should make this a post-commit test instead of pre-commit.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/ExternalTransformBuilder.java
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.beam.sdk.transforms;
+
+import org.apache.beam.sdk.values.PInput;
+import org.apache.beam.sdk.values.POutput;
+
+/** An interface for building a transform from an externally provided configuration. */
```
chamikaramj: Add more documentation to explain how this should be used ?

runners/flink/src/test/java/org/apache/beam/runners/flink/ReadSourcePortableTest.java
```diff
@@ -0,0 +1,120 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameter;
+import org.junit.runners.Parameterized.Parameters;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Tests that Read translation is supported in portable pipelines. */
+@RunWith(Parameterized.class)
+public class ReadSourcePortableTest implements Serializable {
```
tweise: Are there no existing Java PVR tests that cover this?

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/expansion/ExpansionService.java
```diff
@@ -69,14 +82,123 @@
...
+    private static void populateConfiguration(
+        Object config, ExternalTransforms.ExternalConfigurationPayload payload) throws Exception {
+      Converter<String, String> camelCaseConverter =
+          CaseFormat.LOWER_UNDERSCORE.converterTo(CaseFormat.LOWER_CAMEL);
+      for (Map.Entry<String, ExternalTransforms.ConfigValue> entry :
+          payload.getConfigurationMap().entrySet()) {
+        String fieldName = camelCaseConverter.convert(entry.getKey());
+        String coderUrn = entry.getValue().getCoderUrn();
+        final Coder coder;
+        if (BeamUrns.getUrn(RunnerApi.StandardCoders.Enum.VARINT).equals(coderUrn)) {
```
robertwb: Shouldn't we be leveraging the coder rehydration code here? Or is that implicit in the TODO below (which should be made more explicit)?

model/pipeline/src/main/proto/external_transforms.proto
```diff
@@ -0,0 +1,42 @@
...
+package org.apache.beam.model.pipeline.v1;
+
+option go_package = "pipeline_v1";
+option java_package = "org.apache.beam.model.pipeline.v1";
+option java_outer_classname = "ExternalTransforms";
+
+import "beam_runner_api.proto";
+
+message ConfigValue {
+  string coder_urn = 1;
```
robertwb: How are we handling composite coders, e.g. a list of ints? Or should this be a coder_id?

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/graph/GreedyPipelineFuser.java
robertwb: So if a Python Read leaked through, this would try and execute it? It does feel a bit odd to augment the fuser to know about Read--better to augment the fuser to accept a list of known native transforms from the runner invoking it.

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkBatchPortablePipelineTranslator.java
```diff
@@ -240,7 +250,12 @@ private FlinkBatchPortablePipelineTranslator(
 
   @Override
   public Set<String> knownUrns() {
-    return urnToTransformTranslator.keySet();
+    // Do not expose Read as a known URN because we only want to support Read
```
robertwb: I'm not quite following this comment.

sdks/python/apache_beam/runners/portability/flink_runner_test.py
mxm: I'm not sure, all of these tests used to be PostCommit, but we moved them to PreCommit to have less PostCommit breakage. It might be good to have it in PreCommit for now.

mxm:
Rebased to latest master. @robertwb Concerning the fuser now allowing (unbounded) Reads as the root node, would it make sense to have a Runner-provided predicate in the fuser to determine whether the Runner can natively translate the Read transform?

mxm:
Thank you for your reviews. I'm merging this to consolidate what we have now. I think we can move in smaller incremental steps from here.

PR #7876, opened 2019-02-18
[BEAM-4775] Clean up metric protos; support integer distributions, gauges

ryan-williams:
(factored out of #7823)

- Drops extrema protos, adds gauges.
- Drops non-integer metric proto types

Also starts tying together `MetricName` and `MonitoringInfoMetricName`:
- the former is used for "user" metrics (defined by {ptransform,namespace,name}), the latter for "system" metrics ({urn,labels})
- user-metrics are also representable in {urn,labels} form, but as of this PR the Java SDK intercepts attempts to create user-metrics that way (in `MonitoringInfoMetricName.create`) and creates a `MetricName` `AutoValue` instance instead, which preserves existing behavior most straightforwardly.

R: @robertwb, @ajamato 

### DIFF
--- a/model/fn-execution/src/main/proto/beam_fn_api.proto
+++ b/model/fn-execution/src/main/proto/beam_fn_api.proto
@@ -501,48 +501,9 @@ message MonitoringInfoTypeUrns {
 message Metric {
   // (Required) The data for this metric.
   oneof data {
-    CounterData counter_data = 1;
-    DistributionData distribution_data = 2;
-    ExtremaData extrema_data = 3;
-  }
-}
-
-// Data associated with a Counter or Gauge metric.
-// This is designed to be compatible with metric collection
-// systems such as DropWizard.
-message CounterData {
-  oneof value {
-    int64 int64_value = 1;
-    double double_value = 2;
-    string string_value = 3;
-  }
-}
-
-// Extrema messages are used for calculating
-// Top-N/Bottom-N metrics.
-message ExtremaData {
-  oneof extrema {
-    IntExtremaData int_extrema_data = 1;
-    DoubleExtremaData double_extrema_data = 2;
-  }
-}
-
-message IntExtremaData {
-  repeated int64 int_values = 1;
-}
-
-message DoubleExtremaData {
-  repeated double double_values = 2;
-}
-
-// Data associated with a distribution metric.
-// This is based off of the current DistributionData metric.
-// This is not a stackdriver or dropwizard compatible
-// style of distribution metric.
-message DistributionData {
-  oneof distribution {
-    IntDistributionData int_distribution_data = 1;
-    DoubleDistributionData double_distribution_data = 2;
+    int64 counter = 1;
+    IntDistributionData distribution = 2;
+    IntGaugeData gauge = 3;
   }
 }
 
@@ -553,11 +514,9 @@ message IntDistributionData {
   int64 max = 4;
 }
 
-message DoubleDistributionData {
-  int64 count = 1;
-  double sum = 2;
-  double min = 3;
-  double max = 4;
+message IntGaugeData {
+  int64 value = 1;
+  int64 timestamp_ms = 2;
 }
 
 // General MonitoredState information which contains
--- a/runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/LabeledMetrics.java
+++ b/runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/LabeledMetrics.java
@@ -19,17 +19,18 @@ package org.apache.beam.runners.core.metrics;
 
 import org.apache.beam.sdk.metrics.Counter;
 import org.apache.beam.sdk.metrics.DelegatingCounter;
+import org.apache.beam.sdk.metrics.MetricName;
 
 /**
  * Define a metric on the current MetricContainer with a specific URN and a set of labels. This is a
  * more convenient way to collect the necessary fields to repackage the metric into a MonitoringInfo
- * proto later. Intended for internal use only (SDK and RunnerHarness developement).
+ * proto later. Intended for internal use only (SDK and RunnerHarness development).
  */
 public class LabeledMetrics {
   /**
    * Create a metric that can be incremented and decremented, and is aggregated by taking the sum.
    */
-  public static Counter counter(MonitoringInfoMetricName metricName) {
+  public static Counter counter(MetricName metricName) {
     return new DelegatingCounter(metricName);
   }
 }
--- a/runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MonitoringInfoMetricName.java
+++ b/runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MonitoringInfoMetricName.java
@@ -17,12 +17,15 @@
  */
 package org.apache.beam.runners.core.metrics;
 
+import static org.apache.beam.runners.core.metrics.SimpleMonitoringInfoBuilder.PCOLLECTION_LABEL;
+import static org.apache.beam.runners.core.metrics.SimpleMonitoringInfoBuilder.USER_COUNTER_URN_PREFIX;
 import static org.apache.beam.vendor.guava.v20_0.com.google.common.base.Preconditions.checkArgument;
 
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
 import javax.annotation.Nullable;
@@ -39,9 +42,9 @@ public class MonitoringInfoMetricName extends MetricName {
   private String urn;
   @Nullable private String name;
   @Nullable private String namespace;
-  private HashMap<String, String> labels = new HashMap<String, String>();
+  private Map<String, String> labels = new HashMap<>();
 
-  private MonitoringInfoMetricName(String urn, HashMap<String, String> labels) {
+  private MonitoringInfoMetricName(String urn, Map<String, String> labels) {
     checkArgument(!Strings.isNullOrEmpty(urn), "MonitoringInfoMetricName urn must be non-empty");
     checkArgument(labels != null, "MonitoringInfoMetricName labels must be non-null");
     // TODO(ajamato): Move SimpleMonitoringInfoBuilder to beam-runner-core-construction-java
@@ -54,7 +57,7 @@ public class MonitoringInfoMetricName extends MetricName {
 
   /** Parse the urn field into a name and namespace field. */
   private void parseUrn() {
-    if (this.urn.startsWith(SimpleMonitoringInfoBuilder.USER_COUNTER_URN_PREFIX)) {
+    if (this.urn.startsWith(USER_COUNTER_URN_PREFIX)) {
       List<String> split = new ArrayList<String>(Arrays.asList(this.getUrn().split(":")));
       this.name = split.get(split.size() - 1);
       this.namespace = split.get(split.size() - 2);
@@ -85,12 +88,26 @@ public class MonitoringInfoMetricName extends MetricName {
   }
 
   /** @return The labels associated with this MonitoringInfo. */
-  public HashMap<String, String> getLabels() {
+  public Map<String, String> getLabels() {
     return this.labels;
   }
 
-  public static MonitoringInfoMetricName named(String urn, HashMap<String, String> labels) {
-    return new MonitoringInfoMetricName(urn, labels);
+  /**
+   * Polymorphic constructor of {@link MetricName}s.
+   *
+   * <p>If `urn` is a {@link SimpleMonitoringInfoBuilder#USER_COUNTER_URN_PREFIX user metric},
+   * return a {@link MetricName} auto-value (which will {@link Object#equals equal} and {@link
+   * Object#hashCode hash} consistently with other {@link MetricName}s.
+   *
+   * <p>Otherwise, return a concrete {@link MonitoringInfoMetricName} representing a "system"
+   * metric.
+   */
+  public static MetricName named(String urn, Map<String, String> labels) {
+    MetricName metricName = MetricUrns.parseUrn(urn);
+    if (metricName == null) {
+      return new MonitoringInfoMetricName(urn, labels);
+    }
+    return metricName;
   }
 
   @Override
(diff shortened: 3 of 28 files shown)


ryan-williams:
Run Java PreCommit

model/fn-execution/src/main/proto/beam_fn_api.proto
```diff
@@ -553,11 +514,9 @@ message IntDistributionData {
   int64 max = 4;
 }
 
-message DoubleDistributionData {
-  int64 count = 1;
-  double sum = 2;
-  double min = 3;
-  double max = 4;
+message IntGaugeData {
```
ajamato: Is there a use case for this at the moment? Are you implementing these right now?

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/LabeledMetrics.java
```diff
@@ -29,7 +30,7 @@
   /**
    * Create a metric that can be incremented and decremented, and is aggregated by taking the sum.
    */
-  public static Counter counter(MonitoringInfoMetricName metricName) {
+  public static Counter counter(MetricName metricName) {
```
ajamato: Let's keep it as MonitoringInfoMetricName, since this is meant to be an implementation only used by SDK and RunnerHarness authors. For creating system style metrics with URN+labels,

model/fn-execution/src/main/proto/beam_fn_api.proto
```diff
@@ -501,48 +501,9 @@ message MonitoringInfoTypeUrns {
...
-
-// Data associated with a distribution metric.
-// This is based off of the current DistributionData metric.
-// This is not a stackdriver or dropwizard compatible
-// style of distribution metric.
-message DistributionData {
-  oneof distribution {
-    IntDistributionData int_distribution_data = 1;
-    DoubleDistributionData double_distribution_data = 2;
+    int64 counter = 1;
```
ajamato: There was long debate when, and the existing protos were what we could agree on when we introduced this design. While I am fine with this change you are proposing personally. I'd like you to get buy in on the dev list.

https://s.apache.org/beam-fn-api-metrics

https://lists.apache.org/list.html?dev@beam.apache.org:2018-2

I think most of it was in the comments of the doc. So unfortunately, I cannot point you toward too much.

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MonitoringInfoMetricName.java
```diff
@@ -113,11 +136,15 @@ public boolean equals(Object o) {
...
-  public String toString() {
+  public String toString(String delimiter) {
+    if (getNamespace() != null && getName() != null) {
+      return super.toString(delimiter);
+    }
     StringBuilder builder = new StringBuilder();
-    builder.append(this.urn.toString());
-    builder.append(" ");
-    builder.append(this.labels.toString());
+    if (labels.containsKey(PCOLLECTION_LABEL)) {
```
ajamato: just iterate over the keyas and values and add them to the return string, instead of inspecting the labels for specific keys.

Then there is no need to update this code for new labels.

runners/flink/src/main/java/org/apache/beam/runners/flink/metrics/FlinkMetricContainer.java
```diff
@@ -94,121 +93,128 @@ public MetricsContainer getMetricsContainer(String stepName) {
...
+        : null;
+  }
+
   /**
    * Update this container with metrics from the passed {@link MonitoringInfo}s, and send updates
    * along to Flink's internal metrics framework.
    */
-  public void updateMetrics(String stepName, List<MonitoringInfo> monitoringInfos) {
-    MetricsContainer metricsContainer = getMetricsContainer(stepName);
+  public void updateMetrics(List<MonitoringInfo> monitoringInfos) {
```
ajamato: Question about the overall design here. Much of this class doesn't seem flink specific. If its just aggregating the values in memory, perhaps it should just aggregate the MonitoringInfos. Then have some method to extract them in a flink specific format (possibly in another class)

runners/flink/src/main/java/org/apache/beam/runners/flink/metrics/FlinkMetricContainer.java
```diff
@@ -94,121 +93,128 @@ public MetricsContainer getMetricsContainer(String stepName) {
...
-                    intDistributionData.getMin(),
-                    intDistributionData.getMax());
-              } else {
-                LOG.warn("Unsupported DistributionData type: {}", distributionData);
-              }
-            } else if (metric.hasExtremaData()) {
-              ExtremaData extremaData = metric.getExtremaData();
-              LOG.warn("Extrema metric unsupported: {}", extremaData);
-            }
+          if (!monitoringInfo.hasMetric()) {
```
ajamato: Whitelisting the URNs you support would be the preferred method of doing this.

What you've done here will allow various SDK system metrics which get added later on, without the devs of this runner getting any sort of heads up. The intention of the design is to allow the runner to whitelist the specific URNs it wishes to support. You may just wish to support the user metric prefixes. Doing that would drop the element count, execution time, etc. metrics.

ajamato:
@Ardagan You might want to take a look at this as well. To see the proto changes.

model/fn-execution/src/main/proto/beam_fn_api.proto
Ardagan: I don't have strong opinion about removing CounterData, but I'm concerned that we make counter int64. In my opinion, double will be much more flexible.

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MonitoringInfoMetricName.java
ryan-williams: I wrote things that way in several places initially, but ran into interesting problems that we'll need to discuss eventually, so I'll describe them below.

Here's what happens in this PR if I naively append the label kv-pairs: https://github.com/ryan-williams/beam/commit/557628627ae7438a6b688514e81205594238ed0f. 

Note that Flink receives a metric with the unseemly name [`step.PTRANSFORM.step.PCOLLECTION.pcoll.beam.metric.element_count.v1`](https://github.com/ryan-williams/beam/commit/557628627ae7438a6b688514e81205594238ed0f#diff-682be67e606ae40ded3e0087df5f9f8bR136) (which goes e.g. to the Flink web UI, which we've taken pains to make more human-readable).

The first `step` there is added by `MetricKey.toString`, which calls this `MetricName.toString` routine, expecting info about the metric's "name" (URN) only.

## MetricKey vs MetricName

The issue, In my view, is that `MonitoringInfoMetricName` is actually more of a `MetricKey` implementation:
- `MetricKey` ≈ `MonitoringInfoMetricName` ({URN, labels})
  - `stepName` ≈ [labels map with just a `PTRANSFORM` entry]
  - `MetricName` ≈ URN
- `MetricResult` ≈ `MonitoringInfo` ({key, metric value})

I ultimately lay things out in this manner in #7823, folding MIMN into MetricKey.

## Special-casing single-label ptransform/pcollection-scoped metrics

A bigger issue is that I think runners and endpoints are going to want to hard-code [the only 2 cases that we will have for the foreseeable future]: `<ptransform>.<namespace>.<name>` and `<pcollection>.<urn>` (`<ptransform>.<urn>` folds into this nicely as well).

They'll do this for backwards-compat reasons, and/or to make things look nice in e.g. web UIs. I've touched code like this in at least the Samza, Flink, and Spark runners (and that `MetricsHttpSink` we were looking at).

I think it's great that the wire format is extensible to many+new labels, but ended up feeling that the SDK should provide the APIs that deal specifically with the shapes all calling code will care about atm.

Sorry that's a lot! I don't care much what we do in this case, but I think we'll have some harder similar choices shortly 😲 lmk what you think.

runners/flink/src/main/java/org/apache/beam/runners/flink/metrics/FlinkMetricContainer.java
ryan-williams: Interesting point. To paraphrase: ingesting MonitoringInfos into a MetricContainerStepMap (which other runners also use) will be a common path, and code for that should get factored out.

I experimented with keeping fn-api metrics natively in MonitoringInfos in MCSM, alongside traditional Java-SDK-formatted ones, but ended up needing to do a lot of other cleanup in there (#7890) before I could see straight 😄.

runners/flink/src/main/java/org/apache/beam/runners/flink/metrics/FlinkMetricContainer.java
ryan-williams: Interesting. It seems in this case like it's easier and better to just handle all the metrics that come through conforming to the standard format? Or maybe I'm not understanding what you mean.

model/fn-execution/src/main/proto/beam_fn_api.proto
robertwb: I don't remember discussing this, it must have been with someone else. 

I agree that supporting double counters (and especially distributions) would be useful. Also, we haven't really nailed down what the global semantics are for gauge metrics. 

My position was, and remains, that reifying the exact set of supported metric types via a set of nested set of protos is not a stable, long-term solution. It took a long time to come up with this list, and here we are 6 months later changing it. (And change will become harder the more we actually use these protos everywhere.) Instead, data should be either (1) a bytes field or (2) something like

message data {
  repeated int64 int_data = 1;
  repeated double double_data = 2;
  repeated string string_data = 3;
  ...?
}

with the type specifying its encoding into this data attribute. We can then support everything from simple counters to complex distributions without changing the protos, only introducing new types, where the set of types are defined by URNs, and runners simply pass through types they do not understand. This is consistent with the rest of the way the FnAPI works as well (e.g. coders, transforms, ...)

ryan-williams: Thanks Robert. It seems we're stuck between 3 data models:

1. [existing protos](https://github.com/apache/beam/blob/981fb620a1b53d2a564b629f1d1e2d18184f8393/model/fn-execution/src/main/proto/beam_fn_api.proto#L501-L561)
    - {int,double,string} counters
    - {int,double} x {distribution,extrema}
    - unused in SDKs
2. [deprecated protos](https://github.com/apache/beam/blob/981fb620a1b53d2a564b629f1d1e2d18184f8393/model/fn-execution/src/main/proto/beam_fn_api.proto#L687-L714)
    - `int64`-only {counter, distribution, gauge}
    - supported in [Java](https://github.com/apache/beam/blob/981fb620a1b53d2a564b629f1d1e2d18184f8393/sdks/java/core/src/main/java/org/apache/beam/sdk/metrics/MetricQueryResults.java#L32-L39), [Python](https://github.com/apache/beam/blob/981fb620a1b53d2a564b629f1d1e2d18184f8393/sdks/python/apache_beam/metrics/cells.py) SDKs
3. opaque bytes-payloads
    - [Robert's suggestion above](https://github.com/apache/beam/pull/7876#discussion_r258384748)
    - not implemented anywhere afaik

In this PR, I changed (1)'s protos to match the SDK support that is modeled on (2), which is probably wrong!

(3) sgtm but seems like it is not the existing consensus / spec.

I think the larger goal (sending metrics over the job API) requires converting between SDK and proto metrics structures (in both directions), so they need to match.

Maybe there is a way I can side-step that, or maybe we can just get to the real long-term solution here.

model/fn-execution/src/main/proto/beam_fn_api.proto
```diff
@@ -553,11 +514,9 @@ message IntDistributionData {
...
 }
 
-message DoubleDistributionData {
-  int64 count = 1;
-  double sum = 2;
-  double min = 3;
-  double max = 4;
+message IntGaugeData {
+  int64 value = 1;
+  int64 timestamp_ms = 2;
```
ajamato: Can we keep this on the top level of the MonitoringInfo

model/fn-execution/src/main/proto/beam_fn_api.proto
ajamato: @robertwb I'd like to push back a bit about a significant change, where we use opaque bytes payloads.

I am okay with removing the layers in this proto and using the design Ryan has proposed here. But not an extensive rewrite at this stage of the game, which will slow down progress significantly. 

We have protos for the MonitoringInfoTable. This solution is the extensible format, where you can store basically anything and the producer and consumer of this proto can be the only ones who need to understand it (and it could be passed through the RunnerHarness)
https://github.com/apache/beam/blob/197a06852d27c5baea6a4c65894a8841672af7b6/model/fn-execution/src/main/proto/beam_fn_api.proto#L583

But we chose to have separate protos for the common well known metric formats (Metrics being a well defined concept shared by many systems, i.e. a timeseries of data): counter, gauge, distribution, etc.

Additionally, if you wish to modify these formats significantly at a future stage, you can always 
- add to these protos without deleting
- Change the URN version number of a metric
- Upgrade all SDKs to package the metric with a format for the previous and new version number of the URN
- Upgrade all Runners that wish to support it.

Upgrading the URN version is the safe way to do this without introducing a breaking change, as consumers of MonitoringInfos can freely chose which URNs+versions they support.

model/fn-execution/src/main/proto/beam_fn_api.proto
```diff
@@ -501,48 +501,9 @@ message MonitoringInfoTypeUrns {
...
-// This is based off of the current DistributionData metric.
-// This is not a stackdriver or dropwizard compatible
-// style of distribution metric.
-message DistributionData {
-  oneof distribution {
-    IntDistributionData int_distribution_data = 1;
-    DoubleDistributionData double_distribution_data = 2;
+    int64 counter = 1;
+    IntDistributionData distribution = 2;
+    IntGaugeData gauge = 3;
```
ajamato: The one thing worth pointing out here, is that we never introduced this before because this only holds an int64, which is the same as a counter, which holds an int64 (Note you don't need the timestamp inside IntGaugeData), as its already in the MonitoringInfo proto.

So we didn't include both as separate fields in the oneof. You would just specify which way it is aggregated using the MonitoringInfo type field.

ryan-williams:
Thanks for the feedback, these discussions were very useful for me! 

As I mentioned [here](https://github.com/apache/beam/pull/7876#discussion_r258585934), these proto changes are basically backwards from the direction they are supposed to be evolving in.

I'm going to close this out; I think I see how to move the other #7823-associated PRs forward orthogonally, and probably more simply for not needing to deal with new proto structures.

PR #7889, opened 2019-02-19
[BEAM-5381] Fix two issues with generating the pipeline plan for Go SDK

robinp:
- Wrong id was returned for CoGBK expansions, so the expanded transform didn't end up in the subtransform of the parent Scope.
- CoGBK inject node outputs need to include edge name as well, not just input node name, otherwise there's an output name collision (caught by Flink runner's consistency check).

R: @herohde 
CC: @schroederc

### DIFF
--- a/sdks/go/pkg/beam/core/runtime/graphx/translate.go
+++ b/sdks/go/pkg/beam/core/runtime/graphx/translate.go
@@ -336,7 +336,7 @@ func (m *marshaller) expandCoGBK(edge NamedEdge) string {
 	for i, in := range edge.Edge.Input {
 		m.addNode(in.From)
 
-		out := fmt.Sprintf("%v_inject%v", nodeID(in.From), i)
+		out := fmt.Sprintf("%v_%v_inject%v", nodeID(in.From), id, i)
 		m.makeNode(out, kvCoderID, in.From)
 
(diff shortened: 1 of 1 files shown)


aaltay:
R: @lostluck

lostluck:
Run Go Postcommit

lostluck:
Nice catch on the transform scoping! It worked since it was still pointing at the root inject transform and the graph continued from there. I agree that the consistency checker is correct as well.

Consider adding (or amending) the integration test cases to prevent regressions in the future: https://github.com/apache/beam/tree/master/sdks/go/test/integration 
This is making the assumption you have a pipeline shape that you were using when you discovered these errors.

lostluck:
A unit test would be fine and probably preferable instead.

robinp:
@lostluck would add unit test, but until I get some feedback on https://jira.apache.org/jira/browse/BEAM-6724 (which is the next blocker after resolving this), I don't have incentive to work on this bug. If you see what I mean :P

lostluck:
May we close this PR? It's been superceded by the Merged PR https://github.com/apache/beam/pull/8238, which has the same implementation, just with the extra test.

robinp:
Sure, closing! Thanks for the merge!

PR #7915, opened 2019-02-21
Use MonitoringInfo data model in Java SDK metrics

ryan-williams:
- fold `MonitoringInfoMetricName` into `MetricKey` ({URN, labels})
  - `MetricKey.stepName` is expanded to a `MetricLabels` map
  - `MetricContainerStepMap` is now keyed off `MetricLabels` 
  - (existing use-cases just get a `PTRANSFORM` entry with the `stepName`)
- `MetricName` is basically a MonitoringInfo URN
  - previously only modeled user metrics ({namespace,name})
  - can now model system-metric URNs as well
- `MetricResult` is like a full `MonitoringInfo` ({key, metric value})
  - it also contains both "attempted" and "committed", which MonitoringInfo does not, but I'm addressing that elsewhere

R: @ajamato, @Ardagan, @robertwb 

## java/core → model-fn-execution dependency
I waffled on #7867 due to making java/core depend on the `model-fn-execution` proto module. However, talking with @ajamato, it seems like we're ok with that. 

It *may* be possible to pull more java/core metrics classes (Metric{Key,Name,Labels,Container*}, {Counter,Distribution,Gauge}Cell, and maybe more) into runners-core/java, which already depends on the model-fn-execution protos module, and save adding the dep to java/core. Let me know if I should look into that.

If not, someone should sign off on in particular on needing to add `org.apache.beam.model` and vendored gRPC to [this GCP module API surface test](https://github.com/apache/beam/compare/master...ryan-williams:mi-mk?expand=1#diff-e1f378c05ed380fffbfb99e0f819afabR59).


sdks/java/core/src/main/java/org/apache/beam/sdk/metrics/MetricName.java
```diff
@@ -17,63 +17,102 @@
...
+
+  /** Parse the urn field into a name and namespace field. */
+  private boolean parseUrn(boolean raise) {
+    if (namespace != null && name != null) {
+      return true;
+    }
+    String urn = urn();
+    if (urn.startsWith(USER_COUNTER_URN_PREFIX)) {
+      urn = urn.substring(USER_COUNTER_URN_PREFIX.length());
+    } else {
```
Ardagan: if (!urn.startsWith(USER_COUNTER_URN_PREFIX)) {
// raise or return
}
urn = urn.substring(...)

sdks/java/core/src/main/java/org/apache/beam/sdk/metrics/MetricName.java
```diff
@@ -17,63 +17,102 @@
...
+      if (raise) {
+        throw new IllegalStateException(
+            "Attempting to access namespace/name of a non-user metric: " + urn);
+      }
+      return false;
+    }
+    // If it is not a user counter, just use the first part of the URN, i.e. 'beam'
+    List<String> pieces = Splitter.on(':').splitToList(urn);
+    if (pieces.size() != 2) {
```
Ardagan: @ajamato 
I believe we [allow to have colons in namespace](https://github.com/apache/beam/blob/fb425d5c4a7607d78f879e790ddebdc7bb5ea68a/runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/fn/control/UserMonitoringInfoToCounterUpdateTransformer.java#L106).

runners/flink/src/main/java/org/apache/beam/runners/flink/metrics/FlinkMetricContainer.java
```diff
@@ -89,127 +93,138 @@ public FlinkMetricContainer(RuntimeContext runtimeContext) {
...
+
   /**
    * Update this container with metrics from the passed {@link MonitoringInfo}s, and send updates
    * along to Flink's internal metrics framework.
    */
   public void updateMetrics(String stepName, List<MonitoringInfo> monitoringInfos) {
-    MetricsContainer metricsContainer = getMetricsContainer(stepName);
     monitoringInfos.forEach(
         monitoringInfo -> {
-          if (monitoringInfo.hasMetric()) {
```
ryan-williams: A lot of the Flink-related changes here have since been factored out to #7971, I'll try to get that merged, and in the meantime set up a view of this PR rebased on top of that.

sdks/java/core/src/main/java/org/apache/beam/sdk/metrics/MetricName.java
```diff
@@ -17,63 +17,102 @@
...
+      urn = urn.substring(USER_COUNTER_URN_PREFIX.length());
+    } else {
+      if (raise) {
+        throw new IllegalStateException(
+            "Attempting to access namespace/name of a non-user metric: " + urn);
+      }
+      return false;
+    }
+    // If it is not a user counter, just use the first part of the URN, i.e. 'beam'
+    List<String> pieces = Splitter.on(':').splitToList(urn);
```
Ardagan: We got duplication of this logic:
in this PR: MetricUrns.java L66
and in [DF runner](https://github.com/apache/beam/blob/fb425d5c4a7607d78f879e790ddebdc7bb5ea68a/runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/fn/control/UserMonitoringInfoToCounterUpdateTransformer.java#L106)

can be more.

sdks/java/core/src/main/java/org/apache/beam/sdk/metrics/MetricName.java
ajamato: Yes this might be possible. Use the last part of the split as the name, then combine everything else.

Ryan is going to move this to the labels later on, getting rid of all this parsing code anyways though

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/SimpleMonitoringInfoBuilder.java
```diff
@@ -137,17 +115,17 @@ public SimpleMonitoringInfoBuilder setUrn(String urn) {
...
 
-  /**
-   * Sets the urn of the MonitoringInfo to a proper user metric URN for the given params.
-   *
-   * @param namespace
-   * @param name
-   */
-  public SimpleMonitoringInfoBuilder setUrnForUserMetric(String namespace, String name) {
-    this.builder.setUrn(userMetricUrn(namespace, name));
+  public SimpleMonitoringInfoBuilder handleMetricKey(MetricKey key) {
```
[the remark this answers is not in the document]
ryan-williams: ah, yea [I'd renamed this to `setLabelsAndUrn`](https://github.com/apache/beam/pull/7934/files#diff-fcdf8e2f6e21b45ab9096ef5f2ef4391R151) on #7934 for similar reasons; just did the same here

runners/flink/src/main/java/org/apache/beam/runners/flink/metrics/FlinkMetricContainer.java
```diff
@@ -89,127 +93,138 @@ public FlinkMetricContainer(RuntimeContext runtimeContext) {
...
+      metric = create.apply(flinkMetricName, runtimeContext.getMetricGroup(), value);
+      flinkMetricMap.put(flinkMetricName, metric);
     }
+    update.accept(metric, value);
   }
 
-  private void updateGauge(Iterable<MetricResult<GaugeResult>> gauges) {
-    for (MetricResult<GaugeResult> metricResult : gauges) {
-      String flinkMetricName = getFlinkMetricNameString(metricResult.getKey());
+  private void updateCounter(MetricKey metricKey, long attempted) {
```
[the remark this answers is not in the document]
ryan-williams: done

sdks/java/core/src/main/java/org/apache/beam/sdk/metrics/MetricName.java
ryan-williams: done

sdks/java/core/src/main/java/org/apache/beam/sdk/metrics/MetricName.java
ryan-williams: Thanks for flagging this!

I've changed this code (and [`MetricUrns.parseUrn`](https://github.com/apache/beam/pull/7915/files#diff-2a6c540e8bf410b089793ea6efadb291R61)) to match the semantics there ("namespace is everything up to the last `:`, with internal `:` deleted"), and [changed that site](https://github.com/apache/beam/pull/7915/files#diff-a2189a7c97dace5f44f04e16a71b6468R108) to call `MetricUrns.parseUrn`.

Combining this with MetricUrns.parseUrn would also be nice, but they're pretty different contexts and I don't see an easy way to atm.

sdks/java/core/src/main/java/org/apache/beam/sdk/metrics/MetricName.java
ryan-williams: Cool, I changed my handling to match DF, as I mentioned above, thanks.

ryan-williams:
Run Java PreCommit

ajamato:
R: @pabloem

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MetricsContainerStepMap.java
```diff
@@ -32,55 +30,53 @@
...
   /** Returns the container for the given step name. */
-  public MetricsContainerImpl getContainer(String stepName) {
-    if (stepName == null) {
-      // TODO(BEAM-6538): Disallow this in the future, some tests rely on an empty step name today.
-      return getUnboundContainer();
-    }
-    if (!metricsContainers.containsKey(stepName)) {
-      metricsContainers.put(stepName, new MetricsContainerImpl(stepName));
+  public MetricsContainerImpl getContainer(MetricLabels labels) {
+    if (!metricsContainers.containsKey(labels)) {
```
pabloem: It doesnt seem like `MetricLabels` implements a `hashCode` or an `equals` function. I may be missing something, but perhaps we need a test that checks `getContainer(MetricLabels.ptransform("A")).equals(MetricLabels.ptransform("A"))` ?

runners/flink/src/main/java/org/apache/beam/runners/flink/metrics/FlinkMetricContainer.java
```diff
@@ -89,127 +93,138 @@ public FlinkMetricContainer(RuntimeContext runtimeContext) {
...
+          if (!monitoringInfo.hasMetric()) {
+            LOG.info("Skipping metric-less MonitoringInfo: {}", monitoringInfo);
+            return;
+          }
+          Metric metric = monitoringInfo.getMetric();
+
+          MetricKey key = MetricKey.create(monitoringInfo.getUrn(), monitoringInfo.getLabelsMap());
+          MetricName metricName = key.metricName();
+
+          MetricsContainer container = getMetricsContainer(key.labels());
```
pabloem: For my clarification: Does this mean that a user metric can only have as label its PTransform?

ryan-williams:
I don't think this is the approach we're intending to go with. I'm putting more details in a ML post shortly.

PR #7921, opened 2019-02-22
Make data size a criteria for batches when writing with bigqueryio.

danvietduong:
- Respecting bigquery's 10MB API limit
- Making errors from bigquery API more debuggable

------------------------

### DIFF
--- a/sdks/go/pkg/beam/io/bigqueryio/bigquery.go
+++ b/sdks/go/pkg/beam/io/bigqueryio/bigquery.go
@@ -187,8 +187,8 @@ type writeFn struct {
 	Type beam.EncodedType `json:"type"`
 }
 
-// Approximate the size of an element to not exceed the bigquery API limit.
-func getSize(v interface{}) (int, error) {
+// Approximate the size of an element as it would appear in a BQ insert request.
+func getInsertSize(v interface{}) (int, error) {
 	schema, err := bigquery.InferSchema(v)
 	if (err != nil) {
 		return 0, err
@@ -243,10 +243,12 @@ func (f *writeFn) ProcessElement(ctx context.Context, _ int, iter func(*beam.X) 
 	}
 
 	var data []reflect.Value
-	var size = 0
+	// This stores the running byte size estimate of a BQ request. Initializing to 1KB
+	// as an estimate for overall message overhead.
+	var size = 1024
 	var val beam.X
 	for iter(&val) {
-		current, err := getSize(val.(interface{}))
+		current, err := getInsertSize(val.(interface{}))
 		if err != nil {
 			return fmt.Errorf("biquery write error: %v", err)
 		}
--- a/sdks/go/pkg/beam/io/bigqueryio/bigquery.go
+++ b/sdks/go/pkg/beam/io/bigqueryio/bigquery.go
@@ -245,7 +245,7 @@ func (f *writeFn) ProcessElement(ctx context.Context, _ int, iter func(*beam.X) 
 		if len(data) + 1 > writeRowLimit || size + current > writeSizeLimit {
 			// Write rows in batches to comply with BQ limits.
 			if err := put(ctx, table, f.Type.T, data); err != nil {
-				return fmt.Errorf("write error: %d : %v",  size, err)
+				return err
 			}
 			data = nil
 			size = 0
--- a/sdks/go/pkg/beam/io/bigqueryio/bigquery.go
+++ b/sdks/go/pkg/beam/io/bigqueryio/bigquery.go
@@ -28,6 +28,7 @@ import (
 	"cloud.google.com/go/bigquery"
 	"github.com/apache/beam/sdks/go/pkg/beam"
 	"github.com/apache/beam/sdks/go/pkg/beam/core/util/reflectx"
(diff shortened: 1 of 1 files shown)


danvietduong:
R: @lostluck

danvietduong:
Hmmm...  not quite ready. Either I'm not calculating the size right or there is a lot of additional overhead for sending this data.

danvietduong:
OK. I've updated the size estimation a little bit. I'm likely still undercounting, but having some size constraint seems strictly better than none.

sdks/go/pkg/beam/io/bigqueryio/bigquery.go
```diff
@@ -28,12 +28,15 @@ import (
 	"cloud.google.com/go/bigquery"
 	"github.com/apache/beam/sdks/go/pkg/beam"
 	"github.com/apache/beam/sdks/go/pkg/beam/core/util/reflectx"
+	bq "google.golang.org/api/bigquery/v2"
```
lostluck: Is it safe to combine the cloud.google.com API with the V2 API?

Be warned: It's not currently simple to add new Go dependencies, due to gradle and jenkins etc, the lock file will need updating if the v2 package isn't already getting vendored.
There's a low priority task to move to Go Modules instead which would make it easier for everyone involved, but that's a hairy mess WRT beam right now. I can provide more details if necessary.

sdks/go/pkg/beam/io/bigqueryio/bigquery.go
```diff
@@ -198,33 +228,46 @@ func (f *writeFn) ProcessElement(ctx context.Context, _ int, iter func(*beam.X)
...
-		if err := table.Create(ctx, &bigquery.TableMetadata{Schema: mustInferSchema(f.Type.T)}); err != nil {
+		if err := table.Create(ctx, &bigquery.TableMetadata{Schema: schema}); err != nil {
 			return err
 		}
 	}
 
 	var data []reflect.Value
+	// This stores the running byte size estimate of a BQ request. Initializing to 1KB
+	// as an estimate for overall message overhead.
+	var size = 1024
```
lostluck: Please move the 1024 to a constant since it's being used in two places.

sdks/go/pkg/beam/io/bigqueryio/bigquery.go
```diff
@@ -184,6 +187,33 @@ type writeFn struct {
 	Type beam.EncodedType `json:"type"`
 }
 
+// Approximate the size of an element as it would appear in a BQ insert request.
+func getInsertSize(v interface{}, schema bigquery.Schema) (int, error) {
```
lostluck: It seems a little wasteful to encode the data, throw that info away, and then allow the BQ client to encode the data again, but this is certainly precise.

I imagine the reflect approach you had would have worked fine, if things like integers getting converted to strings were accounted for better, and as well as accounting for field names, among other bits of overhead.

sdks/go/pkg/beam/io/bigqueryio/bigquery.go
danvietduong: Internally the cloud api includes v2 api, so I would assume it would be safe, but honestly, this is beyond my depth of the overall build system.

sdks/go/pkg/beam/io/bigqueryio/bigquery.go
danvietduong: Yeah, this was the best I could think of. Do you prefer going down the reflect path? It feels there are more opportunities to get things wrong in that path.

sdks/go/pkg/beam/io/bigqueryio/bigquery.go
danvietduong: Done.

danvietduong:
Thanks for the comments! With regards to the size computation, I'd really like your recommendation on how to proceed. This is the best I could think of.

sdks/go/pkg/beam/io/bigqueryio/bigquery.go
lostluck: It's a precise strategy, so I'm not against it.
The IOs in the Go SDK are not in a great state to begin with, so this is a welcome improvement.

I'd much rather have something that's a bit wasteful but works, and can be improved later, than block on trying to get it perfect.

sdks/go/pkg/beam/io/bigqueryio/bigquery.go
lostluck: I've almost got the right grasp on it to fix it, but we can give it a go.
If Jenkins starts failing, we can revert the PR and re-evaluate.

lostluck:
Go Run PostCommit

lostluck:
Run Go Postcommit

danvietduong:
Anything you need me to do about the post commit failure? It seems like master is having the same problem.

lostluck:
That's a wholely unrelated issue under separate investigation. Apparently providing the same input in the same order to MD5 hashing can lead to different results somehow.

LGTM to me.

@aaltay Please merge. Thanks!

lostluck:
Run Go Postcommit

PR #7922, opened 2019-02-22
[BEAM-6819] Report estimated size as source metadata when splitting.

pedapudi:
Currently, sources don't report SourceMetadata after successful splits; the serialization to a cloud Source omits metadata all together. This prevents Cloud Dataflow from understanding what the size of the resulting splits might be. Especially when subsources are concatenated together (per the split protocol), this results in potentially ambiguous scenarios where it's not feasible for Cloud Dataflow to infer the appropriate fraction of the input represented by a subsource. To help with this, this change populates the estimated size per source. To minimize the expense of calling getEstimatedSizeBytes per subsource, this change also modifies AvroSource to be built with single file metadata which allows it to efficiently compute its estimated size. Moreover, BigQuerySourceBase now leverages constructing Avro subsources using single file metadata.

### DIFF
--- a/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQuerySourceBase.java
+++ b/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQuerySourceBase.java
@@ -17,6 +17,7 @@
  */
 package org.apache.beam.sdk.io.gcp.bigquery;
 
+import static org.apache.beam.sdk.io.FileSystems.match;
 import static org.apache.beam.sdk.io.gcp.bigquery.BigQueryHelpers.createJobIdToken;
 import static org.apache.beam.sdk.io.gcp.bigquery.BigQueryHelpers.getExtractJobId;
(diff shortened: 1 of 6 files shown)


pedapudi:
R: @chamikaramj

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQuerySourceBase.java
```diff
@@ -221,9 +242,18 @@ public T apply(GenericRecord input) {
...
           }
         };
+
     List<BoundedSource<T>> avroSources = Lists.newArrayList();
-    for (ResourceId file : files) {
-      avroSources.add(AvroSource.from(file.toString()).withParseFn(fnWrapper, getOutputCoder()));
+    // If metadata is available, create AvroSources with said metadata in SINGLE_FILE_OR_SUBRANGE
+    // mode.
+    if (metadata != null) {
+      for (MatchResult.Metadata file : metadata) {
```
chamikaramj: nit: change variable name to metadata

chamikaramj:
Run Java PostCommit

PR #7934, opened 2019-02-23
[BEAM-4775] add GetJobMetrics RPC

ryan-williams:
- Send metrics over the job API
  - new job API RPC depends on `MonitoringInfo` protos from the fn api
    - per discussion with @robertwb, I'm going to move `MonitoringInfo` into `model/pipeline`, to avoid introducing a [job API] → [fn API] dep
    - as such, **this PR shouldn't be merged as-is**; that proto move should happen first, but I wanted to get this out for Jenkins and first review pass
- add Python and Java SDK support for portable metrics over the job API
  - portable python flink wordcount outputs the integer `empty_lines` metric!
  - Flink attempted counter PVR cases are now enabled

R: @robertwb @ajamato 
CC: @mxm since I reworked the Flink metrics a bit in particular


ryan-williams:

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MetricUpdatesProtos.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.List;
+import java.util.function.BiConsumer;
+import org.apache.beam.model.fnexecution.v1.BeamFnApi.MonitoringInfo;
+import org.apache.beam.runners.core.construction.metrics.MetricKey;
+
+/** Convert {@link MetricUpdates} to a {@link List} of {@link MonitoringInfo}. */
+public class MetricUpdatesProtos {
+  public static List<MonitoringInfo> toProto(MetricUpdates metricUpdates) {
+    ArrayList<MonitoringInfo> monitoringInfos = new ArrayList<>();
+
```
ajamato: Please share code with 

https://github.com/apache/beam/blob/5b1700db834cf949547a20a30f3ee1270710585a/runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MetricsContainerImpl.java#L147

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MetricsContainerImpl.java
```diff
@@ -139,53 +136,6 @@ public MetricUpdates getUpdates() {
...
-      }
-      builder.setPTransformLabel(metricUpdate.getKey().stepName());
-    }
-    builder.setInt64Value(metricUpdate.getUpdate());
-    builder.setTimestampToNow();
-    return builder.build();
-  }
-
-  /** Return the cumulative values for any metrics in this container as MonitoringInfos. */
-  public Iterable<MonitoringInfo> getMonitoringInfos() {
```
ajamato: Please keep this method, just have it call the counter update conversion here, instead of making all the call sites do it.

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MetricResultsProtos.java
```diff
@@ -0,0 +1,235 @@
...
+    /**
+     * Populate metric-type-specific maps with {@link MonitoringInfo}s from a {@link
+     * BeamFnApi.MetricResults}.
+     */
+    public MetricResultsBuilder(BeamFnApi.MetricResults metrics) {
+      add(metrics.getAttemptedList(), false);
+      add(metrics.getCommittedList(), true);
+    }
+
+    private void add(Iterable<MonitoringInfo> monitoringInfos, Boolean committed) {
```
ajamato: return this to implement builder patern

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MonitoringInfoMetricName.java
```diff
@@ -85,12 +87,33 @@ public String getUrn() {
...
+   *
+   * <p>If `urn` is a {@link SimpleMonitoringInfoBuilder#USER_COUNTER_URN_PREFIX metric}, return a
+   * {@link MetricName} auto-value (which will {@link Object#equals equal} and {@link
+   * Object#hashCode hash} consistently with other {@link MetricName}s.
+   *
+   * <p>Otherwise, return a concrete {@link MonitoringInfoMetricName} representing a "system"
+   * metric.
+   */
+  public static MetricName named(String urn, Map<String, String> labels) {
+    checkArgument(!Strings.isNullOrEmpty(urn), "MonitoringInfoMetricName urn must be non-empty");
```
ajamato: I can't think of a reason that this will ever be called with a user beam URN that needs to be parsed. Are you using this somewhere? Otherwise I don't see a reason to add this.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/InMemoryJobService.java
```diff
@@ -330,6 +335,29 @@ public void describePipelineOptions(
...
 
+  @Override
+  public void getJobMetrics(
+      GetJobMetricsRequest request, StreamObserver<GetJobMetricsResponse> responseObserver) {
+    String invocationId = request.getJobId();
+    LOG.info("Running getJobMetrics for {}", invocationId);
+    try {
+      JobInvocation invocation = getInvocation(invocationId);
+      LOG.info("Found job invocation for metrics: {}", invocation);
+      MetricResults metrics = invocation.getMetrics();
```
ajamato: This code is converting MetricResults -> MonitoringInfos -> MetricResults

We really just need the MonitoringInfos to put on the GetJobMetricResponse.
Let's make the interface on JobInvocation return MonitoringInfos. If existing FLink or Samza code already packages them into MetricResults, then they can convert them to MonitoringInfos there.

Ideally that MetricResults format should only be used in that user level API, and not for internals.

ajamato:
@Ardagan

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MetricResultsProtos.java
```diff
@@ -0,0 +1,235 @@
...
+      processMetric(
+          metric,
+          type,
+          timestamp,
+          counter -> add(metricKey, counter, committed, counters),
+          distribution -> add(metricKey, distribution, committed, distributions),
+          gauge -> add(metricKey, gauge, committed, gauges));
+    }
+
+    private <T> void add(
```
ajamato: With this many params, its not clear what's geting added, please change the name

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MetricResultsProtos.java
```diff
@@ -0,0 +1,235 @@
...
+    private void add(MonitoringInfo monitoringInfo, Boolean committed) {
+      add(
+          keyFromMonitoringInfo(monitoringInfo),
+          monitoringInfo.getType(),
+          monitoringInfo.getTimestamp(),
+          monitoringInfo.getMetric(),
+          committed);
+    }
+
+    private void add(
```
ajamato: Change name

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MetricResultsProtos.java
```diff
@@ -0,0 +1,235 @@
...
+    public MetricResults build() {
+      return new DefaultMetricResults(build(counters), build(distributions), build(gauges));
+    }
+
+    /** Helper for turning a map of result-builders into a sequence of results. */
+    private static <T> List<MetricResult<T>> build(Map<MetricKey, MetricResultBuilder<T>> map) {
+      return map.values().stream().map(MetricResultBuilder::build).collect(toList());
+    }
+
+    private void add(MonitoringInfo monitoringInfo, Boolean committed) {
```
ajamato: Change name

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MetricResultsProtos.java
```diff
@@ -0,0 +1,235 @@
...
+      for (MonitoringInfo monitoringInfo : monitoringInfos) {
+        add(monitoringInfo, committed);
+      }
+    }
+
+    public MetricResults build() {
+      return new DefaultMetricResults(build(counters), build(distributions), build(gauges));
+    }
+
+    /** Helper for turning a map of result-builders into a sequence of results. */
```
ajamato: Please remove helpers if they aren't called in more than one place. I have a feeling some of these add( calls might be in this situation

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MetricResultsProtos.java
```diff
@@ -0,0 +1,235 @@
...
+     * <p>Used while traversing {@link MonitoringInfo} lists where they are stored separately.
+     */
+    private static class MetricResultBuilder<T> {
+      private final MetricKey key;
+      private final T attempted;
+      @Nullable private T committed;
+
+      public MetricResultBuilder(MetricKey key, T attempted) {
+        this.key = key;
+        this.attempted = attempted;
```
ajamato: Add a setAttempted for both
I would say the default is that committed are the only ones emitted by an SDK, which come from the ProcessBundleResponse, while attempted come from the ProcessBundleProgressResponse, requiring a lot more work.

ajamato:
Overall, I think this PR is adding way too much code and complexity. I would rather not submit this. I have spent 2 hours reading this so far and have gone through about half. I'll have to look at it some more tomorrow. But I am not yet convinced everything you've added is necessary at the moment. and much the naming and callback passing is quite confusing so its hard to understand what the various methods are doing.

model/job-management/src/main/proto/beam_job_api.proto
```diff
@@ -155,6 +159,14 @@ message JobMessage {
   }
 }
 
+message GetJobMetricsRequest {
+  string job_id = 1; // (required)
+}
+
+message GetJobMetricsResponse {
+  org.apache.beam.model.fn_execution.v1.MetricResults metrics = 1;
```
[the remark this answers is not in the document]
Ardagan: If we are adding MetricResults we want it to be uniform across languages. This will make it easier to switch between SDKs and simplify some of code logic in the future.
I don't think that MonitoringInfo will change much in the future, so it should be fine to utilize it here directly.

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MetricResultsProtos.java
ryan-williams: This is only used when we know we have an "attempted", and we may or may not have a "committed". All runners today populate "attempted", but only some populate "committed".

Encoding [the knowledge that we always have "attempted" here] makes `build` not have to do any checking; `attempted` is `final` and non-`@Nullable` (unlike `committed`, but that's ok because `MetricResult` also encodes this fact).

I found it cleaner than an approach that pretended to be more general (this is a `private static class` with just the logic it uses), though I understand calling it a `Builder` may be confusing / unidiomatic.

lmk if I should rename it, or still make it a more general "builder"-pattern, or what you think.

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MetricUpdatesProtos.java
ryan-williams: Slightly confused; you're pointing at `MetricsContainerImpl.counterUpdateToMonitoringInfo`, which converts a counter `MetricUpdate` to a `MonitoringInfo`.

In this PR I moved that logic here (a file more explicitly dedicated to MetricUpdate<->MI conversions), removed it there, and generalized it to other metric types. It is still only called from one place, but is reusable if other MetricUpdate<->MI code-paths crop up (it was previously a private method).

It also now uses MI-builder stuff for setting {name,labels} from a MetricKey ([`handleMetricKey` below](https://github.com/apache/beam/pull/7934/files#diff-4f9f0786ec97d7727d0d393e06b2d5f8R49)), so that code is now going through a more widely-shared path.

lmk if I'm missing something.

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MonitoringInfoMetricName.java
ryan-williams: In the MonitoringInfos-everywhere world, it's natural to have MIs that are a mix of user- and system-metrics, and want to turn them into SDK reprs (incl. MetricNames).

In this PR, I rely on [the ability to correctly make an `AutoValue_MetricName` for user-metrics and a `MonitoringInfoMetricName` for system metrics in two places]:
- Flink's bundle-progress listener converts mixed MIs [here]().
- The MetricResultsProtos conversion code-path [uses them](https://github.com/apache/beam/pull/7934/files#diff-e722f484e12b515353e20532d3550465R178) ([`keyFromMonitoringInfo`](https://github.com/apache/beam/pull/7934/files#diff-f0ebadd9b3dbb03171ac530fab654d5dR107) calls [`MIMN.create` below](https://github.com/apache/beam/pull/7934/files#diff-ba1b936a7d4c8d43789e81a816b9bb90R115), which calls this); this is how [JobServicePipelineResult to convert MetricResults proto to an SDK MetricResults](https://github.com/apache/beam/pull/7934/files#diff-284acf2526fcf07c05ee44aefbd9d0c3R134).

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/SimpleMonitoringInfoBuilder.java
```diff
@@ -135,6 +144,23 @@ public SimpleMonitoringInfoBuilder setUrn(String urn) {
     return this;
   }
 
+  public SimpleMonitoringInfoBuilder handleMetricKey(MetricKey key) {
+    MetricName metricName = key.metricName();
+    if (metricName instanceof MonitoringInfoMetricName) {
```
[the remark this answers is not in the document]
ryan-williams: Yea, as I mentioned in the MetricUpdatesProtos comment, in this PR I put that code here, and it's used in multiple places as the canonical way to go from SDK form (MetricKey/MetricName) toward MI form.

SIMB seems like the right place to handle the setting of these MI fields. Everything else now calls this.

I don't think it had a lot to do with MetricUpdates specifically, and the call-site you've linked was a private (non-static) method, inhibiting reuse.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/InMemoryJobService.java
ryan-williams: PipelineResult exposes the SDK-style MetricResults; JobInvocation gets them from there.

Switching to (or adding) MetricResults proto there seemed a bit more involved, but I'll think about it, it might work.

I'd suggest we leave it for a follow-on PR though. This is a case where I tried to play nicely with what the SDK was already doing. We can push for more MI-only APIs everywhere next…

ryan-williams:
[flaky dataflow tests](https://builds.apache.org/job/beam_PreCommit_Java_Examples_Dataflow_Commit/1955/console) ([scan](https://scans.gradle.com/s/fbflw7rtt2ugo/console-log?task=:beam-runners-google-cloud-dataflow-java-examples-streaming:preCommit#L7))
```
> Task :beam-runners-google-cloud-dataflow-java-examples:preCommitLegacyWorker

org.apache.beam.examples.WordCountIT > testE2EWordCount FAILED
    java.lang.RuntimeException at WordCountIT.java:69
        Caused by: java.io.IOException at WordCountIT.java:69

> Task :beam-runners-google-cloud-dataflow-java-examples-streaming:preCommit

org.apache.beam.examples.WordCountIT > testE2EWordCount FAILED
    java.lang.RuntimeException at WordCountIT.java:69
        Caused by: java.io.IOException at WordCountIT.java:69
```

ryan-williams:

runners/flink/src/main/java/org/apache/beam/runners/flink/metrics/FlinkMetricContainer.java
```diff
@@ -98,43 +99,42 @@ public MetricsContainer getMetricsContainer(String stepName) {
...
+            LOG.info("Skipping metric-less MonitoringInfo: {}", monitoringInfo);
+            return;
           }
+          MetricName metricName = MonitoringInfoMetricName.create(monitoringInfo);
+          @Nullable String ptransform = monitoringInfo.getLabelsMap().get(PTRANSFORM_LABEL);
+          MetricsContainer metricsContainer = getMetricsContainer(ptransform);
+          Metric metric = monitoringInfo.getMetric();
+          MetricKey key =
+              MetricKey.create(monitoringInfo.getLabelsMap().get(PTRANSFORM_LABEL), metricName);
+          processMetric(
```
ajamato: processMetric could become a more descriptive name

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MetricResultsProtos.java
ryan-williams: There are a couple meanings of {"attempted", "committed"} floating around. 

You've implied they correspond to MIs that come in via {`ProcessBundleProgressResponse`, `ProcessBundleResponse`}, resp., but there's a pre-fn-api {"physical", "logical"} meaning that most existing code uses.

I'm not trying to resolve these two meanings in this PR. I wrote things to conform to exactly what was already happening, while still supporting the changes I'm attempting to make (which require converting between [new proto] and [existing SDK] representations of metrics).

What was already happening, that I'm preserving, is:
- "attempted" ("physical") is always set
- "committed" ("logical") is sometimes set
  - some runners don't support the latter, so they don't set it (by going through [`MCSM.asAttemptedOnlyMetricResults`](https://github.com/apache/beam/blob/3040767e0312fa73cd08d8f592ce958c484d9fc1/runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MetricsContainerStepMap.java#L114)):
    - [flink](https://github.com/apache/beam/blob/3040767e0312fa73cd08d8f592ce958c484d9fc1/runners/flink/src/main/java/org/apache/beam/runners/flink/metrics/FlinkMetricContainer.java#L146)
    - spark [1](https://github.com/apache/beam/blob/3040767e0312fa73cd08d8f592ce958c484d9fc1/runners/spark/src/main/java/org/apache/beam/runners/spark/metrics/SparkBeamMetric.java#L47) [2](https://github.com/apache/beam/blob/3040767e0312fa73cd08d8f592ce958c484d9fc1/runners/spark/src/main/java/org/apache/beam/runners/spark/SparkPipelineResult.java#L109)
    - samza [1](https://github.com/apache/beam/blob/3040767e0312fa73cd08d8f592ce958c484d9fc1/runners/samza/src/main/java/org/apache/beam/runners/samza/metrics/SamzaMetricsContainer.java#L62) [2](https://github.com/apache/beam/blob/3040767e0312fa73cd08d8f592ce958c484d9fc1/runners/samza/src/main/java/org/apache/beam/runners/samza/SamzaPipelineResult.java#L86)
  - other runners do set both:
    - DataflowMetrics [1](https://github.com/apache/beam/blob/3040767e0312fa73cd08d8f592ce958c484d9fc1/runners/google-cloud-dataflow-java/src/main/java/org/apache/beam/runners/dataflow/DataflowMetrics.java#L143), [2](https://github.com/apache/beam/blob/3040767e0312fa73cd08d8f592ce958c484d9fc1/runners/google-cloud-dataflow-java/src/main/java/org/apache/beam/runners/dataflow/DataflowMetrics.java#L152): "attempted" and "committed" are set to the same value
    - [DirectMetrics](https://github.com/apache/beam/blob/3040767e0312fa73cd08d8f592ce958c484d9fc1/runners/direct-java/src/main/java/org/apache/beam/runners/direct/DirectMetrics.java#L259-L262): both values set, representing "physical" and "logical" values
    - [portable DirectMetrics](https://github.com/apache/beam/blob/3040767e0312fa73cd08d8f592ce958c484d9fc1/runners/direct-java/src/main/java/org/apache/beam/runners/direct/portable/DirectMetrics.java#L263-L266)
      - most portable DirectRunner metrics code seems like copy-pasta from the legacy version. I've consolidated / deleted some of it, but haven't gotten to this yet.
- accordingly, ["committed" is `@Nullable`](https://github.com/apache/beam/blob/3040767e0312fa73cd08d8f592ce958c484d9fc1/sdks/java/core/src/main/java/org/apache/beam/sdk/metrics/MetricResult.java#L65-L66), and ["attempted" is not](https://github.com/apache/beam/blob/3040767e0312fa73cd08d8f592ce958c484d9fc1/sdks/java/core/src/main/java/org/apache/beam/sdk/metrics/MetricResult.java#L68-L69). 

In this `private static` helper, I've written the simplest conversion I can manage between [an attempted metric value] and [optional corresponding committed value that comes in via a different proto field] and a MetricResult that joins them.

lmk if that doesn't make sense!

ryan-williams:
Run Python PreCommit

ryan-williams:
[Some kind of installation error that registered as `:beam-sdks-python:lintPy27` failing](https://scans.gradle.com/s/mna455rvd2eww/console-log?task=:beam-sdks-python:lintPy27)

```
py27-lint inst: /home/jenkins/jenkins-slave/workspace/beam_PreCommit_Python_Commit/src/sdks/python/target/.tox/dist/apache-beam-2.12.0.dev0.zip
--
 
ERROR: invocation failed (exit code 1), logfile: /home/jenkins/jenkins-slave/workspace/beam_PreCommit_Python_Commit/src/sdks/python/target/.tox/py27-lint/log/py27-lint-2.log
ERROR: actionid: py27-lint
msg: installpkg
cmdargs: ['/home/jenkins/jenkins-slave/workspace/beam_PreCommit_Python_Commit/src/sdks/python/target/.tox/py27-lint/bin/python', '/home/jenkins/jenkins-slave/workspace/beam_PreCommit_Python_Commit/src/sdks/python/target/.tox/py27-lint/bin/pip', 'install', '--retries', '10', '/home/jenkins/jenkins-slave/workspace/beam_PreCommit_Python_Commit/src/sdks/python/target/.tox/dist/apache-beam-2.12.0.dev0.zip[test]']
```

ryan-williams:
[Flaky cassandra test](https://builds.apache.org/job/beam_PreCommit_Java_Commit/4582/consoleFull) ([BEAM-6722](https://issues.apache.org/jira/browse/BEAM-6722))
```
> Task :beam-sdks-java-io-cassandra:test

org.apache.beam.sdk.io.cassandra.CassandraIOTest > classMethod FAILED
    java.lang.AssertionError at CassandraIOTest.java:104
```

ryan-williams:
Run Java PreCommit

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/DistributionProtos.java
```diff
@@ -0,0 +1,41 @@
...
+public class DistributionProtos {
+  public static DistributionResult fromProto(IntDistributionData distributionData) {
+    return DistributionResult.create(
+        distributionData.getSum(),
+        distributionData.getCount(),
+        distributionData.getMin(),
+        distributionData.getMax());
+  }
+
+  public static IntDistributionData toProto(DistributionResult distributionResult) {
```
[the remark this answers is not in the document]
ryan-williams: I actually decided to pull these back to this file.

When I put them on `DistributionResult`, that gave sdks/java/core a public-API dependency on the `IntDistributionData` proto (and a bunch of gRPC- and proto-things that come along with it, transitively), [which `GcpCoreApiSurfaceTest` complained about](https://scans.gradle.com/s/xtnseynwu5sqy/console-log?task=:beam-sdks-java-extensions-google-cloud-platform-core:test):

failing api surface test screenshot

I prefer just putting these helpers in this file in runners-core as opposed to adding e.g. `org.apache.beam.model.*` to their public API surface (and everyone else's, they just happen to care enough to have a test…).

I ran into this on other metric-related PRs, and so far have been able to keep it out of master. We may end up needing to just "leak" a lot of proto and gRPC here anyway, so maybe it doesn't matter, but this feels like the more considerate thing to do for now.

lmk what you think

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/GaugeProtos.java
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+package org.apache.beam.runners.core.metrics;
+
+import org.apache.beam.sdk.metrics.GaugeResult;
+import org.apache.beam.vendor.grpc.v1p13p1.com.google.protobuf.Timestamp;
+import org.joda.time.Instant;
+
+/** Convert gauges from proto to SDK representations. */
+public class GaugeProtos {
```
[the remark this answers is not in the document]
ryan-williams: similarly to [above](https://github.com/apache/beam/pull/7934#discussion_r261461766),  [I tried this](https://github.com/apache/beam/pull/7934/files#diff-2fdd57b07c819fee744157907155c430R38), but (shaded) `com.google.protobuf.Timestamp` ends up in the public API surface of downstream modules, which feels icky. I've reverted this back to `GaugeProtos`. 

I don't mind small files of static helpers doing specific things, but lmk if you think I should put this back in core or what.

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/DistributionProtos.java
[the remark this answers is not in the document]
ryan-williams: I actually [end up adding proto things to the sdks/java/core API surface in #7915](https://github.com/apache/beam/pull/7915/files#diff-e1f378c05ed380fffbfb99e0f819afabR59), so maybe it's a moot point…

ryan-williams:
[This is a new one, to me](https://scans.gradle.com/s/kzzeeiswfcejc/console-log?task=:beam-sdks-python-test-suites-dataflow:preCommitIT#L130): `:beam-sdks-python-test-suites-dataflow:preCommitIT`

```
test_streaming_wordcount_it (apache_beam.examples.streaming_wordcount_it_test.StreamingWordCountIT) ... ERROR
--
 
ERROR: test_wordcount_it (apache_beam.examples.wordcount_it_test.WordCountIT)
File "/home/jenkins/jenkins-slave/workspace/beam_PreCommit_Python_Commit/src/sdks/python/apache_beam/examples/wordcount_it_test.py", line 47, in test_wordcount_it
self._run_wordcount_it(wordcount.run)
File "/home/jenkins/jenkins-slave/workspace/beam_PreCommit_Python_Commit/src/sdks/python/apache_beam/examples/wordcount_it_test.py", line 75, in _run_wordcount_it
run_wordcount(test_pipeline.get_full_options_as_args(**extra_opts))
File "/home/jenkins/jenkins-slave/workspace/beam_PreCommit_Python_Commit/src/sdks/python/apache_beam/examples/wordcount.py", line 114, in run
result = p.run()
…
IOError: [Errno 2] No such file or directory: '/tmp/tmpsEVK5k/pickled_main_session'
```

ryan-williams:
Run Python PreCommit

ryan-williams:
[hmm](https://builds.apache.org/job/beam_PreCommit_Python_Phrase/235/console)

> Build timed out (after 120 minutes). Marking the build as aborted.

Some other noise (e.g. `:beam-sdks-python:docs FAILED`) but I'm guessing the timeout is the real culprit. I guess I will try again…

ryan-williams:
Run Python PreCommit

ryan-williams:
Run Portable_Python PreCommit

ryan-williams:
Seems like [the job failed to boot properly](https://builds.apache.org/job/beam_PreCommit_Portable_Python_Commit/1985/console):

```
Using context: Portable_Python ("Run Portable_Python PreCommit")
ERROR: Issue with creating launcher for agent beam9. The agent is being disconnected
[EnvInject] - Loading node environment variables.
ERROR: SEVERE ERROR occurs
```

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MetricUpdatesProtos.java
```diff
@@ -0,0 +1,66 @@
...
+      MetricUpdates.MetricUpdate<T> metricUpdate, BiConsumer<SimpleMonitoringInfoBuilder, T> fn) {
+    SimpleMonitoringInfoBuilder builder = new SimpleMonitoringInfoBuilder(true);
+    MetricKey metricKey = metricUpdate.getKey();
+    builder.setLabelsAndUrnFrom(metricKey);
+
+    fn.accept(builder, metricUpdate.getUpdate());
+    return builder.setTimestampToNow().build();
+  }
+
+  private static <T> void fromMetrics(
```
ajamato: The return type is never used here. Remove it from the signature

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MetricUpdatesProtos.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.function.BiConsumer;
+import org.apache.beam.model.pipeline.v1.MetricsApi.MonitoringInfo;
+import org.apache.beam.sdk.metrics.MetricKey;
+
+/** Convert {@link MetricUpdates} to a {@link List} of {@link MonitoringInfo}. */
+public class MetricUpdatesProtos {
+  public static List<MonitoringInfo> toProto(MetricUpdates metricUpdates) {
+    ArrayList<MonitoringInfo> monitoringInfos = new ArrayList<>();
+
+    fromMetrics(
```
ajamato: Similarily in this code, the BiConsumer usage complicates this and makes it harder to read.
Its simple enough to
(1) make a for loop for each of these calls, iterating through the MetricUpdates
(2) call a helper to do everything else

A lot of this problem comes from the fact that MetricUpdate is just a generic class with no knowledge of what its actually storing. So you can't take advantage of any polymophism with a MetricGaugeUpdate, MetricDistributionUpdate, MetricCounterUpdate, etc. So a lot of the code surrounding this special cases or writes switches for each one. If this polymorphism existed you could tell the MetricUpdate to set itself on a MonitoringInfo/SimpleMonitoringInfoBuilder. That's an option as well, but it will require a larger change. Better to table for now.

A lot of the cases where you pass in a callback seem to be for similar reasons.

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MetricUpdatesProtos.java
```diff
@@ -0,0 +1,66 @@
...
+        metricUpdates.distributionUpdates(),
+        SimpleMonitoringInfoBuilder::setIntDistributionValue);
+    fromMetrics(
+        monitoringInfos, metricUpdates.gaugeUpdates(), SimpleMonitoringInfoBuilder::setGaugeValue);
+
+    return monitoringInfos;
+  }
+
+  private static <T> MonitoringInfo addMetric(
+      MetricUpdates.MetricUpdate<T> metricUpdate, BiConsumer<SimpleMonitoringInfoBuilder, T> fn) {
```
ajamato: please rename fn to setValue

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MonitoringInfoMetricName.java
```diff
@@ -85,12 +87,33 @@ public String getUrn() {
   }
 
   /** @return The labels associated with this MonitoringInfo. */
-  public HashMap<String, String> getLabels() {
+  public Map<String, String> getLabels() {
     return this.labels;
```
ajamato: Would you mind making this return an unmodifyable map?
        return Collections.unmodifiableMap(map);

This would be safer and prevent adding values to it

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MonitoringInfoMetricName.java
```diff
@@ -112,12 +135,27 @@ public boolean equals(Object o) {
...
   @Override
-  public String toString() {
-    StringBuilder builder = new StringBuilder();
-    builder.append(this.urn.toString());
-    builder.append(" ");
-    builder.append(this.labels.toString());
-    return builder.toString();
+  public String toString(String delimiter) {
+    StringBuilder sb = new StringBuilder();
+    if (labels.containsKey(PCOLLECTION_LABEL)) {
```
ajamato: I really don't understand this decision, just make it print all the labels. Why special case PCOLLECTION_LABEL

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/GaugeProtos.java
```diff
@@ -0,0 +1,30 @@
...
+  public static GaugeResult fromProto(long value, Timestamp timestamp) {
+    return GaugeResult.create(
+        value, new Instant(timestamp.getSeconds() * 1000 + timestamp.getNanos() / 1000));
```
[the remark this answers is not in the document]
ryan-williams: good idea, done. There was actually a bug bc I should have been dividing the nanos by 1000000, not 1000 😱 thanks for flagging.

ryan-williams:
Run Java PreCommit

ryan-williams:
Seems like this is going to be chunked into smaller pieces (starting with #8018 / #8022). Closing.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/InMemoryJobService.java
lgajowy: @ryan-williams @ajamato I'd like to get more context here. LMK if I'm missing something.  

Here the `JobInvocation` uses `PipelineResult` that contains a `MetricsResult` rather than `MonitoringInfos`. What would be the best way of getting `MonitoringInfos` in JobInvocation?

Is implementing a `MonitoringInfo portableMetrics()` method in `JobServicePipelineResult` and setting `FutureCallback` with `JobServicePipelineResult instead of `PipelineResult` a good idea? We could use it to call: 

`MonitoringInfo metrics = pipelineResult.portableMetrics();`

instead of : 

`MetricResults metrics = pipelineResult.metrics();`

I don't like that we'd have 2 separate methods for `metrics()` both returning different types but currently I don't see other way to 
> make the interface on JobInvocation return MonitoringInfos

WDYT?

lgajowy: (@up nevermind - I now see that it is a more complex problem than I initially thought. )

ryan-williams: Where did you run into trouble with it @lgajowy?

I thought some version of what you described could make sense.

lgajowy: Sorry I kept you waiting - I was away. I answered on the devlist, here: https://lists.apache.org/thread.html/70803bd35b0b9b6e1202a3baa9872d140492c354e248de07d5d177f6@%3Cdev.beam.apache.org%3E

PR #7936, opened 2019-02-23
Add matchers for python MetricResults, add labels to MetricResults and populate them for dataflow system metrics.

ajamato:
Add matchers for python MetricResults, add labels to MetricResults and populate them for dataflow system metrics.

These matchers can be used for integration testing.

------------------------

### DIFF
--- a/sdks/python/apache_beam/metrics/execution.py
+++ b/sdks/python/apache_beam/metrics/execution.py
@@ -49,8 +49,10 @@ from apache_beam.runners.worker import statesampler
 class MetricKey(object):
   """Key used to identify instance of metric cell.
 
-  Metrics are internally keyed by the name of the step they're associated with
-  and the name of the metric.
+  Metrics are internally keyed by the name of the step they're associated with,
+  the name and namespace (if it is a user defined metric) of the the metric,
+  and any extra label metadata added by the runner specific metric collection
+  service.
   """
   def __init__(self, step, metric, labels=None):
     """Initializes ``MetricKey``.
--- a/sdks/python/apache_beam/metrics/execution.py
+++ b/sdks/python/apache_beam/metrics/execution.py
@@ -66,21 +66,19 @@ class MetricKey(object):
 
   def __eq__(self, other):
     return (self.step == other.step and
-            self.metric == other.metric)
+            self.metric == other.metric and
+            self.labels == other.labels)
 
   def __ne__(self, other):
     # TODO(BEAM-5949): Needed for Python 2 compatibility.
     return not self == other
 
   def __hash__(self):
-    return hash((self.step, self.metric))
+    return hash((self.step, self.metric, frozenset(self.labels)))
 
   def __repr__(self):
     return 'MetricKey(step={}, metric={}, labels={})'.format(
         self.step, self.metric, self.labels)
-
-  def __hash__(self):
-    return hash((self.step, self.metric))
 
 
 class MetricResult(object):
--- a/sdks/python/apache_beam/metrics/execution.py
+++ b/sdks/python/apache_beam/metrics/execution.py
@@ -32,7 +32,6 @@ Available classes:
 
 from __future__ import absolute_import
 
-import pprint
 import threading
 from builtins import object
 from collections import defaultdict
@@ -126,7 +125,7 @@ class MetricResult(object):
         self.key, str(self.committed), str(self.attempted))
 
   def __str__(self):
-    return pprint.pformat(self)
+    return repr(self)
 
   @property
   def result(self):
--- a/sdks/python/apache_beam/metrics/execution.py
+++ b/sdks/python/apache_beam/metrics/execution.py
@@ -32,6 +32,7 @@ Available classes:
 
 from __future__ import absolute_import
 
+import pprint
 import threading
 from builtins import object
 from collections import defaultdict
@@ -51,15 +52,17 @@ class MetricKey(object):
   Metrics are internally keyed by the name of the step they're associated with
   and the name of the metric.
   """
-  def __init__(self, step, metric):
+  def __init__(self, step, metric, labels=None):
     """Initializes ``MetricKey``.
 
     Args:
       step: A string with the step this metric cell is part of.
-      metric: A ``MetricName`` that identifies a metric.
+      metric: A ``MetricName`` namespace+name that identifies a metric.
+      labels: An arbitrary set of labels that also identifies the metric.
     """
     self.step = step
     self.metric = metric
+    self.labels = labels if labels else dict()
 
   def __eq__(self, other):
     return (self.step == other.step and
@@ -73,8 +76,8 @@ class MetricKey(object):
     return hash((self.step, self.metric))
 
   def __repr__(self):
-    return 'MetricKey(step={}, metric={})'.format(
-        self.step, self.metric)
+    return 'MetricKey(step={}, metric={}, labels={})'.format(
+        self.step, self.metric, self.labels)
 
   def __hash__(self):
     return hash((self.step, self.metric))
@@ -121,6 +124,9 @@ class MetricResult(object):
   def __repr__(self):
     return 'MetricResult(key={}, committed={}, attempted={})'.format(
         self.key, str(self.committed), str(self.attempted))
+
+  def __str__(self):
+    return pprint.pformat(self)
 
   @property
   def result(self):
--- a/sdks/python/apache_beam/runners/dataflow/dataflow_metrics.py
+++ b/sdks/python/apache_beam/runners/dataflow/dataflow_metrics.py
@@ -51,6 +51,11 @@ def _get_match(proto, filter_fn):
   return query[0]
 
 
+# V1b3 MetricStructuredName keys to accept and copy to the MetricKey labels.
+WHITELISTED_LABELS = [
+    'execution_step', 'original_name', 'output_user_name', 'step']
+
+
 class DataflowMetrics(MetricResults):
   """Implementation of MetricResults class for the Dataflow runner."""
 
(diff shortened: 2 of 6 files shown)


ajamato:
@Ardagan @ryan-williams PTAL and review these changes. I'm going to use these in a follow up PR to build an integration test.

sdks/python/apache_beam/testing/metric_result_matchers.py
```diff
@@ -0,0 +1,261 @@
...
+      self.committed_value.describe_to(description)
+
+  def describe_mismatch(self, metric_result, mismatch_description):
+    mismatch_description.append_text("was").append_value(metric_result)
+
+
+class DistributionMatcher(BaseMatcher):
+  """A PyHamcrest matcher that validates counter distributions."""
+
+  @with_input_types(
```
robertwb: These decorators are meant to apply to transforms, they have no meaning here. (Also, sum_value, etc. is not a tuple of (int, Matcher).

sdks/python/apache_beam/testing/metric_result_matchers.py
```diff
@@ -0,0 +1,261 @@
...
+      self.min_value.describe_to(description)
+    if self.max_value:
+      description.append_text(" max_value: ")
+      self.max_value.describe_to(description)
+
+  def describe_mismatch(self, distribution_result, mismatch_description):
+    mismatch_description.append_text("was").append_value(distribution_result)
+
+@with_input_types(typing.Tuple[str, Matcher])
+def has_step_name(step):
```
robertwb: It would seam more Pythonic to just expose MetricResultMatcher and let users set the right keywords. (The pattern here is needed in java because one doesn't have keyword arguments.)

sdks/python/apache_beam/testing/metric_result_matchers.py
ajamato: > My first thought when seeing this was asking why we need metrics-specific matchers. If possible, I would make the API for fetching metrics simple enough that one could just use that directly with the standard hamcrest APIs (maybe with a helper or two, such as a DistributionValue matcher).

The main benefit of the matchers, is to allow you to specify these checks, without having to go through the deeper layers of the structure. It also gives us some freedom to change the MetricResult structure, and just change the matcher class. Rather than needing to change all the integration tests which use MetricResult.

ajamato: My original idea was to just populate expected MetricRestuls and compare them. I had done this for MonitoringInfos, but it required writing a lot of code in tests to special case things. Like if you don't want an exact value match. I.e. execution time has an int64_value > 0. With the matchers you can just pass in  greater_than(0) matcher for the value. Then you can reuse the same assertion code in all tests.

We used this technique for the internal metric tests for Dataflow, and it was pretty useful, allowing us to add metric matching to our integration tests without having to write code to query, iterate, etc. over the metrics.

robertwb: But users still have to got through the deeper layers of the structure, and we don't have the freedom to change the MetricsResult due to that, right? In which case it'd be preferable to make the API simple enough that one could write `self.assertGreater(0, some_simple_expression)`. Or perhaps is this really for testing things that are not part of the public API?

(Yes, I'm strongly in favor of being able to use matchers to test things like inequalities.)

sdks/python/apache_beam/testing/metric_result_matchers.py
```diff
@@ -0,0 +1,261 @@
...
+    self.namespace = _matcher_or_equal_to(namespace)
+    self.name = _matcher_or_equal_to(name)
+    self.step = _matcher_or_equal_to(step)
+    self.label_key = _matcher_or_equal_to(label_key)
+    self.label_value = _matcher_or_equal_to(label_value)
+    self.attempted_value = _matcher_or_equal_to(attempted_value)
+    self.committed_value = _matcher_or_equal_to(committed_value)
+
+  def _matches(self, metric_result):
+    if self.namespace is not None and not self.namespace.matches(
```
ryan-williams: any significance to the `is not None` here vs just truthiness check on the next couple string fields below?

sdks/python/apache_beam/testing/metric_result_matchers.py
ajamato: I agree that we do need to make MetricResult suit our use cases. But I don't want to do that right now. I'm happy to delete these matchers after we iterate on it and this becomes much simpler. 

But the matchers themselves will offer a lot of other benefits anyways, so I still think we should proceed with this.

I do see several problems with MetricResults, which need to be discussed on the dev list:
- MetricResults should probably be a proto, instead of re-implemented in every language.
- MetricResults should not make assumptions like: all metrics have a step associate with them
- MetricResults need to support labels in all languages.
- MetricResults don't have a proper way to reflect how a non-user metric should be returned. I.e. where does the URN go? namespace or name field? etc. Should those details from the MonitoringInfo format even 

Further, as we mentioned in the GetMetrics doc, It will be a very challenging task to upgrade the Dataflow Runner to maintain the URN and all those fields through the entire stack, so MetricResults will have to remain in Beta until at least that design and work is complete. We should not guarantee this format yet either, so I do think it needs to be flexible.

robertwb: I would say it's deeper in that the MetricsResults objects (and constructing a query to get a list of them) is generally overkill for "I set a counter in one of my DoFns, and just want to see its value." But that's also deeper than this change, which seems to have immediate use. I just want to be aware when we're writing adapters because the existing API doesn't suit our needs which solves the problem for us but not all our users.

sdks/python/apache_beam/testing/metric_result_matchers.py
pabloem: this is none of my business, but I was in the neighborhood. I imagine an empty string is a valid namespace, so there's that.

ajamato:
Run Portable_Python PreCommit

sdks/python/apache_beam/metrics/execution.py
```diff
@@ -122,6 +125,9 @@ def __repr__(self):
     return 'MetricResult(key={}, committed={}, attempted={})'.format(
         self.key, str(self.committed), str(self.attempted))
 
+  def __str__(self):
+    return pprint.pformat(self)
```
robertwb: Does this actually print a nice representation of self?

sdks/python/apache_beam/runners/dataflow/dataflow_metrics.py
```diff
@@ -51,6 +51,11 @@ def _get_match(proto, filter_fn):
   return query[0]
 
 
+# V1b3 MetricStructuredName keys to accept and copy to the MetricKey labels.
+WHITELISTED_LABELS = [
```
robertwb: Whitelist seems like the wrong name here, maybe V1b3_LABELS or STRUCTURED_NAME_LABELS?

sdks/python/apache_beam/runners/dataflow/dataflow_metrics.py
```diff
@@ -158,6 +180,11 @@ def _populate_metric_results(self, response):
                                  attempted=attempted,
                                  committed=committed))
 
+  def _populate_metric_results(self, response):
+    """Take a list of metrics, and convert it to a list of MetricResult."""
+    result = []
+    self._populate_metrics(response, result, user_metrics=True)
```
robertwb: Do we ever want to call these separately?

sdks/python/apache_beam/runners/dataflow/dataflow_metrics.py
ajamato: Done

sdks/python/apache_beam/metrics/execution.py
ajamato: It does, but I think we can just call repr instead of using a separate ppritn representation. I only used that to debug a bit

sdks/python/apache_beam/runners/dataflow/dataflow_metrics.py
ajamato: My intention here was the make the query() code only return user metrics, and make all metrics accessible when calling all_metrics(). After a discussion and design on the dev list, we can consider adding the system metrics to the regular query() call.

Fixing this up to reflect this, I can see I forgot to make query() only return user metrics.

ajamato:
@robertwb Can we proceed with this one and merge it?

ajamato:
R: @pabloem

sdks/python/apache_beam/runners/dataflow/dataflow_metrics.py
```diff
@@ -200,9 +220,18 @@ def _get_metrics_from_dataflow(self):
       self._cached_metrics = job_metrics
     return job_metrics
 
+  def all_metrics(self):
```
Ardagan: Can we expand query() and __populate_metrics() with "query_user" and "query_v1b3" (based on code, can be query_system) parameters. This will keep only 1 method that will be more flexible than two of currently existing.

Default values should keep code backwards compatible.

ajamato: No, we need to discuss that kind of change on the dev list. This PR introduces he bare minimum to support integration testing. To enhance this API we need to design on the dev list.

sdks/python/apache_beam/metrics/execution.py
```diff
@@ -48,36 +48,38 @@
...
+            self.metric == other.metric and
+            self.labels == other.labels)
 
   def __ne__(self, other):
     # TODO(BEAM-5949): Needed for Python 2 compatibility.
     return not self == other
 
   def __hash__(self):
-    return hash((self.step, self.metric))
+    return hash((self.step, self.metric, frozenset(self.labels)))
```
pabloem: @ajamato @Ardagan - We did not resolve this issue: It seems to me that `labels` are supposed to be a Python set, but a dict is being used here (see line 66).
If labels are a dictionary, then you need to do `frozenset(self.labels.items())` to include the values in the hash and not just the keys.
It seems that labels should be a set. In that case, line 66 should be `self.labels = labels if labels else set()`

PR #7937, opened 2019-02-23
[BEAM-6778] Enable Bundle Finalization in Python SDK harness over FnApi

boyuanzz:
This PR is focus on supporting [Bundle Finalization](https://s.apache.org/beam-finalizing-bundles) in the Python SDK harness level. The finalizer API design is under discussing and would be changed in the future. 
In order to implement the finalization logic inside the sdk harness, there are some simplifications in other components:
1. For the finalizer API, it may be changed by upcoming PR.
2. For the Python Fnapi runner, it simply requests a finalization if the ```ProcessBundleResponse``` sets the ```requires_finalization``` to true.

R: @aaltay @robertwb

### DIFF
--- a/sdks/python/apache_beam/runners/common.py
+++ b/sdks/python/apache_beam/runners/common.py
@@ -292,6 +292,7 @@ class DoFnInvoker(object):
     self.output_processor = output_processor
     self.signature = signature
     self.user_state_context = None
+    self.bundle_finalizer_param = None
 
   @staticmethod
   def create_invoker(
@@ -299,7 +300,8 @@ class DoFnInvoker(object):
       output_processor=None,
       context=None, side_inputs=None, input_args=None, input_kwargs=None,
       process_invocation=True,
-      user_state_context=None):
+      user_state_context=None,
+      bundle_finalizer_param=None):
     """ Creates a new DoFnInvoker based on given arguments.
 
     Args:
@@ -321,6 +323,8 @@ class DoFnInvoker(object):
                             method efficiently.
         user_state_context: The UserStateContext instance for the current
                             Stateful DoFn.
+        bundle_finalizer_param: The param that passed to a process method, which
+                                allows a callback to be registered.
     """
     side_inputs = side_inputs or []
     default_arg_values = signature.process_method.defaults
@@ -333,7 +337,7 @@ class DoFnInvoker(object):
       return PerWindowInvoker(
           output_processor,
           signature, context, side_inputs, input_args, input_kwargs,
-          user_state_context)
+          user_state_context, bundle_finalizer_param)
 
   def invoke_process(self, windowed_value, restriction_tracker=None,
                      output_processor=None,
@@ -423,7 +427,8 @@ class PerWindowInvoker(DoFnInvoker):
   """An invoker that processes elements considering windowing information."""
 
   def __init__(self, output_processor, signature, context,
-               side_inputs, input_args, input_kwargs, user_state_context):
+               side_inputs, input_args, input_kwargs, user_state_context,
+               bundle_finalizer_param):
     super(PerWindowInvoker, self).__init__(output_processor, signature)
     self.side_inputs = side_inputs
     self.context = context
@@ -437,6 +442,7 @@ class PerWindowInvoker(DoFnInvoker):
     self.is_splittable = signature.is_splittable_dofn()
     self.restriction_tracker = None
     self.current_windowed_value = None
+    self.bundle_finalizer_param = bundle_finalizer_param
 
     # Try to prepare all the arguments that can just be filled in
     # without any additional work. in the process function.
@@ -486,6 +492,8 @@ class PerWindowInvoker(DoFnInvoker):
       elif isinstance(d, core.DoFn.StateParam):
         args_with_placeholders.append(ArgPlaceholder(d))
       elif isinstance(d, core.DoFn.TimerParam):
+        args_with_placeholders.append(ArgPlaceholder(d))
+      elif d == core.DoFn.BundleFinalizerParam:
         args_with_placeholders.append(ArgPlaceholder(d))
       else:
         # If no more args are present then the value must be passed via kwarg
@@ -608,6 +616,8 @@ class PerWindowInvoker(DoFnInvoker):
       elif isinstance(p, core.DoFn.TimerParam):
         args_for_process[i] = (
             self.user_state_context.get_timer(p.timer_spec, key, window))
+      elif p == core.DoFn.BundleFinalizerParam:
+        args_for_process[i] = self.bundle_finalizer_param
 
     if additional_kwargs:
       if kwargs_for_process is None:
@@ -694,6 +704,7 @@ class DoFnRunner(Receiver):
 
     self.step_name = step_name
     self.context = DoFnContext(step_name, state=state)
+    self.bundle_finalizer_param = DoFn.BundleFinalizerParam()
 
     do_fn_signature = DoFnSignature(fn)
 
@@ -722,7 +733,8 @@ class DoFnRunner(Receiver):
 
     self.do_fn_invoker = DoFnInvoker.create_invoker(
         do_fn_signature, output_processor, self.context, side_inputs, args,
-        kwargs, user_state_context=user_state_context)
+        kwargs, user_state_context=user_state_context,
+        bundle_finalizer_param=self.bundle_finalizer_param)
 
   def receive(self, windowed_value):
     self.process(windowed_value)
@@ -732,6 +744,9 @@ class DoFnRunner(Receiver):
       return self.do_fn_invoker.invoke_process(windowed_value)
     except BaseException as exn:
       self._reraise_augmented(exn)
+
+  def finalize(self):
+    self.bundle_finalizer_param.finalize_bundle()
 
   def process_with_restriction(self, windowed_value):
     element, restriction = windowed_value.value
--- a/sdks/python/apache_beam/runners/portability/fn_api_runner.py
+++ b/sdks/python/apache_beam/runners/portability/fn_api_runner.py
@@ -1298,6 +1298,15 @@ class BundleManager(object):
     if result.error:
       raise RuntimeError(result.error)
 
+    if result.process_bundle.requires_finalization:
+      finalize_request = beam_fn_api_pb2.InstructionRequest(
+          finalize_bundle=
+          beam_fn_api_pb2.FinalizeBundleRequest(
+              instruction_reference=process_bundle_id
+          ))
+      self._controller.control_handler.push(
+          finalize_request)
+
     return result, split_results
 
 
--- a/sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
+++ b/sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
@@ -21,6 +21,7 @@ import collections
 import logging
 import os
 import random
+import shutil
 import sys
 import tempfile
 import threading
@@ -59,12 +60,6 @@ class FnApiRunnerTest(unittest.TestCase):
 
   def create_pipeline(self):
     return beam.Pipeline(runner=fn_api_runner.FnApiRunner())
-
-  def create_pipeline_with_grpc(self):
-    return beam.Pipeline(
-        runner=fn_api_runner.FnApiRunner(
-            default_environment=beam_runner_api_pb2.Environment(
-                urn=python_urns.EMBEDDED_PYTHON_GRPC)))
 
   def test_assert_that(self):
     # TODO: figure out a way for fn_api_runner to parse and raise the
@@ -807,29 +802,30 @@ class FnApiRunnerTest(unittest.TestCase):
       print(res._monitoring_infos_by_stage)
       raise
 
-  def test_duplicated_callbacks(self):
-    event_recorder = EventRecorder(tempfile.gettempdir())
-    event_recorder.record('test')
+  def test_callbacks_with_exception(self):
     elements_list = ['1', '2']
 
-    class FinalizebleDoFnWithSameRegister(beam.DoFn):
+    def raise_expetion():
+      raise Exception('raise exception when calling callback')
+
+    class FinalizebleDoFnWithException(beam.DoFn):
+
       def process(
           self,
           element,
           bundle_finalizer=beam.DoFn.BundleFinalizerParam):
-        bundle_finalizer.register(lambda: event_recorder.delete_file())
+        bundle_finalizer.register(raise_expetion)
         yield element
 
-    with self.create_pipeline_with_grpc() as p:
+    with self.create_pipeline() as p:
       res = (p
              | beam.Create(elements_list)
-             | beam.ParDo(FinalizebleDoFnWithSameRegister()))
+             | beam.ParDo(FinalizebleDoFnWithException()))
       assert_that(res, equal_to(['1', '2']))
 
   def test_register_finalizations(self):
     event_recorder = EventRecorder(tempfile.gettempdir())
-    elements_list = ['1', '2']
-    expected_res = '1\n2\n'
+    elements_list = ['2', '1']
 
     class FinalizableDoFn(beam.DoFn):
       def process(
@@ -839,7 +835,7 @@ class FnApiRunnerTest(unittest.TestCase):
         bundle_finalizer.register(lambda: event_recorder.record(element))
         yield element
 
-    with self.create_pipeline_with_grpc() as p:
+    with self.create_pipeline() as p:
       res = (p
              | beam.Create(elements_list)
              | beam.ParDo(FinalizableDoFn()))
@@ -847,7 +843,8 @@ class FnApiRunnerTest(unittest.TestCase):
       assert_that(res, equal_to(elements_list))
 
     results = event_recorder.events()
-    self.assertEquals(results, expected_res)
+    event_recorder.cleanup()
+    self.assertEquals(results, sorted(elements_list))
 
 
 class FnApiRunnerTestWithGrpc(FnApiRunnerTest):
@@ -1131,6 +1128,7 @@ _pickled_element_counters = {}
 def _unpickle_element_counter(name):
   return _pickled_element_counters[name]
 
+
 class EventRecorder(object):
   """Used to be registered as a callback in bundle finalization.
 
@@ -1138,26 +1136,25 @@ class EventRecorder(object):
   cannot keep callback records when passing into one DoFn.
   """
   def __init__(self, tmp_dir):
-    self.tmp_dir = tmp_dir
-    self.file_path = os.path.join(self.tmp_dir, uuid.uuid4().hex + '.txt')
+    self.tmp_dir = os.path.join(tmp_dir, uuid.uuid4().hex)
+    os.mkdir(self.tmp_dir)
 
   def record(self, content):
-    tmp_file = open(self.file_path, 'a')
-    tmp_file.write(content + '\n')
-    tmp_file.close()
+    file_path = os.path.join(self.tmp_dir, uuid.uuid4().hex + '.txt')
+    with open(file_path, 'a') as f:
+      f.write(content)
 
   def events(self):
-    tmp_file = open(self.file_path, 'r')
-    content = tmp_file.read()
-    tmp_file.close()
-    self.delete_file()
-    return content
+    content = []
+    record_files = [f for f in os.listdir(self.tmp_dir) if os.path.isfile(
+        os.path.join(self.tmp_dir, f))]
+    for file in record_files:
+      with open(os.path.join(self.tmp_dir, file), 'r') as f:
+        content.append(f.read())
+    return sorted(content)
 
-  def delete_file(self):
-    try:
-      os.remove(self.file_path)
-    except:
-      raise Exception('Trying to delete non-exist file')
+  def cleanup(self):
+    shutil.rmtree(self.tmp_dir)
 
 
 if __name__ == '__main__':
--- a/sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
+++ b/sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
@@ -871,6 +871,9 @@ class FnApiRunnerTestWithBundleRepeat(FnApiRunnerTest):
   def create_pipeline(self):
     return beam.Pipeline(
         runner=fn_api_runner.FnApiRunner(bundle_repeat=3))
+  
+  def test_register_finalizations(self):
+    raise unittest.SkipTest("TODO: Avoid bundle finalizations on repeat.")
 
 
 class FnApiRunnerSplitTest(unittest.TestCase):
@@ -1141,7 +1144,7 @@ class EventRecorder(object):
 
   def record(self, content):
     file_path = os.path.join(self.tmp_dir, uuid.uuid4().hex + '.txt')
-    with open(file_path, 'a') as f:
+    with open(file_path, 'w') as f:
       f.write(content)
 
   def events(self):
--- a/sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
+++ b/sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
@@ -60,6 +60,12 @@ class FnApiRunnerTest(unittest.TestCase):
   def create_pipeline(self):
     return beam.Pipeline(runner=fn_api_runner.FnApiRunner())
 
+  def create_pipeline_with_grpc(self):
+    return beam.Pipeline(
+        runner=fn_api_runner.FnApiRunner(
+            default_environment=beam_runner_api_pb2.Environment(
+                urn=python_urns.EMBEDDED_PYTHON_GRPC)))
+
   def test_assert_that(self):
     # TODO: figure out a way for fn_api_runner to parse and raise the
     # underlying exception.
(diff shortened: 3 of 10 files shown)


boyuanzz:
Run Portable_Python PreCommit

boyuanzz:
@robertwb PLAT. 
Also I'm trying to have better tests for this feature with FnApiRunner. But I found it kind of difficult for me. Here are several of my tries:
1. Use a list to keep track whether the callback gets called:
```python
  def test_finalize(self):
    counter = []
    class FinalizableDoFn(beam.DoFn):
      def process(
          self,
          element,
          bundle_finalizer=beam.DoFn.BundleFinalizerParam()):
        if element == 1:
          bundle_finalizer.register(lambda: counter.append(element))
        if element == 2:
          bundle_finalizer.register(lambda: counter.append(element))
        yield element

     with self.create_pipeline() as p:
      res = (p
             | beam.Create([1, 2])
             | beam.ParDo(FinalizableDoFn()))
      assert_that(counter, equal_to([1, 2]))
```
But it turns out the counter cannot be written properly.

2. Using the user metric inside one ParDo to count the callback, but also doesn't work since FnApiRunner seems like not support metric_containers() yet.

Do you have any idea about better unit tests? Thanks for your help!

sdks/python/apache_beam/runners/common.py
```diff
@@ -237,6 +239,16 @@ def __init__(self, do_fn):
         method = timer_spec._attached_callback
         self.timer_methods[timer_spec] = MethodWrapper(do_fn, method.__name__)
 
+    self._bundle_finalizer = self.get_bundle_finalizer()
+    if self._bundle_finalizer:
+      self.finalize_method = MethodWrapper(
+          self._bundle_finalizer, 'finalize_bundle')
+
+  def get_bundle_finalizer(self):
```
robertwb: This isn't really a method that we want to call; this seems to be leftover from the previous API. I think we can simplify things as follows: 

1) The DoFnRunner holds an instance of a _BundleFinalizerParam, which it passes to the DoFnInvoker.
2) The PerWindowInvoker swaps this out for core.DoFn.BundleFinalizerParam, the same place it's swapping out ElementParam, WindowParam, TimestampParam, etc.
3) Every Operation has new needs_finalization and finalize_bundle methods, which is a no-op except for ParDoOperation which calls a similar operation on DoFnRunner which looks at whether _BundleFinalizerParam has any callbacks/calls them respectively. 
4) The BundleProcessor object has similar methods, which just walk over its set of all operations and calls them.

sdks/python/apache_beam/transforms/core.py
```diff
@@ -1030,7 +1046,9 @@ def to_runner_api_parameter(self, context):
     picked_pardo_fn_data = pickler.dumps(self._pardo_fn_data())
     state_specs, timer_specs = userstate.get_dofn_specs(self.fn)
     from apache_beam.runners.common import DoFnSignature
-    is_splittable = DoFnSignature(self.fn).is_splittable_dofn()
+    do_fn_signature = DoFnSignature(self.fn)
+    is_splittable = do_fn_signature.is_splittable_dofn()
+    is_requests_finalization = do_fn_signature.is_request_finalization()
```
robertwb: We can't know that here, only once finalization callbacks have been registered.

boyuanzz:
> How to test is a really good question, as the only thing bundle finalization can do is side effects. This is especially true if we want to test distributed runners. (Note that even in process, DoFns get serialized and deserialized (with a new copy of their state), so we can't just store lists and expect them to get populated.)
> 
I agree with this point and that's why my demo case doesn't work.

> Here's an idea. You could create an EventRecorder class that has a temporary directory as a parameter. It would have a `record()` method that writes to a random file in that directory (say, `uuid.uuid4().hex + '.txt'`) and a events() method that would return a list of the contents of each `*.txt`. Your finalizers could record events, and your test body could inspect them. Eventually this could be generalized to distributed filesystems as well (but probably not dealing with on first pass.)

Thanks so much for this idea. I was thinking about this yesterday but I'm not sure whether current test framework could do that. I'll try this today, if that works, then there are many scenarios can be tested.

boyuanzz:
@robertwb Could you please do another around of review? I'll follow up with more test cases either in this PR or in one followup PR. Thanks for your help!

sdks/python/apache_beam/runners/common.py
boyuanzz: Changed in the latest version. Thanks!

sdks/python/apache_beam/transforms/core.py
boyuanzz: Yes you are right. Have changed to verify whether there are callbacks registered after finishing bundle processing.

sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -305,10 +316,14 @@ def get_bundle_processor(self, instruction_id, bundle_descriptor_id):
       with state_handler.process_instruction_id(instruction_id):
         yield processor
     finally:
-      del self.active_bundle_processors[instruction_id]
+      requires_finalization = \
+        self.active_bundle_processors[instruction_id].requires_finalization
```
robertwb: OK, I'm seeing why this is an instance variable now. It's odd to "return" whether finalization is required both as a self mutation and in the return value--how about we make this a method (it should be really cheap to compute; probably don't need to cache the result) and use that uniformly everywhere?

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1284,6 +1284,15 @@ def process_bundle(self, inputs, expected_outputs):
       logging.debug('Wait for the bundle to finish.')
       result = result_future.get()
 
+      if result.process_bundle.requires_finalization:
```
robertwb: And not result.error?

robertwb:
Also, link to JIRA with [BEAM-XXX] in title.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1284,6 +1284,15 @@ def process_bundle(self, inputs, expected_outputs):
...
+        finalize_request = beam_fn_api_pb2.InstructionRequest(
+            finalize_bundle=
+            beam_fn_api_pb2.FinalizeBundleRequest(
+                instruction_reference=process_bundle_id
+            ))
+        self._controller.control_handler.push(
```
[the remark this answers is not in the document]
boyuanzz: According to the best effort finalization, seems like it's unnecessary for a runner waiting for the FinalizeResponse right?

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
```diff
@@ -756,6 +756,34 @@ def create_pipeline(self):
...
+            default_environment=beam_runner_api_pb2.Environment(
+                urn=python_urns.EMBEDDED_PYTHON_GRPC)))
+
+  def test_finalize(self):
+    class FinalizableDoFn(beam.DoFn):
+      def process(
+          self,
+          element,
+          bundle_finalizer=beam.DoFn.BundleFinalizerParam):
+        if element == 1:
```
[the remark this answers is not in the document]
boyuanzz: right, this demo test just wanna to show the callback can be data dependent. I'll update this test.

sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -305,10 +316,14 @@ def get_bundle_processor(self, instruction_id, bundle_descriptor_id):
...
+      if not requires_finalization:
+        del self.active_bundle_processors[instruction_id]
     # Outside the finally block as we only want to re-use on success.
-    processor.reset()
-    self.cached_bundle_processors[bundle_descriptor_id].append(processor)
+    if not requires_finalization:
+      processor.reset()
```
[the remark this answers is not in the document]
boyuanzz: If there is a finalization requires for one bundle, then that bundle_processor should stay in active cache until calling on finalize. Added reusing bundle_processor after finalization call. I'll refactor this part after https://github.com/apache/beam/pull/8009 merged.

sdks/python/apache_beam/transforms/core.py
```diff
@@ -344,9 +362,11 @@ class DoFn(WithTypeHints, HasDisplayData, urns.RunnerApiFn):
   TimestampParam = _DoFnParam('TimestampParam')
   WindowParam = _DoFnParam('WindowParam')
   WatermarkReporterParam = _DoFnParam('WatermarkReporterParam')
+  BundleFinalizerParam = _BundleFinalizerParam
```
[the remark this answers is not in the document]
boyuanzz: I couldn't find a usage for WatermarkReporterParam so far. Is there any difference between these 2 ways?

boyuanzz:
@robertwb Could you please take another look?

boyuanzz:
Run Portable_Python PreCommit

boyuanzz:
Run Portable_Python PreCommit

boyuanzz:
Ping

sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -350,6 +362,16 @@ def process_bundle_progress(self, request, instruction_id):
             metrics=processor.metrics() if processor else None,
             monitoring_infos=processor.monitoring_infos() if processor else []))
 
+  def finalize_bundle(self, request, instruction_id):
+    processor = self.bundle_processor_cache.lookup(
+        request.instruction_reference)
+    if processor:
+      finalize_response = processor.finalize_bundle()
```
robertwb: Maybe a 

```
try:
  ...
except:
  self.bundle_processor_cache.discard(...)
  raise
```

sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -362,6 +384,9 @@ def maybe_profile(self, instruction_id):
     else:
       yield
 
+  def has_active_processors(self):
```
robertwb: This isn't correct, as bundle_processor_cache is shared. Hopefully we don't need it though.

sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -186,11 +186,11 @@ def task():
       try:
         self._execute(lambda: worker.do_instruction(work), work)
       finally:
-        # Delete the instruction_id <-> worker mapping
-        self._instruction_id_vs_worker.pop(work.instruction_id, None)
-        # Put the worker back in the free worker pool
-        self.workers.put(worker)
-
+        if not worker.has_active_processors():
```
robertwb: The bundle processors must be kept out of the free pool, but the workers need not be. Instead, grab a fresh worker whenever a finalization request comes through.

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
```diff
@@ -753,6 +753,81 @@ def create_pipeline(self):
...
 
+class EventRecorder(object):
+  def __init__(self):
+    self.work_path = os.getcwd() + '/EventRecorder'
+    if os.path.exists(self.work_path):
+      self.cleanup()
+    os.mkdir(self.work_path)
+
+  def record(self, file_name):
+    file_path = os.path.join(self.work_path, str(file_name) + '.txt')
```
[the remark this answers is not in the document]
boyuanzz: I don't think a finalizer could be called twice unless it's registered twice. If a finalizer callback gets registered twice, then it's the user's choice and I don't think we can say it's not a expected behavior. Instead, how about adding try-catch when calling callbacks in case the situation like, trying to delete one deleted file?

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
```diff
@@ -753,6 +753,81 @@ def create_pipeline(self):
...
+
+
+class FnApiRunnerFinalizeBundleTest(FnApiRunnerTest):
+  def create_pipeline(self):
+    return beam.Pipeline(
+        runner=fn_api_runner.FnApiRunner(
+            default_environment=beam_runner_api_pb2.Environment(
+                urn=python_urns.EMBEDDED_PYTHON_GRPC)))
+
+  def test_no_finalize_call_without_register(self):
```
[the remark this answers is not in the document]
boyuanzz: This test is trying to verify when a DoFn has a FinalizeParam without any registration, the pipeline can run successfully and get correct results.

sdks/python/apache_beam/runners/worker/sdk_worker.py
boyuanzz: If looking at here: https://github.com/apache/beam/blob/05bc3f6b8d46dbbdd6a17b19db01b92fc9c9e276/sdks/python/apache_beam/runners/worker/sdk_worker.py#L212, the SdkHarness is still caching the ```instruction_id -> worker``` map. With BundleProcessorCache, I think this map can be removed and when there is a new request coming in, get a new worker directly. wdyt?

sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -313,15 +322,18 @@ def process_bundle(self, request, instruction_id):
       with bundle_processor.state_handler.process_instruction_id(
           instruction_id):
         with self.maybe_profile(instruction_id):
-          delayed_applications = bundle_processor.process_bundle(instruction_id)
+          delayed_applications, requests_finalization = \
```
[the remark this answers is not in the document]
boyuanzz: Just curious, any reason why not a backslash?

sdks/python/apache_beam/runners/worker/sdk_worker.py
boyuanzz: Implementation in commit https://github.com/apache/beam/pull/7937/commits/46dab24f559a62653f47a794336003e6530e9fc5

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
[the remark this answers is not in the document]
boyuanzz: For https://docs.python.org/2/library/tempfile.html:
I tried to use tempfile.NamedTemporaryFile but it made ParDo fail to initialize:
```
  File "/usr/local/google/home/boyuanz/beam/sdks/python/apache_beam/runners/portability/fn_api_runner_test.py", line 820, in test_register_finalizations
    fn = beam.ParDo(FinalizableDoFn())
  File "/usr/local/google/home/boyuanz/beam/sdks/python/apache_beam/transforms/core.py", line 953, in __init__
    super(ParDo, self).__init__(fn, *args, **kwargs)
  File "/usr/local/google/home/boyuanz/beam/sdks/python/apache_beam/transforms/ptransform.py", line 688, in __init__
    self.fn = pickler.loads(pickler.dumps(self.fn))
  File "/usr/local/google/home/boyuanz/beam/sdks/python/apache_beam/internal/pickler.py", line 258, in loads
    return dill.loads(s)
  File "/usr/local/google/home/boyuanz/beam/sdks/python/.eggs/dill-0.2.9-py2.7.egg/dill/_dill.py", line 317, in loads
    return load(file, ignore)
  File "/usr/local/google/home/boyuanz/beam/sdks/python/.eggs/dill-0.2.9-py2.7.egg/dill/_dill.py", line 305, in load
    obj = pik.load()
  File "/usr/lib/python2.7/pickle.py", line 864, in load
    dispatch[key](self)
  File "/usr/lib/python2.7/pickle.py", line 1221, in load_build
    setstate = getattr(inst, "__setstate__", None)
  File "/usr/lib/python2.7/tempfile.py", line 401, in __getattr__
    file = self.__dict__['file']
KeyError: 'file'
```
Planing using tempfile to create a tmp dir for EventRecorder

boyuanzz:
Run Python PreCommit

boyuanzz:
@robertwb PTAL.Thanks for your help!

boyuanzz:
Run Portable_Python PreCommit

boyuanzz:
Run Python PreCommit

sdks/python/apache_beam/runners/worker/sdk_worker.py
robertwb: Yes, this is great. 

The only issue is that there is a limited number of workers, and we don't want to use them up (or wait for them) for progress/split calls. Given that these worker are stateless and threadsafe, perhaps a reasonable solution would be to create one extra worker that we always use for progress and split (but not for process or bundle finalization). 

Other than that, looks good.

boyuanzz: Setup a thread_pool with one thread for finalization call only: https://github.com/apache/beam/pull/7937/commits/c748ecfc43d11856720e843a115200323bc96586

boyuanzz:
Run Python PreCommit

sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -199,21 +195,26 @@ def task():
         "Currently using %s threads." % len(self._process_thread_pool._threads))
 
   def _request_process_bundle_split(self, request):
-    self._request_process_bundle_action(request)
+    self._request_process_bundle_action(request, self._progress_thread_pool)
+
+  def _request_finalize_bundle(self, request):
+    self._request_process_bundle_action(request, self._finalize_thread_pool)
```
robertwb: We can re-use process_thread_pool for this rather than creating another one.

sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -199,21 +195,26 @@ def task():
...
     def task():
       instruction_reference = getattr(
           request, request.WhichOneof('request')).instruction_reference
-      if instruction_reference in self._instruction_id_vs_worker:
-        self._execute(
-            lambda: self._instruction_id_vs_worker[
-                instruction_reference
-            ].do_instruction(request), request)
+      if instruction_reference not in self._unscheduled_process_bundle:
+        worker = self.workers.get()
```
robertwb: Having available workers here may still be a bottleneck, and the key idea is that we don't want progress/split calls to be blocked on process/finalize ones to complete. 

It should be sufficient to create one extra worker on line 101. Could you drop a TODO there about whether there is value in caching/re-using workers at all?

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
```diff
@@ -60,6 +60,12 @@ class FnApiRunnerTest(unittest.TestCase):
   def create_pipeline(self):
     return beam.Pipeline(runner=fn_api_runner.FnApiRunner())
 
+  def create_pipeline_with_grpc(self):
```
robertwb: We should not be requiring GRPC for these tests. If it only works with the GRPC runner, please move these test to FnApiRunnerTestWithGrpc with a JIRA to fix this.

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
```diff
@@ -801,6 +807,48 @@ def contains_labels(monitoring_info, labels):
...
+    event_recorder = EventRecorder(tempfile.gettempdir())
+    event_recorder.record('test')
+    elements_list = ['1', '2']
+
+    class FinalizebleDoFnWithSameRegister(beam.DoFn):
+      def process(
+          self,
+          element,
+          bundle_finalizer=beam.DoFn.BundleFinalizerParam):
+        bundle_finalizer.register(lambda: event_recorder.delete_file())
```
robertwb: Are you trying to test here that a failing finalizer does not fail the pipeline? If so, make that explicit via a finializer that raises an exception.

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
```diff
@@ -1083,6 +1131,34 @@ def __reduce__(self):
...
 
+class EventRecorder(object):
+  """Used to be registered as a callback in bundle finalization.
+
+  The reason why records are written into a tmp file is, the in-memory dataset
+  cannot keep callback records when passing into one DoFn.
+  """
+  def __init__(self, tmp_dir):
+    self.tmp_dir = tmp_dir
+    self.file_path = os.path.join(self.tmp_dir, uuid.uuid4().hex + '.txt')
```
robertwb: This is not safe to concurrent access. Instead, create a new (random) path like you're doing here in record(), and have events() list all files and return the contents of each, and shutil.rmtree to clean them all up.

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
```diff
@@ -1083,6 +1131,34 @@ def __reduce__(self):
...
+  def record(self, content):
+    tmp_file = open(self.file_path, 'a')
+    tmp_file.write(content + '\n')
+    tmp_file.close()
+
+  def events(self):
+    tmp_file = open(self.file_path, 'r')
+    content = tmp_file.read()
+    tmp_file.close()
+    self.delete_file()
```
robertwb: events() not being idempotent might be surprising.

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
```diff
@@ -1083,6 +1131,34 @@ def __reduce__(self):
...
+    return content
+
+  def delete_file(self):
```
robertwb: I would name this cleanup() to avoid leaking implementation details.

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
```diff
@@ -1083,6 +1131,34 @@ def __reduce__(self):
...
+
+  The reason why records are written into a tmp file is, the in-memory dataset
+  cannot keep callback records when passing into one DoFn.
+  """
+  def __init__(self, tmp_dir):
+    self.tmp_dir = tmp_dir
+    self.file_path = os.path.join(self.tmp_dir, uuid.uuid4().hex + '.txt')
+
+  def record(self, content):
+    tmp_file = open(self.file_path, 'a')
```
robertwb: Nit: preferable to use with statements for file. E.g.

with open(file_path, 'w') as f:
    f.write(...)

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
```diff
@@ -1083,6 +1131,34 @@ def __reduce__(self):
...
+    tmp_file.write(content + '\n')
+    tmp_file.close()
+
+  def events(self):
+    tmp_file = open(self.file_path, 'r')
+    content = tmp_file.read()
+    tmp_file.close()
+    self.delete_file()
+    return content
```
robertwb: Just a thought as to API design: it's nice to make EventRecorder encapsulate its implementation by providing a symmetric API. For example, if I do `record("a"); record("b"); record("c")` I would expect to get `["a", "b", "c"]` back when asking for what was recorded, rather than a single newline-delimited string.

sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -313,15 +314,18 @@ def process_bundle(self, request, instruction_id):
...
               process_bundle=beam_fn_api_pb2.ProcessBundleResponse(
                   residual_roots=delayed_applications,
                   metrics=bundle_processor.metrics(),
-                  monitoring_infos=bundle_processor.monitoring_infos()))
-      # TODO(boyuanz): Don't release here if finalize is needed.
-      self.bundle_processor_cache.release(instruction_id)
+                  monitoring_infos=bundle_processor.monitoring_infos(),
+                  requires_finalization=requests_finalization))
+      # Don't release here if finalize is needed.
+      if not bundle_processor.requires_finalization():
```
robertwb: For consistency, either re-use the `requests_finalization` variable here, or use the `bundle_processor.requires_finalization()` method above instead of a double return value.

sdks/python/apache_beam/runners/worker/sdk_worker.py
boyuanzz: With creating one more worker to handle finalize call sharing the process_thread_pool, it would be possible that the process is blocked by finalize, or the finalize is blocked by process(which seems to be OK).

boyuanzz: caching/re-using worker would be helpful for progress/split. We can still maintain a worker cache map only for progress/split request.

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
boyuanzz: If the current access means multi threads writes to the same file, I would prefer using lock, since If I create a new file for every record call, why there is a need to write content to that file?

sdks/python/apache_beam/runners/worker/sdk_worker.py
boyuanzz: Based on previous discussion: 
> The only issue is that there is a limited number of workers, and we don't want to use them up (or wait for them) for progress/split calls. Given that these worker are stateless and threadsafe, perhaps a reasonable solution would be to create one extra worker that we always use for progress and split (but not for process or bundle finalization).
> 
How about, creating one extra worker inside progress_pool to handle progress/split, creating one extra worker in process_pool to handle finalization. The ```self._worker_count``` will be number of workers processing bundle.

sdks/python/apache_beam/runners/worker/sdk_worker.py
boyuanzz: Just curious, why progress has its own thread_pool?

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
robertwb: If we want to run these tests on a distributed runner (including Flink/Spark/... on a local machine), there'll be no distributed lock to grab. 

You would write content to that file because the content comes from an external source, and you have not control over it other than that it's a bunch of bytes.

sdks/python/apache_beam/runners/worker/sdk_worker.py
robertwb: The intent here is that progress requests and splits should never get queued behind finishing a work item.

sdks/python/apache_beam/runners/worker/sdk_worker.py
robertwb: The concurrency is handled by two different caps, (1) the number of worker instances created and (2) the size of the threadpools. They don't really live in either pool (despite looking like they do, that's why I suggested a TODO to clean this relationship up). 

If we have number of workers >= size of process pool + size of progress pool, we should be fine. So, yes, let's allocate size of process pool + size of progress pool workers. It's fine for finalize to allocate slot(s) in the progress bundle pool.

sdks/python/apache_beam/runners/worker/sdk_worker.py
boyuanzz: Thanks for explanation. Though I still don't get why the progress_pool can serve for the intent you mentioned. But I think the previous instrucion_id_vs_worker is to make sure the progress/split can only call on an active bundle.

sdks/python/apache_beam/runners/worker/sdk_worker.py
boyuanzz: Implemented here: https://github.com/apache/beam/pull/7937/commits/a67e840c7208047f195de04f1c4deb280a4d97d4

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
boyuanzz: Fixed in https://github.com/apache/beam/pull/7937/commits/c5222da50865962311b5f475c9260864452acd0c

sdks/python/apache_beam/runners/worker/sdk_worker.py
boyuanzz: fixed here: https://github.com/apache/beam/pull/7937/commits/03e168102db43873e1170804df4e5f3afe6d0e8d

boyuanzz:

boyuanzz:
Turns out Flink runner doesn't support finalization. So I will skip my tests when running on flink: https://github.com/apache/beam/pull/7937/commits/5363f1e130a37d07951b87d00ec3a0ae62f3d75c

boyuanzz:
All tests have passed. Thanks for your time and help! @robertwb

PR #7939, opened 2019-02-25
[BEAM-6705] Fixes ConcurrentModificationException in RowCoderGenerator

mwalenia:
After running ValidatesRunner test suite for Direct Runner using Java 11 I ran into a ConcurrentModificationException in `RowCoderGenerator.generate` method. 

The cause of the exception was a race condition in a pipeline in `ParDoSchemaTest.testReadAndWrite`, `ParDoSchemaTest.testReadAndWriteMultiOutput` and `ParDoSchemaTest.testReadAndWriteWithSchemaRegistry`, where `RowCoderGenerator.generate` (and, as a result, `HashMap.computeIfAbsent`) was called with the same arguments from different threads. This resulted in tests failing with the exception.

Due to the fact that `HashMap.computeIfAbsent` now throws this exception if an external change is detected, it seems that the simplest solution is to change the map class to `ConcurrentHashMap`, which safeguards against such problems.

@reuvenlax, WDYT?  

------------------------

### DIFF
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/coders/RowCoderGenerator.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/coders/RowCoderGenerator.java
@@ -105,7 +105,7 @@ public abstract class RowCoderGenerator {
   private static final Map<TypeName, StackManipulation> CODER_MAP;
 
   // Cache for Coder class that are already generated.
-  private static Map<UUID, Coder<Row>> generatedCoders = Maps.newConcurrentMap();
+  private static Map<UUID, Coder<Row>> generatedCoders = Maps.newHashMap();
 
   static {
     // Initialize the CODER_MAP with the StackManipulations to create the primitive coders.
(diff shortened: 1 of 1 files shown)


mwalenia:
R: @reuvenlax

reuvenlax:
I ran into something similar, and tried to change to ConcurrentHashMap at
the time. Unfortunately I found that the change is not quite so simple,
because ConcurrentHashMap is not reentrant. In the case of nested Rows,
computeIfAbsent will recursively call generate() again, which could
deadlock everything.

I think the simplest fix to the deadlock would be to replace the recursive
call to generate() (in RowCoderGenerator::getCoder) with an instantiation
of the actual RowCoder class. RowCoder already lazily calls into
RowCoderGenerator the first time encode or decode is called, so this will
remove reentrant call from inside computeIfAbsent. Let me know if you need
some help making that change!

mwalenia:
Thanks for a fast reply! I think I need to understand this part a little bit better in order to modify it properly. If I instantiate the coder in RowCoderGenerator::getCoder, how would I make sure it gets cached in the RowCoderGenerator.generatedCoders map?

Would just calling `RowCoder.coderForFieldType(fieldType)` be sufficient?

reuvenlax:
Maybe a simpler approach for now would be to simply stop using computeIfAbsent, and replace it with an explicit check. Something like this:

Coder<Row> coder = generatedCoders.get(coderId);
if (coder = null) {
  coder = /* compute coder */
  generatedCoders.put(coderId, coder);
}
return coder;

This creates more of a chance of wasted work during a race condition, but will not cause a deadlock. Please add a short comment explaining that computeIfAbsent would deadlock in this situation.

mwalenia:
@reuvenlax Sorry, I force pushed a spotless fix. I reverted the map back to ConcurrentMap.

reuvenlax:
run Dataflow ValidatesRunner

reuvenlax:
run Flink ValidatesRunner

reuvenlax:
run Spark ValidatesRunner

PR #7940, opened 2019-02-25
[BEAM-6733] Use Flink 1.6/1.7 prepareSnapshotPreBarrier to replace BufferedOutputManager

mxm:
For Flink version <= 1.5 the Flink Runner has to buffer any elements which are
emitted during a snapshot because the barrier has already been emitted. This
leads to a lot of code complexity.

Flink version >= 1.6 provides a hook to execute an action before the snapshot
barrier is emitted by the operator. We can remove the buffering in favor of
finishing the current bundle in DoFnOperator's prepareSnapshotPreBarrier. The
1.5/1.6/1.7 build setup allows us to make this change with as little code
duplication as possible.


mxm:
R @tweise 
CC @aljoscha

runners/flink/1.6/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/DoFnOperator.java
```diff
@@ -0,0 +1,959 @@
...
+      elementCount = 0L;
+      lastFinishBundleTime = getProcessingTimeService().getCurrentProcessingTime();
+    }
+  }
+
+  @Override
+  public void prepareSnapshotPreBarrier(long checkpointId) {
+    // Finish the current bundle before the snapshot barrier is sent downstream
+    // This give us a clean state before taking the actual snapshot
+    invokeFinishBundle();
```
mxm: I should highlight that this is the new method used to finish the bundle before the snapshot barrier is emitted and `snapshot(..)` gets called.

runners/flink/1.6/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/DoFnOperator.java
```diff
@@ -0,0 +1,956 @@
...
+    this.currentOutputWatermark = currentOutputWatermark;
+  }
+
+  /** Factory for creating an {@link FlinkOutputManager} from a Flink {@link Output}. */
+  interface OutputManagerFactory<OutputT> extends Serializable {
+    FlinkOutputManager<OutputT> create(Output<StreamRecord<WindowedValue<OutputT>>> output);
+  }
+
+  /** A {@link DoFnRunners.OutputManager} that forwards data to the Flink runtime. */
+  public static class FlinkOutputManager<OutputT> implements DoFnRunners.OutputManager {
```
mxm: The `OutputManager` could be simplified to remove any buffering.

mxm:
Note that we could get rid of `FlinkSplitStateInternals` and `FlinkSplitStateInternalsTest` entirely for Flink > 1.6.

tweise:
Not sure I like duplicating DoFnOperator and then keeping multiple copies in sync. Can we defer this change until we remove 1.5.x support?

mxm:
There is only one copy. I didn't want to copy DoFnOperator but it proved hard to simplify it without a copy. The good thing is that we have tests in place for 1.5/1.6/1.7 which ensure that nothing breaks across these versions.

In my opinion this change gives us more insensitive to drop 1.5 support. Realistically, it will be hard to remain on multiple Flink versions without code diversion.

PR #7947, opened 2019-02-26
[BEAM-6731] Bundle Java ExpansionService with Flink JobServer 

mxm:
We need the Expansion Service to resolve cross-language transforms. This
integrates the service into Flink's JobServerDriver which now provides all services
to translate and execute pipelines.

CC @ibzib 

### DIFF
--- a/runners/java-fn-execution/build.gradle
+++ b/runners/java-fn-execution/build.gradle
@@ -58,3 +58,30 @@ task testDocker(type: Test) {
     includeCategories "org.apache.beam.runners.fnexecution.environment.testing.NeedsDocker"
   }
 }
+
+task runExpansionService (type: JavaExec) {
+  main = "org.apache.beam.runners.fnexecution.expansion.ExpansionService"
+  classpath = sourceSets.main.runtimeClasspath
+  args = [project.findProperty("constructionService.port") ?: "8097"]
+}
+
+task runTestExpansionService (type: JavaExec) {
+  main = "org.apache.beam.runners.fnexecution.expansion.TestExpansionService"
+  classpath = sourceSets.test.runtimeClasspath
+  args = [project.findProperty("constructionService.port") ?: "8097"]
+}
+
+task buildTestExpansionServiceJar(type: Jar) {
+  dependsOn = [shadowJar, shadowTestJar]
+  appendix = "testExpansionService"
+  // Use zip64 mode to avoid "Archive contains more than 65535 entries".
+  zip64 = true
+  manifest {
+    attributes(
+            'Main-Class': 'org.apache.beam.runners.fnexecution.expansion.TestExpansionService'
+    )
+  }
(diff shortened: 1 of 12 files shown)


sdks/python/build.gradle
```diff
@@ -210,7 +210,7 @@ task portableWordCount {
 def portableWordCountTask(name, streaming, crossLanguage) {
   tasks.create(name) {
     dependsOn = ['installGcpTest']
-    mustRunAfter = [':beam-runners-flink_2.11-job-server-container:docker', ':beam-sdks-python-container:docker', ':beam-sdks-java-container:docker', ':beam-runners-core-construction-java:buildTestExpansionServiceJar']
+    mustRunAfter = [':beam-runners-flink_2.11-job-server-container:docker', ':beam-sdks-python-container:docker', ':beam-sdks-java-container:docker', ':beam-runners-java-fn-execution:buildTestExpansionServiceJar']
```
tweise: We can address that separately, but I don't think this task should be tied to building the java container or expansion service jar. If the expansion service is a job server dependency, isn't that redundant here anyways?

runners/java-fn-execution/build.gradle
```diff
@@ -58,3 +58,30 @@ task testDocker(type: Test) {
     includeCategories "org.apache.beam.runners.fnexecution.environment.testing.NeedsDocker"
   }
 }
+
+task runExpansionService (type: JavaExec) {
```
tweise: +1 for relocating this, it is consistent with artifact staging

sdks/python/build.gradle
mxm: Yes, can address this next. It is actually still required because the `TestExpansionService` builds a new jar which contains the test dependencies which would normally not be part of the jar. I did not change the test setup for now.

runners/java-fn-execution/build.gradle
chamikaramj: IIUC java-fn-execution module is for runner specific code related to Fn API execution (job submission to job server, work execution) while core-construction-java contains code related to pipeline construction before submitting job to a runner, right ?

If so, I'm not sure if moving expansion service to the java-fn-execution module is the correct move. It might be weird (or impossible) to maintain expansion service in java-fn-execution while maintaining other SDK logic related to cross-language transform expansion in pipeline construction side.

If the need is to just package expansion service withe the job server offered by the runner, we should be able to do that without moving the expansion service, right ? (given that java-fn-execution depends on core-construction-java)

cc: @ihji @robertwb

mxm: I agree with your arguments. I thought that `fn-execution` might be the better place because other services also live in this module. However, the `ExpansionService` is clearly related to the construction of the pipeline, not its execution. I'll prepare a PR to move it back.

chamikaramj: Thanks Max. Sorry about not commenting before the PR went in.

PR #7952, opened 2019-02-27
[BEAM-6753] Set the stage to make schema coder update compatible

reuvenlax:
Adds a proto representation of schemas.
For Dataflow, embeds this proto in the SchemaCoder representation.
Adds a notion of encoding location, no longer tied to id (not yet hooked into RowCoder.encode). 
Length prefix RowCoder

### DIFF
--- a/model/pipeline/src/main/proto/beam_runner_api.proto
+++ b/model/pipeline/src/main/proto/beam_runner_api.proto
@@ -676,7 +676,6 @@ message Schema {
       Schema row_schema = 5;
       LogicalType logical_type = 6;
     }
-    map<string, bytes> metadata = 7;
   }
 
   message Field {
--- a/model/pipeline/src/main/proto/beam_runner_api.proto
+++ b/model/pipeline/src/main/proto/beam_runner_api.proto
@@ -635,6 +635,62 @@ message StandardCoders {
   }
 }
 
+// Experimental: A representation of a Beam Schema.
+message Schema {
+  enum TypeName {
+    BYTE = 0;
+    INT16 = 1;
+    INT32 = 2;
+    INT64 = 3;
+    DECIMAL = 4;
+    FLOAT = 5;
+    DOUBLE = 6;
+    STRING = 7;
+    DATETIME = 8;
+    BOOLEAN = 9;
+    BYTES = 10;
+    ARRAY = 11;
+    MAP = 13;
+    ROW = 14;
+    LOGICAL_TYPE = 15;
+  }
+
+  message LogicalType {
+    string id = 1;
+    string args = 2;
+    FieldType base_type = 3;
+    bytes serialized_class = 4;
+  }
+
+  message MapType {
+    FieldType key_type = 1;
+    FieldType value_type = 2;
+  }
+
+  message FieldType {
+    TypeName type_name = 1;
+    bool nullable = 2;
+    oneof type_info {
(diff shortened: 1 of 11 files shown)


reuvenlax:
Run Dataflow ValidatesRunner

reuvenlax:
Run Dataflow ValidatesRunner

reuvenlax:
@dpmills

reuvenlax:
R: @dpmills

sdks/java/core/src/main/java/org/apache/beam/sdk/coders/RowCoderGenerator.java
```diff
@@ -285,15 +289,24 @@ public InstrumentedType prepare(InstrumentedType instrumentedType) {
     // per-field Coders.
     static Row decodeDelegate(Schema schema, Coder[] coders, InputStream inputStream)
         throws IOException {
+      int fieldCount = VAR_INT_CODER.decode(inputStream);
```
dpmills: Can we leave this out by just using the length of the BitSet?

reuvenlax: Unfortunately no. BitSet.size() returns the total space used, which rounds up to the next word size.

reuvenlax: as discussed, we don't want to have to reorder fields in the encoding.

dpmills:
LGTM

reuvenlax:
Run Dataflow ValidatesRunner

reuvenlax:
Run Java PreCommit

reuvenlax:
Run Python PreCommit

reuvenlax:
Run Portable_Python PreCommit

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -635,6 +635,61 @@ message StandardCoders {
...
+  enum TypeName {
+    BYTE = 0;
+    INT16 = 1;
+    INT32 = 2;
+    INT64 = 3;
+    DECIMAL = 4;
+    FLOAT = 5;
+    DOUBLE = 6;
+    STRING = 7;
+    DATETIME = 8;
```
kennknowles: Adding schemas to portability needs a design doc and discussion on dev@. Did I miss that?

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -635,6 +635,61 @@ message StandardCoders {
...
+    BOOLEAN = 9;
+    BYTES = 10;
+    ARRAY = 11;
+    MAP = 13;
```
kennknowles: Notably, you don't need this. `FieldType` has a oneof and one of the alternatives can be an atomic type and the others are map, row, array, logical. I would recommend breaking out more types in this oneof, for example a BytesTypes we may likely want metadata indicating a coder. And we discussed - and I think should go ahead with - removing datetime as a primitive type.

reuvenlax:
FYI this was added right now simply for the DataflowRunner; we used the
portability protos for convenience (and because that's where it will go
eventually anyway). pushing things into portability for real indeed does
require a published design first, which is why the portability runners
currently ignore this.

PR #7949, opened 2019-02-26
[BEAM-3713] Add pytest testing infrastructure

udim:
Initial step towards migrating off of nosetests.

- Runs integration tests via tox: `tox -e
py27-gcp-pytest,py36-pytest,etc.`
- `@pytest.mark.{pre,post}commit` replaces `@attr('IT')`
  - Also can mark tests as: no_dataflow, no_direct, etc. to exclude
  tests that don't pass on certain runners.
- conftest.py replaces BeamTestPlugin
- Jenkins reports for all tests run (not just the last nosetests
    invocation), using multiple XML outputs.
- Allows specifying test module in tox cmd line. Example:
```sh
tox -e py27-pytest apache_beam.transforms.window_test
```

Still needs:
- Verification that all the tests that run on nose also run
on pytest.

Nice to have:
- Parallelization, but this requires fixing pickling issues with the
pytest plugins used for this.

------------------------

Thank you for your contribution! Follow this checklist to help us incorporate your contribution quickly and easily:

 - [ ] [**Choose reviewer(s)**](https://beam.apache.org/contribute/#make-your-change) and mention them in a comment (`R: @username`).
 - [x] Format the pull request title like `[BEAM-XXX] Fixes bug in ApproximateQuantiles`, where you replace `BEAM-XXX` with the appropriate JIRA issue, if applicable. This will automatically link the pull request to the issue.
 - [ ] If this contribution is large, please file an Apache [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

Lang | SDK | Apex | Dataflow | Flink | Gearpump | Samza | Spark
--- | --- | --- | --- | --- | --- | --- | ---
Go | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go/lastCompletedBuild/) | --- | --- | --- | --- | --- | ---
[...]

PR #7953, opened 2019-02-27
[BEAM-1251] Fix Python 3 compatibility bug in pickler.py

udim:
This issue was discovered while testing the pytest framework using Python 3 in #7949.

sys.modules changes during iteration. It seems that the intention was to
make a copy using sys.modules.values(), which returns a copy in Python
2, but a view in Python 3.

The issue comes up when using pytest, which implements lazy module
loading.

------------------------

### DIFF
--- a/sdks/python/apache_beam/internal/pickler.py
+++ b/sdks/python/apache_beam/internal/pickler.py
@@ -164,6 +164,17 @@ if 'save_module' in dir(dill.dill):
     obj_id = id(obj)
     if not known_module_dicts or '__file__' in obj or '__package__' in obj:
       if obj_id not in known_module_dicts:
+        # Trigger loading of lazily loaded modules (such as pytest vendored
+        # modules).
+        # This first pass over sys.modules needs to iterate on a copy of
+        # sys.modules since lazy loading modifies the dictionary, hence the use
+        # of list().
+        for m in list(sys.modules.values()):
+          try:
+            _ = m.__dict__
+          except AttributeError:
+            pass
+
         for m in sys.modules.values():
           try:
             if (m


udim:
R: @aaltay 
CC: @tvalentyn , @charlesccychen

sdks/python/apache_beam/internal/pickler.py
```diff
@@ -164,7 +164,7 @@ def new_save_module_dict(pickler, obj):
     obj_id = id(obj)
     if not known_module_dicts or '__file__' in obj or '__package__' in obj:
       if obj_id not in known_module_dicts:
-        for m in sys.modules.values():
+        for m in list(sys.modules.values()):
```
tvalentyn: Do we understand why does sys.modules.values() change during iteration? Is it a side-effect of pytest?

udim:
run python postcommit

sdks/python/apache_beam/internal/pickler.py
udim: Probably, though I don't understand why it sys.modules changes in this loop (assuming no background threads). I believe that accessing `m.__dict__` triggers a lazy loading of the module.

I tested adding an additional for-loop before the original. The first loop iterates on list(sys.modules.value()), while the second on sys.modules.value(). The first loop conditionally accesses m.__dict__ to see which modules change sys.modules. If the second loop succeeds, the touched modules are culprits.

I've isolated this module as a culprit:
py._vendored_packages.apipkg.ApiModule

Maybe related:
pytest has a module import hook for rewriting assertions (no need for self.assertEqual(), etc.):
https://github.com/pytest-dev/pytest/blob/7dcd9bf5add337686ec6f2ee81b24e8424319dba/src/_pytest/assertion/rewrite.py#L277-L301

udim: https://github.com/pytest-dev/py/tree/master/py/_vendored_packages
https://github.com/pytest-dev/py/blob/master/py/__init__.py

So an alternative solution would be to add an additional loop before the original one:
```py
_ = [m.__dict__ for m in list(sys.modules.values())]

for m in sys.modules.values():
  ...
```

udim: @robertwb

udim: @charlesccychen I believe my above suggestion is the more correct one, allowing lazily-loaded modules to initialize before putting their contents in known_module_dicts.

charlesccychen: Thanks @udim--that makes sense.  Are you going to update this change?

udim:
run python postcommit

sdks/python/apache_beam/internal/pickler.py
udim: @charlesccychen updated, ready to merge once all tests pass

charlesccychen:
Thanks, this LGTM.

sdks/python/apache_beam/internal/pickler.py
tvalentyn: Is `list()` essential in line 168? If so please add a comment so that somebody doesn't try to clean it up, otherwise we can remove it.

udim: Right, added more commentary

udim:
run python postcommit

charlesccychen:
Thanks!

PR #7955, opened 2019-02-27
[BEAM-6751] Extend Kafka EOS mode whitelist / Warn instead of throw

mxm:
The FlinkRunner supports exactly-once checkpointing but is blocked from using
Kafka's exactly once mode. The list is hard to maintain and we should rather
warn than throw in case of unsupported Runners. We generally have to rely on the
Runner doing its work correctly and such checks do not prevent bad things from
happening.

### DIFF
--- a/sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
+++ b/sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
@@ -1141,10 +1141,11 @@ public class KafkaIO {
         String runner = options.getRunner().getName();
         if ("org.apache.beam.runners.direct.DirectRunner".equals(runner)
             || runner.startsWith("org.apache.beam.runners.dataflow.")
-            || runner.startsWith("org.apache.beam.runners.spark.")) {
+            || runner.startsWith("org.apache.beam.runners.spark.")
+            || runner.startsWith("org.apache.beam.runners.flink.")) {
           return;
         }
-        throw new UnsupportedOperationException(
+        LOG.warn(
             runner
                 + " is not whitelisted among runners compatible with Kafka exactly-once sink. "
                 + "This implementation of exactly-once sink relies on specific checkpoint guarantees. "


kennknowles:
Whoa, this is kind of an exciting line of code. I had no idea we were doing stuff like this. I agree with your meta-point. We shouldn't have library transforms introspecting the runner _at all_.

But I don't think I agree about Flink's support. This message seems coded, but I assume it means that the transform (like a huge number of IOs) depends on GroupByKey/Reshuffle providing stable input to the downstream transform. This should be replaced by `@RequiresStableInput`, which as I understand it might require waiting for a Flink checkpoint finalization callback or some such, no?

mxm:
You are correct, Flink's KafkaProducer starts a new transaction for every checkpoint and only acknowledges the transaction upon completion of the checkpoint. However, we do not replace KafkaIO with Flink's KafkaProducer because it seems impracticable to map its configuration to Flink's.

After taking a closer look at Beam's `KafkaExactlyOnceSink` it does not look like we can achieve exactly-once semantics with the Flink Runner. We can potentially process elements multiple times if we restore from a checkpoint. 

That is unfortunate because Flink has all the building blocks to ensure exactly once. Also hard to sell to Beam users that are used to exactly once in Flink's KafkaProducer.

A way to fix this would be to provide an (optional) hook in DoFn for committing pending work. I suppose SDF would have something like this. Alternatively, we might have to consider translating KafkaIO to Flink's KafkaProducer.

kennknowles:
Processing an element again is unavoidable in the face of failure, so do you mean that the resulting processing will cause duplicate (non-idempotent) externally visible effects?

I just read up on [Flink's exactly-once KafkaProducer](https://ci.apache.org/projects/flink/flink-docs-stable/dev/connectors/kafka.html) so I can contribute more effectively to this conversation.

There, it reads "during normal working of Flink applications, user can expect a delay in visibility of the records produced into Kafka topics, equal to average time between completed checkpoints." That is exactly the expected cost of implementing `@RequiresStableInput`. I had thought this was untenable for Flink because my (out of date) understanding was that in large scale use checkpoints were set to be exceedingly rare.

The historical cause, which I think you know, is that many Beam IOs were designed before Beam and/or around Dataflow, where you can do things like generate random numbers, shuffle to save the result, then use them and know that they will not be regenerated upon retry. The SparkRunner explicitly materializes after GBK to simulate this (at great cost, I would imagine). The mindset is hard to shake, and lots of people (myself included) don't have a deep understanding of all runners. We probably need to have an explicit push to fix Beam's IOs. It would be great to have a generic way to produce integration tests with failures that could be applied to any runner, to catch erroneous assumptions around durability.

I don't quite understand your last point, about having a callback on a `DoFn` to commit pending work. Theres `@FinishBundle` which must be called before elements are considered processed. I think you mean kind of the converse, like an `OK now it is durable so you can output other things` callback? I believe this is more-or-less what `@RequiresStableInput` achieves, or anyhow what it is hoping to achieve. I can't tell if this problem is fundamental and the model needs an extension (which I'm super OK with) or if it is just the way IOs are written.

kennknowles:
@chamikaramj @rangadi @mxm I wonder if this is now too big for this PR and belongs on the dev@ list. I know there have been prior discussions but I'd like to have a good agreed-upon plan.

mxm:
Thank you for your reply. I was actually in the process of writing an email to the dev list :) Will send it in a bit.

>Processing an element again is unavoidable in the face of failure, so do you mean that the resulting processing will cause duplicate (non-idempotent) externally visible effects?

As it stands, writes are not guaranteed to be idempotent with the FlinkRunner. It makes sense that Spark materializes the GroupByKey, I guess Dataflow does something similar. Flink has a different approach. It sends checkpoint barriers through all channels which allows to take a snapshot of the running pipeline, but it does require a callback after checkpoint completion to commit to Kafka.

>There, it reads "during normal working of Flink applications, user can expect a delay in visibility of the records produced into Kafka topics, equal to average time between completed checkpoints." That is exactly the expected cost of implementing `@RequiresStableInput`.

That is the time between Kafka transaction start and transaction end. The transaction is completed after the checkpoint is complete. 

>The historical cause, which I think you know, is that many Beam IOs were designed before Beam and/or around Dataflow, where you can do things like generate random numbers, shuffle to save the result, then use them and know that they will not be regenerated upon retry

It seems like that is not true for `KafkaIO` because it is not as old as the other transforms, though it might have been inspired by the old transforms.

>I don't quite understand your last point, about having a callback on a `DoFn` to commit pending work. Theres `@FinishBundle` which must be called before elements are considered processed. I think you mean kind of the converse, like an `OK now it is durable so you can output other things` callback?

Flink's KafkaProducer 1) pushes all pending items to Kafka, 2) checkpoints, 3) commits when the checkpoint has completed. To do something similar with Beam, we need to be able to tell the Kafka DoFn that it can now commit the pending Kafka transaction. But since the commit is hidden inside `processElement`, we do not currently have control over it.

rangadi:
@mxm, warning implies that the transform would run, but not provide EOS guarantees on Flink, right? Is that ok? IOW, EOS is ignored?

The incompatibility with Flink was discussed on dev list in Aug 2017 when we added EOS. I am familiar with 2PC commit utility in Flink that was added around that same time to Flink to support EOS for Kafka produced. I had commented on those PRs that time. @mxm, warning implies that the transform would run, but not provide EOS guarantees on Flink, right? Is that ok? IOW, EOS is ignored?

We (including Aljoscha) could not see how 2PC model could be implemented for Flink in Beam context. See https://www.mail-archive.com/dev@beam.apache.org/msg02664.html. This is due to fundamental difference in how Flink's horizontal checkpointing differs from per-stage checkpointing in Dataflow other. 

 If anything has changed in that respect.

@kennknowles, this was added even before @RequiresStableInput was finalized in Beam. If there is better way to achieve the same guarantees, we should improve the implementation. I agree, checking the runner class is a hack, but we thought that is an acceptable compromise compared to not implementing EOS at all. This was in Aug 2017, if there is better way to do it, I don't mind reimplementing it (or work with another contributor to reimplement).

mxm:
>@mxm, warning implies that the transform would run, but not provide EOS guarantees on Flink, right? Is that ok? IOW, EOS is ignored?

Yes, we cannot guarantee EOS with the current architecture. I'm not too fond of this transform check but it does reflect the situation accurately. I think it would be better to check whether the Runner handles `RequiresStableInput` instead. 

We have started discussing this on the mailing: https://lists.apache.org/thread.html/d0c24cb4d421e25980fee1855d1daa68d343700270ed962d642c8f99@%3Cdev.beam.apache.org%3E

I'll leave the check in place until we find a solution. Closing the PR for now.

PR #7956, opened 2019-02-27
[BEAM-6670] Add option to disable reshuffling of JdbcIO

noctune:
Adds a `withReshuffle` method on JdbcIO classes so reshuffling can be disabled.

This can be useful when an operator immediately after `JdbcIO.Read` reshuffles anyway. In my case I am first reading 20 or so tables and then joining them using a `CoGroupByKey`. Since the CoGroupByKey operation would distribute the data anyway, the work to reshuffle the 20 tables is wasted and results in a much more complex graph than necessary.

There are no tests for this, as reshuffling is not easily testable as it should practically have "no effect". However, there are no tests for the original reparallelization logic either.

------------------------

### DIFF
--- a/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcIO.java
+++ b/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcIO.java
@@ -160,7 +160,10 @@ public class JdbcIO {
    * @param <T> Type of the data to be read.
    */
   public static <T> Read<T> read() {
-    return new AutoValue_JdbcIO_Read.Builder<T>().setFetchSize(DEFAULT_FETCH_SIZE).build();
+    return new AutoValue_JdbcIO_Read.Builder<T>()
+        .setFetchSize(DEFAULT_FETCH_SIZE)
+        .setOutputParallelization(true)
+        .build();
   }
 
   /**
@@ -173,6 +176,7 @@ public class JdbcIO {
   public static <ParameterT, OutputT> ReadAll<ParameterT, OutputT> readAll() {
     return new AutoValue_JdbcIO_ReadAll.Builder<ParameterT, OutputT>()
         .setFetchSize(DEFAULT_FETCH_SIZE)
+        .setOutputParallelization(true)
         .build();
   }
 
@@ -399,6 +403,8 @@ public class JdbcIO {
 
     abstract int getFetchSize();
 
+    abstract boolean getOutputParallelization();
+
     abstract Builder<T> toBuilder();
 
     @AutoValue.Builder
@@ -414,6 +420,8 @@ public class JdbcIO {
       abstract Builder<T> setCoder(Coder<T> coder);
 
       abstract Builder<T> setFetchSize(int fetchSize);
+
+      abstract Builder<T> setOutputParallelization(boolean outputParallelization);
 
       abstract Read<T> build();
     }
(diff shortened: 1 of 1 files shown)


aromanenko-dev:
@jkff As an author of this original improvement [BEAM-2803](https://issues.apache.org/jira/browse/BEAM-2803) (side input and reshuffle), could you take a look on this PR? Thanks

jkff:
Thank you, however I am currently not active on Beam so I don't have time to do a full review.
R: @chamikaramj perhaps?

Quick note though - generally we avoid adding any kind of tuning parameters unless there's a very, very good reason (because people almost always end up setting them to a bad value, or to a value that was good at the time it was set but is no longer good and nobody updates it) - the standard we've been using is "some pipeline behaves catastrophically bad in a way that can only be prevented by adding the tuning parameter, and it can not be chosen automatically". Even when this standard is met, it is better to let the user specify hints in declarative rather than operational terms - e.g. how TextIO.read() has a "withHintMatchesManyFiles()" parameter rather than "withHintGoViaReadAll()" (which is what that parameter currently actually does).

It would help if you could e.g. elaborate just how much worse is the performance you observe without specifying this parameter in your pipeline.

Noctune:
I did sort of expect this to be "controversial". :)

>"some pipeline behaves catastrophically bad in a way that can only be prevented by adding the tuning parameter, and it can not be chosen automatically".

Admittedly, it is not fair to say that the presence of reshuffle is *catastrophic*, but it currently does approx. twice as much IO as necessary. I have not had the chance to test it without the reshuffle, but I will see if I can find some time next week.

It also clutters the graph up a lot, which is a problem in Flink as the GUI does *not* perform well for large graphs, so I generally try to minimize the size of the graph.

Although, I suppose the situation I describe could be detected automatically by inspecting the graph and determining if the `JdbcIO` leads directly to a group by. A more general optimization could be to remove sequential reshuffles automatically and have some method of marking certain internal transforms as "cheap" (like constructing `RawUnionValue`s for a `CoGroupByKey`). I.e. merge reshuffles if they are only separated by "cheap" transforms.

 >Even when this standard is met, it is better to let the user specify hints in declarative rather than operational terms

Yeah, I thought about that. I could not really figure out a more suitable name. `withHintDirectlyToGroupBy`, maybe?

aromanenko-dev:
The same, I agree that we should avoid additional "knobs" and we do our best for that, but about 2x performance improvement seems as a reasonable argument for me to add this "knob". 
I propose to add such feature but we could call it `withForcedFusionBreak()` which will be `true` by default and it can be turned off  in case of reshuffle phase after.

noctune:
> I propose to add such feature but we could call it `withForcedFusionBreak()` which will be `true` by default and it can be turned off in case of reshuffle phase after.

Yeah, that seems like a better name. `withHintFusionBreak`, maybe? Makes it clear that it is only a hint and will maybe be ignored some time in the future.

chamikaramj:
Usually, we do not need to add a Reshuffle at the end of read transforms since these transforms usually support parallelizing the read itself (through initial splitting or dynamic work rebalancing). I think JdbcIO is special in this regard since we are running a query from a single worker with no parallelization at all. So most users will benefits from a Reshuffle here but some users might not need it. So I'm fine with adding a hint that will allow disabling this (while keeping reshuffle as the default). For the name, how about withoutOutputParallization(), to denote that this will keep the read operation in a single worker.

Also, please consider raising this in dev list to get other's opinion.

iemejia:
I don't think this deserves the discussion to go to the mailing list. Most of us agree that this is an exceptional case where it makes sense to have this option too, and the performance consequence for the mentioned case is worth the knob. Of course not as a default.
I like @chamikaramj's proposal better, more user friendly, no fancy names. but I will refine it to`withOutputParallelization`. for consistency if it keeps the boolean argument. or without if there is no argument.

@noctune can you please fix this and rebase it so we can merge it. Thanks!

aromanenko-dev:
I like the proposed name by @chamikaramj but I'd prefer1 @iemejia idea to keep a consistency with naming. 
So, my **+1** to `withOutputParallelization()` and `withOutputParallelization(boolean)` where default value is `true`.

chamikaramj:
withOutputParallelization(boolean) with default set to true sounds good :)

iemejia:
Ok we have an agreement :+1:

noctune:
Renamed and rebased. Two test fail, but they seem unrelated.

iemejia:
Run Java PreCommit

PR #7959, opened 2019-02-27
Fixing Python tornadoes test for Google

pabloem:
The tornadoes test, when run from DirectRunner needs a GCS location. When running on Dataflow, it should just use the GCS-based temp location of the pipeline.

r: @markflyhigh

### DIFF
--- a/sdks/python/apache_beam/examples/cookbook/bigquery_tornadoes.py
+++ b/sdks/python/apache_beam/examples/cookbook/bigquery_tornadoes.py
@@ -87,7 +87,7 @@ def run(argv=None):
     rows = p | 'read' >> beam.io.Read(beam.io.BigQuerySource(known_args.input))
     counts = count_tornadoes(rows)
 
-    if p.options.get_all_options()['temp_location'].startswith('gs://'):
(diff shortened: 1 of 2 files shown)


pabloem:
Run Python PostCommit

sdks/python/apache_beam/examples/cookbook/bigquery_tornadoes.py
```diff
@@ -87,14 +87,19 @@ def run(argv=None):
     rows = p | 'read' >> beam.io.Read(beam.io.BigQuerySource(known_args.input))
     counts = count_tornadoes(rows)
 
+    if p.options.get_all_options()['temp_location'].startswith('gs://'):
```
markflyhigh: Seems temp_location has to be a GCS path (https://github.com/apache/beam/blob/master/sdks/python/apache_beam/options/pipeline_options.py#L426). May not need `gs://` check.

pabloem:
Fixed that. Thanks Mark.

pabloem:
Run Python PostCommit

markflyhigh:
LGTM to me. Thanks

PR #7984, opened 2019-03-04
[BEAM-6754] Use subprocess instead of threads in loopback environment

angoenka:
To Take advantage of multi cores, we need to start separate processes to for worker harness.
Using subprocess instead of threads to run the sdk harness.

------------------------

### DIFF
--- a/sdks/python/apache_beam/runners/portability/portable_runner.py
+++ b/sdks/python/apache_beam/runners/portability/portable_runner.py
@@ -17,11 +17,13 @@
 
 from __future__ import absolute_import
 
+import atexit
 import functools
 import itertools
 import json
 import logging
 import os
+import subprocess
 import threading
 import time
 from concurrent import futures
@@ -160,9 +162,13 @@ class PortableRunner(runner.PipelineRunner):
     # This is needed as we start a worker server if one is requested
     # but none is provided.
     if portable_options.environment_type == 'LOOPBACK':
+      use_loopback_process_worker = options.view_as(
+          DebugOptions).lookup_experiment(
+              'use_loopback_process_worker', False)
       portable_options.environment_config, server = (
           BeamFnExternalWorkerPoolServicer.start(
-              sdk_worker_main._get_worker_count(options)))
+              sdk_worker_main._get_worker_count(options),
+              use_process=use_loopback_process_worker))
       globals()['x'] = server
       cleanup_callbacks = [functools.partial(server.stop, 1)]
     else:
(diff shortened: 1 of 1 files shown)


angoenka:
R: @aaltay @mxm @robertwb

angoenka:
Run Portable_Python PreCommit

sdks/python/apache_beam/runners/portability/portable_runner.py
```diff
@@ -445,15 +446,22 @@ def start(cls, worker_threads=1):
...
-      worker_thread.start()
+      command = ['python', '-c',
+                 'from apache_beam.runners.worker.sdk_worker '
+                 'import SdkHarness as SH; '
+                 'SH("%s",worker_count=%d,worker_id="%s").run()' % (
+                     start_worker_request.control_endpoint.url,
+                     self._worker_threads,
+                     start_worker_request.worker_id), ]
+      logging.warn("Starting worker with command %s" % (command))
+      process = subprocess.Popen(command, stdout=subprocess.PIPE, )
```
mxm: intentional `, `?

sdks/python/apache_beam/runners/portability/portable_runner.py
```diff
@@ -445,15 +446,22 @@ def start(cls, worker_threads=1):
...
+
+      def kill_subprocess():
+        process.kill()
```
mxm: Perhaps forcefully kill if it does not exit?

sdks/python/apache_beam/runners/portability/portable_runner.py
```diff
@@ -445,15 +446,22 @@ def start(cls, worker_threads=1):
...
-          worker_id=start_worker_request.worker_id)
-      worker_thread = threading.Thread(
-          name='run_worker_%s' % start_worker_request.worker_id,
-          target=worker.run)
-      worker_thread.daemon = True
-      worker_thread.start()
+      command = ['python', '-c',
+                 'from apache_beam.runners.worker.sdk_worker '
+                 'import SdkHarness as SH; '
+                 'SH("%s",worker_count=%d,worker_id="%s").run()' % (
```
mxm: The renaming here is odd. Assuming this is because of the line length. Just keep this `Sdkharness`?

sdks/python/apache_beam/runners/portability/portable_runner.py
```diff
@@ -445,15 +446,22 @@ def start(cls, worker_threads=1):
...
+                     start_worker_request.control_endpoint.url,
+                     self._worker_threads,
+                     start_worker_request.worker_id), ]
```
mxm: Remove `, `?

robertwb:
Agree we should use multiprocessing. We should also consider making this configurable (especially as loopback is used for lightweight testing.)

sdks/python/apache_beam/runners/portability/portable_runner.py
angoenka: With python, adding a trailing `'` in list is conventional but I think we don't do it in beam.
Removing it.

sdks/python/apache_beam/runners/portability/portable_runner.py
angoenka: There is no parameter for forceful kill. I suppose the kill method will ensure that the subprocress is killed.

angoenka:
> Agree we should use multiprocessing. We should also consider making this configurable (especially as loopback is used for lightweight testing.)

Multiprocessing doesn't seem to work with grpc.

Making it configurable seems right. Using experimental flag which defaults to subprocess.

robertwb:
I don't think the default should be subprocess. More futureproof might be to have a separate numprocesses setting which if 1 would use a threadpool and for > 1 use subprocess.

angoenka:
> I don't think the default should be subprocess. More futureproof might be to have a separate numprocesses setting which if 1 would use a threadpool and for > 1 use subprocess.

Interesting idea. I think that changing the behavior based on number of process (1 and >1) might be bit unintuitive for the user. Also even with 1 process, multiple threads will be created if runner request for multiple SDKHarness.

robertwb:
Fair enough.

I still think it should be opt-in, as the primary advantage of loopback is
that it's as lightweight as possible.

mxm:
Let's keep Python threads as the default for now. We have not yet experimented enough with processes yet.

angoenka:
Sounds good, updated the default to use thread.

PR #7989, opened 2019-03-05
[BEAM-6760] Reducing concurrency in website checks

pabloem:


pabloem:
The website test, it turns out, is real good at finding ways to fail : )

pabloem:
It seems that we just needed to reduce the maximum concurrency and leave the timeout to be its default value (300 seconds). Website tests don't have the problem of having to run to quickly, so I believe this change is acceptable.

r: @alanmyrvold

PR #7828, opened 2019-02-13
Switching google_api_services_dataflow to v1b3-rev20190126

drieber:
Switching google_api_services_dataflow to rev20190126.

------------------------

Thank you for your contribution! Follow this checklist to help us incorporate your contribution quickly and easily:

 - [ ] [**Choose reviewer(s)**](https://beam.apache.org/contribute/#make-your-change) and mention them in a comment (`R: @username`).
 - [ ] Format the pull request title like `[BEAM-XXX] Fixes bug in ApproximateQuantiles`, where you replace `BEAM-XXX` with the appropriate JIRA issue, if applicable. This will automatically link the pull request to the issue.
 - [ ] If this contribution is large, please file an Apache [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

Lang | SDK | Apex | Dataflow | Flink | Gearpump | Samza | Spark
--- | --- | --- | --- | --- | --- | --- | ---
Go | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go/lastCompletedBuild/) | --- | --- | --- | --- | --- | ---
Java | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Apex/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Dataflow/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Flink/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Flink_Batch/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Flink_Streaming/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Gearpump/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Samza/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Spark/lastCompletedBuild/)
[...]

PR #7997, opened 2019-03-06
Revert "StreamingDataflowWorker uses the per-stage transformUserNameT…

pabloem:
…oStateFamily."

This reverts commit 6571c83cfcf6ce3f21414719841f17617bec7724.

I'm hoping I reverted properly.

Passing streaming engine job: https://pantheon.corp.google.com/dataflow/jobsDetail/locations/us-central1/jobs/2019-03-05_18_06_06-14767468226197030508?project=apache-beam-testing


drieber:
I introduced the transform_user_name_to_state_family field for Streaming Engine in 12/26/2018. That made it to the service release in January.
How come we are running with a proto jar that does not contain that field?

drieber:
Is https://github.com/apache/beam/pull/7828 somehow wrong or related?

PR #8000, opened 2019-03-06
[BEAM-6722] Raise embedded Cassandra startup timeout to avoid flakiness in Cassandra tests and avoid NPE in case of timeout

echauchot:

------------------------

### DIFF
--- a/sdks/java/io/cassandra/src/test/java/org/apache/beam/sdk/io/cassandra/CassandraIOTest.java
+++ b/sdks/java/io/cassandra/src/test/java/org/apache/beam/sdk/io/cassandra/CassandraIOTest.java
@@ -126,9 +126,15 @@ public class CassandraIOTest implements Serializable {
 
   @AfterClass
   public static void stopCassandra() {
-    EmbeddedCassandraServerHelper.cleanEmbeddedCassandra();
-    session.close();
-    cluster.close();
+    if (cluster != null && session != null) {
+      EmbeddedCassandraServerHelper.cleanEmbeddedCassandra();
+      session.close();
+      cluster.close();
+    } else {
+      if (cluster != null) {
+        cluster.close();
+      }
+    }
   }
 
   @Before
--- a/sdks/java/io/cassandra/src/test/java/org/apache/beam/sdk/io/cassandra/CassandraIOTest.java
+++ b/sdks/java/io/cassandra/src/test/java/org/apache/beam/sdk/io/cassandra/CassandraIOTest.java
@@ -101,7 +101,7 @@ public class CassandraIOTest implements Serializable {
   public static void startCassandra() throws Exception {
     System.setProperty("cassandra.jmx.local.port", JMX_PORT);
     startupTime = System.currentTimeMillis();
-    EmbeddedCassandraServerHelper.startEmbeddedCassandra();
+    EmbeddedCassandraServerHelper.startEmbeddedCassandra(STARTUP_TIMEOUT);
     cluster = EmbeddedCassandraServerHelper.getCluster();
     session = EmbeddedCassandraServerHelper.getSession();
 


sdks/java/io/cassandra/src/test/java/org/apache/beam/sdk/io/cassandra/CassandraIOTest.java
```diff
@@ -101,7 +101,7 @@
   public static void startCassandra() throws Exception {
     System.setProperty("cassandra.jmx.local.port", JMX_PORT);
     startupTime = System.currentTimeMillis();
-    EmbeddedCassandraServerHelper.startEmbeddedCassandra();
+    EmbeddedCassandraServerHelper.startEmbeddedCassandra(STARTUP_TIMEOUT);
```
mxm: The default used before was `20_000`. Now this is `45_000`. Does this really prevent the NullPointerException?

mxm: Should we try to write the test in a defensive way that it fails with a timeout exception instead of a NullPointerException?

echauchot: @mxm The problem is that the failure is in the AfterClass which is launched even if BeforeClass fails: what happens is the cluster startup in BeforeClass fails with a timeout exception consequently Cluster is null and the Cluster.getMetadata() fails in the AfterClass cleaning method. What I can do is avoid calling the cleaning method if Cluster is null.

echauchot: @mxm just pushed defensive code for it that way: 

- if startup time is above 45s, startCassandra BeForeClass fails with an exception and no NPE in AfterClass because it does nothing in that case.
- If startup time is between 20s and 45s then, all the tests are executed and pass.

sdks/java/io/cassandra/src/test/java/org/apache/beam/sdk/io/cassandra/CassandraIOTest.java
```diff
@@ -101,7 +101,7 @@
   public static void startCassandra() throws Exception {
     System.setProperty("cassandra.jmx.local.port", JMX_PORT);
     startupTime = System.currentTimeMillis();
-    EmbeddedCassandraServerHelper.startEmbeddedCassandra();
+    EmbeddedCassandraServerHelper.startEmbeddedCassandra(500L);
```
mxm: Now the timeout is very low? 500 milliseconds?

sdks/java/io/cassandra/src/test/java/org/apache/beam/sdk/io/cassandra/CassandraIOTest.java
```diff
@@ -126,9 +126,11 @@ public static void startCassandra() throws Exception {
 
   @AfterClass
   public static void stopCassandra() {
-    EmbeddedCassandraServerHelper.cleanEmbeddedCassandra();
-    session.close();
-    cluster.close();
+    if (cluster != null && session != null) {
+      EmbeddedCassandraServerHelper.cleanEmbeddedCassandra();
+      session.close();
+      cluster.close();
```
mxm: Would break this into two if blocks, depending on where we fail, one of them might be non null.

echauchot: well, `session = cluster.connect().` So if cluster is null (because there was an exception in startup), then session is null.  And if not, cluster.connect() is a blocking call to connectAsync future.  That is why I wrote that way. Do you want me to change it ?

mxm: I was worried that if there is an exception creating the session, then the cluster won't be shutdown.

mxm: Ok for me to keep it that way.

sdks/java/io/cassandra/src/test/java/org/apache/beam/sdk/io/cassandra/CassandraIOTest.java
echauchot: oups, I did a test to force a timeout to check the behavior and I forgot to remove it :)

sdks/java/io/cassandra/src/test/java/org/apache/beam/sdk/io/cassandra/CassandraIOTest.java
echauchot: Good point, I guess it can happen that there is an exception in session creation but have the cluster well created, I'll add the code.

sdks/java/io/cassandra/src/test/java/org/apache/beam/sdk/io/cassandra/CassandraIOTest.java
```diff
@@ -126,9 +126,15 @@ public static void startCassandra() throws Exception {
 
   @AfterClass
   public static void stopCassandra() {
-    EmbeddedCassandraServerHelper.cleanEmbeddedCassandra();
-    session.close();
-    cluster.close();
+    if (cluster != null && session != null) {
+      EmbeddedCassandraServerHelper.cleanEmbeddedCassandra();
+      session.close();
```
iemejia: shouldn't this be done before cleaning the embedded server?

sdks/java/io/cassandra/src/test/java/org/apache/beam/sdk/io/cassandra/CassandraIOTest.java
```diff
@@ -126,9 +126,15 @@ public static void startCassandra() throws Exception {
...
+      cluster.close();
+    } else {
+      if (cluster != null){
```
iemejia: this validation is unneeded because is covered by the previous if. Probably is simpler to let it only with if (cluster != null) { ... } and do if (session != null) { ,,, } before.

sdks/java/io/cassandra/src/test/java/org/apache/beam/sdk/io/cassandra/CassandraIOTest.java
echauchot: no because cleanEmbeddedCassandra require both the cluster and the session to drop the keyspace

sdks/java/io/cassandra/src/test/java/org/apache/beam/sdk/io/cassandra/CassandraIOTest.java
echauchot: No, it is not the same validation. the  `if (cluster != null)` is under an `else`. So it is the code for the case `session == null  && cluster != null` i.e. the cluster is created but not the session. This is the case @mxn reported.
Your suggestion is not correct either, because both cluster and session need to be not null to clean cassandra, and we cannot close cluster and session beforewards because they are needed for the cleaning process.

echauchot:
I addressed @iemejia comments. As @iemejia is sick and already approved the PR, and also because flaky tests are urgent to fix, I'll merge the PR.

PR #8026, opened 2019-03-10
[BEAM-6726] explicitly specify signing key

adude3141:
This PR changes the release script to not rely on gpg default key, but forces the release manager to explicitly specify the signgin key which then is used for signing tasks.

------------------------

### DIFF
--- a/release/src/main/scripts/build_release_candidate.sh
+++ b/release/src/main/scripts/build_release_candidate.sh
@@ -56,12 +56,19 @@ read USER_GITHUB_ID
 
 USER_REMOTE_URL=git@github.com:${USER_GITHUB_ID}/beam-site
 
+echo "================Listing all GPG keys================="
+gpg --list-keys --keyid-format LONG --fingerprint --fingerprint
+echo "Please copy the public key which is associated with your Apache account:"
+
+read SIGNING_KEY
+
 echo "================Checking Environment Variables=============="
 echo "beam repo will be cloned into: ${LOCAL_CLONE_DIR}"
 echo "working on release version: ${RELEASE}"
 echo "working on release branch: ${RELEASE_BRANCH}"
 echo "will create release candidate: RC${RC_NUM}"
 echo "Your forked beam-site URL: ${USER_REMOTE_URL}"
+echo "Your signing key: ${SIGNING_KEY}"
 echo "Please review all environment variables and confirm: [y|N]"
 read confirmation
 if [[ $confirmation != "y" ]]; then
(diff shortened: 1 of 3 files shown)


adude3141:
R: @aaltay

release/src/main/scripts/build_release_candidate.sh
```diff
@@ -98,7 +105,8 @@ if [[ $confirmation = "y" ]]; then
   echo "2. new rc tag has created in github."
 
   echo "-------------Staging Java Artifacts into Maven---------------"
-  ./gradlew publish -PisRelease --no-daemon
+  gpg --local-user ${SIGNING_KEY} --output /dev/null --sign ~/.bashrc
```
aaltay: Is this for testing that the key is working?

Would not it break if ~/.bashrc does not exist?

release/src/main/scripts/build_release_candidate.sh
```diff
@@ -56,12 +56,19 @@ read USER_GITHUB_ID
 
 USER_REMOTE_URL=git@github.com:${USER_GITHUB_ID}/beam-site
 
+echo "================Listing all GPG keys================="
+gpg --list-keys --keyid-format LONG --fingerprint --fingerprint
+echo "Please copy the public key which is associated with your Apache account:"
+
+read SIGNING_KEY
```
aaltay: Should we check that this is matching: user.signingkey (from: https://github.com/apache/beam/blob/master/release/src/main/scripts/preparation_before_release.sh#L48) or usage from other scripts?

release/src/main/scripts/build_release_candidate.sh
adude3141: No. It is to ensure the key is unlocked so gpg-agent will just provide access to the key without requesting for user input within gradle call. As gradle is configured to shell out to gpg cli, streams get broken and no input is possible.

And yes, this will break, if .bashrc does not exist. But the same pattern was used before [1], so I just reused that.

Of course, we might reconsider that.

[1] https://github.com/apache/beam/blob/master/release/src/main/scripts/verify_release_build.sh#L140

release/src/main/scripts/build_release_candidate.sh
adude3141: Probably yes.

But we did not check before [1], so I did not bother to implement this. As it probably would require to keep some state across scripts. Currently this is left to manual release verification.

As I tend to assume that these script need some rework anyway, I restricted the scope of this PR to a minimal viable solution to get release enabled on gradle5.

[1] There is no check on signing key set in git config against key put into KEYS file nor against default key used for signing artefacts.

adude3141:
Run Java PreCommit

adude3141:
Run Python PreCommit

release/src/main/scripts/build_release_candidate.sh
aaltay: Fair enough.

release/src/main/scripts/build_release_candidate.sh
aaltay: Sounds good. Do you mind add a JIRA todo comment here to clean this up in all scripts?

adude3141:
Nice. Thx for merging.

PR #8031, opened 2019-03-11
[BEAM-2530] Automatically retrieving JDK 11 containers in Dataflow

pabloem:
r: @aaltay 
r: @mwalenia

### DIFF
--- a/runners/google-cloud-dataflow-java/src/main/java/org/apache/beam/runners/dataflow/DataflowRunner.java
+++ b/runners/google-cloud-dataflow-java/src/main/java/org/apache/beam/runners/dataflow/DataflowRunner.java
@@ -1797,14 +1797,19 @@ public class DataflowRunner extends PipelineRunner<DataflowPipelineJob> {
   @VisibleForTesting
   static String getContainerImageForJob(DataflowPipelineOptions options) {
     String workerHarnessContainerImage = options.getWorkerHarnessContainerImage();
+
+    String javaVersionId =
+        Float.parseFloat(System.getProperty("java.specification.version")) >= 9 ? "java11" : "java";
     if (!workerHarnessContainerImage.contains("IMAGE")) {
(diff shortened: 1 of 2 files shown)


mwalenia:
@pabloem LGTM, thanks!

pabloem:
@aaltay does this look fine to you?

aaltay:
Question, raised from somewhere else:

Could dataflow just use a java11 based container to run both java8 and java11 pipelines? Would not java11 JVM will be compatible with java8 built files?

pabloem:
TBH I dont know what the answer is to this. Maybe yes? I'd like to hear from someone with more java knowledge.

aaltay:
@kennknowles Do you know if Java 11 JVM can backward compatibly run packages built with Java 8? Does Dataflow need a period of time where it offers Java 8 and Java 11 containers?

kennknowles:
Yes, I believe JDK 11 can run JDK 8 compiled code. I'm not aware of any breaking change in this direction.

kennknowles:
I would trust @lukecwik or @adude3141 opinion more than my own.

adude3141:
Unfortunately I do not fully understand the changes here.

I assume we are talking about some container, a beam pipeline is run with? I.e. the container is running beam code and user code.

I further assume, the customer is allowed to provide her code compiled with java8. So will likely not test on java11 herself, but that containers java version would be kind of transparent to her.

Now given a business critical pipeline depending on some third-party library which can not easily be replaced nor upgraded. Which in turn depends for instance on some byte code enhancement lib or does such simple things as
`(URLClassLoader) getClass().getClassLoader()`
which, iirc does not work anymore on java11 (or some such I do not remember).

Wouldn't that mean, that this customer is suddenly locked out of the service? Always assuming that for whatever reason it is not feasible to upgrade those deps.

PR #8037, opened 2019-03-12
Replace assertEquals(boolean, xyz) with assertBoolean(xyz)

coheigea:
Replace assertEquals(boolean, xyz) with assertBoolean(xyz). This makes the test code a bit more concise.


sdks/java/io/amazon-web-services/src/test/java/org/apache/beam/sdk/io/aws/s3/S3WritableByteChannelTest.java
```diff
@@ -148,9 +148,9 @@ public void testAtMostOne() {
     assertTrue(atMostOne(false, false));
     assertFalse(atMostOne(true, true, true));
     assertFalse(atMostOne(true, true, false));
-    assertFalse(atMostOne(true, false, true));
```
ibzib: This test intentionally includes extra boolean values, so I don't think it should be changed.

PR #6884, opened 2018-10-30
[BEAM-4783] Fix invalid parameter to set the partitioner in Spark GbK

iemejia:
Code was refactored too, to make it less error-prone.

**Please** add a meaningful description for your change here

------------------------

Follow this checklist to help us incorporate your contribution quickly and easily:

 - [ ] Format the pull request title like `[BEAM-XXX] Fixes bug in ApproximateQuantiles`, where you replace `BEAM-XXX` with the appropriate JIRA issue, if applicable. This will automatically link the pull request to the issue.
 - [ ] If this contribution is large, please file an Apache [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

It will help us expedite review of your Pull Request if you tag someone (e.g. `@username`) to look at it.

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

Lang | SDK | Apex | Dataflow | Flink | Gearpump | Samza | Spark
--- | --- | --- | --- | --- | --- | --- | ---
Go | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_GradleBuild/lastCompletedBuild/) | --- | --- | --- | --- | --- | ---
Java | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_GradleBuild/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Apex_Gradle/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Dataflow_Gradle/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Flink_Gradle/lastCompletedBuild/) [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Flink/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Gearpump_Gradle/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Samza_Gradle/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Spark_Gradle/lastCompletedBuild/)
[...]

PR #6511, opened 2018-09-27
[BEAM-5519] Remove call to groupByKey in Spark Streaming

kyle-winkelman:
Prevent the duplication of encode/decode cycle in wasted call to groupByKey in spark streaming. It is a waste because after the groupByKey is a call to updateStateByKey which will ensure that the data is shuffled to the correct processing location.

------------------------

Follow this checklist to help us incorporate your contribution quickly and easily:

 - [ ] Format the pull request title like `[BEAM-XXX] Fixes bug in ApproximateQuantiles`, where you replace `BEAM-XXX` with the appropriate JIRA issue, if applicable. This will automatically link the pull request to the issue.
 - [ ] If this contribution is large, please file an Apache [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

It will help us expedite review of your Pull Request if you tag someone (e.g. `@username`) to look at it.

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

Lang | SDK | Apex | Dataflow | Flink | Gearpump | Samza | Spark
--- | --- | --- | --- | --- | --- | --- | ---
Go | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_GradleBuild/lastCompletedBuild/) | --- | --- | --- | --- | --- | ---
Java | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_GradleBuild/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Apex_Gradle/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Dataflow_Gradle/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Flink_Gradle/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Gearpump_Gradle/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Samza_Gradle/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Spark_Gradle/lastCompletedBuild/)
[...]

PR #8042, opened 2019-03-12
[BEAM-6812]: Convert keys to ByteArray in Combine.perKey to make sure hashCode is consistent

jhalaria:
…
The PR makes the CBK consistent with GBK where we convert the keys to a ByteArray before calling group/combine functions.
------------------------

### DIFF
--- a/runners/spark/src/main/java/org/apache/beam/runners/spark/translation/GroupCombineFunctions.java
+++ b/runners/spark/src/main/java/org/apache/beam/runners/spark/translation/GroupCombineFunctions.java
@@ -143,10 +143,10 @@ public class GroupCombineFunctions {
     // Once Spark provides a way to include keys in the arguments of combine/merge functions,
     // we won't need to duplicate the keys anymore.
     // Key has to bw windowed in order to group by window as well.
-    JavaPairRDD<K, WindowedValue<KV<K, InputT>>> inRddDuplicatedKeyPair =
-        rdd.mapToPair(TranslationUtils.toPairByKeyInWindowedValue());
+    JavaPairRDD<ByteArray, WindowedValue<KV<K, InputT>>> inRddDuplicatedKeyPair =
+        rdd.mapToPair(TranslationUtils.toPairByKeyInWindowedValue(keyCoder));
 
-    JavaPairRDD<K, SerializableAccumulator<KV<K, AccumT>>> accumulatedResult =
+    JavaPairRDD<ByteArray, SerializableAccumulator<KV<K, AccumT>>> accumulatedResult =
         inRddDuplicatedKeyPair.combineByKey(
             input ->
                 SerializableAccumulator.of(sparkCombineFn.createCombiner(input), iterAccumCoder),
@@ -160,7 +160,11 @@ public class GroupCombineFunctions {
                         acc1.getOrDecode(iterAccumCoder), acc2.getOrDecode(iterAccumCoder)),
                     iterAccumCoder));
 
-    return accumulatedResult.mapToPair(i -> new Tuple2<>(i._1, i._2.getOrDecode(iterAccumCoder)));
+    return accumulatedResult.mapToPair(
+        i ->
+            new Tuple2<>(
+                CoderHelpers.fromByteArray(i._1.getValue(), keyCoder),
+                i._2.getOrDecode(iterAccumCoder)));
   }
 
   /** An implementation of {@link Reshuffle} for the Spark runner. */
--- a/runners/spark/src/main/java/org/apache/beam/runners/spark/translation/TranslationUtils.java
+++ b/runners/spark/src/main/java/org/apache/beam/runners/spark/translation/TranslationUtils.java
@@ -27,6 +27,7 @@ import org.apache.beam.runners.core.StateInternals;
 import org.apache.beam.runners.core.StateInternalsFactory;
 import org.apache.beam.runners.spark.SparkRunner;
 import org.apache.beam.runners.spark.coders.CoderHelpers;
+import org.apache.beam.runners.spark.util.ByteArray;
 import org.apache.beam.runners.spark.util.SideInputBroadcast;
 import org.apache.beam.sdk.coders.Coder;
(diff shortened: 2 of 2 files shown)


jhalaria:
@iemejia - Please review. I am working on adding tests. But the core partitioner is ready for a review.

runners/spark/src/main/java/org/apache/beam/runners/spark/translation/TransformTranslator.java
```diff
@@ -569,8 +568,8 @@ private static Partitioner getPartitioner(EvaluationContext context) {
     Long bundleSize =
         context.getSerializableOptions().get().as(SparkPipelineOptions.class).getBundleSize();
     return (bundleSize > 0)
-        ? null
-        : new HashPartitioner(context.getSparkContext().defaultParallelism());
+        ? CustomSparkHashPartitioner.of(context.getSparkContext().defaultParallelism())
```
jhalaria: This logic seems to be accidentally flipped?

runners/spark/src/main/java/org/apache/beam/runners/spark/translation/GroupCombineFunctions.java
```diff
@@ -146,20 +147,40 @@
...
-                SerializableAccumulator.of(
-                    sparkCombineFn.mergeValue(input, acc.getOrDecode(iterAccumCoder)),
-                    iterAccumCoder),
-            (acc1, acc2) ->
-                SerializableAccumulator.of(
-                    sparkCombineFn.mergeCombiners(
-                        acc1.getOrDecode(iterAccumCoder), acc2.getOrDecode(iterAccumCoder)),
-                    iterAccumCoder));
+    JavaPairRDD<K, SerializableAccumulator<KV<K, AccumT>>> accumulatedResult;
+    if (partitioner == null) {
```
jhalaria: I don't really like doing this null check and calling different functions `combineByKey` function based on whether the `partitioner` is present or not. I can change the `getPartitioner` function to always return a partitioner and then this way I don't have to do this check here. Thoughts @iemejia ?

jhalaria:
@iemejia : Please take as lot this when you get a chance. Thanks.

iemejia:
Sorry for the delay, have not forgotten about this one, just being swamped in other PRs. @dmvk could you help me review this one please.

dmvk:
Run Portable_Python PreCommit

dmvk:
Nice catch! This problem goes way deeper, we should not ever shuffle raw user data in the first place (we should always use beam coder to serialize data, before partitioning even happens). If you take a look at GBK implementation, it uses ByteArray instead of byte[] (which has correct hashCode implementation).

I guess tests did not catch this because they use primitive types (which spark can serialize on its own) as keys. Also tests are run in a single JVM, therefore hashCode is stable.

Correct fix would be:
- Changing CombinePerKey implementation to convert keys to ByteArray prior calling `combineByKey`
- In case we want to use custom partitioner, we should always check on the first getPartition call, whether hashCode == systemIdentityHashCode and throw an exception if it does; or make sure that we shuflle byte[] only (which may be less expensive than ByteArray object)

runners/spark/src/main/java/org/apache/beam/runners/spark/translation/TransformTranslator.java
dmvk: agreed, this logic is odd, I don't think `EvalutionContext` is enough to properly resolve paralellism for the shuffle operation

In case `bundleSize > 0`, we should reuse paralellism of the input RDD. This way we would get rid of `nullable` case and we can simplify GBK and CPK translations.

Spark's default partitioner does the same thing: org.apache.spark.Partitioner#defaultPartitioner

runners/spark/src/main/java/org/apache/beam/runners/spark/translation/GroupCombineFunctions.java
dmvk: We should get rid of `nullable` case and have one translation only, see note above

jhalaria:
@dmvk - Thank you for looking at the PR. I removed the custom partitioner logic for now. I made the CPK similar to GBK where we convert the key to a `ByteArray`. Will create another PR that provides the ability to add a custom partitioner.

runners/spark/src/main/java/org/apache/beam/runners/spark/translation/TransformTranslator.java
```diff
@@ -569,8 +569,8 @@ private static Partitioner getPartitioner(EvaluationContext context) {
     Long bundleSize =
         context.getSerializableOptions().get().as(SparkPipelineOptions.class).getBundleSize();
     return (bundleSize > 0)
-        ? null
-        : new HashPartitioner(context.getSparkContext().defaultParallelism());
+        ? new HashPartitioner(context.getSparkContext().defaultParallelism())
+        : null;
```
dmvk: Second thoughts, it was correct.

https://github.com/apache/beam/pull/6884/files#r246077919

This was in order to maintain old functionality (bundleSize == 0, which basically means to use predefined parallelism).

I think the old functionality doesn't make much sense as it doesn't scale with input data. I guess someone may use this in order to re-scale "downstream" stage, but there should be a better mechanism to do this.

Any thoughts? @timrobertson100 @kyle-winkelman

jhalaria: I see. I reverted back the changes made to `getPartition`

dmvk:
Run Spark ValidatesRunner

runners/spark/src/main/java/org/apache/beam/runners/spark/translation/TransformTranslator.java
kyle-winkelman: I agree that the old functionality seems strange, but I remember (when I had the logic backwards) that the performance tests for the spark runner were impacted. I think the impact was in streaming mode because if you don't use this HashPartitioner then it actually does a double shuffle of the data. I tried to clean this up but my I never finished getting PR #6511 merged.

PR #8057, opened 2019-03-14
HourlyTeamScore writes to a file

slykar:
@melap  The docs say the data is written to a BigQuery database. It's not what's in the code. The data is written to a text file.


melap:
Thanks for the PR! I looked at this, and it looks like the Python code example writes to BigQuery, and the Java one writes to text files.  For now, we might have to tag the last part of that sentence as language dependent so it displays the appropriate one depending on language toggle. Our current solution at the word level isn't great unfortunately -- you'd need to use something like this:

<span class="language-java">text file</span><span class="language-py">Google BigQuery</span>

@davidcavazos is there a plan to update the code examples to be consistent?

davidcavazos:
Sure, I can update it

davidcavazos:
I tried changing it to use TextIO and it failed with errors about the windowing setup not being supported for TextIO. I then remembered that at the time of writing, we decided to use BigQuery for that reason, so it can't be updated until the product supports that feature.

Melissa's solution on dynamically changing the documentation text depending on the language seems to be the way to go for now. Please go ahead with that change.

slykar:
@melap I've added another commit with the change you suggested after I saw @davidcavazos comment.

PR #8068, opened 2019-03-15
[BEAM-5115] Make ValueProvider API consistent between XmlIO and XmlSource

iemejia:
R: @aromanenko-dev


aromanenko-dev:
Thanks, LGTM in general. 
Could you add a unit test that uses this new added public method `from(ValueProvider<String>)`?

iemejia:
Done, Added missing call to the new signature. Should be ok now @aromanenko-dev

aromanenko-dev:
Thanks! Just one more note - please, apply `gradlew spotlessApply`, otherwise `Spotless PreCommit` is failing

iemejia:
Done, now waiting until Mr Jenkins get happy.

PR #8076, opened 2019-03-18
[BEAM-6619] [BEAM-6593] Add example integration tests to postcommit

Juta:
This is is part of a series of PRs with goal to make Apache Beam PY3 compatible. The proposal with the outlined approach has been documented here: https://s.apache.org/beam-python-3.

This PR adds a more integration tests to the postcommit jobs on direct and dataflow runners.

### DIFF
--- a/sdks/python/apache_beam/examples/complete/game/game_stats.py
+++ b/sdks/python/apache_beam/examples/complete/game/game_stats.py
@@ -306,6 +306,7 @@ def run(argv=None):
           topic=args.topic)
     raw_events = (
         scores
+        | 'DecodeString' >> beam.Map(lambda b: b.decode('utf-8'))
         | 'ParseGameEventFn' >> beam.ParDo(ParseGameEventFn())
         | 'AddEventTimestamps' >> beam.Map(
             lambda elem: beam.window.TimestampedValue(elem, elem['timestamp'])))
--- a/sdks/python/apache_beam/examples/complete/game/game_stats_it_test.py
+++ b/sdks/python/apache_beam/examples/complete/game/game_stats_it_test.py
@@ -33,8 +33,6 @@ Usage:
 from __future__ import absolute_import
 
 import logging
-import os
-import sys
 import time
 import unittest
 import uuid
@@ -51,10 +49,6 @@ from apache_beam.testing.pipeline_verifiers import PipelineStateMatcher
 from apache_beam.testing.test_pipeline import TestPipeline
 
 
-@unittest.skipIf(sys.version_info[0] == 3 and
-                 os.environ.get('RUN_SKIPPED_PY3_TESTS') != '1',
-                 'This test still needs to be fixed on Python 3'
-                 'TODO: BEAM-6711')
 class GameStatsIT(unittest.TestCase):
 
   # Input events containing user, team, score, processing time, window start.
@@ -102,7 +96,8 @@ class GameStatsIT(unittest.TestCase):
 
     for _ in range(message_count):
       self.pub_client.publish(topic.name,
-                              self.INPUT_EVENT % self._test_timestamp)
+                              (self.INPUT_EVENT % self._test_timestamp
+                              ).encode('utf-8'))
 
   def _cleanup_pubsub(self):
     test_utils.cleanup_subscriptions(self.sub_client, [self.input_sub])
--- a/sdks/python/apache_beam/examples/complete/game/leader_board.py
+++ b/sdks/python/apache_beam/examples/complete/game/leader_board.py
@@ -315,6 +315,7 @@ def run(argv=None):
 
     events = (
         scores
+        | 'DecodeString' >> beam.Map(lambda b: b.decode('utf-8'))
         | 'ParseGameEventFn' >> beam.ParDo(ParseGameEventFn())
         | 'AddEventTimestamps' >> beam.Map(
             lambda elem: beam.window.TimestampedValue(elem, elem['timestamp'])))
--- a/sdks/python/apache_beam/examples/complete/game/leader_board_it_test.py
+++ b/sdks/python/apache_beam/examples/complete/game/leader_board_it_test.py
@@ -33,8 +33,6 @@ Usage:
 from __future__ import absolute_import
 
 import logging
-import os
-import sys
 import time
 import unittest
 import uuid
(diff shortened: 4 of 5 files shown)


Juta:
Run Python PostCommit

Juta:
Run Python PostCommit

Juta:
Run Python PostCommit

Juta:
Run Python PostCommit

Juta:
Run Python PreCommit

Juta:
Run Python PostCommit

Juta:
Run Python PostCommit

Juta:
Run Python PostCommit

Juta:
Run Python PostCommit

Juta:
Run Python PostCommit

Juta:
cc: @pabloem @ttanay @tvalentyn

sdks/python/apache_beam/examples/complete/game/game_stats.py
```diff
@@ -111,6 +111,8 @@ def __init__(self):
 
   def process(self, elem):
     try:
+      if isinstance(elem, bytes):
```
tvalentyn: Can we always decode here? It is better to have a clear expectation of input arguments as much as possible on Python 3: either always encoded bytes, or always strings, but not mixing the two.

sdks/python/apache_beam/examples/complete/game/leader_board.py
```diff
@@ -120,6 +120,8 @@ def __init__(self):
 
   def process(self, elem):
     try:
+      if isinstance(elem, bytes):
```
tvalentyn: Can we always decode here? It is better to have a clear expectation of input arguments as much as possible on Python 3: either always encoded bytes, or always strings, but not mixing the two.

sdks/python/apache_beam/examples/complete/game/leader_board_it_test.py
```diff
@@ -53,7 +53,7 @@
 class LeaderBoardIT(unittest.TestCase):
 
   # Input event containing user, team, score, processing time, window start.
-  INPUT_EVENT = 'user1,teamA,10,%d,2015-11-02 09:09:28.224'
+  INPUT_EVENT = b'user1,teamA,10,%d,2015-11-02 09:09:28.224'
```
tvalentyn: I am not sure if this should be bytes. Does `self.pub_client.publish` require bytes? If so, we should encode before passing data to that method.

sdks/python/apache_beam/examples/complete/game/game_stats_it_test.py
```diff
@@ -52,7 +52,7 @@
 class GameStatsIT(unittest.TestCase):
 
   # Input events containing user, team, score, processing time, window start.
-  INPUT_EVENT = 'user1,teamA,10,%d,2015-11-02 09:09:28.224'
+  INPUT_EVENT = b'user1,teamA,10,%d,2015-11-02 09:09:28.224'
```
tvalentyn: I am not sure if this should be bytes. Does self.pub_client.publish require bytes? If so, we should encode before passing data to that method, 

To quote https://docs.python.org/3/howto/pyporting.html ,  `Decode binary data to text as soon as possible, encode text as binary data as late as possible`

sdks/python/apache_beam/examples/complete/game/leader_board.py
Juta: The IT test uses pubsub as input source while the unit test currently passes strings. That's why we cannot always decode

tvalentyn: I am not sure I follow - could you please elaborate - what exactly in either pubsub or unit test makes it so that the input type is not consistent here?

sdks/python/apache_beam/examples/complete/game/leader_board_it_test.py
Juta: yes `self.pub_client.publish` requires bytes. In this case I think specifying the input event as bytes is what is expected because it is directly passed to the client. What do you think?

tvalentyn: So the question is then, to the humans who read and edit this code, should INPUT_EVENT be a textual date or encoded data? I think it's easier to consider it text, up until it's time to feed it to pubsub, then when we can encode it to bytes.

I'd keep INPUT_EVENT as is change publishing to: 

```
event = self.INPUT_EVENT % self._test_timestamp
self.pub_client.publish(event.encode('utf-8'))
```
I would expect users to follow a similar pattern in their pipelines, and they might refer to beam examples for guidance, so I suggest to change this.

sdks/python/apache_beam/examples/complete/game/leader_board.py
Juta: The code for leader_board and game_stats is used both with strings https://github.com/apache/beam/blob/master/sdks/python/apache_beam/examples/complete/game/leader_board_test.py#L45 (unicode strings) and with pubsub messages: https://github.com/apache/beam/blob/master/sdks/python/apache_beam/examples/complete/game/leader_board_it_test.py#L146.

Juta:
Run Python PostCommit

sdks/python/apache_beam/examples/complete/game/leader_board.py
tvalentyn: Ok, the fix would then be to:
1. Expect `elem` to be a string.
2. Decode the messages as they arrive from PubSub using `| 'DecodeString' >> Map(lambda b: b.decode('utf-8')` in the example pipeline. Note that there are several example pipelines that will need this change.

Note that we used to have `ReadStringsFromPubSub` method that was deprecated: https://github.com/apache/beam/blob/a5ed104aeac91781492523808fa4d3df6f2e55f7/sdks/python/apache_beam/io/gcp/pubsub.py#L194.

We may want to revisit this deprecation in the future.

Juta:
Run Python PostCommit

Juta:
Run Python PostCommit

Juta:
Run Python PostCommit

Juta:
Run Python PostCommit

sdks/python/apache_beam/examples/complete/game/leader_board.py
Juta: @tvalentyn I applied this fix

aaltay:
Thank you @Juta and @tvalentyn

PR #8083, opened 2019-03-18
[BEAM-6854] Update amazon-web-services aws-sdk dependency to version 1.11.519

iemejia:
It also unifies AWS SDK versions for the kinesis and amazon-web-services modules

R: @aromanenko-dev

### DIFF
--- a/buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy
+++ b/buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy
@@ -340,30 +340,31 @@ class BeamModulePlugin implements Plugin<Project> {
     // These versions are defined here because they represent
     // a dependency version which should match across multiple
     // Maven artifacts.
-    def generated_grpc_beta_version = "0.44.0"
-    def generated_grpc_ga_version = "1.43.0"
-    def google_cloud_bigdataoss_version = "1.9.16"
-    def google_cloud_spanner_version = "1.6.0"
-    def google_clients_version = "1.27.0"
-    def google_auth_version = "0.12.0"
-    def grpc_version = "1.17.1"
-    def protobuf_version = "3.6.0"
-    def guava_version = "20.0"
-    def netty_version = "4.1.30.Final"
-    def proto_google_common_protos_version = "1.12.0"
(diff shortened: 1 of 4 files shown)


buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy
```diff
@@ -340,30 +340,31 @@ class BeamModulePlugin implements Plugin<Project> {
     // These versions are defined here because they represent
     // a dependency version which should match across multiple
     // Maven artifacts.
+    def apex_core_version = "3.7.0"
```
aromanenko-dev: I see a lot of changes of version variables below. Is it caused by just reordering them? Did you keep the same version values (except `aws_java_sdk_version`)?

sdks/java/io/amazon-web-services/src/test/java/org/apache/beam/sdk/io/aws/sqs/SqsIOTest.java
```diff
@@ -117,7 +117,6 @@ protected void before() {
                   new AWSStaticCredentialsProvider(new BasicAWSCredentials(accessKey, secretKey)))
               .withEndpointConfiguration(
                   new AwsClientBuilder.EndpointConfiguration(endpoint, region))
-              .withRegion(region)
```
aromanenko-dev: Why `withRegion` was dropped? Because region is used in `EndpointConfiguration` creation?

buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy
iemejia: yes exactly it is just a reorder to make things easier to find.

sdks/java/io/amazon-web-services/src/test/java/org/apache/beam/sdk/io/aws/sqs/SqsIOTest.java
iemejia: Yes and now the amazon sdk validates this so the test failed.

aromanenko-dev:
LGTM

PR #8090, opened 2019-03-19
[BEAM-6861] Add support to specify a query in CassandraIO

stankiewicz:
CassandraIO.Read currently selects all fields and maps them to POJO. 
With this change it is possible to select only subset of fields or added computed fields to allow reading things like write Timestamp or other functions. 

------------------------

### DIFF
--- a/sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
+++ b/sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
@@ -37,7 +37,11 @@ import com.google.auto.value.AutoValue;
 import com.google.common.base.Joiner;
 import com.google.common.util.concurrent.ListenableFuture;
 import java.math.BigInteger;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+import java.util.NoSuchElementException;
 import java.util.concurrent.ExecutionException;
 import java.util.function.BiFunction;
 import java.util.stream.Collectors;
@@ -219,12 +223,12 @@ public class CassandraIO {
       return builder().setTable(table).build();
     }
 
-    /** Specify select fields to reduce data or add computed fields * */
+    /** Specify select fields to reduce data or add computed fields. */
     public Read<T> withSelectFields(ValueProvider<List<String>> selectFields) {
       return builder().setSelectFields(selectFields).build();
     }
 
-    /** Specify select fields to reduce data or add computed fields * */
+    /** Specify select fields to reduce data or add computed fields. */
     public Read<T> withSelectFields(List<String> selectFields) {
       checkArgument(
           selectFields != null
--- a/sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
+++ b/sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
@@ -135,7 +135,7 @@ public class CassandraIO {
     abstract ValueProvider<List<String>> hosts();
 
     @Nullable
-    abstract ValueProvider<List<String>> selectFields();
+    abstract ValueProvider<String> query();
 
     @Nullable
     abstract ValueProvider<Integer> port();
@@ -223,19 +223,23 @@ public class CassandraIO {
       return builder().setTable(table).build();
     }
 
-    /** Specify select fields to reduce data or add computed fields. */
-    public Read<T> withSelectFields(ValueProvider<List<String>> selectFields) {
-      return builder().setSelectFields(selectFields).build();
+    /**
+     * Customize select query to reduce data or add computed fields. Provide $CONDITIONS placeholder
+     * inside where statement.
+     */
+    public Read<T> withQuery(ValueProvider<String> query) {
+      return builder().setQuery(query).build();
     }
 
-    /** Specify select fields to reduce data or add computed fields. */
-    public Read<T> withSelectFields(List<String> selectFields) {
+    /**
+     * Customize select query to reduce data or add computed fields. Provide $CONDITIONS placeholder
+     * inside where statement.
+     */
+    public Read<T> withQuery(String query) {
       checkArgument(
-          selectFields != null
-              && selectFields.size() > 0
-              && selectFields.stream().allMatch(s -> s == null || !s.isEmpty()),
-          "Select fields should be valid field names");
-      return builder().setSelectFields(ValueProvider.StaticValueProvider.of(selectFields)).build();
+          query != null && query.length() > 0 && query.contains("$CONDITIONS"),
+          "Query should not be empty. Provide $CONDITIONS placeholder inside where statement.");
+      return builder().setQuery(ValueProvider.StaticValueProvider.of(query)).build();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


stankiewicz:
@iemejia - this is very simple PR, would you have time check it? Or can you recommend someone who would have time? 
It's already green.

sdks/java/io/cassandra/src/test/java/org/apache/beam/sdk/io/cassandra/CassandraIOTest.java
```diff
@@ -270,6 +273,68 @@ public void testReadWithWhere() throws Exception {
...
+                .withWhere(pipeline.newProvider("person_id=10")));
+
+    PAssert.thatSingleton(output.apply("Count", Count.globally())).isEqualTo(1L);
+    PAssert.that(output)
+        .satisfies(
+            input -> {
+              for (Scientist sci : input) {
+                assertNotNull(sci.id);
+                assertNull(sci.name);
+              }
```
iemejia: What happens if you read `nameTs` here? Since objects are mapped I was wondering if it will be zero, probably worth asserting that to validate that the values were not 'filled' (given that they were not projected).

sdks/java/io/cassandra/src/test/java/org/apache/beam/sdk/io/cassandra/CassandraIOTest.java
```diff
@@ -270,6 +273,68 @@ public void testReadWithWhere() throws Exception {
...
+              return null;
+            });
+
+    pipeline.run();
+  }
+
+  @Test
+  public void testSelectFieldsWithFunction() throws Exception {
```
iemejia: Please remove this method, it is almost identical to the previous one or is there something I am missing?

iemejia (changes requested):
Mmm I hadn't seen how this has evolved. It is not good to keep adding more options and methods to the main API for something that in the end is only building a query. Maybe worth to take a look if we can generalize this to allow to pass the full query akin to how `JdbcIO` does `withQuery(String query)` and refactor the existing code to use this, we can also provide some external builder class to build the query (we are doing this in a recent PR on MongoDbIO), but well this last part (the helper builder) is probably worth a different PR.

sdks/java/io/cassandra/src/test/java/org/apache/beam/sdk/io/cassandra/CassandraIOTest.java
stankiewicz: It is testing UDFs (writetime(person_name)), but I can merge this into one test

sdks/java/io/cassandra/src/test/java/org/apache/beam/sdk/io/cassandra/CassandraIOTest.java
stankiewicz: nameTs will be null as it is not returned by Row so it's not mapped. 
Agree on assert.

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
```diff
@@ -483,6 +508,11 @@ private CassandraIO() {}
               .collect(Collectors.joining(","));
 
       List<BoundedSource<T>> sources = new ArrayList<>();
+      String selectFields = "*";
+      if (spec.selectFields() != null) {
+        selectFields = String.join(",", spec.selectFields().get());
```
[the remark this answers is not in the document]
stankiewicz: Will do.

sdks/java/io/cassandra/src/test/java/org/apache/beam/sdk/io/cassandra/CassandraIOTest.java
iemejia: yes better do only one test

iemejia:
@stankiewicz WDYT about the `withQuery` idea, I think we need to better generalize this to have a nicer API and don't add new methods everytime a new requirement comes into place.

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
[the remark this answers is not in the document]
stankiewicz: @iemejia , thanks for feedback - challenge with withQuery in JdbcIO is that it is expected to be full query, like `select a from b` or `select a from b where c='d'`.
In CassandraIO select is generated so it can be split based on key ranges so 'where' part is partially dynamic so where part is set via separate field and I want select fields also set via seperate field/list of fields. 
That's why external builder can be challenge to support every case, unless it gives placeholder for each of those (select fields, where fields). 
So I generally agree we need some more generic but I don't think I could generalize mentioned MongoDB (BEAM-6241) and Cassandra query builder because they are totally different but definately I could try to create such builder for Cassandra that will gather projection/select fields and filter

iemejia:
Select is generated in Cassandra only to add the ranges no? I mean I don't see a huge difference with MongoDbIO who does something similar. And about the field selection well it could be more user friendly to have the method, but it also adds more maintenance work that will be completely fixed just by supporting the full query, We will even support unforeseen cases.

stankiewicz:
so what do you thing about withQuery like in [sqoop](https://sqoop.apache.org/docs/1.4.2/SqoopUserGuide.html#_free_form_query_imports) where user have to provide $CONDITIONS tag so library can insert range filters?

stankiewicz:
to summarise change - if user provides withQuery, I will support where field, otherwise I will produce query based on schema, table name and where field (if exists). 
WDYT @iemejia ?

iemejia:
if user provides `withQuery` we respect the user query, otherwise we build it (as it is the case right now). To test that the select fields works, we provide a `SELECT` with specific fields using the new `withQuery` and we don't add an specific method for the list of fields.

srfrnk:
Sorry for barging in on this.
If there is a need for a more robust query building mechanism we might consider using [Datastux QueryBuilder](https://docs.datastax.com/en/drivers/java/2.0/com/datastax/driver/core/querybuilder/QueryBuilder.html).
I tryed to go for that approach when I added `withWhere` but since their classes are not `Serializable` I could not.
I attempted to fix this problem [here](https://github.com/datastax/java-driver/pull/1163) but it was rejected.
Maybe with enough influence we can manage to make the change?
If not maybe this repo can be forked into Beam and so a working QueryBuilder can be used without having to rebuild it?

WDYT?

stankiewicz:
Hi @srfrnk! 
I don't like the idea of forking driver as we will end up maintaining it. 
I haven't thought about it but I like the idea of programmatic generation of query instead of doing String joins and formats. Challenge here is what we want to expose instead of withWhere() or withQuery() - withQueryBuilder()? If yes it would be nice alternative because we can then just add key ranges dynamically. But then the challenge is how to support value provider for it. 

If we figure out serialisation then we can clean this up a little bit. In the meantime I would try to merge this.

stankiewicz:
@iemejia , WDYT about current implementation?

iemejia:
@stankiewicz Sorry for slow review, I will check ASAP probably today or tuesday at latest. Thanks.

stankiewicz:
@iemejia , no problem :) Thanks for all the effort!

iemejia:
I am reviewing this. Mostly LGTM, I will probably do the missing minor fixes and merge, will confirm later on. Thanks.

stankiewicz:
@iemejia  - thanks a lot!

stankiewicz:
@iemejia , Ismael, any chance closing that before cut-off or are we already too late? Thanks, Radek

iemejia (approved):
LGTM. Merging manually to do some final touches.

iemejia:
Thanks a lot @stankiewicz !

PR #8097, opened 2019-03-20
Improve docs on SpannerWriteResult.

nielm:
Add a note to SpannerIO javadoc that the output PCollection in SpannerWriteResult will only
work in Batch pipelines...

@chamikaramj 

### DIFF
--- a/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/spanner/SpannerIO.java
+++ b/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/spanner/SpannerIO.java
@@ -169,7 +169,10 @@ import org.slf4j.LoggerFactory;
  *
  * The {@link SpannerWriteResult SpannerWriteResult} object contains the results of the transform,
  * including a {@link PCollection} of MutationGroups that failed to write, and a {@link PCollection}
- * that can be used as a completion signal.
+ * that can be used in batch pipelines as a completion signal to {@link
+ * org.apache.beam.sdk.transforms.Wait Wait.OnSignal} to indicate when all input has been written.
+ * Note that in streaming pipelines, this signal will never be triggered as the input is unbounded
+ * and this {@link PCollection} is using the {@link GlobalWindow}.
  *
  * <h3>Batching</h3>
  *
--- a/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/spanner/SpannerWriteResult.java
+++ b/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/spanner/SpannerWriteResult.java
@@ -28,9 +28,16 @@ import org.apache.beam.sdk.values.TupleTag;
 import org.apache.beam.vendor.guava.v20_0.com.google.common.collect.ImmutableMap;
 
 /**
- * A result of {@link SpannerIO#write()} transform. Use {@link #getFailedMutations()} to access
- * failed Mutations. {@link #getOutput()} can be used as a completion signal with the {@link
- * org.apache.beam.sdk.transforms.Wait Wait} transform.
+ * The results of a {@link SpannerIO#write()} transform.
+ *
+ * <p>Use {@link #getFailedMutations()} to access a {@link PCollection} of MutationGroups that
+ * failed to write.
+ *
+ * <p>The {@link PCollection} returned by {@link #getOutput()} can be used in batch pipelines as a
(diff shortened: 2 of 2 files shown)


sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/spanner/SpannerIO.java
```diff
@@ -169,7 +169,10 @@
  *
  * The {@link SpannerWriteResult SpannerWriteResult} object contains the results of the transform,
  * including a {@link PCollection} of MutationGroups that failed to write, and a {@link PCollection}
- * that can be used as a completion signal.
+ * that can be used in batch pipelines as a completion signal to {@link
```
chamikaramj: Do we wait till all output is written before MutationGroups that failed to write are emitted by the DoFn that performed writing ? Otherwise I'm not sure why this will not work in streaming mode.

chamikaramj: Of course this cannot be used as a simple Wait in streaming mode since we are waiting for failures (so if there are no failures we'll wait forever).

nielm: > Do we wait till all output is written before MutationGroups that failed to write are emitted by the DoFn that performed writing ?
No, they are emitted at any time. 

> Of course this cannot be used as a simple Wait in streaming mode since we are waiting for failures (so if there are no failures we'll wait forever).

￼There are 2 output PCollections (see the edit to SpannerWriteResult) 

- 'getOutput()' which will never have any elements, is in the GlobalWindow and will only be closed when the input PCollection to SpannerIO.Write[Grouped] is closed. 
- getFailedMutations() which is also in the GlobalWindow, will only contain mutations that failed to write, and will also only be closed when the input PCollection to SpannerIO.Write[Grouped] is closed. 

So neither stream can be used as a signal to Wait.on() in streaming mode, as they will never signal.

chamikaramj: getOutput() sounds like a batch only feature that was added to satisfy the wait requirement but I think this kind of batch-only features should be generally discouraged in Beam since Beam transforms (and features of those transforms) are expected to work for in both batch and streaming modes. I think a better way might have been to produce PCollection of some sort of a result that would be produced by writing a batch of records to Spanner. (for example, I think we produce a PCollection of filenames for textio). 

@kennknowles @reuvenlax WDYT ?

nielm: I agree, but this is a 'feature' that already exists, I am just trying to improve the docs in this PR to avoid that people use it in a way that does not work.

chamikaramj:
Thanks. I'll go ahead and merge. But we could deprecate/update the feature as needed. Might make sense to bring this up in Beam dev list.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/spanner/SpannerIO.java
kennknowles: Commenting without digging in to the implementation.

Having the output in the global window is appropriate and shouldn't be a problem. Record(s) should be produced when a write is done whether it is in streaming or batch. It seems like a bug to have an output collection for noting completed writes that never has any elements but _does_ have completed writes.

I am not expert enough on `Watch` to say whether to call it a bug or a spec issue or just a limitation of what `Watch` is good for. I am wary of linking the two transform specs versus making this general and also making `Watch` general.

If things are really not working, it might be best to phrase this as a warning about functionality issues, with a link to a Jira, instead of writing it up like it was meant to be this way.

chamikaramj: Niel, could you create a JIRA for updating the functionality to support both batch and streaming and refer to that in the documentation here ?

nielm: [BEAM-6921](https://issues.apache.org/jira/browse/BEAM-6921) created for improving output
[BEAM-6887](https://issues.apache.org/jira/browse/BEAM-6887) created for adding a more useful streaming pipeline.

PR #8102, opened 2019-03-20
[BEAM-6892] Adding support for side inputs on table & schema. Also adding additional params for BQ.

pabloem:
r: @chamikaramj 
This will allow support for time partitioning, and for side inputs coming into the schema and table transforms. This is useful because schemas or table names may be computed during pipeline execution, and side inputs allow users the ability to access them when they're computed.

### DIFF
--- a/sdks/python/apache_beam/io/gcp/bigquery.py
+++ b/sdks/python/apache_beam/io/gcp/bigquery.py
@@ -97,7 +97,7 @@ should be sent to.
 You may also provide a tuple of PCollectionView elements to be passed as side
 inputs to your callable. For example, suppose that one wishes to send
 events of different types to different tables, and the table names are
-computed at pipeline runtime, one may do something like so::
+computed at pipeline runtime, one may do something like the following::
 
     with Pipeline() as p:
       elements = (p | beam.Create([
@@ -115,6 +115,10 @@ computed at pipeline runtime, one may do something like so::
       elements | beam.io.gcp.WriteToBigQuery(
         table=lambda row, table_dict: table_dict[row['type']],
         table_side_inputs=(table_names_dict,))
+
+In the example above, the `table_dict` argument passed to the function in
+`table_dict` is the side input coming from `table_names_dict`, which is passed
+as part of the `table_side_inputs` argument.
 
 Schemas
 ---------
@@ -698,7 +702,9 @@ class BigQueryWriteFn(DoFn):
       retry_strategy: The strategy to use when retrying streaming inserts
         into BigQuery. Options are shown in bigquery_tools.RetryStrategy attrs.
       additional_bq_parameters (dict, callable): A set of additional parameters
-        to be passed when creating a BigQuery table.
+        to be passed when creating a BigQuery table. These are passed when
+        triggering a load job for FILE_LOADS, and when creating a new table for
+        STREAMING_INSERTS.
     """
     self.schema = schema
     self.test_client = test_client
--- a/sdks/python/apache_beam/io/gcp/bigquery.py
+++ b/sdks/python/apache_beam/io/gcp/bigquery.py
@@ -813,7 +813,8 @@ bigquery_v2_messages.TableSchema`. or a `ValueProvider` that has a JSON string,
         fields, repeated fields, or specifying a BigQuery mode for fields
         (mode will always be set to ``'NULLABLE'``).
         If a callable, then it should receive a destination (in the form of
-        a TableReference or a string, and return a str, dict or TableSchema.
+        a TableReference or a string, and return a str, dict or TableSchema, and
+        it should return a str, dict or TableSchema.
       create_disposition (BigQueryDisposition): A string describing what
         happens if the table does not exist. Possible values are:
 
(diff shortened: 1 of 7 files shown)


pabloem:
Run Python PostCommit

pabloem:
Run Python PreCommit

pabloem:
Run Python PostCommit

pabloem:
Run Python PreCommit

pabloem:
This is available in Java via DynamicDestination class: https://beam.apache.org/releases/javadoc/2.11.0/org/apache/beam/sdk/io/gcp/bigquery/DynamicDestinations.html

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -591,6 +592,8 @@ def __init__(
         batch has not been completely filled up.
       retry_strategy: The strategy to use when retrying streaming inserts
         into BigQuery. Options are shown in bigquery_tools.RetryStrategy attrs.
+      additional_bq_parameters (dict, callable): A set of additional parameters
```
chamikaramj: Are the names of these parameters unknown, variable ? If not why not add separate keyword arguments for these ?

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -855,6 +869,15 @@ def __init__(self,
         FILE_LOADS on Batch pipelines.
       insert_retry_strategy: The strategy to use when retrying streaming inserts
         into BigQuery. Options are shown in bigquery_tools.RetryStrategy attrs.
+      additional_bq_parameters (callable): A function that returns a dictionary
+        with additional parameters to pass to BQ when creating / loading data
+        into a table. These can be 'timePartitioning', 'clustering', etc. They
+        are passed directly to the job load configuration. See
+        https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.load
+      table_side_inputs (tuple): A tuple with ``AsSideInput`` PCollections to be
```
chamikaramj: What is "table callable" and "schema callable" ?

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -591,6 +592,8 @@ def __init__(
         batch has not been completely filled up.
       retry_strategy: The strategy to use when retrying streaming inserts
         into BigQuery. Options are shown in bigquery_tools.RetryStrategy attrs.
+      additional_bq_parameters (dict, callable): A set of additional parameters
+        to be passed when creating a BigQuery table.
```
chamikaramj: Also, prob. explain what users are expected to pass here and how they will be used. It's not clear to an end user how this parameter should be constructed and how this will be used by reading the current doc string.

pabloem:
Run Python PreCommit

pabloem:
@chamikaramj I've improved the top-level documentation for the class. Hopefully that'll clarify some of your questions.

I did not add individual parameters, because I did not want to add 20 parameters to the transform (and new ones every time BQ adds something else).

chamikaramj:
Lemme know when this is ready for another look. I think this is mostly done other than few comments I mentioned in the last round.

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -78,15 +78,120 @@
...
+also relies on creating temporary tables when performing file loads.
+
+The WriteToBigQuery transform creates tables using the BigQuery API by
+inserting a load job (see the API reference [1]), or by inserting a new table
+(see the API reference for that [2][3]).
+
+When creating a new BigQuery table, there are a number of extra parameters
+that one may need to specify. For example, clustering, partitioning, data
+encoding, etc. It is possible to provide these additional parameters by
+passing a Python dictionary as `additional_bq_parameters` to the transform.
```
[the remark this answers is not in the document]
pabloem: Pasting my chat response here:

hey cham so for BQ, the reason that  I'm admitting a variable set of properties is that in Java, the BQ sink API did not adequately support users to provide parameters for clustering
which is a newer feature
so when BQ added it, and users wanted to rely on it, they couldnt. And in fact, their pipelines would fail whenever we needed to run multiple loads for the same table with temporary tables
and they had clustered tables
so I decided to accept a dictionary that would make it easy to accept new elements 
but I agree with you it's risky. We could have a whitelist of properties that we admit. WDYT?

pabloem:
Run Python PostCommit

pabloem:
I've addressed all comments, except for the one relating `additional_bq_parameters`. The question remains open: WE could whitelist a set of admissible parameters (in this case, `['clustering', 'partitioning']`). Thoughts?

pabloem:
Run Python PreCommit

pabloem:
Passing postcommits:
https://builds.apache.org/job/beam_PostCommit_Python_Verify_PR/636/
https://builds.apache.org/job/beam_PostCommit_Python3_Verify_PR/273/

sdks/python/apache_beam/io/gcp/bigquery.py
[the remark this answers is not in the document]
chamikaramj: LGTM. But please make sure that Java and Python APIs are consistent in this regard.

chamikaramj:
Please squash commits and self merge.

PR #8103, opened 2019-03-20
Reset Datasource counter to 0 in FinishBundle

boyuanzz:
R: @lostluck 
CC: @ananvay

### DIFF
--- a/sdks/go/pkg/beam/core/runtime/exec/datasource.go
+++ b/sdks/go/pkg/beam/core/runtime/exec/datasource.go
@@ -174,8 +174,9 @@ func (n *DataSource) Process(ctx context.Context) error {
 func (n *DataSource) FinishBundle(ctx context.Context) error {
 	log.Infof(ctx, "DataSource: %d elements in %d ns", atomic.LoadInt64(&n.count), time.Now().Sub(n.start))
 	n.source = nil
+	err := n.Out.FinishBundle(ctx)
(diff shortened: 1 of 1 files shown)


lostluck:
Thanks! LGTM.
As mentioned offline, I'm slightly concerned about the progress for the InstructionResponse_ProcessBundle being wrong (this change renders it always zero), but when that message is receive, the runner shouldn't even be looking at the progress. The message is the signal it's done.

lostluck:
@aaltay Please merge.

sdks/go/pkg/beam/core/runtime/exec/datasource.go
```diff
@@ -174,6 +174,7 @@ func (n *DataSource) Process(ctx context.Context) error {
 func (n *DataSource) FinishBundle(ctx context.Context) error {
 	log.Infof(ctx, "DataSource: %d elements in %d ns", atomic.LoadInt64(&n.count), time.Now().Sub(n.start))
 	n.source = nil
+	atomic.StoreInt64(&n.count, 0)
```
lostluck: I mispoke,please move this to after the finishBundle call, but before returning the error.
The following should be correct, and reduce the number of 

err :=  n.Out.FinishBundle(ctx)
atomic.StoreInt64(&n.count, 0)
return err

lostluck:
LGTM Thanks!
@aaltay Please merge!

PR #8114, opened 2019-03-22
[Go SDK] Use the known varint64 coder for int64s.

lostluck:
It came to my attention that the Go SDK had no way of explicitly triggering known beam coders (in particular, varint64). I checked with Java and Python, and it's apparent that by default, int64s (longs) use the known varint encoding for this datatype, so this change brings this into parity.

Further, the Go SDK was limiting it's use of varints to 32 bits only, which isn't done in the other SDKs. 

This PR:
  * Makes the varint coder explicitly 64 bits, instead of 32 bits.
  * Makes int64s use the known varint coder for beam, allowing runners better knowledge of the data, where applicable.
      * Other numeric types will continue to use CustomCoders (which are then length prefixed for compatibility), unless I hear direction otherwise.
   * Updates the StandardCoder proto documentation for the varint urn, to make it's assumed 64bit-ness explicit.

I'd have liked to make all integers use varint, but the reality is that no subcomponents are permitted for beam's varint, so there's no where to put specific type information the SDK can gleam to produce the correct type right out of the gate. This is particular to Go where int's must be explicitly converted, rather than implicitly. Given the conversion step would be more expensive than the present LP+CustomCoder usage, this asymmetry is the best we can do for now.

------------------------

### DIFF
--- a/sdks/go/pkg/beam/core/graph/coder/coder.go
+++ b/sdks/go/pkg/beam/core/graph/coder/coder.go
@@ -246,7 +246,7 @@ func NewBytes() *Coder {
 
 // NewVarInt returns a new int32 coder using the built-in scheme.
 func NewVarInt() *Coder {
-	return &Coder{Kind: VarInt, T: typex.New(reflectx.Int32)}
+	return &Coder{Kind: VarInt, T: typex.New(reflectx.Int64)}
 }
 
 // IsW returns true iff the coder is for a WindowedValue.
--- a/sdks/go/pkg/beam/core/graph/coder/varint_test.go
+++ b/sdks/go/pkg/beam/core/graph/coder/varint_test.go
@@ -17,6 +17,8 @@ package coder
 
 import (
 	"bytes"
+	"fmt"
+	"math"
 	"testing"
 )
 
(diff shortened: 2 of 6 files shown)


lostluck:
R: @aaltay @youngoli 
Thanks!

lostluck:
Run Go PostCommit

sdks/go/pkg/beam/core/graph/coder/coder.go
```diff
@@ -246,7 +246,7 @@ func NewBytes() *Coder {
 
 // NewVarInt returns a new int32 coder using the built-in scheme.
```
youngoli: Looks like you missed changing this to say "int64 coder"

sdks/go/pkg/beam/core/graph/coder/varint_test.go
```diff
@@ -71,24 +77,26 @@ func TestEncodeDecodeVarInt(t *testing.T) {
 	}
 
 	for _, test := range tests {
-		var buf bytes.Buffer
```
youngoli: Idle thought, but maybe this test could be parallelized? (I just read through the testing docs and noticed testing.T.Parallel)

sdks/go/pkg/beam/core/graph/coder/coder.go
lostluck: Good catch! Updated in my client. I have a small set of other cleanup changes that it can go in with.

sdks/go/pkg/beam/core/graph/coder/varint_test.go
lostluck: It can be, but it's not necessary, the tests run in less than a second anyway.

PR #8121, opened 2019-03-22
[BEAM-6554] Replace latest mentions of FindBugs in build files (part 2)

iemejia:
Sorry to bother you again, I think i missed other ones. It should be ok now.
R: @kennknowles


.test-infra/jenkins/job_PreCommit_Java.groovy
```diff
@@ -41,7 +41,7 @@ builder.build {
         checkStyle {
           pattern('**/build/reports/checkstyle/*.xml')
         }
-        findBugs {
+        spotBugs {
```
dmvk: https://github.com/spotbugs/spotbugs-gradle-plugin/blob/master/src/main/java/com/github/spotbugs/SpotBugsPlugin.java#L122

Should be `spotbugs`.

It seems that extension names are case insensitive, but it may be still worth changing.

iemejia: Thanks for catching this one!

PR #8134, opened 2019-03-25
[BEAM-6906] Update spec on CombineFn and DoFn to clarify mutability of parameters

robinyqiu:
R: @kennknowles @aaltay

### DIFF
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/Combine.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/Combine.java
@@ -341,16 +341,18 @@ public class Combine {
     /**
      * Adds the given input value to the given accumulator, returning the new accumulator value.
      *
-     * <p>For efficiency, the input accumulator may be modified and returned.
+     * @param mutableAccumulator may be modified and returned for efficiency
+     * @param input should not be mutated
      */
-    public abstract AccumT addInput(AccumT accumulator, InputT input);
+    public abstract AccumT addInput(AccumT mutableAccumulator, InputT input);
 
     /**
      * Returns an accumulator representing the accumulation of all the input values accumulated in
      * the merging accumulators.
      *
-     * <p>May modify any of the argument accumulators. May return a fresh accumulator, or may return
-     * one of the (modified) argument accumulators.
+     * @param accumulators only the first accumulator may be modified and returned for efficiency;
+     *     the other accumulators should not be mutated, because they may be shared with other code
+     *     and mutating them could lead to incorrect results or data corruption.
      */
     public abstract AccumT mergeAccumulators(Iterable<AccumT> accumulators);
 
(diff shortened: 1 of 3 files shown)


sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/Combine.java
```diff
@@ -341,16 +341,17 @@ private Combine() {
     /**
      * Adds the given input value to the given accumulator, returning the new accumulator value.
      *
-     * <p>For efficiency, the input accumulator may be modified and returned.
+     * @param accumulator may be modified and returned, for efficiency
```
kennknowles: It would make this even more clear to rename the parameter to `mutableAccumulator`. Just a thought.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/Combine.java
```diff
@@ -341,16 +341,17 @@ private Combine() {
...
     public abstract AccumT addInput(AccumT accumulator, InputT input);
 
     /**
      * Returns an accumulator representing the accumulation of all the input values accumulated in
      * the merging accumulators.
      *
-     * <p>May modify any of the argument accumulators. May return a fresh accumulator, or may return
-     * one of the (modified) argument accumulators.
+     * @param accumulators only the first accumulator may be modified and returned, for efficiency;
+     *     the other accumulators should not be mutated
```
kennknowles: Why should they not be mutated? I would mention that they may be shared with other code and that mutating them could lead to incorrect results / data corruption.

kennknowles:
I think this is https://issues.apache.org/jira/browse/BEAM-6906 (which I just filed)

robinyqiu:
Thanks! I have addressed your comments and tagged the JIRA issue. I would like to separate this PR with a following one that deprecates `mergeAccumulators(Iterable<AccumT>)` if we want to do that. And I still have some concerns with the deprecation, explained below:

1) Deprecating `abstract` method without sacrificing backwards-compatibility seems not easy:
Taking the Java SDK for example, even if we do the deprecation, users will still have to implement the old version of mergeAccumulators and has no option to implement only the new one. This is because after the change we will have:
`@Deprecated public abstract AccumT mergeAccumulators(Iterable<AccumT>)`
and
`public AccumT mergeAccumulators(AccumT, Iterable<AccumT>)` (by default calls the deprecated one)
We have to keep the first one abstract and the second one non-abstract in order to guarantee backwards-compatibility. In this case, we are forcing users to implement a deprecated abstract function, and the new function doesn't help to make the interface cleaner/safer at all.

2) Potential performance penalty:
When we call the deprecated function from the new one, we will have to copy accumulators from one iterable to another. But I think this is a minor concern, compared to the first one.

One idea: what if we keep the old interface, but make the requirement of `mergeAccumulators(Iterable<AccumT>)` even more stringent by saying "non of the accumulator is should be mutated"? As far as I see, this way we keep the interface clean and safe, and the performance penalty from forcing users to create a new empty accumulator inside `mergeAccumulators` is not very high (is this true?). The greatest drawback of this approach is that users who were implementing this function by mutating accumulators will see their programs break without any explicit notice.

Please let me know if you have any idea! I would like to have further discussion on this issue.

robinyqiu:
Ping @kennknowles

PR #8137, opened 2019-03-26
Use self.assertEqual instead of self.assertEquals

boyuanzz:
self.assertEquals is deprecated and use self.assertEqual instead.
R: @aaltay


boyuanzz:
Run Python PreCommit

boyuanzz:
ping.

boyuanzz:
@pabloem Could you please help merge this PR? Thanks so much for your help!

PR #8152, opened 2019-03-27
[DoNotMerge][BEAM-1893] Implementation of CouchbaseIO

EdgarLGB:
Hi all,

Here is the PR for an initial version of CouchbaseIO. Currently, the read part has been accomplished. In general, the idea is to read the data by using a set of sources which is supposed to read a portion of data according to the key range. Knowing that Couchbase does not provide the feature of splitting up a bucket. So firstly I fetch the total number of keys and then divide the key range by the number of threads. Each source will be attributed a part of key range so that a whole set of sources can do a parallel read. 
I'm looking forward to the review. Thanks! (R: @aromanenko-dev @iemejia )

Regards,
Guobao

--- | --- | --- | --- | ---


sdks/java/io/couchbase/src/main/java/org/apache/beam/sdk/io/couchbase/CouchbaseIO.java
```diff
@@ -0,0 +1,359 @@
...
+      this.cluster = cluster;
+      this.bucket = bucket;
+      this.lowerBound = lb;
+      this.upperBound = ub;
+    }
+
+    private Bucket getBucket() {
+      if (cluster == null) {
+        DefaultCouchbaseEnvironment.Builder builder = DefaultCouchbaseEnvironment.builder();
+        if (spec.httpPort() != null) {
```
iemejia: Isn't `httpPort` mandatory, if so remove the `@Nullable` annotation and you won t need this.

sdks/java/io/couchbase/src/main/java/org/apache/beam/sdk/io/couchbase/CouchbaseIO.java
```diff
@@ -0,0 +1,359 @@
...
+          builder.bootstrapHttpDirectPort(spec.httpPort());
+        }
+        if (spec.carrierPort() != null) {
```
iemejia: Same that with `httpPort`

sdks/java/io/couchbase/src/main/java/org/apache/beam/sdk/io/couchbase/CouchbaseIO.java
```diff
@@ -0,0 +1,359 @@
...
+          builder.bootstrapCarrierDirectPort(spec.carrierPort());
+        }
+        cluster = CouchbaseCluster.create(builder.build(), spec.hosts());
+      }
+      if (bucket == null) {
+        // For Couchbase Server, in the previous version than 5.0, the passwordless bucket can be
```
iemejia: Please inline this comments, Beam's limit is 120 characters if I remember well so they should probably fit.

sdks/java/io/couchbase/src/main/java/org/apache/beam/sdk/io/couchbase/CouchbaseIO.java
```diff
@@ -0,0 +1,359 @@
...
+      source.close();
+      return result;
+    }
+  }
+
+  @VisibleForTesting
+  static class CouchbaseSource extends BoundedSource<JsonDocument> {
+
+    private final Read spec;
+    private int itemCount;
```
iemejia: `@Nullable private Long estimatedSizeBytes;`

sdks/java/io/couchbase/src/main/java/org/apache/beam/sdk/io/couchbase/CouchbaseIO.java
```diff
@@ -0,0 +1,359 @@
...
+     * do not estimate the total data size in bytes. In fact, we just fetch the total number of keys
+     * to the target bucket.
+     *
+     * @param options Pipeline options
+     * @return the number of keys to the bucket
+     * @throws Exception throw {@link CouchbaseIOException} when querying Couchbase
+     */
+    @Override
+    public long getEstimatedSizeBytes(PipelineOptions options) throws Exception {
+      String query = String.format("SELECT RAW COUNT(META().id) FROM `%s`", getBucket().name());
```
iemejia: `if (estimatedSizeBytes == null) ...` Estimating the size of the collection could be a heavy operation so this will save you extra calls.

sdks/java/io/couchbase/src/main/java/org/apache/beam/sdk/io/couchbase/CouchbaseIO.java
```diff
@@ -0,0 +1,359 @@
...
+    public Coder<JsonDocument> getOutputCoder() {
+      return SerializableCoder.of(JsonDocument.class);
+    }
+
+    @Override
+    public List<? extends BoundedSource<JsonDocument>> split(
+        long desiredBundleSize, PipelineOptions options) {
+      // If the desiredBundleSize equals to 0, it means that there will be only one bundle of data
+      // to be read.
+      int totalBundle =
```
iemejia: Since itemCount/estimatedSizeBytes could be null you may need to do `long estimatedSizeBytes = getEstimatedSizeBytes(options);` before.

sdks/java/io/couchbase/src/main/java/org/apache/beam/sdk/io/couchbase/CouchbaseIO.java
```diff
@@ -0,0 +1,359 @@
...
+    }
+
+    @Override
+    public JsonDocument getCurrent() throws NoSuchElementException {
+      return source.bucket.get(currentKey);
+    }
+
+    @Override
+    public void close() {
+      // Delegate the disconnection of Couchbase to the method "expand" of CouchbaseSource,
```
iemejia: Better idea to open and deal with the connection lifecycle at the reader level, because the Source does not have a lifecycle.

sdks/java/io/couchbase/src/main/java/org/apache/beam/sdk/io/couchbase/CouchbaseIO.java
```diff
@@ -0,0 +1,359 @@
...
+      // because the Couchbase client instance is shared by all the threads.
+    }
+
+    @Override
+    public BoundedSource<JsonDocument> getCurrentSource() {
```
iemejia: Is `JsonDocument` the only supported type? `Document` maybe? or even arbitrary objects?

sdks/java/io/couchbase/src/main/java/org/apache/beam/sdk/io/couchbase/CouchbaseIO.java
aromanenko-dev: I don't think we need to use `BoundedSource` API since there is no benefit of this here. It could be implemented by using just ParDo. I'll elaborate my point in top-level comment of this review.

EdgarLGB:
Thank you @iemejia and @aromanenko-dev for the code review. I really appreciated it! I stand for the view of @aromanenko-dev that it needs to do some rework. What do you think @iemejia ? Meanwhile, I need to handle the points that @iemejia has mentioned.

iemejia:
I would be surprised if there is not a way to estimate the size of the collection in Couchbase. For example [couch_total_disk_size](https://docs.couchbase.com/server/6.0/rest-api/rest-bucket-stats.html)? or something like that.

I am definitely pro changing it to a ParDo based approach in any case. I just thought that the exercise to write the IO based on the `BoundedSource` API was worth, and thought about doing this next, but why not. Remember however that partitioning should be properly done in this case (You can take a look at SolrIO also for an example).

FInal note if you reimplement it based in ParDo probably is worth to implement `readAll` as a base of read, for an example take a look at `HBaseIO.readAll`. The idea is to pass a collection of `queries` to make it composable with only ParDos and with this produce the transform. Notice that HBaseAll uses annotations the new IO API (SplittableDoFn) PLEASE DON'T USE THOSE!. However produce the splits with a `RestrictionTracker` (ByteKey or OffsetRange or whatever they use to represent Ranges in Couchbase) this is a must for future evolution towards SDF. If you have doubts don't hesitate to ask me @EdgarLGB .

aromanenko-dev:
@iemejia I'm not sure that `couch_total_disk_size` works in this case but It's possible to get a [bucket statistics](https://docs.couchbase.com/server/6.0/rest-api/rest-bucket-info.html) which provides `dataUsed` per bucket. So, we could potentially use this but there is another problem - there is no way (or I missed something?) to obtain mappings between bucket shards and offset/key ranges to create a proper splits. If there were such mapping then I believe we can use `BoundedSource` API.

sdks/java/io/couchbase/src/test/java/org/apache/beam/sdk/io/couchbase/CouchbaseIOTest.java
```diff
@@ -0,0 +1,142 @@
...
+            CouchbaseIO.read()
+                .withHosts(Arrays.asList(couchbase.getContainerIpAddress()))
+                .withHttpPort(couchbase.getMappedPort(HTTP_PORT))
+                .withCarrierPort(couchbase.getMappedPort(CARRIER_PORT))
+                .withBucket(BUCKET_NAME)
+                .withPassword(BUCKET_PWD));
+
+    PAssert.thatSingleton(output.apply("Count", Count.globally()))
+        .isEqualTo(Long.valueOf(SAMPLE_SIZE));
+
```
aromanenko-dev: Please, add the following assert to check that all records were properly read:
`PAssert.that(output).containsInAnyOrder(records);`

iemejia:
@aromanenko-dev If there are no ways to obtain mappings between bucket shards and offset/key ranges how do they partition then? Because without this we may have even problems to apply a `ReadAll`-like with `Restriction`s (ranges) pattern.

EdgarLGB:
@iemejia Well, it seems that this mappings are not exposed to the user according to the [documentation](https://docs.couchbase.com/server/5.0/architecture/core-data-access-vbuckets-bucket-partition.html).

aromanenko-dev:
@iemejia This mappings are on low level and this API is not supposed to be exposed to users, as @EdgarLGB said. So, I think we need to follow an approach based on splitting by number of records and offset, and implement it using `ParDo`

EdgarLGB:
I have finished the rework using ParDo. I'm looking forward to some review. Thanks! @iemejia @aromanenko-dev

iemejia:
Run Java PreCommit

iemejia:
Run Portable_Python PreCommit

aromanenko-dev:
Run Java PreCommit

iemejia:
For info I did a review round of this one and will probably take the commit from @EdgarLGB and add some fixes on top because I think we can still improve it.

EdgarLGB:
Run Java PreCommit

EdgarLGB:
Thanks @iemejia! And additionally, I have made minor changes by resolving the conflict with the master branch.

PR #8153, opened 2019-03-27
[BEAM-6324] Remove withWhere from CassandraIO (superceded by withQuery)

iemejia:
R: @echauchot @pabloem

### DIFF
--- a/sdks/java/io/cassandra/src/test/java/org/apache/beam/sdk/io/cassandra/CassandraIOIT.java
+++ b/sdks/java/io/cassandra/src/test/java/org/apache/beam/sdk/io/cassandra/CassandraIOIT.java
@@ -36,7 +36,6 @@ import org.apache.beam.sdk.options.Validation;
 import org.apache.beam.sdk.testing.PAssert;
 import org.apache.beam.sdk.testing.TestPipeline;
 import org.apache.beam.sdk.transforms.Combine;
-import org.apache.beam.sdk.transforms.Count;
 import org.apache.beam.sdk.transforms.DoFn;
 import org.apache.beam.sdk.transforms.ParDo;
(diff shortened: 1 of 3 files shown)


sdks/java/io/cassandra/src/test/java/org/apache/beam/sdk/io/cassandra/CassandraIOTest.java
```diff
@@ -305,26 +285,6 @@ public void testReadWithQuery() throws Exception {
     pipeline.run();
   }
 
-  @Test
-  public void testReadWithValueProvider() throws Exception {
```
iemejia: This one was removed because the functionality is already covered by delegation from the different methods e.g. `withTable(String) -> withTable(ValueProvider)`

sdks/java/io/cassandra/src/test/java/org/apache/beam/sdk/io/cassandra/CassandraIOIT.java
```diff
@@ -106,12 +105,6 @@ public void testWriteThenRead() {
     runRead();
   }
 
-  @Test
-  public void testWriteThenReadWithWhere() {
```
iemejia: This one was removed because the functionality is already tested in the unit tests via `withQuery`, and because ITs it could be easily subsumed by `testWriteThenReadWithWhere`

pabloem:
This is reasonable. We've synced offline, and the short answer is that allowing the user to compose their queries is much more reasonable. Thanks Ismael.

pabloem:
LGTM.

iemejia:
Thanks Pablo for the eager review. I will take care of the ticket now.

PR #8158, opened 2019-03-28
[BEAM-3072] updates to that the error handling and collected the fail…

NikeNano:
Updates to collect more of the error message. This is done by tracing the error message and collect the output from the subprocess during failure.  In order to do this i have added a dependency on the package traceback. 

First contribution to Apache Beam, so feel free to give as much feedback as possible. 
------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/utils/processes_test.py
+++ b/sdks/python/apache_beam/utils/processes_test.py
@@ -18,6 +18,7 @@
 
 from __future__ import absolute_import
 
+import subprocess
 import unittest
 
 import mock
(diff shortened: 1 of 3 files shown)


NikeNano:
@robinyqiu or  @pabloem cloud you please review. Feel free to give a lot of feedback. First contribution and would love to improve :) Thanks!

NikeNano:
Closed by misstake.  Could you please review the pull request @robinyqiu  or @pabloem. Feel free to leave a lot of feedback, my first contribution and would love to learn as much as possible. Thanks!

NikeNano:
Will check the errors and update.

pabloem:
Hi Niklas! Happy to have your contribution. The PreCommit failed, and as a tip to debug these tests with lots of text, you can go to Details > Console Output, then scroll to the Bottom, and you'll find a link to a Gradle scan like so:
image

In the grade scan you can find exactly which tasks failed (in this case, py27 lint, and py27-gcp). The py27 GCP failure looks unrelated, but the lint failures look related to your change: https://scans.gradle.com/s/4ljrrjerslxiy/console-log?task=:beam-sdks-python:lintPy27

To run these checks before pushing your change you can do `tox -e py27-lint` from within `sdks/python` if you have your virtualenv set up, and tox installed.

In short: Thanks for the contribution. Could you fix the lint issues?

aaltay:
@NikeNano Were you able to test the changes and see that errors are captured as expected?

NikeNano:
@pabloem I will fix it :),  @aaltay as far as i understand the ticket, yes :)

sdks/python/scripts/run_pylint.sh
```diff
@@ -24,6 +24,13 @@
 # The exit-code of the script indicates success or a failure.
 
 # Check that the script is running in a known directory.
+echo "START"
```
NikeNano: my bad will remove this, just wanted to check and try to understand some things....

NikeNano:
@aaltay after thinking about it a bit more, could you elaborate on how "hat errors are captured as expected" and i will add some tests for it. I have set it up so that errors ar captured both if the command is not present compare to default "No such file or director" [1]. If the command is valid but the arguments are incorrect an error is rased giving the command stating that the arguments are wrong. 

@pabloem I have resolved the error in the build concerning the formatting. 

[1] https://stackoverflow.com/questions/24306205/file-not-found-error-when-launching-a-subprocess-containing-piped-commands

aaltay:
@NikeNano According to JIRA if a command fails we only log that such a thing happened but do not log why.

Let's say, we are calling out "pip install a-package-that-does-not-exist", In the error log we would like to see the actual pip error saying that this is an non existent package, in addition to the current error message of a process failed with a non-zero return code. This output will be very likely in stderr.

NikeNano:
Run RAT PreCommi

NikeNano:
Run RAT PreCommit

NikeNano:
@aaltay I have updated and added som tests. The failures atm dosn't seam to be connected to my work but rather something with the jenkins jobs, is that correct?

NikeNano:
@pabloem I have solved all the problems and also added test cases. Would be great if you could provide feedback. 

@aaltay I have updated and added test cases.

pabloem:
Thanks Niklas! Taking a quick look...

NikeNano:
R: @pabloem did you had the opportunity to look at it?

pabloem:
This LGTM. Thanks Niklas! I'll merge now, and if there are more issues, wwe can iterate.

sdks/python/apache_beam/utils/processes_test.py
```diff
@@ -103,5 +104,124 @@ def test_method_forwarding_windows(self, *unused_mocks):
...
+  def setup_class(cls):
+    cls.mock_get_patcher = mock.patch(\
+      'apache_beam.utils.processes.subprocess.check_call')
+    cls.mock_get = cls.mock_get_patcher.start()
+
+  @classmethod
+  def teardown_class(cls):
+    cls.mock_get_patcher.stop()
+
+  def test_oserror_check_call(self):
```
tvalentyn: These tests fail for me when run with 

```
python ./setup.py test -s apache_beam.utils.processes_test.TestErrorHandlingCheckOutput
```
```
ERROR: test_oserror_check_output_message (apache_beam.utils.processes_test.TestErrorHandlingCheckOutput)
  File "/usr/local/home/projects/beam/beam/beam/sdks/python/apache_beam/utils/processes_test.py", line 162, in test_oserror_check_output_message
    self.mock_get.side_effect = OSError()
AttributeError: 'TestErrorHandlingCheckOutput' object has no attribute 'mock_get'
```
I think the reason is that `setup_class` is never called.
But the tests pass when run inside tox suite:  `tox -e py27`.
Looking further to understand why.

NikeNano: I tested and got the same result, can't understand why though. Did you find anything so far?

pabloem: I might be wrong, but I find that unittest expects the method to be `setUpClass` (https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUpClass). Perhaps tox knows to run `setup_class`, but other way of running the test will miss the `setup_class` method because it's not what unittest expects?

tvalentyn: `setup_class` method is called by nose: https://github.com/nose-devs/nose/blob/7c26ad1e6b7d308cafa328ad34736d34028c122a/nose/suite.py#L134. tox suites executes all tests using nose, however when we run individual tests we run plain `unittest.runner`.

tvalentyn: Looks like `s/setup_class/setUpClass` is a quick fix. I'll send a PR. If I run into additional issues, we can consider rolling this back until a better fix is in place.

tvalentyn: Sent https://github.com/apache/beam/pull/8337.

NikeNano: Sounds good to me.

PR #8177, opened 2019-03-30
There is an additional "." in example source code.

lesseasyway:
I think it's a very simple change, so I could skip contribution steps. If not, sorry for my wrong decision.


adude3141:
LGTM. Thanks for your contribution!

PR #8178, opened 2019-03-30
[BEAM-5723] Do not use the default shadow closure when building CassandraIO

kennknowles:
This addresses https://issues.apache.org/jira/browse/BEAM-5723

A part of https://issues.apache.org/jira/browse/BEAM-6620

Instead of extending the default shadow closure, replace it.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/io/cassandra/build.gradle
+++ b/sdks/java/io/cassandra/build.gradle
@@ -18,11 +18,23 @@
 
 plugins { id 'org.apache.beam.module' }
 // Do not relocate guava to avoid issues with Cassandra's version.
-applyJavaNature(shadowClosure: DEFAULT_SHADOW_CLOSURE << {
-  dependencies {
-    exclude(dependency(project.library.java.guava))
-  }
-})
+applyJavaNature(
+    shadowClosure: {
+      dependencies {
+        // is default, but when omitted the default action is to include all runtime deps
+        include(dependency(project.library.java.guava))
+
+        // hack: now exclude the only thing that was included
+        exclude(dependency(project.library.java.guava))
+
+      }
+    },
+    shadowJarValidationExcludes: [
+        "org/apache/beam/**",
+        "com/google/common/**",
+        "com/google/thirdparty/**"
+    ]
+)
 provideIntegrationTestingDependencies()
 enableJavaPerformanceTesting()
 


aromanenko-dev:
Run Java PreCommit

iemejia:
@kennknowles it seems we need to exclude this from `validateShadedJarDoesntLeakNonProjectClasses` too.

kennknowles:
This actually exposed a huge bug in our build, if I understand it. It isn't just the Guava classes used here, but all of the ones relocated by any other module.

kennknowles:
Ah, no it is just that excludes are not additive to the default set.

kennknowles:
Hmm, I am not exactly sure why, but the failure is listing essentially every class. It may have to do with not having this block:

```
dependencies {
  include(dependency(project.library.java.guava))
}
```

Everything else in the `DEFAULT_SHADOW_CLOSURE` is just relocations of Guava.

Almost everything causing the relocation error should be OK, and should not be in the shaded jar.

sdks/java/io/cassandra/build.gradle
```diff
@@ -18,11 +18,23 @@
...
-})
+applyJavaNature(
+    shadowClosure: {
+      dependencies {
+        // is default, but when omitted the default action is to include all runtime deps
+        include(dependency(project.library.java.guava))
+
+        // hack: now exclude the only thing that was included
+        exclude(dependency(project.library.java.guava))
+
```
kennknowles: The default is the `runtime` configuration and it has to be overridden. We never did that because we use the dependency filters.

Something that I think should work instead is this:

```groovy
shadowJar {
  configurations = [project.configurations.compile]
}
```

I tried the above and it did not work. I think maybe it is the order of the concatenation of the `shadowClosure` with the parts of the configuration that are hardcoded into the main plugin. I am not sure. And I'm not sure if there is a var I can just set `dependencies = []`. Basically we want to turn off the shadow plugin here but I don't want to make such a big change.

I manually confirmed the contents of the jar.

sdks/java/io/cassandra/build.gradle
```diff
@@ -18,11 +18,23 @@
...
-applyJavaNature(shadowClosure: DEFAULT_SHADOW_CLOSURE << {
-  dependencies {
-    exclude(dependency(project.library.java.guava))
-  }
-})
+applyJavaNature(
+    shadowClosure: {
+      dependencies {
+        // is default, but when omitted the default action is to include all runtime deps
+        include(dependency(project.library.java.guava))
```
aromanenko-dev: Would something like include(dependency { return false }) be more meaningful or not?

kennknowles:
Thanks for checking so thoroughly!

kennknowles:
It looks like this may have broken `:beam-sdks-java-io-cassandra:test`

kennknowles:
Well, it passed in presubmit. It could be this or Etienne's or a combo of them.

aromanenko-dev:
It passes locally for me, maybe it's a combo or Cassandra-related tests flakiness

PR #8191, opened 2019-04-01
Fixing BigQueryFileLoads test for windows.

pabloem:
r: @udim

### DIFF
--- a/sdks/python/apache_beam/io/gcp/bigquery_file_loads.py
+++ b/sdks/python/apache_beam/io/gcp/bigquery_file_loads.py
@@ -42,7 +42,6 @@ import apache_beam as beam
 from apache_beam import pvalue
 from apache_beam.io import filesystems as fs
 from apache_beam.io.gcp import bigquery_tools
-from apache_beam.io.gcp.internal.clients import bigquery as bigquery_api
 from apache_beam.options import value_provider as vp
 from apache_beam.options.pipeline_options import GoogleCloudOptions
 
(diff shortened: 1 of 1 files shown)


sdks/python/apache_beam/io/gcp/bigquery_file_loads.py
```diff
@@ -90,9 +89,12 @@ def _generate_file_prefix(pipeline_gcs_location):
 
 
 def _make_new_file_writer(file_prefix, destination):
-  if isinstance(destination, bigquery_api.TableReference):
-    destination = '%s:%s.%s' % (
-        destination.projectId, destination.datasetId, destination.tableId)
+  destination = bigquery_tools.get_hashable_destination(destination)
```
udim: Is there a reason to use the `:`? For compatibility with legacy SQL?
https://cloud.google.com/bigquery/docs/reference/standard-sql/migrating-from-legacy-sql#project-qualified_table_names

sdks/python/apache_beam/io/gcp/bigquery_file_loads.py
```diff
@@ -90,9 +89,12 @@ def _generate_file_prefix(pipeline_gcs_location):
...
+
+  # Windows does not allow : on filenames. Replacing with underscore.
+  # Other disallowed characters are:
+  # https://docs.microsoft.com/en-us/windows/desktop/fileio/naming-a-file
+  destination = destination.replace(':', '_')
```
udim: Suggestion: replace the `:` with a dot, '.', since it's consistent with BigQuery standard SQL (see previous comment).

pabloem:
I've replaced `_` with `.`. I'm leaving the `:` to `.` for later.

pabloem:
That's the reason I'm not replacing ATM- because I'm not sure if it could break anything.

PR #8206, opened 2019-04-02
[BEAM-6695] Latest PTransform for Python SDK

ttanay:
Added Latest PTransform and Combine Fns for the Python SDK.
Latest PTransform is used to compute the element(s) with the
latest timestamp from a PCollection.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/transforms/combiners.py
+++ b/sdks/python/apache_beam/transforms/combiners.py
@@ -867,6 +867,8 @@ class PhasedCombineFnExecutor(object):
 class Latest(object):
   """Combiners for computing the latest element"""
 
+  @with_input_types(T)
+  @with_output_types(T)
   class Globally(ptransform.PTransform):
     """Compute the element with the latest timestamp from a
     PCollection."""
@@ -881,6 +883,8 @@ class Latest(object):
               .with_output_types(Tuple[T, TimestampType])
               | core.CombineGlobally(LatestCombineFn()))
 
+  @with_input_types(KV[K, V])
+  @with_output_types(KV[K, V])
   class PerKey(ptransform.PTransform):
     """Compute elements with the latest timestamp for each key
     from a keyed PCollection"""
--- a/sdks/python/apache_beam/transforms/combiners.py
+++ b/sdks/python/apache_beam/transforms/combiners.py
@@ -45,7 +45,6 @@ from apache_beam.typehints import TypeVariable
 from apache_beam.typehints import Union
 from apache_beam.typehints import with_input_types
 from apache_beam.typehints import with_output_types
-from apache_beam.typehints.decorators import _check_instance_type
 from apache_beam.utils.timestamp import Duration
 from apache_beam.utils.timestamp import Timestamp
 
@@ -888,7 +887,6 @@ class Latest(object):
 
     @staticmethod
     def add_timestamp(element, timestamp=core.DoFn.TimestampParam):
-      _check_instance_type(KV[K, V], element)
       key, value = element
       return [(key, (value, timestamp))]
 
--- a/sdks/python/apache_beam/transforms/combiners.py
+++ b/sdks/python/apache_beam/transforms/combiners.py
@@ -885,8 +885,8 @@ class Latest(object):
     @staticmethod
     def add_timestamp(element, timestamp=core.DoFn.TimestampParam):
       _check_instance_type(KV[T, T], element)
-      (K, V) = element
-      return [(K, (V, timestamp))]
+      key, value = element
+      return [(key, (value, timestamp))]
 
     def expand(self, pcoll):
       return (pcoll
--- a/sdks/python/apache_beam/transforms/combiners.py
+++ b/sdks/python/apache_beam/transforms/combiners.py
@@ -46,6 +46,8 @@ from apache_beam.typehints import Union
 from apache_beam.typehints import with_input_types
 from apache_beam.typehints import with_output_types
 from apache_beam.typehints.decorators import _check_instance_type
+from apache_beam.utils.timestamp import Duration
+from apache_beam.utils.timestamp import Timestamp
 
 __all__ = [
     'Count',
@@ -61,6 +63,7 @@ __all__ = [
 T = TypeVariable('T')
 K = TypeVariable('K')
 V = TypeVariable('V')
+TimestampType = Union[int, long, float, Timestamp, Duration]
 
 
 class Mean(object):
@@ -876,6 +879,7 @@ class Latest(object):
     def expand(self, pcoll):
       return (pcoll
               | core.ParDo(self.add_timestamp)
+              .with_output_types(Tuple[T, TimestampType])
               | core.CombineGlobally(LatestCombineFn()))
 
   class PerKey(ptransform.PTransform):
@@ -884,17 +888,18 @@ class Latest(object):
 
     @staticmethod
     def add_timestamp(element, timestamp=core.DoFn.TimestampParam):
-      _check_instance_type(KV[T, T], element)
+      _check_instance_type(Tuple[T, T], element)
       key, value = element
       return [(key, (value, timestamp))]
 
     def expand(self, pcoll):
       return (pcoll
               | core.ParDo(self.add_timestamp)
+              .with_output_types(Tuple[T, Tuple[T, TimestampType]])
               | core.CombinePerKey(LatestCombineFn()))
 
 
-@with_input_types(KV[T, T])
+@with_input_types(Tuple[T, TimestampType])
 @with_output_types(T)
 class LatestCombineFn(core.CombineFn):
   """CombineFn to get the element with the latest timestamp
--- a/sdks/python/apache_beam/transforms/combiners.py
+++ b/sdks/python/apache_beam/transforms/combiners.py
@@ -45,6 +45,7 @@ from apache_beam.typehints import TypeVariable
 from apache_beam.typehints import Union
 from apache_beam.typehints import with_input_types
 from apache_beam.typehints import with_output_types
+from apache_beam.typehints.decorators import _check_instance_type
 
 __all__ = [
     'Count',
@@ -53,6 +54,7 @@ __all__ = [
     'Top',
     'ToDict',
     'ToList',
+    'Latest'
     ]
 
 # Type variables
@@ -858,3 +860,60 @@ class PhasedCombineFnExecutor(object):
 
   def extract_only(self, accumulator):
     return self.combine_fn.extract_output(accumulator)
+
+
+class Latest(object):
+  """Combiners for computing the latest element"""
+
+  class Globally(ptransform.PTransform):
+    """Compute the element with the latest timestamp from a
+    PCollection."""
+
+    @staticmethod
+    def add_timestamp(element, timestamp=core.DoFn.TimestampParam):
+      return [(element, timestamp)]
+
+    def expand(self, pcoll):
+      return (pcoll
+              | core.ParDo(self.add_timestamp)
+              | core.CombineGlobally(LatestCombineFn()))
+
+  class PerKey(ptransform.PTransform):
+    """Compute elements with the latest timestamp for each key
+    from a keyed PCollection"""
+
+    @staticmethod
+    def add_timestamp(element, timestamp=core.DoFn.TimestampParam):
+      _check_instance_type(KV[T, T], element)
+      (K, V) = element
+      return [(K, (V, timestamp))]
+
+    def expand(self, pcoll):
+      return (pcoll
+              | core.ParDo(self.add_timestamp)
+              | core.CombinePerKey(LatestCombineFn()))
+
+
+@with_input_types(KV[T, T])
+@with_output_types(T)
+class LatestCombineFn(core.CombineFn):
+  """CombineFn to get the element with the latest timestamp
+  from a PCollection."""
+
+  def create_accumulator(self):
+    return (None, window.MIN_TIMESTAMP)
+
+  def add_input(self, accumulator, element):
+    if accumulator[1] > element[1]:
+      return accumulator
+    else:
+      return element
+
+  def merge_accumulators(self, accumulators):
+    result = self.create_accumulator()
+    for accumulator in accumulators:
+      result = self.add_input(result, accumulator)
+    return result
+
+  def extract_output(self, accumulator):
+    return accumulator[0]
--- a/sdks/python/apache_beam/transforms/combiners.py
+++ b/sdks/python/apache_beam/transforms/combiners.py
@@ -888,14 +888,14 @@ class Latest(object):
 
     @staticmethod
     def add_timestamp(element, timestamp=core.DoFn.TimestampParam):
-      _check_instance_type(Tuple[T, T], element)
+      _check_instance_type(KV[K, V], element)
       key, value = element
       return [(key, (value, timestamp))]
 
     def expand(self, pcoll):
       return (pcoll
               | core.ParDo(self.add_timestamp)
-              .with_output_types(Tuple[T, Tuple[T, TimestampType]])
+              .with_output_types(KV[K, Tuple[T, TimestampType]])
               | core.CombinePerKey(LatestCombineFn()))
 
 
--- a/sdks/python/apache_beam/transforms/combiners_test.py
+++ b/sdks/python/apache_beam/transforms/combiners_test.py
@@ -397,25 +397,31 @@ class CombineTest(unittest.TestCase):
 class LatestTest(unittest.TestCase):
 
   def test_globally(self):
-    l = [window.GlobalWindows.windowed_value(3, 100),
-         window.GlobalWindows.windowed_value(1, 200),
-         window.GlobalWindows.windowed_value(2, 300)]
+    l = [window.TimestampedValue(3, 100),
+         window.TimestampedValue(1, 200),
+         window.TimestampedValue(2, 300)]
     with TestPipeline() as p:
-      pc = p | Create(l)
+      # Map(lambda x: x) PTransform is added after Create here, because when
+      # a PCollection of TimestampedValues is created with Create PTransform,
+      # the timestamps are not assigned to it. Adding a Map forces the
+      # PCollection to go through a DoFn so that the PCollection consists of
+      # the elements with timestamps assigned to them instead of a PCollection
+      # of TimestampedValue(element, timestamp).
+      pc = p | Create(l) | Map(lambda x: x)
       latest = pc | combine.Latest.Globally()
       assert_that(latest, equal_to([2]))
 
   def test_globally_empty(self):
     l = []
     with TestPipeline() as p:
-      pc = p | Create(l)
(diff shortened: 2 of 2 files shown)


ttanay:
Run Portable_Python PreCommit

ttanay:
Run Python PreCommit

ttanay:
Run RAT PreCommit

ttanay:

ttanay:
Run Portable_Python PreCommit

aaltay:
R: @robinyqiu

robinyqiu:
Hi Tanay! Thanks for the contribution. Currently the Python preCommit are all failling. I can review the code while we wait for the preCommit to pass.

ttanay:
Hi @robinyqiu! Sorry I was AFK. I'll fix the failing Python Pre-Commit before the review. I'll ping you once the tests pass.

ttanay:
Hi @robinyqiu the tests have passed. It is ready for review :)

ttanay:
**Observation while working on this:**
A PCollection of TimestampedValues can be created with a DoFn that adds the timestamp info along with the value in a TimestampedValue object, in a ParDo PTransform. This approach is used in the Java SDK.
But, this is not possible in the Python SDK because when the runner evaluates the ParDo, it converts the element of TimestampedValue to a WindowedValue with the same value and timestamp as the TimestampedValue object.
But, this value used for the WindowedValue object is not the TimestampedValue object, which should be the case.

I have a PR on my fork with a failing test to illustrate this: https://github.com/ttanay/beam/pull/6

I'm not sure whether a PCollection of TimestampedValues would be of much use. I found a tuple of (value, timestamp) to be a useful replacement for it. 
But, I don't know whether there may be other cases where this may be needed.

sdks/python/apache_beam/transforms/combiners_test.py
```diff
@@ -392,5 +394,92 @@ def test_global_fanout(self):
       assert_that(result, equal_to([49.5]))
 
 
+class LatestTest(unittest.TestCase):
+
+  def test_globally(self):
+    l = [window.GlobalWindows.windowed_value(1, 100),
```
robinyqiu: Maybe shuffle the example inputs a bit to make them look more random? Like [(2, 100), (3, 300), (1, 200)]? Same below.

sdks/python/apache_beam/transforms/combiners_test.py
```diff
@@ -392,5 +394,92 @@ def test_global_fanout(self):
...
+                    (3, 300)]
+    merged_accumulator = self.fn.merge_accumulators(accumulators)
+    self.assertEquals(merged_accumulator, (3, 300))
+
+  def test_extract_output(self):
+    accumulator = (1, 100)
+    output = self.fn.extract_output(accumulator)
+    self.assertEquals(output, 1)
+
+  def test_input_type_violation(self):
```
robinyqiu: I think this test should go to `LatestTest` instead of `LatestCombineFnTest`?

sdks/python/apache_beam/transforms/combiners_test.py
ttanay: Sure! I'll make the change.

sdks/python/apache_beam/transforms/combiners_test.py
ttanay: This test - `test_input_type_violation` is used to test the type imposed on the input to the CombineFn [here](https://github.com/ttanay/beam/blob/e99e4845410e78d77f9e1f82246c851cf38f308c/sdks/python/apache_beam/transforms/combiners.py#L897). 
The test - `test_per_key_type_violation` is used to check that the input is a KV pair in the DoFn - `add_timestamp` of class `PerKey` [here](https://github.com/ttanay/beam/blob/e99e4845410e78d77f9e1f82246c851cf38f308c/sdks/python/apache_beam/transforms/combiners.py#L887). 

I think `test_per_key_type_violation` can be written better to display the same behaviour. I'll make that change.

ttanay:
Run Portable_Python PreCommit

ttanay:
Run RAT PreCommit

ttanay:

ttanay:
Run Python PreCommit

ttanay:
Run Python PreCommit

ttanay:
Hey @robinyqiu I have made the changes. PTAL.

sdks/python/apache_beam/transforms/combiners.py
```diff
@@ -858,3 +860,60 @@ def merge_only(self, accumulators):
...
+              | core.CombineGlobally(LatestCombineFn()))
+
+  class PerKey(ptransform.PTransform):
+    """Compute elements with the latest timestamp for each key
+    from a keyed PCollection"""
+
+    @staticmethod
+    def add_timestamp(element, timestamp=core.DoFn.TimestampParam):
+      _check_instance_type(KV[T, T], element)
+      (K, V) = element
```
robinyqiu: Maybe consider changing these local variables to (k, v) since K, V are type variables already defined.

aaltay: Or in a similar vein to element, timestamp etc, use key, value = element. Also I believe you do not need the paranthesis on the left.

sdks/python/apache_beam/transforms/combiners.py
```diff
@@ -858,3 +860,60 @@ def merge_only(self, accumulators):
...
+      return [(K, (V, timestamp))]
+
+    def expand(self, pcoll):
+      return (pcoll
+              | core.ParDo(self.add_timestamp)
+              | core.CombinePerKey(LatestCombineFn()))
+
+
+@with_input_types(KV[T, T])
```
robinyqiu: Using KV as type hint seems a bit weird to me because the input is really a (element, timestamp) tuple. Maybe we can just use something like Tuple[E, T] as input type and E as output type. For more about Python type hints, see: https://beam.apache.org/documentation/sdks/python-type-safety/

sdks/python/apache_beam/transforms/combiners.py
```diff
@@ -45,6 +45,7 @@
 from apache_beam.typehints import Union
 from apache_beam.typehints import with_input_types
 from apache_beam.typehints import with_output_types
+from apache_beam.typehints.decorators import _check_instance_type
```
robinyqiu: Functions whose names start with underscore are intended to be used only internally. So we shouldn't import _check_instance_type from another package.

sdks/python/apache_beam/transforms/combiners.py
ttanay: Thank You @robinyqiu and @aaltay. I'll make the change.

sdks/python/apache_beam/transforms/combiners.py
ttanay: I agree. It should be a Tuple rather than a KV.

Also, only types int, float, long, Timestamp and Duration can be interpreted as timestamps[1] so they can be compared to `window.MIN_TIMESTAMP` in the accumulator. So, having the timestamp's type as T would be less restrictive. 
Does it make sense to create a new type - `TimestampT/TimestampType = Union[int, float, long, Timestamp, Duration]` in typehints.py or the combiners module? This would make the input type `Tuple[T, TimestampType]`.

[1] https://github.com/apache/beam/blob/master/sdks/python/apache_beam/utils/timestamp.py#L71

sdks/python/apache_beam/transforms/combiners.py
ttanay: I tried using the decorator `with_input_types`. But, it was comparing the typehint with `WindowedValue`, whereas, the type of the element in the DoFn was a `tuple` which was actually the `WindowedValue` object's value attribute.

sdks/python/apache_beam/transforms/combiners.py
robinyqiu: That's a good suggestion. We can add the definition in this module.
Why is Duration also allowed?

sdks/python/apache_beam/transforms/combiners.py
robinyqiu: `with_input_types` won't work here. Are there other public functions that does the check available? I still believe importing private functions is not a good approach.

sdks/python/apache_beam/transforms/combiners.py
ttanay: Both `Timestamp` and `Duration` have an attribute - `micros`. This is used to compare the two.

https://github.com/apache/beam/blob/master/sdks/python/apache_beam/utils/timestamp.py#L161-L165

sdks/python/apache_beam/transforms/combiners.py
ttanay: There are no public functions that do the same/something similar, except for `with_input_types` AFAIK. 

I think the problem here is that we want to check the type of the element which is different at pipeline construction-time and runtime.
While creating the PCollection, the type of the elements is `WindowedValue` and inside the DoFn, the type of the element is `Tuple`. 

If the input is an object of WindowedValue, any DoFn will have its element param as the value attribute of the WindowedValue.  I think the type evaluated at pipeline construction-time should also be the same. 
If that is the case, then `with_input_types` will work as intended here.

sdks/python/apache_beam/transforms/combiners.py
```diff
@@ -884,17 +888,18 @@ class PerKey(ptransform.PTransform):
 
     @staticmethod
     def add_timestamp(element, timestamp=core.DoFn.TimestampParam):
-      _check_instance_type(KV[T, T], element)
+      _check_instance_type(Tuple[T, T], element)
```
robinyqiu: I believe you have accidentally changed KV to Tuple here, I think here we do want to check the input has KV[K, V] type.

sdks/python/apache_beam/transforms/combiners.py
```diff
@@ -884,17 +888,18 @@ class PerKey(ptransform.PTransform):
...
       key, value = element
       return [(key, (value, timestamp))]
 
     def expand(self, pcoll):
       return (pcoll
               | core.ParDo(self.add_timestamp)
+              .with_output_types(Tuple[T, Tuple[T, TimestampType]])
```
robinyqiu: Same here. I believe it should be KV[K, Tuple[T, TimestampType]]

ttanay: Sorry. My bad. I'll change it.

ttanay:
Hi @robinyqiu 
I've made the changes. PTAL :smile:

robinyqiu:
LGTM, and thank you very much for making these changes!

ttanay:
Thank you @robinyqiu! :smile:

sdks/python/apache_beam/transforms/combiners_test.py
```diff
@@ -392,5 +394,112 @@ def test_global_fanout(self):
...
+    l_3_tuple = [window.GlobalWindows.windowed_value((1, 2, 3), 800),
+                 window.GlobalWindows.windowed_value((4, 5, 6), 200),
+                 window.GlobalWindows.windowed_value((7, 8, 9), 700)]
+
+    with self.assertRaises(TypeCheckError):
+      with TestPipeline() as p:
+        pc = p | Create(l_int)
+        _ = pc | beam.ParDo(combine.Latest.PerKey.add_timestamp)
+
+    with self.assertRaises(TypeCheckError):
```
aaltay: What is being tested here? Why passing KV values here will result in an assert?

sdks/python/apache_beam/transforms/combiners.py
```diff
@@ -53,12 +56,14 @@
...
     'ToDict',
     'ToList',
+    'Latest'
     ]
 
 # Type variables
 T = TypeVariable('T')
 K = TypeVariable('K')
 V = TypeVariable('V')
+TimestampType = Union[int, long, float, Timestamp, Duration]
```
aaltay: Would not this line fail in python 3 because long is not defined?

/cc @tvalentyn

sdks/python/apache_beam/transforms/combiners.py
```diff
@@ -858,3 +863,62 @@ def merge_only(self, accumulators):
...
+              .with_output_types(Tuple[T, TimestampType])
+              | core.CombineGlobally(LatestCombineFn()))
+
+  class PerKey(ptransform.PTransform):
+    """Compute elements with the latest timestamp for each key
+    from a keyed PCollection"""
+
+    @staticmethod
+    def add_timestamp(element, timestamp=core.DoFn.TimestampParam):
+      _check_instance_type(KV[K, V], element)
```
aaltay: You can drop this. The reason is, during pipeline construction these validation will happen, the best thing to do at transform level is to use typehints as much as possible to express what the transform is doing. These are limited but it is better than introducing one off checks inside transforms.

See Top transform for an example: https://github.com/apache/beam/blob/master/sdks/python/apache_beam/transforms/combiners.py#L314


 (This will also be good for not using an internal function.)

sdks/python/apache_beam/transforms/combiners.py
tvalentyn: It would, but `from past.builtins import long` on line 31 makes this code py3-compatible.

sdks/python/apache_beam/transforms/combiners.py
ttanay: Thank You @aaltay! I'll go through it and make the necessary changes.

sdks/python/apache_beam/transforms/combiners_test.py
ttanay: This was to test that the `_check_instance_type` used above raises an Exception. The last positional argument to `window.GlobalWindows.windowed_value` is the timestamp of each element. So, the input PCollections are those that violate the KV constraint.

ttanay:
@aaltay @robinyqiu 
I agree that `_check_instance_type` is not the way to go. 

The need for type validation inside the staticmethod `add_timestamp` of PerKey exists because when using `with_input_types`, the type evaluation does not evaluate PCollections of WindowedValue correctly as needed in this case - the value attribute of a WindowedValue. 
Eg: 
This is from the test: [test_per_key](https://github.com/ttanay/beam/blob/beam-6695/sdks/python/apache_beam/transforms/combiners_test.py#L415)  
```python
# Elements to create PCollection from
elem_list = [window.GlobalWindows.windowed_value(('a', 1), 300),
         window.GlobalWindows.windowed_value(('b', 3), 100),
         window.GlobalWindows.windowed_value(('a', 2), 200)]

with TestPipeline() as p:
      pc = p | Create(elem_list)
      latest = pc | combine.Latest.PerKey()
      assert_that(latest, equal_to([('a', 1), ('b', 3)]))
```
In the staticmethod `add_timestamp` of Latest PTransform, the type of `pc` is evaluated as WindowedValue when using `with_input_types`, whereas, it should be KV since the value attribute of the WindowedValue object is a KV(`('a', 1)`) which is the case when using `_check_instance_type`.

Is this expected behaviour? If not, I'd love to fix this. It would solve the problem.

aaltay:
Try this instead:
elem_list = [('a', TimestampedValue(1, 300)), ('b', TimestampedValue(1, 100)), ('a', TimestampedValue(2, 200))]

In your example, the inputs you are passing are not in type Tuple[K, V]. For example, whatever test you come up with, it should be valid to do Create(elem_list) | GroupByKey(), and that input will work on PerKey transform.

I will defer to @robinyqiu for the remainder of the review.

ttanay:
Thank You @aaltay :smile: 

I tried using TimestampedValue, but, I need to set the timestamp so it is accessible as the TimestampParam in the DoFn - `add_timestamp`, but, it doesn't seem to do that :/ 

@robinyqiu I've removed the call to `_check_instance_type` and the related test - `test_per_key_add_timestamp_element_type_violation` as suggested. :metal:

aaltay:
Thank you @ttanay for your patience.

TimestampedValue is supposed to pack its timestamp. I expected that TimestampParam  should work with that, if not, that might be a bug.

ttanay:
Hi @aaltay sorry for late reply. 
I'll investigate, create a JIRA for it and work on a fix.

robinyqiu:
> In your example, the inputs you are passing are not in type Tuple[K, V]. For example, whatever test you come up with, it should be valid to do Create(elem_list) | GroupByKey(), and that input will work on PerKey transform.
Hi @ttanay , I agree with what @aaltay said here. In other words, we should add type annotations here
`@with_input_types(T) \
 @with_output_types(T)
class Globally(ptransform.PTransform):`
and here
`@with_input_types(KV[K, V]) \
 @with_output_types(KV[K, V]) \
 class PerKey(ptransform.PTransform): ...`

ttanay:
Thank You @robinyqiu! The action points and the hack were really helpful. I now understand why the hack was neccessary. This approach of type-checking looks much cleaner!

sdks/python/apache_beam/transforms/combiners_test.py
```diff
@@ -392,5 +394,86 @@ def test_global_fanout(self):
...
+         window.GlobalWindows.windowed_value(2, 300)]
+    with TestPipeline() as p:
+      pc = p | Create(l)
+      latest = pc | combine.Latest.Globally()
+      assert_that(latest, equal_to([2]))
+
+  def test_globally_empty(self):
+    l = []
+    with TestPipeline() as p:
+      pc = p | Create(l)
```
robinyqiu: I think we should add an explicit `Map` here as well, similar to what we did below, for the sake of consistency.

(This test is passing now only because in the `assert_that` function there is an implicit `Map` transform added to the pipeline.)

sdks/python/apache_beam/transforms/combiners_test.py
```diff
@@ -392,5 +394,86 @@ def test_global_fanout(self):
...
+      latest = pc | combine.Latest.Globally()
+      assert_that(latest, equal_to([None]))
+
+  def test_per_key(self):
+    l = [window.GlobalWindows.windowed_value(('a', 1), 300),
+         window.GlobalWindows.windowed_value(('b', 3), 100),
+         window.GlobalWindows.windowed_value(('a', 2), 200)]
+    with TestPipeline() as p:
+      pc = p | Create(l) | Map(lambda x: x)
```
robinyqiu: Maybe we should add a short comment here to explain why this hack is needed.

sdks/python/apache_beam/transforms/combiners_test.py
```diff
@@ -392,5 +394,86 @@ def test_global_fanout(self):
       assert_that(result, equal_to([49.5]))
 
 
+class LatestTest(unittest.TestCase):
+
+  def test_globally(self):
+    l = [window.GlobalWindows.windowed_value(3, 100),
```
robinyqiu: Consider using `TimestampedValue` instead of `windowed_value`? I think it will make the code easier to understand. Same below.

ttanay:
Made the changes. @robinyqiu PTAL :)

robinyqiu:
Thank you @ttanay for your patience!

@aaltay

aaltay:
Thank you @ttanay and @robinyqiu. This was a great collaboration between you two!

PR #8218, opened 2019-04-03
[BEAM-3545] Support Metrics in Go SDK CombineFns

lostluck:
It was determined that CombineFns should reasonably have metrics. This PR does two things:
* Populate the CombineFn PTransform id into a context.
* Fix an infinite recursive loop in the context code, and added a proper test case for it.

These were caught by a stackoverflow from the latter point, due to the former point.
While it's unlikely for a future transform to be missing it's bundleID (that's applied very early in the stack) it's likely to be missing it's ptransformID, if some new operation is added (eg. SplittableDoFns, timers, etc). Now it will clearly indicate (ptransformID unset) in this case. 

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/go/pkg/beam/core/metrics/metrics.go
+++ b/sdks/go/pkg/beam/core/metrics/metrics.go
@@ -82,15 +82,19 @@ func (ctx *beamCtx) Value(key interface{}) interface{} {
 	switch key {
 	case bundleKey:
 		if ctx.bundleID == "" {
-			if id := ctx.Value(key); id != nil {
+			if id := ctx.Context.Value(key); id != nil {
 				ctx.bundleID = id.(string)
+			} else {
+				return nil
 			}
 		}
 		return ctx.bundleID
 	case ptransformKey:
 		if ctx.ptransformID == "" {
-			if id := ctx.Value(key); id != nil {
+			if id := ctx.Context.Value(key); id != nil {
 				ctx.ptransformID = id.(string)
+			} else {
+				return nil
 			}
 		}
 		return ctx.ptransformID
@@ -118,8 +122,13 @@ func SetPTransformID(ctx context.Context, id string) context.Context {
 	return &beamCtx{Context: ctx, ptransformID: id}
 }
 
+const (
+	bundleIdUnset     = "(bundle id unset)"
+	ptransformIdUnset = "(ptransform id unset)"
+)
+
 func getContextKey(ctx context.Context, n name) key {
-	key := key{name: n, bundle: "(bundle id unset)", ptransform: "(ptransform id unset)"}
+	key := key{name: n, bundle: bundleIdUnset, ptransform: ptransformIdUnset}
 	if id := ctx.Value(bundleKey); id != nil {
 		key.bundle = id.(string)
 	}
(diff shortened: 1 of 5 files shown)


lostluck:
R: @youngoli Please take a look, and ask questions as needed.

sdks/go/pkg/beam/core/metrics/metrics_test.go
```diff
@@ -25,6 +25,44 @@ import (
...
 
+// TestRobustness validates metrics not panicking if the context doesn't
+// have the bundle or transform ID.
+func TestRobustness(t *testing.T) {
+	ns, n := "Test", "myCount"
+	m := NewCounter(ns, n)
+	m.Inc(context.Background(), 3)
+	ptCtx := SetPTransformID(context.Background(), "MY_TRANSFORM")
+	m.Inc(ptCtx, 3)
+	bCtx := SetBundleID(context.Background(), "MY_BUNDLE")
```
youngoli: Should probably use bID here to be consistent with the other tests.

sdks/go/pkg/beam/core/runtime/exec/translate.go
```diff
@@ -375,13 +375,15 @@ func (b *builder) makeLink(from string, id linkID) (Node, error) {
 
 			switch op {
 			case graph.ParDo:
-				n := &ParDo{UID: b.idgen.New(), PID: id.to, Inbound: in, Out: out}
+				n := &ParDo{UID: b.idgen.New(), PID: transform.UniqueName, Inbound: in, Out: out}
```
youngoli: Just checking, this change (id.to -> transform.UniqueName) is because the value of id.to is a name for the transform local to the link, right? i.e. it's not unique within the entire graph?

sdks/go/pkg/beam/core/metrics/metrics.go
```diff
@@ -82,15 +82,19 @@ func (ctx *beamCtx) Value(key interface{}) interface{} {
 	switch key {
 	case bundleKey:
 		if ctx.bundleID == "" {
-			if id := ctx.Value(key); id != nil {
+			if id := ctx.Context.Value(key); id != nil {
```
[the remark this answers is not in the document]
lostluck: This is where the infinite recursion was happening, using ctx.Context.Value is the fix!
Value is a method on the context.Context interface which is ordinarily promoted to beamCtx via embedding. But we're overriding it on beamCtx. In other words, the Value being called was the one we're looking at, instead of the embedded Context's Value. In short, if the value didn't exist, it would recursively call itself forever.
As you've noted, calling ctx.Context.Value makes it explicit we want the embedded types implementation, rather than the beamCtx.Value. There's no benefit in this case to call beamCtx.Value since we already know this level doesn't have the bundleID (or ptransformID).

sdks/go/pkg/beam/core/runtime/exec/translate.go
lostluck: Correct, and while this works OK, the unique names are unambiguous, and supposed to be required by the runner spec.

sdks/go/pkg/beam/core/metrics/metrics_test.go
lostluck: Done. Also removed the variables for namespace and name (ns, n) because they weren't re-used in this small test.

sdks/go/pkg/beam/core/metrics/metrics.go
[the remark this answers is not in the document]
youngoli: Ah, that makes sense now. The piece I missed was that the current function was beamCtx.Value.

lostluck:
R: @aaltay Please review and merge. Thanks!

PR #8220, opened 2019-04-03
[BEAM-6935] Spark portable runner: implement side inputs

ibzib:
1. During translation, convert each side input's corresponding RDD into bytes, which bytes are then broadcast to the Spark context. (This is necessary because in Spark we cannot broadcast an RDD directly like how we can broadcast a Dataset in Flink.)
2. Pass reference to the resulting Spark broadcast variable along with a decoder for those bytes to the executable stage function.
3. When the executable stage function needs to read the side inputs, it gets the broadcasted bytes value, then uses the decoder to convert the bytes to the side input value.

Notes:
- Converted `FlinkBatchSideInputHandlerFactory` to a pluggable `BatchSideInputHandlerFactory` to connect Spark implementation with portability framework.
- For now, testing is still just a log statement with the side input value. More formal verification is still a few PRs away.
- It is quite possible we will want to improve performance in the future; this is just a starting point.

R: @angoenka @iemejia 

--- | --- | --- | --- | ---

### DIFF
--- a/runners/spark/src/main/java/org/apache/beam/runners/spark/translation/BoundedDataset.java
+++ b/runners/spark/src/main/java/org/apache/beam/runners/spark/translation/BoundedDataset.java
@@ -46,6 +46,7 @@ public class BoundedDataset<T> implements Dataset {
   private Iterable<WindowedValue<T>> windowedValues;
   private Coder<T> coder;
   private JavaRDD<WindowedValue<T>> rdd;
+  private List<byte[]> clientBytes;
 
   BoundedDataset(JavaRDD<WindowedValue<T>> rdd) {
     this.rdd = rdd;
@@ -67,6 +68,14 @@ public class BoundedDataset<T> implements Dataset {
               .map(CoderHelpers.fromByteFunction(windowCoder));
     }
     return rdd;
+  }
+
+  List<byte[]> getBytes(WindowedValue.WindowedValueCoder<T> wvCoder) {
+    if (clientBytes == null) {
+      JavaRDDLike<byte[], ?> bytesRDD = rdd.map(CoderHelpers.toByteFunction(wvCoder));
+      clientBytes = bytesRDD.collect();
+    }
+    return clientBytes;
   }
 
   Iterable<WindowedValue<T>> getValues(PCollection<T> pcollection) {
--- a/runners/spark/src/main/java/org/apache/beam/runners/spark/translation/SparkBatchPortablePipelineTranslator.java
+++ b/runners/spark/src/main/java/org/apache/beam/runners/spark/translation/SparkBatchPortablePipelineTranslator.java
@@ -22,10 +22,12 @@ import static org.apache.beam.runners.fnexecution.translation.PipelineTranslator
 
 import java.io.IOException;
 import java.util.Collections;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import javax.annotation.Nullable;
 import org.apache.beam.model.pipeline.v1.RunnerApi;
+import org.apache.beam.model.pipeline.v1.RunnerApi.ExecutableStagePayload.SideInputId;
 import org.apache.beam.model.pipeline.v1.RunnerApi.PCollection;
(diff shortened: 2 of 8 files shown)


runners/spark/src/main/java/org/apache/beam/runners/spark/translation/BoundedDataset.java
```diff
@@ -69,6 +70,14 @@
     return rdd;
   }
 
+  List<byte[]> getBytes(WindowedValue.WindowedValueCoder<T> wvCoder) {
+    if (clientBytes == null) {
+      JavaRDDLike<byte[], ?> bytesRDD = rdd.map(CoderHelpers.toByteFunction(wvCoder));
+      clientBytes = bytesRDD.collect();
```
angoenka: Is bytesRDD.collect expected to be invoked only once?

ibzib: `getBytes` implicitly assumes a 1:1 relationship between `this` and the `wvCoder` argument, meaning the value of `bytesRDD.collect` would always be the same for each BoundedDataset object and therefore would be cacheable. (This is similar for the existing `getValues` function.)

angoenka: I would ideally make these method stateless but I think its fine for now as its the same in getValue

ibzib:
Run Java PreCommit

ibzib:
Run Java PreCommit

ibzib:
Run Java PreCommit

angoenka:
Run Java PreCommit

ibzib:
When running this on Python, I'm running into some issues with Spark broadcast variables. When I try to get the value of a broadcast variable, a NullPointerException is thrown. This probably has something to do with the way Spark passes broadcast variables via closures. I will need to update this PR when I have a solution.

ibzib:
> When running this on Python, I'm running into some issues with Spark broadcast variables.

False alarm, there was a different underlying issue. It seems to be working fine.

runners/spark/src/main/java/org/apache/beam/runners/spark/translation/BoundedDataset.java
```diff
@@ -46,6 +46,7 @@
   private Iterable<WindowedValue<T>> windowedValues;
   private Coder<T> coder;
   private JavaRDD<WindowedValue<T>> rdd;
+  private List<byte[]> clientBytes;
```
iemejia: kinda have issues with that name but cannot think at the moment on something better apart of the even worse `bytes`

runners/spark/src/main/java/org/apache/beam/runners/spark/translation/SparkBatchPortablePipelineTranslator.java
```diff
@@ -191,6 +209,29 @@ private static void translateImpulse(
...
+    BoundedDataset<T> dataset = (BoundedDataset<T>) context.popDataset(collectionId);
+    PCollectionNode collectionNode = PipelineNode.pCollection(collectionId, collection);
+    WindowedValueCoder<T> coder;
+    try {
+      coder =
+          (WindowedValueCoder<T>) WireCoders.instantiateRunnerWireCoder(collectionNode, components);
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+    List<byte[]> bytes = dataset.getBytes(coder);
```
iemejia: again uneasy about naming but without real propositions, maybe `dataAsBytes` (mmm nothing seems good enough).

iemejia:
This means that we can have now WordCount in place, have you already tried it? Next probably will need to document a bit the user side + website (I can help with that).

iemejia:
As with the other one I let you deal with the JIRA, not sure if you have a different jira for the streaming case.

ibzib:
> This means that we can have now WordCount in place, have you already tried it?

I have tried it and it works, but we still have to merge a few more PRs before it's ready to put on the website.

PR #8082, opened 2019-03-18
[BEAM-6853] Add sdk-worker-parallelism and environment-cache-millis to python sdk

angoenka:
These options are already added in the java. We need to add them to python sdk to use them in python.

------------------------

Thank you for your contribution! Follow this checklist to help us incorporate your contribution quickly and easily:

 - [ ] [**Choose reviewer(s)**](https://beam.apache.org/contribute/#make-your-change) and mention them in a comment (`R: @username`).
 - [ ] Format the pull request title like `[BEAM-XXX] Fixes bug in ApproximateQuantiles`, where you replace `BEAM-XXX` with the appropriate JIRA issue, if applicable. This will automatically link the pull request to the issue.
 - [ ] If this contribution is large, please file an Apache [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

Lang | SDK | Apex | Dataflow | Flink | Gearpump | Samza | Spark
--- | --- | --- | --- | --- | --- | --- | ---
Go | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go/lastCompletedBuild/) | --- | --- | --- | --- | --- | ---
Java | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Apex/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Dataflow/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Flink/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Flink_Batch/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Flink_Streaming/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Gearpump/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Samza/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Spark/lastCompletedBuild/)
[...]

PR #8036, opened 2019-03-12
[BEAM-5663] Add Python 3.6 and Python 3.7 precommit test suites

RobbeSneyders:
This is is part of a series of PRs with goal to make Apache Beam PY3 compatible. The proposal with the outlined approach has been documented here: https://s.apache.org/beam-python-3.

This PR checks all unit tests for Python 3.6 and 3.7 to get an overview of remaining work.

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

Lang | SDK | Apex | Dataflow | Flink | Gearpump | Samza | Spark
--- | --- | --- | --- | --- | --- | --- | ---
Go | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go/lastCompletedBuild/) | --- | --- | --- | --- | --- | ---
Java | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Apex/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Dataflow/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Flink/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Flink_Batch/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Flink_Streaming/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Gearpump/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Samza/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Spark/lastCompletedBuild/)
[...]

PR #8225, opened 2019-04-04
[BEAM-6942]  Make modifications to pipeline options to be visible to all views.

tvalentyn:
This PR:
-  fixes a bug in pipeline options implementation that makes changes of multi-valued pipeline options in one view to be invisible by other views of the same pipeline options. The bug manifests when  multi-valued options are not empty before modification.
- adds documentation and unit tests for codepaths affected by the bug and the fix.
- cleans up experiment propagation in Dataflow runner and removes the no longer required propagation of dataflow_kms_key.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/options/pipeline_options.py
+++ b/sdks/python/apache_beam/options/pipeline_options.py
@@ -117,11 +117,13 @@ class _BeamArgumentParser(argparse.ArgumentParser):
 
 
 class PipelineOptions(HasDisplayData):
-  """Pipeline options class used as container for command line options.
+  """Subclasses of this class are used as containers for command line options.
 
-  The class is essentially a wrapper over the standard argparse Python module
+  These classes are wrappers over the standard argparse Python module
   (see https://docs.python.org/3/library/argparse.html).  To define one option
-  or a group of options you subclass from PipelineOptions::
+  or a group of options, create a subclass from PipelineOptions.
+
+  Example Usage::
 
     class XyzOptions(PipelineOptions):
 
@@ -134,11 +136,19 @@ class PipelineOptions(HasDisplayData):
   described in the argparse public documentation.
 
   Pipeline objects require an options object during initialization.
-  This is obtained simply by initializing an options class as defined above::
+  This is obtained simply by initializing an options class as defined above.
+
+  Example Usage::
 
     p = Pipeline(options=XyzOptions())
     if p.options.xyz == 'end':
       raise ValueError('Option xyz has an invalid value.')
+
+  Instances of PipelineOptions or any of its subclass have access to values
+  defined by other PipelineOption subclasses (see get_all_options()), and
+  can be converted to an instance of another PipelineOptions subclass
+  (see view_as()). All views share the underlying data structure that stores
+  option key-value pairs.
 
   By default the options classes will use command line arguments to initialize
   the options.
@@ -150,7 +160,7 @@ class PipelineOptions(HasDisplayData):
     arguments and then parse the command line specified by flags or by default
     the one obtained from sys.argv.
 
-    The subclasses are not expected to require a redefinition of __init__.
+    The subclasses of PipelineOptions should not redefine __init__.
 
     Args:
       flags: An iterable of command line arguments to be used. If not specified
@@ -158,19 +168,39 @@ class PipelineOptions(HasDisplayData):
 
       **kwargs: Add overrides for arguments passed in flags.
     """
+    # self._flags stores a list of not yet parsed arguments, typically,
+    # command-line flags. This list is shared across different views.
+    # See: view_as().
     self._flags = flags
-    self._all_options = kwargs
-    parser = _BeamArgumentParser()
 
+    # Build parser that will parse options recognized by the [sub]class of
+    # PipelineOptions whose object is being instantiated.
+    parser = _BeamArgumentParser()
     for cls in type(self).mro():
       if cls == PipelineOptions:
         break
       elif '_add_argparse_args' in cls.__dict__:
         cls._add_argparse_args(parser)
-    # The _visible_options attribute will contain only those options from the
-    # flags (i.e., command line) that can be recognized. The _all_options
-    # field contains additional overrides.
+
+    # The _visible_options attribute will contain options that were recognized
+    # by the parser.
     self._visible_options, _ = parser.parse_known_args(flags)
+
+    # self._all_options is initialized with overrides to flag values,
+    # provided in kwargs, and will store key-value pairs for options recognized
+    # by current PipelineOptions [sub]class and its views that may be created.
+    # See: view_as().
+    # This dictionary is shared across different views, and is lazily updated
+    # as each new views are created.
+    # Users access this dictionary store via __getattr__ / __setattr__ methods.
+    self._all_options = kwargs
+
+    # Initialize values of keys defined by this class.
+    for option_name in self._visible_option_list():
+      # Note that options specified in kwargs will not be overwritten.
+      if option_name not in self._all_options:
+        self._all_options[option_name] = getattr(self._visible_options,
+                                                 option_name)
 
   @classmethod
   def _add_argparse_args(cls, parser):
@@ -246,7 +276,40 @@ class PipelineOptions(HasDisplayData):
     return self.get_all_options(True)
 
   def view_as(self, cls):
+    """Returns a view of current object as provided PipelineOption subclass.
+
+    Example Usage::
+
+      options = PipelineOptions(['--runner', 'Direct', '--streaming'])
+      standard_options = options.view_as(StandardOptions)
+      if standard_options.streaming:
+        # ... start a streaming job ...
+
+    Note that options objects may have multiple views, and modifications
+    of values in any view-object will apply to current object and other
+    view-objects.
+
+    Args:
+      cls: PipelineOptions class or any of its subclasses.
+
+    Returns:
+      An instance of cls that is intitialized using options contained in current
+      object.
+
+    """
     view = cls(self._flags)
+    for option_name in view._visible_option_list():
+      # Initialize values of keys defined by a cls.
+      #
+      # Note that we do initialization only once per key to make sure that
+      # values in _all_options dict are not-recreated with each new view.
+      # This is important to make sure that values of multi-options keys are
+      # backed by the same list across multiple views, and that any overrides of
+      # pipeline options already stored in _all_options are preserved.
+      if option_name not in self._all_options:
+        self._all_options[option_name] = getattr(view._visible_options,
+                                                 option_name)
+    # Note that views will still store _all_options of the source object.
     view._all_options = self._all_options
     return view
 
@@ -264,7 +327,7 @@ class PipelineOptions(HasDisplayData):
     if name[:2] == name[-2:] == '__':
       return object.__getattribute__(self, name)
     elif name in self._visible_option_list():
-      return self._all_options.get(name, getattr(self._visible_options, name))
+      return self._all_options[name]
     else:
       raise AttributeError("'%s' object has no attribute '%s'" %
                            (type(self).__name__, name))
--- a/sdks/python/apache_beam/options/pipeline_options.py
+++ b/sdks/python/apache_beam/options/pipeline_options.py
@@ -70,9 +70,9 @@ class _BeamArgumentParser(argparse.ArgumentParser):
       @classmethod
 
       def _add_argparse_args(cls, parser):
-        parser.add_value_provider_argument('--vp-arg1', default='start')
-        parser.add_value_provider_argument('--vp-arg2')
-        parser.add_argument('--non-vp-arg')
+        parser.add_value_provider_argument('--vp_arg1', default='start')
+        parser.add_value_provider_argument('--vp_arg2')
+        parser.add_argument('--non_vp_arg')
 
   """
   def add_value_provider_argument(self, *args, **kwargs):
@@ -801,13 +801,13 @@ class PortableOptions(PipelineOptions):
               '"<ENV_VAL>"} }. All fields in the json are optional except '
               'command.'))
     parser.add_argument(
-        '--sdk-worker-parallelism', default=None,
+        '--sdk_worker_parallelism', default=None,
         help=('Sets the number of sdk worker processes that will run on each '
               'worker node. Default is 1. If 0, it will be automatically set '
               'by the runner by looking at different parameters (e.g. number '
               'of CPU cores on the worker machine).'))
     parser.add_argument(
-        '--environment-cache-millis', default=0,
+        '--environment_cache_millis', default=0,
         help=('Duration in milliseconds for environment cache within a job. '
               '0 means no caching.'))
 
--- a/sdks/python/apache_beam/options/pipeline_options.py
+++ b/sdks/python/apache_beam/options/pipeline_options.py
@@ -117,7 +117,7 @@ class _BeamArgumentParser(argparse.ArgumentParser):
 
 
 class PipelineOptions(HasDisplayData):
-  """Subclasses of this class are used as containers for command line options.
+  """This class and subclasses are used as containers for command line options.
 
   These classes are wrappers over the standard argparse Python module
   (see https://docs.python.org/3/library/argparse.html).  To define one option
@@ -160,7 +160,7 @@ class PipelineOptions(HasDisplayData):
     arguments and then parse the command line specified by flags or by default
     the one obtained from sys.argv.
 
-    The subclasses of PipelineOptions should not redefine __init__.
+    The subclasses of PipelineOptions do not need to redefine __init__.
 
     Args:
       flags: An iterable of command line arguments to be used. If not specified
--- a/sdks/python/apache_beam/options/pipeline_options_test.py
+++ b/sdks/python/apache_beam/options/pipeline_options_test.py
@@ -120,6 +120,16 @@ class PipelineOptionsTest(unittest.TestCase):
           '--mock_multi_option', action='append', help='mock multi option')
       parser.add_argument('--option with space', help='mock option with space')
 
+  # Use with MockOptions in test cases where multiple option classes are needed.
+  class FakeOptions(PipelineOptions):
+
+    @classmethod
+    def _add_argparse_args(cls, parser):
+      parser.add_argument('--fake_flag', action='store_true', help='fake flag')
+      parser.add_argument('--fake_option', help='fake option')
+      parser.add_argument(
+          '--fake_multi_option', action='append', help='fake multi option')
+
   def test_display_data(self):
     for case in PipelineOptionsTest.TEST_CASES:
       options = PipelineOptions(flags=case['flags'])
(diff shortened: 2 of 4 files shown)


tvalentyn:
Run Python Postcommit

tvalentyn:
Run Python PostCommit

tvalentyn:
R: @udim 
cc: @aaltay, @robertwb

tvalentyn:
Run Python PostCommit

tvalentyn:
retest this please

tvalentyn:
Run Portable_Python PreCommit

tvalentyn:
Python SDK PostCommit Tests

tvalentyn:
Run Portable_Python PreCommit

tvalentyn:
Python SDK PostCommit Tests

tvalentyn:
Run Python PostCommit

tvalentyn:
Run Portable_Python PreCommit

sdks/python/apache_beam/options/pipeline_options.py
```diff
@@ -796,13 +796,13 @@ def _add_argparse_args(cls, parser):
...
     parser.add_argument(
-        '--sdk-worker-parallelism', default=None,
+        '--sdk_worker_parallelism', default=None,
         help=('Sets the number of sdk worker processes that will run on each '
               'worker node. Default is 1. If 0, it will be automatically set '
               'by the runner by looking at different parameters (e.g. number '
               'of CPU cores on the worker machine).'))
     parser.add_argument(
-        '--environment-cache-millis', default=0,
+        '--environment_cache_millis', default=0,
```
tvalentyn: @goenka, @robertwb, I rootcaused remaining failure on this PR to this typo. 
With this PR, but without fixing this typo, PortableRunner tests fail because the value `environment_cache_millis=10000` set in https://github.com/apache/beam/blob/37556f2cd3ce32e23e16cb653632eafd7169c8e2/sdks/python/build.gradle#L182 is propagated, and then we end up losing correct path to a temporary location stored in the environment while running the test. However I don't fully understand why the test without any changes currently passes. It is possible that another error is lurking somewhere. Do you have any ideas? Also, where is  `environment_cache_millis` actually used? 
cc: @aaltay.

tvalentyn:
@robertwb, I'd appreciate your feedback on this change, in particular whether you think modifications of views of PipelineOptions should modify the options, or instead, views should have been made read-only.
Also, do you think methods of PipelineOptions class need to be made thread-safe if modifying views is allowed. Thank you.

robertwb:
I looked at the first commit, and I think this change makes sense--modification of the view should change the (one source of truth) underlying data. I don't think thread safety is required here. 

Don't really have any light to shed on the environment_cache_millis changes though. @angoenka would know more about this.

tvalentyn:
@ehudm, this is ready for review.

tvalentyn:
@udim, this is ready for review.

sdks/python/apache_beam/options/pipeline_options.py
udim: This might be an incomplete implementation. I found this still open bug: https://issues.apache.org/jira/browse/BEAM-6853

tvalentyn: Thanks, left a comment there.

tvalentyn: Looks like these flags are used in https://github.com/apache/beam/blob/master/runners/flink/src/main/java/org/apache/beam/runners/flink/translation/functions/ReferenceCountingFlinkExecutableStageContextFactory.java.
We should look how PortableOptions from Python SDK are currently passed to the Flink Java Runner, as something may be off there.

tvalentyn: @udim pointed me to  the codepath that allows adding options not defined in any of PipelineOptions subclasses: https://github.com/apache/beam/blob/120394be6ca979892bce98bdea6ca01439b22476/sdks/python/apache_beam/options/pipeline_options.py#L227 After looking further, I think I understand what happens:
1) PortableRunner sends a request to get pipeline supported by any given Runner implementing FnAPI: https://github.com/apache/beam/blob/684f8130284a7c7979773300d04e5473ca0ac8f3/sdks/python/apache_beam/runners/portability/portable_runner.py#L251
2) Flink runner replies that `environment_cache_millis` is a supported option.
3) We dynamically try to make `environment_cache_millis` as an option recognizable by arguments the parser in https://github.com/apache/beam/blob/120394be6ca979892bce98bdea6ca01439b22476/sdks/python/apache_beam/runners/portability/portable_runner.py#L281, by specifying add_extra_args_fn.
4)  PortableOptions declares `environment-cache-millis` as an option. Even though the name contains dashes, `argparse` will store the value for this option in a dictionary with a key where dashes are normalized to underscores: https://docs.python.org/dev/library/argparse.html#dest.
5) With this PR, since the value for `environment-cache-millis` is NOT specified on the command line (but a value with underscores is set to 10000), `self._all_options['environment_cache_millis']` is initialized to 0, the current default value. This value takes precedence in `get_all_options()`: https://github.com/apache/beam/blob/120394be6ca979892bce98bdea6ca01439b22476/sdks/python/apache_beam/options/pipeline_options.py#L237, and causes the test to fail.
6) Without this PR, the value for `environment_cache_millis` in `self._all_options` will be initialized to 0 at the first access https://github.com/apache/beam/blob/120394be6ca979892bce98bdea6ca01439b22476/sdks/python/apache_beam/options/pipeline_options.py#L267, but this never happens because `environment_cache_millis` is not used anywhere in Python SDK, so the test passes.

tvalentyn: Further thoughts on this: 
1. Why is some subset of Flink runner options defined in PortableOptions. It seems that we should remove them, since they are not used by PortableRunner? Any thoughts, @angoenka, @robertwb ?
2. If we need to keep `environment_cache_millis` in `PortableOptions`, is the default 0 reasonable?
3. We should avoid using dashes in pipeline options. Emitting a warning seems to be the safest way, failing or converting them to underscores will be breaking changes.

udim: Thanks for the thorough investigation!

sdks/python/apache_beam/options/pipeline_options.py
```diff
@@ -117,28 +117,34 @@ def error(self, message):
...
-  or a group of options you subclass from PipelineOptions::
+  or a group of options, create a subclass from PipelineOptions:
 
-    class XyzOptions(PipelineOptions):
-
-      @classmethod
-      def _add_argparse_args(cls, parser):
-        parser.add_argument('--abc', default='start')
-        parser.add_argument('--xyz', default='end')
+  |  class XyzOptions(PipelineOptions):
```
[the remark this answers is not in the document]
tvalentyn: Thanks, I found a better way to fix docs errors.

tvalentyn:
@aaltay PTAL. If you don't have any comments, let's merge this once tests pass.

tvalentyn:
Thanks for review, @udim.

sdks/python/apache_beam/options/pipeline_options.py
```diff
@@ -70,9 +70,9 @@ class TemplateUserOptions(PipelineOptions):
       @classmethod
 
       def _add_argparse_args(cls, parser):
-        parser.add_value_provider_argument('--vp-arg1', default='start')
-        parser.add_value_provider_argument('--vp-arg2')
-        parser.add_argument('--non-vp-arg')
+        parser.add_value_provider_argument('--vp_arg1', default='start')
```
aaltay: Why do we change these examples? Are not -vp-arg1 a valid argument?

sdks/python/apache_beam/options/pipeline_options.py
```diff
@@ -117,11 +117,13 @@ def error(self, message):
 
 
 class PipelineOptions(HasDisplayData):
-  """Pipeline options class used as container for command line options.
+  """Subclasses of this class are used as containers for command line options.
```
aaltay: nit: This class is also used as a container.

sdks/python/apache_beam/options/pipeline_options.py
```diff
@@ -150,28 +160,48 @@ def __init__(self, flags=None, **kwargs):
     arguments and then parse the command line specified by flags or by default
     the one obtained from sys.argv.
 
-    The subclasses are not expected to require a redefinition of __init__.
+    The subclasses of PipelineOptions should not redefine __init__.
```
aaltay: What happens if they define an __init__ and call super's __init__ as well?

sdks/python/apache_beam/options/pipeline_options.py
```diff
@@ -246,7 +276,40 @@ def display_data(self):
...
+      #
+      # Note that we do initialization only once per key to make sure that
+      # values in _all_options dict are not-recreated with each new view.
+      # This is important to make sure that values of multi-options keys are
+      # backed by the same list across multiple views, and that any overrides of
+      # pipeline options already stored in _all_options are preserved.
+      if option_name not in self._all_options:
+        self._all_options[option_name] = getattr(view._visible_options,
+                                                 option_name)
+    # Note that views will still store _all_options of the source object.
```
aaltay: Would not this result in views that expose attributes that are not meant to be part of that view?

sdks/python/apache_beam/options/pipeline_options.py
```diff
@@ -738,13 +801,13 @@ def _add_argparse_args(cls, parser):
               '"<ENV_VAL>"} }. All fields in the json are optional except '
               'command.'))
     parser.add_argument(
-        '--sdk-worker-parallelism', default=None,
+        '--sdk_worker_parallelism', default=None,
```
aaltay: Would not this change be backward incompatible? Does argparse do an auto conversion?

sdks/python/apache_beam/options/pipeline_options.py
aaltay: Some of these could be removed now that there is RunnerOptions. @mxm will have context on this.

sdks/python/apache_beam/options/pipeline_options.py
tvalentyn: Argparse will use the same key for storing flags with dashes and flags with underscores, but argparse will not consider them the same flags.  See also my comment above: https://github.com/apache/beam/pull/8225#discussion_r274140134

tvalentyn: So if we declare a flag with dashes, argparse will ignore a flag with underscores passed on the command line.

sdks/python/apache_beam/options/pipeline_options.py
tvalentyn: I want to discourage using dashes in the name. Argparse considers them valid, but as @udim has checked, flags with dashes passed in `**kwargs` parameter of PipelineOptions constructor will not overwrite flags with dashes defined in `_add_argparse_args()` method. Argparse normalizes dashes to underscores, but Beam does not. I was thinking about sending a PR to emit a warning when dashes are used to encourage users to stop using them.

sdks/python/apache_beam/options/pipeline_options.py
tvalentyn: Things should work. Do you prefer a different wording? `The subclasses of PipelineOptions do not need to redefine __init__` perhaps?

sdks/python/apache_beam/options/pipeline_options.py
tvalentyn: No, what is exposed by a view will be defined by `getattr` which calls `_visible_option_list`, which returns `_visible_options`, which are the options recognized by the parser which are populated during  `view = cls(self._flags)`. The view may still store default values defined by previous views, but should not expose them. It would be prudent to add a test for this, looks like there isn't one already.

I think there will be undefined behavior if multiple options define the same flag with different default values, and the flag is not set on the command line.

sdks/python/apache_beam/options/pipeline_options.py
tvalentyn: I think `PortableOptions` should contain options that all Portable Runners are expected to support and `RunnerOptions` may be something that will be obtained from the runner and represent runner-specific options that SDK cannot be aware of. Looks like current `RunnerOptions` class is not used, instead we try to get runner-specific options via JobApi, see https://github.com/apache/beam/blob/e4cf63b9c69cdd8ce37208bb301595f5f2af12c8/sdks/python/apache_beam/runners/portability/portable_runner.py#L281 and above. More changes may be needed to create PipelineOptions subclasses dynamically... Oh well. This is beyond the scope of this PR.

sdks/python/apache_beam/options/pipeline_options.py
tvalentyn: Fair point. Updated.

sdks/python/apache_beam/options/pipeline_options.py
tvalentyn: Changed this, ptal.

sdks/python/apache_beam/options/pipeline_options.py
tvalentyn: Added a test.

tvalentyn:
Thanks, @aaltay, PTAL.

tvalentyn:
Precommit failure looks unrelated: https://github.com/apache/beam/pull/8036#issuecomment-482442925.

tvalentyn:
Run Python PreCommit

sdks/python/apache_beam/options/pipeline_options.py
mxm: @tvalentyn Thanks for taking the time to look at this. I agree that the pipeline options retrieval is not obvious and can cause some confusion. However, before we had to replicate options in all SDKs which proofed to be error-prone.

>  Why is some subset of Flink runner options defined in PortableOptions? It seems that we should remove them, since they are not used by PortableRunner? 

`EnvironmentCacheMillis` is only used by the Flink Runner but it was assumed to be applicable to all portable Runners. I can't speak for Dataflow, but I know that it will be useful for Spark as well. It's used for batch workloads where subsequent operators want to reuse the same environment, e.g. read a local file produced by the preceding operator. I think we could find a better solution for this in the future.

The only other option I see is `sdkWorkerParallelism`. We have experimented with this for the Flink Runner with different work loads and I'm sure other Runners may want to have control over the Harness parallelism.

> If we need to keep environment_cache_millis in PortableOptions, is the default 0 reasonable?

`0` is the default in the Java SDK: https://github.com/apache/beam/commit/63ddad144c1ca7066673b0756c25962f24c899ee#diff-a42abc13cb0431ddc407d59b6023c873R92

>Looks like current RunnerOptions class is not used, instead we try to get runner-specific options via JobApi

`RunnerOptions` is merely a place holder. The `DescribePipelineOptions` request returns all available options from the Java SDK, including their default values.

tvalentyn: - Thanks, @mxm. I think the meaning of PortableOptions, and the relationship PortableOptions vs RunnerOptions is not obvious and may confuse users. 
- Keeping options common for all/most runners in PortableOptions SGTM. One possibility can be to unify PortableOptions and RunnerOptions in one class (RunnerOptions?) and document that the list of options in this class is not exhaustive. 
- We'd need to add instructions how a user could find which options are supported by a portable runner X.

sdks/python/apache_beam/options/pipeline_options.py
aaltay: A warning would be good in that case. (We can resolve this comment.)

sdks/python/apache_beam/options/pipeline_options.py
aaltay: New versions looks good, thanks.

sdks/python/apache_beam/options/pipeline_options.py
aaltay: Great, thanks!

sdks/python/apache_beam/options/pipeline_options.py
aaltay: This is a bit confusing.

Would a user's pipeline break if they have been using the 'sdk-worker-parallelism' version in command line or by directly passing to pipeline options in kwargs?

aaltay:
Left one last question related to backward compatibility of 'sdk-worker-parallelism' change. Otherwise this change looks good and we do not need to block on open questions.

Thank you!

sdks/python/apache_beam/options/pipeline_options.py
tvalentyn: Yes, this would have been a backwards-incompatible change. Luckily, looks like sdk-worker-parallelism didn't make it into 2.12.0: https://github.com/apache/beam/blob/release-2.12.0/sdks/python/apache_beam/options/pipeline_options.py#L719.

tvalentyn: The two options with dashes were added in https://github.com/apache/beam/pull/8082

tvalentyn: This may help clarify argparse behavior:
```
In [1]: import argparse

In [2]: parser=argparse.ArgumentParser()

In [3]: parser.add_argument('--flag-with-dash')
Out[3]: _StoreAction(option_strings=['--flag-with-dash'], dest='flag_with_dash', nargs=None, const=None, default=None, type=None, choices=None, help=None, metavar=None)                                          

In [5]: parser.parse_args(['--flag-with-dash=abc'])
Out[5]: Namespace(flag_with_dash='abc')

In [6]: parser.parse_args(['--flag_with_dash=I am using underscores instead of dashes in flag name!'])
usage: ipython [-h] [--flag-with-dash FLAG_WITH_DASH]
ipython: error: unrecognized arguments: --flag_with_dash=I am using underscores instead of dashes in flag name!
An exception has occurred, use %tb to see the full traceback.

In [7]: parser.add_argument('--flag_with_dash')
Out[7]: _StoreAction(option_strings=['--flag_with_dash'], dest='flag_with_dash', nargs=None, const=None, default=None, type=None, choices=None, help=None, metavar=None)                                          

In [8]: parser.parse_args(['--flag_with_dash=Now I can use both dashes and underscores'])
Out[8]: Namespace(flag_with_dash='Now I can use both dashes and underscores')

In [9]: parser.parse_args(['--flag-with-dash=Same store is used for flags with dashes and underscores'])
Out[9]: Namespace(flag_with_dash='Same store is used for flags with dashes and underscores')
```

tvalentyn:
Thanks for detailed review, PTAL, @aaltay .

sdks/python/apache_beam/options/pipeline_options.py
aaltay: Awesome, sounds like we are safe.

sdks/python/apache_beam/options/pipeline_options.py
mxm: I think we could get rid of RunnerOptions. We created this class just as a dummy when we removed the FlinkOptions from the Python SDK after option fetching had been implemented. 

Not a bad idea to rename rename PortableOptions to RunnerOptions, but we would have to check other SDKs for consistency, e.g. Java has PortablePipelineOptions which are also transferred to the Python SDK.

PR #8251, opened 2019-04-08
[BEAM-7029] Add KafkaIO.Read as external transform

mxm:
This adds KafkaIO.Read as an external transform and includes a Python wrapper (ReadFromkafka) for convenience.

CC @tweise @robertwb @ihji @lukecwik  

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
+++ b/sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
@@ -21,12 +21,15 @@ import static org.apache.beam.vendor.guava.v20_0.com.google.common.base.Precondi
 import static org.apache.beam.vendor.guava.v20_0.com.google.common.base.Preconditions.checkNotNull;
 import static org.apache.beam.vendor.guava.v20_0.com.google.common.base.Preconditions.checkState;
 
+import com.google.auto.service.AutoService;
 import com.google.auto.value.AutoValue;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.lang.reflect.Method;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -35,17 +38,22 @@ import java.util.Set;
 import javax.annotation.Nullable;
 import org.apache.beam.sdk.annotations.Experimental;
 import org.apache.beam.sdk.coders.AtomicCoder;
+import org.apache.beam.sdk.coders.ByteArrayCoder;
 import org.apache.beam.sdk.coders.CannotProvideCoderException;
 import org.apache.beam.sdk.coders.Coder;
 import org.apache.beam.sdk.coders.CoderRegistry;
 import org.apache.beam.sdk.coders.KvCoder;
 import org.apache.beam.sdk.coders.NullableCoder;
+import org.apache.beam.sdk.coders.VarIntCoder;
+import org.apache.beam.sdk.coders.VarLongCoder;
+import org.apache.beam.sdk.expansion.ExternalTransformRegistrar;
 import org.apache.beam.sdk.io.Read.Unbounded;
 import org.apache.beam.sdk.io.UnboundedSource;
 import org.apache.beam.sdk.io.UnboundedSource.CheckpointMark;
 import org.apache.beam.sdk.options.PipelineOptions;
 import org.apache.beam.sdk.options.ValueProvider;
 import org.apache.beam.sdk.transforms.DoFn;
+import org.apache.beam.sdk.transforms.ExternalTransformBuilder;
 import org.apache.beam.sdk.transforms.MapElements;
 import org.apache.beam.sdk.transforms.PTransform;
 import org.apache.beam.sdk.transforms.ParDo;
@@ -57,6 +65,7 @@ import org.apache.beam.sdk.values.PBegin;
 import org.apache.beam.sdk.values.PCollection;
 import org.apache.beam.sdk.values.PDone;
 import org.apache.beam.vendor.guava.v20_0.com.google.common.annotations.VisibleForTesting;
+import org.apache.beam.vendor.guava.v20_0.com.google.common.base.Charsets;
 import org.apache.beam.vendor.guava.v20_0.com.google.common.base.Joiner;
 import org.apache.beam.vendor.guava.v20_0.com.google.common.collect.ImmutableList;
 import org.apache.beam.vendor.guava.v20_0.com.google.common.collect.ImmutableMap;
@@ -351,7 +360,8 @@ public class KafkaIO {
     abstract Builder<K, V> toBuilder();
 
     @AutoValue.Builder
-    abstract static class Builder<K, V> {
+    abstract static class Builder<K, V>
+        implements ExternalTransformBuilder<External.Configuration, PBegin, PCollection<KV<K, V>>> {
       abstract Builder<K, V> setConsumerConfig(Map<String, Object> config);
 
       abstract Builder<K, V> setTopics(List<String> topics);
@@ -386,6 +396,121 @@ public class KafkaIO {
       abstract Builder<K, V> setOffsetConsumerConfig(Map<String, Object> offsetConsumerConfig);
 
       abstract Read<K, V> build();
+
+      @Override
+      public PTransform<PBegin, PCollection<KV<K, V>>> buildExternal(
+          External.Configuration config) {
+        ImmutableList.Builder<String> listBuilder = ImmutableList.builder();
+        for (byte[] topic : config.topics) {
+          listBuilder.add(utf8String(topic));
+        }
+        setTopics(listBuilder.build());
+
+        String keyDeserializerClassName = utf8String(config.keyDeserializer);
+        Class keyDeserializer = resolveClass(keyDeserializerClassName);
+        setKeyDeserializer(keyDeserializer);
+        setKeyCoder(resolveCoder(keyDeserializer));
+
+        String valueDeserializerClassName = utf8String(config.valueDeserializer);
+        Class valueDeserializer = resolveClass(valueDeserializerClassName);
+        setValueDeserializer(valueDeserializer);
+        setValueCoder(resolveCoder(valueDeserializer));
+
+        Map<String, Object> consumerConfig = new HashMap<>();
+        for (KV<byte[], byte[]> kv : config.consumerConfig) {
+          String key = utf8String(kv.getKey());
+          String value = utf8String(kv.getValue());
+          consumerConfig.put(key, value);
+        }
+        // Key and Value Deserializers always have to be in the config.
+        consumerConfig.put(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, keyDeserializer.getName());
+        consumerConfig.put(
+            ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, valueDeserializer.getName());
+        setConsumerConfig(consumerConfig);
+
+        // Set required defaults
+        setTopicPartitions(Collections.emptyList());
+        setConsumerFactoryFn(Read.KAFKA_CONSUMER_FACTORY_FN);
+        setMaxNumRecords(Long.MAX_VALUE);
+        setCommitOffsetsInFinalizeEnabled(false);
+        setTimestampPolicyFactory(TimestampPolicyFactory.withProcessingTime());
+        // We do not include Metadata until we can encode KafkaRecords cross-language
+        return build().withoutMetadata();
+      }
+
+      private static Coder resolveCoder(Class deserializer) {
+        for (Method method : deserializer.getDeclaredMethods()) {
+          if (method.getName().equals("deserialize")) {
+            Class<?> returnType = method.getReturnType();
+            if (returnType.equals(Object.class)) {
+              continue;
+            }
+            if (returnType.equals(byte[].class)) {
+              return ByteArrayCoder.of();
+            } else if (returnType.equals(Integer.class)) {
+              return VarIntCoder.of();
+            } else if (returnType.equals(Long.class)) {
+              return VarLongCoder.of();
+            } else {
+              throw new RuntimeException("Couldn't infer Coder from " + deserializer);
+            }
+          }
+        }
+        throw new RuntimeException("Couldn't resolve coder for Deserializer: " + deserializer);
+      }
+
+      private static Class resolveClass(String className) {
+        try {
+          return Class.forName(className);
+        } catch (ClassNotFoundException e) {
+          throw new RuntimeException("Could not find deserializer class: " + className);
+        }
+      }
+
+      private static String utf8String(byte[] bytes) {
+        return new String(bytes, Charsets.UTF_8);
+      }
+    }
+
+    /**
+     * Exposes {@link KafkaIO.TypedWithoutMetadata} as an external transform for cross-language
+     * usage.
+     */
+    @AutoService(ExternalTransformRegistrar.class)
+    public static class External implements ExternalTransformRegistrar {
+
+      public static final String URN = "beam:external:java:kafka:read:v1";
+
+      @Override
+      public Map<String, Class<? extends ExternalTransformBuilder>> knownBuilders() {
+        return ImmutableMap.of(URN, AutoValue_KafkaIO_Read.Builder.class);
+      }
+
+      /** Parameters class to expose the transform to an external SDK. */
+      public static class Configuration {
+
+        // All byte arrays are UTF-8 encoded strings
+        private Iterable<KV<byte[], byte[]>> consumerConfig;
+        private Iterable<byte[]> topics;
+        private byte[] keyDeserializer;
+        private byte[] valueDeserializer;
+
+        public void setConsumerConfig(Iterable<KV<byte[], byte[]>> consumerConfig) {
+          this.consumerConfig = consumerConfig;
+        }
+
+        public void setTopics(Iterable<byte[]> topics) {
+          this.topics = topics;
+        }
+
+        public void setKeyDeserializer(byte[] keyDeserializer) {
+          this.keyDeserializer = keyDeserializer;
+        }
+
+        public void setValueDeserializer(byte[] valueDeserializer) {
+          this.valueDeserializer = valueDeserializer;
+        }
+      }
     }
 
     /** Sets the bootstrap servers for the Kafka consumer. */
--- a/sdks/python/apache_beam/io/external/kafka.py
+++ b/sdks/python/apache_beam/io/external/kafka.py
@@ -1,0 +1,152 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+"""
+  PTransforms for supporting Kafka in Python pipelines. These transforms do not
+  run a Kafka client in Python. Instead, they expand to ExternalTransform and
+  utilize the Java SDK's Kafka IO. The expansion service will insert Kafka Java
+  transforms before the pipeline is executed. Users currently have to provide
+  the address of the Java expansion service. Flink Users can use the built-in
+  expansion service of the Flink Runner's job server.
+"""
+
+from __future__ import absolute_import
+
+from apache_beam import ExternalTransform
+from apache_beam import pvalue
+from apache_beam.coders import BytesCoder
+from apache_beam.coders import IterableCoder
+from apache_beam.coders import TupleCoder
+from apache_beam.coders.coders import LengthPrefixCoder
+from apache_beam.portability.api.external_transforms_pb2 import ConfigValue
+from apache_beam.portability.api.external_transforms_pb2 import ExternalConfigurationPayload
+from apache_beam.transforms import ptransform
+
+
+class ReadFromKafka(ptransform.PTransform):
+  """
+    An external PTransform which reads from Kafka and returns a KV pair for
+    each item in the specified Kafka topics. If no Kafka Deserializer for
+    key/value is provided, then the data will be returned as a raw byte array.
+
+    Note: To use this transform, you need to start the Java expansion service.
+    Please refer to the portability documentation on how to do that. The
+    expansion service address has to be provided when instantiating this
+    transform. During pipeline translation this transform will be replaced by
+    the Java SDK's KafkaIO.
+
+    If you start Flink's job server, the expansion service will be started on
+    port 8097. This is also the configured default for this transform. For a
+    different address, please set the expansion_service parameter.
+
+    For more information see:
+    - https://beam.apache.org/documentation/runners/flink/
+    - https://beam.apache.org/roadmap/portability/
+
+    Note: Runners need to support translating Read operations in order to use
+    this source. At the moment only the Flink Runner supports this.
+  """
+
+  # Returns the key/value data as raw byte arrays
+  byte_array_deserializer = 'org.apache.kafka.common.serialization.' \
+                            'ByteArrayDeserializer'
+
+  def __init__(self, consumer_config,
+               topics,
+               key_deserializer=byte_array_deserializer,
+               value_deserializer=byte_array_deserializer,
+               expansion_service='localhost:8097'):
+    """
+    Initializes a read operation from Kafka.
+
+    :param consumer_config: A dictionary containing the consumer configuration.
+    :param topics: A list of topic strings.
+    :param key_deserializer: A fully-qualified Java class name of a Kafka
+                             Deserializer for the topic's key, e.g.
+                             'org.apache.kafka.common.
+                             serialization.LongDeserializer'.
+                             Default: 'org.apache.kafka.common.
+                             serialization.ByteArrayDeserializer'.
+    :param value_deserializer: A fully-qualified Java class name of a Kafka
+                               Deserializer for the topic's value, e.g.
+                               'org.apache.kafka.common.
+                               serialization.LongDeserializer'.
+                               Default: 'org.apache.kafka.common.
+                               serialization.ByteArrayDeserializer'.
+    :param expansion_service: The address (host:port) of the ExpansionService.
+    """
+    super(ReadFromKafka, self).__init__()
+    self._urn = 'beam:external:java:kafka:read:v1'
+    self.consumer_config = consumer_config
+    self.topics = topics
+    self.key_deserializer = key_deserializer
+    self.value_deserializer = value_deserializer
+    self.expansion_service = expansion_service
+
+  def expand(self, pbegin):
+    if not isinstance(pbegin, pvalue.PBegin):
+      raise Exception("ReadFromKafka must be a root transform")
+
+    args = {
+        'consumer_config':
+            ReadFromKafka._encode_map(self.consumer_config),
+        'topics':
+            ReadFromKafka._encode_list(self.topics),
+        'key_deserializer':
+            ReadFromKafka._encode_str(self.key_deserializer),
+        'value_deserializer':
+            ReadFromKafka._encode_str(self.value_deserializer),
+    }
+
+    payload = ExternalConfigurationPayload(configuration=args)
+    return pbegin.apply(
+        ExternalTransform(
+            self._urn,
+            payload.SerializeToString(),
+            self.expansion_service))
+
+  @staticmethod
+  def _encode_map(dict_obj):
+    kv_list = [(key.encode('utf-8'), val.encode('utf-8'))
+               for key, val in dict_obj.items()]
+    coder = IterableCoder(TupleCoder(
+        [LengthPrefixCoder(BytesCoder()), LengthPrefixCoder(BytesCoder())]))
+    coder_urns = ['beam:coder:iterable:v1',
+                  'beam:coder:kv:v1',
+                  'beam:coder:bytes:v1',
+                  'beam:coder:bytes:v1']
+    return ConfigValue(
+        coder_urn=coder_urns,
+        payload=coder.encode(kv_list))
+
+  @staticmethod
+  def _encode_list(list_obj):
+    encoded_list = [val.encode('utf-8') for val in list_obj]
+    coder = IterableCoder(LengthPrefixCoder(BytesCoder()))
+    coder_urns = ['beam:coder:iterable:v1',
+                  'beam:coder:bytes:v1']
+    return ConfigValue(
+        coder_urn=coder_urns,
+        payload=coder.encode(encoded_list))
+
+  @staticmethod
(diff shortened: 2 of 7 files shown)


sdks/python/apache_beam/io/external/kafka.py
```diff
@@ -0,0 +1,102 @@
...
+    self._urn = 'beam:external:java:kafka:read:v1'
+    self.consumer_config = consumer_config
+    self.topics = topics
+    self.key_deserializer = key_deserializer
+    self.value_deserializer = value_deserializer
+    self.expansion_service = expansion_service
+
+  def expand(self, pbegin):
+    if not isinstance(pbegin, pvalue.PBegin):
+      raise Exception("GenerateSequence must be a root transform")
```
ihji: GenerateSequence -> ReadFromKafka

sdks/python/apache_beam/io/external/kafka.py
```diff
@@ -0,0 +1,102 @@
...
+
+    args = {
+        'consumer_config':
+            KafkaRead._encode_map(self.consumer_config),
```
ihji: KafkaRead -> ReadFromKafka

sdks/python/apache_beam/io/external/kafka.py
```diff
@@ -0,0 +1,102 @@
...
+        'topics':
+            KafkaRead._encode_list(self.topics),
```
ihji: KafkaRead -> ReadFromKafka

sdks/python/apache_beam/io/external/kafka.py
```diff
@@ -0,0 +1,102 @@
...
+        'key_deserializer':
+            KafkaRead._encode_str(self.key_deserializer),
```
ihji: KafkaRead -> ReadFromKafka

sdks/python/apache_beam/io/external/kafka.py
```diff
@@ -0,0 +1,102 @@
...
+        'value_deserializer':
+            KafkaRead._encode_str(self.value_deserializer),
```
ihji: KafkaRead -> ReadFromKafka

sdks/python/apache_beam/io/external/kafka.py
mxm: Ah, thanks. I renamed this before opening the PR and missed those.

sdks/python/apache_beam/io/external/kafka.py
mxm: Thanks for spotting.

mxm:
Thanks for taking a look @lukecwik. The design document for this is https://s.apache.org/beam-cross-language-io. Approach (2) in the design doc was the original proposal for configuring cross-language transforms. It is very close to what you are suggesting. The PR #7875 originally had this design, but we changed it to match approach (1) after a discussion in the PR, on the mailing list, and in a meeting which @chamikaramj summarized on the mailing list.

It is not necessarily a contradiction to support both, an explicit Proto for each transform, and a generic approach. However, the main objective for cross-language transforms was that there should not be the need for users to write Proto. So the generic approach was deemed more important.

>Also, how do you "choose" what the output type of the PCollection produced by Kafka.Read is?
If someone passes in the org.apache.kafka.common.serialization.LongDeserializer like in your example, the returned PCollection may contain some Java SDK specific coder. Would it make sense to have a mapping from beam coder URNs to kafka deserializers instead?

Good point. I think it makes sense to define a mapping of supported coders for Kafka deserializers, since we can only support standard SDK coders.

mxm:
Run Java PreCommit

mxm:
Run Java PreCommit

mxm:
Flaky Hadoop-related test: https://builds.apache.org/job/beam_PreCommit_Java_Commit/5269/console

mxm:
Run Python PreCommit

chamikaramj:
@lukecwik I summarized the current design (adapter from Max's original design doc) in an email to dev list: https://lists.apache.org/thread.html/b99ba8527422e31ec7bb7ad9dc3a6583551ea392ebdc5527b5fb4a67@%3Cdev.beam.apache.org%3E

The idea is to minimize the amount of work a transform author has to do to make a given transform available to remote SDKs. With the current solution, the transform authors simply have to provide a builder that can be used to build the transform from a given config object (POJO). Framework will generate a proto from this that defines the wire format. Remote SDKs can invoke arbitrary transforms that are exposed through the expansion API (and has a builder available) by using generic ExternalTransform. Also, transform authors may define nicer wrappers (of ExternalTransform) if they choose to do so.

Of course this means that a cross-SDK interaction can only be done thorough standard coders that are defined for both languages. These will be used for (1) encoding parameters used to build the rewmote transform (2) encoding data transferred between SDKs at runtime. A pipeline author has to be aware of this and has to make sure that transforms at SDK boundaries use standard coders. We hope the solution will become more generic once Beam SQL becomes available (where we should be able to do a Row to Row conversion from SDK A to SDK B for generic types).

mxm:
Run Python PreCommit

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
```diff
@@ -386,6 +397,114 @@
...
+          throw new RuntimeException("Could not find deserializer class: " + className);
+        }
+      }
+
+      private static String utf8String(byte[] bytes) {
+        return new String(bytes, Charsets.UTF_8);
+      }
+    }
+
+    /** Exposes GenerateSequence as an external transform for cross-language usage. */
```
chamikaramj: s/GenerateSequence/KafkaIO.Read

sdks/python/apache_beam/io/external/kafka.py
```diff
@@ -0,0 +1,102 @@
...
+from apache_beam.coders import BytesCoder
+from apache_beam.coders import IterableCoder
+from apache_beam.coders import TupleCoder
+from apache_beam.coders.coders import LengthPrefixCoder
+from apache_beam.portability.api.external_transforms_pb2 import ConfigValue
+from apache_beam.portability.api.external_transforms_pb2 import ExternalConfigurationPayload
+from apache_beam.transforms import ptransform
+
+
+class ReadFromKafka(ptransform.PTransform):
```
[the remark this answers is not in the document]
chamikaramj: Thanks. I think we should also provide end-to-end instructions on how a user can use this transform in a pipeline, including for example, steps for starting up the expansion service (here or elsewhere) given the this is the first real world source that some of our users might try to use out of the box in the next Beam release.

sdks/python/apache_beam/runners/portability/flink_runner_test.py
```diff
@@ -172,6 +173,20 @@ def test_external_transform(self):
 
         assert_that(res, equal_to([i for i in range(1, 10)]))
 
+      # We do not run KafkaIO but check that the transform
+      # is expanded by the ExpansionService without an error.
+      p = self.create_pipeline()
```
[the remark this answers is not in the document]
chamikaramj: Sg. I think a full end-to-end test will definitely be helpful.

lukecwik:
Thanks @mxm and @chamikaramj for the links. I read through them and have a better understanding of the details.

Using the decoded PCollection<keyType, valueType> for KafkaIO.Read will cover many user usecases but from experience with the GCP pubsub native source for Dataflow and the Python SDK users always wanted to get additional attributes from the pubsub message and only passing through the "data" meant that this didn't satisfy what many users wanted (and exposing a few additional attributes at a time wasn't great). This was easy for pubsub since they used proto as the canonical wire format. As a future follow-up it may be useful to also expose a Java KafkaIO.Read that produces org.apache.kafka.common.record.Record encoded as a byte[]. This would mean that the output type of KafkaIO.Read for the cross language transform is PCollection<byte[]> and push all the decoding logic into the Python SDK. This would mean that the user could use any "deserializer" they want but this puts a greater burden on the language that wants to consume the records since they will need to be able to decode such messages. Many times this additional logic in the downstream SDK isn't difficult to implement since you can rely on a language specific source library to do the parsing. Alternatively, if there aren't many language specific libraries for the source format, it may be wise to produce an intermediate format such as proto/json/... (which can generate language specific bindings or are very well supported in almost all languages) which sends all the data across.

chamikaramj:
Makes sense in general. If a IO connector offers a transform that produces records as bytes, it makes sense to support pushing deserializers to remote SDKs.

Looks like there's no such a read transform for Java `KafkaIO` today though. `KafkaIO.TypedWithoutMetadata` produces a `PCollection<KV<K, V>>` while `KafkIO.Read` produces a `PCollection<KafkaRecord<K, V>>` where a `KafkaRecord` wraps` K`, `V` values (and other metadata) extracted from a native Kafka record. Also, `KafkIO.Read`  will not work for cross-language transforms till we convert it to use SDF (instead of `UnboundedSource`).

mxm:
@lukecwik That's correct. During implementation I discovered that data is always wrapped into `KafakRecord` and went with `Read.TypedWithoutMetaData` to get around this. Like you pointed out, we need either 1) a `KafkaRecord` standard coder (perhaps something applicable to other messaging systems as well or 2) `byte[]` where we implement all coding in the SDK. For both we need to provide Coders in the SDK, but the benefit of having a standard coder would be a language-independent encoding scheme which is important when dealing with cross-language pipelines.

> Also, KafkIO.Read will not work for cross-language transforms till we convert it to use SDF (instead of UnboundedSource).

I think we will have to support Read in portable Runners for now until SDF is finalized and an adapter for UnboundedSource is available. For the Flink Runner this is already the case and other Runners can add it easily. It's not so bad because, after all, it is transparent to the user.

sdks/python/apache_beam/io/external/kafka.py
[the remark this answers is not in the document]
chamikaramj: I think yes.  Otherwise users will have no way of knowing how to use this in a pipeline. I think the instructions can be as simple as steps to get the expansion service running (or a pointer to a script for doing this) and providing the address to the ReadFromKafka transform (no need to mention runner setup).

I agree that when things are more stable, we should develop a Webpage for cross-language transforms with complete instructions, examples (for transform authors and users), etc. (not in this PR).

mxm:
Run Java PreCommit

lukecwik:
@mxm Supporting the whole KafkaRecord could be a TODO for the future. For now you could just produce KV<byte[], byte[]> with the intent that the receiving end performs any Kafka key/value -> language specific type decoding (instead of passing in the key / value deserializers to the Java SDK). This will produce a version that has a lot of value for many people since all "data" types can be supported in the Python SDK by exposing a UDF that users can implement to perform the decoding in the Python SDK.

Also, to support the whole KafkaRecord in the future, Kafka must already have a well defined "encoding" format for compatibility across all its client SDKs so it would be great if we could use that and then use the parsing in the Kafka client libraries to do that work. Our coder spec would be beam:coder:kafka_record:v1 and just point to a certain version of the KafkaRecord spec.

mxm:
Run Portable_Python PreCommit

mxm:
@lukecwik I think your experience with Pubsub on Dataflow is a good reminder of the requirements users have when reading from a message queue. For simple use cases, only having the data might be fine but ultimately users will ask for the metadata (partition id, partition offset, timestamp). Defining a standard coder for this seems inevitable.

For now, going with the current approach here to just ship the data seems to be fine. The `KV<byte[], byte[]>`, coder agnostic way makes sense to me because it gives users freedom to implement their own encoding. Note that this is already possible if users configure the `ByteArrayDeserializer`.

I'm wondering whether we should keep the mapping from Kafka Deserializer to standard coders. Most Kafka users are familiar with Java and are used to Kafka Deserializers. The obvious drawback of this approach is that we need to maintain this mapping and we can't support all Deserializers. We could make `KV<byte[], byte[]>` the default and infer types of if Deserializers if they are provided by the user. This might be a more flexible approach.

@lukecwik @chamikaramj WDYT?

chamikaramj:
I think it makes sense to keep a mapping from deserializers to standard coders in Python SDK. This way users can configure next step of the python pipeline to consume data using the corresponding coder. The issue (as you mentioned) is keeping such a list up to date with what happens in the Java SDK. But probably changes to Kafka deserializers are rare enough to justify this.

lukecwik:
I agree with @chamikaramj about making the decoding happen within the Python SDK for the same reason.

sdks/python/apache_beam/io/external/kafka.py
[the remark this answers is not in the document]
mxm: I've included a pointer to the Flink Job Server which starts the Java expansion service. Additionally, I've linked documentation pages which explain how to start the Flink Job Server. Should be sufficient for now.

mxm:
I've made `KV<byte[], byte[]>` the default. Optionally, users can specify a different Deserializer. I've updated the documentation and modified the end-to-end-test in `flink_runner_test.py` to fully execute (and check failure due to missing Kafka cluster). I think we can merge this and then add `WriteToKafka` next.

chamikaramj:
LGTM. Thanks.

mxm:
Thanks for the reviews!

mxm:
Unrelated `WatchTest` failures: https://builds.apache.org/job/beam_PreCommit_Java_Commit/5332/

manuelaguilar:
@mxm I've been trying to use the ReadFromKafka tranform (kafka.py) via Flink with the portable runner, but it seems the value coder resolves to ByteArrayCoder instead of KafkaRecordCoder. Am I missing a registration step? My transform is similar to the one in the unit test (flink_runner_test.py).

Here are some logs with some extra INFO I added:

`
[Source: Read From Kafka/ExternalTransform(beam:external:java:kafka:read:v1)/KafkaIO.Read/Read(KafkaUnboundedSource) -> Flat Map -> [1]Read From Kafka/ExternalTransform(beam:external:java:kafka:read:v1)/Remove Kafka Metadata (1/4)] INFO org.apache.beam.sdk.coders.LengthPrefixCoder - Creating length prefix coder with value coder class org.apache.beam.sdk.coders.ByteArrayCoder
[grpc-default-executor-1] INFO org.apache.beam.fn.harness.FnHarness - Fn Harness started
[grpc-default-executor-1] INFO org.apache.beam.fn.harness.FnHarness - Entering instruction processing loop
[Source: Read From Kafka/ExternalTransform(beam:external:java:kafka:read:v1)/KafkaIO.Read/Read(KafkaUnboundedSource) -> Flat Map -> [1]Read From Kafka/ExternalTransform(beam:external:java:kafka:read:v1)/Remove Kafka Metadata (2/4)] INFO org.apache.beam.runners.core.construction.CoderTranslators - Getting length prefix coder for type class org.apache.beam.sdk.coders.ByteArrayCoder
[Source: Read From Kafka/ExternalTransform(beam:external:java:kafka:read:v1)/KafkaIO.Read/Read(KafkaUnboundedSource) -> Flat Map -> [1]Read From Kafka/ExternalTransform(beam:external:java:kafka:read:v1)/Remove Kafka Metadata (2/4)] INFO org.apache.beam.sdk.coders.LengthPrefixCoder - Creating length prefix coder with value coder class org.apache.beam.sdk.coders.ByteArrayCoder
[Source: Read From Kafka/ExternalTransform(beam:external:java:kafka:read:v1)/KafkaIO.Read/Read(KafkaUnboundedSource) -> Flat Map -> [1]Read From Kafka/ExternalTransform(beam:external:java:kafka:read:v1)/Remove Kafka Metadata (3/4)] INFO org.apache.beam.runners.core.construction.CoderTranslators - Getting length prefix coder for type class org.apache.beam.sdk.coders.ByteArrayCoder
[Source: Read From Kafka/ExternalTransform(beam:external:java:kafka:read:v1)/KafkaIO.Read/Read(KafkaUnboundedSource) -> Flat Map -> [1]Read From Kafka/ExternalTransform(beam:external:java:kafka:read:v1)/Remove Kafka Metadata (3/4)] INFO org.apache.beam.sdk.coders.LengthPrefixCoder - Creating length prefix coder with value coder class org.apache.beam.sdk.coders.ByteArrayCoder
[Source: Read From Kafka/ExternalTransform(beam:external:java:kafka:read:v1)/KafkaIO.Read/Read(KafkaUnboundedSource) -> Flat Map -> [1]Read From Kafka/ExternalTransform(beam:external:java:kafka:read:v1)/Remove Kafka Metadata (4/4)] INFO org.apache.beam.sdk.coders.LengthPrefixCoder - Creating length prefix coder with value coder class org.apache.beam.sdk.coders.ByteArrayCoder
...
[Source: Read From Kafka/ExternalTransform(beam:external:java:kafka:read:v1)/KafkaIO.Read/Read(KafkaUnboundedSource) -> Flat Map -> [1]Read From Kafka/ExternalTransform(beam:external:java:kafka:read:v1)/Remove Kafka Metadata (4/4)] INFO org.apache.beam.sdk.coders.LengthPrefixCoder - Encoding value of type class org.apache.beam.sdk.io.kafka.KafkaRecord with coder of type class org.apache.beam.sdk.coders.ByteArrayCoder
[pool-13-thread-1] INFO org.apache.beam.sdk.io.kafka.KafkaUnboundedSource - Reader-3: Returning from consumer pool loop
[Source: Read From Kafka/ExternalTransform(beam:external:java:kafka:read:v1)/KafkaIO.Read/Read(KafkaUnboundedSource) -> Flat Map -> [1]Read From Kafka/ExternalTransform(beam:external:java:kafka:read:v1)/Remove Kafka Metadata (4/4)] INFO org.apache.flink.runtime.taskmanager.Task - Source: Read From Kafka/ExternalTransform(beam:external:java:kafka:read:v1)/KafkaIO.Read/Read(KafkaUnboundedSource) -> Flat Map -> [1]Read From Kafka/ExternalTransform(beam:external:java:kafka:read:v1)/Remove Kafka Metadata (4/4) (6eabba22b9485eeaf271d72fffdc08b1) switched from RUNNING to FAILED.
java.lang.ClassCastException: org.apache.beam.sdk.io.kafka.KafkaRecord cannot be cast to [B
        at org.apache.beam.sdk.coders.ByteArrayCoder.encode(ByteArrayCoder.java:44)
        at org.apache.beam.sdk.coders.LengthPrefixCoder.encode(LengthPrefixCoder.java:66)
        at org.apache.beam.sdk.values.ValueWithRecordId$ValueWithRecordIdCoder.encode(ValueWithRecordId.java:112)
        at org.apache.beam.sdk.values.ValueWithRecordId$ValueWithRecordIdCoder.encode(ValueWithRecordId.java:87)
        at org.apache.beam.sdk.util.WindowedValue$FullWindowedValueCoder.encode(WindowedValue.java:537)
        at org.apache.beam.sdk.util.WindowedValue$FullWindowedValueCoder.encode(WindowedValue.java:486)
        at org.apache.beam.sdk.util.CoderUtils.encodeToSafeStream(CoderUtils.java:82)
        at org.apache.beam.sdk.util.CoderUtils.encodeToByteArray(CoderUtils.java:66)
        at org.apache.beam.sdk.util.CoderUtils.encodeToByteArray(CoderUtils.java:51)
        at org.apache.beam.sdk.util.CoderUtils.clone(CoderUtils.java:141)
        at org.apache.beam.runners.flink.translation.types.CoderTypeSerializer.copy(CoderTypeSerializer.java:89)
        at org.apache.flink.streaming.runtime.tasks.OperatorChain$CopyingChainingOutput.pushToOperator(OperatorChain.java:577)
        at org.apache.flink.streaming.runtime.tasks.OperatorChain$CopyingChainingOutput.collect(OperatorChain.java:554)
        at org.apache.flink.streaming.runtime.tasks.OperatorChain$CopyingChainingOutput.collect(OperatorChain.java:534)
        at org.apache.flink.streaming.api.operators.AbstractStreamOperator$CountingOutput.collect(AbstractStreamOperator.java:718)
        at org.apache.flink.streaming.api.operators.AbstractStreamOperator$CountingOutput.collect(AbstractStreamOperator.java:696)
        at org.apache.flink.streaming.api.operators.StreamSourceContexts$ManualWatermarkContext.processAndCollectWithTimestamp(StreamSourceContexts.java:310)
        at org.apache.flink.streaming.api.operators.StreamSourceContexts$WatermarkContext.collectWithTimestamp(StreamSourceContexts.java:409)
        at org.apache.beam.runners.flink.translation.wrappers.streaming.io.UnboundedSourceWrapper.emitElement(UnboundedSourceWrapper.java:334)
        at org.apache.beam.runners.flink.translation.wrappers.streaming.io.UnboundedSourceWrapper.run(UnboundedSourceWrapper.java:239)
        at org.apache.flink.streaming.api.operators.StreamSource.run(StreamSource.java:93)
        at org.apache.flink.streaming.api.operators.StreamSource.run(StreamSource.java:57)
        at org.apache.flink.streaming.runtime.tasks.SourceStreamTask.run(SourceStreamTask.java:97)
        at org.apache.flink.streaming.runtime.tasks.StoppableSourceStreamTask.run(StoppableSourceStreamTask.java:45)
        at org.apache.flink.streaming.runtime.tasks.StreamTask.invoke(StreamTask.java:300)
        at org.apache.flink.runtime.taskmanager.Task.run(Task.java:711)
        at java.lang.Thread.run(Thread.java:748)
`

mxm:
Thanks for trying out the Kafka consumer. There are limitations for the consumer to work correctly. The biggest issue is the structure of KafkaIO itself, which uses a combination of the source interface and DoFns to generate the desired output. The problem is that the source interface is natively translated by Flink to support unbounded sources in portability, while the DoFn runs in a Java environment. To transfer data between the two a coder needs to be involved. It happens to be that the initial read does not immediately drop the KafakRecord structure which does not work together well with our current assumption of only supporting "standard coders" present in all SDKs.

There are several possible solutions:
1. Make the DoFn which drops the KafkaRecordCoder a native Java transform in the Flink Runner
2. Modify KafkaIO to immediately drop the KafkaRecord structure
3. Add the KafkaRecordCoder to all SDKs
4. Add a generic coder, e.g. AvroCoder to all SDKs

For a workaround which uses (3), please see this patch which is not a proper fix but adds KafkaRecordCoder to the SDK such that it can be used encode/decode records: https://github.com/mxm/beam/commit/b31cf99c75b3972018180d8ccc7e73d311f4cfed 

I'm planning to fix the issue in the coming weeks.

mxm:
Created a JIRA issue: https://jira.apache.org/jira/browse/BEAM-7870

manuelaguilar:
@mxm I tried the patch and I can see that the right coder is applied to the transform. Thank you.

My next step is to have the transform output (KafkaRecords) to be passed to my Python pipeline.

How can I handover KafkaRecord elements from Java to Python? Will it work by calling an external transform in the Python pipeline or is there a simpler way to translate the ReadFromKafka output into a Python implementation of KafkaRecord (e.g. confluent_kafka). 

For my pipeline test, I added a Map function 'passthrough' (it just logs the message and yields the element) in my Python pipeline after reading from Kafka: 

`
messages = ( p | 'Read From Kafka' >> ReadFromKafka(
                                          consumer_config={ ......
                                          },
                                          ......
                                          expansion_service='localhost:8097'))`
`
messages | 'Passthrough' >> beam.Map(passthrough)
`

As of now, this is the error I'm getting while the portable runner is reading from Kafka topics

`
[grpc-default-executor-1] ERROR org.apache.beam.fn.harness.control.BeamFnControlClient - Exception while trying to handle InstructionRequest 5 java.lang.IllegalArgumentException: Expected DoFn to be FunctionSpec with URN urn:beam:dofn:javasdk:0.1, but URN was beam:dofn:pickled_python_info:v1
        at org.apache.beam.vendor.guava.v20_0.com.google.common.base.Preconditions.checkArgument(Preconditions.java:416)
        at org.apache.beam.runners.core.construction.ParDoTranslation.doFnWithExecutionInformationFromProto(ParDoTranslation.java:572)
        at org.apache.beam.runners.core.construction.ParDoTranslation.getDoFn(ParDoTranslation.java:282)
        at org.apache.beam.fn.harness.DoFnPTransformRunnerFactory$Context.<init>(DoFnPTransformRunnerFactory.java:197)
        at org.apache.beam.fn.harness.DoFnPTransformRunnerFactory.createRunnerForPTransform(DoFnPTransformRunnerFactory.java:96)
        at org.apache.beam.fn.harness.DoFnPTransformRunnerFactory.createRunnerForPTransform(DoFnPTransformRunnerFactory.java:64)
        at org.apache.beam.fn.harness.control.ProcessBundleHandler.createRunnerAndConsumersForPTransformRecursively(ProcessBundleHandler.java:194)
        at org.apache.beam.fn.harness.control.ProcessBundleHandler.createRunnerAndConsumersForPTransformRecursively(ProcessBundleHandler.java:163)
        at org.apache.beam.fn.harness.control.ProcessBundleHandler.processBundle(ProcessBundleHandler.java:290)
        at org.apache.beam.fn.harness.control.BeamFnControlClient.delegateOnInstructionRequestType(BeamFnControlClient.java:160)
        at org.apache.beam.fn.harness.control.BeamFnControlClient.lambda$processInstructionRequests$0(BeamFnControlClient.java:144)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)`

mxm:
This error occurs when the output coder can not be inferred correctly from the Python function. The following should work:

```python
messages | 'Passthrough' >> beam.Map(passthrough).with_output_types(KV[bytes, bytes])
```

manuelaguilar:
@mxm I tried the hinted type but I'm still getting the same error. 'passthrough' is a local Python function:

`
def passthrough(element):
    logging.info("Received element {}".format(element))
    yield element
`

And the error:

`
[grpc-default-executor-0] ERROR org.apache.beam.fn.harness.control.BeamFnControlClient - Exception while trying to handle InstructionRequest 5 java.lang.IllegalArgumentException: Expected DoFn to be FunctionSpec with URN urn:beam:dofn:ja
vasdk:0.1, but URN was beam:dofn:pickled_python_info:v1
`

Does portability allow having both external and local transforms in a pipeline in different languages? From what I read, the pipeline is submitted to the runner for execution, how does the runner handle cross language pipelines?

In this case, the execution of the pipeline occurs in the Flink server (Java), and it's not expecting a Python urn.

manuelaguilar:
@mxm I logged the pipeline request (./sdks/java/harness/src/main/java/org/apache/beam/fn/harness/control/BeamFnControlClient.java): from there we can see the 'Passthrough" transform payload has the python urn embedded in it:

`
[grpc-default-executor-0] INFO org.apache.beam.fn.harness.control.BeamFnControlClient - Sending instruction request instruction_id: "3"
register {
  process_bundle_descriptor {
    id: "5"
    transforms {
      key: "ref_AppliedPTransform_Passthrough_4"
      value {
        spec {
          urn: "urn:beam:transform:pardo:v1"
          payload: "\n\333\a\n\261\a\n beam:dofn:pickled_python_info:v1\032\214\aeNp9kmtv0zAUhpOurCXduG2MywYb16WwJu3uwBiFjTE0LUgBwydkOYmbZEtin8RBqyASCHXwhR/LP8CtJo1KCNmyjo/Pa9nPeb+WdJdw4gYUO5TEhkhJknVYGmeGy1KqbZEoIk5EP6SEc5pus51EA6X+DdQCSjqq8JS5NMtgxPXCKDJwf9Wwm1IiKO7kiStCJhVlfeg8YsTDosupBmdQdYt59J3cwGgPKjZU9T1VjtL+RBuN/1SUL4ryXVV8VXkLZ60eaHWkHkGtB2PolxmwmJrEz+Nms9VOaWY4id8xIuaSyPRdjjM3DbnIzCh0TN4VAUsWjTUzCwVtyE8fEp9m5l/fbywarSWj2eBdWdaIwiQ/ahytr+LVZYP6/t+V5ikosw/K4F0YR9WNiMSORzbh3P7vMlLgPCp1ErjQg4t1AZdsmHAxdvIwEmGCseZTQYRINZgcohPGnKUCx8zLI0noMpr+T4dgqgdXbLiKqljKXYExXDuG6zZMoxonWSaClOV+ADPWvzrkUrmBG0FZD06wj+y7bTQrFOVAVTxVOSid8JdTldF71VNkI25aaMKmLg0/UW+ORjSmiZj7XMDsMcyhSsR8P0x8uIXKYdJhcBuN9h9MBNz5AXdR5UQA93pwH80fks4hwYJhJ/Qhp2kX514He9TLucFDTmUbBnjng5m9SVRW1CkV9LrkWbfhwRBPjGMyCIIZyyrgIaoNTIaDMBEZLAwbXR4M8oZHJUciWJppr9/0fbjbT2vQkC43CjALaOqoFiY8F4PrMmhZaIzl4jSxaOXHsORkApZtWOnBqg1rPVgv4JEeLAT9ix4X8KSADT1oWcGg+qkjX0dSP+PUxdIjm8FKLuCZDe0TxfMCXhSwdarYdtBIv/Jl0M6dj7BTwCvLErBr/AG4tUWj\022%ref_Environment_default_environment_1"
        }
        inputs {
          key: "0"
          value: "ref_PCollection_PCollection_1"
        }
        outputs {
          key: "None"
          value: "ref_PCollection_PCollection_2"
        }
        unique_name: "Passthrough"
      }
    }
`

mxm:
>Does portability allow having both external and local transforms in a pipeline in different languages? 

Yes it does. You can mix Python, Java code, and Go code in the same pipeline. Each have their own environment via docker containers. That's also why we are limited to standard coders to transfer data between the different environments. 

It is a bit tricky though for sources where we actually mix three different environments a) Java via docker b) Python via docker c) native Java via the Flink runner to run the old style unbounded sources, i.e. Kafka IO. The latter code path will be removed once we have an adapter to run old style sources via SplittableDoFn. If that does not make sense to you, that's alright :) 

Not sure what the issue is here. Earlier I ran a semantically identical pipeline with a ReadFromKafka and a pass-through Python transform, followed by a WriteToKafka. Will take a closer look when I get a chance (currently traveling).

manuelaguilar:
@mxm The issue was related to invalid docker image names. 

My first issue (invalid docker image name) was solved by creating a plain Unix account and building all the images there. 

A small issue I noticed after was when deploying the portable runner in Linux for a pipeline that runs multiple languages (Java + Python).  The flink server does not have the USER environment variable defined, so it can't find the right payload image to use on the second transform and probably onwards. I hardcoded the USER environment variable in the flink server Dockerfile and it now finds the right image for the right payload.

The patch works and I'm able to consume Kafka payload using the portable runner. 

Another thing I noticed is that although the deserializer I specified in the ReadFromKafka transform is for ByteArray (both for key and value), the Python pipeline is consuming it as a string without further conversion.

Would it make sense to have the KafkaRecord metadata as an option in the portable pipeline? (to be able to access partitions, offsets, timestamps, and other KafkaRecord properties).

I'll continue running some performance tests to see how Flink scales up with a few million records in the portability framework.

mxm:
Great to hear that you resolved the issue. The user variable could probably be inferred from the user's home directory.

>Another thing I noticed is that although the deserializer I specified in the ReadFromKafka transform is for ByteArray (both for key and value), the Python pipeline is consuming it as a string without further conversion.

The string consumption will only work for Python 2 though where bytes and strings are treated equally.

>Would it make sense to have the KafkaRecord metadata as an option in the portable pipeline? (to be able to access partitions, offsets, timestamps, and other KafkaRecord properties).

That would make perfect sense. However, we wanted to avoid having to include KafkaRecord as a type across SDKs. We figured it might be sufficient to have only the raw data available. As you saw, it is not yet possible to completely use the KafkaIO consumer without the KafkaRecord cross-language coder. Hopefully, we will have fixed that soon and introduce a structure for preserving KafkaRecord meta data across SDKs.

Curious to learn about your performance tests.

manuelaguilar:
@mxm It seems performance depends mostly on the sink. I've been able to get 3000 msg/sec with a file sink (which doesn't complete the final write when I cancel the job via Flink), and 2300 msg/sec with a recently patched version of the mongodb sink. The pipeline had a map transform to get the value from every KV element. This was done on a quad core Intel(R) Xeon(R) CPU E5-2695 v4 @ 2.10GHz virtual machine with 8GB memory using a standalone Flink docker image as runner endpoint.
I have observed our Google Dataflow n1-standard-1 instance (1vCPU, 3.75GB memory) can process 2000 msg/sec per worker (using the respective Dataflow runner). This was using a Java dataflow template.

PR #8252, opened 2019-04-08
[BEAM-6934] Fixing timer firing timing issue

pabloem:
What is odd about this problem is that:
- We would expect the timer set to `100` to fire when the watermark advances to 100 - but instead, it fires together with the timer set to `1000` when the watermark advances to infinity.

On the other hand, if we set `advance_watermark_to(101)`, the test behaves properly without any need for modifications.

BUT, what's even more odd is that when doing the actual check in trigger.py as to whether the timer should be fired, we do check for `<=`. Strange.:
https://github.com/apache/beam/blob/master/sdks/python/apache_beam/transforms/trigger.py#L1273

Nonetheless, this change does fix the new test (which was previously failing) - and I believe it does correct the behavior. So. Yes.

I'll keep hunting for bugs.

r: @aaltay 
cc: @tweise

### DIFF
--- a/sdks/python/apache_beam/runners/common.py
+++ b/sdks/python/apache_beam/runners/common.py
@@ -789,7 +789,7 @@ class DoFnRunner(Receiver):
 
   def _reraise_augmented(self, exn):
     if getattr(exn, '_tagged_with_step', False) or not self.step_name:
-      raise exn
+      raise
     step_annotation = " [while running '%s']" % self.step_name
     # To emulate exception chaining (not available in Python 2).
     try:
--- a/sdks/python/apache_beam/runners/common.py
+++ b/sdks/python/apache_beam/runners/common.py
@@ -789,7 +789,7 @@ class DoFnRunner(Receiver):
 
   def _reraise_augmented(self, exn):
     if getattr(exn, '_tagged_with_step', False) or not self.step_name:
-      raise
+      raise exn
     step_annotation = " [while running '%s']" % self.step_name
     # To emulate exception chaining (not available in Python 2).
     try:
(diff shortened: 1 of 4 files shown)


sdks/python/apache_beam/runners/common.py
```diff
@@ -789,7 +789,7 @@ def finalize(self):
 
   def _reraise_augmented(self, exn):
     if getattr(exn, '_tagged_with_step', False) or not self.step_name:
-      raise
+      raise exn
```
aaltay: Would not this override the stack trace? I believe it should be either raise or raise exn[0], exn[1], exn[2] format.

sdks/python/apache_beam/transforms/userstate_test.py
```diff
@@ -366,6 +367,47 @@ def expiry_callback(self, buffer=DoFn.StateParam(BUFFER_STATE),
...
+                     .advance_watermark_to(0)
+                     .add_elements([('key', 'value')])
+                     .advance_watermark_to(100))
+
+      _ = (p
+           | test_stream
+           | beam.ParDo(BagStateClearingStatefulDoFn())
+           | beam.ParDo(self.record_dofn()))
+
+    self.assertEqual(
```
aaltay: Is it possible to check whether `emit_values` fired or not at all?

sdks/python/apache_beam/runners/common.py
pabloem: Ooops. Good catch.

sdks/python/apache_beam/transforms/userstate_test.py
pabloem: I've added a check for that. Thanks!

PR #8271, opened 2019-04-10
Refactoring code from direct runner, and adding unit test for processing time timers.

pabloem:
I went through this code by trying to study the state and timers implementation in the direct runner. I added a bit of documentation, a unit test, and refactored a couple places for clarity.

r: @aaltay 

---
Thoughts for myself:

I realized that currently, it is not possible to create a sequence generator in Python, because there is no way to have access to the watermark (other than using a test stream). (How about watermark holds? - i dont know enough about any of this basically : ))

If we use a sequence generator using a `beam.Create` transform, the watermark immediately goes to infinity, and the timers are not processed, because the pipeline ends.

Anyway.

### DIFF
--- a/sdks/python/apache_beam/runners/common.py
+++ b/sdks/python/apache_beam/runners/common.py
@@ -547,7 +547,7 @@ class PerWindowInvoker(DoFnInvoker):
       try:
         self.current_windowed_value = windowed_value
         self.restriction_tracker = restriction_tracker
-        return self._invoke_per_window(
+        return self._invoke_process_per_window(
             windowed_value, additional_args, additional_kwargs,
             output_processor)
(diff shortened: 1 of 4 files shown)


aaltay:
> I realized that currently, it is not possible to create a sequence generator in Python,
Is not the GenerateRecords you have here is a sequence generator?

aaltay:
> I realized that currently, it is not possible to create a sequence generator in Python,
Is not the GenerateRecords you have here is a sequence generator?

udim:
I believe this PR broke TestDirectRunner-based tests in Python postcommit tests. 
https://issues.apache.org/jira/browse/BEAM-7111

PR #8289, opened 2019-04-12
[BEAM-7064] Reject BigQuery data with sub-millisecond precision, instead of losing data

kennknowles:
Previously, reading from BQ and converting to Beam rows - either via SQL or just using schema-driven transforms - would cause timestamps to be truncated at milliseconds. We don't support higher precision, but this rejects pipelines that would otherwise lose data.

------------------------

--- | --- | --- | --- | ---


kennknowles:
R: @amaliujia too

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/AvroUtils.java
```diff
@@ -73,8 +69,19 @@ public static Object convertAvroFormat(Field beamField, Object value) {
       default:
         throw new RuntimeException("Does not support converting unknown type value");
     }
+  }
 
-    return ret;
+  private static ReadableInstant safeToMillis(Object value) {
+    long subMilliPrecision = ((long) value) % 1000;
```
amaliujia: How do we know that that the value returned is not millis but micro?

amaliujia:
Run SQL PostCommit

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/AvroUtils.java
kennknowles: It is because this `AvroUtils` is part of the BigQuery connector, not general. And we know BigQuery is always micros. Is that your question?

amaliujia: Yes. I was asking because input is an object.

amaliujia: I was also wondering where long is generated. If BQ timestamp converts to long which contains micros, will that lose precision?

I did some math:

BigQuery doc says TIMESTAMP type range is `0001-01-01 00:00:00 to 9999-12-31 23:59:59.999999 UTC`

so it should ranges from `-6.127488×10¹⁶ to 2.4976512×10¹⁷` in micros, calculated by `((0001-01-01 / 9999-12-31) -/+ 1970) * 12 * 30 * 24 *3600 * 1000 * 1000`, which falls into INT64's range. 

So no precision loss.

amaliujia:
Run Java PostCommit

amaliujia:
LGTM

kennknowles:
Not user friendly at all! An attempt to limit data loss. The real solution will be nano precision, a logical type at the Beam schema level. We will have to reimplement date time functions.

reuvenlax:
So I'm a bit worried about this change: there may be users who are ok with losing timestamp precision (e.g. for log events, they may perceive losing nano precision on event timestamps as a noop). Maybe for now we should allow them to override this behavior with a flag?

kennknowles:
I'm cool with an opt-in flag.

amaliujia:
With a flag, which behavior should be the default?

Aggressive plan is to fail users pipelines so make exception approach as the default. 

Or we can be more cautious to allow truncating timestamp by default.

It's more like a short term pain (possibly breaking change to some users) vs long term pain (transparent precision truncation that is hard to detect for both existing and new users).

kennknowles:
I feel strongly that we should not truncate by default.

kennknowles:
Updated to have a flag that can be set in table properties. I thought this code was shared with BigQuery schema conversions, but it does not appear to be.

kennknowles:
Because of how the TableRow to Beam Row conversion is currently expressed a bit differently (as a map of typename to parser), it will be a moderate refactor to share these conversion options. I don't want to confuse this review with that, but will start work on it since we don't want to lose data there either.

kennknowles:
run java precommit

kennknowles:
run sql postcommit

reuvenlax:
+1 the flag should opt in not opt out.

kennknowles:
run java precommit

PR #8300, opened 2019-04-13
[BEAM-7035] Compatible wire representation for timers in Python SDK

tweise:
Fix timer encoding on Python SDK side, support cancellation of timers by timer ID in the Flink runner and add VR test coverage for timer clear/cancel.

Adding @lukecwik who will be familiar with the timestamp encoding workarounds :)

Also see #8253

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/DoFnOperator.java
+++ b/runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/DoFnOperator.java
@@ -1014,7 +1014,7 @@ public class DoFnOperator<InputT, OutputT> extends AbstractStreamOperator<Window
     @Override
     public void setTimer(TimerData timer) {
       try {
-        String contextTimerId = getContextTimerId(timer);
+        String contextTimerId = getContextTimerId(timer.getTimerId(), timer.getNamespace());
         // Only one timer can exist at a time for a given timer id and context.
         // If a timer gets set twice in the same context, the second must
         // override the first. Thus, we must cancel any pending timers
@@ -1051,15 +1051,15 @@ public class DoFnOperator<InputT, OutputT> extends AbstractStreamOperator<Window
 
     void cleanupPendingTimer(TimerData timer) {
       try {
-        pendingTimersById.remove(getContextTimerId(timer));
+        pendingTimersById.remove(getContextTimerId(timer.getTimerId(), timer.getNamespace()));
       } catch (Exception e) {
         throw new RuntimeException("Failed to cleanup state with pending timers", e);
       }
     }
 
     /** Unique contextual id of a timer. Used to look up any existing timers in a context. */
-    private String getContextTimerId(TimerData timer) {
-      return timer.getTimerId() + timer.getNamespace().stringKey();
+    private String getContextTimerId(String timerId, StateNamespace namespace) {
+      return timerId + namespace.stringKey();
     }
 
     /** @deprecated use {@link #deleteTimer(StateNamespace, String, TimeDomain)}. */
@@ -1071,7 +1071,11 @@ public class DoFnOperator<InputT, OutputT> extends AbstractStreamOperator<Window
 
     @Override
     public void deleteTimer(StateNamespace namespace, String timerId, TimeDomain timeDomain) {
-      throw new UnsupportedOperationException("Canceling of a timer by ID is not yet supported.");
+      try {
+        cancelPendingTimerById(getContextTimerId(timerId, namespace));
+      } catch (Exception e) {
+        throw new RuntimeException("Failed to cancel timer", e);
+      }
     }
 
     /** @deprecated use {@link #deleteTimer(StateNamespace, String, TimeDomain)}. */
--- a/runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
+++ b/runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
@@ -387,7 +387,6 @@ public class ExecutableStageDoFnOperator<InputT, OutputT> extends DoFnOperator<I
         stateBackendLock.lock();
         getKeyedStateBackend().setCurrentKey(encodedKey);
         if (timerData.getTimestamp().isAfter(BoundedWindow.TIMESTAMP_MAX_VALUE)) {
-          // TODO: "Canceling of a timer by ID is not yet supported."
           timerInternals.deleteTimer(
               timerData.getNamespace(), timerData.getTimerId(), timerData.getDomain());
         } else {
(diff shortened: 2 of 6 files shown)


tweise:
Run Java PreCommit

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
```diff
@@ -378,14 +379,20 @@ public ByteBuffer getCurrentKey() {
...
+      LOG.debug("Setting timer: {} {}", timerElement, timerData);
       // KvToByteBufferKeySelector returns the key encoded
       ByteBuffer encodedKey = (ByteBuffer) keySelector.getKey(timerElement);
       // We have to synchronize to ensure the state backend is not concurrently accessed by the
       // state requests
       try {
         stateBackendLock.lock();
         getKeyedStateBackend().setCurrentKey(encodedKey);
-        timerInternals.setTimer(timerData);
+        if (timerData.getTimestamp().isAfter(GlobalWindow.TIMESTAMP_MAX_VALUE)) {
```
mxm: Should this be `MAX_TIMESTAMP` (`Long.MAX_VALUE`)? There can be windows between `GlobalWindow.MAX_TIMEMSTAMP` and `MAX_TIMESTAMP` because the former is a standard day before the latter. Actually, `MAX_TIMESTAMP` would be good as a value because triggers using it as a timestamp can never fire because the Watermark cannot go past `MAX_TIMESTAMP`.

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/DoFnOperator.java
```diff
@@ -1071,7 +1071,11 @@ public void deleteTimer(StateNamespace namespace, String timerId) {
 
     @Override
     public void deleteTimer(StateNamespace namespace, String timerId, TimeDomain timeDomain) {
-      throw new UnsupportedOperationException("Canceling of a timer by ID is not yet supported.");
+      try {
+        cancelPendingTimerById(getContextTimerId(timerId, namespace));
```
mxm: This is not enough to clean the timer. The method call will only evict the timer meta information used to delete timers by id (Flink does not allow to delete timers by id, you always need to supply the original timestamp of the the timer). See the deleteTimer method below for the additional delete step on Flink's timer service.

sdks/python/apache_beam/coders/coder_impl.py
```diff
@@ -491,13 +491,29 @@ def estimate_size(self, value, nested=False):
...
+
+  TODO: SDK agnostic encoding
+  For interoperability with Java SDK, encoding needs to match
+  that of the Java SDK InstantCoder.
+  https://github.com/apache/beam/blob/f5029b4f0dfff404310b2ef55e2632bbacc7b04f/sdks/java/core/src/main/java/org/apache/beam/sdk/coders/InstantCoder.java#L79
+  """
 
   def encode_to_stream(self, value, out, nested):
-    out.write_bigendian_int64(value.micros)
+    millis = value.micros // 1000
```
mxm: Does that mean that timestamps were incorrect before?

sdks/python/apache_beam/runners/worker/bundle_processor.py
```diff
@@ -381,8 +381,12 @@ def set(self, ts):
         windowed_value.WindowedValue(
             (self._key, dict(timestamp=ts)), ts, (self._window,)))
 
-  def clear(self, timestamp):
-    self._receiver.receive((self._key, dict(clear=True)))
+  def clear(self):
+    dummy_millis = int(common_urns.constants.MAX_TIMESTAMP_MILLIS.constant) + 1
```
mxm: Did you mean to use GlobalWindow.MAX_TIMESTAMP here?

I think this should be just MAX_TIMESTAMP (see above).

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/DoFnOperator.java
mxm: Have to correct myself, I thought you were calling `cleanupPendingTimer` here, but `cancelPendingTimerById` does delete the timer correctly. Sorry about that.

sdks/python/apache_beam/coders/coder_impl.py
tweise: Yup, and timers only working because they would fire on the final watermark.

sdks/python/apache_beam/runners/worker/bundle_processor.py
tweise: No, it refers to https://github.com/apache/beam/blob/7d6b7b89f5832d5d375059f9c1980181b443835b/model/pipeline/src/main/proto/beam_runner_api.proto#L40

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
tweise: The constant is actually defined in BoundedWindow https://github.com/apache/beam/blob/a9723d954a92391e9ca79130a199a8eaf38953f2/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/windowing/BoundedWindow.java#L57

I will change this to avoid confusion.

tweise:
Run Python PreCommit

sdks/python/apache_beam/runners/worker/bundle_processor.py
mxm: Wouldn't this cause an overflow then? We cannot go past the maximum timestamp.

tweise: Long.MAX_VALUE is `9223372036854775807`, which is different from `MAX_TIMESTAMP_MILLIS`. The resulting timer timestamp is `Timer{timestamp=294247-01-10T04:00:54.776Z, payload=[B@5aec2a63}`. (The code in this PR is covered by a test, if there was an overflow the test would fail.)

mxm: Ok, max timestamp is then `Long.MAX_VALUE / 1000`. Seems safe to add 1 to it then.

tweise:
@lukecwik can you please clarify what changes you request? I assume you are referring to work that should be done to clean up this timestamp and timer mess in general. That's certainly reasonable outside of this PR, which is fixing a bug that we are blocked on.

lukecwik:
Use standard_coders.yaml to perform the timer test. You can do that by copying over the missing timer definition into the python one or attempt to solve BEAM-7015 as well and migrate to use the standard_coders.yaml found underneath the model/ folder.

It would be nice to discuss the discrepancy on the ML but that shouldn't block this change from going in.

tweise:
@lukecwik makes sense. I will see to make the change to the test and also bring the overall topic to the ML.

pabloem:
Thanks Thomas for fixing this. This LGTM. IIUC, cancelling timers will be done by setting their timestamp beyond the maximum timestamp. Is that correct?

tweise:
> Thanks Thomas for fixing this. This LGTM. IIUC, cancelling timers will be done by setting their timestamp beyond the maximum timestamp. Is that correct?

That's right. There is currently no way to encode "cancel timer", so this is using the timestamp trick.

PR #8301, opened 2019-04-13
[BEAM-7070] JOIN condition should accept field access

amaliujia:
Currently BeamSQL's join rel implementation has limitations:
1. only deal with RexInputRef and direct cast to RexInputRef without checking.
2. only check if join condition is `() [AND () AND ...]`
3. treat non equal operator (e.g. <) as equal operator (e.g. =)

and more

It's a big effort to improve join rel to make it work properly in every cases. In BeamSQL,  there is a also a lack if enough test cases and test utils.

This PR tries to kick off the party and
1. make JOIN condition work on field access. 
2. reject unsupported RexNode by an exception.

This PR is only an attempt to start to generalize join condition.

This PR also partially solves https://jira.apache.org/jira/browse/BEAM-7071.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamJoinRel.java
+++ b/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamJoinRel.java
@@ -21,6 +21,7 @@ import static org.apache.beam.sdk.schemas.Schema.toSchema;
 import static org.apache.beam.sdk.values.PCollection.IsBounded.UNBOUNDED;
 import static org.joda.time.Duration.ZERO;
 
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -259,14 +260,11 @@ public class BeamJoinRel extends Join implements BeamRelNode {
       // the name of the join field is not important
       Schema extractKeySchemaLeft =
           pairs.stream()
-              .map(pair -> BeamJoinRel.getFieldBasedOnRexNode(leftSchema, pair.getKey(), 0))
+              .map(pair -> getFieldBasedOnRexNode(leftSchema, pair.getKey(), 0))
               .collect(toSchema());
       Schema extractKeySchemaRight =
           pairs.stream()
-              .map(
-                  pair ->
-                      BeamJoinRel.getFieldBasedOnRexNode(
-                          rightSchema, pair.getValue(), leftRowColumnCount))
+              .map(pair -> getFieldBasedOnRexNode(rightSchema, pair.getValue(), leftRowColumnCount))
               .collect(toSchema());
 
       SchemaCoder<Row> extractKeyRowCoder = SchemaCoder.of(extractKeySchemaLeft);
@@ -505,14 +503,41 @@ public class BeamJoinRel extends Join implements BeamRelNode {
       return schema.getField(((RexInputRef) rexNode).getIndex() - leftRowColumnCount);
     } else if (rexNode instanceof RexFieldAccess) {
       // need to extract field of Struct/Row.
-      RexFieldAccess fieldAccess = (RexFieldAccess) rexNode;
-      RexInputRef inputRef = (RexInputRef) fieldAccess.getReferenceExpr();
-      Field structField = schema.getField(inputRef.getIndex() - leftRowColumnCount);
-      // Should use RexFieldAccess to get field of Struct field.
-      return structField.getType().getRowSchema().getField(fieldAccess.getField().getIndex());
+      getFieldBasedOnRexFieldAccess(schema, (RexFieldAccess) rexNode, leftRowColumnCount);
     }
 
     throw new UnsupportedOperationException("Does not support " + rexNode.getType() + " in JOIN.");
+  }
+
+  private static Field getFieldBasedOnRexFieldAccess(
+      Schema schema, RexFieldAccess rexFieldAccess, int leftRowColumnCount) {
+    ArrayDeque<RexFieldAccess> fieldAccessStack = new ArrayDeque<>();
+    fieldAccessStack.push(rexFieldAccess);
+
+    RexFieldAccess curr = rexFieldAccess;
+    while (curr.getReferenceExpr() instanceof RexFieldAccess) {
+      curr = (RexFieldAccess) curr.getReferenceExpr();
+      fieldAccessStack.push(curr);
+    }
+
+    // curr.getReferenceExpr() is not a RexFieldAccess. Check if it is RexInputRef, which is only
+    // allowed RexNode type in RexFieldAccess.
+    if (!(curr.getReferenceExpr() instanceof RexInputRef)) {
+      throw new UnsupportedOperationException(
+          "Does not support " + curr.getReferenceExpr().getType() + " in JOIN.");
+    }
+
+    // curr.getReferenceExpr() is a RexInputRef.
+    RexInputRef inputRef = (RexInputRef) curr.getReferenceExpr();
+    Field curField = schema.getField(inputRef.getIndex() - leftRowColumnCount);
+
+    // pop RexFieldAccess from stack one by one to know the final field type.
+    while (fieldAccessStack.size() > 0) {
+      curr = fieldAccessStack.pop();
+      curField = curField.getType().getRowSchema().getField(curr.getField().getIndex());
+    }
+
+    return curField;
   }
 
   private List<Pair<RexNode, RexNode>> extractJoinRexNodes() {
@@ -557,7 +582,7 @@ public class BeamJoinRel extends Join implements BeamRelNode {
     if (rexNode instanceof RexInputRef) {
       return ((RexInputRef) rexNode).getIndex();
     } else if (rexNode instanceof RexFieldAccess) {
-      return ((RexInputRef) ((RexFieldAccess) rexNode).getReferenceExpr()).getIndex();
+      return getColumnIndex(((RexFieldAccess) rexNode).getReferenceExpr());
     }
 
     throw new UnsupportedOperationException("Cannot get column index from " + rexNode.getType());
--- a/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamJoinRel.java
+++ b/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamJoinRel.java
@@ -33,6 +33,7 @@ import org.apache.beam.sdk.extensions.sql.BeamSqlTable;
 import org.apache.beam.sdk.extensions.sql.impl.transform.BeamJoinTransforms;
 import org.apache.beam.sdk.extensions.sql.impl.utils.CalciteUtils;
 import org.apache.beam.sdk.schemas.Schema;
+import org.apache.beam.sdk.schemas.Schema.Field;
 import org.apache.beam.sdk.schemas.SchemaCoder;
 import org.apache.beam.sdk.transforms.MapElements;
 import org.apache.beam.sdk.transforms.PTransform;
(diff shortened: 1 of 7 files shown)


amaliujia:
R: @kennknowles

amaliujia:
retest this please

akedin:
One thing to note is that Calcite doesn't seem to recognize EquiJoins if field access is in use: https://issues.apache.org/jira/browse/CALCITE-2898

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamJoinRel.java
```diff
@@ -248,16 +250,24 @@ private boolean isSideInputLookupJoin() {
...
-      List<Pair<Integer, Integer>> pairs =
-          extractJoinColumns(leftRelNode.getRowType().getFieldCount());
+      List<Pair<RexNode, RexNode>> pairs = extractJoinRexNodes();
 
       // build the extract key type
       // the name of the join field is not important
       Schema extractKeySchemaLeft =
-          pairs.stream().map(pair -> leftSchema.getField(pair.getKey())).collect(toSchema());
+          pairs.stream()
+              .map(pair -> BeamJoinRel.getFieldBasedOnRexNode(leftSchema, pair.getKey(), 0))
```
akedin: you don't need the `BeamJoinRel` to reference the static method in this class

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamJoinRel.java
```diff
@@ -488,22 +499,38 @@ private Schema buildNullSchema(Schema schema) {
...
 
-  private List<Pair<Integer, Integer>> extractJoinColumns(int leftRowColumnCount) {
+  private static Field getFieldBasedOnRexNode(
+      Schema schema, RexNode rexNode, int leftRowColumnCount) {
+    if (rexNode instanceof RexInputRef) {
+      return schema.getField(((RexInputRef) rexNode).getIndex() - leftRowColumnCount);
+    } else if (rexNode instanceof RexFieldAccess) {
+      // need to extract field of Struct/Row.
+      RexFieldAccess fieldAccess = (RexFieldAccess) rexNode;
+      RexInputRef inputRef = (RexInputRef) fieldAccess.getReferenceExpr();
```
akedin: Is the nested field access possible here? If it is, would the `fieldAccess.getReferenceExpr()` return `RexFieldAccess`?

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/transform/BeamJoinTransforms.java
```diff
@@ -44,26 +46,48 @@
...
       return KV.of(row, input);
     }
 
+    @SuppressWarnings("unused")
     private Schema.Field toField(Schema schema, Integer fieldIndex) {
       Schema.Field original = schema.getField(fieldIndex);
       return original.withName("c" + fieldIndex);
     }
+
+    private Object getValue(
```
akedin: This supports only one level nesting. How hard would it be to implement the linked list of the field indices? E.g. `serializableRexNode.isNested()`, `serializableRexNode.getNestedSerializableRexNode()`. This way the logic can probably be generalized a bit. And you probably won't need the class hierarchy, it would be a linked list with an int payload, and you stop when you reach the end of the list.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/utils/SerializableRexNode.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+package org.apache.beam.sdk.extensions.sql.impl.utils;
+
+import java.io.Serializable;
+import org.apache.calcite.rex.RexFieldAccess;
+import org.apache.calcite.rex.RexInputRef;
+import org.apache.calcite.rex.RexNode;
+
+/** SerializableRexNode. */
+public abstract class SerializableRexNode implements Serializable {
```
akedin: It is confusing to call these `RexNodes`, rename to something like `FieldIndex`?

amaliujia:
Reading that relevant mail and seems to me that Calcite does not recognize equi-join if field acess is in use, when using Enumerablexxx. 

Are we using anyone Enumeralbexxx of JOIN in BeamSQL? (thining we don't use that).

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/utils/SerializableRexNode.java
```diff
@@ -0,0 +1,56 @@
...
+  public abstract int getIndex();
+
+  public SerializableRexNode getReferenceNode() {
```
akedin: Also unclear what reference node is, I would probably call it something like `getNestedFieldIndex()`

sdks/java/testing/nexmark/src/test/java/org/apache/beam/sdk/nexmark/queries/QueryTest.java
```diff
@@ -188,12 +188,14 @@ public void query5MatchesModelStreaming() {
 
   @Test
   @Category(NeedsRunner.class)
+  @Ignore("https://jira.apache.org/jira/browse/BEAM-7072")
```
akedin: Can you add some description to jira?

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamJoinRel.java
amaliujia: Done

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamJoinRel.java
amaliujia: You are absolutely right. `row.row.row.field` should have `ref.fc.fc`. Improved to support longer field access chain.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/transform/BeamJoinTransforms.java
amaliujia: That's a good idea. 
InputRef -> Index
FieldAccess -> List of Index

Changed.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/utils/SerializableRexNode.java
amaliujia: Totally changed here so there is no more `getReferenceNode`

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/utils/SerializableRexNode.java
amaliujia: The following work is to add function call support in JOIN condition, e.g. CAST(col) = field access. How to name that then?

I named these new classes to be similar with `RexNode` to preserve some meaning from `RexNode`. Say when see `SerializableRexCall`, it should remind that within this class, there should be a calcite operator with a list of `SerializableRexNode` as operands.

amaliujia:
Run Java PostCommit

amaliujia:
Run Java PostCommit

amaliujia:
Run Java PostCommit

amaliujia:
Run Java PostCommit

amaliujia:
All green. PLAT

amaliujia:
Ping

sdks/java/testing/nexmark/src/test/java/org/apache/beam/sdk/nexmark/queries/QueryTest.java
akedin: Did it work before without `@Ignore`?

amaliujia: There was a bug (which is also fixed by this PR by throwing exception), if there is a `JOIN a > b AND c < d`, BeamSQL executes it as `JOIN a = b AND c = d`. From code you can see if there is a [AND](https://github.com/apache/beam/blob/master/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamJoinRel.java#L499), Beam code goes to extract JOIN column without checking what's [condition](https://github.com/apache/beam/blob/master/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamJoinRel.java#L518).

And these ignored queries are using non equi-join, which BeamSQL has never supported.

amaliujia: So. Join Rel implementation is a hack and needs lots of improvement.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/utils/SerializableRexNode.java
akedin: I am not sure, I think we will need to think about it a bit more. At this point we're starting to replicate Calcite's relational model representation just because we need to serialize some parts of it.

Also, at the moment this code is used only in `ExtractJoinFields` which doesn't imply function calls. Maybe it's just naming.

Can you share the design for function calls? My concern is that we will need to re-implement that part as well, even though now we have codegen that does similar thing for us in calc rel. Can we re-use that part somehow? E.g. insert a projection which will be backed up by codegen for field accesses and function calls?

sdks/java/testing/nexmark/src/test/java/org/apache/beam/sdk/nexmark/queries/QueryTest.java
akedin: Whoa that was bad

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/utils/SerializableRexNode.java
amaliujia: I can see your concern.

Ok, I will directly work on https://jira.apache.org/jira/browse/BEAM-6112 then.

By doing so, we will have a detailed design/implementation on how to support FunctionCall, ColumnRef and FieldAccess together in Beam.

That should address your concern.

sdks/java/testing/nexmark/src/test/java/org/apache/beam/sdk/nexmark/queries/QueryTest.java
kennknowles: Yikes. One thought that we can do is if there is any equi-join then we can do it first, and then do the rest as a filter operation after that. We could do this in `toPTransform` or in rewrite rules. It might be hard in rewrite rules because the transformation is not an "optimization" unless we leave it at infinite cost until it is rewritten to what we can execute. I'm not sure how easy that will be to express.

Are there tests that unimplementable joins are rejected?

amaliujia: I see. I will explore a general way to supports these: equi-join, AND/OR, function call/operators, Column Ref and FieldAccess. 


There are no tests to test rejections of unsupported JOIN. I will add some.

amaliujia:
Thanks!

As there are many future works so I come up with this doc to track those:
https://docs.google.com/document/d/1vDiE4HR5ZdbZypIf1vzyFy9yKmAMWBu6BzBKfP7JhPc/edit?usp=sharing

kennknowles:
Did you share the doc on the dev list?

amaliujia:
Yes I did.

PR #8304, opened 2019-04-15
[BEAM-7075] Create Redis embedded server on @BeforeClass and simplify tests

iemejia:
I have also refactored most unit tests to make them more unit test like with a considerable performance gain. Also relies only on IPv4 to speed tests still more.

R: @aromanenko-dev


iemejia:

iemejia:
Thanks JB, Clean up is not needed if we take care of using a different key or key prefix in each test, this PR does that. We could make it more robust by cleaning the data on `@After` with a minor performance cost in the future.

PR #8317, opened 2019-04-15
[BEAM-2939] Add a new sdf E2E test without defer_remainder

boyuanzz:
The new test is partially taken from test_sdf. The difference is, test_sdf produces defer_remainder in DoFn.process(), verse the new test only yield elements within the range.
R: @lukecwik 
CC: @robertwb

### DIFF
--- a/sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
+++ b/sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
@@ -364,23 +364,25 @@ class FnApiRunnerTest(unittest.TestCase):
 
   def test_sdf(self):
 
+    class ExpandingStringsDoFn(beam.DoFn):
+      def process(self, element, restriction_tracker=ExpandStringsProvider()):
+        assert isinstance(
+            restriction_tracker,
+            restriction_trackers.OffsetRestrictionTracker), restriction_tracker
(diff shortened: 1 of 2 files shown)


boyuanzz:
Run RAT PreCommit

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
```diff
@@ -411,6 +395,24 @@ def process(self, element, restriction_tracker=ExpandStringsProvider()):
       self.assertEqual(1, len(counters))
       self.assertEqual(counters[0].committed, len(''.join(data)))
 
+  def test_sdf_expansion_without_defer_remainder(self):
```
lukecwik: rename to test_sdf

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
```diff
@@ -366,22 +366,6 @@ def test_sdf(self):
```
lukecwik: rename test_sdf to test_sdf_with_sdf_initiated_checkpointing

might also want to reorder this test so that it appears after the other one.

lukecwik:
The Flink test seems to be legitimately failing

PR #8322, opened 2019-04-16
[BEAM-7029] Add KafkaIO.Write as an external transform

mxm:
This adds KafkaIO.Write as an external transform and includes a Python
wrapper (WriteToKafka) for convenience.

CC @chamikaramj @ihji @robertwb @tweise 

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/container/Dockerfile
+++ b/sdks/java/container/Dockerfile
@@ -23,6 +23,11 @@ ADD target/slf4j-api.jar /opt/apache/beam/jars/
 ADD target/slf4j-jdk14.jar /opt/apache/beam/jars/
 ADD target/beam-sdks-java-harness.jar /opt/apache/beam/jars/
 
+# Required to run cross-language pipelines with KafkaIO
+# TODO May be removed once custom environments are supported
+ADD target/beam-sdks-java-io-kafka.jar /opt/apache/beam/jars/
+ADD target/kafka-clients.jar /opt/apache/beam/jars/
+
 ADD target/linux_amd64/boot /opt/apache/beam/
 
 ENTRYPOINT ["/opt/apache/beam/boot"]
--- a/sdks/java/container/Dockerfile-java11
+++ b/sdks/java/container/Dockerfile-java11
@@ -23,6 +23,11 @@ ADD target/slf4j-api.jar /opt/apache/beam/jars/
 ADD target/slf4j-jdk14.jar /opt/apache/beam/jars/
 ADD target/beam-sdks-java-harness.jar /opt/apache/beam/jars/
 
+# Required to run cross-language pipelines with KafkaIO
+# TODO May be removed once custom environments are supported
+ADD target/beam-sdks-java-io-kafka.jar /opt/apache/beam/jars/
+ADD target/kafka-clients.jar /opt/apache/beam/jars/
+
 ADD target/linux_amd64/boot /opt/apache/beam/
 
 ENTRYPOINT ["/opt/apache/beam/boot"]
--- a/sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
+++ b/sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
@@ -458,18 +458,6 @@ public class KafkaIO {
         }
         throw new RuntimeException("Couldn't resolve coder for Deserializer: " + deserializer);
       }
-
-      private static Class resolveClass(String className) {
-        try {
-          return Class.forName(className);
-        } catch (ClassNotFoundException e) {
-          throw new RuntimeException("Could not find deserializer class: " + className);
-        }
-      }
-
-      private static String utf8String(byte[] bytes) {
-        return new String(bytes, Charsets.UTF_8);
-      }
     }
 
     /**
@@ -486,7 +474,7 @@ public class KafkaIO {
         return ImmutableMap.of(URN, AutoValue_KafkaIO_Read.Builder.class);
       }
 
-      /** Parameters class to expose the transform to an external SDK. */
+      /** Parameters class to expose the Read transform to an external SDK. */
       public static class Configuration {
 
         // All byte arrays are UTF-8 encoded strings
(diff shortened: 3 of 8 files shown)


mxm:
Run Python PreCommit

sdks/java/harness/build.gradle
```diff
@@ -55,6 +55,8 @@ dependencies {
   compile library.java.joda_time
   compile library.java.slf4j_api
   compile library.java.vendored_grpc_1_13_1
+  // For executing KafkaIO, e.g. as an external transform
+  compile project(path: ":beam-sdks-java-io-kafka", configuration: "shadow")
```
mxm: Note: `runtime` scope is sufficient.

mxm: We might need a better way to make cross-language artifacts available. This also causes some problems with our build system.

```
FAILURE: Build failed with an exception.

* What went wrong:
Circular dependency between the following tasks:
:beam-runners-direct-java:buildDependents
\--- :beam-sdks-java-io-kafka:buildDependents
     \--- :beam-sdks-java-harness:buildDependents
          +--- :beam-runners-direct-java:buildDependents (*)
          \--- :beam-runners-java-fn-execution:buildDependents
               \--- :beam-runners-direct-java:buildDependents (*)
```

ihji: I got a similar error while working on #8174. I've solved the problem by not packaging a required dependency together but just staging a dependency jar with #8340.

mxm: Thanks for mentioning this. I was also thinking about an option to stage jars from Python. I think it would be nicer if the Jars were automatically provided for the user. The user has to download the KafkaIO jar now and supply its path?

The reason why we can't add KafkaIO to the SDK Harness is that Kafka depends on the direct-runner which has a dependency on the SDK Harness. I'm not sure why the direct-runner has to depend on the SDK harness.

lukecwik: KafkaIO should have at most a test level dependency on the direct runner. Also the direct runner was meant to become the ULR and the interim the SDK harness was added so the direct runner would be able to launch and run Java pipelines "portably" without using a container. Now that the new environments exist, this shouldn't be needed anymore.

ihji: @mxm I think it would be good if expansion service also provides additional dependencies when it responds to expansion requests. In case of Java, not only serialized objects themselves but also enclosing (or referenced) classes are needed for deserialization. For the completeness of serialized codes, they should be bundled together.

lukecwik: @ihji The expansion service should return PTransforms that point to an environment with the needed jars.

mxm: @lukecwik @ihji Could you have a look? I've added kafka-io to the SDK Harness container.

mxm:
Planning to merge this later today if there are no further comments.

sdks/java/harness/build.gradle
tweise: I also think the expansion service should (in the future) have an option to specify additional dependencies that are not part of the environment.

tweise:
We should allow the user to specify the environment for the external transform. Would like to be able to use EMBEDDED when using the Flink runner.

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
```diff
@@ -1325,12 +1325,89 @@ public void populateDisplayData(DisplayData.Builder builder) {
...
+
+      private static Class resolveClass(String className) {
+        try {
+          return Class.forName(className);
+        } catch (ClassNotFoundException e) {
+          throw new RuntimeException("Could not find Serializer class: " + className);
+        }
+      }
+
+      private static String utf8String(byte[] bytes) {
```
aromanenko-dev: We have the same method for Read transform. Do we have similar function in any utils class already? If not, it'd make sense to put it into `org.apache.beam.sdk.util.StringUtils` to avoid code duplication.

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
```diff
@@ -1325,12 +1325,89 @@ public void populateDisplayData(DisplayData.Builder builder) {
...
+                .updateProducerProperties(producerConfig)
+                .withKeySerializer(keySerializer)
+                .withValueSerializer(valSerializer)
+                .withTopic(topic);
+        setWriteRecordsTransform(writeRecords);
+
+        return build();
+      }
+
+      private static Class resolveClass(String className) {
```
aromanenko-dev: Does it makes sense to move into `org.apache.beam.sdk.util.SerializableUtils` ?

mxm:
I already discovered this before opening this PR and created https://jira.apache.org/jira/browse/BEAM-7084. I think this can be solved independently of this PR.

sdks/java/container/Dockerfile
```diff
@@ -22,6 +22,8 @@ MAINTAINER "Apache Beam <dev@beam.apache.org>"
 ADD target/slf4j-api.jar /opt/apache/beam/jars/
 ADD target/slf4j-jdk14.jar /opt/apache/beam/jars/
 ADD target/beam-sdks-java-harness.jar /opt/apache/beam/jars/
+ADD target/beam-sdks-java-io-kafka.jar /opt/apache/beam/jars/
```
chamikaramj: Please add a TODO to remove these when we support custom environments.

sdks/java/container/Dockerfile-java11
```diff
@@ -22,6 +22,8 @@ MAINTAINER "Apache Beam <dev@beam.apache.org>"
 ADD target/slf4j-api.jar /opt/apache/beam/jars/
 ADD target/slf4j-jdk14.jar /opt/apache/beam/jars/
 ADD target/beam-sdks-java-harness.jar /opt/apache/beam/jars/
+ADD target/beam-sdks-java-io-kafka.jar /opt/apache/beam/jars/
```
chamikaramj: Ditto (here and below).

sdks/python/apache_beam/io/external/kafka.py
```diff
@@ -118,35 +118,112 @@ def expand(self, pbegin):
...
+  """
+    An external PTransform which writes KV data to a specified Kafka topic.
+    If no Kafka Serializer for key/value is provided, then key/value are
+    assumed to be byte arrays.
+
+    Note: To use this transform, you need to start the Java expansion service.
+    Please refer to the portability documentation on how to do that. The
+    expansion service address has to be provided when instantiating this
+    transform. During pipeline translation this transform will be replaced by
+    the Java SDK's KafkaIO.
```
chamikaramj: Prob. move this information to the top of the module (instead of repeating for read and write transforms).

sdks/python/apache_beam/runners/portability/flink_runner_test.py
```diff
@@ -189,12 +190,28 @@ def test_external_transforms(self):
...
                                               'common.serialization.'
                                               'LongDeserializer',
-                           expansion_service=expansion_address))
+                           expansion_service=get_expansion_service()))
       self.assertTrue('No resolvable bootstrap urls given in bootstrap.servers'
                       in str(ctx.exception),
                       'Expected to fail due to invalid bootstrap.servers, but '
                       'failed due to:\n%s' % str(ctx.exception))
 
+      # We just test the expansion but do not execute.
```
chamikaramj: FYI: @ihji is looking into adding a validates runner test suite for cross-language transforms that includes Kafka.

mxm: Yes, I saw the PR (#8397). That will be great to extend our test coverage here.

sdks/python/apache_beam/io/external/kafka.py
mxm: Wasn't sure how this is displayed in IDEs for Python development but seems fair not to repeat it.

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
mxm: I think it makes sense not to repeat this. Let me see if I can find something. `StringUtils` seems like a good place otherwise.

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
mxm: +1

mxm: Actually, I've moved the method to the root class because I didn't know whether this would be useful beyond this scope. That way at least both inner classes can use it.

mxm:
Thanks for all the comments!

PR #8338, opened 2019-04-17
[BEAM-2939] Initial SyntheticSDF as Source and add an Synthetic pipeline to sdf test

boyuanzz:
Create a SyntheticSDFAsSource to produce records like a source and add a fnapi runner test.
R: @pabloem
cc: @laschmidt @robertwb @lukecwik

### DIFF
--- a/sdks/python/apache_beam/testing/synthetic_pipeline.py
+++ b/sdks/python/apache_beam/testing/synthetic_pipeline.py
@@ -345,6 +345,14 @@ class SyntheticSDFAsSource(beam.DoFn):
     p
     | beam.Create([description1, description2,...])
     | beam.ParDo(SyntheticSDFAsSource())
+
+  NOTE:
+    The SDF.process() will have different param content between defining a DoFn
+    and runtime.
+    When defining an SDF.process, the restriction_tracker should be a
+    `RestrictionProvider`.
+    During runtime, the DoFnRunner.process_with_sized_restriction() will feed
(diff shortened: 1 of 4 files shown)


boyuanzz:
Run Portable_Python PreCommit

boyuanzz:
Run Python PreCommit

boyuanzz:
Run Python PreCommit

boyuanzz:
Kindly ping : )

sdks/python/apache_beam/testing/synthetic_pipeline.py
```diff
@@ -250,6 +252,95 @@ def default_output_coder(self):
...
+  """
+
+  def initial_restriction(self, element):
+    return (0, element['num_records'])
+
+  def create_tracker(self, restriction):
+    return restriction_trackers.OffsetRestrictionTracker(
+        restriction[0], restriction[1])
+
+  def split(self, element, restriction):
```
pabloem: Also for my edification, can `split` be called on restrictions that have already been split? If so, would the implementation of this method change? (because it seems that it calculates splits based on the 'global/unsplit' info coming from the element).

TBH you know this way better than me, so I may be barking at the wrong tree : )

boyuanzz:
So sorry for the late! Could you please take another look? @pabloem

pabloem:
LGTM. Thanks Boyuan!

PR #8349, opened 2019-04-18
[SQL] Google Cloud Data Catalog table provider

akedin:
Table provider for [Data Catalog](https://cloud.google.com/data-catalog/)
------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/extensions/sql/datacatalog/build.gradle
+++ b/sdks/java/extensions/sql/datacatalog/build.gradle
@@ -1,0 +1,63 @@
+import groovy.json.JsonOutput
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * License); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an AS IS BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+plugins { id 'org.apache.beam.module' }
+
+applyJavaNature(
+  testShadowJar: true
+)
+
+dependencies {
+  shadow library.java.grpc_google_cloud_datacatalog_v1beta1
+  shadow library.java.proto_google_cloud_datacatalog_v1beta1
+  provided project(path: ":beam-sdks-java-extensions-sql", configuration: "shadow")
+
+  // For Data Catalog GRPC client
+  provided library.java.grpc_all
+  provided library.java.google_auth_library_oauth2_http
+  provided library.java.grpc_netty
+  provided library.java.netty_tcnative_boringssl_static
+
+  // Dependencies for the example
+  provided project(path: ":beam-sdks-java-io-google-cloud-platform", configuration: "shadow")
+  provided library.java.vendored_guava_20_0
+  provided library.java.slf4j_api
+  provided library.java.slf4j_simple
+}
+
+task runDataCatalogExample(type: JavaExec) {
+  description = "Run SQL example of how to use Data Catalog table provider"
+  main = "org.apache.beam.sdk.extensions.sql.example.BeamSqlDataCatalogExample"
+  classpath = sourceSets.main.runtimeClasspath
+
+  def runner = project.findProperty('runner') ?: 'DirectRunner'
+  def queryString = project.findProperty('queryString') ?: ''
+  def outputFilePrefix = project.findProperty('outputFilePrefix') ?: ''
+  def gcpProject = project.findProperty('gcpProject') ?: 'apache-beam-testing'
+  def gcsTempRoot = project.findProperty('gcsTempRoot') ?: 'gs://temp-storage-for-end-to-end-tests/'
+
+  args = [
+    "--runner=${runner}",
+    "--queryString=${queryString}",
+    "--outputFilePrefix=${outputFilePrefix}",
+    "--project=${gcpProject}",
+    "--tempLocation=${gcsTempRoot}",
+  ]
+}
--- a/sdks/java/extensions/sql/datacatalog/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/datacatalog/DataCatalogPipelineOptions.java
+++ b/sdks/java/extensions/sql/datacatalog/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/datacatalog/DataCatalogPipelineOptions.java
@@ -1,0 +1,35 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.beam.sdk.extensions.sql.meta.provider.datacatalog;
+
+import org.apache.beam.sdk.options.Default;
+import org.apache.beam.sdk.options.Description;
+import org.apache.beam.sdk.options.PipelineOptions;
+import org.apache.beam.sdk.options.Validation;
+
+/** Pipeline options for Data Catalog table provider. */
+public interface DataCatalogPipelineOptions extends PipelineOptions {
+
+  /** DataCatalog endpoint. */
+  @Description("Data catalog endpoint.")
+  @Validation.Required
+  @Default.String("datacatalog.googleapis.com")
+  String getDataCatalogEndpoint();
+
+  void setDataCatalogEndpoint(String dataCatalogEndpoint);
+}
--- a/sdks/java/extensions/sql/datacatalog/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/datacatalog/SchemaUtils.java
+++ b/sdks/java/extensions/sql/datacatalog/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/datacatalog/SchemaUtils.java
@@ -1,0 +1,98 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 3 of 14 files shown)


sdks/java/extensions/sql/datacatalog/build.gradle
```diff
@@ -0,0 +1,59 @@
...
+ * limitations under the License.
+ */
+
+plugins { id 'org.apache.beam.module' }
+
+applyJavaNature()
+
+dependencies {
+  provided project(path: ":beam-sdks-java-extensions-sql", configuration: "shadow")
+  provided project(path: ":beam-sdks-java-io-google-cloud-platform", configuration: "shadow")
```
amaliujia: Why depending `:beam-sdks-java-io-google-cloud-platform`?

akedin: Example ends up using the `BigQueryTableProvider` or `PubsubTableProvider`, i'll move this dependency to the example section here

sdks/java/extensions/sql/datacatalog/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/datacatalog/SchemaUtils.java
```diff
@@ -0,0 +1,98 @@
...
+          .put("TYPE_DATETIME", FieldType.DATETIME)
+          .put("TYPE_DOUBLE", FieldType.DOUBLE)
+          .put("TYPE_FLOAT", FieldType.DOUBLE)
+          .put("TYPE_INT32", FieldType.INT32)
+          .put("TYPE_INT64", FieldType.INT64)
+          .put("TYPE_STRING", FieldType.STRING)
+          .put("TYPE_TIME", FieldType.logicalType(new CalciteUtils.TimeType()))
+          .put("TYPE_TIMESTAMP", FieldType.DATETIME)
+          .put(
+              "TYPE_MAP<TYPE_STRING, TYPE_STRING>",
```
amaliujia: Why MAP has only TYPE_STRING as Key and Value type? Is it because data catalog only allows STRING in MAP?

sdks/java/extensions/sql/datacatalog/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/datacatalog/DataCatalogPipelineOptions.java
```diff
@@ -0,0 +1,35 @@
...
+import org.apache.beam.sdk.options.PipelineOptions;
+import org.apache.beam.sdk.options.Validation;
+
+/** Pipeline options for Data Catalog table provider. */
+public interface DataCatalogPipelineOptions extends PipelineOptions {
+
+  /** DataCatalog endpoint. */
+  @Description("Required. Data catalog endpoint.")
+  @Validation.Required
+  @Default.String("datacatalog.googleapis.com")
```
amaliujia: Will this never change?

This it might change, we could remove default value so users will have to specify a working endpoint.

sdks/java/extensions/sql/datacatalog/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/datacatalog/SchemaUtils.java
amaliujia: Copy single comment tp a batch of review comment:

Why MAP has only TYPE_STRING as Key and Value type? Is it because data catalog only allows STRING in MAP?

akedin: No, this is how [`PubsubMessage` works](https://github.com/apache/beam/blob/b953645ed6db837d24284d7fe1fe091e7309f821/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/pubsub/PubsubMessage.java#L32), same in [`CREATE TABLE`](https://beam.apache.org/documentation/dsls/sql/create-external-table/#syntax-2).

sdks/java/extensions/sql/datacatalog/build.gradle
akedin: Done

sdks/java/extensions/sql/datacatalog/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/datacatalog/SchemaUtils.java
```diff
@@ -0,0 +1,98 @@
...
+    FieldType fieldType = getBeamFieldType(column);
+    Field field = Field.of(name, fieldType);
+
+    // set the nullable flag, or convert to a list if repeated
+    if (Strings.isNullOrEmpty(column.getMode()) || "NULLABLE".equals(column.getMode())) {
+      field = field.withNullable(true);
+    } else if ("REQUIRED".equals(column.getMode())) {
+      field = field.withNullable(false);
+    } else if ("REPEATED".equals(column.getMode())) {
+      field = Field.of(name, FieldType.array(fieldType));
```
amaliujia: How's nullability on element type of ARRAY. Say ARRAY<INT64> but it could be (null, null, 12, 13, null...)

sdks/java/extensions/sql/datacatalog/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/datacatalog/SchemaUtils.java
amaliujia: Is there a reason that don't support MAP in a general way?

sdks/java/extensions/sql/datacatalog/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/datacatalog/DataCatalogPipelineOptions.java
akedin: This should never change, it's the same for all GCP services. If it does, users still have an option to override this. We use similar approach for DF, BQ, others:

https://github.com/apache/beam/blob/98faabff70b58935b7cbc7efeacc8ec1a850479d/sdks/python/apache_beam/options/pipeline_options.py#L413

https://github.com/apache/beam/blob/279a05604b83a54e8e5a79e13d8761f94841f326/runners/google-cloud-dataflow-java/src/main/java/org/apache/beam/runners/dataflow/options/DataflowPipelineDebugOptions.java#L54

amaliujia: makes sense as it looks like permanent.

sdks/java/extensions/sql/datacatalog/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/datacatalog/SchemaUtils.java
akedin: There is no general maps support in BQ: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types

This is just to re-use whatever we already have for Pubsub. Currently we don't model JSON payloads as maps, we model them as structs, so no maps will be needed there as well.

DataCatalog can allow everything but I don't want to add something seemingly generic on our side that will be hard to switch from.

sdks/java/extensions/sql/datacatalog/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/datacatalog/SchemaUtils.java
akedin: BQ does not allow nulls in arrays: https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#array-type

JSON Arrays can potentially include `nulls`, not sure whether we should allow this or not

akedin:
run java precommit

amaliujia:
I realize there is a different understanding on data catalog as a external table:

My understanding to use data catalog as external metadata source in general. So in it's own implementations, especially type conversions to Beam, should be a complete conversion to BeamSQL types (either schema field types or BeamSQL's logical types). BeamSQL types shouldn't be limited to some special needs on, say MAP<STRING, STRING>. BeamSQL types should compose on primitive types (include array, struct and map), logical types and nullability. And schema from data catalog will also maintain composability, unless data catalog cannot express BeamSQL type's composability.

Other data sources might be a subset of BeamSQL type system. E.g. BQ.

By doing so data catalog will only rely on BeamSQL type system and open to other data sources that are compatible with BeamSQL.

I would agree not to achieve this goal in this PR, not because Pubsub' MAP<STRING, STRING> support or BQ does not support MAP,  but because this goal is too big for a single PR.

akedin:
I agree with high level picture. But DataCatalog schema specification does not satisfy general purpose at the moment. Neither our table providers nor Beam SQL support any wide use cases in terms of schemas and sources.

sdks/java/extensions/sql/datacatalog/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/datacatalog/SchemaUtils.java
amaliujia: I am seeing that type is just a string in data catalog: https://github.com/googleapis/googleapis/blob/b7f88d8827df8a78e1bb1d56b0f56a1bc23c6359/google/cloud/datacatalog/v1beta1/schema.proto#L39

It is so bad for BeamSQL. `MAP<STRING, STRING>` is an example that suffers due to this design, which is also not extensible (need a entry look up for any other MAP).

sdks/java/extensions/sql/datacatalog/build.gradle
```diff
@@ -0,0 +1,59 @@
...
+ * limitations under the License.
+ */
+
+plugins { id 'org.apache.beam.module' }
+
+applyJavaNature()
+
+dependencies {
+  provided project(path: ":beam-sdks-java-extensions-sql", configuration: "shadow")
+  shadow library.java.grpc_all
```
amaliujia: Can we repackage grpc libraies as we have seen leaked grpc that caused class conflict?

amaliujia: Or make it provided to remove it from shadowed jar.

akedin: i'll try

akedin: relocation wasn't as trivial as I expected, made it provided for now, will try to figure out the relocation later if needed

akedin:
run sql postcommit

akedin:
run java postcommit

sdks/java/extensions/sql/datacatalog/build.gradle
amaliujia: Sounds good

akedin:
+ @kennknowles @apilloud

sdks/java/extensions/sql/datacatalog/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/datacatalog/DataCatalogPipelineOptions.java
```diff
@@ -0,0 +1,35 @@
...
+import org.apache.beam.sdk.options.Default;
+import org.apache.beam.sdk.options.Description;
+import org.apache.beam.sdk.options.PipelineOptions;
+import org.apache.beam.sdk.options.Validation;
+
+/** Pipeline options for Data Catalog table provider. */
+public interface DataCatalogPipelineOptions extends PipelineOptions {
+
+  /** DataCatalog endpoint. */
+  @Description("Required. Data catalog endpoint.")
```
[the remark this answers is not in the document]
akedin: Makes sense. Removing `required` from the description

akedin:
run sql postcommit

akedin:
@amaliujia I will merge this if you don't have objections

PR #8354, opened 2019-04-19
[BEAM-5379] Go Modules versioning support

nfisher:

------------------------

--- | --- | --- | --- | ---


nfisher:
R: @lostluck

lukecwik:
You'll need to add an Apache 2.0 License header to your newly added files.

lostluck:
If it were this simple I'd have done it already. :) I too hoped it was this simple, but here are at least 2 problems with just adding the go.mod and go.sum files:

These are totally up for discussion BTW, I'm willing to be convinced whichever way, these are just the conclusions I came to when I thought it through.

1. Beam testing (all those lovely precommit/post commit tests) uses Gradle to manage these tasks. To do so with go we use GoGradle We need to remove the gogradle lock file which it uses to manage it's only vendored versions of things, and validate that GoGradle works with Go Modules, and outside of a GOPATH. Otherwise we're defeating the prime benefit of GoModules, which is hermetic reproducible builds, and we'd be testing the SDK under different conditions/and dependencies than users would receive them.
* Two dependency files (effectively) would mean there's additional hoops to add dependencies rather than just running the go tools. We'd also need to run the gradle tools. This impedes adding useful dependencies, and contributions from other Go users.
* This is the part that needs validation, and is why I haven't simply done it yet, as it requires time to handle.
  * For example I don't know if GoGradle operates properly outside of a GoPath.

2.  Putting these here makes the claim that the SDK is v2.X, when it should be a v0.X.
* I do not want folks to mistakenly assume the SDK is production ready. Due to the IOs, it's not.
* There are some parts that should be v2.X, since they're not part of the SDK, just things written in Go, like the container boot code, or they should be clients of the SDK. I posit the integration & regression tests, and the examples should be outside of the user package, and depend on a clear version of the user package. See 
* Fortunately this divide is clear at sdks/go/pkg/beam , so we can put a modules at sdks/go (as you've done), and at sdks/go/pkg/beam. 
* Then it's a matter of using git tags with a sub path prefix (sdks/go/pkg/beam/v0.X) 
* And validating that things continue to work. 
See https://github.com/golang/go/wiki/Modules#faqs--multi-module-repositories for clearer instructions and examples.

Now, while they're inter-related issues, we could probably do 2 separate from 1, as long as we don't live in that mixed state for too long.

See 
https://docs.google.com/document/d/1ZjP30zNLWTu_WzkWbgY8F_ZXlA_OWAobAD9PuohJxPg/edit#heading=h.drpipq762xi7 for some lengthy thoughts on they whys for most of this.

nfisher:
I'm assuming there's a tonne of context I lack! Hopefully this wall of text yields a fruitful discussion through my naive questions and poking at the boundary of required vs desired.

I gave a first reading through the Gdoc you referenced thanks for that. I probably need another pass or two to ensure I grasp the nuance.

> If it were this simple I'd have done it already. :)

I assumed as much but was curious whether there was anything that prevented build/test with the naivest/smallest change possible. I suppose I could've equally discussed it in the issue but I sometimes find working code is easier to identify and clarify concrete concerns and issues.

> Beam testing ... uses Gradle to manage these tasks.
> prime benefit of GoModules, which is hermetic reproducible builds

Is GoGradle a hard requirement or is it merely the desire and ability to employ a common build tool? If an alternative reproducible build/test cycle could be implemented by other means which still provided Gradle build targets, precommit, etc would that be considered/tick the boxes? From a dependency graph basis is Go modules not an adequate mechanism to yield a hermetic reproducible build (e.g. go.sum pins detailed version info for the full dependency graph including transitive deps)? Tangental but on the topic of hermetic builds presumably Go tool-chain compatibility is a concern? I've seen a number of libraries cite they will support the latest N minor versions of Go. Is there a goal for Beam to do similar (e.g. support Go v{1.N & 1.N-1} currently {v1.11 & v1.12})? Implies there's a build matrix required for version verification is there one already?

> makes the claim that the SDK is v2.X, when it should be a v0.X

Understanding that more broadly 0.X is a common idiom to indicate "there be dragons" and "API changes are highly probable". If it's explicit in the SDK documentation for Go does the version number *really* matter? Spinning it from another angle could a v2.X tagging be a "compatible with/to" in the Beam graph definition sense rather than a measure of maturity? As a further example, am I correct in stating that the Python and Java libraries haven't always had feature parity but have carried the same version at a given tag/commit?

As a user I think I would find a capability matrix a more informative measure of utility than version at this stage in the project as versioning is often subjective and somewhat arbitrary. Naively I feel in the Go community (or at least within projects I've consumed) there's often "fluidity/fluctuation" in API's even within minor versions.

> Fortunately this divide is clear at sdks/go/pkg/beam , so we can put a modules at sdks/go (as you've done), and at sdks/go/pkg/beam.

Probably an example where my lack of context/familiarity with the codebase is failing me. Am I correct in understanding that `sdks/go` is the "private" code and CLI executables and `pkg` is the public API where `sdks/go`, `sdks/go/pkg/beam`, and user code will often be compiled into a single binary executable? If yes, as a user I think it's fair to expect them to be the same version. I did note that beamctl exists in the cmd folder so I assume there's also client/server tooling to consider and perhaps that's the larger concern you're outlining?

lostluck:
I'm fluent in wall of text, and am happy to provide all the context I have available. Lets get this done!
+1 to discussing here at the code. I agree entirely.

On Go Gradle:
* I am ambivalent about keeping the plugin, and modifying things to direct go tool usage (likely shell scripts) so long as the tasks can continue to be triggered via Gradle. It simplifies things for test infrastructure that way.
  * There was a bit of [discussion on ditching gogradle recently, and I got reasonable pushback](https://lists.apache.org/thread.html/2754c89ef5f4ad850b98fa558ed245a5bdaa099c04bdcfbd0a34171f@%3Cdev.beam.apache.org%3E)
  * If we can address those concerns (I think we can), we can probably do it with the scripty + go tools approach.
* Go Modules (committing both a go.mod & go.sum) do accomplish the goal of hermetic reproducible builds. There's a reasonable fear of external tooling by those who work with gradle more regularly.
* Go Gradle likely sets up a GO_PATH which would disable modules anyway by default without setting the right ENV variable setting, and then vendor the packages listed in the lock file.
* You can find all the different Go "targets" by searching for [applyGoNature] in this repo.(https://github.com/apache/beam/search?q=applyGoNature&unscoped_q=applyGoNature) 

On the v0.X:
* Sadly, as much as I like saying "it's in the documentation", discoverability of those statements is hard. The version number (such as it will display in a user's go.mod) is an unambiguous statement that's much harder to miss.
 * "There be Dragons" is largely due to the lack of mature native Go IO connectors. As I outlined recently in a dev list post on the [Go SDK status](https://lists.apache.org/thread.html/8f729da2d3009059d7a8b2d8624446be161700dcfa953939dd3530c6@%3Cdev.beam.apache.org%3E), the connectors have received very little, if any love, and aren't currently tested meaningfully.
  * I can't in good conscious recommend users depend on those IOs at this time. Even implicitly via the version numbering, since they won't do the job Beam promises, which is allow you to write code that can scale.
  * It is possible to write scalable batch sinks, with certain techniques (resharding via CoGBKs, expanding to 1 DoFn per input file, etc), but I'm biased to either Cross Language IO, or runner availability of Splittable Do Fns (even just the initial splitting), so we can implement the Go support for that, and directly have the final solution, rather than migrate through a few stages.
* There's also the matter of importing packages that are v2: They need to have the v2 in the import path etc.
* The actual version number largely speaks to the passage of time, but it is useful for runners and services to be able to say "The Go SDK versions X-Z all work on the FooBar runner!", and they can also maintain their own testing to be able to deprecate support for such versions later.

On Python and Java SDK versions being in sync:
* This is a nuance from the history of the project. Python and Java have "pre-portability" implementations, where the SDK and the runner were tightly coupled on the worker. That is, every SDK+Runner pair required it's own Worker. As you can imagine, this doesn't scale. Everything (runners+ SDKs) *had* to be at the same version. 
* Portability (via the FnAPI) allows each SDK to have it's own worker side harness, that communicates to a runner side harness. Then each runner , and each SDK merely maintains it's own half, simplifying both SDK and runner development.
* The Go SDK has only ever had a "portable" implementation.

On Independant Language Versions:
* I personally don't see it making sense long term for SDKs to be on the same version. It's useful for users to know "the SDK didn't change, so my pipeline code will be the same break" 
* Why should the Go SDK get a patch release when Java as a bug to fix?
* If Generics ever happen for Go, and adopting them requires making API breaking changes, why should Java and Python get a major version bump for the sake of Go?
* That it's compatible with runners on Beam vFoo.Bar, and that can be tested independantly.

On the version of Go being used:
* Currently, IIRC the Jenkin's machines are using Go1.10. Our GoGradle is currently [configured to use go1.10](https://github.com/apache/beam/blob/9c8a8dc241296ad5003124c8bcb94a50bb53beb5/buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy#L1310) and I'm not sure if we updated the Jenkins machines to have Go 1.12 when they recently broke.
* We *should* provide some kind of statement of what versions of Go are supported, but this is another reason for being v0.X, we haven't made those kinds of commitments yet.
* There's similar work being done for tracking supported Java and Python versions, so this seems reasonable.

On having 2 modules, rather than one:
* It's useful to be able to "retarget" the tests independently of the version of the Go SDK being used, eg to find regressions run performance comparisons against different minor versions etc.
  * Validating compatibility of Go SDK versions X against Runner Version Y
* Sets us up better for the non-zero eventuality that the repo splits per SDK language one portability is complete. There would still be some work to do of course WRT path changes (unless we adopt Vanity Import URLs), but the modules would be independent.
* Clean split of dependencies for things used to vet the SDK against different runners, vs what users actually need to use the SDK.

I might be repeating myself from the doc a bit, but re-writing is handy in making thoughts more concrete.

nfisher:
@lukecwik do generated files need the license as well?

`go.sum` is generated from the `go.mod` and is basically equivalent to a package lock file in other package managers. AFAICT I'm not sure `go.sum` files allow for a comment but I haven't explored this extensively beyond copy-pasting the existing header.

lostluck:
I do believe that even generated files need the license comment.

Turns out resolving this versioning thing in some capacity will help solve an issue around generating the go versions of protos too. There's currently lots of churn since the versions of protoc-gen-go and the go proto implementation will change with every new person adding protos, since those versions aren't locked presently (but should be).

Have you made additional headway on this in the meantime? At this point the temporary 2 lockfile state is probably preferable, and easier to handle than the "no one can meaningfully update Go protos" state we're currently in.

nfisher:
Had family in town so doing the royal tour so to speak. Likely to look at it Sunday evening. Want to map out all of the concerns including those outlined in the email. It feels some of the concerns raised are more in direct relation to the python build is that a correct understanding? Or is python used for integration tests across all language SDKs?

lostluck:
Ah! No worries. Life/Family >>> code!

That matches my understanding as well. Python requires a whole virtual
environment set up to ensure hermetic builds while Go Modules handle that
in the tools without additional installations. We shouldn't need worry
about clearing the Jenkins machines versioning caches, since Go should be
able to manage that efficiently for us.

I just checked with Yifan who recently handled the Jenkins machine update:
Each of the languages handles their own integration test set up
independently.
In short, if we can address the issues observed by Python's set up, we
should be golden.

forsaken628:
in go 1.16 Module-aware mode is enabled by default

PR #8363, opened 2019-04-19
Adding Python samples to the Stateful Processing post.

pabloem:
Also add a couple more pydocs to python transforms.

r: @kennknowles 
cc: @mxm

### DIFF
--- a/sdks/python/apache_beam/transforms/trigger.py
+++ b/sdks/python/apache_beam/transforms/trigger.py
@@ -77,7 +77,7 @@ class AccumulationMode(object):
 class _StateTag(with_metaclass(ABCMeta, object)):
   """An identifier used to store and retrieve typed, combinable state.
 
-  The given tag must be unique for this stage."""
+  The given tag must be unique for this step."""
 
   def __init__(self, tag):
     self.tag = tag
--- a/sdks/python/apache_beam/transforms/trigger.py
+++ b/sdks/python/apache_beam/transforms/trigger.py
@@ -96,9 +96,8 @@ class _ValueStateTag(_StateTag):
 class _CombiningValueStateTag(_StateTag):
   """StateTag pointing to an element, accumulated with a combiner.
 
-  The given tag must be unique for this stage.  If CombineFn is None then
-  all elements will be returned as a list, otherwise the given CombineFn
-  will be applied (possibly incrementally and eagerly) when adding elements."""
+  The given tag must be unique for this step. The given CombineFn will be
+  applied (possibly incrementally and eagerly) when adding elements."""
 
   # TODO(robertwb): Also store the coder (perhaps extracted from the combine_fn)
   def __init__(self, tag, combine_fn):
--- a/sdks/python/apache_beam/transforms/trigger.py
+++ b/sdks/python/apache_beam/transforms/trigger.py
@@ -67,8 +67,7 @@ __all__ = [
 
 
 class AccumulationMode(object):
-  """Controls what to do with data when a trigger fires multiple times.
-  """
+  """Controls what to do with data when a trigger fires multiple times."""
   DISCARDING = beam_runner_api_pb2.AccumulationMode.DISCARDING
   ACCUMULATING = beam_runner_api_pb2.AccumulationMode.ACCUMULATING
   # TODO(robertwb): Provide retractions of previous outputs.
@@ -78,10 +77,7 @@ class AccumulationMode(object):
 class _StateTag(with_metaclass(ABCMeta, object)):
   """An identifier used to store and retrieve typed, combinable state.
 
-  The given tag must be unique for this stage.  If CombineFn is None then
-  all elements will be returned as a list, otherwise the given CombineFn
-  will be applied (possibly incrementally and eagerly) when adding elements.
-  """
+  The given tag must be unique for this stage."""
 
   def __init__(self, tag):
     self.tag = tag
@@ -98,7 +94,11 @@ class _ValueStateTag(_StateTag):
 
 
 class _CombiningValueStateTag(_StateTag):
-  """StateTag pointing to an element, accumulated with a combiner."""
+  """StateTag pointing to an element, accumulated with a combiner.
+
+  The given tag must be unique for this stage.  If CombineFn is None then
+  all elements will be returned as a list, otherwise the given CombineFn
+  will be applied (possibly incrementally and eagerly) when adding elements."""
 
   # TODO(robertwb): Also store the coder (perhaps extracted from the combine_fn)
   def __init__(self, tag, combine_fn):
--- a/sdks/python/apache_beam/transforms/userstate_test.py
+++ b/sdks/python/apache_beam/transforms/userstate_test.py
@@ -31,7 +31,6 @@ from apache_beam.runners.common import DoFnSignature
 from apache_beam.testing.test_pipeline import TestPipeline
 from apache_beam.testing.test_stream import TestStream
 from apache_beam.testing.util import equal_to
-from apache_beam.transforms import combiners
 from apache_beam.transforms import userstate
 from apache_beam.transforms.combiners import ToListCombineFn
 from apache_beam.transforms.combiners import TopCombineFn
@@ -501,8 +500,7 @@ class StatefulDoFnOnDirectRunnerTest(unittest.TestCase):
 
   def test_index_assignment(self):
     class IndexAssigningStatefulDoFn(DoFn):
-      INDEX_STATE = CombiningValueStateSpec('index', VarIntCoder(),
-                                            combiners.SumCombineFn())
+      INDEX_STATE = CombiningValueStateSpec('index', sum)
 
       def process(self, element, state=DoFn.StateParam(INDEX_STATE)):
         unused_key, value = element
--- a/sdks/python/apache_beam/transforms/userstate_test.py
+++ b/sdks/python/apache_beam/transforms/userstate_test.py
@@ -504,8 +504,8 @@ class StatefulDoFnOnDirectRunnerTest(unittest.TestCase):
 
       def process(self, element, state=DoFn.StateParam(INDEX_STATE)):
         unused_key, value = element
-        next_index = state.read() or 0
-        yield (value, next_index)
+        current_index = state.read()
+        yield (value, current_index)
         state.add(1)
 
     with TestPipeline() as p:
--- a/sdks/python/apache_beam/transforms/userstate_test.py
+++ b/sdks/python/apache_beam/transforms/userstate_test.py
@@ -31,6 +31,7 @@ from apache_beam.runners.common import DoFnSignature
 from apache_beam.testing.test_pipeline import TestPipeline
 from apache_beam.testing.test_stream import TestStream
 from apache_beam.testing.util import equal_to
+from apache_beam.transforms import combiners
 from apache_beam.transforms import userstate
 from apache_beam.transforms.combiners import ToListCombineFn
 from apache_beam.transforms.combiners import TopCombineFn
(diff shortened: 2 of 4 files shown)


pabloem:
Run Website PreCommit

mxm:
Thanks! Looks good. Have you tested this with an actual project? Just asking because we do not have checks for the doc code.

CC @robertwb

pabloem:
The indexing dofn is run as part of the unit tests for user state. The rest of the code I haven't ran... it's just a translation from the Java code to Python code (using BagState, as Python doesn't have a plain ValueState).

website/src/_posts/2017-02-13-stateful-processing.md
```diff
@@ -501,29 +543,60 @@ new DoFn<KV<UserId, Event>, KV<UserId, Prediction>>() {
 ```py
 # State and timers are not yet supported in Beam's Python SDK.
 # Watch this space!
```
robertwb: You can remove these comment lines now.

website/src/_posts/2017-02-13-stateful-processing.md
```diff
@@ -501,29 +543,60 @@ new DoFn<KV<UserId, Event>, KV<UserId, Prediction>>() {
...
+
+    model_state.clear()
+    model_state.add(model)
+
+    if (previous_prediction is None
+        or self.should_output_prediction(
+            previous_prediction, new_prediction)):
+        previous_pred_state.clear()
+        previous_pred_state.add(new_prediction)
+        yield (user, new_prediction)
```
robertwb: 2 space indentation.

mxm:
On this note, could we add `ValueState` to the Python SDK. Perhaps just as a simple wrapper around ListState?

website/src/_posts/2017-02-13-stateful-processing.md
pabloem: Done.

website/src/_posts/2017-02-13-stateful-processing.md
pabloem: Done.

website/src/_posts/2017-02-13-stateful-processing.md
```diff
@@ -293,8 +293,15 @@ new DoFn<KV<MyKey, MyValue>, KV<Integer, KV<MyKey, MyValue>>>() {
 ```
 
 ```py
-# State and timers are not yet supported in Beam's Python SDK.
-# Watch this space!
+class IndexAssigningStatefulDoFn(DoFn):
+  INDEX_STATE = BagStateSpec('index', VarIntCoder())
```
robertwb: As per the discussion, please update this to use CombiningState. (Possibly elsewhere as well.)

website/src/_posts/2017-02-13-stateful-processing.md
```diff
@@ -499,31 +541,60 @@ new DoFn<KV<UserId, Event>, KV<UserId, Prediction>>() {
...
+              previous_pred_state=beam.DoFn.StateParam(PREVIOUS_PREDICTION),
+              model_state=beam.DoFn.StateParam(MODEL_STATE)):
+    user = user_event[0]
+    event = user_event[1]
+    model = model_state.read()
+    previous_prediction = previous_pred_state.read()
+
+    new_prediction = model.prediction(event)
+    model.add(event)
+
```
robertwb: Use ModelFromEventsFn above instead of read-modify-write?

pabloem:
@robertwb I've addressed some of your comments. I added a public `SumCombineFn`, but then this also unblocked `combiners.Sum.PerKey`, and `combiners.Sum.Globally`. PErhaps only the CombineFn? WDYT?

pabloem:
ok we can review this after merging https://github.com/apache/beam/pull/8402

pabloem:
@robertwb ready for review

sdks/python/apache_beam/transforms/userstate_test.py
```diff
@@ -500,14 +500,13 @@ def process(self, element, ts=DoFn.TimestampParam):
 
   def test_index_assignment(self):
     class IndexAssigningStatefulDoFn(DoFn):
-      INDEX_STATE = BagStateSpec('index', VarIntCoder())
+      INDEX_STATE = CombiningValueStateSpec('index', sum)
 
       def process(self, element, state=DoFn.StateParam(INDEX_STATE)):
         unused_key, value = element
-        next_index, = list(state.read()) or [0]
+        next_index = state.read() or 0
```
robertwb: The `or 0` is no longer needed.

website/src/_posts/2017-02-13-stateful-processing.md
```diff
@@ -293,8 +293,14 @@ new DoFn<KV<MyKey, MyValue>, KV<Integer, KV<MyKey, MyValue>>>() {
...
 
 ```py
-# State and timers are not yet supported in Beam's Python SDK.
-# Watch this space!
+class IndexAssigningStatefulDoFn(DoFn):
+  INDEX_STATE = CombiningStateSpec('index', sum)
+
+  def process(self, element, state=DoFn.StateParam(INDEX_STATE)):
+    unused_key, value = element
+    next_index = state.read() or 0
```
robertwb: Same.

sdks/python/apache_beam/transforms/trigger.py
```diff
@@ -98,7 +94,11 @@ def with_prefix(self, prefix):
 
 
 class _CombiningValueStateTag(_StateTag):
-  """StateTag pointing to an element, accumulated with a combiner."""
+  """StateTag pointing to an element, accumulated with a combiner.
+
+  The given tag must be unique for this stage.  If CombineFn is None then
```
kennknowles: What is a stage? This is not a concept in the Beam model. The tag must be unique for a step / transform / DoFn.

kennknowles: And signing yourself up for a lifetime of `if combine_fn is not None` doesn't seem very pleasant...

website/src/_posts/2017-02-13-stateful-processing.md
```diff
@@ -499,31 +540,59 @@ new DoFn<KV<UserId, Event>, KV<UserId, Prediction>>() {
...
+      previous_pred_state.add(new_prediction)
+      yield (user, new_prediction)
 ```
 
 Let's walk through it,
 
- - You have two state cells declared, `@StateId("model")` to hold the current
-   state of the model for a user and `@StateId("previousPrediction")` to hold
+ - You have two state cells declared,
+   `@StateId("model")/StateParam(MODEL_STATE)` to hold the current
```
kennknowles: Move the slash out of the code font, at least. I wonder if there is another way to present both Python and Java. Perhaps a more long-form way explanation, or a separate table and drop the syntax from the paragraph.

website/src/_posts/2017-02-13-stateful-processing.md
```diff
@@ -499,31 +540,59 @@ new DoFn<KV<UserId, Event>, KV<UserId, Prediction>>() {
...
+   state of the model for a user and
+   `@StateId("previousPrediction")/StateParam(PREVIOUS_PREDICTION)` to hold
    the prediction output previously.
  - Access to the two state cells by annotation in the `@ProcessElement` method
-   is as before.
- - You read the current model via `modelState.read()`. Because state is also
-   per-key-and-window, this is a model just for the UserId of the Event
-   currently being processed.
+   is as before, and in Python by using an argument placeholder.
```
kennknowles: You know, at this point it is starting to see like it should be documentation rather than editing the blog post. I say this because blog posts are dated and it is not great to go back and rewrite them like this. I know it gets a lot of hits, but probably we just need docs and they can receive all the hits.

website/src/_posts/2017-02-13-stateful-processing.md
```diff
@@ -278,7 +278,7 @@ new DoFn<KV<MyKey, MyValue>, KV<Integer, KV<MyKey, MyValue>>>() {
 
   // A state cell holding a single Integer per key+window
   @StateId("index")
-  private final StateSpec<ValueState<Integer>> indexSpec = 
+  private final StateSpec<ValueState<Integer>> indexSpec =
```
kennknowles: The first change to the file should be a bold and loud header about the blog post being updated.

website/src/_posts/2017-02-13-stateful-processing.md
mxm: I agree, we shouldn't be rewriting the prose section of the blog post, but I still think that adding the Python code to the blog post has value.

sdks/python/apache_beam/transforms/trigger.py
pabloem: Removed that line.

website/src/_posts/2017-02-13-stateful-processing.md
pabloem: Done.

website/src/_posts/2017-02-13-stateful-processing.md
mxm: I believe you can add span around the Java/Python code, e.g. `<span language=py>Python code</span>` which will only show the selected language. This way the prose is not distorted with two languages here.

sdks/python/apache_beam/transforms/trigger.py
```diff
@@ -78,10 +77,7 @@ class AccumulationMode(object):
 class _StateTag(with_metaclass(ABCMeta, object)):
   """An identifier used to store and retrieve typed, combinable state.
 
-  The given tag must be unique for this stage.  If CombineFn is None then
-  all elements will be returned as a list, otherwise the given CombineFn
-  will be applied (possibly incrementally and eagerly) when adding elements.
-  """
+  The given tag must be unique for this stage."""
```
robertwb: s/stage/step/

website/src/_posts/2017-02-13-stateful-processing.md
```diff
@@ -499,31 +543,59 @@ new DoFn<KV<UserId, Event>, KV<UserId, Prediction>>() {
...
+   state of the model for a user and
+   `@StateId("previousPrediction")/StateParam(PREVIOUS_PREDICTION)` to hold
    the prediction output previously.
  - Access to the two state cells by annotation in the `@ProcessElement` method
-   is as before.
- - You read the current model via `modelState.read()`. Because state is also
-   per-key-and-window, this is a model just for the UserId of the Event
-   currently being processed.
+   is as before, and in Python by using an argument placeholder.
+ - You read the current model via `modelState.read()` / `model_state.read()`.
```
robertwb: I would omit the python version of the code in the prose. I think the mapping is clear enough, especially with the disclaimer above.

website/src/_posts/2017-02-13-stateful-processing.md
robertwb: I agree we need documentation, but I also think it's worth updating this post (and not holding that up further). I would exclude the prose changes except the ", and in Python by using an argument placeholder" phrase which I think may not be as obvious as the others. (We could do fancy things with language-specific spans, but I would also be fine letting the prose remain, with that one addition, as it was.)

mxm: +1 I also think it is worth getting this in as a reference point before we add further documentation.

sdks/python/apache_beam/transforms/trigger.py
pabloem: Done.

website/src/_posts/2017-02-13-stateful-processing.md
pabloem: I think it does no harm. Is it okay if I leave it?

pabloem:
what should we do about this?

website/src/_posts/2017-02-13-stateful-processing.md
```diff
@@ -499,31 +543,59 @@ new DoFn<KV<UserId, Event>, KV<UserId, Prediction>>() {
...
  - If the prediction has changed a significant amount since the last time you
-   output, you emit it via `context.output(...)` and save the prediction using
-   `previousPredictionState.write(...)`. Here the decision is relative to the
-   prior prediction output, not the last one computed - realistically you might
-   have some complex conditions here.
+   output, you emit it via `context.output(...)` or `yield` in Python, and
+   save the prediction using `previousPredictionState.write(...)`, or
+   `previous_pred_state.clear()` and `previous_pred_state.add(...)` in Python.
+   Here the decision is relative to the prior prediction output, not the last
+   one computed - realistically you might have some complex conditions here.
```
mxm: I think the only debatable change is this prose which is hard to read because it intermingles Java and Python. You can conditionally display text with a `span` element which has `class=languagepy` or `class=languagejava`.

Other than that, I'm absolutely for getting this in.

website/src/_posts/2017-02-13-stateful-processing.md
robertwb: It harms the readability. Having both variations is not worth stuttering of repeating tiny variations in the code. (As Max says, we can use a span with langauge classes, or I would be OK with just leaving the Java version in the prose as it was originally written.)

mxm:
@kennknowles Are you ok with this now that the prose changes have been reverted?

PR #8364, opened 2019-04-19
[BEAM-7122] Adding an accumulating trigger test

pabloem:
Testing that a trigger fires as expected for accumulating outputs.


pabloem:
r: @aaltay

pabloem:
Run Python Precommit

PR #8390, opened 2019-04-24
 [BEAM-7043] Add DynamoDBIO

cmachgodaddy:
This PR is going to add a new component which will enable user read and write data to AWS DynamoDb

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/io/amazon-web-services/build.gradle
+++ b/sdks/java/io/amazon-web-services/build.gradle
@@ -29,6 +29,7 @@ dependencies {
   compile project(path: ":sdks:java:core", configuration: "shadow")
   compile library.java.aws_java_sdk_cloudwatch
   compile library.java.aws_java_sdk_core
+  compile library.java.aws_java_sdk_dynamodb
   compile library.java.aws_java_sdk_s3
   compile library.java.aws_java_sdk_sns
   compile library.java.aws_java_sdk_sqs
@@ -47,6 +48,7 @@ dependencies {
   testCompile library.java.mockito_core
   testCompile library.java.junit
   testCompile 'org.elasticmq:elasticmq-rest-sqs_2.12:0.14.1'
+  testCompile 'org.testcontainers:localstack:1.11.2'
   testRuntimeOnly library.java.slf4j_jdk14
   testRuntimeOnly project(":runners:direct-java")
 }
--- a/sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/dynamodb/AttributeValueCoder.java
+++ b/sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/dynamodb/AttributeValueCoder.java
@@ -1,0 +1,166 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.beam.sdk.io.aws.dynamodb;
+
+import com.amazonaws.services.dynamodbv2.model.AttributeValue;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.nio.ByteBuffer;
+import java.util.List;
+import java.util.stream.Collectors;
+import org.apache.beam.sdk.coders.AtomicCoder;
+import org.apache.beam.sdk.coders.BooleanCoder;
+import org.apache.beam.sdk.coders.ByteArrayCoder;
+import org.apache.beam.sdk.coders.Coder;
+import org.apache.beam.sdk.coders.CoderException;
+import org.apache.beam.sdk.coders.ListCoder;
+import org.apache.beam.sdk.coders.MapCoder;
+import org.apache.beam.sdk.coders.StringUtf8Coder;
+
+/** A {@link Coder} that serializes and deserializes the {@link AttributeValue} objects. */
+public class AttributeValueCoder extends AtomicCoder<AttributeValue> {
+
+  /** Data type of each value type in AttributeValue object. */
+  private enum AttributeValueType {
+    s, // for String
+    n, // for Number
+    b, // for Byte
+    sS, // for List of String
+    nS, // for List of Number
+    bS, // for List of Byte
+    m, // for Map of String and AttributeValue
+    l, // for list of AttributeValue
+    bOOL, // for Boolean
+    nULLValue, // for null
+  }
+
+  private static final AttributeValueCoder INSTANCE = new AttributeValueCoder();
+
+  private static final ListCoder<String> LIST_STRING_CODER = ListCoder.of(StringUtf8Coder.of());
+  private static final ListCoder<byte[]> LIST_BYTE_CODER = ListCoder.of(ByteArrayCoder.of());
+
+  private static final ListCoder<AttributeValue> LIST_ATTRIBUTE_CODER =
+      ListCoder.of(AttributeValueCoder.of());
+  private static final MapCoder<String, AttributeValue> MAP_ATTRIBUTE_CODER =
+      MapCoder.of(StringUtf8Coder.of(), AttributeValueCoder.of());
+
+  private AttributeValueCoder() {}
+
+  public static AttributeValueCoder of() {
+    return INSTANCE;
+  }
+
+  @Override
+  public void encode(AttributeValue value, OutputStream outStream) throws IOException {
+
+    if (value.getS() != null) {
+      StringUtf8Coder.of().encode(AttributeValueType.s.toString(), outStream);
+      StringUtf8Coder.of().encode(value.getS(), outStream);
+    } else if (value.getN() != null) {
+      StringUtf8Coder.of().encode(AttributeValueType.n.toString(), outStream);
+      StringUtf8Coder.of().encode(value.getN(), outStream);
+    } else if (value.getBOOL() != null) {
+      StringUtf8Coder.of().encode(AttributeValueType.bOOL.toString(), outStream);
+      BooleanCoder.of().encode(value.getBOOL(), outStream);
+    } else if (value.getB() != null) {
+      StringUtf8Coder.of().encode(AttributeValueType.b.toString(), outStream);
+      ByteArrayCoder.of().encode(convertToByteArray(value.getB()), outStream);
+    } else if (value.getSS() != null) {
+      StringUtf8Coder.of().encode(AttributeValueType.sS.toString(), outStream);
+      LIST_STRING_CODER.encode(value.getSS(), outStream);
+    } else if (value.getNS() != null) {
+      StringUtf8Coder.of().encode(AttributeValueType.nS.toString(), outStream);
+      LIST_STRING_CODER.encode(value.getNS(), outStream);
+    } else if (value.getBS() != null) {
+      StringUtf8Coder.of().encode(AttributeValueType.bS.toString(), outStream);
+      LIST_BYTE_CODER.encode(convertToListByteArray(value.getBS()), outStream);
+    } else if (value.getL() != null) {
+      StringUtf8Coder.of().encode(AttributeValueType.l.toString(), outStream);
+      LIST_ATTRIBUTE_CODER.encode(value.getL(), outStream);
+    } else if (value.getM() != null) {
+      StringUtf8Coder.of().encode(AttributeValueType.m.toString(), outStream);
+      MAP_ATTRIBUTE_CODER.encode(value.getM(), outStream);
+    } else if (value.getNULL() != null) {
+      StringUtf8Coder.of().encode(AttributeValueType.nULLValue.toString(), outStream);
+      BooleanCoder.of().encode(value.getNULL(), outStream);
+    } else {
+      throw new CoderException("Unknown Type");
+    }
+  }
+
+  @Override
+  public AttributeValue decode(InputStream inStream) throws IOException {
+    AttributeValue attrValue = new AttributeValue();
+
+    String type = StringUtf8Coder.of().decode(inStream);
+    AttributeValueType attrType = AttributeValueType.valueOf(type);
+
+    switch (attrType) {
+      case s:
+        attrValue.setS(StringUtf8Coder.of().decode(inStream));
+        break;
+      case n:
+        attrValue.setN(StringUtf8Coder.of().decode(inStream));
+        break;
+      case bOOL:
+        attrValue.setBOOL(BooleanCoder.of().decode(inStream));
+        break;
+      case b:
+        attrValue.setB(ByteBuffer.wrap(ByteArrayCoder.of().decode(inStream)));
+        break;
+      case sS:
+        attrValue.setSS(LIST_STRING_CODER.decode(inStream));
+        break;
+      case nS:
+        attrValue.setNS(LIST_STRING_CODER.decode(inStream));
+        break;
+      case bS:
+        attrValue.setBS(convertToListByteBuffer(LIST_BYTE_CODER.decode(inStream)));
+        break;
+      case l:
+        attrValue.setL(LIST_ATTRIBUTE_CODER.decode(inStream));
+        break;
+      case m:
+        attrValue.setM(MAP_ATTRIBUTE_CODER.decode(inStream));
+        break;
+      case nULLValue:
+        attrValue.setNULL(BooleanCoder.of().decode(inStream));
+        break;
+      default:
+        throw new CoderException("Unknown Type");
+    }
+
+    return attrValue;
+  }
+
+  private List<byte[]> convertToListByteArray(List<ByteBuffer> listByteBuffer) {
+    return listByteBuffer.stream().map(this::convertToByteArray).collect(Collectors.toList());
+  }
+
+  private byte[] convertToByteArray(ByteBuffer buffer) {
+    byte[] bytes = new byte[buffer.remaining()];
+    buffer.get(bytes);
+    buffer.position(buffer.position() - bytes.length);
+    return bytes;
+  }
+
+  private List<ByteBuffer> convertToListByteBuffer(List<byte[]> listByteArr) {
+    return listByteArr.stream().map(ByteBuffer::wrap).collect(Collectors.toList());
+  }
+}
--- a/sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/dynamodb/DynamoDBIO.java
+++ b/sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/dynamodb/DynamoDBIO.java
@@ -1,0 +1,535 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.beam.sdk.io.aws.dynamodb;
+
+import static org.apache.beam.vendor.guava.v20_0.com.google.common.base.Preconditions.checkArgument;
+
+import com.amazonaws.regions.Regions;
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDB;
+import com.amazonaws.services.dynamodbv2.model.AmazonDynamoDBException;
+import com.amazonaws.services.dynamodbv2.model.AttributeValue;
+import com.amazonaws.services.dynamodbv2.model.BatchWriteItemRequest;
+import com.amazonaws.services.dynamodbv2.model.ScanRequest;
+import com.amazonaws.services.dynamodbv2.model.ScanResult;
+import com.amazonaws.services.dynamodbv2.model.WriteRequest;
+import com.google.auto.value.AutoValue;
+import java.io.IOException;
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import javax.annotation.Nullable;
+import org.apache.beam.sdk.annotations.Experimental;
+import org.apache.beam.sdk.coders.Coder;
+import org.apache.beam.sdk.coders.ListCoder;
+import org.apache.beam.sdk.coders.MapCoder;
+import org.apache.beam.sdk.coders.SerializableCoder;
+import org.apache.beam.sdk.coders.StringUtf8Coder;
+import org.apache.beam.sdk.metrics.Counter;
+import org.apache.beam.sdk.metrics.Metrics;
+import org.apache.beam.sdk.transforms.Create;
+import org.apache.beam.sdk.transforms.DoFn;
+import org.apache.beam.sdk.transforms.PTransform;
+import org.apache.beam.sdk.transforms.ParDo;
+import org.apache.beam.sdk.transforms.Reshuffle;
+import org.apache.beam.sdk.transforms.SerializableFunction;
+import org.apache.beam.sdk.util.BackOff;
+import org.apache.beam.sdk.util.BackOffUtils;
+import org.apache.beam.sdk.util.FluentBackoff;
+import org.apache.beam.sdk.util.Sleeper;
+import org.apache.beam.sdk.values.KV;
+import org.apache.beam.sdk.values.PBegin;
+import org.apache.beam.sdk.values.PCollection;
+import org.apache.beam.sdk.values.TypeDescriptor;
+import org.apache.beam.vendor.guava.v20_0.com.google.common.annotations.VisibleForTesting;
+import org.apache.beam.vendor.guava.v20_0.com.google.common.collect.ImmutableSet;
+import org.apache.http.HttpStatus;
+import org.joda.time.Duration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * {@link PTransform}s for writing to <a href="https://aws.amazon.com/dynamodb/">DynamoDB</a>.
+ *
+ * <h3>Writing to DynamoDB</h3>
+ *
+ * <p>Example usage:
+ *
+ * <pre>{@code
+ * PCollection<T> data = ...;
+ * data.apply(
+ *           DynamoDBIO.<WriteRequest>write()
+ *               .withWriteRequestMapperFn(
+ *                   (SerializableFunction<T, KV<String, WriteRequest>>)
+ *                       //Transforming your T data into KV<String, WriteRequest>
+ *                       t -> KV.of(tableName, writeRequest))
+ *               .withRetryConfiguration(
+ *                    DynamoDBIO.RetryConfiguration.create(5, Duration.standardMinutes(1)))
+ *               .withAwsClientsProvider(new BasicSnsProvider(accessKey, secretKey, region));
+ * }</pre>
+ *
+ * <p>As a client, you need to provide at least the following things:
+ *
+ * <ul>
+ *   <li>Retry configuration
+ *   <li>Specify AwsClientsProvider. You can pass on the default one BasicSnsProvider
+ *   <li>Mapper function with a table name to map or transform your object into KV<tableName,
+ *       writeRequest>
+ * </ul>
+ *
+ * <h3>Reading from DynamoDB</h3>
+ *
+ * <p>Example usage:
+ *
+ * <pre>{@code
+ * PCollection<List<Map<String, AttributeValue>>> output =
+ *     pipeline.apply(
+ *             DynamoDBIO.<List<Map<String, AttributeValue>>>read()
+ *                 .withAwsClientsProvider(new BasicSnsProvider(accessKey, secretKey, region))
+ *                 .withScanRequestFn(
+ *                     (SerializableFunction<Void, ScanRequest>)
+ *                         input -> new ScanRequest(tableName).withTotalSegments(1))
+ *                 .items());
+ * }</pre>
+ *
+ * <p>As a client, you need to provide at least the following things:
+ *
+ * <ul>
+ *   <li>Specify AwsClientsProvider. You can pass on the default one BasicSnsProvider
+ *   <li>ScanRequestFn, which you build a ScanRequest object with at least table name and total
+ *       number of segment. Note This number should base on the number of your workers
+ * </ul>
+ */
+@Experimental(Experimental.Kind.SOURCE_SINK)
+public final class DynamoDBIO {
+  public static <T> Read<T> read() {
+    return new AutoValue_DynamoDBIO_Read.Builder().build();
+  }
+
+  public static <T> Write<T> write() {
+    return new AutoValue_DynamoDBIO_Write.Builder().build();
+  }
+
+  /** Read data from DynamoDB and return ScanResult. */
+  @AutoValue
+  public abstract static class Read<T> extends PTransform<PBegin, PCollection<T>> {
+    @Nullable
+    abstract AwsClientsProvider getAwsClientsProvider();
+
+    @Nullable
+    abstract SerializableFunction<Void, ScanRequest> getScanRequestFn();
+
+    @Nullable
+    abstract Integer getSegmentId();
+
+    @Nullable
+    abstract SerializableFunction<ScanResult, T> getScanResultMapperFn();
+
+    @Nullable
+    abstract Coder<T> getCoder();
+
+    abstract Builder<T> toBuilder();
+
+    @AutoValue.Builder
+    abstract static class Builder<T> {
+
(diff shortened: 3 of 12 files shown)


cmachgodaddy:
@lukecwik , @NathanHowell , @reuvenlax , @randomsamples, @jxlewis, @kennknowles Can you please review this PR?

jhalaria:
@iemejia - Please take a look.

cmachgodaddy:
retest this please

sdks/java/io/amazon-web-services/src/test/java/org/apache/beam/sdk/io/aws/dynamodb/AwsClientProviderMock.java
```diff
@@ -0,0 +1,48 @@
...
+package org.apache.beam.sdk.io.aws.dynamodb;
+
+import com.amazonaws.services.cloudwatch.AmazonCloudWatch;
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDB;
+import org.mockito.Mockito;
+
+/**
+ * Mocking AwsClientProvider instance so it can take i different {@link AmazonDynamoDB} object type.
+ */
+public class AwsClientProviderMock implements AwsClientsProvider {
```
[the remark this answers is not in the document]
cmachgodaddy: @iemejia , thanks for your comments. Working to address them

cmachgodaddy:
@chamikaramj , @lukecwik , @mxm , can you guys review this PR? Thanks

iemejia:
@cmachgodaddy You don't need everyone to review your PR. We commonly do reviews by one or two people, in particular people who maintain this area of the code, so in this case probably a second better candidate for reviewer would be @aromanenko-dev.

cmachgodaddy:
@iemejia and @aromanenko-dev , can you review it again? Have refactored it a bit.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/dynamodb/DynamoDBIO.java
```diff
@@ -0,0 +1,536 @@
...
+      abstract DynamoDBIO.RetryConfiguration.Builder setRetryPredicate(
+          RetryPredicate retryPredicate);
+
+      abstract DynamoDBIO.RetryConfiguration build();
+    }
+
+    /**
+     * An interface used to control if we retry the BatchWriteItemRequest call when a {@link
+     * Throwable} occurs. If {@link RetryPredicate#test(Object)} returns true, {@link Write} tries
+     * to resend the requests to the Solr server if the {@link RetryConfiguration} permits it.
```
aromanenko-dev: Please, change "Solr server" to "DynamoDB server"

buildSrc/src/main/groovy/org/apache/beam/gradle/Repositories.groovy
```diff
@@ -54,6 +54,9 @@ class Repositories {
 
       // Apache release snapshots
       maven { url "https://repository.apache.org/content/repositories/releases" }
+
+      // Mulesoft public repository
+      maven { url "https://repository.mulesoft.org/nexus/content/repositories/public" }
```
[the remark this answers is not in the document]
cmachgodaddy: That is the repos that host dynamodblocal for unit test. It's a bit confused, the artifact can be found here, https://mvnrepository.com/artifact/com.amazonaws/DynamoDBLocal/1.11.477, but it's actually hosted at mulesoft.

sdks/java/io/amazon-web-services/build.gradle
```diff
@@ -48,11 +51,24 @@ dependencies {
   shadowTest library.java.mockito_core
   shadowTest library.java.junit
   shadowTest library.java.slf4j_jdk14
+  testImplementation group: 'com.amazonaws', name: 'DynamoDBLocal', version:'1.11.477'
   shadowTest group: 'org.elasticmq', name: 'elasticmq-rest-sqs_2.12', version: '0.14.1'
   testRuntimeOnly project(path: ":beam-runners-direct-java")
 }
 
+// This is to copy native dependencies (sql-lite) for embedded dynamodb, for testing
+final copyNativeDeps = tasks.register('copyNativeDeps', Copy) {
```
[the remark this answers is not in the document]
cmachgodaddy: @iemejia and @aromanenko-dev , yes it's needed to copy native dll to the build/libs directory, so we can use danamolocaldb. Ref: https://stackoverflow.com/questions/34401039/run-dynamodb-local-as-part-of-a-gradle-java-project
But if you have different way of doing this, please let me know.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/dynamodb/AttributeValueCoder.java
```diff
@@ -0,0 +1,179 @@
...
+import org.apache.beam.sdk.coders.AtomicCoder;
+import org.apache.beam.sdk.coders.BooleanCoder;
+import org.apache.beam.sdk.coders.ByteArrayCoder;
+import org.apache.beam.sdk.coders.CoderException;
+import org.apache.beam.sdk.coders.ListCoder;
+import org.apache.beam.sdk.coders.MapCoder;
+import org.apache.beam.sdk.coders.StringUtf8Coder;
+
+/** Provide a mechanism to encode and decode AttributeValue object. */
+public class AttributeValueCoder extends AtomicCoder<AttributeValue> implements Serializable {
```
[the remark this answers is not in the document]
cmachgodaddy: Not quite to understand your question much. The `AttributeValueCoder` works without implementing Serializable interface. Next commit will remove it. The IO's reader and writer doesn't work without `AttributeValueCoder`. Let me know if I still didn't answer your question?

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/dynamodb/AttributeValueCoder.java
```diff
@@ -0,0 +1,179 @@
...
+    n, // for Number
+    b, // for Byte
+    sS, // for List of String
+    nS, // for List of Number
+    bS, // for List of Byte
+    m, // for Map of String and AttributeValue
+    l, // for list of AttributeValued
+    bOOL, // for Boolean
+    nULLValue, // for null
+  };
```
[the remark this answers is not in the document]
cmachgodaddy: Done!

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/dynamodb/AttributeValueCoder.java
```diff
@@ -0,0 +1,179 @@
...
+
+  private static final AttributeValueCoder INSTANCE = new AttributeValueCoder();
+
+  private static final StringUtf8Coder STRING_CODER = StringUtf8Coder.of();
```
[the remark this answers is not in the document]
cmachgodaddy: I replaced all primitive type coding by `StringUtf8Coder.of().encode()`, but for List and Map I left them so we can easily to read the code. But feel free to let me know if you still want to use that style? (not a big deal)

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/dynamodb/AttributeValueCoder.java
```diff
@@ -0,0 +1,179 @@
...
+        break;
+      default:
+        throw new CoderException("Unknown Type");
+    }
+
+    return attrValue;
+  }
+
+  private List<byte[]> convertToListByteArray(List<ByteBuffer> listByteBuffer) {
+    List<byte[]> listByteArray = new ArrayList<>();
```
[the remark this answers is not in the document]
cmachgodaddy: Done!

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/dynamodb/AttributeValueCoder.java
```diff
@@ -0,0 +1,179 @@
...
+
+  private byte[] convertToByteArray(ByteBuffer buffer) {
+    byte[] bytes = new byte[buffer.remaining()];
+    buffer.get(bytes);
+    buffer.position(buffer.position() - bytes.length);
+    return bytes;
+  }
+
+  private List<ByteBuffer> convertToListByteBuffer(List<byte[]> listByteArr) {
+    List<ByteBuffer> byteBufferList = new ArrayList<>();
```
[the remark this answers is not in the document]
cmachgodaddy: Done! Coming commit will replace it.

buildSrc/src/main/groovy/org/apache/beam/gradle/Repositories.groovy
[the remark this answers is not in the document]
cmachgodaddy: Adding more info .... The jar file is published by AWS for testing read/write against dynamo locally. This dynamolocal also is related to your question below, about the `copyNativeDeps` gradle task, which help setting up dynamodb. 
I can see we have 3 options here for unit test:
1. using this jar
2. mock the db
3. using docker
What would you guys recommend?

cmachgodaddy:
@iemejia , just refactor it again to add withScanRequestFn as you suggested above. Would you pls review? Regarding the ParDo for Read, I love this idea, but prefer to do it in a separate PR.

cmachgodaddy:
@iemejia , @aromanenko-dev , ping again :-)

aromanenko-dev:
Run Java PreCommit

aromanenko-dev:

aromanenko-dev:

aromanenko-dev:
@cmachgodaddy Thank you for addressing the review comments. In general, it's looks fine for me but let's wait for @iemejia response.
In the same time, could you rebase (to avoid merging conflicts) and squash all commits into one?

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/dynamodb/AttributeValueCoder.java
```diff
@@ -0,0 +1,166 @@
...
+  /** Data type of each value type in AttributeValue object. */
+  private enum AttributeValueType {
+    s, // for String
+    n, // for Number
+    b, // for Byte
+    sS, // for List of String
+    nS, // for List of Number
+    bS, // for List of Byte
+    m, // for Map of String and AttributeValue
+    l, // for list of AttributeValued
```
lukecwik:
```suggestion
    l, // for list of AttributeValue
```

cmachgodaddy: Good catch @lukecwik . And thank for your review!

iemejia:
Hi sorry @cmachgodaddy got trapped on other stuff, will do second ASAP. I am aware of this PR and have it in my TODO. Sorry again.

sdks/java/io/amazon-web-services/src/test/java/org/apache/beam/sdk/io/aws/dynamodb/DynamoDBConfigMock.java
```diff
@@ -0,0 +1,59 @@
...
+ * limitations under the License.
+ */
+package org.apache.beam.sdk.io.aws.dynamodb;
+
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDB;
+import javax.annotation.Nullable;
+import org.apache.beam.sdk.options.ValueProvider;
+
+/** Mock DynamoDBConfiguration to pass it over the write to writer for testing Writer's retries. */
+public class DynamoDBConfigMock extends DynamoDBIO.DynamoDBConfiguration {
```
iemejia: Can you please remove this class and do the configuration of the AmazonDynamoDB class by following an approach like the one on KinesisIO: Two methods: `withAWSClientsProvider(AWSClientsProvider awsClientsProvider)` and `withAWSClientsProvider(
        String awsAccessKey, String awsSecretKey, Regions region, String serviceEndpoint)`. `AWSClientsProvider` can be like this:

```java
public interface AWSClientsProvider extends Serializable {
  AmazonDynamoDB getAmazonDynamoDB();
}
```

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/dynamodb/DynamoDBIO.java
```diff
@@ -0,0 +1,501 @@
...
+      abstract Builder setScanRequestFn(SerializableFunction<Void, ScanRequest> fn);
+
+      abstract Read build();
+    }
+
+    public Read withDynamoDBConfiguration(DynamoDBConfiguration dynamoDBConfiguration) {
+      return toBuilder().setDynamoDBConfiguration(dynamoDBConfiguration).build();
+    }
+
+    public Read withScanRequestFn(SerializableFunction<Void, ScanRequest> fn) {
```
iemejia: Given that ScanRequest is Serializable what is the reason to not pass this directly (I mean without wrapping it in the SerializableFunction). Can you change it to do it directly if there are no constraints to do so?

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/dynamodb/DynamoDBIO.java
```diff
@@ -0,0 +1,501 @@
...
+                new BasicAWSCredentials(getAwsAccessKey().get(), getAwsSecretKey().get())));
+      }
+      return builder.build();
+    }
+  }
+
+  /** Read data from DynamoDB and return PCollection<Map<String, AttributeValue>>. */
+  @AutoValue
+  public abstract static class Read
+      extends PTransform<PBegin, PCollection<Map<String, AttributeValue>>> {
```
iemejia: Any reason to not return here better `ScanResult` ? we can if needed provide an extra transformation to extract those into `Map<String, AttributeValue>` but the more generic we can let the API the better.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/dynamodb/DynamoDBIO.java
```diff
@@ -0,0 +1,501 @@
...
+            || (throwable instanceof AmazonDynamoDBException
+                && ELIGIBLE_CODES.contains(((AmazonDynamoDBException) throwable).getStatusCode())));
+      }
+    }
+  }
+
+  /** Implementation of {@link #write}. */
+  @AutoValue
+  public abstract static class Write
+      extends PTransform<PCollection<BatchWriteItemRequest>, PCollectionTuple> {
```
iemejia: Why this returns a `PCollectionTuple`? to have multiple outputs?  Is this really needed? Can we achieve the same by returning `PCollection<Void>`?

cmachgodaddy: This is to make it consistent with SNSIO. Of course, I can try with Void.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/dynamodb/DynamoDBIO.java
cmachgodaddy: This is to make it consistent with other IO e.i. Kinesis, SNS, SQS, ... They are all return one item/record at a time. In Dynamo, a Map represent for just one record, which its content is a list of attributes/columns. Let me know if we still want ScanResult, instead?

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/dynamodb/DynamoDBIO.java
cmachgodaddy: Yes, we can try that. I though you meant to implement the withScanRequestFn in lambda style, so ...

sdks/java/io/amazon-web-services/src/test/java/org/apache/beam/sdk/io/aws/dynamodb/DynamoDBConfigMock.java
cmachgodaddy: Sure, will get back that AWSClientProvider. But this Configuration approach is copied from JDBC IO one

iemejia: Yes other IOs do it like that but for Amazon is probably a good idea to keep it similar, thanks!
A future idea we have is to make it configurable via options because users mostly already will have `AwsOptions` around, but do not do this for this PR we have to discuss this first in the mailing list.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/dynamodb/DynamoDBIO.java
iemejia: Mmm no maybe that was a confusion, the most straight forward the API is the better.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/dynamodb/DynamoDBIO.java
iemejia: Yes, it is probably better to let `ScanResult` so please change it, that way users can decide what to do with it and we will solve issues of missing features if Amazon decides to add more stuff to it which we will have to maintain otherwise.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/dynamodb/DynamoDBIO.java
iemejia: Oh wow I think we missed that one in the review of SnsIO. Now I have to check on why it is like that, thanks for the explanation. And thanks for being receptive for the  change suggestions.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/dynamodb/DynamoDBIO.java
cmachgodaddy: @iemejia , I understand your point, and I am trying to make it return ScanResult. But, this is not an easy change at all, and seem to not align with the existing IOs model. First, we have to `extends PTransform<PBegin, PCollection<ScanResult>>` , and the ScanResult already contain a list of item. This is like a collection contains a collection, but the outsider one always have one item (note: can't do `extends PTransform<PBegin, ScanResult>` directly without PCollection). Second, if we  return ScanResult, what would we do with the Reader's getCurrent() and get advance()? since they are supposed to return one item at time from ScanResult? Any idea of how it would be done with ScanResult?

randomsamples: @iemejia I think we are a little confused about the request to return ScanResult b/c that type represents a collection of results so it would seem natural to "unroll" the results and pass them on to downstream transforms for individual processing. If the goal is to allow more control on how consumers handle ScanResults, what do you think about changing DynamoDBIO API to include something similar to JdbcIO's RowMapper class, but with signature f: ScanResult -> [T]?

randomsamples: Maybe this not too different from what you were referring to here @iemejia?
> we can if needed provide an extra transformation to extract those into Map<String, AttributeValue>

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/dynamodb/DynamoDBIO.java
cmachgodaddy: @iemejia , have tried sending ScanReqeust directly, but doesn't work. The ScanReqeust seem to be serializable, but it also extends `com.amazonaws.AmazonWebServiceRequest`, which is not serializable, and AmazonWebServiceRequest contains quite a lot of objects/properties that are not serializable. When testing passing ScanRequest directly, it fails most of unit tests that have more than one split (or totalSegment > 1). Any idea of how it could be done beside `SerializableFunction<Void, ScanRequest>` ?

iemejia: No, that was what I was afraid of, ok so let's let it like it is then. Thanks.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/dynamodb/DynamoDBIO.java
iemejia: Yes that's exactly the idea @randomsamples. I am trying to do the default to be as generic and  closer to the Amazon API as possible. That way the users can decide what to do with such objects afterwards which guarantees (1) less maintenance from our side since woe already cover the generic case and (2) that we don't need to change our API if Amazon adds something new to their default Result object to support it.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/dynamodb/DynamoDBIO.java
cmachgodaddy: @iemejia , I have changed it to use `PCollection<Void>`, but couldn't find a way to return multiple outputs. Something similar to this JDBC one, https://github.com/apache/beam/blob/master/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcIO.java#L842, it also return just one output if using `PCollection<Void>`. Any idea of how this would be done?

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/dynamodb/DynamoDBIO.java
iemejia: @cmachgodaddyThe IO model is to output elements from requests, in many cases these are selfcontained records where the mapping is done 1 to 1, but there are already IOs that follow other  patterns e.g. `HBaseIO` does the same and it returns a `Result` that contains column families and columns that the user can decide to iterate in further transforms and produce more outputs (user can decide if it uses it more as a FlatMap than as a Map).

> note: can't do extends PTransform<PBegin, ScanResult> directly without Pcollection.

Yes you are right, maybe I was not clear in my first message. I intended to say `extends PTransform<PBegin, PCollection<ScanResult>>`

> Second, if we return ScanResult, what would we do with the Reader's getCurrent() and getAdvance()? since they are supposed to return one item at time from ScanResult? Any idea of how it would be done with ScanResult?

We can just add a facility  to extract the values, but we may not need to, the user can do that in its own DoFn logic.

One extra advantage of this pattern is that it will allow us to do in the future a transform of the ReadAll kind, e.g. `PTransform<PCollection<ScanRequest>, Pcollection<ScanResult>>` and with this we may do queries to multiples tables in the same pipeline.

cmachgodaddy: Thanks for your further explanation, @iemejia , I am trying this pattern with pointers that @randomsamples shared above.

cmachgodaddy: Hi @iemejia and @randomsamples, I have a closer look at HBaseIO, JdbcIO, and even CassandraIO while trying to refactor this DynamoIO base on this comment, to return ScanResult directly to user, but then realized we can't simply just return ScanResult. Here are the reasons and what I found: 
1. HBaseIO's Reader return Result, which represent for **one** row, but ScanResult in Dynamo represent for a list of row, `ScanResult.getItems()` , which returns List<Map<String, AttributeValue>>
Ref to HBase Reader: https://github.com/apache/beam/blob/master/sdks/java/io/hbase/src/main/java/org/apache/beam/sdk/io/hbase/HBaseIO.java#L460
2. JdbcIO's Reader use RowMapper to send result type of T to user, but each "result" it send is one row, not a whole set of row or a list of row. (While ScanResult in Dynamo is a set of rows, again) https://github.com/apache/beam/blob/master/sdks/java/io/jdbc/src/test/java/org/apache/beam/sdk/io/jdbc/JdbcTestHelper.java#L34 . This pointer is more clear, which create a row from the "result" object sent from Reader.
3. Similar for Cassandra....
With that being said, the ScanResult is incomparable with other "result" objects sent from other Readers to user. Second, all of those Readers iterate through each row and send to user, so if in DynamoIO, we send the ScanResult object to user directly, we don't iterate through each row, and which we will lose the ability to do parallel scan in Dynamo and split in Beam.
Hope I explained it well this one. And please let me know if I missed anything? or if you have any idea?

cmachgodaddy: @iemejia ?

iemejia: Sorry for late answer. I am trying to understand why this is different, given that each partition is being queried by one individual `ScanRequest` that assigns an independent segment, so each read of a 'partition' ends up being the responsibility of an individual worker, so doing this won't affect parallelism, and make the mapping more straight forward if we want to move it in the future to a ParDo based implementation like those IOs that use the `ReadAll` approach.

I know that this sounds contra intuitive because of the 1 to 1 mapping instead of 1 to n but it really does not make a big difference.

Only possible issue I see is that if there is an unbalanced partition it can make the read slow, but this will be the case if we 'extract' the data from the object so in the end it will be similar..

iemejia: Sorry for late answer. I am trying to understand why this is different, given that each partition is being queried by one individual `ScanRequest` that assigns an independent segment, so each read of a 'partition' ends up being the responsibility of an individual worker, so doing this won't affect parallelism, and make the mapping more straight forward if we want to move it in the future to a ParDo based implementation like those IOs that use the `ReadAll` approach.

I know that this sounds contra intuitive because of the 1 to 1 mapping instead of 1 to n but it really does not make a big difference.

Only possible issue I see is that if there is an unbalanced partition it can make the read slow, but this will be the case if we 'extract' the data from the object so in the end it will be similar..

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/dynamodb/DynamoDBIO.java
iemejia: Let me look into this and tell you, I don't have a use case at hand but worth checking.

iemejia: Ah wait I have misread this. What you mean with multiple outputs, can't we just use a `PCollectionTuple` and connect each output PCollection to the different outputting IOs?

cmachgodaddy: Hi @iemejia , thanks for your response. As we discussed, the `PCollectionTuple` is unnecessary, since we just need to return one output. But implementing with PCollection<Void> will not give us a option to turn any output. I have put together my internal PR for your convenience to review the implementation of PCollection<BatchWriteItemResult> for the Writer, if you don't mind? https://github.com/cmachgodaddy/beam/pull/1/files#diff-127519743356d99d2d217126be4f240eR316

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/dynamodb/DynamoDBIO.java
cmachgodaddy: Thanks for the response again, @iemejia . I good to go with the ScanResult approach, just can't picture how client/user will consume it? I have `stage` implementation of it in my internal repo PR, https://github.com/cmachgodaddy/beam/pull/1/files#diff-127519743356d99d2d217126be4f240eR233
. But again I'm not quite sure know how user is going to implement the RowMapper interface and return one item or a list of item?
https://github.com/cmachgodaddy/beam/pull/1/files#diff-392c3e0413082935785b6e8f8864a3e2R179
I am getting stuck right here actually in my test case, https://github.com/cmachgodaddy/beam/pull/1/files#diff-fc5ff3bb6a92b5649de50d3ef3fb5a13R74 
Look forward to your advice?

cmachgodaddy: @iemejia , I have tried several times to implement the Reader to return ScanResult with RowMapper (like JDBC's one, as suggested above), but still couldn't get the unit test work. Would you mind to take a look at my repo?

Here is the RowMapper: https://github.com/cmachgodaddy/beam/blob/ScanResultApproach/sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/dynamodb/DynamoDBIO.java#L119

And this is my ReadFn which call the RowMapper and pass the ScanResult to user:
https://github.com/cmachgodaddy/beam/blob/ScanResultApproach/sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/dynamodb/DynamoDBIO.java#L214

And, here is the implementation of the RowMapper from user/client side:
https://github.com/cmachgodaddy/beam/blob/ScanResultApproach/sdks/java/io/amazon-web-services/src/test/java/org/apache/beam/sdk/io/aws/dynamodb/DynamoDBIOTestHelper.java#L177

The question is should user/client return ScanResult directly or a List<Items> which attracted from ScanResult? Neither ways, unit testing both cases are getting stuck.
https://github.com/cmachgodaddy/beam/blob/ScanResultApproach/sdks/java/io/amazon-web-services/src/test/java/org/apache/beam/sdk/io/aws/dynamodb/DynamoDBIOTest.java#L74

I am not sure if you understand the issue I am dealing with this ScanResult approach? I would love to jump on a call to explain more and answer any questions that you have. Please let me know when you're available for a chat? Or if you have any idea to speed up this PR since it's has been out there for more than a couple of months?

iemejia: Yes, I  agree with you, this PR has taken way too long. Please excuse me if my delayed review has contributed to that. However I am a bit busy this weekend, so I will take a look and give you feedback on monday.

The next release of Beam (2.14.0) will be the 19/06 so if you want we can try to finish it before that date. I will mark the Jira to be included in next release and we’ll see if we can make it.

If you prefer to discuss more interactively you can ping me at the [ASF slack](http://the-asf.slack.com/) You can [request an invite here](https://s.apache.org/slack-invite). Notice that I am based in Europe so consider the timezone difference for this.

cmachgodaddy: @iemejia , let try to get it in 19/06. Can you review my last commit? and the issue that I have with ScanResult in my repo?

iemejia: Is there any technical reason that impeaches returning `ScanResult`? I already explained above my reasons to prefer this, summarizing:

1. I prefer to keep the AWS object available (in case they expand their API).
2. Parallelism is not affected because the user will have one ScanResult per worker so he can just extract and iterate the values.

I understand that (2) may seem less user-friendly, but it is in order to be consistent with (1). It is mostly about being 'independent' of the evolution of the Amazon Dynamo API and of future requests from users around things not available because of not exposing the `ScanResult` API.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/dynamodb/DynamoDBIO.java
iemejia: Since this is a sink you may not really care about multiple outputs in the return type of the `Write` definition. You can achieve multiple outputs by creating subgraphs in the pipeline with `PCollectionTuple` and then chaining to each outputting path the write transform.

We could have decided to make the signature `PTransform<PCollection<BatchWriteItemRequest>, PDone>` too, but I asked explictly to use the `PCollection<Void>` as a return type to enable in the future the possibility to have patterns like waiting for a PCollection write before another, that's the reason why  (sorry if I was not explicit about this).

I still have to check `SnsIO`  to see why they took that approach but I would not be surprised if we can achieve the same without needing to return a `PCollectionTuple` in the Write.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/dynamodb/DynamoDBIO.java
cmachgodaddy: @iemejia , of course, I love to go with your approach but, yes there is a technical channel to implement the Reader with ScanResult. Would you  mind to look at my explanation and my repo above, for more details?

iemejia:
@cmachgodaddy I think I answered all the pending questions. Please write me if you still have questions or comments.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/dynamodb/DynamoDBIO.java
cmachgodaddy: @iemejia , I think we are not in synced. Of course, I understand your point. But, like I explained above, using PCollection<Void> or PDone as return type, will not give us any way to return any result to user. 

> @iemejia , I have changed it to use PCollection<Void>, but couldn't find a way to return multiple outputs. Something similar to this JDBC one, https://github.com/apache/beam/blob/master/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcIO.java#L842, it also return just one output if using PCollection<Void>. Any idea of how this would be done?

cmachgodaddy:
@iemejia , I put together my implementation for the Reader with ScanResult (as return type) in the last commit, do you mind to take a look? However the test case of the Reader is failing, so I would love to have your instructions to get it work. And regarding the Writer with PCollection<Void> as return type as you requested, I still couldn't get it work (the reasons I explained above). Can you give me some instructions as well?

iemejia:
Merged manually to do some minor fixes.

PR #8394, opened 2019-04-24
[BEAM-2857] Implementing WriteToFiles transform for fileio (Python)

pabloem:
r: @chamikaramj

### DIFF
--- a/sdks/python/apache_beam/io/fileio.py
+++ b/sdks/python/apache_beam/io/fileio.py
@@ -30,6 +30,11 @@ The transforms in this file include ``WriteToFiles``, which allows you to write
 a ``beam.PCollection`` to files, and gives you many options to customize how to
 do this.
 
+The ``WriteToFiles`` transform supports bounded and unbounded PCollections
+(i.e. it can be used both batch and streaming pipelines). For streaming
+pipelines, it currently does not have support for multiple trigger firings
+on the same window.
+
 File Naming
 -----------
 One of the parameters received by ``WriteToFiles`` is a function specifying how
@@ -315,7 +320,7 @@ def destination_prefix_naming():
       kwargs['start'] = window.start.to_utc_datetime().isoformat()
       kwargs['end'] = window.end.to_utc_datetime().isoformat()
 
-    # TODO(pabloem): Add support for PaneInfo
+    # TODO(BEAM-3759): Add support for PaneInfo
     # If the PANE is the ONLY firing in the window, we don't add it.
     #if pane and not (pane.is_first and pane.is_last):
     #  kwargs['pane'] = pane.index
@@ -509,8 +514,6 @@ class WriteToFiles(beam.PTransform):
 
 
 def _create_writer(base_path, writer_key):
-  # TODO(pabloem) Is this a good place to create a temporary directory?
-
   try:
     filesystems.FileSystems.mkdirs(base_path)
   except IOError:
--- a/sdks/python/apache_beam/io/fileio.py
+++ b/sdks/python/apache_beam/io/fileio.py
@@ -23,15 +23,73 @@ and its metadata; and ``ReadMatches``, which takes in a ``PCollection`` of file
 metadata records, and produces a ``PCollection`` of ``ReadableFile`` objects.
 These transforms currently do not support splitting by themselves.
 
+Writing to Files
+================
+
+The transforms in this file include ``WriteToFiles``, which allows you to write
+a ``beam.PCollection`` to files, and gives you many options to customize how to
+do this.
+
+File Naming
+-----------
+One of the parameters received by ``WriteToFiles`` is a function specifying how
+to name the files that are written. This is a function that takes in the
+following parameters:
+
+- window
+- pane
+- shard_index
+- total_shards
+- compression
+- destination
+
+It should return a file name that is unique for a combination of these
+parameters.
+
+The default naming strategy is to name files
+in the format
+`$prefix-$start-$end-$pane-$shard-of-$numShards$suffix$compressionSuffix`,
+where:
+
+- `$prefix` is, by default, `"output"`.
+- `$start` and `$end` are the boundaries of the window for the data being
+  written. These are omitted if we're using the Global window.
+- `$pane` is the index for the number of firing for a window.
+- `$shard` and `$numShards` are the current shard number, and the total number
+  of shards for this window firing.
+- `$suffix` is, by default, an empty string, but it can be set by the user via
+  ``default_file_naming``.
+
+Dynamic Destinations
+--------------------
+If the elements in the input ``beam.PCollection`` can be partitioned into groups
+that should be treated differently (e.g. some events are to be stored as CSV,
+while some others are to be stored as Avro files), it is possible to do this
+by passing a `destination` parameter to ``WriteToFiles``. Something like the
+following::
+
+    my_pcollection | beam.io.fileio.WriteToFiles(
+          path='/my/file/path',
+          destination=lambda record: 'avro' if record['type'] == 'A' else 'csv',
+          sink=lambda dest: AvroSink() if dest == 'avro' else CsvSink(),
+          file_naming=beam.io.fileio.destination_prefix_naming())
+
+In this transform, depending on the type of a record, it will be written down to
+a destination named `'avro'`, or `'csv'`. The value returned by the
+`destination` call is then passed to the `sink` call, to determine what sort of
+ sink will be used for each destination. The return type of the `destination`
+parameter can be anything, as long as elements can be grouped by it.
+
+
 No backward compatibility guarantees. Everything in this module is experimental.
 """
 
 from __future__ import absolute_import
 
+import collections
 import logging
 import random
 import uuid
-from collections import defaultdict
 
 from past.builtins import unicode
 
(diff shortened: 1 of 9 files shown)


pabloem:
Run Python PostCommit

pabloem:
Run Python PreCommit

pabloem:
Passing postcommit: https://builds.apache.org/job/beam_PostCommit_Python_Verify_PR/593/
PAssing py3 postcommit: https://builds.apache.org/job/beam_PostCommit_Python3_Verify_PR/230/

udim:
still reviewing...

pabloem:
ugh... I'm so sorry Udi. The design doc for this is here: https://s.apache.org/fileio-beam-python
it's light on details.. but i guess... it would be helpful : /

sdks/python/apache_beam/transforms/core.py
```diff
@@ -1982,7 +1983,11 @@ def process(self, element, timestamp=DoFn.TimestampParam,
...
 
-  def __init__(self, windowfn, **kwargs):
+  def __init__(self,
+               windowfn,
+               trigger=None,
+               accumulation_mode=None,
+               timestamp_combiner=None):
     """Initializes a WindowInto transform.
 
     Args:
```
udim: Could you add the allowable types for `windowfn` (Windowing and WindowFn?).

sdks/python/apache_beam/testing/util.py
```diff
@@ -141,6 +145,21 @@ def _equal(actual):
   return _equal
 
 
+def matches_all(expected):
```
udim: Is this different from `contains_in_any_order` above?

pabloem: The `contains_in_any_order` seems to require a set of hashable items. It simply does `collections.Counter()` on the iterable, and compares both counters. `matches_all`  takes in an iterable of elements or matchers, and matches them 1:1 properly.

sdks/python/apache_beam/io/fileio.py
```diff
@@ -169,3 +179,476 @@ def __init__(self, compression=None, skip_directories=True):
...
+    if not self._temp_directory:
+      temp_location = (
+          p.options.view_as(GoogleCloudOptions).temp_location
+          or self.path.get())
+      dir_uid = str(uuid.uuid4())
+      self._temp_directory = StaticValueProvider(
+          str,
+          filesystems.FileSystems.join(temp_location,
+                                       '.temp%s' % dir_uid))
+      logging.info('Added temporary directory %s', self._temp_directory.get())
```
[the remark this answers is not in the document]
pabloem: No. The transform may be run in streaming, so I don't think there's a good moment in the pipeline execution to perform a cleanup of the temp directory.

pabloem:
Hi Udi. I've addressed some of your comments. Some things pending to finalize:

- How do we clean up temporary files?
- In general, how do we deal with failures?

Some things that are worth looking at:

- I added documentation at the top of the file. LMK if that clarifies some things.
- Default sink is now a text sink that encodes strings to UTF8 encoding.

sdks/python/apache_beam/testing/util.py
udim: gotcha

pabloem:
Run Python PreCommit

pabloem:
I've updated the cleanup code. Any error other than `BeamIOError` should be a retry. Also, I believe it should be okay to only clean up all of the files, and not the directory (as we'd have to check for batch vs streaming).
Thoughts?

udim:
From what I understand deletion of temporary files is best effort. There is currently no mechanism to clean up old temporary files left over due to an error. Perhaps we could, in the future, add a TTL or some other way to tell if a temporary file can be safely deleted.

sdks/python/apache_beam/io/fileio.py
```diff
@@ -554,6 +561,18 @@ def process(self, element):
                        r.pane,
                        destination)
 
+    logging.info('Cautiously removing temporary files for destination %s',
+                 destination)
+    self._remove_temporary_files([r.file_name for r in file_results])
```
udim: There's an implicit `return None` at the end of this method. Are `None`s handled correctly?

udim:
Looks good, except for the new fileio.py comment I left above.

sdks/python/apache_beam/io/fileio.py
pabloem: I believe this method will return a generator due to the above yield keyword, so this does not return None.

pabloem:
> From what I understand deletion of temporary files is best effort. There is currently no mechanism to clean up old temporary files left over due to an error. Perhaps we could, in the future, add a TTL or some other way to tell if a temporary file can be safely deleted.

Do you mean when the pipeline fails? (e.g. for batch?)

You're right about that, that if the pipeline fails, we won't be able to clean up the bundles that did not run. But how could we know this ahead of time? : /

Currently, all files will be cleaned up for any bundle that succeeds. If the bundle does not succeed, then we should not clean up the files.

pabloem:

udim:
Yes, it applies to cases where the pipeline fails, but could also be in cases where it is cancelled or updated (does the temp directory change in that case?).

We could format temp directory names to have the job ID in them, so it's possible to check if the job is still running and delete temp files if it isn't. This process would not be run as part of the pipeline.
Regarding temp directory deletion, I believe that for batch jobs they should be removed at the end (because it's better to clean up if we can). For streaming jobs, there's no hook for cleanup during cancellation and updates AFAIK.

Regarding failed bundles, assuming that the pipeline has not failed, it should be possible to scan for and delete any remaining temporary files.

udim:
cc: @chamikaramj as I'm OOO next week.

pabloem:
@chamikaramj

sdks/python/apache_beam/io/fileio.py
```diff
@@ -502,12 +502,13 @@ def expand(self, pcoll):
...
-                            self.path, self.file_naming_fn)))
+                            self.path, self.file_naming_fn,
+                            self._temp_directory)))
 
     return file_results
 
 
-def _create_writer(base_path, file_name):
+def _create_writer(base_path, writer_key):
   # TODO(pabloem) Is this a good place to create a temporary directory?
```
chamikaramj: Please create a JIRA or remove this.

sdks/python/apache_beam/io/fileio.py
```diff
@@ -562,16 +569,21 @@ def process(self, element):
...
+  def _remove_temporary_files(self, writer_key):
     try:
-      filesystems.FileSystems.delete(files)
+      prefix = filesystems.FileSystems.join(
+          self.temporary_directory.get(), str(abs(hash(writer_key))))
+      match_result = filesystems.FileSystems.match(['%s*' % prefix])
+      orphaned_files = [m.path for m in match_result[0].metadata_list]
+
+      logging.debug('Deleting orphaned files: %s', orphaned_files)
+      filesystems.FileSystems.delete(orphaned_files)
```
chamikaramj: Shouldn't we be ignoring missing source files to handle workitem retries ?

sdks/python/apache_beam/io/fileio.py
```diff
@@ -170,3 +237,487 @@ def __init__(self, compression=None, skip_directories=True):
...
+              'compression': ''}
+    if total_shards is not None and shard_index is not None:
+      kwargs['shard'] = int(shard_index)
+      kwargs['total_shards'] = int(total_shards)
+
+    if window != GlobalWindow():
+      kwargs['start'] = window.start.to_utc_datetime().isoformat()
+      kwargs['end'] = window.end.to_utc_datetime().isoformat()
+
+    # TODO(pabloem): Add support for PaneInfo
```
chamikaramj: Create a JIRA and mention here ?

sdks/python/apache_beam/io/fileio.py
```diff
@@ -23,17 +23,83 @@
...
+================
+
+The transforms in this file include ``WriteToFiles``, which allows you to write
+a ``beam.PCollection`` to files, and gives you many options to customize how to
+do this.
+
+File Naming
+-----------
+One of the parameters received by ``WriteToFiles`` is a function specifying how
+to name the files that are written. This is a function that takes in the
```
chamikaramj: Also document that this is supported for streaming and how it is supported.

sdks/python/apache_beam/io/fileio.py
udim: I agree that we should ignore errors here. Note that `orphaned_files` are taken from a recent `filesystems.FileSystems.match()` call, so it doesn't make a difference if this is a retry.

sdks/python/apache_beam/io/fileio.py
pabloem: Done.

sdks/python/apache_beam/io/fileio.py
pabloem: Yes. All we're doing is logging the failures (in debug level) to be aware of them.

sdks/python/apache_beam/io/fileio.py
pabloem: Done

sdks/python/apache_beam/io/fileio.py
pabloem: Done.

chamikaramj:
LGTM. Thanks.

Please fix-up and self-merge after tests pass.

pabloem:

chamikaramj:
Also, please fix the conflict.

pabloem:
Run Portable_Python PreCommit

pabloem:
Run Python PreCommit

yupbank:
Hey, So following the doc for which `AvroSink`  clearly doesn't exist and the `beam.io.avroio._AvroSink` would require much more info. For the context, i'm looking into ways to do `WriteToParquet` from a pubsub input. and `WriteToFiles` provides the closest thing. But the document is a bit confusing.

PR #8396, opened 2019-04-24
[BEAM-7138] keep Java serialized coder in wire coder construction

ihji:
keep Java serialized coder in wire coder construction. otherwise, we can't use `SerializableCoder` as a wire coder and it will possibly cause an issue when read transform is executed.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/wire/LengthPrefixUnknownCoders.java
+++ b/runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/wire/LengthPrefixUnknownCoders.java
@@ -22,6 +22,7 @@ import java.util.function.Predicate;
 import org.apache.beam.model.pipeline.v1.RunnerApi;
 import org.apache.beam.model.pipeline.v1.RunnerApi.Coder;
 import org.apache.beam.model.pipeline.v1.RunnerApi.Components;
+import org.apache.beam.runners.core.construction.CoderTranslation;
 import org.apache.beam.runners.core.construction.ModelCoders;
 import org.apache.beam.sdk.coders.ByteArrayCoder;
 import org.apache.beam.sdk.coders.LengthPrefixCoder;
@@ -55,12 +56,16 @@ public class LengthPrefixUnknownCoders {
     //     request to replace the coder with a byte array coder.
     //  2) the requested coder is a known coder but not a length prefix coder. In this case we
     //     rebuild the coder by recursively length prefixing any unknown component coders.
-    //  3) the requested coder is an unknown coder. In this case we either wrap the requested coder
+    //  3) the requested coder is a serialized Java coder. In this case we wrap the coder with
+    //     a length prefix coder.
+    //  4) the requested coder is an unknown coder. In this case we either wrap the requested coder
     //     with a length prefix coder or replace it with a length prefix byte array coder.
     if (ModelCoders.LENGTH_PREFIX_CODER_URN.equals(urn)) {
       return replaceWithByteArrayCoder ? lengthPrefixedByteArrayCoderId : coderId;
     } else if (ModelCoders.urns().contains(urn)) {
       return addForModelCoder(coderId, components, replaceWithByteArrayCoder);
+    } else if (CoderTranslation.JAVA_SERIALIZED_CODER_URN.equals(urn)) {
+      return addWrappedWithLengthPrefixCoder(coderId, components);
     } else {
       return replaceWithByteArrayCoder
           ? lengthPrefixedByteArrayCoderId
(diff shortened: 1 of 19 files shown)


ihji:
R: @mxm

mxm:
Actually, just came across this as well when implementing TestStream for PVR. We need to look into why the Flink tests fail here.

ihji:
@mxm I encountered the `ClassCastException` when I tried to use `.withMaxNumRecords` in `KafkaIO`. It generates a `PCollection` of `AutoValue_BoundedReadFromUnboundedSource_Shard` and current implementation of WireCoder doesn't support Java serialized coder which this `PCollection` needs to use.

ihji:
I'll look into what's wrong in the test.

mxm:
Thanks for explaining. Good find. Looks like this breaks the Dataflow tests now.

mxm:
Any updates for this? I would like to use this for the TestStream PR.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/wire/LengthPrefixUnknownCoders.java
```diff
@@ -55,12 +56,16 @@ public static String addLengthPrefixedCoder(
...
-    //  3) the requested coder is an unknown coder. In this case we either wrap the requested coder
+    //  3) the requested coder is a serialized Java coder. In this case we wrap the coder with
+    //     a length prefix coder.
+    //  4) the requested coder is an unknown coder. In this case we either wrap the requested coder
     //     with a length prefix coder or replace it with a length prefix byte array coder.
     if (ModelCoders.LENGTH_PREFIX_CODER_URN.equals(urn)) {
       return replaceWithByteArrayCoder ? lengthPrefixedByteArrayCoderId : coderId;
     } else if (ModelCoders.urns().contains(urn)) {
       return addForModelCoder(coderId, components, replaceWithByteArrayCoder);
+    } else if (CoderTranslation.JAVA_SERIALIZED_CODER_URN.equals(urn)) {
```
lukecwik: This isn't what your looking for. Runners and SDKs have a duality in the coder representation. See [Encoding and Decoding Unknown Types](This isn't the correct place to fix this. It is likely that a portable runner isn't storing "two" coder representations, one for itself and one for the SDK. See [Encoding and Decoding Unknown Types](https://docs.google.com/document/d/1IGduUqmhWDi_69l9nG8kw73HZ5WI5wOps9Tshl5wpQA/edit#heading=h.sh4d5klmtfis).

The SDK may want to represent a type `Foo` but the runner can't encode or decode this type. The runner instead uses a representation which has the same wire encoding by using `LengthPrefix<Byte[]>` for its coder while the SDK would be told to use a `LengthPrefix<Foo>`. Both have the same wire representation.

lukecwik: You'll want to find the location that is calling this code and see if it is using the incorrect representation (or not remembering this duality).

ihji: Thanks for the comment. Will look into the code you mentioned!

ihji:
The problem is that we're trying to support [`Read` transform](https://github.com/apache/beam/blob/0cc6b671de91719ea80b7aae5164a5500eff3d83/runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkBatchPortablePipelineTranslator.java#L554) in `FlinkRunner`. According to the current portability wire coder design, any transform handled by a runner itself can only generate a `PCollection` with either a standard or a length-prefixed byte array coder. If we have `Read` transform which generates a custom Java class, it will throw a `ClassCastException`. `PortableRunner` substitutes `Read` with `Impulse+DoFn` to avoid this problem. We can do the same thing in expansion service (#8491) and it fixes the issue I tried to solve with this PR.

mxm:
Thanks for your comments @lukecwik. And thanks @ihji for taking the time to break down the problem. It makes perfect sense, but I'm not sure about the solution proposed in #8491. Reads are deliberately translated by the Runner to be able to support unbounded sources.

ihji:
> Reads are deliberately translated by the Runner to be able to support unbounded sources.

This might not be true for portability framework. I don't know whether unbounded `Read` transform works just fine on Flink portable runner or not. But, if it does, I think it's because `FlinkStreamingPortablePipelineTranslator` does not use `WireCoder` for translating `Read` transform. In case of `FlinkBatchPortablePipelineTranslator`, it uses
```
outputCoder = WireCoders.instantiateRunnerWireCoder(collectionNode, pipeline.getComponents());
```
When `Read` transform run by Flink runner itself produces `PCollection` of something that should be encoded with `SerializableCoder`, it will throw the exception because `SerializableCoder` is not supported by a runner wire coder.

If we generate elements with SDF, everything should be okay since the source `PCollection` is the output of `DoFn` and SDK harness supports any coders.

ihji:
Closing this PR. This PR does not fix the source of the problem.

mxm:
@ihji That's true, unbounded sources in portability are limited to the standard coders. Custom classes using the custom coders like SerlizableCoder are indeed not supported.

PR #8399, opened 2019-04-25
[BEAM-7112] Timer race with state cleanup - take two

tweise:
There was still a race condition when cleanup and user timer are triggered together in finish bundle. This is resolved by deferring state cleanup to the next bundle.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
+++ b/runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
@@ -35,6 +35,7 @@ import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.BiConsumer;
+import java.util.function.Consumer;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import org.apache.beam.model.fnexecution.v1.BeamFnApi.ProcessBundleProgressResponse;
@@ -42,7 +43,6 @@ import org.apache.beam.model.fnexecution.v1.BeamFnApi.ProcessBundleResponse;
 import org.apache.beam.model.fnexecution.v1.BeamFnApi.StateKey.TypeCase;
 import org.apache.beam.model.pipeline.v1.RunnerApi;
 import org.apache.beam.runners.core.DoFnRunner;
-import org.apache.beam.runners.core.DoFnRunners;
 import org.apache.beam.runners.core.LateDataUtils;
 import org.apache.beam.runners.core.StateInternals;
 import org.apache.beam.runners.core.StateNamespace;
@@ -118,7 +118,6 @@ public class ExecutableStageDoFnOperator<InputT, OutputT> extends DoFnOperator<I
   private transient SdkHarnessDoFnRunner<InputT, OutputT> sdkHarnessRunner;
   private transient FlinkMetricContainer flinkMetricContainer;
   private transient long backupWatermarkHold = Long.MIN_VALUE;
-  private transient ArrayDeque<InternalTimer<?, TimerInternals.TimerData>> cleanupTimers;
 
   /** Constructor. */
   public ExecutableStageDoFnOperator(
@@ -189,7 +188,6 @@ public class ExecutableStageDoFnOperator<InputT, OutputT> extends DoFnOperator<I
             flinkMetricContainer.updateMetrics(stepName, response.getMonitoringInfosList());
           }
         };
-    cleanupTimers = new ArrayDeque<>();
 
     // This will call {@code createWrappingDoFnRunner} which needs the above dependencies.
     super.open();
@@ -398,35 +396,8 @@ public class ExecutableStageDoFnOperator<InputT, OutputT> extends DoFnOperator<I
     }
   }
 
-  @SuppressWarnings("ByteBufferBackingArray")
-  private void fireCleanupTimers() {
-    while (!cleanupTimers.isEmpty()) {
-      InternalTimer<?, TimerInternals.TimerData> timer = cleanupTimers.remove();
-      final ByteBuffer encodedKey = (ByteBuffer) timer.getKey();
-      try {
-        // still need to process as timer, see CleanupTimer
-        stateBackendLock.lock();
-        if (LOG.isDebugEnabled()) {
-          LOG.debug(
-              "State cleanup for {} {}",
-              Arrays.toString(encodedKey.array()),
-              timer.getNamespace().getNamespace());
-        }
-        getKeyedStateBackend().setCurrentKey(encodedKey);
-        super.fireTimer(timer);
-      } finally {
-        stateBackendLock.unlock();
-      }
-    }
-  }
-
   @Override
   public void fireTimer(InternalTimer<?, TimerInternals.TimerData> timer) {
-    if (CleanupTimer.GC_TIMER_ID.equals(timer.getNamespace().getTimerId())) {
-      // hold cleanup until bundle is complete
-      cleanupTimers.add(timer);
-      return;
-    }
     final ByteBuffer encodedKey = (ByteBuffer) timer.getKey();
     // We have to synchronize to ensure the state backend is not concurrently accessed by the state
     // requests
@@ -526,9 +497,6 @@ public class ExecutableStageDoFnOperator<InputT, OutputT> extends DoFnOperator<I
                 // at this point the bundle is finished, allow the watermark to pass
                 // we are restoring the previous hold in case it was already set for side inputs
                 setPushedBackWatermark(backupWatermarkHold);
-                // fire cleanup timers, they can only execute after the bundle is complete
-                // as they remove the state that the timer callback may rely on
-                fireCleanupTimers();
                 super.processWatermark(mark);
               } catch (Exception e) {
                 throw new RuntimeException(
@@ -538,10 +506,6 @@ public class ExecutableStageDoFnOperator<InputT, OutputT> extends DoFnOperator<I
       }
     }
     super.processWatermark(mark);
-    // if this was the final watermark, then no callback was scheduled
-    if (mark.getTimestamp() >= BoundedWindow.TIMESTAMP_MAX_VALUE.getMillis()) {
-      fireCleanupTimers();
-    }
   }
 
   private static class SdkHarnessDoFnRunner<InputT, OutputT>
@@ -741,10 +705,29 @@ public class ExecutableStageDoFnOperator<InputT, OutputT> extends DoFnOperator<I
         executableStage.getUserStates().stream()
             .map(UserStateReference::localName)
             .collect(Collectors.toList());
-    StateCleaner stateCleaner = new StateCleaner(userStates, windowCoder, keyedStateInternals);
 
-    return DoFnRunners.defaultStatefulDoFnRunner(
-        doFn, sdkHarnessRunner, windowingStrategy, cleanupTimer, stateCleaner);
+    KeyedStateBackend<ByteBuffer> stateBackend = getKeyedStateBackend();
+    StateCleaner stateCleaner =
+        new StateCleaner(userStates, windowCoder, () -> stateBackend.getCurrentKey());
+
+    return new StatefulDoFnRunner<InputT, OutputT, BoundedWindow>(
+        sdkHarnessRunner, windowingStrategy, cleanupTimer, stateCleaner) {
+      @Override
+      public void finishBundle() {
+        // Before cleaning up state, first finish bundle for all underlying DoFnRunners
+        super.finishBundle();
+        // execute cleanup after the bundle is complete
+        if (!stateCleaner.cleanupQueue.isEmpty()) {
+          try {
+            stateBackendLock.lock();
+            stateCleaner.cleanupState(
+                keyedStateInternals, (key) -> stateBackend.setCurrentKey(key));
+          } finally {
+            stateBackendLock.unlock();
+          }
+        }
+      }
+    };
   }
 
   static class CleanupTimer<InputT> implements StatefulDoFnRunner.CleanupTimer<InputT> {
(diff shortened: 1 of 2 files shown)


runners/flink/src/test/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperatorTest.java
```diff
@@ -449,6 +466,92 @@ public void testEnsureStateCleanupWithKeyedInputStateCleaner() throws Exception
...
+        getOperator(
+            mainOutput, Collections.emptyList(), outputManagerFactory, windowingStrategy, keyCoder);
+
+    @SuppressWarnings("unchecked")
+    RemoteBundle bundle = Mockito.mock(RemoteBundle.class);
+    when(stageBundleFactory.getBundle(any(), any(), any())).thenReturn(bundle);
+
+    String timerInputId = "timerInput";
+    AtomicBoolean timerInputReceived = new AtomicBoolean();
+    Instant windowEnd = new Instant(1000);
```
mxm: End of window is `windowEnd  - 1` because the end if not inclusive. See `IntervalWindow`.

I'd suggest removing this variable and using `window.maxTimestamp()` instead to track the end of the window.

runners/flink/src/test/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperatorTest.java
```diff
@@ -449,6 +466,92 @@ public void testEnsureStateCleanupWithKeyedInputStateCleaner() throws Exception
...
+
+    DoFnOperator.FlinkTimerInternals timerInternals =
+        (DoFnOperator.FlinkTimerInternals) Whitebox.getInternalState(operator, "timerInternals");
+    Collection<?> cleanupTimers = (Collection) Whitebox.getInternalState(operator, "cleanupTimers");
+
+    TimerInternals.TimerData timerData =
+        TimerInternals.TimerData.of(
+            timerInputId,
+            StateNamespaces.window(windowCoder, window),
+            windowEnd.plus(1),
```
mxm:
```suggestion
            window.maxTimestamp(),
```

You cannot set a timer later than the end of the window.

runners/flink/src/test/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperatorTest.java
```diff
@@ -449,6 +466,92 @@ public void testEnsureStateCleanupWithKeyedInputStateCleaner() throws Exception
...
+    KeyedStateBackend<ByteBuffer> keyedStateBackend = operator.getKeyedStateBackend();
+    ByteBuffer key =
+        FlinkKeyUtils.encodeKey(one.getValue().getKey(), keyCoder, Coder.Context.NESTED);
+    keyedStateBackend.setCurrentKey(key);
+
+    DoFnOperator.FlinkTimerInternals timerInternals =
+        (DoFnOperator.FlinkTimerInternals) Whitebox.getInternalState(operator, "timerInternals");
+    Collection<?> cleanupTimers = (Collection) Whitebox.getInternalState(operator, "cleanupTimers");
+
+    TimerInternals.TimerData timerData =
```
mxm: Would rename to `userTimer`.

runners/flink/src/test/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperatorTest.java
```diff
@@ -449,6 +466,92 @@ public void testEnsureStateCleanupWithKeyedInputStateCleaner() throws Exception
...
+            StateNamespaces.window(windowCoder, window),
+            windowEnd.plus(1),
+            TimeDomain.EVENT_TIME);
+    timerInternals.setTimer(timerData);
+
+    testHarness.processElement(new StreamRecord<>(one));
+    verify(receiver).accept(one);
+
+    // fire user and cleanup timer while bundle in progress
+    operator.processWatermark(new Watermark(windowEnd.getMillis() + 2));
```
mxm:
```suggestion
    operator.processWatermark(new Watermark(window.maxTimestamp() + 2));
```

runners/flink/src/test/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperatorTest.java
```diff
@@ -449,6 +466,92 @@ public void testEnsureStateCleanupWithKeyedInputStateCleaner() throws Exception
...
+    assertFalse("watermark hold", timerInputReceived.get());
```
mxm:
```suggestion
    assertFalse("Watermark must be held back until bundle is complete.", timerInputReceived.get());
```

runners/flink/src/test/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperatorTest.java
```diff
@@ -449,6 +466,92 @@ public void testEnsureStateCleanupWithKeyedInputStateCleaner() throws Exception
...
+    assertThat(cleanupTimers, hasSize(0));
+
+    operator.invokeFinishBundle();
+    assertTrue("timer triggered", timerInputReceived.get());
```
mxm:
```suggestion
    assertTrue("Timer should have been triggered.", timerInputReceived.get());
```

runners/flink/src/test/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperatorTest.java
mxm: `windowEnd.getMillis() + 2 == window.maxTimestamp() + 2 + 1` (one more millisecond than needed, the gc timer for the window fires when the watermark has reached `window.maxTimestamp() + 2`)

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
```diff
@@ -401,7 +402,14 @@ private void setTimer(WindowedValue<InputT> timerElement, TimerInternals.TimerDa
   @SuppressWarnings("ByteBufferBackingArray")
   private void fireCleanupTimers() {
     while (!cleanupTimers.isEmpty()) {
-      InternalTimer<?, TimerInternals.TimerData> timer = cleanupTimers.remove();
+      KV<RemoteBundle, InternalTimer<?, TimerInternals.TimerData>> kv = cleanupTimers.peek();
+      if (kv.getKey() != null && kv.getKey() == sdkHarnessRunner.remoteBundle) {
+        // user timer and cleanup may trigger together in finish bundle,
+        // defer state cleanup to allow user timers to complete
+        return;
```
mxm: Wouldn't it be easier to fire timers only after finishing the bundle? That way we would guarantee for user timers to be fired.

mxm:
Thanks Thomas! I think we should simplify the firing logic of the cleanup timers but looks good otherwise.

runners/flink/src/test/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperatorTest.java
tweise: This is supposed to trigger both the user timer and the state cleanup. User timer is scheduled after cleanup since I want to verify that the execution gets reordered. Added comments.

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
tweise: This gets executed during finish bundle and will ensure that the cleanup timers fire after the bundle only.

runners/flink/src/test/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperatorTest.java
tweise: It is the intention to set it this way, after the cleanup timer. The window only matters for scheduling the cleanup timer in this test.

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
mxm: Let's move it to where it belongs, i.e. after the bundle has been finished.

tweise:
Run Java PreCommit

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
```diff
@@ -812,21 +796,35 @@ public boolean isForWindow(
...
     }
 
     @Override
     public void clearForWindow(BoundedWindow window) {
-      // Executed in the context of onTimer(..) where the correct key will be set
-      for (String userState : userStateNames) {
-        StateNamespace namespace = StateNamespaces.window(windowCoder, window);
-        BagState<?> state = stateInternals.state(namespace, StateTags.bag(userState, null));
-        state.clear();
+      cleanupQueue.add(KV.of(keyedStateBackend.get(), window));
```
mxm:
```suggestion
      // Executed in the context of onTimer(..) where the correct key will be set
      ByteBuffer key = keyedStateBackend.get();
      cleanupQueue.add(KV.of(key, window));
```

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
```diff
@@ -563,6 +527,7 @@ public void processWatermark(Watermark mark) throws Exception {
 
     private RemoteBundle remoteBundle;
     private FnDataReceiver<WindowedValue<?>> mainInputReceiver;
+    private ArrayDeque<KV<ByteBuffer, BoundedWindow>> cleanupWindows;
```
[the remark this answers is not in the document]
mxm: Unused field.

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
```diff
@@ -592,6 +557,7 @@ public SdkHarnessDoFnRunner(
       }
       this.windowCoder = windowCoder;
       this.outputQueue = new LinkedBlockingQueue<>();
+      this.cleanupWindows = new ArrayDeque<>();
```
mxm:
```suggestion
```

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
```diff
@@ -812,21 +796,35 @@ public boolean isForWindow(
...
+    private final Supplier<ByteBuffer> keyedStateBackend;
 
-    StateCleaner(List<String> userStateNames, Coder windowCoder, StateInternals stateInternals) {
+    StateCleaner(
+        List<String> userStateNames, Coder windowCoder, Supplier<ByteBuffer> keyedStateBackend) {
       this.userStateNames = userStateNames;
       this.windowCoder = windowCoder;
-      this.stateInternals = stateInternals;
+      this.keyedStateBackend = keyedStateBackend;
+      this.cleanupQueue = new ArrayDeque<>();
```
mxm: I think we have a consistency problem here because this is not persisted in checkpoints.

tweise: Hmm, that should not be the case when the cleanup is performed as part of finishing the bundle. Need to see how that is possible.

tweise:
We don't need to checkpoint the cleanup queue since it does not cross the bundle boundary. But we need to look into another issue (outside of this PR): The finish bundle callback starts a new bundle when the watermark advances, which can lead to inconsistent state snapshot.

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
```diff
@@ -812,21 +796,35 @@ public boolean isForWindow(
 
     private final List<String> userStateNames;
     private final Coder windowCoder;
-    private final StateInternals stateInternals;
+    private final ArrayDeque<KV<ByteBuffer, BoundedWindow>> cleanupQueue;
```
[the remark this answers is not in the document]
tweise: No need to use use an interface here.

tweise:
Run Java PreCommit

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
mxm: Right, if after `invokeFinishBundle()` has been called the cleanup queue is always cleared as part of however many additional bundles, then we are fine. If we delay any cleanup for after the checkpoint has completed, then we would have a problem.

mxm:
>The finish bundle callback starts a new bundle when the watermark advances, which can lead to inconsistent state snapshot.

I think the solution could be to checkpoint cleanup timers and delay execution of the timers for until after the checkpoint has completed. Otherwise we have to checkpoint any possible emitted elements during completing the bundle within `snapshotState`.

mxm:
Run Java PreCommit

mxm:
Run Python PreCommit

mxm:
Build successful but it stuck collecting results: https://builds.apache.org/job/beam_PreCommit_Java_Commit/5674/
```
06:13:42 BUILD SUCCESSFUL in 21m 32s
06:13:42 966 actionable tasks: 932 executed, 34 from cache
```

mxm:

mxm:

tweise:
> I think this is good to merge. Let's tackle problems with the checkpoint interaction next. On Flink >= 1.6 this would be trivial to solve by flushing in`preSnapshotBarrier`. For 1.5 we will have to do one of the following:
> 
> 1. Checkpoint cleanup timers and defer execution for after the checkpoint
> 2. Buffer emitted elements during snapshot (this is already done but there is no key scope)

The issues is that a new bundles is created during finish bundle, which is actually not related to this PR. Tracked in: https://issues.apache.org/jira/browse/BEAM-7171

PR #8414, opened 2019-04-26
[BEAM-6526] Add ReadFiles and ParseFiles transforms for AvroIO

iemejia:
I finally fixed the issue with the breaking tests. PTAL again 
R: @lgajowy

### DIFF
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/io/AvroIO.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/io/AvroIO.java
@@ -283,6 +283,23 @@ public class AvroIO {
         .build();
   }
 
+  /**
+   * Like {@link #read}, but reads each file in a {@link PCollection} of {@link
+   * FileIO.ReadableFile}, returned by {@link FileIO#readMatches}.
+   *
+   * <p>You can read {@link GenericRecord} by using {@code #readFiles(GenericRecord.class)} or
+   * {@code #readFiles(new Schema.Parser().parse(schema))} if the schema is a String.
+   */
+  public static <T> ReadFiles<T> readFiles(Class<T> recordClass) {
+    return new AutoValue_AvroIO_ReadFiles.Builder<T>()
+        .setMatchConfiguration(MatchConfiguration.create(EmptyMatchTreatment.ALLOW_IF_WILDCARD))
+        .setRecordClass(recordClass)
+        .setSchema(ReflectData.get().getSchema(recordClass))
+        .setInferBeamSchema(false)
+        .setDesiredBundleSizeBytes(DEFAULT_BUNDLE_SIZE_BYTES)
+        .build();
+  }
+
   /** Like {@link #read}, but reads each filepattern in the input {@link PCollection}. */
   public static <T> ReadAll<T> readAll(Class<T> recordClass) {
     return new AutoValue_AvroIO_ReadAll.Builder<T>()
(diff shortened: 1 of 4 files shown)


lgajowy:
I'm on vacation until 5th may. 😎 Please reassign reviewer if it cannot wait. :)

iemejia:
Ok, not mega urgent but maybe worth merging because I have other 2 PRs pending on this one. So chaing reviewer, thanks for the info @lgajowy 

R: @chamikaramj can you please take a look at this one.
Note, there was already a first round on this one at PR #7672. But it was closed due to stalenessbecause I did not update it rapidly.

sdks/java/core/src/main/java/org/apache/beam/sdk/io/AvroIO.java
```diff
@@ -283,17 +283,31 @@
         .build();
   }
 
+  /**
+   * Like {@link #read}, but reads each file in a {@link PCollection} of {@link
+   * FileIO.ReadableFile}, returned by {@link FileIO#readMatches}.
```
chamikaramj: ",for example, returned by {@link FileIO#readMatches}" since manually constructing a PCollection of ReadableFile objects will work too.

sdks/java/core/src/main/java/org/apache/beam/sdk/io/AvroIO.java
```diff
@@ -308,6 +322,17 @@
         .build();
   }
 
+  /** Like {@link #readGenericRecords(Schema)}, but for {@link FileIO.ReadableFile} collections. */
```
chamikaramj: Probably clearer to say: "but for a {@link PCollection} of {@link FileIO.ReadableFile}, for example, returned by {@link FileIO#readMatches}."

sdks/java/core/src/main/java/org/apache/beam/sdk/io/AvroIO.java
```diff
@@ -330,6 +355,11 @@
     return readGenericRecords(new Schema.Parser().parse(schema));
   }
 
+  /** Like {@link #readGenericRecords(String)}, but for {@link FileIO.ReadableFile} collections. */
```
chamikaramj: Ditto.

sdks/java/core/src/main/java/org/apache/beam/sdk/io/AvroIO.java
```diff
@@ -283,17 +283,31 @@
...
   }
 
+  /**
+   * Like {@link #read}, but reads each file in a {@link PCollection} of {@link
+   * FileIO.ReadableFile}, returned by {@link FileIO#readMatches}.
+   *
+   * <p>You can read {@link GenericRecord} by using {@code #readFiles(GenericRecord.class)} or
+   * {@code #readFiles(new Schema.Parser().parse(schema))} if the schema is a String.
+   */
+  public static <T> ReadFiles<T> readFiles(Class<T> recordClass) {
```
chamikaramj: Please consider adding an integration test for this (similar to https://github.com/apache/beam/blob/master/sdks/java/io/file-based-io-tests/src/test/java/org/apache/beam/sdk/io/avro/AvroIOIT.java).

sdks/java/core/src/main/java/org/apache/beam/sdk/io/AvroIO.java
iemejia: done

sdks/java/core/src/main/java/org/apache/beam/sdk/io/AvroIO.java
iemejia: done

sdks/java/core/src/main/java/org/apache/beam/sdk/io/AvroIO.java
iemejia: will do, I have missed that one because it was not in the same package, thanks.

iemejia:
Run Java AvroIO Performance Test

iemejia:
Run Java AvroIO Performance Test

iemejia:
Updates done, PTAL again @chamikaramj 
Side note, I added `ParseFiles` too and had to expose an enum in FileIO to be able to configure it in the IT (this was a curious issue because we had a a public method `FileIO.withDirectoryTreatment` that received a non public enum `DirectoryTreatment`).

chamikaramj:
LGTM. Thanks.

PR #8418, opened 2019-04-26
[BEAM-6479] Deprecate AvroIO.RecordFormatter

iemejia:
Another really simple one in the same subject (FileIO/AvroIO)
R: @chamikaramj

### DIFF
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/io/AvroIO.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/io/AvroIO.java
@@ -1452,7 +1452,13 @@ public class AvroIO {
   }
   /////////////////////////////////////////////////////////////////////////////
 
-  /** Formats an element of a user type into a record with the given schema. */
+  /**
+   * Formats an element of a user type into a record with the given schema.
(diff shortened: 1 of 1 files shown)


sini:
Question, as I use this... in my FileIO dynamic write we have a mixed PCollection of hundreds of different types of binary avro records that we re-serialize using this sink. Would materializing an intermediate PCollection of GenericRecords, as is the proposed alternative, not increase the size and cost of shuffle operations if the collection needed to be flushed to disk as the schema is duplicated for every record? Or was this also a problem with the RecordFormatter transform step and I was just oblivious to it?

iemejia:
What is proposed here is that Beam users should not use a `RecordFormatter` to convert from `ElementT` to `GenericRecord` when writing, but better do this explicitly in an extra `ParDo.of(new DoFn<ElementT, GenericRecord>)` before. There is in practice little difference (apart of the fact that using ParDo is more composable and  'Beamish').  Full materialization should not happen because DoFns are normally fusioned by the runner, of course there could be a small minor extra cost because of the extra DoFn instantiation but probably negligable.

PR #8430, opened 2019-04-29
Changes to SDF API to use DoFn Params

pabloem:
A minimal change to the SDF API for consistency and clarity of expectations.

### DIFF
--- a/sdks/python/apache_beam/runners/common.py
+++ b/sdks/python/apache_beam/runners/common.py
@@ -239,8 +239,8 @@ class DoFnSignature(object):
 
   def get_restriction_provider(self):
     result = _find_param_with_default(self.process_method,
-                                      default_as_type=RestrictionProvider)
-    return result[1] if result else None
+                                      default_as_type=DoFn.RestrictionParam)
+    return result[1].restriction_provider if result else None
 
   def _validate(self):
     self._validate_process()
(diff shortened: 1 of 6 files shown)


sdks/python/apache_beam/runners/common.py
```diff
@@ -239,8 +239,8 @@ def __init__(self, do_fn):
 
   def get_restriction_provider(self):
```
aaltay: Rename the method as well?

aaltay: Scratch that.

charlesccychen:
I think you may need to update the invoker logic in common.py: https://github.com/apache/beam/blob/ef67378fec2afadbe4a5303e61ac0427ced1faa0/sdks/python/apache_beam/runners/common.py#L477

pabloem:
Thanks for looking @charlesccychen. I believe the parameter is repalced in invoke_process, and the relevant change is in line 541: https://github.com/apache/beam/pull/8430/files#diff-dc4f89b2c45e945478f3fb5a0c482976R541

pabloem:
Run Python PostCommit

pabloem:
Run Python PostCommit

pabloem:
Run Python PostCommit

pabloem:
In Java, users provide a `TableDestination` object. The `TableDestination` interface has individual functions for each table attribute, and doesn't currently support clustering - only time partitioning. The API that we are defining allows us to support time partitioning, and clustering (and any new BQ feature if necessary).

See https://beam.apache.org/releases/javadoc/2.12.0/org/apache/beam/sdk/io/gcp/bigquery/TableDestination.html

pabloem:
Run Python PostCommit

pabloem:
r: @robertwb  PTAL : )

pabloem:
This may require updates to documentation elsewhere.

PR #8453, opened 2019-05-01
[BEAM-6985] TypeHints Py3 Error: Native type compatibility tests fail on Python 3.7+ Updates

NikeNano:
These changes are related to the difference in the typing module. In short the issubclass  is removed for typing after python 3.5 and thus this needs to be handled. For more information about the difference and change in typing see:

- https://stackoverflow.com/questions/42027923/mypy-differences-in-isinstance-and-issubclass-from-python-3-5-to-3-6-in-paramete

- https://github.com/python/typing/issues/136

In order to accomplish the same behaviour as issubclass did for beam, python2 in python3  the attribute .\_\_origon\_\_ is used for some of the data structures, see below. Using .\_\_origin\_\_  result in that both parent and derived is evaluated as subclass of each other. To solve this the attribute .\_\_args\_\_ is used,  _\_args\_\_ gives the arguments of the typing object. For further information see:
https://stackoverflow.com/questions/48572831/how-to-access-the-type-arguments-of-typing-generic

The comparison is splited in the following cases:

- evaluation of typing.[Dict,List,Set,Tuple], there are some special handling of Namedtuples(which is based upon the original code). The evaluation for these types are done using .\_\_orign\_\_ in order to also check the parent vs derived the attribute .\_\_args\_\_ are used to check for arguments
- union evaluation of typing.Union, this case is different in the sense that typing.Union.\_\_origin\_\_ result in attribute error. However for example typing.Union[int, typing.List[int]].\_\_origin\_\_ gives typing.Union therefore the comparison is: derived.__origin__ is parent 
- any is compared using type

------------------------

--- | --- | --- | --- | ---


NikeNano:
R: @RobbeSneyders

I saw that some of the commits where not done with the correct naming, hope that is still ok. Happy for all feedback to improve upon my contributions to Beam :)

NikeNano:
I will check and fix the formatting later this evening (CET) :)

NikeNano:
All test running now :)

NikeNano:
@aaltay, @tvalentyn saw that I should you guys as well in it. According to https://beam.apache.org/roadmap/python-sdk/#python-3-support

tvalentyn:
Thanks a lot @NikeNano , I just noticed that you were working on https://issues.apache.org/jira/browse/BEAM-6985. I'll take a look at the PR, stay tuned.

NikeNano:
Ahh missed the naming with one in the PR, should I change it to 6985?

tvalentyn:
Sure, that wouldn't hurt and is a few clicks in the UI :-)

NikeNano:
Made some testa on apache_beam.typehints.typed_pipeline_test.NativeTypesTest and seams like this patch might solve the problems related to [BEAM-6987].

sdks/python/apache_beam/typehints/native_type_compatibility.py
```diff
@@ -84,19 +84,38 @@ def _safe_issubclass(derived, parent):
     issubclass(derived, parent), or False if a TypeError was raised.
   """
   try:
-    return issubclass(derived, parent)
+    if type(derived) == typing._GenericAlias:
```
tvalentyn: You could consider following to reduce the nesting in try-catch.
```
  try:
    if hasattr(typing, '_GenericAlias') && (type(derived) == typing._GenericAlias):
      # Python 3.7+
      out = issubclass(derived.__origin__, parent.__origin__)
    else:
      out = issubclass(derived, parent)
    return out
  except TypeError:
    return False
```

tvalentyn: Do we also need to compare `issubclass(derived.__origin__, parent)`?

Also how about we add targeted unit tests just for `_safe_issublass`?

sdks/python/apache_beam/typehints/native_type_compatibility.py
```diff
@@ -84,19 +84,38 @@ def _safe_issubclass(derived, parent):
...
+      return out
+    except TypeError:
+      return False
 
 
 def _match_issubclass(match_against):
   return lambda user_type: _safe_issubclass(user_type, match_against)
 
 
+def _matcher_(derived, parent):
```
tvalentyn: Let's make it a nested function inside _match_same_type:

```
def _match_same_type(match_against):
  def matcher(user_type):
      # Compare user_type and match_against here
  
  return lambda user_type: match(user_type)
```

sdks/python/apache_beam/typehints/native_type_compatibility.py
```diff
@@ -84,19 +84,38 @@ def _safe_issubclass(derived, parent):
...
+  """ Follow structure of _safe_issubclass
+  but uses __orign__ due to updates in typing package of python3"""
+  try:
+    return derived.__origin__ is parent
```
tvalentyn: Do we by  chance need to compare all combinations here?

e.g. type1 is type2 or type1.__origin__ is type2 or type1 is type2.__origin__?

sdks/python/apache_beam/typehints/native_type_compatibility.py
NikeNano: Sounds good!

sdks/python/apache_beam/typehints/native_type_compatibility.py
NikeNano: Sounds good, will also add tests tomorrow morning.

sdks/python/apache_beam/typehints/native_type_compatibility.py
NikeNano: Hmm don't know if that is needed. Will do some tests and try it.

tvalentyn:
Thanks, @NikeNano , please ping this PR once it's ready for review. You can say "PTAL"=please take another look or something like that.

sdks/python/apache_beam/typehints/native_type_compatibility.py
NikeNano: I have added the test. I don't think we need to add the reversed order. Partly since the test cases works great and I believe they have good coverage. Second I think the order has a purpose, we only want to match if the derived (class) is a subclass of the parent(classinfo).  See: https://docs.python.org/3/library/functions.html#issubclass. 

Let me know if you disagree and I will update :)

sdks/python/apache_beam/typehints/native_type_compatibility.py
NikeNano: Same as above, but would be great if you could elaborate on why you think it might be needed? Maybe I am missing something.

sdks/python/apache_beam/typehints/native_type_compatibility_test.py
```diff
@@ -103,6 +103,31 @@ def test_convert_to_beam_types(self):
         native_type_compatibility.convert_to_beam_types(typing_types),
         beam_types)
 
+  def test_is_sub_class(self):
+    self.assertTrue(native_type_compatibility._safe_issubclass(\
```
tvalentyn: We don't need '\' at the end of the line, since there is a parenthesis. I think you'd need two extra spaces on the next line to avoid lint complaints.
Also, some of these scenarios failed - see PreCommit tests Details page at the end of the PR.
You can also run individual tests via:
```
python ./setup.py test -s apache_beam.typehints.native_type_compatibility_test
```

sdks/python/apache_beam/typehints/native_type_compatibility_test.py
```diff
@@ -37,10 +37,10 @@ class _TestClass(object):
   pass
 
 
-@unittest.skipIf(sys.version_info >= (3, 7, 0) and
-                 os.environ.get('RUN_SKIPPED_PY3_TESTS') != '1',
-                 'This test still needs to be fixed on Python 3.7. '
-                 'See BEAM-6985')
+#@unittest.skipIf(sys.version_info >= (3, 7, 0) and
```
tvalentyn: We can just remove these lines.

sdks/python/apache_beam/typehints/native_type_compatibility_test.py
NikeNano: Working on it atm :), thanks for the tips 👍

sdks/python/apache_beam/typehints/native_type_compatibility_test.py
NikeNano: Will do!

NikeNano:
R: @tvalentyn @RobbeSneyders @aaltay   PTAL

sdks/python/apache_beam/typehints/native_type_compatibility.py
```diff
@@ -84,7 +84,13 @@ def _safe_issubclass(derived, parent):
     issubclass(derived, parent), or False if a TypeError was raised.
   """
   try:
-    return issubclass(derived, parent)
+    if hasattr(typing, '_GenericAlias') and \
+      (type(derived) == typing._GenericAlias):
```
cclauss: Could we please use parens to avoid adding a backslash (see PEP8P)?
```python
    if (hasattr(typing, '_GenericAlias') and
        type(derived) == typing._GenericAlias):
```
One whitespace character to the right of the backslash and the script breaks on a change that is invisible to the reader.

sdks/python/apache_beam/typehints/native_type_compatibility_test.py
```diff
@@ -103,6 +98,64 @@ def test_convert_to_beam_types(self):
...
+    self.assertTrue(native_type_compatibility._safe_issubclass(
+        parent=typing.Dict,
+        derived=typing.Dict[bytes, int]))
+    self.assertFalse(native_type_compatibility._safe_issubclass(
+        parent=typing.List,
+        derived=typing.Dict[bytes, int]))
+    self.assertTrue(native_type_compatibility._safe_issubclass(
+        parent=typing.List,
+        derived=typing.List[bytes]))
+    self.assertFalse(native_type_compatibility._safe_issubclass(
```
tvalentyn: This is a duplicate of line 105.
Also, it is sufficient to focus on different classes of usecases for the function. 
For example between
```
    self.assertTrue(native_type_compatibility._safe_issubclass(
        parent=typing.Set,
        derived=typing.Set[int]))
```
and
```
    self.assertTrue(native_type_compatibility._safe_issubclass(
        parent=typing.Tuple,
        derived=typing.Tuple[int]))
```
I would keep only one scenario.

sdks/python/apache_beam/typehints/native_type_compatibility_test.py
```diff
@@ -103,6 +98,64 @@ def test_convert_to_beam_types(self):
...
+        parent=typing.List,
+        derived=typing.Set[float]))
+    self.assertTrue(native_type_compatibility._safe_issubclass(
+        parent=typing.Tuple,
+        derived=typing.Tuple[int]))
+    self.assertFalse(native_type_compatibility._safe_issubclass(
+        parent=typing.List,
+        derived=typing.Tuple[bytes]))
+
+  @unittest.skipIf(sys.version_info >= (2, 7, 0),
```
tvalentyn: Why is this a correct behavior for this function to return different results on Py2 and Py3?

sdks/python/apache_beam/typehints/native_type_compatibility_test.py
```diff
@@ -103,6 +98,64 @@ def test_convert_to_beam_types(self):
...
+        parent=typing.List[bytes],
+        derived=typing.List))
+    self.assertTrue(native_type_compatibility._safe_issubclass(
+        parent=typing.Set[int],
+        derived=typing.Set))
+    self.assertTrue(native_type_compatibility._safe_issubclass(
+        parent=typing.Tuple[int],
+        derived=typing.Tuple))
+
+  @unittest.skipIf(sys.version_info.major != '3',
```
tvalentyn: It is  discouraged to compare to exact "3" version, as one day there may be Python 4, see also: https://docs.python.org/3/howto/pyporting.html#use-feature-detection-instead-of-version-detection

```if sys.version_info.major < 3 will be better```

File: sdks/python/apache_beam/typehints/native_type_compatibility_test.py (hunk 7 above)
Comment (NikeNano) [reply sdks/python/apache_beam/typehints/native_type_compatibility_test.py]: True, removed the duplicate

File: sdks/python/apache_beam/typehints/native_type_compatibility_test.py (hunk 8 above)
Comment (NikeNano) [reply sdks/python/apache_beam/typehints/native_type_compatibility_test.py]: To be honest I don't know if it is an advantage or not. I realised there was a difference and made the test to point it out. Maybe I also should have pointed it out more clearly in the PR as well. Based upon your comment I assume that we want to keep the Py2 behaviour and will investigate further if I can achieve that with minimal changes.

Comment (tvalentyn) [reply sdks/python/apache_beam/typehints/native_type_compatibility_test.py]: I looked a little closer into the code of native_type_compatibility module.

The high-level goal of this change is to make sure that the correspondence between `typing` types and `apache_beam.typehints` types, defined by `convert_to_beam_type`, is established in the same way across all versions of Python. 

`test_convert_to_beam_type` should help us ensure this. As long as this test has sufficient coverage, we don't need to test all the helpers.  

`_is_subclass` is a helper function, and it's parent argument is one of typing.{Dict,List,Set}, so we can remove `_is_sublcass_order` test cases, which pass subscripted types, e.g. List[int].

There are several interesting differences in how following expressions are evaluated  different Py3.6 / Py3.7 versions, and there may be more differences in other Python versions.

```
type(typing.Any)
type(typing.Dict)
type(typing.List)
type(typing.Set)
type(typing.Tuple)
type(typing.Union)

type(typing.Any)
type(typing.Dict[int])
type(typing.List[int])
type(typing.Set[int])
type(typing.Tuple[int])
type(typing.Union[int,str])

typing.Dict[int].__origin__
typing.List[int].__origin__
typing.Set[int].__origin__
typing.Tuple[int].__origin__
typing.Union[int,str].__origin__

typing.Any.__origin__
typing.Dict.__origin__
typing.List.__origin__
typing.Set.__origin__
typing.Tuple.__origin__
typing.Union.__origin__
```
Taking these differences into account, we need to make sure that whatever comparisons we are doing in `_match_same_type` and `_is_subclass` make sense, and we not just making a change that makes a test pass. A reader unfamiliar with this code should also be able to understand the details of implementation. 

Looking at the evaluation of expressions above, I think we will be better off if we replace `_match_same_type` into separate `_match_union` and `_match_any` helpers.

**Comment by tvalentyn** [comment]
Also cc: udim@ who is working on improving type-checking in Beam.

**Comment by tvalentyn** [comment]
Also cc'ing: udim@ who is working on improving type-checking in Beam.

File: sdks/python/apache_beam/typehints/native_type_compatibility_test.py (hunk 8 above)
Comment (NikeNano) [reply sdks/python/apache_beam/typehints/native_type_compatibility_test.py]: Yes, But what comparison makes sense. Would be like the comparison to go both ways or one way? I will remove the helper tests and split it.

File: sdks/python/apache_beam/typehints/native_type_compatibility.py [hunk 10]
```diff
@@ -95,8 +101,15 @@ def _match_issubclass(match_against):
 
 def _match_same_type(match_against):
   # For Union types. They can't be compared with isinstance either, so we
-  # have to compare their types directly.
-  return lambda user_type: type(user_type) == type(match_against)
+  # Have to compare their types directly.
```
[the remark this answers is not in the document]
NikeNano: Makes sens will update

sdks/python/apache_beam/typehints/native_type_compatibility.py
```diff
@@ -95,8 +101,15 @@ def _match_issubclass(match_against):
...
+
+  def matcher(derived, parent):
+    try:
+      return derived.__origin__ is parent
```
udim: Is there a unit test for this? I'm wondering why this code is not:
```suggestion
      return derived.__origin__ is parent.__origin__
```

sdks/python/apache_beam/typehints/native_type_compatibility_test.py
udim: Not that this test is always skipped because `major` is an int.

sdks/python/apache_beam/typehints/native_type_compatibility_test.py
```diff
@@ -103,6 +98,64 @@ def test_convert_to_beam_types(self):
...
+    self.assertTrue(native_type_compatibility._safe_issubclass(
+        parent=typing.Tuple,
+        derived=typing.Tuple[int]))
+    self.assertFalse(native_type_compatibility._safe_issubclass(
+        parent=typing.List,
+        derived=typing.Tuple[bytes]))
+
+  @unittest.skipIf(sys.version_info >= (2, 7, 0),
+                   'Order dosent matter in python 3')
+  def test_is_sub_class_order(self):
```
udim: This test never runs because there's an identically named test below it.

sdks/python/apache_beam/typehints/native_type_compatibility_test.py
tvalentyn: > But what comparison makes sense
It would help if you can make a recommendation, and explain why particular comparisons need to be made and why they make sense. It is sometimes easy to make a change to silence the test, but it should also be clear why a particular change is the right thing to do.

NikeNano:
PTLA @tvalentyn I have added a test for the ordering to make sure the behaviour is the same for python2 vs python3.

NikeNano:
> Thanks, @NikeNano. #8590 expands on this change. Would you be ok to keep the discussion on what needs to happen in that PR? We probably don't need two changes.

Sure!

tvalentyn:
@NikeNano should we close this PR since #8590 is merged?

NikeNano:
> @NikeNano should we close this PR since #8590 is merged?

Yes :)

PR #8458, opened 2019-05-01
[BEAM-6356] Use TFRecord to store cache results using PCollection's PCoder

leo-unc:
To enable Python Interactive CacheManager to work with arbitrary types such as Google protobufs,
it should start using pcoder attached to PCollection instead of hard-coded one. It changes the file format for cache results to TFRecord to take advantage of direct writing and reading binary data.

R: @aaltay


aaltay:
LGTM. Thank you.

@qinyeli could you also take a look at it?

qinyeli:
Thanks, Hennadiy!
Verified that it works with Jupyter notebook. LGTM

PR #8460, opened 2019-05-01
Fixing fileio tests for windows

pabloem:
These unit tests relied on `/` being the delimiter in unix, and don't work on windows.
r: @chamikaramj

### DIFF
--- a/sdks/python/apache_beam/io/fileio_test.py
+++ b/sdks/python/apache_beam/io/fileio_test.py
@@ -22,6 +22,7 @@ from __future__ import absolute_import
 import csv
 import io
 import logging
+import os
 import sys
 import unittest
 
@@ -40,7 +41,7 @@ class MatchTest(_TestCaseWithTempDirCleanUp):
 
   def test_basic_two_files(self):
     files = []
-    tempdir = '%s/' % self._new_tempdir()
+    tempdir = '%s%s' % (self._new_tempdir(), os.sep)
 
     # Create a couple files to be matched
     files.append(self._create_temp_file(dir=tempdir))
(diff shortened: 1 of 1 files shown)


sdks/python/apache_beam/io/fileio_test.py
```diff
@@ -40,14 +41,16 @@ class MatchTest(_TestCaseWithTempDirCleanUp):
 
   def test_basic_two_files(self):
     files = []
-    tempdir = '%s/' % self._new_tempdir()
+    tempdir = self._new_tempdir()
```
chamikaramj: Can you use the context manager here ? https://github.com/apache/beam/blob/master/sdks/python/apache_beam/testing/test_utils.py#L41

This is used by other file-based IO tests.

sdks/python/apache_beam/io/fileio_test.py
```diff
@@ -56,25 +59,24 @@ def test_match_all_two_directories(self):
...
+      d = self._new_tempdir()
       directories.append(d)
 
       files.append(self._create_temp_file(dir=d))
       files.append(self._create_temp_file(dir=d))
 
     with TestPipeline() as p:
       files_pc = (p
-                  | beam.Create(directories)
+                  | beam.Create([FileSystems.join(d, '*')
```
chamikaramj: Also, prob. just using os.sep should work (instead of changing the pipeline).

pabloem:
I've switched to use `os.sep`. As for the context manager, I'd like to delay that change, since this one also uses utilities provided by Beam. WDYT?

chamikaramj:
Yeah, that's fine. Probably create a newbie JIRA to unify utilities.

LGTM. Thanks.

PR #8464, opened 2019-05-02
Adding documentation to DirectRunner functions.

pabloem:
r: @aaltay

### DIFF
--- a/sdks/python/apache_beam/runners/direct/evaluation_context.py
+++ b/sdks/python/apache_beam/runners/direct/evaluation_context.py
@@ -32,6 +32,10 @@ from apache_beam.utils import counters
 
 
 class _ExecutionContext(object):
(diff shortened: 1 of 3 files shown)


robertwb:
The "Also marking Fn API Runner as supporting metrics." seems like a much bigger change that should probably be pulled out into its own PR.

pabloem:
oops - I guess I got confused about5 this. Thanks Robert. I think this is ready for review now.

aaltay:
I do not follow where "Also marking Fn API Runner as supporting metrics." is happening. Otherwise change looks good. 

Please update the PR description and self merge.

pabloem:
That marking FnAPI runner...etc stuff is not happening anymore. This is apure pydoc PR now. Thanks Ahmet.

PR #8465, opened 2019-05-02
[BEAM-6605] Deprecate TextIO.readAll and TextIO.ReadAll transform

iemejia:
I also rewrote code to depend directly on ReadFiles instead of ReadAll for consistency (but kept ReadAll in the unit tests so we notice if somehow it breaks).

R: @chamikaramj


iemejia:
Run Java TextIO Performance Test

chamikaramj:
LGTM. Thanks.

PR #8473, opened 2019-05-02
[BEAM-7173] Avoiding schema autodetection by default in WriteToBigQuery

pabloem:
@tvalentyn 
cc: @aaltay

### DIFF
--- a/sdks/python/apache_beam/io/gcp/bigquery.py
+++ b/sdks/python/apache_beam/io/gcp/bigquery.py
@@ -786,11 +786,7 @@ class BigQueryWriteFn(DoFn):
     logging.debug('Creating or getting table %s with schema %s.',
                   table_reference, schema)
 
-    if schema == SCHEMA_AUTODETECT:
-      raise ValueError('Schema auto-detection is not supported for streaming '
-                       'inserts into BigQuery. Only for File Loads.')
-    else:
-      table_schema = self.get_table_schema(schema)
+    table_schema = self.get_table_schema(schema)
 
     if table_reference.projectId is None:
       table_reference.projectId = vp.RuntimeValueProvider.get_value(
@@ -1139,6 +1135,11 @@ bigquery_v2_messages.TableSchema):
           GoogleCloudOptions).project
 
     method_to_use = self._compute_method(p, p.options)
+
+    if (method_to_use == WriteToBigQuery.Method.STREAMING_INSERTS
+        and self.schema == SCHEMA_AUTODETECT):
+      raise ValueError('Schema auto-detection is not supported for streaming '
+                       'inserts into BigQuery. Only for File Loads.')
 
     if method_to_use == WriteToBigQuery.Method.STREAMING_INSERTS:
       # TODO: Support load jobs for streaming pipelines.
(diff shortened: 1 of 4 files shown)


pabloem:
Run Python PreCommit

pabloem:
Run Python PostCommit

tvalentyn:
Run Python Postcommit

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -666,7 +666,12 @@ def _create_table_if_needed(self, table_reference, schema=None):
     logging.debug('Creating or getting table %s with schema %s.',
                   table_reference, schema)
 
-    table_schema = self.get_table_schema(schema)
+    if schema == SCHEMA_AUTODETECT:
```
tvalentyn: Are we assuming that this helper only gets called only in streaming inserts codepath? 
Perhaps we should strengthen the condition to something like: `if schema == SCHEMA_AUTODETECT and  method == STREAMING_INSERTS`, or move this check somewhere up the call stack to fail early?

sdks/python/apache_beam/io/gcp/bigquery_test.py
```diff
@@ -531,6 +531,7 @@ def test_value_provider_transform(self):
            | "WriteWithMultipleDests2" >> beam.io.gcp.bigquery.WriteToBigQuery(
                table=value_provider.StaticValueProvider(
                    str, '%s:%s' % (self.project, output_table_2)),
+               schema=bigquery.SCHEMA_AUTODETECT,
```
tvalentyn: Note that SCHEMA_AUTODETECT if BQ has to ingest `bytes` data into a table that has `bytes` field.
Also, should we have a test with and without autodetect behavior?

tvalentyn:
I think we should loop in Cham, who reviewed previous BQ PRs, perhaps once the tests pass.
//cc: @chamikaramj

pabloem:
Run Python PostCommit

pabloem:
Run Python PostCommit

pabloem:
Run Python PostCommit

pabloem:
Run Python PostCommit

pabloem:
Run Portable_Python PreCommit

pabloem:
Run Python Precommit

pabloem:
Run Python PostCommit

pabloem:
Run Python PostCommit

pabloem:
Run Python PostCommit

pabloem:
Run Python PostCommit

sdks/python/apache_beam/io/gcp/bigquery_test.py
pabloem: There are tests with and without this feature running in that file. WDYT?

tvalentyn: I think that should be fine and we don't need to duplicate every test twice, but you might want to do a one-off run of integration tests where the value of autodetection flag is changed to the opposite value, to make sure there are no surprises.

tvalentyn:
LGTM, thank you! Please squash commits before the merge.

PR #8474, opened 2019-05-02
[BEAM-7210] Fix non-deterministic row access

reuvenlax:


sdks/java/core/src/test/java/org/apache/beam/sdk/transforms/ParDoSchemaTest.java
```diff
@@ -388,7 +388,7 @@ public void testInferredSchemaPipeline() {
                     new DoFn<Inferred, String>() {
                       @ProcessElement
                       public void process(@Element Row row, OutputReceiver<String> r) {
-                        r.output(row.getString(0) + ":" + row.getInt32(1));
+                        r.output(row.getString("stringField") + ":" + row.getInt32("integerField"));
```
mxm: So basically it is not safe to use AutoValue with Schemas. Should we convert the example to use a Pojo instead (like it previously did)? Should we throw an error that schemas do not work deterministically in this case?

reuvenlax: It is perfectly safe. What's not safe is assuming that the order of the schema fields will match the order of the methods in your class. In fact it's technically not safe for POJOs either, as AFAICT Java makes no guarantees there either.

mxm: I just wondered whether we should throw an error when users try to access fields via a non-deterministic position? It seems like this could cause subtle bugs when two fields have the same type.

reuvenlax:
So this is deterministic as long as you resolve the ids beforehand. E.g.
the following is deterministic:

@DefaultFieldSchema(AutoValueSchema.class)
@AutoValue
class Foo {...}
PCollection<Foo> foos = readFoos();
int userIdField = foos.getSchema().indexOf("userId");
foos.apply(ParDo.of(....
        void process(@Element Row row) {
           row.get(userIdField);
        }

In fact all of the schema transforms (Group, Join, etc.) do just this for
efficiency. They resolve the field names into field positions at expansion
time, and then always access by field id inside the actual transform.

Reuven

mxm:
I see. So on Rows users need to ensure that they either know the structure of the Row or go through the Schema which returns the correct index for a field name. Thanks!

reuvenlax:
Yes. I think most users will access the Row by name though, not by index.
If users have an object (POJO, Avro, AutoValue, etc.) with a matching
schema, they can instead just access via that object and not bother with
the Row object at all.

PR #8485, opened 2019-05-03
Fixed bug where fieldIdsAccessed order could be inconsistent with getFieldsAccessed

TheNeuralBit:
In several places we make assumptions that `fieldIdsAccessed` returned a Set with the same iteration order as `getFieldsAccessed`, but this is not the case. Most notably this resulted in an error in `CoGroupTest` when joining by multiple keys. This bug could cause the pipeline to return results with (country, user) rather than (user, country).

This PR makes `fieldIdsAccessed` collect field ids into a List instead to maintain the original order.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/FieldAccessDescriptor.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/FieldAccessDescriptor.java
@@ -23,6 +23,7 @@ import static org.apache.beam.vendor.guava.v20_0.com.google.common.base.Precondi
 import com.google.auto.value.AutoOneOf;
 import com.google.auto.value.AutoValue;
 import java.io.Serializable;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
@@ -130,7 +131,7 @@ public abstract class FieldAccessDescriptor implements Serializable {
   abstract static class Builder {
     abstract Builder setAllFields(boolean allFields);
 
-    abstract Builder setFieldsAccessed(Set<FieldDescriptor> fieldsAccessed);
+    abstract Builder setFieldsAccessed(List<FieldDescriptor> fieldsAccessed);
 
     abstract Builder setNestedFieldsAccessed(
         Map<FieldDescriptor, FieldAccessDescriptor> nestedFieldsAccessedById);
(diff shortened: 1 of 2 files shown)


TheNeuralBit:
Run Java PreCommit

TheNeuralBit:
R: @reuvenlax

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/FieldAccessDescriptor.java
```diff
@@ -316,11 +316,14 @@ public FieldAccessDescriptor withOrderByFieldInsertionOrder() {
...
+  /**
+   * Return the field ids accessed. Should not be called until after {@link #resolve} is called.
+   * Iteration order is consistent with {@link #getFieldsAccessed}.
+   */
+  public List<Integer> fieldIdsAccessed() {
     return getFieldsAccessed().stream()
         .map(FieldDescriptor::getFieldId)
-        .collect(Collectors.toSet());
+        .collect(Collectors.toList());
   }
```
reuvenlax: maybe instead we should make getFieldsAccessed return a List, instead of storing it as a set internally?

TheNeuralBit: Yeah that's probably a good idea. I can add that change here.

TheNeuralBit: How does this look?

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/FieldAccessDescriptor.java
```diff
@@ -379,6 +376,13 @@ public FieldAccessDescriptor resolve(Schema schema) {
       field = fillInMissingQualifiers(field, schema);
       fields.add(field);
     }
+
+    if (!getFieldInsertionOrder()) {
+      // Re-order fields based on field ID, rather than keeping them in insertion order
+      Function<FieldDescriptor, Integer> extract =
+          (Function<FieldDescriptor, Integer> & Serializable) FieldDescriptor::getFieldId;
+      Collections.sort(fields, Comparator.comparing(extract));
```
reuvenlax: Does it work to simply write Collections.sort(fields, Comparator.comparing(FieldDescriptor::getFieldId))?

We needed the verbose version before because it was captured inside the Set class, so had to be serializable. I suspect we can drop back to the simple version now.

TheNeuralBit: Good point, that does seem to work now :+1:

reuvenlax:
lgtm. Thanks for the fix!

PR #8494, opened 2019-05-04
[SQL] Simplify limit collect in BeamEnumerableConverter

amaliujia:
limit collect can be further simplified such that `toRowList` and `toEnumerable` can share the same code.  

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamEnumerableConverter.java
+++ b/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamEnumerableConverter.java
@@ -21,6 +21,7 @@ import static org.apache.beam.vendor.guava.v20_0.com.google.common.base.Precondi
 import static org.apache.calcite.avatica.util.DateTimeUtils.MILLIS_PER_DAY;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
@@ -141,19 +142,25 @@ public class BeamEnumerableConverter extends ConverterImpl implements Enumerable
   }
 
   static List<Row> toRowList(PipelineOptions options, BeamRelNode node) {
+    checkAndFailOnNonDirectRunner(options, "toRowList is only available in DirectRunner.");
+
     if (node instanceof BeamIOSinkRel) {
       throw new UnsupportedOperationException("Does not support BeamIOSinkRel in toRowList.");
     } else if (isLimitQuery(node)) {
-      throw new UnsupportedOperationException("Does not support queries with LIMIT in toRowList.");
+      return new ArrayList<>(limitCollectRows(options, node));
     }
-    return collectRows(options, node).stream().collect(Collectors.toList());
+
+    return new ArrayList<>(collectRows(options, node));
   }
 
   static Enumerable<Object> toEnumerable(PipelineOptions options, BeamRelNode node) {
+    checkAndFailOnNonDirectRunner(
+        options, "SELECT without INSERT is only supported in DirectRunner in SQL Shell.");
+
     if (node instanceof BeamIOSinkRel) {
       return count(options, node);
     } else if (isLimitQuery(node)) {
-      return limitCollect(options, node);
+      return Linq4j.asEnumerable(rowToAvaticaAndUnboxValues(limitCollectRows(options, node)));
     }
     return Linq4j.asEnumerable(rowToAvaticaAndUnboxValues((collectRows(options, node))));
   }
@@ -205,13 +212,6 @@ public class BeamEnumerableConverter extends ConverterImpl implements Enumerable
     long id = options.getOptionsId();
     Queue<Row> values = new ConcurrentLinkedQueue<>();
 
-    checkArgument(
-        options
-            .getRunner()
-            .getCanonicalName()
-            .equals("org.apache.beam.runners.direct.DirectRunner"),
-        "collectRowList is only available in direct runner.");
-
     Collector.globalValues.put(id, values);
 
     runCollector(options, node);
@@ -220,16 +220,18 @@ public class BeamEnumerableConverter extends ConverterImpl implements Enumerable
     return values;
   }
 
-  private static Enumerable<Object> limitCollect(PipelineOptions options, BeamRelNode node) {
-    long id = options.getOptionsId();
-    ConcurrentLinkedQueue<Row> values = new ConcurrentLinkedQueue<>();
-
+  private static void checkAndFailOnNonDirectRunner(PipelineOptions options, String errorMessage) {
     checkArgument(
         options
             .getRunner()
             .getCanonicalName()
             .equals("org.apache.beam.runners.direct.DirectRunner"),
-        "SELECT without INSERT is only supported in DirectRunner in SQL Shell.");
+        errorMessage);
+  }
+
+  private static Queue<Row> limitCollectRows(PipelineOptions options, BeamRelNode node) {
+    long id = options.getOptionsId();
+    ConcurrentLinkedQueue<Row> values = new ConcurrentLinkedQueue<>();
 
     int limitCount = getLimitCount(node);
 
@@ -242,7 +244,7 @@ public class BeamEnumerableConverter extends ConverterImpl implements Enumerable
       values.remove();
     }
 
-    return Linq4j.asEnumerable(rowToAvaticaAndUnboxValues(values));
+    return values;
   }
 
   private static class Collector extends DoFn<Row, Void> {
--- a/sdks/java/extensions/sql/src/test/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamEnumerableConverterTest.java
+++ b/sdks/java/extensions/sql/src/test/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamEnumerableConverterTest.java
@@ -44,8 +44,10 @@ import org.apache.calcite.jdbc.JavaTypeFactoryImpl;
 import org.apache.calcite.linq4j.Enumerable;
 import org.apache.calcite.linq4j.Enumerator;
 import org.apache.calcite.plan.RelOptCluster;
+import org.apache.calcite.plan.RelTraitSet;
 import org.apache.calcite.plan.volcano.VolcanoPlanner;
 import org.apache.calcite.prepare.RelOptTableImpl;
+import org.apache.calcite.rel.RelCollations;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeSystem;
 import org.apache.calcite.rex.RexBuilder;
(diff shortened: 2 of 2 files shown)


amaliujia:
@akedin @apilloud

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamEnumerableConverter.java
```diff
@@ -141,19 +141,25 @@ public static PipelineOptions createPipelineOptions(Map<String, String> map) {
...
 
   static Enumerable<Object> toEnumerable(PipelineOptions options, BeamRelNode node) {
+    checkAndFailOnNonDirectRunner(
+        options, "SELECT without INSERT is only supported in DirectRunner in SQL Shell.");
+
     if (node instanceof BeamIOSinkRel) {
       return count(options, node);
     } else if (isLimitQuery(node)) {
-      return limitCollect(options, node);
+      return Linq4j.asEnumerable(rowToAvaticaAndUnboxValues(limitCollectRows(options, node)));
```
akedin: Looking at this I think it is possible to refactor this even further. `toRowList()` and `toEnumerable()` are essentially the same thing now with minor differences. Would it be possible to do the following?

* add support for count to `toRowList()` as mentioned in another comment;
* don't call `limitCollectRows()` here, but call `toRowList()` instead;
* this would probably make this method a one-line call to `Linq4j.asEnumerable(rowToAvaticaAndUnboxValues..`

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamEnumerableConverter.java
```diff
@@ -141,19 +141,25 @@ public static PipelineOptions createPipelineOptions(Map<String, String> map) {
   }
 
   static List<Row> toRowList(PipelineOptions options, BeamRelNode node) {
+    checkAndFailOnNonDirectRunner(options, "toRowList is only available in DirectRunner.");
+
     if (node instanceof BeamIOSinkRel) {
       throw new UnsupportedOperationException("Does not support BeamIOSinkRel in toRowList.");
```
akedin: Can we support count here? I.e. if we make `count()` return a singleton `Collection<Row>` with a single field and element?

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamEnumerableConverter.java
```diff
@@ -220,16 +219,18 @@ private static void runCollector(PipelineOptions options, BeamRelNode node) {
...
     checkArgument(
         options
             .getRunner()
             .getCanonicalName()
             .equals("org.apache.beam.runners.direct.DirectRunner"),
-        "SELECT without INSERT is only supported in DirectRunner in SQL Shell.");
+        errorMessage);
+  }
+
+  private static Queue<Row> limitCollectRows(PipelineOptions options, BeamRelNode node) {
```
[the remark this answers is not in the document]
amaliujia: I think the queue is to guarantee order of rows for cases where ORDER BY is used?

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamEnumerableConverter.java
amaliujia: There is no direct usage on it: people when want access the count can call `toEnumerable`. 

`toRowList` is used for cases when plain Row is required. 

Count, instead, is an int without associated schema.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamEnumerableConverter.java
```diff
@@ -141,19 +141,25 @@ public static PipelineOptions createPipelineOptions(Map<String, String> map) {
   }
 
   static List<Row> toRowList(PipelineOptions options, BeamRelNode node) {
+    checkAndFailOnNonDirectRunner(options, "toRowList is only available in DirectRunner.");
+
     if (node instanceof BeamIOSinkRel) {
       throw new UnsupportedOperationException("Does not support BeamIOSinkRel in toRowList.");
     } else if (isLimitQuery(node)) {
-      throw new UnsupportedOperationException("Does not support queries with LIMIT in toRowList.");
+      return limitCollectRows(options, node).stream().collect(Collectors.toList());
```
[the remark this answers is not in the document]
amaliujia: updated.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamEnumerableConverter.java
amaliujia: limit collect relies on `limitRun`, which requires cancel pipeline based on collected value size. That's the reason limit collect and collect are in different code path.

akedin: I am not sure I understand. I am not proposing to re-use the code between `limitCollect` and `collect` parts. What I mean is:

Change `toRowList()` to something like:

```
static List<Row> toRowList(PipelineOptions options, BeamRelNode node) {
    Collection<Row> rows;
    if (node instanceof BeamIOSinkRel) {
     rows = count();
    } else if (isLimitQuery(node)) {
      rows = limitCollectRows(options, node);
    } else {
      rows = collectRows(options, node);
    }
    return new ArrayList<>(rows);
  }
```

Then  `toEnumerable` may become something like this:

```
static Enumerable<Object> toEnumerable(PipelineOptions options, BeamRelNode node) {
    Collection<Row> rows = toRowList(options, node);
    return Linq4j.asEnumerable(rowToAvaticaAndUnboxValues(rows));
}
```

I may be missing something though.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamEnumerableConverter.java
akedin: I am thinking this would allow to reduce the code duplication, see my other comment.

All relational operations though have a schema. My understanding is that when I'm running `SELECT 1` I still get a relation as the result. Same for `SELECT COUNT(x)`.

Now `count()` here is an int without an associated schema in the current implementation which doesn't allow to use it in `toRowList()` as you mentioned. However if you wrap the value in a row with a schema, it probably should work. And even if it's not used explicitly it might allow to re-use the code between `toRowList` and `toEnumerable`.

And in `toEnumerable`, the call to `rowToAvaticaAndUnboxValues` should handle the single row with a single value as well, so that you don't have to special case it there.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamEnumerableConverter.java
amaliujia: I see.

What you are missing is `count` does not collect rows: it uses metrics API to count (https://github.com/apache/beam/blob/master/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamEnumerableConverter.java#L370).

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamEnumerableConverter.java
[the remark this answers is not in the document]
amaliujia: I think there is an order to maintain. Think about this case:

SELECT * FROM a ORDER BY a.col_a DESC LIMIT 5. 

The result should be displayed in SQL shell with a order, which means our collected rows are in a queue (or is a list), rather than a bag of objects.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamEnumerableConverter.java
amaliujia: It might be feasible to re-implement count to make it return a `List<Row>{ROW<int>(count)}`, but it seems over-kill for achieving the goal of merging toRowList and toEnumerable.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamEnumerableConverter.java
amaliujia: Ah this is the last comment I saw, but it explains why there is a propose to merge count for both API.

I wouldn't call this proposal as a way to reduce code duplication: there is no duplicated code now because toRowList does not support count()

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamEnumerableConverter.java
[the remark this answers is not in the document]
akedin: That is true for this case. However the fact that the results internally are ordered is just an implementation detail. Nothing stops us from using queues only for orderby-limit case but use multisets, trees, or anything else instead for other classes of queries. The `toEnumerable()` API doesn't provide the ordering guarantees in general, it's defined at a different level.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamEnumerableConverter.java
amaliujia: Basically if there is a need to support `count` for `toRowList`, it's better to use the same `count` for both API. 

It's just debatable that if `count` is needed for `toRowList`.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamEnumerableConverter.java
akedin: It does not matter how `count()` calculates the results, it can return a constant or a random result or call some service in the cloud. It's a function in a relational programming model, so the result is expected to be a relation. Sure, it feels not necessary to wrap a single value into a row so that it can be immediately unwrapped (in one case). But my argument is:
 * it is natural to define a function in such environment as returning Rows, so it will not be unexpected to the readers;
 * it is harder to read two duplicate versions of semantically the same code and figure out the differences, then to understand why we have to wrap a single value in a row;
 * no special casing;
 * it is less code, less things to read, break and test;

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamEnumerableConverter.java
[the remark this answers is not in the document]
amaliujia: I can try to say if `Collection` interface will not affect implementation order. If it does maintain the order, I can switch to `Collection` and interpreter it after.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamEnumerableConverter.java
akedin: It's not needed. But we are in a situation like this:
 * it is not needed, so we have a special case for it which throws an exception, and if we need it we will have to implement it;
 * it's free to add support for it (-ish, modulo wrapping/unwrapping the value into a row) which remove special casing and make it simpler;

PR #8499, opened 2019-05-05
[BEAM-5865] Create optional auto-balancing sharding function for Flink

JozoVilcek:
This change has to parts.
* modify WriteFiles to be able to specify sharding function
* introduce an option for FlinkRunner to use sharding function which distributes shards evenly among workers

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkPipelineOptions.java
+++ b/runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkPipelineOptions.java
@@ -237,7 +237,12 @@ public interface FlinkPipelineOptions
   void setAllowNonRestoredState(Boolean allowNonRestoredState);
 
   @Description(
-      "Flag indicating whether auto-balance sharding for WriteFiles transform should be enabled.")
+      "Flag indicating whether auto-balance sharding for WriteFiles transform should be enabled. "
+          + "This might prove useful in streaming use-case, where pipeline needs to write quite many events "
+          + "into files, typically divided into N shards. Default behavior on Flink would be, that some workers "
+          + "will receive more shards to take care of than others. This cause workers to go out of balance in "
+          + "terms of processing backlog and memory usage. Enabling this feature will make shards to be spread "
+          + "evenly among available workers in improve throughput and memory usage stability.")
   @Default.Boolean(false)
   Boolean isAutoBalanceWriteFilesShardingEnabled();
 
--- a/runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkPipelineOptions.java
+++ b/runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkPipelineOptions.java
@@ -235,4 +235,11 @@ public interface FlinkPipelineOptions
   Boolean getAllowNonRestoredState();
 
   void setAllowNonRestoredState(Boolean allowNonRestoredState);
+
+  @Description(
+      "Flag indicating whether auto-balance sharding for WriteFiles transform should be enabled.")
+  @Default.Boolean(false)
+  Boolean isAutoBalanceWriteFilesShardingEnabled();
+
+  void setAutoBalanceWriteFilesShardingEnabled(Boolean autoBalanceWriteFilesShardingEnabled);
 }
--- a/runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkStreamingPipelineTranslator.java
+++ b/runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkStreamingPipelineTranslator.java
@@ -17,27 +17,47 @@
  */
 package org.apache.beam.runners.flink;
 
-import java.util.Collections;
+import static org.apache.beam.runners.core.construction.PTransformTranslation.WRITE_FILES_TRANSFORM_URN;
+
+import java.io.IOException;
+import java.io.Serializable;
+import java.nio.ByteBuffer;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.concurrent.ThreadLocalRandom;
 import org.apache.beam.runners.core.construction.PTransformReplacements;
 import org.apache.beam.runners.core.construction.PTransformTranslation;
+import org.apache.beam.runners.core.construction.ReplacementOutputs;
 import org.apache.beam.runners.core.construction.UnconsumedReads;
 import org.apache.beam.runners.core.construction.WriteFilesTranslation;
 import org.apache.beam.sdk.Pipeline;
+import org.apache.beam.sdk.coders.Coder;
+import org.apache.beam.sdk.coders.CoderException;
+import org.apache.beam.sdk.coders.ShardedKeyCoder;
+import org.apache.beam.sdk.coders.VarIntCoder;
 import org.apache.beam.sdk.io.FileBasedSink;
+import org.apache.beam.sdk.io.ShardingFunction;
 import org.apache.beam.sdk.io.WriteFiles;
 import org.apache.beam.sdk.io.WriteFilesResult;
 import org.apache.beam.sdk.options.PipelineOptions;
 import org.apache.beam.sdk.runners.AppliedPTransform;
+import org.apache.beam.sdk.runners.PTransformMatcher;
 import org.apache.beam.sdk.runners.PTransformOverrideFactory;
 import org.apache.beam.sdk.runners.TransformHierarchy;
 import org.apache.beam.sdk.transforms.PTransform;
+import org.apache.beam.sdk.util.CoderUtils;
 import org.apache.beam.sdk.values.PCollection;
 import org.apache.beam.sdk.values.PCollectionView;
 import org.apache.beam.sdk.values.PValue;
+import org.apache.beam.sdk.values.ShardedKey;
 import org.apache.beam.sdk.values.TupleTag;
 import org.apache.beam.vendor.guava.v20_0.com.google.common.annotations.VisibleForTesting;
+import org.apache.flink.runtime.state.KeyGroupRangeAssignment;
 import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
 import org.apache.flink.util.Preconditions;
 import org.slf4j.Logger;
@@ -169,6 +189,29 @@ class FlinkStreamingPipelineTranslator extends FlinkPipelineTranslator {
     }
   }
 
+  static PTransformMatcher writeFilesNeedsOverrides() {
+    return application -> {
+      if (WRITE_FILES_TRANSFORM_URN.equals(
+          PTransformTranslation.urnForTransformOrNull(application.getTransform()))) {
+        try {
+          FlinkPipelineOptions options =
+              application.getPipeline().getOptions().as(FlinkPipelineOptions.class);
+          ShardingFunction shardingFn =
+              WriteFilesTranslation.getShardingFunction((AppliedPTransform) application);
+          return WriteFilesTranslation.isRunnerDeterminedSharding((AppliedPTransform) application)
+              || (options.isAutoBalanceWriteFilesShardingEnabled() && shardingFn == null);
+        } catch (IOException exc) {
+          throw new RuntimeException(
+              String.format(
+                  "Transform with URN %s failed to parse: %s",
+                  WRITE_FILES_TRANSFORM_URN, application.getTransform()),
+              exc);
+        }
+      }
+      return false;
+    };
+  }
+
   @VisibleForTesting
   static class StreamingShardedWriteFactory<UserT, DestinationT, OutputT>
       implements PTransformOverrideFactory<
@@ -211,6 +254,23 @@ class FlinkStreamingPipelineTranslator extends FlinkPipelineTranslator {
         if (WriteFilesTranslation.isWindowedWrites(transform)) {
           replacement = replacement.withWindowedWrites();
         }
+
+        if (options.isAutoBalanceWriteFilesShardingEnabled()) {
+          Preconditions.checkArgument(
+              options.getParallelism() > 0,
+              "Parallelism is required to be set in FlinkPipelineOptions when isAutoBalanceWriteFilesShardingEnabled");
+          Preconditions.checkArgument(
+              options.getMaxParallelism() > 0,
+              "MaxParallelism is required to be set in FlinkPipelineOptions when isAutoBalanceWriteFilesShardingEnabled");
+
+          replacement =
+              replacement.withShardingFunction(
+                  new FlinkAutoBalancedShardKeyShardingFunction<>(
+                      options.getParallelism(),
+                      options.getMaxParallelism(),
+                      sink.getDynamicDestinations().getDestinationCoder()));
+        }
+
         return PTransformReplacement.of(
             PTransformReplacements.getSingletonMainInput(transform),
             replacement.withNumShards(numShards));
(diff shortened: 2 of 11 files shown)


JozoVilcek:
retest this please

JozoVilcek:
@reuvenlax @mxm

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkStreamingPipelineTranslator.java
```diff
@@ -211,6 +254,23 @@ boolean canTranslate(T transform, FlinkStreamingTranslationContext context) {
...
           replacement = replacement.withWindowedWrites();
         }
+
+        if (options.isAutoBalanceWriteFilesShardingEnabled()) {
+          Preconditions.checkArgument(
+              options.getParallelism() > 0,
+              "Parallelism is required to be set in FlinkPipelineOptions when isAutoBalanceWriteFilesShardingEnabled");
+          Preconditions.checkArgument(
+              options.getMaxParallelism() > 0,
+              "MaxParallelism is required to be set in FlinkPipelineOptions when isAutoBalanceWriteFilesShardingEnabled");
```
mxm: Why is this required, couldn't we assume `maxParallelism == parallelism` if not provided?

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkStreamingPipelineTranslator.java
```diff
@@ -222,7 +282,117 @@ boolean canTranslate(T transform, FlinkStreamingTranslationContext context) {
...
     public Map<PValue, ReplacementOutput> mapOutputs(
         Map<TupleTag<?>, PValue> outputs, WriteFilesResult<DestinationT> newOutput) {
-      return Collections.emptyMap();
+      return ReplacementOutputs.tagged(outputs, newOutput);
+    }
+  }
+
+  /**
+   * Flink has a known problem of unevenly assigning keys to key groups (and then workers) for cases
+   * that number of keys is not >> key groups. This is typical scenario when writing files, where
```
mxm: I wouldn't say this is a Flink problem, it just comes with the nature of hashing algorithms (Murmur hash in this case) when the number of keys is not much bigger than the number of partitions.

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkStreamingPipelineTranslator.java
```diff
@@ -222,7 +282,117 @@ boolean canTranslate(T transform, FlinkStreamingTranslationContext context) {
...
+      if (cache.size() < shardNumber) {
+        for (int i = 0; i < shardNumber; i++) {
+          generateInternal(new CacheKey(destinationKey, i));
+        }
+      }
+
+      return generateInternal(new CacheKey(destinationKey, shardNumber));
+    }
+
+    private ShardedKey<Integer> generateInternal(CacheKey key) {
```
mxm: Rename to `getShardedKey` and add a Precondition that the cache has been created upfront.

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkStreamingPipelineTranslator.java
```diff
@@ -222,7 +282,117 @@ boolean canTranslate(T transform, FlinkStreamingTranslationContext context) {
...
+
+      int destinationKey =
+          Arrays.hashCode(CoderUtils.encodeToByteArray(destinationCoder, destination));
+      // we need to ensure that keys are always stable no matter at which worker they
+      // are created and what is an order of observed shard numbers
+      if (cache.size() < shardNumber) {
+        for (int i = 0; i < shardNumber; i++) {
+          generateInternal(new CacheKey(destinationKey, i));
+        }
+      }
```
mxm: Could be refactored to a method named `generateInitialShardedKeys` which only takes care of the initial generation.

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkStreamingPipelineTranslator.java
```diff
@@ -222,7 +282,117 @@ boolean canTranslate(T transform, FlinkStreamingTranslationContext context) {
...
+      while (true) {
+        salt++;
+        ShardedKey<Integer> shk = ShardedKey.of(Objects.hash(key.key, salt), key.shard);
+        int targetPartition = key.shard % parallelism;
+
+        // create effective key in the same way Beam/Flink will do so we can see if it gets
+        // allocated to the partition we want
+        ByteBuffer effectiveKey;
+        try {
+          byte[] bytes = CoderUtils.encodeToByteArray(ShardedKeyCoder.of(VarIntCoder.of()), shk);
```
mxm: use `FlinkKeyUtils`

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkStreamingPipelineTranslator.java
```diff
@@ -222,7 +282,117 @@ boolean canTranslate(T transform, FlinkStreamingTranslationContext context) {
...
+   * workers by Flink.
+   */
+  private static class FlinkAutoBalancedShardKeyShardingFunction<UserT, DestinationT>
+      implements ShardingFunction<UserT, DestinationT> {
+
+    private final int parallelism;
+    private final int maxParallelism;
+    private final Coder<DestinationT> destinationCoder;
+    private final Map<CacheKey, ShardedKey<Integer>> cache = new HashMap<>();
+    private final Set<Integer> usedSalts = new HashSet<>();
```
mxm: Does this still serve a purpose?

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkStreamingPipelineTranslator.java
```diff
@@ -222,7 +282,117 @@ boolean canTranslate(T transform, FlinkStreamingTranslationContext context) {
...
+      this.maxParallelism = maxParallelism;
+      this.destinationCoder = destinationCoder;
+    }
+
+    @Override
+    public ShardedKey<Integer> assignShardKey(
+        DestinationT destination, UserT element, int shardCount) throws Exception {
+
+      if (shardNumber == -1) {
+        shardNumber = ThreadLocalRandom.current().nextInt(shardCount);
```
mxm: Is there any benefit to starting with a random shard number as opposed to `0`?

sdks/java/core/src/test/java/org/apache/beam/sdk/io/WriteFilesTest.java
```diff
@@ -278,6 +280,36 @@ public void testExpandShardedWrite() throws IOException {
         WriteFiles.to(makeSimpleSink()).withNumShards(20));
   }
 
+  /**
+   * Test that WriteFiles with a configured sharding function distributes elements according to
+   * assigned key. Test use simple sharding which co-locate same elements into same shard.
+   */
+  @Test
+  @Category(NeedsRunner.class)
+  public void testCustomShardingFunctionWrite() throws IOException {
```
mxm: Could we also add a test for `FlinkAutoBalancedShardKeyShardingFunction`?

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkStreamingPipelineTranslator.java
```diff
@@ -222,7 +282,117 @@ boolean canTranslate(T transform, FlinkStreamingTranslationContext context) {
...
+          effectiveKey = ByteBuffer.wrap(bytes);
+        } catch (CoderException e) {
+          throw new RuntimeException(e);
+        }
+
+        int partition =
+            KeyGroupRangeAssignment.assignKeyToParallelOperator(
+                effectiveKey, maxParallelism, parallelism);
+
+        if (partition == targetPartition && !usedSalts.contains(salt)) {
```
mxm: I know this will succeed by likelihood but I'm wondering whether we should put an upper limit here in case the loop does not return.

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkStreamingPipelineTranslator.java
JozoVilcek: I think Flik does use `KeyGroupRangeAssignment .computeDefaultMaxParallelism()` and not directly parallelism. I guess I was just being conservative. I will make it work with defaults

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkStreamingPipelineTranslator.java
JozoVilcek: True. Does this sounds better?
```
Flink assigns elements to parallel operators (workers) based their key group. Key group is determined based on a Murmur hash of element's key. This will have skew distribution in case when number of keys is not >> key groups, which is typical scenario when writing files, where one do not want to end up with too many small files.
```

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkStreamingPipelineTranslator.java
JozoVilcek: I just took it from WriteFiles, where it says
```
// We want to desynchronize the first record sharding key for each instance of
// ApplyShardingKey, so records in a small PCollection will be statistically balanced.
```

sdks/java/core/src/test/java/org/apache/beam/sdk/io/WriteFilesTest.java
JozoVilcek: Yes

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkStreamingPipelineTranslator.java
mxm: Ah that's true. I wasn't aware that we set a higher max parallelism, just in case the user wants to scale up.

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkStreamingPipelineTranslator.java
mxm: Sounds good!

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkStreamingPipelineTranslator.java
```diff
@@ -325,74 +357,63 @@ private FlinkAutoBalancedShardKeyShardingFunction(
...
+          }
+          ShardedKey<Integer> shk = ShardedKey.of(Objects.hash(key, salt), shard);
+          int targetPartition = shard % parallelism;
+
+          // create effective key in the same way Beam/Flink will do so we can see if it gets
+          // allocated to the partition we want
+          ByteBuffer effectiveKey;
+          try {
+            byte[] bytes = CoderUtils.encodeToByteArray(ShardedKeyCoder.of(VarIntCoder.of()), shk);
+            effectiveKey = ByteBuffer.wrap(bytes);
```
mxm: Could you use `FlinkKeyUtils.encodeKey(K key, Coder<K> keyCoder)` here? Just to ensure that this work correctly in case we change the encoding.

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkStreamingPipelineTranslator.java
```diff
@@ -325,74 +357,63 @@ private FlinkAutoBalancedShardKeyShardingFunction(
...
+          } catch (CoderException e) {
+            throw new RuntimeException(e);
+          }
+
+          int partition =
+              KeyGroupRangeAssignment.assignKeyToParallelOperator(
+                  effectiveKey, maxParallelism, parallelism);
+
+          if (partition == targetPartition && !usedSalts.contains(salt)) {
```
mxm: Why do we need `!usedSalts.contains(salt)`?

runners/flink/src/test/java/org/apache/beam/runners/flink/FlinkStreamingPipelineTranslatorTest.java
```diff
@@ -0,0 +1,131 @@
...
+/** Tests if overrides are properly applied. */
+public class FlinkStreamingPipelineTranslatorTest {
+
+  @Test
+  public void testAutoBalanceShardKeyResolvesMaxParallelism() {
+
+    assertThat(
+        new FlinkAutoBalancedShardKeyShardingFunction<>(3, -1, StringUtf8Coder.of())
+            .getMaxParallelism(),
+        equalTo(128));
```
mxm: `equalTo(FlinkKeyGroupAssignment.getMaxParallelism())`?

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkStreamingPipelineTranslator.java
JozoVilcek: shard key identity is defined by destination (it's hash), shardNumber and salt. Salt is integer from we are searching for, to make sure the "real key" (destination, shardNumber) is landed to correct worker. Salt is restarted for each shard key to zero. 

I guess I was paranoid in the beginning to avoid key collisions. But right now, I do not see a reason why not to drop it.

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkStreamingPipelineTranslator.java
JozoVilcek: Sorry, I forgot about this. Will do

JozoVilcek: `FlinkKeyUtils` is package private. I assume it is OK to make it public?

runners/flink/src/test/java/org/apache/beam/runners/flink/FlinkStreamingPipelineTranslatorTest.java
JozoVilcek: I can not find `` ... Can only do `equalTo(KeyGroupRangeAssignment.computeDefaultMaxParallelism(parallelism))` what is done in sharding function, if you like it better than hardcoded value.

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkStreamingPipelineTranslator.java
```diff
@@ -211,6 +261,20 @@ boolean canTranslate(T transform, FlinkStreamingTranslationContext context) {
...
+              replacement.withShardingFunction(
+                  new FlinkAutoBalancedShardKeyShardingFunction<>(
+                      options.getParallelism(),
+                      options.getMaxParallelism(),
+                      sink.getDynamicDestinations().getDestinationCoder()));
+        }
+
         return PTransformReplacement.of(
             PTransformReplacements.getSingletonMainInput(transform),
             replacement.withNumShards(numShards));
```
JozoVilcek: I have a problem here. I found out that now, my explicit number of shards gets always override here to runner determined. 
I see that `replacement` is recreated from scratch and and must be created full. It seems that `num shards` is not explicit in payload but gets injected here and there during the transform `expand` (just guessing here)?.
It seems I can not just put sharding function to existing WriteFiles context, but needs to create fresh new one. Am I right?

I need a help here. Is there an easy way of accessing `numShards`? Do I need to change `WriteFilesPayload` and make it explicit ? @reuvenlax what approach would you suggest to resolve this?

runners/flink/src/test/java/org/apache/beam/runners/flink/FlinkStreamingPipelineTranslatorTest.java
mxm: Yes, that's what I meant.

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkStreamingPipelineTranslator.java
JozoVilcek: Never mind. I talked with @mxm and noticed that overrides have access to proto and WriteFiles transform. I got derailed by runner determined sharding code and thought that accessing stuff from proto is the only way. But it seems it is now so I should be good here

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkPipelineOptions.java
```diff
@@ -235,4 +235,11 @@
   Boolean getAllowNonRestoredState();
 
   void setAllowNonRestoredState(Boolean allowNonRestoredState);
+
+  @Description(
+      "Flag indicating whether auto-balance sharding for WriteFiles transform should be enabled.")
```
mxm: Could you expand this comment to explain why this is useful and when?

mxm:
@reuvenlax Any comments on this? I think this is good to merge otherwise.

mxm:
@JozoVilcek Apologies, but could you rebase once more?

mxm:
Thanks! If you want this to be included in the release, you will have to backport this to release-2.13.0.

JozoVilcek:
Next release is fine. I can do patched versions privately. Thank for the merge and support!

PR #8507, opened 2019-05-06
Add UsesSchema category to Schema transform tests

iemejia:
I decided to categorize them class wide in case someone forgets to add the annotation on new methods.
For some context this comes from testing support of both Spark and Flink runners.

R: @reuvenlax


reuvenlax:
lgtm

PR #8509, opened 2019-05-06
Skipping fileio tests on windows

pabloem:
r: @chamikaramj


pabloem:
Using method from unittestdocs: https://docs.python.org/2/library/unittest.html#skipping-tests-and-expected-failures

chamikaramj:
LGTM

PR #8512, opened 2019-05-06
[BEAM-7235] StreamingDataflowWorker creates commit stream only when commit available

scwhittle:
Otherwise it can create a commit stream that may exceed it's deadline and spam the logs.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/StreamingDataflowWorker.java
+++ b/runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/StreamingDataflowWorker.java
@@ -1459,7 +1459,9 @@ public class StreamingDataflowWorker {
     Commit commit = null;
     while (running.get()) {
       // Batch commits as long as there are more and we can fit them in the current request.
-      CommitWorkStream commitStream = streamPool.getStream();
+      // We lazily initialize the commit stream to make sure that we only create one after
+      // we have a commit.
+      CommitWorkStream commitStream = null;
       int commits = 0;
       while (running.get()) {
         // There may be a commit left over from the previous iteration but if not, pull one.
@@ -1487,6 +1489,9 @@ public class StreamingDataflowWorker {
         final Windmill.WorkItemCommitRequest request = commit.getRequest();
         final int size = commit.getSize();
         commit.getWork().setState(State.COMMITTING);
+        if (commitStream == null) {
+          commitStream = streamPool.getStream();
+        }
         if (commitStream.commitWorkItem(
             state.computationId,
             request,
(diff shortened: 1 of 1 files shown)


scwhittle:
R: @lukecwik

runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/StreamingDataflowWorker.java
```diff
@@ -1459,7 +1459,9 @@ private void streamingCommitLoop() {
     Commit commit = null;
     while (running.get()) {
       // Batch commits as long as there are more and we can fit them in the current request.
-      CommitWorkStream commitStream = streamPool.getStream();
+      // We lazily initialize the commit stream to make sure that we only create one after
+      // we have a commit.
+      CommitWorkStream commitStream = null;
```
lukecwik: Why not move this declaration below to line 1492 instead of assigning null here.

Then you wouldn't have to check if things are null or not.

scwhittle: We want to add a batch of things to the same stream and then call flush.
If the declaration was moved to 1492, the commit stream would only be used for a single call to CommitWorkItem.

It would be nice if this method could be broken into some helper methods.  However I don't see a clean way to do so because of the logic where a unbatched Commit needs to be consumed by a subsequent stream.

lukecwik: Thanks, makes sense.

runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/StreamingDataflowWorker.java
```diff
@@ -1508,8 +1513,10 @@ private void streamingCommitLoop() {
           break;
         }
       }
-      commitStream.flush();
-      streamPool.releaseStream(commitStream);
+      if (commitStream) {
```
lukecwik:
```suggestion
      if (commitStream != null) {
```

scwhittle: Oops, fixed,

I ran the following ./gradlew :beam-runners-google-cloud-dataflow-java:build does that not actually build this? Or perhaps I ran it in the wrong branch.

lukecwik: You built the runner that submits the pipeline and not the worker component.

I think you meant to do `./gradlew :beam-runners-google-cloud-dataflow-java:worker:build`

PR #8515, opened 2019-05-07
[BEAM-7238] Make sfl4j bindings runtimeOnly/testRuntimeOnly

iemejia:
R: @mxm

### DIFF
--- a/runners/core-java/build.gradle
+++ b/runners/core-java/build.gradle
@@ -45,6 +45,6 @@ dependencies {
   shadowTest library.java.mockito_core
   shadowTest library.java.junit
   shadowTest library.java.slf4j_api
-  shadowTest library.java.slf4j_simple
   shadowTest library.java.jackson_dataformat_yaml
+  testRuntimeOnly library.java.slf4j_simple
 }
--- a/sdks/java/extensions/sql/build.gradle
+++ b/sdks/java/extensions/sql/build.gradle
@@ -86,12 +86,12 @@ dependencies {
   provided project(path: ":beam-sdks-java-io-kafka", configuration: "shadow")
   provided project(path: ":beam-sdks-java-io-google-cloud-platform", configuration: "shadow")
   provided library.java.kafka_clients
-  shadowTest library.java.slf4j_jdk14
   shadowTest library.java.junit
   shadowTest library.java.hamcrest_core
   shadowTest library.java.hamcrest_library
   shadowTest library.java.mockito_core
   shadowTest library.java.quickcheck_core
+  shadowTestRuntimeClasspath library.java.slf4j_jdk14
 
   // Dependencies that we don't directly reference
   permitUnusedDeclared "com.jayway.jsonpath:json-path:2.4.0"
--- a/sdks/java/extensions/sql/datacatalog/build.gradle
+++ b/sdks/java/extensions/sql/datacatalog/build.gradle
@@ -39,7 +39,7 @@ dependencies {
   provided project(path: ":beam-sdks-java-io-google-cloud-platform", configuration: "shadow")
   provided library.java.vendored_guava_20_0
   provided library.java.slf4j_api
-  provided library.java.slf4j_simple
+  shadowTestRuntimeClasspath library.java.slf4j_simple
 }
 
 task runDataCatalogExample(type: JavaExec) {
--- a/runners/flink/flink_runner.gradle
+++ b/runners/flink/flink_runner.gradle
@@ -109,7 +109,6 @@ dependencies {
   shadowTest library.java.junit
   shadowTest library.java.mockito_core
   shadowTest library.java.google_api_services_bigquery
-  shadowTest library.java.slf4j_simple
   shadowTest project(path: ":beam-sdks-java-io-google-cloud-platform", configuration: "shadow")
   shadowTest library.java.jackson_dataformat_yaml
   shadowTest "org.apache.flink:flink-core:$flink_version:tests"
@@ -117,6 +116,7 @@ dependencies {
   shadowTest "org.apache.flink:flink-streaming-java_2.11:$flink_version:tests"
   shadowTest "org.apache.flink:flink-test-utils_2.11:$flink_version"
   shadowTest project(":beam-sdks-java-harness")
+  testRuntimeOnly library.java.slf4j_simple
   validatesRunner project(path: ":beam-sdks-java-core", configuration: "shadowTest")
   validatesRunner project(path: ":beam-runners-core-java", configuration: "shadowTest")
   validatesRunner project(path: project.path, configuration: "shadow")
--- a/runners/java-fn-execution/build.gradle
+++ b/runners/java-fn-execution/build.gradle
@@ -41,7 +41,7 @@ dependencies {
   testCompile library.java.hamcrest_core
   testCompile library.java.hamcrest_library
   testCompile library.java.mockito_core
-  testCompile library.java.slf4j_simple
+  testRuntimeOnly library.java.slf4j_simple
 }
 
 test {
--- a/sdks/java/core/build.gradle
+++ b/sdks/java/core/build.gradle
@@ -82,7 +82,6 @@ dependencies {
   provided library.java.hamcrest_library
   shadowTest library.java.jackson_dataformat_yaml
   shadowTest library.java.guava_testlib
-  shadowTest library.java.slf4j_jdk14
   shadowTest library.java.mockito_core
   shadowTest library.java.hamcrest_core
   shadowTest library.java.hamcrest_library
@@ -90,4 +89,5 @@ dependencies {
   shadowTest library.java.quickcheck_core
   shadowTest library.java.avro_tests
   shadowTest library.java.zstd_jni
+  testRuntimeOnly library.java.slf4j_jdk14
 }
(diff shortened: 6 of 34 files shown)


runners/core-java/build.gradle
```diff
@@ -45,6 +45,6 @@ dependencies {
   shadowTest library.java.mockito_core
   shadowTest library.java.junit
   shadowTest library.java.slf4j_api
-  shadowTest library.java.slf4j_simple
   shadowTest library.java.jackson_dataformat_yaml
+  testRuntimeOnly library.java.slf4j_simple
```
mxm: I wonder, is `testRuntimeOnly` safe because it circumvents the shading? I suppose we do not shade `sl4j_simple`.

iemejia: Seems so. In any case is as you said, we should not shade it just guarantee is available at runtime.

lukecwik: Right now most modules don't set [testShadowJar](https://github.com/apache/beam/blob/a6a6b8efcbd000299e1ffb808eae131156198455/buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy#L96) to be true. There was some work to make a few modules test that way but it hasn't been done for the rest. I think people assumed that vendoring would replace shading everywhere (but vendoring work has stalled) so testing with the shadow jar stalled.

You can use the [shadowTestRuntimeClasspath](https://github.com/apache/beam/blob/a6a6b8efcbd000299e1ffb808eae131156198455/buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy#L917) for projects which run tests using the shadowTest jar. Otherwise the testRuntimeOnly is the right place to do this. (Should we rename shadowTestRuntimeClasspath to be shadowTestRuntimeOnly to match the naming convention within Gradle already?)

sdks/java/extensions/sql/jdbc/build.gradle
```diff
@@ -34,12 +34,13 @@ dependencies {
   compile project(path: ":beam-sdks-java-extensions-sql", configuration: "shadow")
   compile "jline:jline:2.14.6"
   compile "sqlline:sqlline:1.4.0"
-  compile library.java.slf4j_jdk14
   compile library.java.guava
   testCompile project(path: ":beam-sdks-java-io-google-cloud-platform", configuration: "shadowTest")
   testCompile library.java.junit
   testCompile library.java.hamcrest_core
   testCompile library.java.hamcrest_library
+  testRuntimeOnly library.java.slf4j_jdk14
```
lukecwik:
```suggestion
  shadowTestRuntimeClasspath library.java.slf4j_jdk14
```

sdks/java/extensions/sql/build.gradle
```diff
@@ -86,12 +86,12 @@ dependencies {
   provided project(path: ":beam-sdks-java-io-kafka", configuration: "shadow")
   provided project(path: ":beam-sdks-java-io-google-cloud-platform", configuration: "shadow")
   provided library.java.kafka_clients
-  shadowTest library.java.slf4j_jdk14
   shadowTest library.java.junit
   shadowTest library.java.hamcrest_core
   shadowTest library.java.hamcrest_library
   shadowTest library.java.mockito_core
   shadowTest library.java.quickcheck_core
+  testRuntimeOnly library.java.slf4j_jdk14
```
lukecwik:
```suggestion
  shadowTestRuntimeClasspath library.java.slf4j_jdk14
```

sdks/java/extensions/sql/datacatalog/build.gradle
```diff
@@ -39,7 +39,7 @@ dependencies {
   provided project(path: ":beam-sdks-java-io-google-cloud-platform", configuration: "shadow")
   provided library.java.vendored_guava_20_0
   provided library.java.slf4j_api
-  provided library.java.slf4j_simple
+  runtimeOnly library.java.slf4j_simple
```
lukecwik:
```suggestion
  shadowTestRuntimeClasspath library.java.slf4j_simple
```

runners/core-java/build.gradle
iemejia: Thanks for all this information. i was not at all aware of such functionalities. I agree the renaming makes sense but let's do that in a different PR.
Will modules using testShadowJar be tested using the 'shaded' jar? If yes this is great!

sdks/java/extensions/sql/jdbc/build.gradle
iemejia: done

sdks/java/extensions/sql/datacatalog/build.gradle
iemejia: done

sdks/java/extensions/sql/build.gradle
iemejia: done

iemejia:
Did all suggested changes. Could you PTAL again @lukecwik.

runners/core-java/build.gradle
lukecwik: Yes, modules that set testShadowJar will run their unit tests using the shaded jars

sdks/java/extensions/sql/jdbc/build.gradle
```diff
@@ -34,12 +34,13 @@ dependencies {
   compile project(path: ":beam-sdks-java-extensions-sql", configuration: "shadow")
   compile "jline:jline:2.14.6"
   compile "sqlline:sqlline:1.4.0"
-  compile library.java.slf4j_jdk14
   compile library.java.guava
   testCompile project(path: ":beam-sdks-java-io-google-cloud-platform", configuration: "shadowTest")
   testCompile library.java.junit
   testCompile library.java.hamcrest_core
   testCompile library.java.hamcrest_library
+  shadowTestRuntimeClasspath library.java.slf4j_jdk14
```
lukecwik: I don't believe this module is ever meant to be used as a library that is consumed by others and this is why the compile scope is used for slf4j_jdk14

runners/samza/job-server/build.gradle
```diff
@@ -24,18 +24,19 @@ apply plugin: 'application'
...
+    append "reference.conf"
+  },
 )
 
 dependencies {
   compile project(path: ":beam-runners-samza", configuration: "shadow")
   compile group: "org.slf4j", name: "jcl-over-slf4j", version: dependencies.create(project.library.java.slf4j_api).getVersion()
-  compile library.java.slf4j_simple
   shadow library.java.guava
+  shadowTestRuntimeClasspath library.java.slf4j_simple
```
lukecwik: I believe the Samza job server is meant to be a standalone application and compile is the correct scope.

runners/reference/job-server/build.gradle
```diff
@@ -34,7 +35,7 @@ description = "Apache Beam :: Runners :: Reference :: Job Server"
 dependencies {
   compile project(path: ":beam-runners-direct-java", configuration: "shadow")
   compile project(path: ":beam-runners-java-fn-execution", configuration: "shadow")
-  compile library.java.slf4j_simple
+  shadowTestRuntimeClasspath library.java.slf4j_simple
```
lukecwik: I believe the reference job server is also meant to be a standalone application and we want the logging backend to be included so that logging is output.

runners/flink/job-server/flink_job_server.gradle
```diff
@@ -82,10 +83,10 @@ dependencies {
   validatesPortableRunner project(path: ":beam-runners-core-java", configuration: "shadowTest")
   validatesPortableRunner project(path: ":beam-runners-reference-java", configuration: "shadowTest")
   compile project(path: ":beam-sdks-java-extensions-google-cloud-platform-core", configuration: "shadow")
-  compile library.java.slf4j_simple
 //  TODO: Enable AWS and HDFS file system.
   // For resolving external transform requests
   compile project(path: ":beam-sdks-java-io-kafka", configuration: "shadow")
+  shadowTestRuntimeClasspath library.java.slf4j_simple
```
lukecwik: The Flink job server is meant to be a standalone application and not a library which is why we include the slf4j backend as part of it.

runners/samza/job-server/build.gradle
iemejia: done

runners/reference/job-server/build.gradle
iemejia: done

runners/flink/job-server/flink_job_server.gradle
iemejia: done

sdks/java/extensions/sql/jdbc/build.gradle
iemejia: done

iemejia:
Thanks!

runners/core-java/build.gradle
iemejia: This is great improvement, I will try to check in which modules it is worth to do the switch (in particular the ones who do shade stuff to enable it.

sdks/java/extensions/sql/jdbc/build.gradle
mxm: @lukecwik Thanks for catching these.

PR #8516, opened 2019-05-07
Fixing fileio tests for windows

pabloem:
r: @aaltay 
I've tested this for windows, and got fileio tests to pass. See golink at go-slash-fiotpemfx

### DIFF
--- a/sdks/python/apache_beam/io/fileio.py
+++ b/sdks/python/apache_beam/io/fileio.py
@@ -129,9 +129,10 @@ class _ReadMatchesFn(beam.DoFn):
                 if isinstance(file_metadata, (str, unicode))
                 else file_metadata)
 
-    if metadata.path.endswith('/') and self._skip_directories:
+    if ((metadata.path.endswith('/') or metadata.path.endswith('\\'))
+        and self._skip_directories):
       return
-    elif metadata.path.endswith('/'):
+    elif metadata.path.endswith('/') or metadata.path.endswith('\\'):
       raise BeamIOError(
           'Directories are not allowed in ReadMatches transform.'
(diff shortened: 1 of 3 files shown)


sdks/python/apache_beam/io/fileio.py
```diff
@@ -129,9 +129,10 @@ def process(self, file_metadata):
                 if isinstance(file_metadata, (str, unicode))
                 else file_metadata)
 
-    if metadata.path.endswith('/') and self._skip_directories:
+    if ((metadata.path.endswith('/') or metadata.path.endswith('\\'))
```
aaltay: Can you use `os.sep` (https://docs.python.org/3.3/library/os.html?highlight=os.sep#os.sep) ?

sdks/python/apache_beam/io/filesystem.py
```diff
@@ -679,7 +679,7 @@ def match(self, patterns, limits=None):
 
     def _match(pattern, limit):
       """Find all matching paths to the pattern provided."""
-      if pattern.endswith('/'):
+      if pattern.endswith('/') or pattern.endswith('\\'):
```
aaltay: Do we other places that needs this change?

sdks/python/apache_beam/io/fileio.py
pabloem: I'd like to check for both, as we may want to access GCS files in windowss, which use `/`.

sdks/python/apache_beam/io/filesystem.py
pabloem: I didn't find any other checks for `'/'`.

PR #8530, opened 2019-05-08
[BEAM-6988] solved problem related to updates of the str object

NikeNano:
Update to Beam 6988. Removed comment since it was wrong. Will update 

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/typehints/typed_pipeline_test.py
+++ b/sdks/python/apache_beam/typehints/typed_pipeline_test.py
@@ -56,7 +56,7 @@ class MainInputTest(unittest.TestCase):
     result = ['a', 'bb', 'c'] | beam.Map(str.upper)
     self.assertEqual(['A', 'BB', 'C'], sorted(result))
 
-    result = ['xa', 'bbx', 'xcx'] | beam.Map(str.strip, 'x')
+    result = ['xa', 'bbx', 'xcx'] | beam.Map(lambda x: x.strip("x"))
     self.assertEqual(['a', 'bb', 'c'], sorted(result))
 
     result = ['1', '10', '100'] | beam.Map(int)


NikeNano:
R: @aaltay  @fredo838  @Juta  @tvalentyn

NikeNano:
No ;( sorry for having two badly looked through PR atm. I will look at

tvalentyn:
No worries!

NikeNano:
BACKGROUND: I thought I understood the reason for the error(thought it was to difference in str.split in python2 vs python3) and changed in the test. However @tvalentyn pointed out that that was not the case. Thus it is unclear why it fails and changing in the test is not a good solution. 

The test fails in "/beam/sdks/python/apache_beam/typehints/decorators.py", line 298, in getcallargs_forhints" when calling the function inspect.getcallargs(func, *packed_typeargs, **typekwargs). This function fails due to that in python3 to few positional(1 is passed) arguments are passed where two args are need(self,chars). This result in that the test fails with the error strip() missing 1 required positional argument: 'chars'.  In python 2 one positional argument is passed and one arg is required. I would be very happy for some help on this @udim(know you mentioned earlier you would be happy to help) since I can't figure it out. Have really tried my best :)

What i have figured out: 
- There is a difference in the inspect.getcallargs between python2 and python3
- There seams to be a difference in the required arguments when using str.split in python2 vs python3

If you have any tips/questions/advice on how to solve it, let me know :)

fredo838:
I tested [this](https://github.com/apache/beam/blob/5c7ee600ba8560102f61622c894633955d208f87/sdks/python/apache_beam/typehints/decorators.py#L123-L148) function, for python 2.7 - 3.5 - 3.6 - 3.7. 

The result of calling getfullargspec(fstr.split)
2.7 FullArgSpec(args=['_'],  ...)
3.5 FullArgSpec(args=['self'],  ...)
3.6 FullArgSpec(args=['self'],  ...)
3.7 FullArgSpec(args=['self', 'sep', 'maxsplit],  ...)

Seems that is related to the 'monkeypatch' specified in [here](https://github.com/apache/beam/blob/5c7ee600ba8560102f61622c894633955d208f87/sdks/python/apache_beam/typehints/decorators.py#L274), which is already an issue in JIRA ([BEAM-5490](https://issues.apache.org/jira/browse/BEAM-5490), which was partially fixed by [PR-6781](https://github.com/apache/beam/pull/6781), but this assumed all python 3 versions behaved the same. Author of this PR is @tvalentyn , so I guess that makes a full circle. I think we need to rework the typehints file to remove the monkey patch. It shouldn't be necessairy in python3.7, because the  `inspect` module _can_ now take str.split as input, while it fails on python3.6 and python3.5. But then how would we go about on python3.5-3.6? @thoughts?

udim:
getargspec and getfullargspec has not worked on builtin functions before 3.7.
What we're getting in 3.5 and 3.6 is the result of:
```
>>> typehints.decorators.getfullargspec(str.strip.__call__)
FullArgSpec(args=['self'], varargs='args', varkw='kwargs', defaults=None, kwonlyargs=[], kwonlydefaults=None, annotations={})
```
Note that this is a bogus argspec with `self`, `args`, and `kwargs`.

Does anyone know what corner case these lines handle?
https://github.com/apache/beam/blob/5c7ee600ba8560102f61622c894633955d208f87/sdks/python/apache_beam/typehints/decorators.py#L131-L133

udim:
In `apache_beam.typehints.typed_pipeline_test:MainInputTest.test_non_function`, `getcallargs_forhints` is called twice:
https://github.com/apache/beam/blob/ef4b2ef7e5fa2fb87e1491df82d2797947f51be9/sdks/python/apache_beam/transforms/ptransform.py#L761-L762

The second time fails, since type_hints is equals to `((<class 'str'>,), {})`, but inspect.getcallargs fails because it thinks str.strip has 2 required positional args (self and chars). I opened a bug for this (https://bugs.python.org/issue36920), but I believe we can use inspect.Signature instead in Python 3.

The code generating the type hint is I believe:
https://github.com/apache/beam/blob/5c7ee600ba8560102f61622c894633955d208f87/sdks/python/apache_beam/typehints/decorators.py#L330-L331

So this brings up some questions (cc: @robertwb)
1. Are partial type hints okay?
1. Is the type hints heuristic from decorators.py above useful? Should it stay?

To summarize: the type hints generated for str.strip are partial and inspect.getcallargs is buggy and believes that str.strip has 2 required positional args instead of 1.
I think that if we decide to not guess type hints for builtin functions in Python 3 (the code in decorators.py), then this will solve the issue.
Another option is to replace inspect.getcallargs with inspect.signature(fn).bind(...), which seems to be less buggy, but only works for builtins in 3.7.

robertwb:
I'm not exactly sure what you mean by "partial type hints" but we should attempt to provide as much information as possible, but not fail if we can't. We'll probably just have to work around 3.5 and 3.6, referencing the bug. 

As for (2), this should probably stay until inspect works well on builtin methods.

udim:
`fn.__objclass__` is a partial type hint for some builtin methods, since it doesn't provide type information to all possible arguments.
In Py3.7:
```py
>>> inspect.signature(str.isupper).bind('a')  # okay
<BoundArguments (self='a')>
>>> inspect.signature(str.strip).bind('a')  # missing optional arg chars
<BoundArguments (self='a')>
>>> inspect.signature(str.join).bind('a')  # missing required arg iterable
TypeError: missing a required argument: 'iterable'
```
We can avoid the TypeError above by using `bind_partial()`, but the type information for arguments beyond `self` will be missing from all builtins. Hence, these are partial type hints.

udim:
BTW, I'm still trying to come up with a fix for 3.7.

PR #8535, opened 2019-05-08
[BEAM-6693] ApproximateUnique transform for Python SDK

Hannah-Jiang:
This PR is adding a PTransform for estimating the number of distinct elements in a PCollection and the number of distinct values associated with each key in a PCollection KVs.

It offers the same API as its Java counterpart: https://github.com/apache/beam/blob/11a977b8b26eff2274d706541127c19dc93131a2/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/ApproximateUnique.java.

**Transform names:** 
ApproximateUnique.Globally()
ApproximateUnique.PerKey()

------------------------

--- | --- | --- | --- | ---


sdks/python/apache_beam/transforms/core.py
```diff
@@ -518,12 +525,12 @@ def _process_argspec_fn(self):
   def is_process_bounded(self):
     """Checks if an object is a bound method on an instance."""
     if not isinstance(self.process, types.MethodType):
-      return False # Not a method
+      return False  # Not a method
```
Hannah-Jiang: Format of some lines are changed after I reformat the code.
Code changes start from line 2289.

sdks/python/apache_beam/transforms/core.py
```diff
@@ -2270,3 +2286,163 @@ def to_runner_api_parameter(self, unused_context):
   @PTransform.register_urn(common_urns.primitives.IMPULSE.urn, None)
   def from_runner_api_parameter(unused_parameter, unused_context):
     return Impulse()
+
+
+class ApproximateUniqueGlobally(PTransform):
```
Hannah-Jiang: Since the file is growing, is it better to separate each transform to it's own file? I am happy to make that change.

sdks/python/setup.py
```diff
@@ -125,6 +125,7 @@ def get_version():
     'pyvcf>=0.6.8,<0.7.0; python_version < "3.0"',
     'pyyaml>=3.12,<4.0.0',
     'typing>=3.6.0,<3.7.0; python_version < "3.5.0"',
+    'mmh3>=2.5.1; python_version >= "2.7"',
```
aaltay: A few questions:
- What is this dependency? The pypi page says this is a wrapper. Does it require other things to be installed? Is this the only option we have?
- You do need the python_version >= "2.7" part, because all our sdks only support py >= 2.7
- Can you add an upper bound here. Maybe <3.0.0

sdks/python/apache_beam/transforms/core.py
aaltay: How did you reformat the code? As long as linter still passes, this is fine. However it generally easier to review when format changes are separated from other changes.

sdks/python/apache_beam/transforms/core.py
aaltay: That make sense. Also probably core.py is not the right place anyway.

Hannah-Jiang: Does stats.py sound good?

Hannah-Jiang:
Not ready to review.

sdks/python/apache_beam/transforms/core.py
Hannah-Jiang: I used reformat tool provided by IntelliJ. However, I don't think this tool is the most optimal one and I changed the format back.

sdks/python/setup.py
Hannah-Jiang: Ahmet, you are correct, we need this package only for python 2.7. With python 3, buildin hash function is different than the one from 2.7 and it works well. For 2.7, we still need to use mmh3. mmh3 is a wrapper of MurmurHash3, however, MurmurHash3 only supports from python 3.2. I don't think mmh3 requires other things to be installed. Please let me know if we want to investigate more.

pabloem:
Run Python PreCommit

Hannah-Jiang:
Run Python PreCommit

Hannah-Jiang:
R: @aaltay 
Code is ready to be reviewed.

sdks/python/apache_beam/transforms/stats.py
```diff
@@ -0,0 +1,206 @@
...
+    self._sample_size = sample_size
+
+  def create_accumulator(self, *args, **kwargs):
+    return _LargestUnique(self._sample_size)
+
+  @staticmethod
+  def add_input(accumulator, element, *args, **kwargs):
+    try:
+      if 'mmh3' in sys.modules:
+        accumulator.add(mmh3.hash64(str(element))[1])
```
pabloem: You can do something like:
```
try:
  Import mmh3
  hash_fn = mmh3.hash64
Except ImportError:
  hash_fn = hash
```

To avoid doing this check here.

sdks/python/apache_beam/transforms/stats.py
```diff
@@ -0,0 +1,206 @@
...
+  _NO_VALUE_ERR_MSG = 'Either size or error should be set. Received {}.'
+  _MULTI_VALUE_ERR_MSG = 'Either size or error should be set. ' \
+                         'Received {size = %s, error = %s}.'
+  _INPUT_SIZE_ERR_MSG = 'ApproximateUnique needs a size >= 16 for an error ' \
+                        '<= 0.50. In general, the estimation error is about ' \
+                        '2 / sqrt(sample_size). Received {size = %s}.'
+  _INPUT_ERROR_ERR_MSG = 'ApproximateUnique needs an estimation error ' \
+                         'between 0.01 and 0.50. Received {error = %s}.'
+
+  def __init__(self, **kwargs):
```
aaltay: Curious, why do we prefer this definition over `def __init__(self, size=None, error=None):`? Would not that be more explicit?

sdks/python/apache_beam/transforms/stats.py
```diff
@@ -0,0 +1,206 @@
...
+    log(1 - sampleSize/sampleSpace) / log(1 - 1/sampleSpace) ~ sampleSize
+    and hence estimate ~ sampleSize * HASH_SPACE_SIZE / sampleSpace
+    as one would expect.
+    """
+
+    if len(self._sample_heap) < self._sample_size:
+      return len(self._sample_heap)
+    else:
+      sample_space_size = sys.maxsize - 1.0 * self._min_hash
+      est = math.log1p(-self._sample_size/ sample_space_size) \
```
aaltay: What is the source of this math? Is there a link that one can follow and understand what is happening here?

sdks/python/apache_beam/transforms/stats.py
```diff
@@ -0,0 +1,206 @@
...
+  def create_accumulator(self, *args, **kwargs):
+    return _LargestUnique(self._sample_size)
+
+  @staticmethod
+  def add_input(accumulator, element, *args, **kwargs):
+    try:
+      if 'mmh3' in sys.modules:
+        accumulator.add(mmh3.hash64(str(element))[1])
+      else:
+        accumulator.add(hash(str(element)))
```
aaltay: Is builtin hash good enough for statistical purposes?

sdks/python/apache_beam/transforms/stats.py
```diff
@@ -0,0 +1,206 @@
...
+  def __init__(self, sample_size):
+    self._sample_size = sample_size
+
+  def create_accumulator(self, *args, **kwargs):
+    return _LargestUnique(self._sample_size)
+
+  @staticmethod
+  def add_input(accumulator, element, *args, **kwargs):
+    try:
+      if 'mmh3' in sys.modules:
```
aaltay: Why don't we use `hash()` in python2?

sdks/python/apache_beam/transforms/stats.py
```diff
@@ -0,0 +1,206 @@
...
+        accumulator.add(hash(str(element)))
+      return accumulator
+    except Exception as e:
+      raise RuntimeError("Runtime exception: %s", e)
+
+  def merge_accumulators(self, accumulators, *args, **kwargs):
+    merged_accumulator = self.create_accumulator()
+    for accumulator in accumulators:
+      for i in accumulator._sample_heap:
+        merged_accumulator.add(i)
```
aaltay: Is it possible to offer a fast merge for 2 accumulators? Like a merge two heaps in a single call? Perhaps file a JIRA for future optimization of this.

sdks/python/apache_beam/transforms/stats.py
Hannah-Jiang: I think mmh3 performs better than builtin hash() in python3. Now I am doing some testing, with builtin hash(), ~3% of estimation error is greater than max error when population is big. I will test more to see if we need to use mmh3 for python 3.

sdks/python/apache_beam/transforms/stats.py
```diff
@@ -0,0 +1,206 @@
...
+           >> (CombineGlobally(ApproximateUniqueCombineDoFn(self._sample_size)))
+
+  @staticmethod
+  def _get_sample_size_from_est_error(est_err):
+    """
+    :return: sample size
+
+    Calculate sample size from estimation error
+    """
+    return int(math.ceil(4.0 / math.pow(est_err, 2.0)))
```
Hannah-Jiang: I forgot to remove int() here, will fix it at next commit.

sdks/python/apache_beam/transforms/stats.py
```diff
@@ -0,0 +1,205 @@
...
+    Given above sample_size approximate,
+    est = log1p(-sample_size/sample_space) / log1p(-1/sample_space)
+      * hash_space / sample_space
+    """
+
+    if len(self._sample_heap) < self._sample_size:
+      return len(self._sample_heap)
+    else:
+      sample_space_size = sys.maxsize - 1.0 * self._min_hash
+      est = math.log1p(-self._sample_size / sample_space_size) \
```
pabloem: In Beam we usually do multi-line statements with parenthesis. Would you be okay changing this to parenthesis-based instead of `\`-escaped?

sdks/python/apache_beam/transforms/stats.py
```diff
@@ -0,0 +1,205 @@
...
+  def __init__(self, sample_size):
+    self._sample_size = sample_size
+
+  def create_accumulator(self, *args, **kwargs):
+    return _LargestUnique(self._sample_size)
+
+  @staticmethod
+  def add_input(accumulator, element, *args, **kwargs):
+    try:
+      accumulator.add(mmh3.hash64(str(element))[1])
```
pabloem: I don't think we want to do `str` here. We probably want to encode the element into its byte-representation. @robinyqiu what does Java do in this case?

sdks/python/apache_beam/transforms/stats.py
```diff
@@ -0,0 +1,206 @@
...
+            / sample_space_size
+
+      return round(est)
+
+
+class ApproximateUniqueCombineDoFn(CombineFn):
+  """
+  ApproximateUniqueCombineDoFn computes an estimate of the number of
+  unique values that were combined.
+  """
```
pabloem: Would you add display data for this combine? See for example here: https://github.com/apache/beam/blob/master/sdks/python/apache_beam/io/gcp/bigquery_file_loads.py#L174-L179

sdks/python/apache_beam/transforms/stats.py
pabloem: oops this is duplicated.

aaltay:
@Hannah-Jiang could you also please update the PR description?

sdks/python/setup.py
robertwb: Could you clarify why the builtin hash function does not work well in 2.7?

sdks/python/apache_beam/transforms/stats.py
```diff
@@ -0,0 +1,206 @@
...
+      sample_space_size = sys.maxsize - 1.0 * self._min_hash
+      est = math.log1p(-self._sample_size/ sample_space_size) \
+            / math.log1p(-1 / sample_space_size) \
+            * self._HASH_SPACE_SIZE \
+            / sample_space_size
+
+      return round(est)
+
+
+class ApproximateUniqueCombineDoFn(CombineFn):
```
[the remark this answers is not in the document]
robertwb: Yes, users should have access to the CombineFn (e.g. for state). (Agree it should not be called a DoFn).

sdks/python/apache_beam/transforms/stats.py
robertwb: This assumes that distinct values never have the same string representation, which may not be true. At the very least this should be documented, better would be to use a different encoding (do we have access to a Coder here?)

sdks/python/apache_beam/transforms/stats.py
robinyqiu: Oh I just saw this question. I think we should use coder (should be accessible) instead of calling `str`, as @robertwb mentioned in his comment below.

sdks/python/apache_beam/transforms/stats.py
Hannah-Jiang: I was not able to find a document, instead I added more explanation about the estimate equation.

sdks/python/apache_beam/transforms/stats.py
Hannah-Jiang: It ends up that we should use mmh3 for python 3 as well. mmh3 has better estimation performance than builtin hash.

sdks/python/setup.py
Hannah-Jiang: Thanks for asking this question. I was able to dig into more and found an unexpected reason.
From 3.4, Python used SpiHash algorithm, before that, it used FNV algorithm. Though SpiHash reduced collision, I don't think it's noticeable from my test. So I saw how hash values are distributed. Graphs at https://docs.google.com/spreadsheets/d/1QpNAorUhUY1Nq3b4QRMfbB9Rupq3ZC9FshunIyRRfH0/edit#gid=0 shows how hash values are distributed with different hash algorithm given exact the same data set. Builtin hash function with Python 2 is not as uniformly distributed as other hash functions (builtin hash function with Python 3 and mmh3). Though it's not obvious from the graphs that hash values from mmh3 algorithms are more uniformed distributed than the ones from builtin hash function with python 3, test results shows that mmh3 has more uniformly distributed hash values.

Why does it matter to approximate unique count? We are calculating sample space with hash values and calculate population space from the sample space. More uniformly distributed hash values give us more accurate sample space thus we can calculate more accurate population space. The difference is more obvious when we have more duplicate records with input.

A test case is generating 10000 elements within range of [0, 1000]. This test consistently fails if we use builtin hash function with py2.7, and fails sometimes with builtin hash function with py3 and hasn't fail with mmh3.

Hannah-Jiang: mmh3
py2
py3

sdks/python/apache_beam/transforms/stats_test.py
```diff
@@ -0,0 +1,384 @@
...
+    for (x, y) in test_input:
+      actual_count_dict[x].add(y)
+
+    pipeline = TestPipeline()
+    result = (pipeline
+              | 'create' >> beam.Create(test_input)
+              | 'get_estimate'
+              >> beam.ApproximateUniquePerKey(size=sample_size)
+              | 'compare'
+              >> beam.ParDo(lambda x: [abs(x[1] - len(actual_count_dict[x[0]]))
```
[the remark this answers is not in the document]
Hannah-Jiang: It is fixed now.

sdks/python/apache_beam/transforms/stats.py
Hannah-Jiang: It is added.

sdks/python/apache_beam/transforms/stats.py
Hannah-Jiang: It is fixed.

sdks/python/apache_beam/transforms/stats.py
Hannah-Jiang: I changed to use coder here. Thanks for the comment.

Hannah-Jiang:
Forgot to mention, the code is ready to be reviewed with new fixes.

sdks/python/apache_beam/transforms/__init__.py
```diff
@@ -26,6 +26,7 @@
 from apache_beam.transforms.ptransform import *
 from apache_beam.transforms.timeutil import TimeDomain
 from apache_beam.transforms.util import *
+from apache_beam.transforms.stats import *
```
aaltay: nit; put this in sorted order.

sdks/python/setup.py
```diff
@@ -125,6 +125,7 @@ def get_version():
     'pyvcf>=0.6.8,<0.7.0; python_version < "3.0"',
     'pyyaml>=3.12,<4.0.0',
     'typing>=3.6.0,<3.7.0; python_version < "3.5.0"',
+    'mmh3>=2.5.1',
```
aaltay: Put an upper bound?

sdks/python/apache_beam/transforms/stats.py
```diff
@@ -0,0 +1,216 @@
...
+        self._max_est_err = error
+
+  def expand(self, pcoll):
+    coder = coders.registry.get_coder(pcoll)
+    return pcoll \
+           | 'CountGlobalUniqueValues' \
+           >> (CombineGlobally(ApproximateUniqueCombineFn(self._sample_size,
+                                                          coder)))
+
+  def display_data(self):
```
pabloem: Sorry about the confusion. Composite transforms cannot contain display data. Can you add it to the combinefn?

Hannah-Jiang: Thanks for catching it. It was fixed at most recent commit.

Hannah-Jiang:

pabloem:
@robinyqiu any other comments?

pabloem:

sdks/python/apache_beam/transforms/stats.py
```diff
@@ -0,0 +1,215 @@
...
+    """
+    :return: sample size
+
+    Calculate sample size from estimation error
+    """
+    # math.ceil in python 2.7 returns float, while it returns int in python 3.
+    return int(math.ceil(4.0 / math.pow(est_err, 2.0)))
+
+
+class ApproximateUniquePerKey(ApproximateUniqueGlobally):
```
robinyqiu: Sharing code by letting `ApproximateUniquePerKey` to extend `ApproximateUniqueGlobally` seems a bit weird to me. How about we create a wrapper class `ApproximateUnique` and make `PerKey` and `Globally` two inner classes, as what we usually do (see https://github.com/apache/beam/blob/a71f305402efe050c9dcf5ef305141a66efb2953/sdks/python/apache_beam/transforms/combiners.py#L112)?

We can abstract the input checking logic into a static method and call them from both constructors.

sdks/python/apache_beam/transforms/stats.py
```diff
@@ -0,0 +1,215 @@
...
+from apache_beam.transforms.core import *
+from apache_beam.transforms.ptransform import PTransform
+
+__all__ = [
+    'ApproximateUniqueGlobally',
+    'ApproximateUniquePerKey',
+]
+
+
+class ApproximateUniqueGlobally(PTransform):
```
robinyqiu: We can add type annotation to the `PTransform`s here using `@with_input_type` and `@with_output_type`.

See https://github.com/apache/beam/blob/a71f305402efe050c9dcf5ef305141a66efb2953/sdks/python/apache_beam/transforms/core.py#L1788

robinyqiu:
Added some comments.

sdks/python/apache_beam/transforms/stats.py
```diff
@@ -0,0 +1,236 @@
...
+    """
+    :return: sample size
+
+    Calculate sample size from estimation error
+    """
+    #math.ceil in python2.7 returns a float, while it returns an int in python3.
+    return int(math.ceil(4.0 / math.pow(est_err, 2.0)))
+
+  @typehints.with_input_types(T)
+  @typehints.with_output_types(T)
```
robinyqiu: I think the output type here should be a numeric type, not the same input type `T`. Same below.

sdks/python/apache_beam/transforms/stats.py
```diff
@@ -0,0 +1,236 @@
...
+  _MULTI_VALUE_ERR_MSG = 'Either size or error should be set. ' \
+                         'Received {size = %s, error = %s}.'
+  _INPUT_SIZE_ERR_MSG = 'ApproximateUnique needs a size >= 16 for an error ' \
+                        '<= 0.50. In general, the estimation error is about ' \
+                        '2 / sqrt(sample_size). Received {size = %s}.'
+  _INPUT_ERROR_ERR_MSG = 'ApproximateUnique needs an estimation error ' \
+                         'between 0.01 and 0.50. Received {error = %s}.'
+
+  @staticmethod
+  def get_input_params(size=None, error=None):
```
robinyqiu: Nit: rename to `parse_input_params`?

robinyqiu:
Thanks for the changes. I left a few comments.

A quick reminder: after you have addressed all reviewers' comments, you can squash all the commits into one (see https://stackoverflow.com/a/5189600/9963322).

pabloem:
I can squash on merge.

robinyqiu:
LGTM. Thanks for the changes!

jonathanasdf:
Should https://beam.apache.org/documentation/transforms/python/overview/ be updated?

aaltay:
> Should [beam.apache.org/documentation/transforms/python/overview](https://beam.apache.org/documentation/transforms/python/overview/) be updated?

ApproximateUnique is already on the list. Am I missing something?

jonathanasdf:
It's on the list but says "Not available" in the description

PR #8558, opened 2019-05-11
[BEAM-7131] Spark: cache executable stage output to prevent re-computation

ibzib:
When an executable stage has more than one consumer, it currently results in the executable stage being re-computed for each extra consumer. This behavior is undesirable because we are potentially fusing a large amount of work into some executable stages, in addition to thorny problems like side effects (see the [Jira](https://issues.apache.org/jira/browse/BEAM-7131)) which, though not generally encouraged, should not be breaking things so consistently and opaquely.

R: @robertwb @angoenka 

--- | --- | --- | --- | ---

### DIFF
--- a/runners/spark/src/main/java/org/apache/beam/runners/spark/translation/SparkBatchPortablePipelineTranslator.java
+++ b/runners/spark/src/main/java/org/apache/beam/runners/spark/translation/SparkBatchPortablePipelineTranslator.java
@@ -112,11 +112,22 @@ public class SparkBatchPortablePipelineTranslator {
     QueryablePipeline p =
         QueryablePipeline.forTransforms(
             pipeline.getRootTransformIdsList(), pipeline.getComponents());
-    // Pre-scan pipeline to count which pCollections are consumed as inputs more than once so their
-    // corresponding RDDs can later be cached.
     for (PipelineNode.PTransformNode transformNode : p.getTopologicallyOrderedTransforms()) {
+      // Pre-scan pipeline to count which pCollections are consumed as inputs more than once so
+      // their corresponding RDDs can later be cached.
       for (String inputId : transformNode.getTransform().getInputsMap().values()) {
-        context.incrementConsumptionCount(inputId);
+        context.incrementConsumptionCountBy(inputId, 1);
+      }
+      // Executable stage consists of two parts: computation and extraction. This means the result
+      // of computation is an intermediate RDD, which we might also need to cache.
+      if (transformNode.getTransform().getSpec().getUrn().equals(ExecutableStage.URN)) {
+        context.incrementConsumptionCountBy(
+            getExecutableStageIntermediateId(transformNode),
+            transformNode.getTransform().getOutputsMap().size());
+      }
+      for (String outputId : transformNode.getTransform().getOutputsMap().values()) {
+        WindowedValueCoder outputCoder = getWindowedValueCoder(outputId, pipeline.getComponents());
+        context.putCoder(outputId, outputCoder);
       }
     }
     for (PipelineNode.PTransformNode transformNode : p.getTopologicallyOrderedTransforms()) {
@@ -138,11 +149,10 @@ public class SparkBatchPortablePipelineTranslator {
 
   private static void translateImpulse(
       PTransformNode transformNode, RunnerApi.Pipeline pipeline, SparkTranslationContext context) {
-    Coder<byte[]> coder = ByteArrayCoder.of();
     BoundedDataset<byte[]> output =
         new BoundedDataset<>(
-            Collections.singletonList(new byte[0]), context.getSparkContext(), coder);
-    context.pushDataset(getOutputId(transformNode), output, coder);
+            Collections.singletonList(new byte[0]), context.getSparkContext(), ByteArrayCoder.of());
+    context.pushDataset(getOutputId(transformNode), output);
   }
 
   private static <K, V> void translateGroupByKey(
@@ -182,15 +192,12 @@ public class SparkBatchPortablePipelineTranslator {
                   context.serializablePipelineOptions,
                   AggregatorsAccumulator.getInstance()));
     }
-    String outputId = getOutputId(transformNode);
-    WindowedValueCoder<KV<K, V>> outputCoder = getWindowedValueCoder(outputId, components);
-    context.pushDataset(outputId, new BoundedDataset<>(groupedByKeyAndWindow), outputCoder);
+    context.pushDataset(getOutputId(transformNode), new BoundedDataset<>(groupedByKeyAndWindow));
   }
 
   private static <InputT, OutputT, SideInputT> void translateExecutableStage(
       PTransformNode transformNode, RunnerApi.Pipeline pipeline, SparkTranslationContext context) {
 
-    RunnerApi.Components pipelineComponents = pipeline.getComponents();
     RunnerApi.ExecutableStagePayload stagePayload;
     try {
       stagePayload =
@@ -203,7 +210,7 @@ public class SparkBatchPortablePipelineTranslator {
     Dataset inputDataset = context.popDataset(inputPCollectionId);
     JavaRDD<WindowedValue<InputT>> inputRdd = ((BoundedDataset<InputT>) inputDataset).getRDD();
     Map<String, String> outputs = transformNode.getTransform().getOutputsMap();
-    BiMap<String, Integer> outputMap = createOutputMap(outputs.values());
+    BiMap<String, Integer> outputExtractionMap = createOutputMap(outputs.values());
 
     ImmutableMap.Builder<String, Tuple2<Broadcast<List<byte[]>>, WindowedValueCoder<SideInputT>>>
         broadcastVariablesBuilder = ImmutableMap.builder();
@@ -222,24 +229,39 @@ public class SparkBatchPortablePipelineTranslator {
         new SparkExecutableStageFunction<>(
             stagePayload,
             context.jobInfo,
-            outputMap,
+            outputExtractionMap,
             broadcastVariablesBuilder.build(),
             MetricsAccumulator.getInstance());
     JavaRDD<RawUnionValue> staged = inputRdd.mapPartitions(function);
+    String intermediateId = getExecutableStageIntermediateId(transformNode);
+    context.pushDataset(
+        intermediateId,
+        new Dataset() {
+          @Override
+          public void cache(String storageLevel, Coder<?> coder) {
+            StorageLevel level = StorageLevel.fromString(storageLevel);
+            staged.persist(level);
+          }
 
-    // Prevent potentially expensive re-computation of executable stage
-    SparkPipelineOptions sparkOptions =
-        context.serializablePipelineOptions.get().as(SparkPipelineOptions.class);
-    if (!sparkOptions.isCacheDisabled() && outputs.size() > 1) {
-      StorageLevel storageLevel = StorageLevel.fromString(sparkOptions.getStorageLevel());
-      staged.persist(storageLevel);
-    }
+          @Override
+          public void action() {
+            // Empty function to force computation of RDD.
+            staged.foreach(TranslationUtils.emptyVoidFunction());
+          }
+
+          @Override
+          public void setName(String name) {
+            staged.setName(name);
+          }
+        });
+    // pop dataset to mark RDD as used
+    context.popDataset(intermediateId);
 
     for (String outputId : outputs.values()) {
       JavaRDD<WindowedValue<OutputT>> outputRdd =
-          staged.flatMap(new SparkExecutableStageExtractionFunction<>(outputMap.get(outputId)));
-      WindowedValueCoder<OutputT> outputCoder = getWindowedValueCoder(outputId, pipelineComponents);
-      context.pushDataset(outputId, new BoundedDataset<>(outputRdd), outputCoder);
+          staged.flatMap(
+              new SparkExecutableStageExtractionFunction<>(outputExtractionMap.get(outputId)));
+      context.pushDataset(outputId, new BoundedDataset<>(outputRdd));
     }
     if (outputs.isEmpty()) {
       // After pipeline translation, we traverse the set of unconsumed PCollections and add a
@@ -250,8 +272,7 @@ public class SparkBatchPortablePipelineTranslator {
           staged.flatMap((rawUnionValue) -> Collections.emptyIterator());
       context.pushDataset(
           String.format("EmptyOutputSink_%d", context.nextSinkId()),
-          new BoundedDataset<>(outputRdd),
-          null);
+          new BoundedDataset<>(outputRdd));
     }
   }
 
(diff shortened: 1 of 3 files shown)


runners/spark/src/test/java/org/apache/beam/runners/spark/SparkPortableExecutionTest.java
```diff
@@ -163,4 +168,75 @@ public void process(ProcessContext context) {
...
+    JobInvocation jobInvocation =
+        SparkJobInvoker.createJobInvocation(
+            "testParDoWithSideEffects",
+            "testParDoWithSideEffectsRetrievalToken",
+            sparkJobExecutor,
+            pipelineProto,
+            options.as(SparkPipelineOptions.class));
+    jobInvocation.start();
+    while (jobInvocation.getState() != Enum.DONE) {
+      Thread.sleep(1000);
```
angoenka: Shall we use SameThreadExecutor[1] to avoid sleeping.

[1] https://google.github.io/guava/releases/17.0/api/docs/com/google/common/util/concurrent/MoreExecutors.html#sameThreadExecutor()

runners/spark/src/test/java/org/apache/beam/runners/spark/SparkPortableExecutionTest.java
```diff
@@ -163,4 +168,75 @@ public void process(ProcessContext context) {
...
+   */
+  @Test(timeout = 120_000)
+  public void testParDoWithSideEffects() throws Exception {
+    PipelineOptions options = PipelineOptionsFactory.create();
+    options.setRunner(CrashingRunner.class);
+    options
+        .as(PortablePipelineOptions.class)
+        .setDefaultEnvironmentType(Environments.ENVIRONMENT_EMBEDDED);
+    Pipeline pipeline = Pipeline.create(options);
+    Path path = FileSystems.getDefault().getPath(TEST_DIR_PREFIX, UUID.randomUUID().toString());
```
angoenka: We can use TemporaryFolder to avoid cleanup later

Example https://github.com/apache/beam/blob/b953645ed6db837d24284d7fe1fe091e7309f821/sdks/java/core/src/test/java/org/apache/beam/sdk/io/LocalFileSystemTest.java#L61

runners/spark/src/test/java/org/apache/beam/runners/spark/SparkPortableExecutionTest.java
```diff
@@ -163,4 +168,75 @@ public void process(ProcessContext context) {
...
+            "testParDoWithSideEffects",
+            "testParDoWithSideEffectsRetrievalToken",
+            sparkJobExecutor,
+            pipelineProto,
+            options.as(SparkPipelineOptions.class));
+    jobInvocation.start();
+    while (jobInvocation.getState() != Enum.DONE) {
+      Thread.sleep(1000);
+    }
+    Assert.assertTrue(dir.exists());
```
angoenka: We should clean the directory. Please refer to https://github.com/apache/beam/blob/b953645ed6db837d24284d7fe1fe091e7309f821/sdks/java/core/src/test/java/org/apache/beam/sdk/io/LocalFileSystemTest.java#L61 for automatic cleanup.

runners/spark/src/main/java/org/apache/beam/runners/spark/translation/SparkBatchPortablePipelineTranslator.java
```diff
@@ -224,6 +224,11 @@ private static void translateImpulse(
             MetricsAccumulator.getInstance());
     JavaRDD<RawUnionValue> staged = inputRdd.mapPartitions(function);
 
+    // Prevent potentially expensive re-computation of executable stage
+    if (outputs.size() > 1) {
+      staged.cache();
```
angoenka: For correctness purpose, we should also cache RDDs which have multiple consumers.

Example, ES1 has 1 output PC1 which is consumed by GBKa and GBKb. GBKa is consumed by ESa and GBKb is consumed by ESb. In this case ES1 will be executed twice.

ibzib: Yes, good call. Do you think it would be better to cache all intermediate RDDs (including those for other transforms such as GBK, flatten, etc.), like the legacy Spark runner does, or cache just those outputted by the exec stage function?

angoenka:
Run Samza ValidatesRunner

angoenka:
Ignore Run Samza ValidatesRunner

angoenka:
Run Samza ValidatesRunner

angoenka:
Run Seed Job

angoenka:
Run Seed Job

angoenka:
sorry for hijacking your pull request :)

runners/spark/src/main/java/org/apache/beam/runners/spark/translation/SparkBatchPortablePipelineTranslator.java
ibzib: I went for caching all the intermediate RDDs (can be turned off by `setCacheDisabled(true)`). PTAL

robertwb: I was about to comment the same. 

I think caching all intermediate RDDs will be prohibitively expensive, and disabling them all has re-execution issues. Did you try the flatten + materialize approach we discussed earlier?

ibzib: Flattening the output RDDs (or "leaves") seems to have no effect on re-computation. As far as I understand, Spark caches nothing unless explicitly commanded by the user.

robertwb: That's really unfortunate. 

So the current design is to cache, when the (by default True flag is enabled) (1) collections that are used more than once and (2) all outputs of transforms with multiple outputs.

That seems a reasonable first step. I think we can do better  (e.g. cache all but one output from multi-output transforms, uncache once we're done with all consumers), but could probably call that future work. 

This PR also seems to change a bunch of stuff with how coding works. Could you factor that out into a separate PR?

ibzib: (1) and (2) are correct.

I added the coder as an argument to `SparkTranslationContext::pushDataset` because the coder is used in the Spark runner's bespoke caching implementation: https://github.com/apache/beam/blob/fdedb4c41e32606c64c8c1385d12444998dc4a2b/runners/spark/src/main/java/org/apache/beam/runners/spark/translation/BoundedDataset.java#L110-L117

runners/spark/src/main/java/org/apache/beam/runners/spark/translation/SparkBatchPortablePipelineTranslator.java
```diff
@@ -224,10 +227,19 @@ private static void translateImpulse(
             MetricsAccumulator.getInstance());
     JavaRDD<RawUnionValue> staged = inputRdd.mapPartitions(function);
 
+    // Prevent potentially expensive re-computation of executable stage
+    SparkPipelineOptions sparkOptions =
+        context.serializablePipelineOptions.get().as(SparkPipelineOptions.class);
+    if (!sparkOptions.isCacheDisabled() && outputs.size() > 1) {
+      StorageLevel storageLevel = StorageLevel.fromString(sparkOptions.getStorageLevel());
+      staged.persist(storageLevel);
```
robertwb: It seems a bit odd that we do caching here and also implicitly on pushDataset as determined by the incrementConsumptionCount calls. 

Could we instead walk over the graph once, above, determining the total set of collections that need caching rather than spreading out the logic. (The caching itself would likely still have to happen in pushDataset.) I think this would be cleaner to follow and easier to update in the future.

robertwb: If we do this, we could also possibly compute/store the coder required for caching there too, rather than plumbing it all over.

runners/spark/src/main/java/org/apache/beam/runners/spark/translation/SparkBatchPortablePipelineTranslator.java
```diff
@@ -111,6 +112,25 @@ public void translate(final RunnerApi.Pipeline pipeline, SparkTranslationContext
...
+      }
+      // Executable stage consists of two parts: computation and extraction. This means the result
+      // of computation is an intermediate RDD, which we might also need to cache.
+      if (transformNode.getTransform().getSpec().getUrn().equals(ExecutableStage.URN)) {
+        context.incrementConsumptionCountBy(
+            getExecutableStageIntermediateId(transformNode),
+            transformNode.getTransform().getOutputsMap().size());
+      }
+      for (String outputId : transformNode.getTransform().getOutputsMap().values()) {
+        RunnerApi.Components components = pipeline.getComponents();
```
robertwb: Move this out of the loop.

runners/spark/src/main/java/org/apache/beam/runners/spark/translation/SparkBatchPortablePipelineTranslator.java
```diff
@@ -111,6 +112,25 @@ public void translate(final RunnerApi.Pipeline pipeline, SparkTranslationContext
...
+        WindowedValueCoder outputCoder = getWindowedValueCoder(outputId, components);
```
robertwb: Nit: I would just inline this line. (Or, alternatively, is there any reason context does not have a handle to components itself?)

runners/spark/src/main/java/org/apache/beam/runners/spark/translation/SparkBatchPortablePipelineTranslator.java
```diff
@@ -111,6 +112,25 @@ public void translate(final RunnerApi.Pipeline pipeline, SparkTranslationContext
...
         QueryablePipeline.forTransforms(
             pipeline.getRootTransformIdsList(), pipeline.getComponents());
+    for (PipelineNode.PTransformNode transformNode : p.getTopologicallyOrderedTransforms()) {
+      // Pre-scan pipeline to count which pCollections are consumed as inputs more than once so
+      // their corresponding RDDs can later be cached.
+      for (String inputId : transformNode.getTransform().getInputsMap().values()) {
+        context.incrementConsumptionCountBy(inputId, 1);
+      }
+      // Executable stage consists of two parts: computation and extraction. This means the result
+      // of computation is an intermediate RDD, which we might also need to cache.
```
robertwb: Is this preferable to extracting the RDD corresponding to the (potentially expensive) extraction? (In that case, rather than special casing ExecutableStage, we could simply increment all outputs of multi-output transforms.)

runners/spark/src/main/java/org/apache/beam/runners/spark/translation/SparkBatchPortablePipelineTranslator.java
```diff
@@ -130,9 +150,10 @@ private static void urnNotFound(
 
   private static void translateImpulse(
       PTransformNode transformNode, RunnerApi.Pipeline pipeline, SparkTranslationContext context) {
+    Coder<byte[]> coder = ByteArrayCoder.of();
```
robertwb: Revert.

runners/spark/src/main/java/org/apache/beam/runners/spark/translation/SparkBatchPortablePipelineTranslator.java
```diff
@@ -111,6 +112,25 @@ public void translate(final RunnerApi.Pipeline pipeline, SparkTranslationContext
     QueryablePipeline p =
         QueryablePipeline.forTransforms(
             pipeline.getRootTransformIdsList(), pipeline.getComponents());
+    for (PipelineNode.PTransformNode transformNode : p.getTopologicallyOrderedTransforms()) {
+      // Pre-scan pipeline to count which pCollections are consumed as inputs more than once so
+      // their corresponding RDDs can later be cached.
+      for (String inputId : transformNode.getTransform().getInputsMap().values()) {
+        context.incrementConsumptionCountBy(inputId, 1);
```
robertwb: I was thinking that rather than putting consumption counts on the context, we would just compute them locally and then inform the context of the full set of collections that need caching (together with their coders). But we can keep it this way if you'd prefer.

ibzib: Ack, I think I'm just going to leave it the way it is -- it somehow feels better to pass integers than maps. :man_shrugging:

runners/spark/src/main/java/org/apache/beam/runners/spark/translation/SparkBatchPortablePipelineTranslator.java
ibzib: Done

runners/spark/src/main/java/org/apache/beam/runners/spark/translation/SparkBatchPortablePipelineTranslator.java
ibzib: Done

runners/spark/src/main/java/org/apache/beam/runners/spark/translation/SparkBatchPortablePipelineTranslator.java
ibzib: We have to cache the output of both the executable stage itself and the extraction function in order to fully prevent re-computation (tested in `testExecStageWithMultipleOutputs` and `testExecStageWithMultipleConsumers`, respectively). The immediate output of the stage (ie the input to extraction) doesn't correspond to any RunnerApi-level pCollection, being a runner-specific detail, which is why it's special-cased here.

runners/spark/src/main/java/org/apache/beam/runners/spark/translation/SparkBatchPortablePipelineTranslator.java
ibzib: Done. (Mostly my reasoning here is that the translator is responsible for all of the messy pipeline reading, leaving `SparkTranslationContext` responsible only for execution-level details).

runners/spark/src/main/java/org/apache/beam/runners/spark/translation/SparkTranslationContext.java
```diff
@@ -17,6 +17,7 @@
  */
 package org.apache.beam.runners.spark.translation;
 
+import com.sun.istack.Nullable;
```
kennknowles, line 20: This is probably not the `Nullable` you want. CC @tysonjh since I found this from a failure on his PR.

tysonjh: Correct, use javax nullable instead please.

I have local changes to replace the other occurrences of the istack one.

PR #8565, opened 2019-05-13
[BEAM-7239] Do not close DataSource on Teardown on JdbcIO

iemejia:
DataSource implementations do not necessarily implement close() and even
if so, it may make sense not to close them systematically on @Teardown
because they could be provided for a longer lifecycle than the DoFn
thread, e.g. PoolableDataSource should not be instantiated per DoFn
in particular for Streaming Jobs but by VM.
(See [BEAM-7230](https://issues.apache.org/jira/browse/BEAM-7230) for details.

R: @aromanenko-dev 
CC: @jbonofre

### DIFF
--- a/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcIO.java
+++ b/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcIO.java
@@ -769,9 +769,6 @@ public class JdbcIO {
     @Teardown
     public void teardown() throws Exception {
       connection.close();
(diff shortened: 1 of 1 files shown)


iemejia:
Run Java JdbcIO Performance Test

jbonofre:
Thanks, let me take a look.

jbonofre:
The Jira is talking about lot of `Connection` which is closed in `teardown` already. I'm OK to close the `DataSource` itself but I'm not sure it will actually limit the connections or it's a side effect.

iemejia:
Thanks JB!

PR #8583, opened 2019-05-15
Add retry strategy when TableRow or TableCell is null from BQ response.

boyuanzz:
Patched TrafficRoutesIT and TrafficMaxLaneFlowIT.
R: @angoenka

### DIFF
--- a/examples/java/src/test/java/org/apache/beam/examples/complete/TrafficMaxLaneFlowIT.java
+++ b/examples/java/src/test/java/org/apache/beam/examples/complete/TrafficMaxLaneFlowIT.java
@@ -19,22 +19,31 @@ package org.apache.beam.examples.complete;
 
 import static org.junit.Assert.assertEquals;
 
+import com.google.api.client.util.BackOff;
+import com.google.api.client.util.BackOffUtils;
+import com.google.api.client.util.Sleeper;
 import com.google.api.services.bigquery.model.QueryResponse;
 import org.apache.beam.examples.complete.TrafficMaxLaneFlow.FormatMaxesFn;
 import org.apache.beam.examples.complete.TrafficMaxLaneFlow.TrafficMaxLaneFlowOptions;
 import org.apache.beam.sdk.extensions.gcp.options.GcpOptions;
+import org.apache.beam.sdk.extensions.gcp.util.BackOffAdapter;
 import org.apache.beam.sdk.io.gcp.testing.BigqueryClient;
 import org.apache.beam.sdk.options.PipelineOptionsFactory;
 import org.apache.beam.sdk.testing.TestPipeline;
+import org.apache.beam.sdk.util.FluentBackoff;
+import org.joda.time.Duration;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /** End-to-end tests of TrafficMaxLaneFlowIT. */
 @RunWith(JUnit4.class)
 public class TrafficMaxLaneFlowIT {
+  private static final Logger LOG = LoggerFactory.getLogger(TrafficMaxLaneFlowIT.class);
   private TrafficMaxLaneFlowOptions options;
   private final String timestamp = Long.toString(System.currentTimeMillis());
   private final String outputDatasetId = "traffic_max_lane_flow_" + timestamp;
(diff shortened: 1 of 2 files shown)


boyuanzz:
Run Java PostCommit

boyuanzz:
Run Java PostCommit

examples/java/src/test/java/org/apache/beam/examples/complete/TrafficRoutesIT.java
```diff
@@ -63,14 +69,22 @@ public void testE2ETrafficRoutes() throws Exception {
...
+    BackOff backoff = BackOffAdapter.toGcpBackOff(backoffFactory.backoff());
+    String res = "empty_result";
+    do {
+      QueryResponse response =
+          this.bqClient.queryWithRetries(
+              String.format(
+                  "SELECT count(*) as total FROM [%s:%s.%s]",
+                  this.projectId, this.outputDatasetId, this.outputTable),
+              this.projectId);
+      if (response.getRows() != null && response.getRows().get(0).getF() != null) {
```
angoenka: 1. Don't we need null check at every location?
2. Can we compare TableRow (response.getRows().get(0)) object to a dummy TableRow(new TableRow().set....) object with the expected values?

examples/java/src/test/java/org/apache/beam/examples/complete/TrafficRoutesIT.java
```diff
@@ -63,14 +69,22 @@ public void testE2ETrafficRoutes() throws Exception {
...
+        res = response.getRows().get(0).getF().get(0).getV().toString();
```
angoenka: We should terminate the `while` loop after we get the response

examples/java/src/test/java/org/apache/beam/examples/complete/TrafficRoutesIT.java
boyuanzz: 1. If every location means not only the first row, then the result table is expected to have only one row: select count.
2. I don't think TableRow is comparable directly but it can be serialized to string, which can serve as a matcher.

examples/java/src/test/java/org/apache/beam/examples/complete/TrafficRoutesIT.java
boyuanzz: Sorry for missing  a break.

boyuanzz:
Run Java PostCommit

boyuanzz:
@angoenka Would you like to do another round of review : D

boyuanzz:
Run Java PreCommit

boyuanzz:
Run Java PostCommit

examples/java/src/test/java/org/apache/beam/examples/complete/TrafficMaxLaneFlowIT.java
```diff
@@ -63,14 +72,28 @@ public void testE2ETrafficMaxLaneFlow() throws Exception {
...
-        this.bqClient.queryWithRetries(
-            String.format(
-                "SELECT count(*) as total FROM [%s:%s.%s]",
-                this.projectId, this.outputDatasetId, this.outputTable),
-            this.projectId);
-    String res = response.getRows().get(0).getF().get(0).getV().toString();
+      try {
+        res = response.getRows().get(0).getF().get(0).getV().toString();
+        break;
+      } catch (Exception e) {
```
angoenka: Shall we limit catch to NullPointerException?

boyuanzz: The overall idea is we want to retry certain times to get ride of any Exception caused by BQ not by our service. I would prefer using Exception. Instead, how about also logging stacktrace of exception?

angoenka: I think their are 2 sort of cases that needs retry.
1. Error in getting response from BQ (which is not happening in the try-catch block.
2. Early result causing null pointer. (happening in the try-catch block) 
As for logging , I think we should log stack trace after we fail the last retry.

examples/java/src/test/java/org/apache/beam/examples/complete/TrafficRoutesIT.java
```diff
@@ -63,14 +72,30 @@ public void testE2ETrafficRoutes() throws Exception {
     this.options.setBigQueryDataset(this.outputDatasetId);
     this.options.setBigQueryTable(this.outputTable);
     TrafficRoutes.runTrafficRoutes(options);
+    FluentBackoff backoffFactory =
+        FluentBackoff.DEFAULT.withMaxRetries(4).withInitialBackoff(Duration.standardSeconds(1L));
+    Sleeper sleeper = Sleeper.DEFAULT;
+    BackOff backoff = BackOffAdapter.toGcpBackOff(backoffFactory.backoff());
+    String res = "empty_result";
+    int attemptNum = 1;
```
angoenka: Unused variable

boyuanzz:
Thanks Ankur!

PR #8585, opened 2019-05-15
[BEAM-6829] Use transform/pcollection name for metric namespace if none provided

mxm:
In non-portable pipelines, transforms are always scoped by an operator. Metric
names do not have to be unique across transforms.

In portable pipelines, duplicate metric names inevitably occur if no unique
namespace has been set, due to multiple transforms being fused together inside a
single operator.

This avoids duplicate metric names by using the transform name as the namespace,
if none has been provided. It is still possible to provide a custom namespace.

--- | --- | --- | --- | ---

### DIFF
--- a/runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MonitoringInfoMetricName.java
+++ b/runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MonitoringInfoMetricName.java
@@ -52,6 +52,8 @@ public class MonitoringInfoMetricName extends MetricName {
   public String getNamespace() {
     if (labels.containsKey(MonitoringInfoConstants.Labels.NAMESPACE)) {
       return labels.getOrDefault(MonitoringInfoConstants.Labels.NAMESPACE, null);
+    } else if (labels.containsKey(MonitoringInfoConstants.Labels.PTRANSFORM)) {
+      return labels.getOrDefault(MonitoringInfoConstants.Labels.PTRANSFORM, null);
     } else {
       return urn.split(":", 2)[0];
     }
--- a/runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MonitoringInfoMetricName.java
+++ b/runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MonitoringInfoMetricName.java
@@ -54,11 +54,11 @@ public class MonitoringInfoMetricName extends MetricName {
       // User-generated metric
       return labels.getOrDefault(MonitoringInfoConstants.Labels.NAMESPACE, null);
     } else if (labels.containsKey(MonitoringInfoConstants.Labels.PCOLLECTION)) {
-      // System-generated metric, prepend with a colon
-      return ":" + labels.getOrDefault(MonitoringInfoConstants.Labels.PCOLLECTION, null);
+      // System-generated metric
+      return labels.getOrDefault(MonitoringInfoConstants.Labels.PCOLLECTION, null);
     } else if (labels.containsKey(MonitoringInfoConstants.Labels.PTRANSFORM)) {
-      // System-generated metric, prepend with a colon
-      return ":" + labels.getOrDefault(MonitoringInfoConstants.Labels.PTRANSFORM, null);
+      // System-generated metric
+      return labels.getOrDefault(MonitoringInfoConstants.Labels.PTRANSFORM, null);
     } else {
       return urn.split(":", 2)[0];
     }
--- a/runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MonitoringInfoMetricName.java
+++ b/runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MonitoringInfoMetricName.java
@@ -51,9 +51,14 @@ public class MonitoringInfoMetricName extends MetricName {
   @Override
   public String getNamespace() {
     if (labels.containsKey(MonitoringInfoConstants.Labels.NAMESPACE)) {
+      // User-generated metric
       return labels.getOrDefault(MonitoringInfoConstants.Labels.NAMESPACE, null);
+    } else if (labels.containsKey(MonitoringInfoConstants.Labels.PCOLLECTION)) {
+      // System-generated metric, prepend with a colon
+      return ":" + labels.getOrDefault(MonitoringInfoConstants.Labels.PCOLLECTION, null);
     } else if (labels.containsKey(MonitoringInfoConstants.Labels.PTRANSFORM)) {
-      return labels.getOrDefault(MonitoringInfoConstants.Labels.PTRANSFORM, null);
+      // System-generated metric, prepend with a colon
+      return ":" + labels.getOrDefault(MonitoringInfoConstants.Labels.PTRANSFORM, null);
     } else {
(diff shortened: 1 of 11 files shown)


mxm:
@tweise This gets rid of these warnings:

```
WARN org.apache.flink.metrics.MetricGroup - Name collision: Group already contains a Metric with the name ...
```

mxm:
@tweise Have you had a chance to take a look?

tweise:
Unfortunately not yet. This will take a bit more time to verify the correct propagation of the metrics to our backend.

tweise:
With this change, I still see the duplicate metrics warning:
```
[Source: Impulse -> [2]write/Write/WriteImpl/DoOnce/{FlatMap(<lambda at core.py:2254>), Map(decode)} -> [1]write/Write/WriteImpl/InitializeWrite -> (Map -> ToKeyedWorkItem, Map -> ToKeyedWorkItem, Map -> ToKeyedWorkItem) (1/1)] WARN org.apache.flink.metrics.MetricGroup - Name collision: Group already contains a Metric with the name 'beam.metric:element_count:v1'. Metric will not be reported.[localhost, taskmanager, 0bd0adb5-1fba-49c7-811f-c8ae76f83bb8, BeamApp-tweise-0531134309-499ce3fa, [2]write/Write/WriteImpl/DoOnce/{FlatMap(<lambda at core.py:2254>), Map(decode)}, 0]
```

But I also suspect that the proposed fix will have other side effects, since it is changing the metric name (at least that's how it would appear in our metric system).

The metrics as logged with the slf4j reporter before this change look like this:
```
localhost.taskmanager.87150cf7-cc34-412c-9057-656edef0590c.BeamApp-tweise-0531054435-7c2c37e1.[1]write/Write/WriteImpl/FinalizeWrite.0.numRecordsIn: 1
localhost.taskmanager.87150cf7-cc34-412c-9057-656edef0590c.BeamApp-tweise-0531054435-7c2c37e1.[3]{read, split, pair_with_one}.0.beam.metric:element_count:v1: 1
localhost.taskmanager.87150cf7-cc34-412c-9057-656edef0590c.BeamApp-tweise-0531054435-7c2c37e1.ToKeyedWorkItem.0.numRecordsOut: 0
localhost.taskmanager.87150cf7-cc34-412c-9057-656edef0590c.BeamApp-tweise-0531054435-7c2c37e1.[3]{read, split, pair_with_one}.0.ref_AppliedPTransform_read/Read/ReadSplits_16.metric:pardo_execution_time:finish_bundle_msecs:v1: 0
```
In our backend this translates to `<prefix>.operator.beam-metric-element_count-v1.gauge.sum`

If we replace `beam` with `<transform_name>`, then we would effectively change the metric name. Instead, shouldn't the transform name be part of the scope?

tweise:
Looks as if the change applies to some metrics (in the log snipped `pardo_execution_time`, but not `element_count`). But I believe we need to keep the metric name intact and augment the scope.

mxm:
@tweise You should be seeing only a fraction of the previous warnings. The reason why you see some of the warnings is that there is no "PTRANSFORM" set for the metric. In this case it looks like we do not have enough information to de-duplicate metrics. This could also be an issue with how the Beam metric label is populated.

> Instead, shouldn't the transform name be part of the scope?

The metric group is the best way to define the scope of a metric as part of an ExecutableStage operator which serves metrics for multiple transforms running inside an SDK Harness. We cannot make the transform name part of the operator scope because it hosts multiple transforms with overlapping metric names.

mxm:
@tweise I've revisited this problem and found two other important fixes in addition to the first commit:

1. We also need to include PCollection-scoped metrics, e.g. num elements
2. The metric reporting was doubled in ExecutableDoFnOperator. Once by the operator,
    once by the wrapping metrics operator.

In my tests I could not find any duplicate metrics anymore. Plus, the scoping of the metrics is done correctly, whether it is user metrics, transform metrics, or pcollection metrics.

mxm:
Test failures are unrelated: https://builds.apache.org/job/beam_PreCommit_Java_Commit/7949/
```
Test Result (2 failures / +2)

    org.apache.beam.sdk.io.TFRecordIOTest.testReadInvalidRecord
    org.apache.beam.sdk.transforms.ParDoLifecycleTest.testTeardownCalledAfterExceptionInStartBundleStateful
```

tweise:
Sounds good, I will try it again with the latest changes.

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MonitoringInfoMetricName.java
```diff
@@ -51,7 +51,14 @@ private MonitoringInfoMetricName(String urn, Map<String, String> labels) {
...
   public String getNamespace() {
     if (labels.containsKey(MonitoringInfoConstants.Labels.NAMESPACE)) {
+      // User-generated metric
       return labels.getOrDefault(MonitoringInfoConstants.Labels.NAMESPACE, null);
+    } else if (labels.containsKey(MonitoringInfoConstants.Labels.PCOLLECTION)) {
+      // System-generated metric, prepend with a colon
+      return ":" + labels.getOrDefault(MonitoringInfoConstants.Labels.PCOLLECTION, null);
+    } else if (labels.containsKey(MonitoringInfoConstants.Labels.PTRANSFORM)) {
+      // System-generated metric, prepend with a colon
+      return ":" + labels.getOrDefault(MonitoringInfoConstants.Labels.PTRANSFORM, null);
```
mxm: Looking at this again, we may not want to add the colon. My concern here was possible clash with the user metrics namespace. However, the system metric names already contain a `beam.` prefix which might be sufficient.

mxm:
>Sounds good, I will try it again with the latest changes.

Thanks, curious to see the results! :)

tweise:
Run Java PreCommit

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MonitoringInfoMetricName.java
tweise: Agreed, there should be no need for this prefix. I'm going to remove it. I wouldn't be surprised though if we revisit the metric names as we test this with a real world metric backend. The current names are quite verbose. Examples:

`36read/Read/Reshuffle/RemoveRandomKeys.None/beam:env:docker:v1:0:beam:metric:user_distribution {NAMESPACE=__main__.WordExtractingDoFn, PTRANSFORM=ref_AppliedPTransform_split_17, NAME=word_len_dist}: DistributionResult{sum=114, count=29, min=0, max=11}`

(user metric created as `self.word_lengths_dist = Metrics.distribution(self.__class__, 'word_len_dist')`

or 

`17read/Read/Impulse.None/beam:env:docker:v1:0:beam:metric:sampled_byte_size:v1 {PCOLLECTION=ref_PCollection_PCollection_1}: DistributionResult{sum=13, count=1, min=13, max=13}, 6format.None/beam:env:docker:v1:0:beam:metric:sampled_byte_size:v1 {PCOLLECTION=ref_PCollection_PCollection_17}: DistributionResult{sum=395, count=18, min=19, max=29}`

tweise:
Manual test shows that duplicate metric warnings are gone.

PR #8597, opened 2019-05-17
[BEAM-7190] Enable file based token auth for samza portable runner

lhaiesp:
For Samza and potentially other portable runners who do not use docker and need to run on multi-tenant environment, there is a need to secure the communication between sdk worker and runner. Currently the SSL/TLS in portability is half done.

However, after investigation we found that it's sufficient to just
1. Use loopback address. So that the traffic is not exposed to external network
2. Enforce authentication. So that only the valid users can connect to the ports. 

With the two steps above, it won't be necessary to enable TLS. Because the data channel is only local and one needs root privilege to eavesdrop the local traffic.

A trivial way to do authentication is to share a secret token through file system (e.g. set the file permission to be 600, i.e. -rw-------) . Next we introduce a customized interpreter for both the gRPC client and server to provide and verify this token (see GrpcFileTokenAuthProvider.java and token_auth_interceptor.py). The server can then deny any connection attempts that do not have the right token.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/GrpcFileTokenAuthProvider.java
+++ b/runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/GrpcFileTokenAuthProvider.java
@@ -1,0 +1,57 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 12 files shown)


angoenka:
- Can we make the token creation and authentication modular and pluggable so that it can be added to other runners as well by setting a pipeline option. 
- We will also need secure channel to encrypt the content to call it truely secure.

lhaiesp:
@angoenka can you be a little bit more specific about making it modular? Right now runners instantiate and start gRPC servers on their own. What I could do now is to put more logic into a util function but ultimately each runner needs to integrate with it themselves.

On your second point, the channel is effectively "secure" because 1) traffic is loopback only so there is no real network packet 2) the loopback traffic can not be eavesdropped unless by a root user, who can basically do anything including reading the private keys on file on in memory

angoenka:
We are essentially creating a new header interceptor to validate the information in the header. We can make it modular by letting users specify the interceptor when starting job server.
The jobserver can convey the interceptor class name to the Samza/Flink/Spark ExecutableStage function which will add the interceptor when creating the Channels. A similar option can be passed to SDK Harness to add an arbitrary grpc interceptor.

To implement the use case of token based validation, Samza jobserver can simply instrument these options correctly.

lhaiesp:
Sorry for being inactive here. We were busy with many internal issues when onboarding users to production. I'm back to work on it this week. Will consume the feedback and make the change. Thanks!

PR #8598, opened 2019-05-17
[BEAM-7322] Add threshold to PubSub unbounded source

tims:
This PR adds a threshold to the pubsub unbounded source to ensure that the watermark advances for unbounded streams. It's approaches the problem identically to how the Kinesis source deals with it.

@rezarokni can you review? also any advice on how to write a test for this?

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/pubsub/PubsubUnboundedSource.java
+++ b/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/pubsub/PubsubUnboundedSource.java
@@ -993,8 +993,7 @@ public class PubsubUnboundedSource extends PTransform<PBegin, PCollection<Pubsub
       boolean hasEnoughSamples =
           minReadTimestampMsSinceEpoch.isSignificant()
               || minUnreadTimestampMsSinceEpoch.isSignificant();
-      boolean isStale =
-          lastWatermarkMsSinceEpoch < (nowMsSinceEpoch - UPDATE_THRESHOLD.getMillis());
+      boolean isStale = lastReceivedMsSinceEpoch < (nowMsSinceEpoch - UPDATE_THRESHOLD.getMillis());
       return hasEnoughSamples || isStale;
     }
 
--- a/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/pubsub/PubsubUnboundedSource.java
+++ b/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/pubsub/PubsubUnboundedSource.java
@@ -982,17 +982,19 @@ public class PubsubUnboundedSource extends PTransform<PBegin, PCollection<Pubsub
 
     /**
      * In case of streams with low traffic, {@link MovingFunction} could never get enough samples in
-     * {@link PubsubUnboundedSource#SAMPLE_PERIOD} to move watermark. To prevent this situation, we need to
-     * check if watermark is stale (it was not updated during {@link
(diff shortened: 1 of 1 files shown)


tims:
R: @rezarokni

tims:
R: @kennknowles

kennknowles:
@reuvenlax @slavachernyak are probably better than myself for this code review

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/pubsub/PubsubUnboundedSource.java
```diff
@@ -973,6 +980,24 @@ public Instant getWatermark() {
...
+     *
+     * @param nowMsSinceEpoch - current timestamp
+     * @return should the watermark be updated
+     */
+    private boolean shouldUpdate(long nowMsSinceEpoch) {
+      boolean hasEnoughSamples =
+          minReadTimestampMsSinceEpoch.isSignificant()
+              || minUnreadTimestampMsSinceEpoch.isSignificant();
+      boolean isStale =
+          lastWatermarkMsSinceEpoch < (nowMsSinceEpoch - UPDATE_THRESHOLD.getMillis());
```
reuvenlax: This check doesn't test whether the watermark hasn't been updated, it's actually testing whether the watermark is old. However the watermark might be old because the data is old (maybe the pipeline was started on old data and is catching up). I think you actually want to store the current time whenever you allow the watermark to advance, and ensure that you never stop it for too long.

slavachernyak: I agree with reuvenlax, what we need here is a notion of "how long its been since the watermark was updated" not "how old is the watermark" 

Also, it might be better if we defer deciding that we want to update the watermark due to staleness not only until it hasn't been updated in a while, but also until we haven't seen any new data arrive for a while. It seems that as long as data is continuously arriving, hasEnoughSamples should eventually be true, or are we worried about data arriving so infrequently it is aged out of the minReadTimestampMsSinceEpoch MovingFunction before it can ever let isSignificant evaluate to true?

tims: Oops... 
I fixed it to use lastReceivedMsSinceEpoch. 

I am still struggling to write a test for this, which would have caught this, I'm finding it difficult to understand the PubsubUnboundedSourceTest code, but plodding away. 

I was trying to adapt the readManyMessages tests and ran into some issues with PubsubTestClient state not being cleared if a test fails to close the factory in the @after clause, which then breaks all other pubsub tests! This is because the PubsubTestClientFactory cannot be closed if it still has pending messages. Any test guidance appreciated. 

Also, what do you think of the idea of making SAMPLE_PERIOD configurable? So that we could advance the watermark very optimistically if we wanted to (eg: when demonstrating and teaching Beam)? When done, this PR will ensure very low traffic subscriptions' watermarks advance, but it will still have a 2 minute delay. Could be a separate PR.

tims: > ... It seems that as long as data is continuously arriving, hasEnoughSamples should eventually be true, or are we worried about data arriving so infrequently it is aged out of the minReadTimestampMsSinceEpoch MovingFunction before it can ever let isSignificant evaluate to true?

In my tests, I found that if a single message arrives with intervals of ten seconds, after 10 minutes it still hasn't advanced the watermark. So that might be the issues, I'm trying to write tests to verify it.

tvalentyn:
What's next on this PR?

nullobject:
Is this still a thing? I ran into this problem today with the latest version of Beam (2.19).

PR #8607, opened 2019-05-17
[BEAM-7342] Extend SyntheticPipeline map steps to be splittable.

laraschmidt:
Extending SyntheticPipeline map steps to be splittable.

R: @lukecwik 
R: @boyuanzz 
 
--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/testing/synthetic_pipeline.py
+++ b/sdks/python/apache_beam/testing/synthetic_pipeline.py
@@ -62,7 +62,7 @@ except ImportError:
 def parse_byte_size(s):
   suffixes = 'BKMGTP'
   if s[-1] in suffixes:
-    return int(float(s[:-1]) * 1024**suffixes.index(s[-1]))
+    return int(float(s[:-1]) * 1024 ** suffixes.index(s[-1]))
 
   return int(s)
 
@@ -223,7 +223,7 @@ def getSyntheticSDFStep(per_element_delay_sec=0,
       # the given number of seconds, so we only sleep enough to make up for
       # overheads not incurred elsewhere.
       to_sleep = self._per_bundle_delay_sec - (
-        time.time() - self._start_time)
+          time.time() - self._start_time)
 
       # Ignoring sub-millisecond sleep times.
       if to_sleep >= 1e-3:
@@ -300,8 +300,8 @@ class SyntheticSource(iobase.BoundedSource):
       self._initial_splitting_distribution_parameter = 0
     self._dynamic_splitting = (
       'none' if (
-        'splitPointFrequencyRecords' in input_spec
-        and input_spec['splitPointFrequencyRecords'] == 0)
+          'splitPointFrequencyRecords' in input_spec
+          and input_spec['splitPointFrequencyRecords'] == 0)
       else 'perfect')
     if 'delayDistribution' in input_spec:
       if input_spec['delayDistribution']['type'] != 'const':
@@ -310,9 +310,9 @@ class SyntheticSource(iobase.BoundedSource):
           'distributions of type \'const\'. Received %s.',
           input_spec['delayDistribution']['type'])
       self._sleep_per_input_record_sec = (
-        float(input_spec['delayDistribution']['const']) / 1000)
+          float(input_spec['delayDistribution']['const']) / 1000)
       if (self._sleep_per_input_record_sec and
-        self._sleep_per_input_record_sec < 1e-3):
+          self._sleep_per_input_record_sec < 1e-3):
         raise ValueError(
           'Sleep time per input record must be at least 1e-3.'
           ' Received: %r', self._sleep_per_input_record_sec)
@@ -431,10 +431,10 @@ class SyntheticSDFSourceRestrictionProvider(RestrictionProvider):
     estimate_size = element_size * element['num_records']
     if element['initial_splitting'] == 'zipf':
       desired_num_bundles = (
-        element['initial_splitting_num_bundles'] or
-        div_round_up(estimate_size,
-                     element[
-                       'initial_splitting_desired_bundle_size']))
+          element['initial_splitting_num_bundles'] or
+          div_round_up(estimate_size,
+                       element[
+                         'initial_splitting_desired_bundle_size']))
       samples = np.random.zipf(
         element['initial_splitting_distribution_parameter'],
         desired_num_bundles)
@@ -505,10 +505,10 @@ class SyntheticSDFAsSource(beam.DoFn):
   """
 
   def process(
-    self,
-    element,
-    restriction_tracker=beam.DoFn.RestrictionParam(
-      SyntheticSDFSourceRestrictionProvider())):
+      self,
+      element,
+      restriction_tracker=beam.DoFn.RestrictionParam(
+        SyntheticSDFSourceRestrictionProvider())):
     for k in range(*restriction_tracker.current_restriction()):
       if not restriction_tracker.try_claim(k):
         return
@@ -524,7 +524,7 @@ class ShuffleBarrier(beam.PTransform):
             | beam.Map(rotate_key)
             | beam.GroupByKey()
             | 'Ungroup' >> beam.FlatMap(
-        lambda elm: [(elm[0], v) for v in elm[1]]))
+          lambda elm: [(elm[0], v) for v in elm[1]]))
 
 
 class SideInputBarrier(beam.PTransform):
@@ -533,8 +533,8 @@ class SideInputBarrier(beam.PTransform):
     return (pc
             | beam.Map(rotate_key)
             | beam.Map(
-        lambda elem, ignored: elem,
-        beam.pvalue.AsIter(pc | beam.FlatMap(lambda elem: None))))
+          lambda elem, ignored: elem,
+          beam.pvalue.AsIter(pc | beam.FlatMap(lambda elem: None))))
 
 
 def merge_using_gbk(name, pc1, pc2):
@@ -544,8 +544,8 @@ def merge_using_gbk(name, pc1, pc2):
   pc2_with_key = pc2 | (name + 'AttachKey2') >> beam.Map(lambda x: (x, x))
 
   grouped = (
-    {'pc1': pc1_with_key, 'pc2': pc2_with_key} |
-    (name + 'Group') >> beam.CoGroupByKey())
+      {'pc1': pc1_with_key, 'pc2': pc2_with_key} |
+      (name + 'Group') >> beam.CoGroupByKey())
   return (grouped |
           (name + 'DeDup') >> beam.Map(lambda elm: elm[0]))  # Ignoring values
 
@@ -712,7 +712,7 @@ def run(argv=None):
 
     pc_list = []
     num_roots = 2 ** (len(known_args.steps) - 1) if (
-      barrier == 'merge-gbk' or barrier == 'merge-side-input') else 1
+        barrier == 'merge-gbk' or barrier == 'merge-side-input') else 1
     for read_no in range(num_roots):
       pc_list.append((p | ('Read %d' % read_no) >> beam.io.Read(source)))
 
--- a/sdks/python/apache_beam/testing/synthetic_pipeline.py
+++ b/sdks/python/apache_beam/testing/synthetic_pipeline.py
@@ -45,9 +45,9 @@ import apache_beam as beam
 from apache_beam.io import WriteToText
 from apache_beam.io import iobase
 from apache_beam.io import range_trackers
+from apache_beam.io import restriction_trackers
 from apache_beam.io.restriction_trackers import OffsetRange
 from apache_beam.io.restriction_trackers import OffsetRestrictionTracker
-from apache_beam.io import restriction_trackers
 from apache_beam.options.pipeline_options import PipelineOptions
 from apache_beam.options.pipeline_options import SetupOptions
 from apache_beam.testing.test_pipeline import TestPipeline
@@ -105,18 +105,17 @@ class SyntheticStep(beam.DoFn):
     if to_sleep >= 1e-3:
       time.sleep(to_sleep)
 
-    def process(self, element):
-      print("normal step")
-      if self._per_element_delay_sec >= 1e-3:
-        time.sleep(self._per_element_delay_sec)
-      filter_element = False
-      if self._output_filter_ratio > 0:
-        if np.random.random() < self._output_filter_ratio:
-          filter_element = True
+  def process(self, element):
+    if self._per_element_delay_sec >= 1e-3:
+      time.sleep(self._per_element_delay_sec)
+    filter_element = False
+    if self._output_filter_ratio > 0:
+      if np.random.random() < self._output_filter_ratio:
+        filter_element = True
 
-      if not filter_element:
-        for _ in range(self._output_records_per_input_record):
-          yield element
+    if not filter_element:
+      for _ in range(self._output_records_per_input_record):
+        yield element
 
 
 class NonLiquidShardingOffsetRangeTracker(OffsetRestrictionTracker):
@@ -149,19 +148,15 @@ class SyntheticSDFStepRestrictionProvider(RestrictionProvider):
     self._initial_splitting_uneven_chunks = initial_splitting_uneven_chunks
     self._disable_liquid_sharding = disable_liquid_sharding
     self._size_estimate_override = size_estimate_override
-    print('here')
-    print(disable_liquid_sharding)
 
   def initial_restriction(self, element):
     return (0, self._num_records)
 
   def create_tracker(self, restriction):
     if self._disable_liquid_sharding:
-      print('here!')
       return NonLiquidShardingOffsetRangeTracker(restriction[0],
                                                  restriction[1])
     else:
-      print('here2!')
       # OffsetRange.new_tracker returns a RangeTracker - not RestrictionTracker.
       return OffsetRestrictionTracker(restriction[0], restriction[1])
 
@@ -169,7 +164,7 @@ class SyntheticSDFStepRestrictionProvider(RestrictionProvider):
     ''' Randomly split the restriction into the right number of bundles.'''
     elems = restriction[1] - restriction[0]
     bundles = self._initial_splitting_num_bundles
-    randomNums = [np.random.randint(0, elems - 1) for x in
+    randomNums = [np.random.randint(0, elems - 1) for _ in
                   range(0, bundles - 1)]
     print randomNums
     randomNums.append(0)
@@ -181,7 +176,9 @@ class SyntheticSDFStepRestrictionProvider(RestrictionProvider):
 
   def split(self, element, restriction):
     elems = restriction[1] - restriction[0]
-    if self._initial_splitting_uneven_chunks and self._initial_splitting_num_bundles > 1 and elems > 1:
+    if (self._initial_splitting_uneven_chunks and
+        self._initial_splitting_num_bundles > 1 and
+        elems > 1):
       return self.split_randomly(restriction)
     else:
       offsets_per_split = max(1, (elems // self._initial_splitting_num_bundles))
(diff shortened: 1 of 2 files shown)


sdks/python/apache_beam/testing/synthetic_pipeline.py
```diff
@@ -115,6 +115,108 @@ def process(self, element):
...
+      bundle_ranges.append((start_position, final_position))
+      start_position = final_position
+    return bundle_ranges
+
+  def restriction_size(self, element, restriction):
+    (start, stop) = restriction
+    element_size = len(element) if isinstance(element, str) else 1
+    return (stop - start) * element_size
+
+""" A function which returns a SyntheticSDFStep with given parameters. """
```
boyuanzz: We would prefer comments going below the function signature, like:
```python
def getSyntheticSDFStep():
"""some comments"""
```
Otherwise, lint check may complain. You can run ./gradlew :sdks:python:lintPy27  to run lint check.

sdks/python/apache_beam/testing/synthetic_pipeline.py
```diff
@@ -450,6 +552,9 @@ def _parse_steps(json_str):
         for each input element to a step.
     (4) output_filter_ratio - the probability at which a step may filter out a
         given element by not producing any output for that element.
+    (5) splittable - if the step should be splittable.
+    (6) initial_splitting_num_bundles - number of bundles initiall split if step
```
lukecwik:
```suggestion
    (6) initial_splitting_num_bundles - number of bundles initial split if step
```

sdks/python/apache_beam/testing/synthetic_pipeline.py
```diff
@@ -496,7 +607,9 @@ def parse_args(args):
            '    Defaults to 0.'
            '(3) An integer "output_records_per_input_record". Defaults to 1.'
            '(4) A float "output_filter_ratio" in the range [0, 1] . '
-           '    Defaults to 0.')
+           '    Defaults to 0.'
+           '(5) A bool "splittable" that defaults to false.'
+           '(6) A integer "initial_splitting_num_bundles". Defaults to 2.')
```
lukecwik: How about we do something like 8, so that its larger then splitting that LS would provide.

sdks/python/apache_beam/testing/synthetic_pipeline.py
```diff
@@ -115,6 +115,108 @@ def process(self, element):
       for _ in range(self._output_records_per_input_record):
         yield element
 
+class SyntheticSDFStepRestrictionProvider(RestrictionProvider):
+  """A `RestrictionProvider` for SyntheticSDFStep.
+
+  In initial_restriction and split that operates on num_records and ignores
+  source description (element).
+
```
lukecwik:
```suggestion
```

sdks/python/apache_beam/testing/synthetic_pipeline.py
```diff
@@ -115,6 +115,108 @@ def process(self, element):
...
+  """
+
+  def __init__(self, num_records, initial_splitting_num_bundles):
+    self._num_records = num_records
+    self._initial_splitting_num_bundles = initial_splitting_num_bundles
+
+  def initial_restriction(self, element):
+    return (0, self._num_records)
```
lukecwik: Can we return an OffsetRange instead of a tuple?

sdks/python/apache_beam/testing/synthetic_pipeline.py
```diff
@@ -115,6 +115,108 @@ def process(self, element):
...
+      if i == num_bundles - 1:
+        final_position = stop_position  # Final bundle goes to end
+      bundle_ranges.append((start_position, final_position))
+      start_position = final_position
+    return bundle_ranges
+
+  def restriction_size(self, element, restriction):
+    (start, stop) = restriction
+    element_size = len(element) if isinstance(element, str) else 1
+    return (stop - start) * element_size
```
lukecwik: Why the multiplication?

sdks/python/apache_beam/testing/synthetic_pipeline.py
```diff
@@ -115,6 +115,108 @@ def process(self, element):
...
+        restriction[0], restriction[1])
+
+  def split(self, element, restriction):
+    bundle_ranges = []
+    start_position, stop_position = restriction
+    if self._initial_splitting_num_bundles < 2:
+      bundle_ranges.append((start_position, stop_position))
+      return bundle_ranges
+    num_bundles = self._initial_splitting_num_bundles
+    num_records_per_bundle = (stop_position - start_position) // num_bundles
```
lukecwik: What about using https://github.com/apache/beam/blob/069fc3de95bd96f34c363308ad9ba988ab58502d/sdks/python/apache_beam/io/restriction_trackers.py#L53 to perform the splitting logic?

This would fall out naturally if we used OffsetRange as the restriction type.

sdks/python/apache_beam/testing/synthetic_pipeline.py
```diff
@@ -115,6 +115,108 @@ def process(self, element):
...
+      if to_sleep >= 1e-3:
+        time.sleep(to_sleep)
+
+    def process(self,
+                element,
+                restriction_tracker=beam.DoFn.RestrictionParam(
+                    SyntheticSDFStepRestrictionProvider(
+                        output_records_per_input_record,
+                        initial_splitting_num_bundles))):
+      if self._per_element_delay_sec >= 1e-3:
```
lukecwik: We prevent this from happening in the constructor so the check is unnecessary

sdks/python/apache_beam/testing/synthetic_pipeline.py
```diff
@@ -115,6 +115,108 @@ def process(self, element):
...
+      filter_element = False
+      if self._output_filter_ratio > 0:
+        if np.random.random() < self._output_filter_ratio:
+          filter_element = True
+
+      for k in range(*restriction_tracker.current_restriction()):
+        if not restriction_tracker.try_claim(k):
+          return
+
+        if self._per_element_delay_sec >= 1e-3:
```
lukecwik: We prevent this from happening in the constructor so the check is unnecessary

sdks/python/apache_beam/testing/synthetic_pipeline.py
boyuanzz: Any reason why prefer using OffRange rather than a (start_pos, end_pos)? If using OffsetRange instead of a simple tuple, then a RestrictionProvider can only create an OffRangeTracker in process fn, right? I'm not sure whether there are other links of tracker expected to be created, though.

lukecwik: The SDF is not required to use an OffsetRangeTracker. Any restriction tracker that uses OffsetRange as the restriction type is valid.

sdks/python/apache_beam/testing/synthetic_pipeline.py
boyuanzz: I believe the original purpose is to get the total size of current element. Imaging current element is a list of records, then the total size would be (stop - start) * element_size, where element_size would be a similar attr as _num_records in SyntheticSDFStepRestrictionProvider.

sdks/python/apache_beam/testing/synthetic_pipeline.py
laraschmidt: removed this entirely actually. We only need it per record so not sure why it was there twice.

sdks/python/apache_beam/testing/synthetic_pipeline.py
laraschmidt: Still need to check if it's greater than 0, right? updated.

sdks/python/apache_beam/testing/synthetic_pipeline.py
laraschmidt: Yeah that's right. It's a real size we need to provide, not just the number of elements which is not really useful in determining size.

sdks/python/apache_beam/testing/synthetic_pipeline.py
laraschmidt: Sounds good.

sdks/python/apache_beam/testing/synthetic_pipeline.py
laraschmidt: Still have to check if it's greater then 0 I think.

sdks/python/apache_beam/testing/synthetic_pipeline.py
laraschmidt: Sounds good.

laraschmidt:
I addressed the comments, tests pass, and linter is happy. Should be ready for another round of review. Thanks!
@lukecwik @boyuanzz

sdks/python/apache_beam/testing/synthetic_pipeline.py
```diff
@@ -116,6 +118,149 @@ def process(self, element):
...
+      # OffsetRange.new_tracker returns a RangeTracker - not RestrictionTracker.
+      return OffsetRestrictionTracker(restriction[0], restriction[1])
+
+  def split_randomly(self, restriction):
+    ''' Randomly split the restriction into the right number of bundles.'''
+    elems = restriction[1] - restriction[0]
+    bundles = self._initial_splitting_num_bundles
+    randomNums = [np.random.randint(0, elems - 1) for _ in
+                  range(0, bundles - 1)]
+    print randomNums
```
lukecwik:
```suggestion
```

sdks/python/apache_beam/testing/synthetic_pipeline.py
```diff
@@ -116,6 +118,149 @@ def process(self, element):
         yield element
 
 
+class NonLiquidShardingOffsetRangeTracker(OffsetRestrictionTracker):
+  """ A OffsetRangeTracker that doesn't allow splitting. """
```
lukecwik:
```suggestion
  """An OffsetRangeTracker that doesn't allow splitting."""
```

sdks/python/apache_beam/testing/synthetic_pipeline.py
```diff
@@ -116,6 +118,149 @@ def process(self, element):
...
+    return  # Don't split.
+
+  def checkpoint(self):
+    return # Don't split.
+
+
+class SyntheticSDFStepRestrictionProvider(RestrictionProvider):
+  """A `RestrictionProvider` for SyntheticSDFStep.
+
+  An initial_restriction and split that operate on num_records and ignore
```
lukecwik:
```suggestion
  An initial_restriction and split that operate on num_records and ignores
```

sdks/python/apache_beam/testing/synthetic_pipeline.py
```diff
@@ -116,6 +118,149 @@ def process(self, element):
...
+  def create_tracker(self, restriction):
+    if self._disable_liquid_sharding:
+      return NonLiquidShardingOffsetRangeTracker(restriction[0],
+                                                 restriction[1])
+    else:
+      # OffsetRange.new_tracker returns a RangeTracker - not RestrictionTracker.
+      return OffsetRestrictionTracker(restriction[0], restriction[1])
+
+  def split_randomly(self, restriction):
+    ''' Randomly split the restriction into the right number of bundles.'''
```
lukecwik:
```suggestion
    '''Randomly split the restriction into the right number of bundles.'''
```

sdks/python/apache_beam/testing/synthetic_pipeline.py
```diff
@@ -116,6 +118,149 @@ def process(self, element):
...
+    print randomNums
+    randomNums.append(0)
+    randomNums.append(elems)
+    randomNums.sort()
+    for i in range(1, len(randomNums)):
+      yield (restriction[0] + randomNums[i - 1],
+             restriction[0] + randomNums[i])
+
+  def split(self, element, restriction):
+    elems = restriction[1] - restriction[0]
```
lukecwik: Can we factor out the logic here: https://github.com/apache/beam/blob/7943d5ff74df200a821d72d5ba24558f523eb322/sdks/python/apache_beam/testing/synthetic_pipeline.py#L196

and use it in both places via a shared method.

Use zipf distribution for "uneven" splits and falling back to the `_initial_splitting_num_bundles` logic for "even" size bundles.

We could either take a parameter or default to a zipf distribution with a 3.0 factor for now.

sdks/python/apache_beam/testing/synthetic_pipeline.py
```diff
@@ -116,6 +118,149 @@ def process(self, element):
...
+
+def getSyntheticSDFStep(per_element_delay_sec=0,
+                        per_bundle_delay_sec=0,
+                        output_records_per_input_record=1,
+                        output_filter_ratio=0,
+                        initial_splitting_num_bundles=8,
+                        initial_splitting_uneven_chunks=False,
+                        disable_liquid_sharding=False,
+                        size_estimate_override=None,):
+  """ A function which returns a SyntheticSDFStep with given parameters. """
```
lukecwik:
```suggestion
  """A function which returns a SyntheticSDFStep with given parameters."""
```

sdks/python/apache_beam/testing/synthetic_pipeline.py
```diff
@@ -116,6 +118,149 @@ def process(self, element):
...
+      filter_element = False
+      if self._output_filter_ratio > 0:
+        if np.random.random() < self._output_filter_ratio:
+          filter_element = True
+
+      for k in range(*restriction_tracker.current_restriction()):
+        if not restriction_tracker.try_claim(k):
+          return
+
+        if self._per_element_delay_sec > 0:
```
lukecwik: In the synthetic pipeline map step, the per element delay is how much we should wait in time per **input** element and not per output that we produce.

I think we should divide the wait time by num elements we are going to emit and perform these smaller sleeps to make progress through the SDF, wdyt?

sdks/python/apache_beam/testing/synthetic_pipeline.py
```diff
@@ -496,7 +663,9 @@ def parse_args(args):
            '    Defaults to 0.'
            '(3) An integer "output_records_per_input_record". Defaults to 1.'
            '(4) A float "output_filter_ratio" in the range [0, 1] . '
-           '    Defaults to 0.')
+           '    Defaults to 0.'
+           '(5) A bool "splittable" that defaults to false.'
+           '(6) A integer "initial_splitting_num_bundles". Defaults to 8.')
```
lukecwik:
```suggestion
           '(6) An integer "initial_splitting_num_bundles". Defaults to 8.')
```

sdks/python/apache_beam/testing/synthetic_pipeline_test.py
```diff
@@ -69,6 +69,102 @@ def testSyntheticStep(self):
...
+          synthetic_pipeline.getSyntheticSDFStep(0, 0.5, 10))
+      assert_that(
+          pcoll | beam.combiners.Count.Globally(), equal_to([100]))
+
+    elapsed = time.time() - start
+    # TODO(chamikaramj): Fix the flaky time based bounds.
+    self.assertTrue(0.5 <= elapsed <= 3, elapsed)
+
+  def testSyntheticStepSplitProvider(self):
+    provider = synthetic_pipeline.SyntheticSDFStepRestrictionProvider(5, 2,
```
lukecwik: nit: It should be easier to read if you format the following way (ditto for the lines below)
```
provider = synthetic_pipeline.SyntheticSDFStepRestrictionProvider(
    5, 2, False, False, None)
```

sdks/python/apache_beam/testing/synthetic_pipeline_test.py
```diff
@@ -69,6 +69,102 @@ def testSyntheticStep(self):
...
+
+    prior_stop = start
+    for r in ranges:
+      self.assertEquals(r[0], prior_stop)
+      prior_stop = r[1]
+    self.assertEquals(prior_stop, stop)
+    self.assertEquals(len(ranges), bundles)
+
+  def testSyntheticStepSplitProviderUnevenChunks(self):
+
```
lukecwik:
```suggestion
```

sdks/python/apache_beam/testing/synthetic_pipeline.py
```diff
@@ -116,6 +118,149 @@ def process(self, element):
         yield element
 
 
+class NonLiquidShardingOffsetRangeTracker(OffsetRestrictionTracker):
+  """ A OffsetRangeTracker that doesn't allow splitting. """
+
+  def try_split(self, split_offset):
+    return  # Don't split.
```
boyuanzz: For non-op purpose, maybe ```pass``` fits better here?

sdks/python/apache_beam/testing/synthetic_pipeline_test.py
```diff
@@ -111,7 +207,12 @@ def testSplitAtFraction(self):
...
 
   def run_pipeline(self, barrier, writes_output=True):
-    steps = [{'per_element_delay': 1}, {'per_element_delay': 1}]
+    steps = [{
+        'per_element_delay': 1
+    }, {
+        'per_element_delay': 1,
+        'splittable': True
+    }]
     args = ['--barrier=%s' % barrier, '--runner=DirectRunner',
```
boyuanzz: I would prefer using ```--runner=SwitchDirecrtRunner```, which will using a ```FnApiRunner``` instead of one non-portable ```DirectRunner``` if trying to test SDF. An SDF can be splittable only over fnapi. I'm also curious why it works on ```DirectRunner``` without involving SDF expansion and SDF-specific operation.
I may not have best judgement here though. @lukecwik what do you think here?

lukecwik: The DirectRunner is currently doing an expansion here: https://github.com/apache/beam/blob/ccbad612a49ad1eacc5f515da6a3fe417b9cc7ce/sdks/python/apache_beam/runners/direct/direct_runner.py#L221

boyuanzz: Good to know. Thanks!

laraschmidt: The point of adding it there was mostly just to check that the plumbing and args were set up correctly and to just use it somewhere so I think it should be fine. The splitting is tested via unit tests.

sdks/python/apache_beam/testing/synthetic_pipeline.py
laraschmidt: Done. I thought about making the config name the same but initial_splitting seems less obvious then if it supports uneven chunks.

Otherwise I'll just set it at 3.0 unless we think we'll need to mess with that.

laraschmidt:
Ready for another round. Thanks! @boyuanzz @lukecwik

sdks/python/apache_beam/testing/synthetic_pipeline.py
```diff
@@ -76,8 +78,34 @@ def rotate_key(element):
...
+    stop = start + int(num_total_records * relative_bundle_sizes[index])
+    bundle_ranges.append((start, stop))
+    start = stop
+    index += 1
+  return bundle_ranges
+
+
 class SyntheticStep(beam.DoFn):
   """A DoFn of which behavior can be controlled through prespecified parameters.
+
```
lukecwik:
```suggestion
```

sdks/python/apache_beam/testing/synthetic_pipeline.py
lukecwik: You can also factor out the offsets_per_split logic that you have below to reuse https://github.com/apache/beam/blob/2471e3e4656806f4b1aefa2d677ef0e3256b41f1/sdks/python/apache_beam/testing/synthetic_pipeline.py#L214
I'm just trying to ensure that the two behaviors don't diverge in how the splits are computed.

sdks/python/apache_beam/testing/synthetic_pipeline.py
laraschmidt: Seems reasonable. Done.

laraschmidt:
@lukecwik @boyuanzz Ready for another round, thanks!

lukecwik:
Run Python PreCommit

PR #8613, opened 2019-05-18
[BEAM-7356] Support generic field selections in Filter.

reuvenlax:
This allows filtering on nested fields

### DIFF
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/transforms/Filter.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/transforms/Filter.java
@@ -94,9 +94,9 @@ public class Filter {
       abstract SerializableFunction<FieldT, Boolean> getPredicate();
 
       @Nullable
-      abstract Schema getOutputSchema();
+      abstract Schema getSelectedSchema();
 
-      abstract boolean getUnbox();
+      abstract boolean getSelectsSingleField();
 
       abstract Builder<FieldT> toBuilder();
 
@@ -107,9 +107,9 @@ public class Filter {
 
         abstract Builder<FieldT> setPredicate(SerializableFunction<FieldT, Boolean> predicate);
 
-        abstract Builder<FieldT> setOutputSchema(@Nullable Schema outputSchema);
+        abstract Builder<FieldT> setSelectedSchema(@Nullable Schema selectedSchema);
 
-        abstract Builder<FieldT> setUnbox(boolean unbox);
+        abstract Builder<FieldT> setSelectsSingleField(boolean unbox);
 
         abstract FilterDescription<FieldT> build();
       }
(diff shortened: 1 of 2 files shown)


reuvenlax:
R: @robinyqiu

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/transforms/Filter.java
```diff
@@ -84,95 +87,128 @@
...
-    private final Map<List<Integer>, SerializableFunction<Row, Boolean>> fieldIdsFilters =
-        Maps.newHashMap();
+    @AutoValue
+    abstract static class FilterDescription<FieldT> implements Serializable {
+      abstract FieldAccessDescriptor getFieldAccessDescriptor();
+
+      abstract SerializableFunction<FieldT, Boolean> getPredicate();
+
+      @Nullable
+      abstract Schema getOutputSchema();
```
robinyqiu: Nit: I found the name `outputSchema` a bit confusing, maybe something like `selectedSchema` or 'accessedSchema' is a better name?

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/transforms/Filter.java
```diff
@@ -84,95 +87,128 @@
...
+
+      abstract boolean getUnbox();
```
robinyqiu: Nit: same as above, maybe `selectedSingleField` or `accessedSingleField`?

reuvenlax:
"generic" in the PR title meant generic field access strings. This PR
allows passing "x.y.z" into filter to filter on subfields, while the
previous Filter using did not allow this.

I made one change to actual Java generics, to make the transform better
able to infer types from lambdas. Previously you had to cast the type
inside your lambda, and now that is not necessary.

Good point on confusing names. I'll try and fix those.

robinyqiu:
Thanks for the explanation! That answers my question.

Another quick question:
I think we can't use `java.util.function.Predicate<FieldT>` to replace `SerializableFunction<FieldT, Boolean>` defined in Beam because `Predicate` doesn't implement `Serializable`, right? Then maybe adding something like `SerializablePredicate` to the class of functional interfaces defined by Beam will be helpful.

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/transforms/Filter.java
reuvenlax: Done

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/transforms/Filter.java
reuvenlax: done

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/transforms/Filter.java
robinyqiu: Thank you for the changes! LGTM.

PR #8629, opened 2019-05-20
[BEAM-6284] Improve error message on waitUntilFinish.

Ardagan:
Allow for infinite wait.

Seems that [BEAM-6284] is relevant to this issue, even though logs are not available any more.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/runners/google-cloud-dataflow-java/src/main/java/org/apache/beam/runners/dataflow/DataflowPipelineJob.java
+++ b/runners/google-cloud-dataflow-java/src/main/java/org/apache/beam/runners/dataflow/DataflowPipelineJob.java
@@ -38,6 +38,7 @@ import java.util.concurrent.atomic.AtomicReference;
 import javax.annotation.Nullable;
 import org.apache.beam.runners.dataflow.options.DataflowPipelineOptions;
 import org.apache.beam.runners.dataflow.util.MonitoringUtil;
+import org.apache.beam.runners.dataflow.util.MonitoringUtil.JobMessagesHandler;
 import org.apache.beam.sdk.PipelineResult;
 import org.apache.beam.sdk.extensions.gcp.util.BackOffAdapter;
 import org.apache.beam.sdk.metrics.MetricResults;
@@ -53,6 +54,7 @@ import org.slf4j.LoggerFactory;
 
 /** A DataflowPipelineJob represents a job submitted to Dataflow using {@link DataflowRunner}. */
 public class DataflowPipelineJob implements PipelineResult {
+
   private static final Logger LOG = LoggerFactory.getLogger(DataflowPipelineJob.class);
 
   /** The id for the job. */
@@ -92,6 +94,8 @@ public class DataflowPipelineJob implements PipelineResult {
 
   static final Duration STATUS_POLLING_INTERVAL = Duration.standardSeconds(2);
 
+  static final Duration DEFAULT_MAX_BACKOFF = Duration.standardMinutes(2);
+
   static final double DEFAULT_BACKOFF_EXPONENT = 1.5;
 
   /** The amount of polling retries for job status and messages information. */
@@ -103,7 +107,9 @@ public class DataflowPipelineJob implements PipelineResult {
       FluentBackoff.DEFAULT
           .withInitialBackoff(MESSAGES_POLLING_INTERVAL)
           .withMaxRetries(MESSAGES_POLLING_RETRIES)
-          .withExponent(DEFAULT_BACKOFF_EXPONENT);
+          .withExponent(DEFAULT_BACKOFF_EXPONENT)
+          .withMaxBackoff(DEFAULT_MAX_BACKOFF);
+
   protected static final FluentBackoff STATUS_BACKOFF_FACTORY =
       FluentBackoff.DEFAULT
           .withInitialBackoff(STATUS_POLLING_INTERVAL)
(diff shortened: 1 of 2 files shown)


Ardagan:
@kennknowles @akedin

Ardagan:
@amaliujia

amaliujia:
I am not familiar with this piece of code. Maybe also ask in dev@ to see who is also able to review this change?

Ardagan:
run java postcommit

runners/google-cloud-dataflow-java/src/main/java/org/apache/beam/runners/dataflow/DataflowPipelineJob.java
```diff
@@ -261,96 +275,123 @@ State waitUntilFinish(
...
-          if (remaining.isLongerThan(Duration.ZERO)) {
-            backoff =
-                BackOffAdapter.toGcpBackOff(
-                    MESSAGES_BACKOFF_FACTORY.withMaxCumulativeBackoff(remaining).backoff());
-          } else {
-            // If there is no time remaining, don't bother backing off.
-            backoff = BackOff.STOP_BACKOFF;
-          }
+      // We can stop if the job is done.
+      if (state.isTerminal()) {
```
akedin: I suggest refactoring this a bit further. Right now it's hard to find the body of the `if`:

```
if (state.isTerminal()) {
  logTerminalState(state);
  return state;
}
```

runners/google-cloud-dataflow-java/src/main/java/org/apache/beam/runners/dataflow/DataflowPipelineJob.java
```diff
@@ -261,96 +275,123 @@ State waitUntilFinish(
...
-                      getReplacedByJob().getJobId()));
-              break;
-            default:
-              LOG.info("Job {} failed with status {}.", getJobId(), state);
-          }
-          return state;
-        }
+      exception = processJobMessages(messageHandler, monitor);
+
+      if (exception != null) {
```
akedin: previously we would reset backoff before `continue`, is it the right thing to not do this anymore?

Ardagan: Previous code is actually too layered and that's why I did refactored it a lot. backoff.reset() is under if(!hasError) condition.

We did reset backoff only in case if there was no error, ie status is not UNKNOWN. That caused us to fail upon reaching max retry count while receiving UNKNOWN state, not upon reaching timeout.

My change separates two cases:
a) If we fail to get status, we do not reset backoff and will fail due to exceptions upon attempt to get job status
b) If we actually receive UNKNOWN status, we will wait until timeout.

runners/google-cloud-dataflow-java/src/main/java/org/apache/beam/runners/dataflow/DataflowPipelineJob.java
Ardagan: Yeah, was thinking of that, but kept it here since it was one of the main parts of this function as I seen it.
Will refactor it out.

Ardagan:
UPD:
Confirmed that State.UNKNOWN is not supposed to be terminal on API side.

runners/google-cloud-dataflow-java/src/main/java/org/apache/beam/runners/dataflow/DataflowPipelineJob.java
akedin: Let me try to summarize the main flow to see if I understand it correctly:

**Previous Flow**
* get job state:
  * get non-`UNKNOWN` state -> **reset backoff** -> continue loop if not terminal;
     * will timeout at max duration or get a terminal state; correct behavior;
  * get `IOException`, same as:
  * get `UNKNOWN` state -> continue loop unconditionally;
     * does not **not** reset backoff;
     * can exceed number of allowed attempts fast, not waiting for max allowed duration;

**New Flow**
* get job state:
  * get non-`UNKNOWN` or `UNKNOWN` state -> **reset backoff** -> continue loop if not terminal;
    * can only receive `UNKNOWN` explicitly;
    * will timeout at max duration or get a terminal state; correct behavior;
  * get `IOException` -> continue loop unconditionally:
    * can exceed number of attempts instead of waiting for max allowed time;

akedin: In this case the logic seems right. I would probably try to organize the body of the loop to emphasize the flow though, something along the lines of:

```
Optional<State> state = tryGetState(); 
if (!state.isPresent() || !tryProcessJobMessages()) {
  continue;
}

if (state.get().isTerminal()) {
  return state.get();
}

resetAttemptsCount();
```

Hope this makes sense

Ardagan: You are right and your code example makes sense. The problem that didn't let me to get that state is that I want to propagate exception to outside of the loop. Unfortunately, java can not pass method arguments by reference, so there's no clear way to return state or exception except defining explicit class. And that would be a bit of an overkill in this case.

Ardagan:
run java postcommit

Ardagan:
Run Dataflow ValidatesRunner

akedin:
run java postcommit

Ardagan:
Run Dataflow ValidatesRunner

amaliujia:
Is there a reason that this PR's commits were not squashed?

akedin:
I squashed it when merging

amaliujia:
I see. Squash and merge will not update this PR directly.

PR #8636, opened 2019-05-21
[BEAM-7368] Flink + Python + gbk load test

lgajowy:
This adds jenkins jobs that run python gbk load test on flink. This is only one load test now but run periodicaly and on demand (phrase triggering). 

@agoenka could you take a look? 

cc: @pabloem

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/.test-infra/dataproc/create_flink_cluster.sh
+++ b/.test-infra/dataproc/create_flink_cluster.sh
@@ -28,6 +28,7 @@
 #    ARTIFACTS_DIR: Url to bucket where artifacts will be stored for staging (optional)
 #    FLINK_DOWNLOAD_URL: Url to Flink .tar archive to be installed on the cluster
 #    FLINK_NUM_WORKERS: Number of Flink workers
+#    FLINK_TASKMANAGER_SLOTS: Number of slots per Flink task manager
 #    DETACHED_MODE: Detached mode: should the SSH tunnel run in detached mode?
 #
 #    Example usage:
@@ -38,6 +39,7 @@
 #    ARTIFACTS_DIR=gs://<bucket-for-artifacts> \
 #    FLINK_DOWNLOAD_URL=http://archive.apache.org/dist/flink/flink-1.7.0/flink-1.7.0-bin-hadoop28-scala_2.12.tgz \
 #    FLINK_NUM_WORKERS=2 \
+#    FLINK_TASKMANAGER_SLOTS=1 \
 #    DETACHED_MODE=false \
 #    ./create_flink_cluster.sh
 #
@@ -57,6 +59,10 @@ DOCKER_INIT="$GCS_BUCKET/$INIT_ACTIONS_FOLDER_NAME/docker.sh"
 
 # Flink properties
 FLINK_LOCAL_PORT=8081
+
+# By default each taskmanager has one slot - use that value to avoid sharing SDK Harness by multiple tasks.
+FLINK_TASKMANAGER_SLOTS="${FLINK_TASKMANAGER_SLOTS:=1}"
+
 TASK_MANAGER_MEM=10240
 YARN_APPLICATION_MASTER=""
 
@@ -114,7 +120,8 @@ function start_tunnel() {
 
 function create_cluster() {
   local metadata="flink-snapshot-url=${FLINK_DOWNLOAD_URL},"
-  metadata+="flink-start-yarn-session=true"
+  metadata+="flink-start-yarn-session=true,"
+  metadata+="flink-taskmanager-slots=${FLINK_TASKMANAGER_SLOTS}"
 
   [[ -n "${HARNESS_IMAGES_TO_PULL:=}" ]] && metadata+=",beam-sdk-harness-images-to-pull=${HARNESS_IMAGES_TO_PULL}"
   [[ -n "${JOB_SERVER_IMAGE:=}" ]] && metadata+=",beam-job-server-image=${JOB_SERVER_IMAGE}"
--- a/.test-infra/dataproc/create_flink_cluster.sh
+++ b/.test-infra/dataproc/create_flink_cluster.sh
@@ -129,9 +129,12 @@ function create_cluster() {
   local image_version=$DATAPROC_VERSION
   echo "Starting dataproc cluster. Dataproc version: $image_version"
 
+  # Create one extra Dataproc VM for Flink's Job Manager
+  local num_dataproc_workers="$(($FLINK_NUM_WORKERS + 1))"
+
   # Docker init action restarts yarn so we need to start yarn session after this restart happens.
   # This is why flink init action is invoked last.
-  gcloud dataproc clusters create $CLUSTER_NAME --num-workers=$FLINK_NUM_WORKERS --initialization-actions $DOCKER_INIT,$BEAM_INIT,$FLINK_INIT --metadata "${metadata}", --image-version=$image_version --zone=$GCLOUD_ZONE --quiet
+  gcloud dataproc clusters create $CLUSTER_NAME --num-workers=$num_dataproc_workers --initialization-actions $DOCKER_INIT,$BEAM_INIT,$FLINK_INIT --metadata "${metadata}", --image-version=$image_version --zone=$GCLOUD_ZONE --quiet
 }
 
 function main() {
--- a/.test-infra/dataproc/init-actions/flink.sh
+++ b/.test-infra/dataproc/init-actions/flink.sh
@@ -56,6 +56,8 @@ readonly START_FLINK_YARN_SESSION_DEFAULT=true
 # Set this to install flink from a snapshot URL instead of apt
 readonly FLINK_SNAPSHOT_URL_METADATA_KEY='flink-snapshot-url'
 
+# Set this to define how many task slots are there per flink task manager
+readonly FLINK_TASKMANAGER_SLOTS_METADATA_KEY='flink-taskmanager-slots'
 
 
 function err() {
@@ -121,7 +123,17 @@ function configure_flink() {
     grep 'spark\.executor\.cores' /etc/spark/conf/spark-defaults.conf \
       | tail -n1 \
       | cut -d'=' -f2)
-  local flink_taskmanager_slots="$(($spark_executor_cores * 2))"
+
+  local slots="$(/usr/share/google/get_metadata_value \
+    "attributes/${START_FLINK_YARN_SESSION_METADATA_KEY}" \
+    || echo "${START_FLINK_YARN_SESSION_DEFAULT}")"
+
+  local flink_taskmanager_slots_default="$(($spark_executor_cores * 2))"
+
+  # if provided, use user defined number of slots.
+  local flink_taskmanager_slots="$(/usr/share/google/get_metadata_value \
+    "attributes/${FLINK_TASKMANAGER_SLOTS_METADATA_KEY}" \
+    || echo "${flink_taskmanager_slots_default}")"
 
   # Determine the default parallelism.
   local flink_parallelism=$(python -c \
--- a/.test-infra/dataproc/init-actions/flink.sh
+++ b/.test-infra/dataproc/init-actions/flink.sh
@@ -113,22 +113,12 @@ function configure_flink() {
   # NB: This assumes > 1 worker node.
   local num_taskmanagers="$(($num_workers - 1))"
 
-  # Determine the number of task slots per worker.
-  # TODO: Dataproc does not currently set the number of worker cores on the
-  # master node. However, the spark configuration sets the number of executors
-  # to be half the number of CPU cores per worker. We use this value to
-  # determine the number of worker cores. Fix this hack when
-  # yarn.nodemanager.resource.cpu-vcores is correctly populated.
-  local spark_executor_cores=$(\
-    grep 'spark\.executor\.cores' /etc/spark/conf/spark-defaults.conf \
-      | tail -n1 \
-      | cut -d'=' -f2)
+  local num_cores="$(grep -c processor /proc/cpuinfo)"
+  local flink_taskmanager_slots_default=$num_cores
 
   local slots="$(/usr/share/google/get_metadata_value \
     "attributes/${START_FLINK_YARN_SESSION_METADATA_KEY}" \
     || echo "${START_FLINK_YARN_SESSION_DEFAULT}")"
-
-  local flink_taskmanager_slots_default="$(($spark_executor_cores * 2))"
 
   # if provided, use user defined number of slots.
   local flink_taskmanager_slots="$(/usr/share/google/get_metadata_value \
--- a/.test-infra/jenkins/CommonTestProperties.groovy
+++ b/.test-infra/jenkins/CommonTestProperties.groovy
@@ -27,8 +27,9 @@ class CommonTestProperties {
     enum Runner {
         DATAFLOW("DataflowRunner"),
         SPARK("SparkRunner"),
-        FLINK("TestFlinkRunner"),
+        FLINK("FlinkRunner"),
         DIRECT("DirectRunner"),
+        PORTABLE("PortableRunner")
 
         def RUNNER_DEPENDENCY_MAP = [
                 JAVA: [
@@ -40,6 +41,7 @@ class CommonTestProperties {
                 PYTHON: [
                         DATAFLOW: "TestDataflowRunner",
                         DIRECT: "DirectRunner",
+                        PORTABLE: "PortableRunner"
                 ]
         ]
 
--- a/.test-infra/jenkins/CommonTestProperties.groovy
+++ b/.test-infra/jenkins/CommonTestProperties.groovy
@@ -27,7 +27,7 @@ class CommonTestProperties {
     enum Runner {
         DATAFLOW("DataflowRunner"),
         SPARK("SparkRunner"),
-        FLINK("FlinkRunner"),
+        FLINK("TestFlinkRunner"),
         DIRECT("DirectRunner"),
         PORTABLE("PortableRunner")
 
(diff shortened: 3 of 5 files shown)


lgajowy:
Run seed job

lgajowy:
Run Load Tests Python GBK Flink Batch

lgajowy:
Run seed job

lgajowy:
Run Load Tests Python GBK Flink Batch

lgajowy:
Run seed job

lgajowy:
Run Load Tests Python GBK Flink Batch

lgajowy:
Run seed job

lgajowy:
Run Load Tests Python GBK Flink Batch

lgajowy:
Run seed job

lgajowy:
Run Load Tests Python GBK Flink Batch

lgajowy:
Run seed job

lgajowy:
Run Load Tests Python GBK Flink Batch

lgajowy:
Run seed job

lgajowy:
Run Load Tests Python GBK Flink Batch

lgajowy:
Run seed job

lgajowy:
Run Load Tests Python GBK Flink Batch

.test-infra/jenkins/Infrastructure.groovy
```diff
@@ -0,0 +1,109 @@
...
+        common.setGradleSwitches(delegate)
+        tasks(":beam-runners-flink_2.11-job-server-container:docker")
+        switches("-Pdocker-repository-root=${repositoryRoot}")
+        switches("-Pdocker-tag=${dockerTag}")
+      }
+
+      shell("echo \" Tagging Flink Job Server's image\"...")
+      shell("docker tag ${image} ${imageTag}")
+      shell("echo \" Configure docker credentials\"...")
+      shell("docker-credential-gcr configure-docker")
```
lgajowy: @yifanzou @tvalentyn since yesterday you seem to have some more background on this ( 😃): is this safe to use here (won't this affect other jobs and credentials configuration)?

lgajowy:
@angoenka pinging kindly :) 

Adding @tvalentyn @yifanzou for additional checks if I didn't introduce changes breaking other jobs credentials.

.test-infra/jenkins/Infrastructure.groovy
```diff
@@ -0,0 +1,109 @@
...
+    String clusterName = getClusterName(clusterNamePrefix)
+    String artifactsDir="${gcsBucket}/${clusterName}"
+
+    context.steps {
+      environmentVariables {
+        env("GCLOUD_ZONE", "us-central1-a")
+        env("CLUSTER_NAME", clusterName)
+        env("GCS_BUCKET", gcsBucket)
+        env("FLINK_DOWNLOAD_URL", 'http://archive.apache.org/dist/flink/flink-1.5.6/flink-1.5.6-bin-hadoop28-scala_2.11.tgz')
+        env("FLINK_NUM_WORKERS", workerCount)
```
aaltay: Should we use a later version of flink? Does Beam support 1.8? Otherwise 1.7?

.test-infra/jenkins/Infrastructure.groovy
yifanzou: It should be fine to use this command. This will not affect other jobs base on our experiments.

The problem was that I've run the 'gcloud auth configure-docker' on the build workers while installing the GCR credential helper. It wrote an additional incompatible creds pair "marketplace.gcr.io": "gcloud" into the docker config. I reset the config files on build workers so that all affected tests were executed successfully.

yifanzou: It's only required to do once. It's already setup on the workers, so you don't have to run this every time.

.test-infra/jenkins/Infrastructure.groovy
```diff
@@ -0,0 +1,109 @@
...
+        rootBuildScriptDir(common.checkoutDir)
+        common.setGradleSwitches(delegate)
+        tasks(":beam-sdks-${sdkName}-container:docker")
+        switches("-Pdocker-repository-root=${repositoryRoot}")
+        switches("-Pdocker-tag=${dockerTag}")
+      }
+      shell("echo \" Tagging Harness' image\"...")
+      shell("docker tag ${image} ${imageTag}")
+      shell("echo \" Configure docker credentials\"...")
+      shell("docker-credential-gcr configure-docker")
```
yifanzou: same as comments below.

.test-infra/jenkins/Infrastructure.groovy
tvalentyn: +1 to @yifanzou's comments.

lgajowy: Great, thanks for clarifying!

.test-infra/jenkins/Infrastructure.groovy
lgajowy: Good catch - I should have used at least 1.7 here. It's the highest version I was able to use here. I know that there's 1.8 support in beam too but this caused problems - since flink 1.8 flink is packaged without hadoop28 and I cannot setup a cluster on dataproc due to that. 

I created JIRA addressing problems with 1.8: https://issues.apache.org/jira/browse/BEAM-7440

.test-infra/jenkins/Infrastructure.groovy
lgajowy: ok

lgajowy:
Run seed job

lgajowy:
Run Load Tests Python GBK Flink Batch

lgajowy:
Run Load Tests Python GBK Flink Batch

lgajowy:
Run seed job

lgajowy:
Run Load Tests Python GBK Flink Batch

lgajowy:
There's a caveat I've just noticed - the cluster won't get deleted when some step fails before "teardownDataproc". This, in turn, will cause leaving Dataproc clusters not deleted (only in case of failures, but still - this surely can happen). 

A possible candidate to solve this problem: https://jenkinsci.github.io/job-dsl-plugin/#method/javaposse.jobdsl.dsl.helpers.publisher.MavenPublisherContext.postBuildScripts

I have to fix this before merging the PR. I will post an update shortly.

.test-infra/jenkins/CommonTestProperties.groovy
```diff
@@ -27,8 +27,9 @@ class CommonTestProperties {
     enum Runner {
         DATAFLOW("DataflowRunner"),
         SPARK("SparkRunner"),
-        FLINK("TestFlinkRunner"),
+        FLINK("FlinkRunner"),
```
mxm: You might want to use `TestFlinkRunner` which has the advantage that pipelines shut down for finite sources.

.test-infra/jenkins/Infrastructure.groovy
mxm: Support for Flink 1.8 has been added recently (will be part of 2.13.0). Yes, you will have to include the Hadoop jar in the classpath, or add a Gradle dependency on Hadoop for your tests.

.test-infra/jenkins/CommonTestProperties.groovy
lgajowy: ok

lgajowy:
Run seed job

.test-infra/jenkins/Infrastructure.groovy
```diff
@@ -0,0 +1,111 @@
...
+      }
+
+      shell('echo Setting up flink cluster')
+      shell("cd ${common.makePathAbsolute('src/.test-infra/dataproc/')}; ./create_flink_cluster.sh")
+    }
+  }
+
+  static void teardownDataproc(def context, String jobName) {
+    context.publishers {
+      postBuildScripts {
```
lgajowy: This requires a plugin - I created an issue for this: https://issues.apache.org/jira/browse/BEAM-7457 

The goal here is to cleanup the cluster regardless of the test result.

lgajowy:
I implemented the postBuildScript (mentioned above) deleting the Dataproc cluster to be run regardless of the job result. It requires a plugin though - I created an issue for this: https://issues.apache.org/jira/browse/BEAM-7457

https://jenkinsci.github.io/job-dsl-plugin/#method/javaposse.jobdsl.dsl.helpers.publisher.MavenPublisherContext.postBuildScripts

.test-infra/jenkins/Infrastructure.groovy
```diff
@@ -0,0 +1,111 @@
...
+
+class Infrastructure {
+
+  static void prepareSDKHarness(def context, SDK sdk, String repositoryRoot, String dockerTag) {
+    context.steps {
+      String sdkName = sdk.name().toLowerCase()
+      String image = "${repositoryRoot}/${sdkName}"
+      String imageTag = "${image}:${dockerTag}"
+
+      shell("echo \"Building SDK harness for ${sdkName} SDK.\"")
```
angoenka: We can use "print" here.

.test-infra/jenkins/Infrastructure.groovy
```diff
@@ -0,0 +1,111 @@
...
+      gradle {
+        rootBuildScriptDir(common.checkoutDir)
+        common.setGradleSwitches(delegate)
+        tasks(":beam-sdks-${sdkName}-container:docker")
+        switches("-Pdocker-repository-root=${repositoryRoot}")
+        switches("-Pdocker-tag=${dockerTag}")
+      }
+      shell("echo \"Tagging Harness' image\"...")
```
angoenka: We can use "print" here.

.test-infra/jenkins/Infrastructure.groovy
angoenka: Shall we parameterize it for later flink updates and multiversion tests.

.test-infra/jenkins/job_LoadTests_GBK_Flink_Python.groovy
```diff
@@ -0,0 +1,80 @@
...
+                        environment_type: 'DOCKER'
+
+                ]
+        ]
+
+def loadTest = { scope, triggeringContext ->
+  scope.description('Runs Java GBK load tests on Flink runner in batch mode')
+  commonJobProperties.setTopLevelMainJobProperties(scope, 'master', 240)
+
+  infra.prepareSDKHarness(scope, testConfiguration.sdk, dockerRegistryRoot, 'latest')
```
angoenka: Just a Note: With cross language pipelines, we will need multiple sdk builds.
We don't have to do anything here.

.test-infra/jenkins/job_LoadTests_GBK_Flink_Python.groovy
```diff
@@ -0,0 +1,80 @@
...
+                jobProperties: [
+                        job_name            : "load_tests_Python_Flink_Batch_GBK_1_${now}",
+                        publish_to_big_query: false,
+                        project             : 'apache-beam-testing',
+                        metrics_dataset     : 'load_test',
+                        metrics_table       : "python_flink_batch_GBK_1",
+                        input_options       : '\'{"num_records": 200000000,"key_size": 1,"value_size":9}\'',
+                        iterations          : 1,
+                        fanout              : 1,
+                        parallelism         : 5,
```
angoenka: The cluster should have equal or more Tasks Slots (Slots per TaskManager(Worker) * Number of TaskManagers(Worker)).
Also, because of the way SDKHarness are shared in Flink taskmanager, I would recommend having 1 TaskSlot per worker an change numWorker to match parallelism in the test.

lgajowy:
Run seed job

.test-infra/jenkins/Infrastructure.groovy
lgajowy: Even if I do, what's described [here](https://stackoverflow.com/questions/7742472/groovy-script-in-jenkins-println-output-disappears-when-called-inside-class-envi/7754410) to allow using `print` / `println` from groovy in helper classes,  we will still get the messages printed in the Seed Job during the job definition creation/update. The goal is to have the line printed during the job execution - this is why it uses "shell" step.

.test-infra/jenkins/job_LoadTests_GBK_Flink_Python.groovy
lgajowy: Yes - in the future we can simply use this method multiple times to prepare any images we will need in the cluster for cross lang pipelines. :)

.test-infra/jenkins/job_LoadTests_GBK_Flink_Python.groovy
lgajowy: Could you explain a little bit more on how the SdkHarness are shared in the task manager? the problem I see is that multiple slots located on the same virtual machine can require SDK Harnesses at the same time - this will cause endpoint collisions

lgajowy: Could you elaborate a little bit more on how the SdkHarness are shared in the task manager? Is the problem that multiple slots located on the same virtual machine can require SDK Harnesses at the same time?

.test-infra/jenkins/Infrastructure.groovy
lgajowy: Parametrization is indeed a low hanging fruit - I'll fix this

lgajowy:
@angoenka Thanks for the review. I posted some questions.

.test-infra/jenkins/job_LoadTests_GBK_Flink_Python.groovy
angoenka: The SDKHarness across multiple tasks for the same job. So if a TaskManager has multiple slots and get multiple task for the same job then it will share the same SDKHarness.

angoenka:
@lgajowy Added my comments. Please let me know if you need more info.

lgajowy:
Run seed job

lgajowy:
Run seed job

lgajowy:
Run Load Tests Python GBK Flink Batch

lgajowy:
@angoenka thanks - all clear now. I posted the fixes (sorry for the delay). Could you take a look again?

lgajowy:
Oops - forgot to increase the number of workers from 3 to 5 and parallelism was 5. At least we know for sure that postBuildScripts cleanup the Dataproc cluster. Fixing.

lgajowy:
Run seed job

lgajowy:
Run Load Tests Python GBK Flink Batch

lgajowy:
Same error. Sorry for calling you to early @angoenka. I'll investigate this and keep you posted.

.test-infra/dataproc/create_flink_cluster.sh
```diff
@@ -57,6 +59,10 @@ DOCKER_INIT="$GCS_BUCKET/$INIT_ACTIONS_FOLDER_NAME/docker.sh"
 
 # Flink properties
 FLINK_LOCAL_PORT=8081
+
+# By default each taskmanager has one slot - use that value to avoid sharing SDK Harness by multiple tasks.
+FLINK_TASKMANAGER_SLOTS="${FLINK_TASKMANAGER_SLOTS:=1}"
```
lgajowy: @angoenka @mxm could you take a look at the error here: https://builds.apache.org/job/beam_LoadTests_Python_GBK_Flink_Batch_PR/14/console

According to my understanding, we have now 5 workers, each one having 1 slot. The error states that there is to few resources to run. Did I miss something?

lgajowy: More importantly, how can I determine how many workers with one slots should I have to succeed? Or maybe (for some reason) we cannot have 1 slot per worker? 🤔

mxm: What parallelism do you use? Generally, the total number of task slots needs to match the parallelism.

.test-infra/dataproc/init-actions/flink.sh
```diff
@@ -121,7 +123,17 @@ function configure_flink() {
     grep 'spark\.executor\.cores' /etc/spark/conf/spark-defaults.conf \
       | tail -n1 \
       | cut -d'=' -f2)
-  local flink_taskmanager_slots="$(($spark_executor_cores * 2))"
```
mxm: Could we duplicate the spark code here? Relying on an external config file is a bit strange.

.test-infra/dataproc/create_flink_cluster.sh
lgajowy: parralelism = 5, there are 5 Dataproc workers but I just found out that the flink.sh init action creates 4 task_managers. I think this is the culprit. I will take a look later.

.test-infra/dataproc/init-actions/flink.sh
lgajowy: Not sure - this is taken from the original init action - I will investigate that too.

.test-infra/dataproc/create_flink_cluster.sh
lgajowy: suspected line I'm talking about: https://github.com/apache/beam/blob/5ee4cf4e4880782492ec26f2b454a6df9b25f1e2/.test-infra/dataproc/init-actions/flink.sh#L112

lgajowy:
Run seed job

lgajowy:
Run Load Tests Python GBK Flink Batch

.test-infra/dataproc/init-actions/flink.sh
lgajowy: I "paraphrased" the default behavior of Dataproc while it was generating the spark-default.conf file. LMK if it's not too cryptic. 

Another approach we could have here is to simply set 1 slot by default and provide slot number through metadata otherwise. Not cryptic and explicit. 

Which approach feels better to you?

lgajowy:
Run seed job

lgajowy:
Run Load Tests Python GBK Flink Batch

lgajowy:
@mxm @angoenka I provided fixes. Could you take a look?

mxm:
Thanks @lgajowy. Looks good!

PR #8638, opened 2019-05-21
[release-2.13.0][BEAM-7230] Make PoolableDataSourceProvider a static singleton

iemejia:
This is a cherry pick to fix an instantiation issue that produces too many open connections on JdbcIO

R: @angoenka 
CC: @jbonofre @aromanenko-dev


jbonofre:
I will merge once Jenkins is happy.

iemejia:
Merged eagerly so it gets in once the vote starts. Thanks again JB!

PR #8639, opened 2019-05-21
[BEAM-7084] Allow custom environments for external transforms

mxm:
This adds an option to Python's ExternalTransform to specify the environment
alongside with its configuration. This allows users to use other environments
than the default Docker Java environment.

R @robertwb 
CC @tweise @angoenka 

--- | --- | --- | --- | ---

### DIFF
--- a/model/job-management/src/main/proto/beam_expansion_api.proto
+++ b/model/job-management/src/main/proto/beam_expansion_api.proto
@@ -46,6 +46,9 @@ message ExpansionRequest {
   // A namespace (prefix) to use for the id of any newly created
   // components.
   string namespace = 3;
+
+  // The type of environment to use for the expanded transform.
+  org.apache.beam.model.pipeline.v1.Environment environment = 4;
 }
 
 message ExpansionResponse {
--- a/sdks/python/apache_beam/runners/portability/flink_runner_test.py
+++ b/sdks/python/apache_beam/runners/portability/flink_runner_test.py
@@ -77,7 +77,7 @@ if __name__ == '__main__':
     _use_subprocesses = True
 
     conf_dir = None
-    expansion_port = None
+    expansion_service = None
 
     @classmethod
     def tearDownClass(cls):
@@ -116,7 +116,8 @@ if __name__ == '__main__':
       tmp_dir = mkdtemp(prefix='flinktest')
 
       cls._create_conf_dir()
-      cls.expansion_port = expansion_port
+      # TODO Move expansion address into PipelineOptions
+      cls._expansion_service = 'localhost:' + str(expansion_port)
 
       try:
         return [
@@ -161,15 +162,15 @@ if __name__ == '__main__':
       raise unittest.SkipTest("BEAM-4781")
 
     def test_external_transforms(self):
-      # TODO Move expansion address resides into PipelineOptions
-      def get_expansion_service():
-        return "localhost:" + str(self.expansion_port)
 
       with self.create_pipeline() as p:
         res = (
             p
             | GenerateSequence(start=1, stop=10,
-                               expansion_service=get_expansion_service()))
+                               expansion_service=
+                               FlinkRunnerTest._expansion_service,
+                               environment_type="EMBEDDED",
+                               environment_config=""))
 
         assert_that(res, equal_to([i for i in range(1, 10)]))
 
@@ -189,7 +190,8 @@ if __name__ == '__main__':
                            value_deserializer='org.apache.kafka.'
                                               'common.serialization.'
                                               'LongDeserializer',
-                           expansion_service=get_expansion_service()))
+                           expansion_service=
+                           FlinkRunnerTest._expansion_service))
       self.assertTrue('No resolvable bootstrap urls given in bootstrap.servers'
                       in str(ctx.exception),
                       'Expected to fail due to invalid bootstrap.servers, but '
@@ -209,7 +211,8 @@ if __name__ == '__main__':
                       value_serializer='org.apache.kafka.'
                                        'common.serialization.'
                                        'ByteArraySerializer',
-                      expansion_service=get_expansion_service()))
+                      expansion_service=
+                      FlinkRunnerTest._expansion_service))
 
     def test_flattened_side_input(self):
       # Blocked on support for transcoding
--- a/sdks/python/apache_beam/runners/portability/portable_runner.py
+++ b/sdks/python/apache_beam/runners/portability/portable_runner.py
@@ -95,31 +95,40 @@ class PortableRunner(runner.PipelineRunner):
   @staticmethod
   def _create_environment(options):
     portable_options = options.view_as(PortableOptions)
+    return PortableRunner._resolve_environment(
+        portable_options.environment_type,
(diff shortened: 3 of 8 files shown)


robertwb:
I don't think it makes sense for the caller to specify an environment for the expansion service to use... What we could do, however, is enumerate a set of known/preferred environments (enumerated in the existing components) which an expansion service could use if it deems them compatible.

mxm:
Where do we infer the known/preferred environments from? Existing environments in the Proto do not help us for external transforms because they will use a different environment. Do you want to add an extra message in the Proto?

I think specifying an environment would be the easiest way to use an external transform which does not belong to the standard Java environment. At least for now. Note that specifying an environment is optional.

robertwb:
Yes, we could add extra environments to the proto. 

It feels like this is working around the core issue that expansion services should be responsible for specifying their environments. If the non-standard environment is preferred, possibly this should be done as part of starting up the expansion service.

mxm:
I'm not convinced the environment type should be set for the life time of the expansion service. The expansion server is a potential long-running service and deciding upfront on the environment seems to lack flexibility. 

As of today, users can specify the main environment via the PipelineOptions, shouldn't this also include environments for external transforms? The expansion service could still override them if it can find a better environment.

tweise:
Being able to specify the default environment when starting the job service / expansion service may be a good option, and both are not exclusive. We already have a few job service parameters that serve similar purpose.

mxm:
I agree, setting up the environment up front on the Expansion Service is definitely an option. I think that an additional configuration of the preferred expansion environment via the pipeline options would be desirable for the user, who can already do that for the main environment of the pipeline.

robertwb:
There may be two problems that we are trying to solve here. 

(1) An expansion service that produces transforms tagged with the "vanilla" docker environment but relies on dependencies not in this environment. In my book, such an expansion service is simply broken and it's not the user's responsibility to fix it. 

(2) An expansion service vending transforms that can be part of a variety of environments (e.g. subprocess, embedded, or docker) which wants input into which one the user would prefer (or force?). It's a bit unclear how it would specify dependencies (artifacts?) in addition, or given an environment how it would detect that its transforms are compatible with it (absent a fixed list).

model/job-management/src/main/proto/beam_expansion_api.proto
```diff
@@ -46,6 +46,9 @@ message ExpansionRequest {
   // A namespace (prefix) to use for the id of any newly created
   // components.
   string namespace = 3;
+
+  // The type of environment to use for the expanded transform.
+  org.apache.beam.model.pipeline.v1.Environment environment = 4;
```
robertwb: It'd be better to refer to environments already in (or placed in) the components above, if just to facilitate sharing with existing transforms.

sdks/python/apache_beam/runners/portability/flink_runner_test.py
```diff
@@ -116,7 +116,8 @@ def _subprocess_command(cls, job_port, expansion_port):
       tmp_dir = mkdtemp(prefix='flinktest')
 
       cls._create_conf_dir()
-      cls.expansion_port = expansion_port
+      # TODO Move expansion address into PipelineOptions
```
robertwb: Not sure this makes sense as a pipeline option, there may be different services to use for different transforms.

sdks/python/apache_beam/runners/portability/portable_runner.py
```diff
@@ -95,31 +95,40 @@ def default_docker_image():
...
       environment_urn = common_urns.environments.EXTERNAL.urn
-    elif portable_options.environment_type:
-      if portable_options.environment_type.startswith('beam:env:'):
-        environment_urn = portable_options.environment_type
+    elif environment_type == 'EMBEDDED':
+      environment_urn = environment_type
+    elif environment_type:
+      # Allow embedded environment for testing
+      if environment_type.startswith('beam:env:') or \
+            environment_type == 'EMBEDDED':
```
robertwb: Here we seem to be embedding knowledge about Java's understanding about the special string "EMBEDDED" meaning "java embedded." There's also a python "embedded" that is separate from this. https://github.com/apache/beam/blob/release-2.13.0/sdks/python/apache_beam/portability/python_urns.py#L38 On the other hand, specifying "DOCKER" would give the Python docker image, right? Seems we should be doing something more principled here.

mxm: No, the environment_type is a URN. `Embedded` is a special testing URN that I didn't want to add as a standard urn. So this alongside with the payload works across all SDKs.

sdks/python/apache_beam/runners/portability/flink_runner_test.py
mxm: I copied over this TODO. We discussed a while ago that adding a registry in pipeline options for expansion services might make sense.

model/job-management/src/main/proto/beam_expansion_api.proto
mxm: Completely agree, makes sense to add the environments to components.

mxm:
>(1) An expansion service that produces transforms tagged with the "vanilla" docker environment but relies on dependencies not in this environment. In my book, such an expansion service is simply broken and it's not the user's responsibility to fix it.

What you describe is the current state of the expansion service which always uses the standard Docker environment. The idea of the PR was to give users the option to configure the environment. There are more sophisticated ways to determining the correct environment but I think they are out of scope here.

>(2) An expansion service vending transforms that can be part of a variety of environments (e.g. subprocess, embedded, or docker) which wants input into which one the user would prefer (or force?). It's a bit unclear how it would specify dependencies (artifacts?) in addition, or given an environment how it would detect that its transforms are compatible with it (absent a fixed list).

If a user has created a dedicated environment for the desired external transforms, I don't see a reason to prevent him/her from using it. What you are describing is a more advanced process which we will work out in the discussions around artifact resolution. We could include additional information about the environment which would be used by the Expansion Server to compare it against other environments. Again, the idea here was just to enable use of custom environments.

sdks/python/apache_beam/runners/portability/portable_runner.py
robertwb: It doesn't have to be a standard URN, but if we're spreading it across multiple codebases we should probably qualify it because it has a specific meaning. Maybe "beam:env:embedded_java_for_testing:v1" or something like that then. It'd break in strange ways if I tried to use this with a python expansion service, or on a non-Java runner. What you really want to say is "run this as cheaply as possible."

(Also, I still think it's inconsistent to have a "EMBEDDED" mean a java-runner-specific thing thing and "DOCKER" a python-image-spawning one and "PROCESS" something else altogether.)

mxm: Agree, we should give it a proper, private URN. Yes, reusing the existing Python code here has some problematic implications. We probably only want to accept the URN, or come up with a better way to specify that we are using the embedded environment of an SDK (e.g. `Embedded:Java`)

mxm:
Could be worth taking this to the mailing list. Just sent a mail.

PR #8641, opened 2019-05-21
[BEAM-7443] Create a BoundedSource -> SDF wrapper in Python SDK

boyuanzz:
Create a wrapper tp wrapping python BoundedSource into SDF in python SDK. 
Major components of wrapper included:
* SDFBoundedSourceWrapper
* SDFBoundedSourceRestrictionTracker
* SDFBoundedSourceDoFn

In dataflow_runner and fn_api_runner, if pipeline running with "use_sdf_bounded_source" option, the runner will override Read(BoundedSource) with SDFBoundedSourceWrapper(BoundedSource). No support for other runners for now.

R: @lukecwik @robertwb 
CC: @laraschmidt

### DIFF
--- a/sdks/python/apache_beam/io/iobase.py
+++ b/sdks/python/apache_beam/io/iobase.py
@@ -845,6 +845,16 @@ class Read(ptransform.PTransform):
     super(Read, self).__init__()
     self.source = source
 
+  @staticmethod
+  def get_desired_chunk_size(total_size):
+    total_size
+    if total_size:
+      # 1MB = 1 shard, 1GB = 32 shards, 1TB = 1000 shards, 1PB = 32k shards
+      chunk_size = max(1 << 20, 1000 * int(math.sqrt(total_size)))
+    else:
+      chunk_size = 64 << 20  # 64mb
+    return chunk_size
+
   def expand(self, pbegin):
     from apache_beam.options.pipeline_options import DebugOptions
     from apache_beam.transforms import util
@@ -857,13 +867,8 @@ class Read(ptransform.PTransform):
       source = self.source
 
       def split_source(unused_impulse):
-        total_size = source.estimate_size()
-        if total_size:
-          # 1MB = 1 shard, 1GB = 32 shards, 1TB = 1000 shards, 1PB = 32k shards
-          chunk_size = max(1 << 20, 1000 * int(math.sqrt(total_size)))
-        else:
-          chunk_size = 64 << 20  # 64mb
-        return source.split(chunk_size)
+        return source.split(
+            self.get_desired_chunk_size(self.source.estimate_size()))
 
       return (
           pbegin
@@ -1122,6 +1127,8 @@ class RestrictionTracker(object):
     Methods of the class ``RestrictionTracker`` including this method may get
     invoked by different threads, hence must be made thread-safe, e.g. by using
     a single lock object.
+
+    TODO(BEAM-7473): Remove thread safety requirements from API implementation.
     """
     raise NotImplementedError
 
@@ -1148,6 +1155,8 @@ class RestrictionTracker(object):
     Methods of the class ``RestrictionTracker`` including this method may get
     invoked by different threads, hence must be made thread-safe, e.g. by using
     a single lock object.
+
+    TODO(BEAM-7473): Remove thread safety requirements from API implementation.
     """
 
     raise NotImplementedError
@@ -1168,9 +1177,100 @@ class RestrictionTracker(object):
     invoked by different threads, hence must be made thread-safe, e.g. by using
     a single lock object.
 
+    TODO(BEAM-7473): Remove thread safety requirements from API implementation.
+
     Returns: ``True`` if current restriction has been fully processed.
     Raises:
       ~exceptions.ValueError: if there is still any unclaimed work remaining.
+    """
+    raise NotImplementedError
+
+  def try_split(self, fraction_of_remainder):
+    """Splits current restriction based on fraction_of_remainder.
+
+    If splitting the current restriction is possible, the current restriction is
+    split into a primary and residual restriction pair. This invocation updates
+    the ``current_restriction()`` to be the primary restriction effectively
+    having the current ``DoFn.process()`` execution responsible for performing
+    the work that the primary restriction represents. The residual restriction
+    will be executed in a separate ``DoFn.process()`` invocation (likely in a
+    different process). The work performed by executing the primary and residual
+    restrictions as separate ``DoFn.process()`` invocations MUST be equivalent
+    to the work performed as if this split never occurred.
+
+    The ``fraction_of_remainder`` should be used in a best effort manner to
+    choose a primary and residual restriction based upon the fraction of the
+    remaining work that the current ``DoFn.process()`` invocation is responsible
+    for. For example, if a ``DoFn.process()`` was reading a file with a
+    restriction representing the offset range [100, 200) and has processed up to
+    offset 130 with a fraction_of_remainder of 0.7, the primary and residual
+    restrictions returned would be [100, 179), [179, 200) (note: current_offset
+    + fraction_of_remainder * remaining_work = 130 + 0.7 * 70 = 179).
+
+    It is very important for pipeline scaling and end to end pipeline execution
+    that try_split is implemented well.
+
+    Args:
+      fraction_of_remainder: A hint as to the fraction of work the primary
+        restriction should represent based upon the current known remaining
+        amount of work.
+
+    Returns:
+      (primary_restriction, residual_restriction) if a split was possible,
+      otherwise returns ``None``.
+
+    ** Thread safety **
+
+    Methods of the class ``RestrictionTracker`` including this method may get
+    invoked by different threads, hence must be made thread-safe, e.g. by using
+    a single lock object.
+
+    TODO(BEAM-7473): Remove thread safety requirements from API implementation.
+    """
+    raise NotImplementedError
+
+  def try_claim(self, position):
+    """ Attempts to claim the block of work in the current restriction
+    identified by the given position.
+
+    If this succeeds, the DoFn MUST execute the entire block of work. If it
+    fails, the ``DoFn.process()`` MUST return ``None`` without performing any
+    additional work or emitting output (note that emitting output or performing
+    work from ``DoFn.process()`` is also not allowed before the first call of
+    this method).
+
+    Args:
+      position: current position that wants to be claimed.
+
+    Returns: ``True`` if the position can be claimed as current_position.
+    Otherwise, returns ``False``.
+
+    ** Thread safety **
+
+    Methods of the class ``RestrictionTracker`` including this method may get
+    invoked by different threads, hence must be made thread-safe, e.g. by using
+    a single lock object.
+
+    TODO(BEAM-7473): Remove thread safety requirements from API implementation.
+    """
+    raise NotImplementedError
+
+  def defer_remainder(self, watermark=None):
+    """ Invokes checkpoint() in an SDF.process().
+
+    TODO(BEAM-7472): Remove defer_remainder() once SDF.process() uses
+    ``ProcessContinuation``.
+
+    Args:
+      watermark
+    """
+    raise NotImplementedError
+
+  def deferred_status(self):
+    """ Returns deferred_residual with deferred_watermark.
+
+    TODO(BEAM-7472): Remove defer_status() once SDF.process() uses
+    ``ProcessContinuation``.
     """
     raise NotImplementedError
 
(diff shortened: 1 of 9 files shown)


sdks/python/apache_beam/io/iobase.py
```diff
@@ -855,24 +864,44 @@ def expand(self, pbegin):
...
-
-      return (
-          pbegin
-          | core.Impulse()
-          | 'Split' >> core.FlatMap(split_source)
-          | util.Reshuffle()
-          | 'ReadSplits' >> core.FlatMap(lambda split: split.source.read(
-              split.source.get_range_tracker(
-                  split.start_position, split.stop_position))))
+      # My original design was to have a SDFBoundedSourceWrapper PTransform, which will
```
lukecwik: Makes sense.

sdks/python/apache_beam/io/iobase.py
```diff
@@ -855,24 +864,44 @@ def expand(self, pbegin):
...
+      # replace Read() when constructing  a pipeline. We can also choose to expand Read
+      # direcly using SDF, which is simpler from my point of view.
+      if 'sdf_bounded_source' in debug_options.experiments:
+        from apache_beam.io.sdf_restriction_provider import SDFBoundedSourceRestrictionProvider
+
+        def create_sdf_bounded_source_dofn():
+          chunk_size = self._get_desired_chunk_size()
```
lukecwik: Currently this is being done before pipeline execution but we should move this logic to when the pipeline is executing.

sdks/python/apache_beam/io/sdf_restriction_provider.py
```diff
@@ -0,0 +1,73 @@
...
+    for source_bundle in source_bundles:
+      yield (source_bundle.start_position, source_bundle.stop_position)
+
+  def restriction_size(self, element, restriction):
+    # More precisely, this function returns the weight, rather the actual size.
+    return restriction[1] - restriction[0]
+
+
+
+class SDFBoundedSourceWrapper(ptransform.PTransform):
```
lukecwik: Can you use this PTransform instead of having a duplicate definition of the DoFn inside of iobase.py#876

sdks/python/apache_beam/io/iobase.py
```diff
@@ -910,6 +913,67 @@ def from_runner_api_parameter(parameter, context):
     Read.from_runner_api_parameter)
 
 
+class SDFBoundedSourceWrapper(ptransform.PTransform):
+  """A ``PTransform`` that uses SDF to read from a ``BoundedSource``.
+  NOTE: This transform can only be used with beam_fn_api enabled.
```
lukecwik: docstring format should be 
```
one liner.

additional details.
```

The blank line after the first line is important.

sdks/python/apache_beam/io/sdf_restriction_provider.py
```diff
@@ -0,0 +1,56 @@
...
+
+from __future__ import absolute_import
+
+from apache_beam.io.restriction_trackers import SDFBoundedSourceRestrictionTracker
+from apache_beam.transforms.core import RestrictionProvider
+
+__all__ = ['SDFBoundedSourceRestrictionProvider']
+
+
+class SDFBoundedSourceRestrictionProvider(RestrictionProvider):
```
lukecwik: Ditto on making this a class defined within SDFBoundedSourceWrapper and not expose it to users.

sdks/python/apache_beam/io/sdf_restriction_provider.py
```diff
@@ -0,0 +1,56 @@
...
+                                       restriction[1]))
+
+  def split(self, element, restriction):
+    # Invoke source.split to get initial splitting results.
+    source_bundles = self._source.split(self._desired_chunk_size)
+    for source_bundle in source_bundles:
+      self._restriction_size_map[(source_bundle.start_position,
+                                  source_bundle.stop_position)]\
+        = source_bundle.weight
+      yield (source_bundle.start_position, source_bundle.stop_position)
```
lukecwik: I don't believe you can assume that the source you have as the element equal to the source being returned within SourceBundle.

If you take a look at the file based source, you can see that the SourceBundle may return a ConcatSource or a _SingleFileSource.

I believe you'll want to make the "restriction" a SourceBundle and use the source from the SourceBundle.

The initial restriction could return the SourceBundle(None, source, None, None). Conveniently, this would allow you to return source_bundle.weight and fall back to source_bundle.source.estimate_size() if source_bundle.weight is None.

sdks/python/apache_beam/runners/dataflow/dataflow_runner.py
```diff
@@ -94,11 +94,16 @@ class DataflowRunner(PipelineRunner):
   # Imported here to avoid circular dependencies.
   # TODO: Remove the apache_beam.pipeline dependency in CreatePTransformOverride
   from apache_beam.runners.dataflow.ptransform_overrides import CreatePTransformOverride
+  from apache_beam.runners.dataflow.ptransform_overrides import ReadPTransformOverride
 
   _PTRANSFORM_OVERRIDES = [
       CreatePTransformOverride(),
   ]
 
+  SDF_PTRANSFORM_OVERRIDES = [
```
lukecwik:
```suggestion
  _SDF_PTRANSFORM_OVERRIDES = [
```

sdks/python/apache_beam/runners/dataflow/dataflow_runner.py
```diff
@@ -371,6 +376,8 @@ def run_pipeline(self, pipeline, options):
     # done before Runner API serialization, since the new proto needs to contain
     # any added PTransforms.
     pipeline.replace_all(DataflowRunner._PTRANSFORM_OVERRIDES)
+    if apiclient._use_sdf_bounded_source(options):
+      pipeline.replace_all(DataflowRunner.SDF_PTRANSFORM_OVERRIDES)
```
lukecwik:
```suggestion
      pipeline.replace_all(DataflowRunner._SDF_PTRANSFORM_OVERRIDES)
```

sdks/python/apache_beam/io/iobase.py
```diff
@@ -910,6 +913,67 @@ def from_runner_api_parameter(parameter, context):
     Read.from_runner_api_parameter)
 
 
+class SDFBoundedSourceWrapper(ptransform.PTransform):
```
lukecwik: I don't believe we want to make this class public. It should be an implementation detail of how a runner is performing the replacement with eventually the Read PTransform performing the replacement.

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
```diff
@@ -478,6 +479,22 @@ def process(
...
+      # Setup experiment option to enable using SDFBoundedSourceWrapper
+      if not 'use_sdf_bounded_source' in experiments:
+        experiments.append('use_sdf_bounded_source')
+
+      p._options.view_as(DebugOptions).experiments = experiments
+
+      actual = (
+          p | beam.io.Read(RangeSource(0, 4))
+      )
+      assert_that(actual, equal_to([0, 1, 2, 3]))
```
lukecwik: This won't verify that the PTransform replacement actually occurred since if the replacement didn't occur, you would still get the same pipeline result.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -273,6 +281,10 @@ def run_pipeline(self, pipeline, options):
         pipeline_options.DirectOptions).direct_runner_bundle_repeat
     self._profiler_factory = profiler.Profile.factory_from_options(
         options.view_as(pipeline_options.ProfilingOptions))
+
+    if experiments and 'use_sdf_bounded_source' in experiments:
```
lukecwik:
```suggestion
    if 'use_sdf_bounded_source' in experiments:
```

You already guarantee that experiments is not `None` when creating the variable

sdks/python/apache_beam/io/iobase.py
```diff
@@ -910,6 +913,67 @@ def from_runner_api_parameter(parameter, context):
...
+  """
+
+  def __init__(self, source):
+    if not isinstance(source, BoundedSource):
+      raise RuntimeError('SDFBoundedSourceWrapper can only wrap BoundedSource')
+    super(SDFBoundedSourceWrapper, self).__init__()
+    self.source = source
+
+  def _get_desired_chunk_size(self):
+    total_size = self.source.estimate_size()
```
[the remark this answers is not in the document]
boyuanzz: Thanks for helping review!  _get_desired_chunk_size could be a static function of Read(PTransform) as long as it's not possible that Read PTransform would be deprecated before SDFWrapper.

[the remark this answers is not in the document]
lukecwik: sgtm

sdks/python/apache_beam/io/sdf_restriction_provider.py
boyuanzz: I don't think I make this assumption here. Taking ```FileBaseSource``` as an example, ```FileBaseSource.split()``` returns SourceBundle constructed based on ```_SingleFileSource```:
https://github.com/apache/beam/blob/master/sdks/python/apache_beam/io/filebasedsource.py#L271 .

As for falling back to source_bundle.source.estimate_size(), actually I don't think it would work at least for ```_SingleFileSource```. If looking into constructing SourceBundle in split: https://github.com/apache/beam/blob/master/sdks/python/apache_beam/io/filebasedsource.py#L294, the stop_pos will be set to INF, which causes source_bundle.source.estimate_size()(https://github.com/apache/beam/blob/master/sdks/python/apache_beam/io/filebasedsource.py#L302) return inaccurate size.

lukecwik: The `SDFBoundedSourceRestrictionProvider` is initialized with the source object that is passed into `Read`. After splitting, the `SDFBoundedSourceRestrictionProvider` creates a tracker using that source from `Read` and not the source contained inside the `SourceBundle`.

Good point about not relying on source_bundle.source.estimate_size() and only using the weight if it is provided.

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
boyuanzz: Adding one more test case to verify overrides: https://github.com/apache/beam/pull/8641/commits/2392ae94101c71135466837547838d4b8bc1e090

sdks/python/apache_beam/io/iobase.py
```diff
@@ -1180,15 +1180,47 @@ def check_done(self):
     raise NotImplementedError
 
   def try_split(self, fraction_of_remainder):
+    """Splits current restriction based on fraction_of_remainder.
+
+    Invoked when SDK receiving ProcessBundleSplitRequest during processing
+    bundle.
+
+    Args:
+      fraction_of_remainder: a fraction of (cur_pos, stop_pos).
```
lukecwik: fraction_of_remainder: A hint as to the fraction of work the primary restriction should represent based upon the current known remaining amount of work.

sdks/python/apache_beam/io/iobase.py
```diff
@@ -1180,15 +1180,47 @@ def check_done(self):
...
+    ((start_pos, split_pos), (split_pos, stop_pos)).
+
+    ** Thread safety **
+    Accessing to position and checkpoint status should be guarded by a single
+    lock object.
+    """
     raise NotImplementedError
 
   def try_claim(self, position):
+    """ Claims position as current_position.
```
lukecwik: Try using the javadoc from https://github.com/apache/beam/blob/c0de4570b186793ff477f4a5bd86cc27191e2999/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/RestrictionTracker.java#L28 to write this version.

pabloem:
I'm just commenting on this to get mail updates:)

sdks/python/apache_beam/io/iobase.py
```diff
@@ -1182,37 +1182,69 @@ def check_done(self):
...
-    Accessing to position should be guarded by a single lock object.
+
+    Methods of the class ``RestrictionTracker`` including this method may get
+    invoked by different threads, hence must be made thread-safe, e.g. by using
+    a single lock object.
     """
     raise NotImplementedError
 
   def defer_remainder(self, watermark=None):
-    """ Invokes checkpoint() in an SDF.process()
```
lukecwik: I filed https://issues.apache.org/jira/browse/BEAM-7472

I would add this as a TODO and just mark these methods as deprecated.

lukecwik:
Please update the thread safety sections with a TODO pointing at BEAM-7473 to remove the thread safety requirements.

boyuanzz:
All comments have been addressed. Would you like to do another round of review? @lukecwik

sdks/python/apache_beam/io/iobase.py
```diff
@@ -1168,12 +1177,102 @@ def check_done(self):
...
+
+    Args:
+      watermark
+    """
+    raise NotImplementedError
+
+  def deferred_status(self):
+    """ Returns deferred_residual with deferred_watermark.
+
+    TODO(BEAM-7472): Mark defer_status as deprecated once SDF.process() uses
```
lukecwik:
```suggestion
    TODO(BEAM-7472): Remove deferred_status once SDF.process() uses
```

sdks/python/apache_beam/io/iobase.py
```diff
@@ -1168,12 +1177,102 @@ def check_done(self):
...
+    a single lock object.
+
+    TODO(BEAM-7473): Remove thread safety requirements from API implementation.
+    """
+    raise NotImplementedError
+
+  def defer_remainder(self, watermark=None):
+    """ Invokes checkpoint() in an SDF.process().
+
+    TODO(BEAM-7472): Mark defer_remainder as deprecated once SDF.process() uses
```
lukecwik:
```suggestion
    TODO(BEAM-7472): Remove defer_remainder once SDF.process() uses
```

boyuanzz:
All checks have passed :D @lukecwik

PR #8648, opened 2019-05-22
[BEAM-3934] Use try with resources to close the reader on JavaReadViaImpulse

iemejia:
R: @lukecwik

### DIFF
--- a/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/JavaReadViaImpulse.java
+++ b/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/JavaReadViaImpulse.java
@@ -132,11 +132,12 @@ public class JavaReadViaImpulse {
   @VisibleForTesting
   static class ReadFromBoundedSourceFn<T> extends DoFn<BoundedSource<T>, T> {
     @ProcessElement
-    public void readSoruce(ProcessContext ctxt) throws IOException {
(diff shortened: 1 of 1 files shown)


iemejia:

lukecwik:

iemejia:
It seems the class of the container used by "Run Python_PVR_Flink PreCommit" is not getting updated and this is breaking the tests. Any hint on why this is happening @robertwb ?

robertwb:
I have no idea. It should be taking these on as a dependency. Maybe it's pulling an old cache from somewhere?

iemejia:
Mmm I am worried to merge and then breaking all tests after, anyone who may know/fix this ?

iemejia:
@angoenka maybe?

angoenka:
I think the failure is caused by missmatch in class version of Flink Job Server and Extension service. 
Checking how we build the extension service

angoenka:
seems like we are not rebuilding the java container which is required now as we run java transforms while in python tests. 
Jira: https://issues.apache.org/jira/browse/BEAM-7422

angoenka:
However, on a side note. This highlights that the containers are shared across tests in a machine.
Their is a race condition when multiple tests runs on on machine building container and can be a cause of flakiness.

iemejia:

iemejia:
Thanks a lot @angoenka for creating the related issue and for the quick fix.
What could be done to address the possible flakiness issue?

PR #8649, opened 2019-05-22
[BEAM-7114] Move dot pipeline graph renderer to runners-core-construc…

iemejia:
Please tell me if you prefer a different module or package as a destination.

R: @kennknowles 
CC: @xinyuiscool


iemejia:
Thanks @xinyuiscool I will merge it once I get the tests passing.

PR #8659, opened 2019-05-23
Documenting and refactoring fn api runner code

pabloem:
WIP

### DIFF
--- a/sdks/python/apache_beam/runners/portability/fn_api_runner.py
+++ b/sdks/python/apache_beam/runners/portability/fn_api_runner.py
@@ -650,16 +650,14 @@ class FnApiRunner(runner.PipelineRunner):
                                 bundle_processor.DATA_OUTPUT_URN):
         pcoll_id = transform.spec.payload
         if transform.spec.urn == bundle_processor.DATA_INPUT_URN:
-          target = transform.unique_name, only_element(transform.outputs)
           if pcoll_id == fn_api_runner_transforms.IMPULSE_BUFFER:
-            data_input[target] = [ENCODED_IMPULSE_VALUE]
+            data_input[transform.unique_name] = [ENCODED_IMPULSE_VALUE]
           else:
-            data_input[target] = pcoll_buffers[pcoll_id]
+            data_input[transform.unique_name] = pcoll_buffers[pcoll_id]
           coder_id = pipeline_components.pcollections[
               only_element(transform.outputs.values())].coder_id
         elif transform.spec.urn == bundle_processor.DATA_OUTPUT_URN:
-          target = transform.unique_name, only_element(transform.inputs)
-          data_output[target] = pcoll_id
+          data_output[transform.unique_name] = pcoll_id
           coder_id = pipeline_components.pcollections[
               only_element(transform.inputs.values())].coder_id
         else:
--- a/sdks/python/apache_beam/runners/portability/fn_api_runner.py
+++ b/sdks/python/apache_beam/runners/portability/fn_api_runner.py
@@ -421,7 +421,6 @@ class FnApiRunner(runner.PipelineRunner):
     data_input, data_side_input, data_output = self._extract_endpoints(
         stage, pipeline_components, data_api_service_descriptor, pcoll_buffers)
 
-    # TODO(pabloem) Describe what is a process bundle descriptor
     process_bundle_descriptor = beam_fn_api_pb2.ProcessBundleDescriptor(
         id=self._next_uid(),
         transforms={transform.unique_name: transform
--- a/sdks/python/apache_beam/runners/portability/fn_api_runner.py
+++ b/sdks/python/apache_beam/runners/portability/fn_api_runner.py
@@ -83,6 +83,7 @@ ENCODED_IMPULSE_VALUE = beam.coders.WindowedValueCoder(
 
 
 class BeamFnControlServicer(beam_fn_api_pb2_grpc.BeamFnControlServicer):
+  """TODO"""
 
   UNSTARTED_STATE = 'unstarted'
   STARTED_STATE = 'started'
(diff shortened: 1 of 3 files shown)


pabloem:
Run Python PreCommit

pabloem:
r: @boyuanzz

sdks/python/apache_beam/runners/portability/fn_api_runner_transforms.py
```diff
@@ -1018,8 +1018,8 @@ def fuse(producer, consumer):
       for output in transform.outputs.values():
         producers_by_pcoll[output] = stage
 
-  logging.debug('consumers\n%s', consumers_by_pcoll)
-  logging.debug('producers\n%s', producers_by_pcoll)
+  #logging.debug('consumers\n%s', consumers_by_pcoll)
+  #logging.debug('producers\n%s', producers_by_pcoll)
```
pabloem: I commented these out because they are very noisy and don't provide much value. Someone debugging them can uncomment them, but otherwise, I think it makes sense to not have them.

robertwb: Logging.debug is fairly low level, but if it's too verbose for even that then go ahead and delete these lines rather than comment them out.

pabloem: What happens here is that when a test fails in `tox`, debug logging gets activated, and we currently log almost every proto related to portability, so it's a ton of logging, and it's very hard to read.

When running the tests with `python -m unittest module.TestClass.test_case`, this is not a problem. Debug logging does not get activated on failure, so it's easier to read through. 

It would be nice to have a middle ground - but I lean towards removing them until someone finds them useful in debugging, and we can putthem back. Thoughts?

boyuanzz: Vote for deleting them. If a dev needs these info, the dev can add these logs.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -396,48 +405,11 @@ def iterable_state_write(values, element_coder_impl):
...
-                create_buffer_id(transform.inputs[tag]), si.access_pattern)
-      return data_input, data_side_input, data_output
-
     logging.info('Running %s', stage.name)
-    logging.debug('       %s', stage)
-    data_input, data_side_input, data_output = extract_endpoints(stage)
+    data_input, data_side_input, data_output = self._extract_endpoints(
+        stage, pipeline_components, data_api_service_descriptor, pcoll_buffers)
 
+    # TODO(pabloem) Describe what is a process bundle descriptor
```
boyuanzz: The ProcessBundleDescriptor is defined here: https://github.com/apache/beam/blob/c85f0f51634e65fe332e8bbe18fbf0689eb44d0f/model/fn-execution/src/main/proto/beam_fn_api.proto#L149. Maybe we can link the proto definition here. Hope the comments of proto would be enough.

pabloem:
```suggestion
```

pabloem: I've removed the line.

sdks/python/apache_beam/runners/portability/fn_api_runner_transforms.py
pabloem: Done.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -838,9 +887,9 @@ def push(self, request):
     if not request.instruction_id:
       self._uid_counter += 1
       request.instruction_id = 'control_%s' % self._uid_counter
-    logging.debug('CONTROL REQUEST %s', request)
+    logging.debug('CONTROL REQUEST ID %s', request.instruction_id)
```
boyuanzz: When debugging fnapi request and response, the instruction id may not be enough for that purpose. Sometimes instruction type + instruction id would be more helpful. But perhaps we can delete these debug logs and let dev adds them if needed. wdyt?

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1220,11 +1288,14 @@ def process_bundle(self, inputs, expected_outputs):
     # Actually start the bundle.
     if registration_future and registration_future.get().error:
       raise RuntimeError(registration_future.get().error)
-    process_bundle = beam_fn_api_pb2.InstructionRequest(
+
+    # Runner requests the SDK to fulfill the following instruction:
```
boyuanzz: I'm not sure whether I understand this comment correctly, but from what I know so far, the runner sends ProcessBundleRequest to sdk and waiting for the response.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
pabloem: Done.

pabloem:
Run Python PreCommit

sdks/python/apache_beam/runners/portability/fn_api_runner.py
pabloem:
```suggestion
```

sdks/python/apache_beam/runners/portability/fn_api_runner_transforms.py
robertwb: Yes, delete. The value of these logs is low now that this stuff is basically working (and it'd be good to fix the tox issue, but we shouldn't wait for that.)

tvalentyn:
Thanks for helping with code clarity, Pablo.

PR #8665, opened 2019-05-23
Refine runners/reference/java dependencies and fix minor static analysis issues

iemejia:
R: @robertwb


iemejia:
Is the reference job server going to stay there, or are there plans to move it to other module?

robertwb:
The code changes look fine to me. 

I think it would make sense to move this to a runners/portable module. (Open to other ideas though.)

iemejia:
I think @angoenka said he was looking at this. I was just not sure if it was finally happening or it was just going to state in reference.

PR #8680, opened 2019-05-24
Showing the Java environment used by Dataflow in the UI

pabloem:


pabloem:

pabloem:
I need to check how these look in the UI

pabloem:
This is showing up in the user agent - probably not what we want:
image

runners/google-cloud-dataflow-java/src/main/java/org/apache/beam/runners/dataflow/DataflowRunner.java
```diff
@@ -326,8 +326,15 @@ public static DataflowRunner fromOptions(PipelineOptions options) {
     }
 
     DataflowRunnerInfo dataflowRunnerInfo = DataflowRunnerInfo.getDataflowRunnerInfo();
+    // Adding the Java version to the SDK name for user's and support convenience.
+    String javaVersion =
+        Float.parseFloat(System.getProperty("java.specification.version")) >= 9
+            ? "(JDK 11 environment)"
+            : "(JRE 8 environment)";
     String userAgent =
```
Ardagan: Can you elaborate why we mix jre and jdk here?

My understanding of this article is that java.specification.version adheres to current jre specification.
https://docs.oracle.com/javase/7/docs/technotes/guides/versioning/spec/versioning2.html

pabloem: Yeah. For Java 11 there is no "JRE" - the only distribution for Java is the JDK. The containers that we provide have the Java 11 JDK.

For Java 8, there is a development distribution, and a runtime distribution (JRE). For the containers we provide the JRE.

Ardagan: Thank you for clarification.

tvalentyn:
What's next on this PR?

pabloem:
Run Dataflow ValidatesRunner Java 11

pabloem:
This looks like this:
image
image

r: @aaltay

pabloem:

pabloem:

pabloem:

pabloem:
Run Dataflow ValidatesRunner Java 11

pabloem:
Run Dataflow ValidatesRunner Java 11

pabloem:

pabloem:

pabloem:
Run Dataflow ValidatesRunner Java 11

pabloem:
I guess this change affects the SDK version matching, so we can't use it.

pabloem:
This is what it will look like. Is this reasonable?
r: @aaltay 

> image

PR #8683, opened 2019-05-24
[BEAM-5791] splitting support over FnAPI for Go SDK

ananvay:
Adding support for splitting over FnApi for Go SDK.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/go/pkg/beam/core/runtime/exec/datasource.go
+++ b/sdks/go/pkg/beam/core/runtime/exec/datasource.go
@@ -189,7 +189,7 @@ func (n *DataSource) FinishBundle(ctx context.Context) error {
 	err := n.Out.FinishBundle(ctx)
 	n.count = 0
 	n.splitPos = math.MaxInt64
-	n.mu.Unlock()
+	defer n.mu.Unlock()
 	return err
 }
 
@@ -202,12 +202,10 @@ func (n *DataSource) String() string {
 	return fmt.Sprintf("DataSource[%v] Coder:%v Out:%v", n.SID, n.Coder, n.Out.ID())
 }
 
-// IncrementCountAndCheckSplit increments DataSource.count by one and checks to
-// make sure the new value is smaller than the promised split point. If the new
-// value is greater than or equal to the split point, calls the FinishBundle and
-// returns true to indicate that the caller should stop processing elements and
-// exit. If the new value is before the split point (or if the split point
-// hasn't been set), returns false.
+// IncrementCountAndCheckSplit increments DataSource.count by one and checks if
+// the caller should abort further element processing, and finish the bundle.
+// Returns true if the new value of count is greater than or equal to the split
+// point, and false otherwise.
 func (n *DataSource) IncrementCountAndCheckSplit(ctx context.Context) bool {
 	b := false
 	n.mu.Lock()
@@ -248,20 +246,15 @@ func (n *DataSource) Split(splits []int64, frac float32) (int64, error) {
 	}
 	n.mu.Lock()
 	c := n.count
-	// Find the smallest split index that we haven't yet processed.
-	p := math.MaxInt64
-
-	for i := int(0); i < len(splits); i++ {
-		if splits[i] >= c {
-			p = i
-			break
+	// Find the smallest split index that we haven't yet processed, and set
+	// the promised split position to this value.
+	for _, s := range splits {
+		if s >= c {
+			n.splitPos = s
+			fs := n.splitPos
+			n.mu.Unlock()
+			return fs, nil
 		}
-	}
-	if p < len(splits) {
-		n.splitPos = splits[p]
-		fs := n.splitPos
-		n.mu.Unlock()
-		return fs, nil
 	}
 	n.mu.Unlock()
 	// If we can't find a suitable split point from the requested choices,
--- a/sdks/go/pkg/beam/core/runtime/exec/datasource.go
+++ b/sdks/go/pkg/beam/core/runtime/exec/datasource.go
@@ -19,7 +19,8 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"sync/atomic"
+	"math"
+	"sync"
 	"time"
 
 	"github.com/apache/beam/sdks/go/pkg/beam/core/graph/coder"
@@ -34,9 +35,12 @@ type DataSource struct {
 	Coder *coder.Coder
 	Out   Node
 
-	source DataManager
-	count  int64
-	start  time.Time
+	source   DataManager
+	count    int64
+	splitPos int64
+	start    time.Time
+
+	mu sync.Mutex
 }
 
 func (n *DataSource) ID() UnitID {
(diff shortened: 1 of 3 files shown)


ananvay:
@lostluck

ananvay:
@lostluck PTAL.

ananvay:
@udim PTAL.

sdks/go/pkg/beam/core/runtime/exec/datasource.go
```diff
@@ -189,6 +202,23 @@ func (n *DataSource) String() string {
 	return fmt.Sprintf("DataSource[%v] Coder:%v Out:%v", n.SID, n.Coder, n.Out.ID())
 }
 
+// IncrementCountAndCheckSplit increments DataSource.count by one and checks to
+// make sure the new value is smaller than the promised split point. If the new
+// value is greater than or equal to the split point, calls the FinishBundle and
```
lostluck: This code no longer calls FinishBundle directly, so we should remove that part of the comment.

What do you think of this alternative documentation? :

IncrementCountAndCheckSplit increments DataSource.count by one and checks if the caller should abort further element processing, and finish the bundle.
Returns true if the new value of count is greater than or equal to the split point, and false otherwise.

sdks/go/pkg/beam/core/runtime/exec/datasource.go
```diff
@@ -200,5 +230,41 @@ func (n *DataSource) Progress() ProgressReportSnapshot {
...
+// if successful. Returns an error when unable to split.
+func (n *DataSource) Split(splits []int64, frac float32) (int64, error) {
+	if splits == nil {
+		return 0, fmt.Errorf("failed to split: requested splits were empty")
+	}
+	if n == nil {
+		return 0, fmt.Errorf("failed to split at requested splits: {%v}, DataSource not initialized", splits)
+	}
+	n.mu.Lock()
+	c := n.count
```
lostluck: Since we're in the "locked" zone here, we should probably avoid copying the value, for readability.

sdks/go/pkg/beam/core/runtime/exec/datasource.go
```diff
@@ -200,5 +230,41 @@ func (n *DataSource) Progress() ProgressReportSnapshot {
...
+	// Find the smallest split index that we haven't yet processed.
+	p := math.MaxInt64
+
+	for i := int(0); i < len(splits); i++ {
```
lostluck: Consider replacing the two loops with the following single loop, and then we can also remove p as well.
```suggestion
	for _, s := range splits {
		if s >= n.count {
			n.splitPos = s
			fs := n.splitPos
    			n.mu.Unlock()
			return fs, nil 
		}
	}
	n.mu.Unlock()
	...
```
(I might have missed something, but I think these two are the same, and this one is direct)

sdks/go/pkg/beam/core/runtime/harness/harness.go
```diff
@@ -247,11 +247,35 @@ func (c *control) handleInstruction(ctx context.Context, req *fnpb.InstructionRe
...
 		log.Debugf(ctx, "PB Split: %v", msg)
+		ref := msg.GetInstructionReference()
+		c.mu.Lock()
+		plan, ok := c.active[ref]
+		c.mu.Unlock()
+		if !ok {
+			return fail(id, "execution plan for %v not found", ref)
+		}
+
+		ds := msg.GetDesiredSplits()["0"]
```
lostluck: It's OK if there isn't any user friendly reason, but is there some documentation we can use for why it's "0" the whole time?

ananvay: hm, basically b/c we can only split at the root node that reads from fnapi (aka "0").
Added comment to state this.

sdks/go/pkg/beam/core/runtime/exec/datasource.go
ananvay: done

sdks/go/pkg/beam/core/runtime/exec/datasource.go
ananvay: I'm also using 'c' outside the lock to print the value that we saw when we held the lock...

sdks/go/pkg/beam/core/runtime/exec/datasource.go
ananvay: done.

sdks/go/pkg/beam/core/runtime/exec/datasource.go
```diff
@@ -173,10 +183,13 @@ func (n *DataSource) Process(ctx context.Context) error {
...
 func (n *DataSource) FinishBundle(ctx context.Context) error {
-	log.Infof(ctx, "DataSource: %d elements in %d ns", atomic.LoadInt64(&n.count), time.Now().Sub(n.start))
+	n.mu.Lock()
+	log.Infof(ctx, "DataSource: %d elements in %d ns", n.count, time.Now().Sub(n.start))
 	n.source = nil
 	err := n.Out.FinishBundle(ctx)
-	atomic.StoreInt64(&n.count, 0)
+	n.count = 0
+	n.splitPos = math.MaxInt64
+	n.mu.Unlock()
```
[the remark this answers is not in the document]
ananvay: done.

ananvay:
Thanks @lostluck ! Updated, PTAL.

sdks/go/pkg/beam/core/runtime/exec/datasource.go
```diff
@@ -173,10 +183,13 @@ func (n *DataSource) Process(ctx context.Context) error {
...
 func (n *DataSource) FinishBundle(ctx context.Context) error {
-	log.Infof(ctx, "DataSource: %d elements in %d ns", atomic.LoadInt64(&n.count), time.Now().Sub(n.start))
+	n.mu.Lock()
+	log.Infof(ctx, "DataSource: %d elements in %d ns", n.count, time.Now().Sub(n.start))
 	n.source = nil
 	err := n.Out.FinishBundle(ctx)
-	atomic.StoreInt64(&n.count, 0)
+	n.count = 0
+	n.splitPos = math.MaxInt64
+	defer n.mu.Unlock()
```
lostluck: The defer needs to be immeadiately after the n.mu.Lock() call to be effective. If FinishBundle panics, this line would never be reached.

defer can be read as  "no matter how this FinishBundle method exits, ensure this code runs." Kind of like Java's "finally" in a "try {} catch {}finally {}" paradigm.

sdks/go/pkg/beam/core/runtime/exec/datasource.go
lostluck: I missed that! Thanks!

sdks/go/pkg/beam/core/runtime/exec/datasource.go
ananvay: done!

ananvay:
thanks!

lostluck:
@aaltay Please merge! Thanks!

udim:
Hi, could you please annotate the description of the PR with the associated JIRA issue?
(and open one if there isn't)

ananvay:
https://issues.apache.org/jira/browse/BEAM-5791

This change is the "unreading elements already sent by the runner" part.

ananvay:
Also related: https://issues.apache.org/jira/browse/BEAM-6655

udim:
run go postcommit

PR #8689, opened 2019-05-25
[BEAM-7426] Add underscore support on Field lexer

alexvanboxel:
Underscore is a common used word delimiter in field names, the current
FieldSpecifierNotationLexer only support alpha-numeric values for field
name character.

The upcoming Protobuf schema support will emit underscores in the field
names, so field names should support underscore.


alexvanboxel:
@reuvenlax this PR would make it possible to use proto field names as they contain underscores in general. This would make the following code work:

```
            .apply(Filter.<FooBar>create().whereFieldName("test_name", u -> u.equals("abc")))
            .apply(Select.fieldNames("test_name"))
            .apply(ParDo.of(new DoFn<Row, String>() {
```

It seems that Filter works, but Select doesn't without the fix. Even so, I think the lexer is too restrictive and needs to be revised later on.

PR #8690, opened 2019-05-26
[BEAM-7274] Implement the Protobuf schema provider

alexvanboxel:
The default implementation for Protobuf messages. Supports both
generated as dynamic messages.

----
NOTE: This implementation doesn't support Unknown Fields. I like to have a discussion with the reviewers how todo this.


alexvanboxel:
@reuvenlax here is the first draft of Protobuf support. I hope the unit test `testCoder` and `testCoderOnDynamic` proves that having `oneof`'s work as nullable fields (if you look at the Java API., they are actually treated as such. I think almost all permutations are unit-tested.

Things to discuss: Unknown fields.

reuvenlax:
I'm still reviewing this. A few high-level thoughts off the bat:
* I'm not thrilled about creating a ProtoRow object, as I've so far attempted to prevent Row from becoming an "open" class that can by extended by users. I'd be happier if we could do this the way we did Avro classes, by generating a set of getters/setters (it would probably be more efficient as well)
 * There are various places where logical types would be preferable - e.g. all the places where multiple proto types map to a single Beam schema type. As written today the conversion is lossy - if you selected just that field, there would be no way to get the original proto type back; LogicalTypes provide a simple way of fixing this, though we could defer that to a future PR.

alexvanboxel:
@reuvenlax I'll wait for your complete review, but keep in mind that I wanted a uniform way for handling concrete typed classes and DynamicType (DynamicType do require access via the FieldDescriptor). I had trouble getting it in the same way as the AvroUtils, but maybe this is due to my unfamiliarity with it. (I'm open to try to convert it, as most of the effort is spend in the tests anyway. The for internal use only note in the Java doc didn't seem encouraging to use though.

On LogicalTypes: Are you talking about Fixed32/Fixed64/UInt32/UInt64? I simply took the Scalar mapping described here: https://developers.google.com/protocol-buffers/docs/proto#scalar

If you are talking about Timestamp... well I like to have this configurable. Having it default to DATETIME will be a helper for making it available in Beam SQL or in output IO's like BigQuery.

alexvanboxel:
Addition info, so I tried to get the logic using the FieldValueGetter with **partial** success, see experiment: https://github.com/alexvanboxel/beam/commit/d1a8e16a6d8a48b899ed769e4d0837937c31977c

My main issue is with RowWithGetters private `get(Map/List)Value` values that assume they have enough information to construct the embedded complex data structures. I could get rid of my ProtoRow if I could let the handling of the List/Row/Map to it's dedicated `FieldValueGetter`

alexvanboxel:
I got it to work with the RowWithGetters but now without modifying it. See https://github.com/alexvanboxel/beam/commit/1da5715c74d1f3c5de66ae4ebcddcc148bfa7915

I keep on handling the List/Map/Row with the Getters as I need more context then the other implementations. Personally I find this better and more flexible then doing it in the `RowWithGetters` as you have more ways to give context (example: proto map is a repeatable message).

I looked for reference to the `GenericRecord` in Avro, but was disappointed that the `GenericRecord` gets materialised in a `RowWithStorage`. I'm quite sure with having the getters that handle collection this could be changed to `RowWithGetters`

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoDescriptorSerializer.java
```diff
@@ -0,0 +1,137 @@
...
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Utility class that makes a Descriptor class Serializable. This should be part of the ProtoBuf
+ * library but for moving forward this is copied to the Beam code base.
+ *
+ * <p>Follow https://github.com/protocolbuffers/protobuf/issues/6101 for the state of the feature
+ * request in ProtoBuf.
+ */
```
reuvenlax: Could you instead convert the Descriptor to a proto (using the Descriptor.toProto method) and store that? Then I think you won't need this class at all.

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoRow.java
```diff
@@ -0,0 +1,556 @@
...
+import org.joda.time.Instant;
+
+/**
+ * ProtoRow extends the Row and does late materialisation. It hold a reference to the original proto
+ * message and has an overlay on each field of the proto message. It doesn't have it's own Coder as
+ * it relies on the SchemaCoder to (de)serialize the message over the wire.
+ *
+ * <p>Each row has a FieldOverlay that handles specific field conversions, as well has special
+ * overlays for Well Know Types, Repeatables, Maps and Nullables.
+ */
```
reuvenlax: As mentioned above I think we can create this functionality without this class (though we will need a lot of the code in this class). For now I'm reviewing this class as is, and later we can see if we can use the getter/setter/builder framework instead.

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoRow.java
```diff
@@ -0,0 +1,556 @@
...
+    @Override
+    public Object convertGetObject(Object object) {
+      // return object;
+      return ((ByteString) object).toByteArray();
+    }
+
+    @Override
+    public void set(Message.Builder message, Object value) {
+      if (value != null && ((byte[]) value).length > 0) {
+        // Protobuf messages BYTES doesn't like emptry bytes?!
```
reuvenlax: typo

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoRow.java
```diff
@@ -0,0 +1,556 @@
...
+    WrapperOverlay(ProtoSchema protoSchema, Descriptors.FieldDescriptor fieldDescriptor) {
+      this.fieldDescriptor = fieldDescriptor;
+      this.valueDescriptor = fieldDescriptor.getMessageType().findFieldByName("value");
+      this.value = protoSchema.createFieldLayer(valueDescriptor, false);
+    }
+
+    @Override
+    public ValueT get(Message message) {
+      if (message.hasField(fieldDescriptor)) {
+
```
reuvenlax: empty line

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoRow.java
```diff
@@ -0,0 +1,556 @@
...
+    public Instant get(Message message) {
+      if (message.hasField(fieldDescriptor)) {
+        Message wrapper = (Message) message.getField(fieldDescriptor);
+        return convertGetObject(wrapper);
+      }
+      return null;
+    }
+
+    @Override
+    public Instant convertGetObject(Object object) {
```
reuvenlax: this loses precision.

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoSchema.java
```diff
@@ -0,0 +1,326 @@
...
+    return new RowToMessageFunction(this);
+  }
+
+  private void writeObject(ObjectOutputStream oos) throws IOException {
+    oos.defaultWriteObject();
+    if (DynamicMessage.class.equals(this.rawType)) {
+      if (this.descriptor == null) {
+        throw new RuntimeException("DynamicMessages require provider a Descriptor to the coder.");
+      }
+      ProtoDescriptorSerializer.writeObject(oos, this.descriptor);
```
reuvenlax: As mentioned above - just serialize this.descriptor.toProto() instead

sdks/java/extensions/protobuf/src/test/java/org/apache/beam/sdk/extensions/protobuf/ProtoSchemaTest.java
```diff
@@ -0,0 +1,506 @@
...
+
+    provider.add("", Proto3SchemaMessages.Complex.getDescriptor());
+    RowCoder rowCoder = RowCoder.of(schemaCoder.getSchema());
+
+    byte[] schemaCoderBytes = SerializableUtils.serializeToByteArray(schemaCoder);
+    SchemaCoder<Message> schemaCoderCoded =
+        (SchemaCoder<Message>) SerializableUtils.deserializeFromByteArray(schemaCoderBytes, "");
+    byte[] rowCoderBytes = SerializableUtils.serializeToByteArray(rowCoder);
+    RowCoder rowCoderCoded =
+        (RowCoder) SerializableUtils.deserializeFromByteArray(rowCoderBytes, "");
```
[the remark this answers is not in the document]
alexvanboxel: This is my understanding of what happens when Coder and Rows are encoded/decoded. I wanted to make sure everything is properly serialized.

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoSchema.java
```diff
@@ -0,0 +1,326 @@
...
+  private transient ArrayList<ProtoRow.FieldOverlay> getters;
+
+  ProtoSchema(Class rawType, Descriptors.Descriptor descriptor, Map<String, Class> overlayClasses) {
+    this.rawType = rawType;
+    this.descriptor = descriptor;
+    this.typeMapping = overlayClasses;
+    init();
+  }
+
+  static Schema.FieldType convertType(Descriptors.FieldDescriptor.Type type) {
```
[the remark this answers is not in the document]
alexvanboxel: I was following the Proto Java implementation here: https://developers.google.com/protocol-buffers/docs/proto ( [3] 64-bit or unsigned 32-bit integers are always represented as long when decoded, but can be an int if an int is given when setting the field. In all cases, the value must fit in the type represented when set ). Anyway, can be discussed on the JIRA ticket.

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoRow.java
alexvanboxel: Did you look at my experiment for converting it to (at least the getter framework): see my comment, just before your review comments.

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoRow.java
alexvanboxel: Like to make this switchable via options, Instant will be enough for most Beam use-cases and will make it work nicely with Beam SQL.

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoRow.java
reuvenlax: Yes. I want to review this first though.

alexvanboxel:
ok, I was wondering if I should take action on this. Get it in master and
then rebase the other section.

 _/
_/ Alex Van Boxel

reuvenlax:
Yes. I'd prefer to keep Row as a somewhat sealed class and not introduce extra subtypes. However there was enough other things to review in this PR that I figured I could review that first.

alexvanboxel:
OK, I'll fix the issues you mentioned, although I doubt that I can do the
.toProto as it probably lacks
contextual information to reconstruct the descriptors, as the Descriptor
has a lot of possibilities to
reference other components like Enums and other Messages.

But I'll try...

 _/
_/ Alex Van Boxel

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoSchema.java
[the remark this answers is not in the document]
alexvanboxel: Added Jira: https://issues.apache.org/jira/browse/BEAM-7518

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoRow.java
alexvanboxel: Same ticket: https://issues.apache.org/jira/browse/BEAM-7518

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoDescriptorSerializer.java
alexvanboxel: So I removed this class as it was not super elegant, but .toProto() is not a solution as it only works in one way. A descriptor is always contained within a FileDescriptor.
I've introduced a Domain class that could group descriptors that are related with each other. This is an important component that a ProtoBuf schema registry should provision to make all Descriptors crosswired. See the Javadoc in the ProtoDomain class.

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoSchema.java
alexvanboxel: see ProtoDomain comment.

alexvanboxel:
> Yes. I'd prefer to keep Row as a somewhat sealed class and not introduce extra subtypes. However there was enough other things to review in this PR that I figured I could review that first.

Answered (and fixed comments), I also rebased the experiment to remove the ProtoRow as a subtype: https://github.com/alexvanboxel/beam/commit/1629557d6a6114b18ef4e3526b01900e5466a6e5

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoSchema.java
```diff
@@ -0,0 +1,365 @@
...
+    }
+
+    @Override
+    public Row apply(Message input) {
+      return Row.withSchema(protoSchema.schema)
+          .withFieldValueGettersHandleCollections(true)
+          .withFieldValueGetters(new ProtoOverlayFactory(), input)
+          .build();
+
+      // return new ProtoRow(protoSchema.schema, protoSchema.getters, input);
```
reuvenlax: remove?

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoSchema.java
```diff
@@ -0,0 +1,365 @@
...
+    return domain;
+  }
+
+  /** Overlay. */
+  public static class ProtoOverlayFactory implements Factory<List<FieldValueGetter>> {
+
+    public ProtoOverlayFactory() {}
+
+    @Override
+    public List create(Class<?> clazz, Schema schema) {
```
reuvenlax: should this return List<FieldValueGetter?>

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoSchemaProvider.java
```diff
@@ -0,0 +1,140 @@
...
+  public ProtoSchemaProvider() {
+    this.domain = STATIC_COMPILED_DOMAIN;
+  }
+
+  public ProtoSchemaProvider(ProtoDomain domain) {
+    this.domain = domain;
+  }
+
+  private SchemaCoder ensure(Class rawType) {
+    String urn = "class:" + rawType.getName();
```
reuvenlax: why do you need to convert this to a string? Can't you just make the Class itself the map key?

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoSchemaProvider.java
```diff
@@ -0,0 +1,140 @@
...
+ * external registry that maps those URN's with Descriptors.
+ */
+@Experimental(Experimental.Kind.SCHEMAS)
+public class ProtoSchemaProvider implements SchemaProvider {
+  private static final Logger LOG = LoggerFactory.getLogger(ProtoSchemaProvider.class);
+
+  private static final ProtoDomain STATIC_COMPILED_DOMAIN = new ProtoDomain();
+  private ProtoDomain domain;
+  private Map<String, SchemaCoder> cache = new HashMap<>();
+  private Map<String, Class> overlayClasses = new HashMap<>();
```
reuvenlax: these maps aren't thread safe

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoSchemaProvider.java
```diff
@@ -0,0 +1,140 @@
...
+  public ProtoSchemaProvider(ProtoDomain domain) {
+    this.domain = domain;
+  }
+
+  private SchemaCoder ensure(Class rawType) {
+    String urn = "class:" + rawType.getName();
+    SchemaCoder coder = cache.get(urn);
+    if (coder == null) {
+      return add(urn, rawType);
+    }
```
reuvenlax: Map.computeIfAbsent is a better way of implementing this pattern.

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoSchemaProvider.java
```diff
@@ -0,0 +1,140 @@
...
+ * <p>It's possible to inherit this class for a specific implementation that communicates with an
+ * external registry that maps those URN's with Descriptors.
+ */
+@Experimental(Experimental.Kind.SCHEMAS)
+public class ProtoSchemaProvider implements SchemaProvider {
+  private static final Logger LOG = LoggerFactory.getLogger(ProtoSchemaProvider.class);
+
+  private static final ProtoDomain STATIC_COMPILED_DOMAIN = new ProtoDomain();
+  private ProtoDomain domain;
+  private Map<String, SchemaCoder> cache = new HashMap<>();
```
reuvenlax: is there a reason to cache SchemaCoder here?

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoSchemaProvider.java
alexvanboxel: Cache is completely removed, as using the ProtoSchemaProvider is now only used for Proto messages pre-compiled into the pipeline. Using DynamicTypes should be done though ProtoSchema builder.

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoSchemaProvider.java
alexvanboxel: Cache has been remoed.

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoSchemaProvider.java
alexvanboxel: As the functionality of the DynamicTypes are removed for the ProtoSchemaProvider the cache is not useful anymore. Most of the Dynamic behaviour is moved to ProtoDomain.

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoSchema.java
alexvanboxel: yes, changed to include it.

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoSchemaProvider.java
alexvanboxel: Maps are move, they are now only filled up in a builder at Pipeline construction time, and at build() are put in an ImmutableMap

sdks/java/core/src/main/java/org/apache/beam/sdk/values/Row.java
```diff
@@ -554,6 +555,12 @@ public Builder withFieldValueGetters(
       return this;
     }
 
+    /** The FieldValueGetters will handle the conversion for Arrays, Maps and Rows. */
+    public Builder withFieldValueGettersHandleCollections(boolean collectionHandledByGetter) {
+      this.collectionHandledByGetter = collectionHandledByGetter;
+      return this;
+    }
```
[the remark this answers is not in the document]
alexvanboxel: OK, maybe the name could be better, but it means that the the FieldValueGetters also handle collections like ARRAY, ROW, MAP. If you go to the original implementation ( https://github.com/apache/beam/blob/master/sdks/java/core/src/main/java/org/apache/beam/sdk/values/RowWithGetters.java#L90 ) you see that RowWithGetters does naive handling of ARRAY, ROW and MAP, for protobuf you need more context (the descriptor) to handle them. That's why I need to disable the naive mapping and let the FieldValueGetters handle ARRAY, ROW and MAP.

Feel free to suggest a better name though.

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoRow.java
alexvanboxel: ProtoRow completely removed. Everything delegated to RowWithGetters.

alexvanboxel:
@reuvenlax I think I tackled all your concerns you posted. The build is failing, but this is due to a warning in Jenkins that has nothing todo with this PR (I investigated). The PR is also rebased.

- Reverted changes to Row so it's back to the original implementation, except for the need of `collectionHandledByGetter` see remark above. (better name always welcome)
- ProtoDescriptorSerializer is gone. I defined a ProtoDomain class that the user needs to use themself if they want to use DynamicMessages. having one ProtoDomain for all messages that belong together will make sure object equality works. 
- ProtoSchemaProvider only handles compiled Proto classes, it will give an error on DynamicMessages.
- ProtoSchema is the workhorse, it's a bit like AvroSchemaUtils.

It would be great having this in the next release. From then on we can do incremental changes like:

- Documentation
- Option handling
- Logical types

I also backported the ProtoDomain on the DynamicMessage PR to align it with this.

Sorry it took so long, but I needed to prioritize something different, but related with Proto (a proto schema registry).

alexvanboxel:
Run Java PreCommit

alexvanboxel:
Requesting @pablo to be added as reviewer.

alexvanboxel:
Requesting @pabloem as a reviewer.

sdks/java/core/src/main/java/org/apache/beam/sdk/values/Row.java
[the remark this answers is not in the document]
TheNeuralBit: @reuvenlax is this the last remaining concern you have for this PR?

https://github.com/apache/beam/pull/8690#issuecomment-497695744 is a good reference for the motivation of this as well.

I'm not crazy about the way this is implemented since it's adding state to `RowWithGetter` that will get checked every time a collection field is accessed. I can't think of a better way to do it without some non-trivial refactoring though. Some ideas:

1.  Rather than storing `collectionHandledByGetter` in `RowWithGetters` and changing behavior based on it, have two alternate `RowWithGetters` implementations, one with the special handling for collections and one without. I think this is still an improvement over the separate `ProtoRow` class that was rejected since it's not explicitly tied to a particular `SchemaProvider`, in fact it sounds like it could be re-used for Avro `GenericRecord` instances.
1. Move the  special logic for collection getters into those `SchemaProvider`/`FieldValueGetter` implementations that need it, and make `RowWithGetters` always behave as if `collectionHandledByGetter` is true. I think this could be a cleaner approach? But it's challenging because that special logic stores cached values that wouldn't be appropriate to move into the `FieldValueGetter` implementations. Maybe a `SchemaProvider` could have some way to indicate that a certain row is expensive to access and should be cached in `RowWithGetters`?

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoSchemaProvider.java
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Schema provider for Protobuf messages. The provider is able to handle pre compiled Message file
+ * without external help. For Dynamic Messages a Descriptor needs to be registered up front on a
+ * specific URN.
+ *
+ * <p>It's possible to inherit this class for a specific implementation that communicates with an
+ * external registry that maps those URN's with Descriptors.
+ */
+@Experimental(Experimental.Kind.SCHEMAS)
+public class ProtoSchemaProvider implements SchemaProvider {
```
TheNeuralBit: Is the need for `withFieldValueGettersHandleCollections(true)` the only reason this can't be made to implement `GetterBasedSchemaProvider`?

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoSchema.java
```diff
@@ -0,0 +1,563 @@
...
+ * iterpret and reconstruct messages.
+ *
+ * <ul>
+ *   <li>Protobuf oneOf fields are mapped to nullable fields and flattened into the parent row.
+ *   <li>Protobuf primitives are mapped to it's nullable counter part.
+ *   <li>Protobuf maps are mapped to nullable maps, where empty maps are mapped to the null value.
+ *   <li>Protobuf repeatables are mapped to nullable arrays, where empty arrays are mapped to the
+ *       null value.
+ *   <li>Protobuf enums are mapped to non-nullable string values.
+ *   <li>Enum map to their string representation
```
TheNeuralBit: nit: Are these two lines saying the same thing?

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoSchema.java
```diff
@@ -0,0 +1,563 @@
...
+ *
+ * <p>ProtoDomain is an integral part of a ProtoSchema, it it contains all the information needed to
+ * iterpret and reconstruct messages.
+ *
+ * <ul>
+ *   <li>Protobuf oneOf fields are mapped to nullable fields and flattened into the parent row.
+ *   <li>Protobuf primitives are mapped to it's nullable counter part.
+ *   <li>Protobuf maps are mapped to nullable maps, where empty maps are mapped to the null value.
+ *   <li>Protobuf repeatables are mapped to nullable arrays, where empty arrays are mapped to the
+ *       null value.
```
TheNeuralBit: Why not just use non-nullable maps and arrays and produce empty ones when the proto ones are empty?

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoFieldOverlay.java
```diff
@@ -0,0 +1,525 @@
...
+
+    @Override
+    public Object convertGetObject(FieldDescriptor fieldDescriptor, Object in) {
+      return in.toString();
+    }
+
+    @Override
+    public void set(Message.Builder message, Object value) {
+      //     builder.setField(fieldDescriptor,
+      // convertSetObject(row.getString(fieldDescriptor.getName())));
```
TheNeuralBit: nit: can you remove this?

sdks/java/extensions/protobuf/src/test/java/org/apache/beam/sdk/extensions/protobuf/ProtoSchemaValuesTest.java
```diff
@@ -0,0 +1,670 @@
...
+    ProtoSchema protoSchema =
+        ProtoSchema.newBuilder(domain).forDescriptor(proto.getDescriptorForType());
+
+    toRowFunction = protoSchema.getSchemaCoder().getToRowFunction();
+    fromRowFunction = protoSchema.getSchemaCoder().getFromRowFunction();
+  }
+
+  @Test
+  public void testRowAndBack() {
+    setup();
```
TheNeuralBit: I wonder if this could be re-structured a bit so that you could use `@Before` instead of calling the setup functions yourself. Maybe if you have two different Parameterized test classes that draw from the same data? One for message and one for dynamic message. You can have multiple suites in one outer class if you annotate it with `@RunWith(Enclosed.class)` (like I did [here](https://github.com/apache/beam/pull/9446/files#diff-cd81b3c0b9fb7c9081a9be718cc5cdd1R77))

This is definitely a nitpicky thing... totally fine to leave it as is.

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoFieldOverlay.java
alexvanboxel: removed

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoSchema.java
alexvanboxel: Business wise this makes more sense. As proto doesn't distinguish between empty map/array and not set. It's also more performant as if the default would be an empty map/array it would result in a bigger Row. It would be an option to make it later configurable.

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoSchema.java
alexvanboxel: Removed

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoSchemaProvider.java
alexvanboxel: No, each data type has it's own Provider (eg. AvroSchemaProvider). This class makes sure Proto can be auto mapped.

sdks/java/core/src/main/java/org/apache/beam/sdk/values/Row.java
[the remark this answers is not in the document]
alexvanboxel: I've refactored from your input: I've created a RowWithGettersCachedCollection that inherits for RowWithGetters. This cached is the default.

alexvanboxel:
This PR is now rebased against master.This removed ProtoDomain (and it's tests) from this PR as it was part of the DymamicMessage PR that is now part of master. I think I've incorporated most of the comment of the review in this PR as well.

alexvanboxel:
@TheNeuralBit / @pabloem can you re-review. Thanks.

sdks/java/core/src/main/java/org/apache/beam/sdk/values/Row.java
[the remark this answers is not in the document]
reuvenlax: Can you help me understand this a bit more? Why does it not work to cache lists for protocol buffers? We saw repeated array conversion to be a big problem (which is why we cache them). I'm wondering if we could instead cache a lazy array like we do with iterables.

I'll take a closer look at this code to figure it out.

reuvenlax:
Another comment: this delegates everything to Proto's reflection API, which can be quite inefficient. Compare with AvroSchema where we delegate straight to generated classes when they exist. Reflection might be necessary for the case of DynamicMessage (though in that case I think we should use RowWithStorage instead of RowWithGetters), but shouldn't be necessary when we have generated classes.

alexvanboxel:
> Another comment: this delegates everything to Proto's reflection API, which can be quite inefficient. Compare with AvroSchema where we delegate straight to generated classes when they exist. 

**Totally agree, but** I wanted first a solution that works for every use-case. Optimization can happen afterward. I still have **plans for iterative improvements**, certainly related to protobuf options first, then optimize for performance (but I have to make performance tests for this to measure this)

> Reflection might be necessary for the case of DynamicMessage (though in that case I think we should use RowWithStorage instead of RowWithGetters), but shouldn't be necessary when we have generated classes.

RowWithStorage could be a solution when the Row is materialized (with the getToRowFunction), but having the reflection API is still needed it you use the schema without materializing. Example just you the filter on a column function on a schema aware collection but keeping the original type, you don't want to materialize the a big proto just to filter on 1 column. Again, I like to measure this, but we need a base line.

sdks/java/extensions/protobuf/src/main/java/org/apache/beam/sdk/extensions/protobuf/ProtoSchemaProvider.java
alexvanboxel: Let me elaborate. The **ProtoSchemaProvider** delegates the functionality from to/fromRow to the ProtoSchema class that is, yes, optimized for dynamic types. The ProtoSchema class managed most of the descriptors. that's it doesn't the default GetterBasedProvider.

If we would optimize for generated classes this could switch (but I prefer this to do in a separate PR later on).

reuvenlax:
BTW I've checked out the PR to play with it a bit more. My worry is that in its current form "optimizing later" will not be so simple, since it doesn't currently fit in with the framework used by AvroSchema, AutoValueSchema, JavaBeanSchema, etc.. If you don't mind, I'll try and play with it a bit to see if I can get it to fit into that framework a bit better.

sdks/java/core/src/main/java/org/apache/beam/sdk/values/Row.java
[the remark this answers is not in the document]
alexvanboxel: Would we make a ticket out of this or is this blocking?

alexvanboxel:
> BTW I've checked out the PR to play with it a bit more. My worry is that in its current form "optimizing later" will not be so simple, since it doesn't currently fit in with the framework used by AvroSchema, AutoValueSchema, JavaBeanSchema, etc.. If you don't mind, I'll try and play with it a bit to see if I can get it to fit into that framework a bit better.

I don't mind, the thing that worries me is that it will miss the 2.18 as well (as it missed the 2.17 already). As we are depending in our business on this feature we are currently building our own version of Beam and that's a pain. For me the DynamicMessage is very important... I was thinking for the normal use-case going to a complete other codepath.

alexvanboxel:
Closing this PR as it's supersede by several other PR that are already on master: Last remaining PR for full proto support is: https://github.com/apache/beam/pull/10502

Option/Metadata support has been split of and can be followed:
- https://github.com/apache/beam/pull/10413 (options implementation)
- https://github.com/apache/beam/pull/10529 (options support out of proto)

PR #8691, opened 2019-05-27
[BEAM-7428] Output the timestamp on elements in ReadAllViaFileBasedSource

iemejia:
I also refactored the way the Read is done by adding an additional `DoFn` to remove the repeated code in `runners/core/core-construction-java`
R: @chamikaramj

### DIFF
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/io/ReadAllViaFileBasedSource.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/io/ReadAllViaFileBasedSource.java
@@ -18,8 +18,14 @@
 package org.apache.beam.sdk.io;
 
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.Serializable;
 import org.apache.beam.sdk.annotations.Experimental;
 import org.apache.beam.sdk.coders.Coder;
+import org.apache.beam.sdk.coders.CoderException;
+import org.apache.beam.sdk.coders.CustomCoder;
+import org.apache.beam.sdk.coders.SerializableCoder;
 import org.apache.beam.sdk.io.FileIO.ReadableFile;
 import org.apache.beam.sdk.io.fs.MatchResult.Metadata;
 import org.apache.beam.sdk.io.fs.ResourceId;
@@ -61,7 +67,9 @@ public class ReadAllViaFileBasedSource<T>
     return input
         .apply("Split into ranges", ParDo.of(new SplitIntoRangesFn(desiredBundleSizeBytes)))
         .apply("Reshuffle", Reshuffle.viaRandomKey())
-        .apply("Read ranges", ParDo.of(new ReadFileRangesFn<>(createSource)))
+        .apply("Create sources", ParDo.of(new CreateBoundedSourceFn<>(createSource)))
+        .setCoder(new BoundedSourceCoder<>())
+        .apply("Read ranges", ParDo.of(new ReadFromBoundedSourceFn<>()))
         .setCoder(coder);
   }
 
@@ -86,10 +94,11 @@ public class ReadAllViaFileBasedSource<T>
     }
   }
 
-  private static class ReadFileRangesFn<T> extends DoFn<KV<ReadableFile, OffsetRange>, T> {
+  private static class CreateBoundedSourceFn<T>
+      extends DoFn<KV<ReadableFile, OffsetRange>, BoundedSource<T>> {
     private final SerializableFunction<String, ? extends FileBasedSource<T>> createSource;
 
-    private ReadFileRangesFn(
+    private CreateBoundedSourceFn(
         SerializableFunction<String, ? extends FileBasedSource<T>> createSource) {
       this.createSource = createSource;
     }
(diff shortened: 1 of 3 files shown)


sdks/java/core/src/main/java/org/apache/beam/sdk/io/ReadAllViaFileBasedSource.java
```diff
@@ -98,15 +108,53 @@ private ReadFileRangesFn(
...
+   * SerializableCoder#structuralValue(Serializable)} assumes that coded elements support object
+   * equality (https://issues.apache.org/jira/browse/BEAM-3807). By default, Coders compare equality
+   * by serialized bytes, which we want in this case. It is usually safe to depend on coded
+   * representation here because we only compare objects on bundle commit, which compares
+   * serializations of the same object instance.
+   *
+   * <p>BoundedSources are generally not used as PCollection elements, so we do not expose this
+   * coder for wider use.
+   */
+  public static class BoundedSourceCoder<T> extends CustomCoder<BoundedSource<T>> {
```
iemejia: Wondering if I should put thisas an inner class of BoundedSource or other place or if is ok here..
Also if we should auto register this Coder in `BoundedSource`.

iemejia:
It seems that now that it is taking the timestamp into account some tests (in particular watch/continous reading tests break with something like this:

```
org.apache.beam.sdk.Pipeline$PipelineExecutionException: java.lang.IllegalArgumentException: Cannot output with timestamp -290308-12-21T19:59:05.225Z. Output timestamps must be no earlier than the timestamp of the current input (2019-05-28T08:22:29.859Z) minus the allowed skew (0 milliseconds). See the DoFn#getAllowedTimestampSkew() Javadoc for details on changing the allowed skew.
	at org.apache.beam.runners.direct.DirectRunner$DirectPipelineResult.waitUntilFinish(DirectRunner.java:348)
	at org.apache.beam.runners.direct.DirectRunner$DirectPipelineResult.waitUntilFinish(DirectRunner.java:318)
```

@chamikaramj Do you think the assertion is wrong? or any suggestion on the best road to take to fix this issue.

sdks/java/core/src/main/java/org/apache/beam/sdk/io/ReadAllViaFileBasedSource.java
chamikaramj: Probably this should be it's own file if public ?

I don't think there's a reason to register since I don't think using BoundedSources as elements of a PCollection should be a recommended pattern for Beam (even though we do it here as a stop gap till we get SDF).

sdks/java/core/src/main/java/org/apache/beam/sdk/io/ReadAllViaFileBasedSource.java
```diff
@@ -98,15 +107,53 @@ private ReadFileRangesFn(
...
+
+    @Override
+    public BoundedSource<T> decode(InputStream inStream) throws CoderException, IOException {
+      return coder.decode(inStream);
+    }
+  }
+
+  /** Reads elements contained within an input {@link BoundedSource}. */
+  // TODO: Extend to be a Splittable DoFn.
+  public static class ReadFromBoundedSourceFn<T> extends DoFn<BoundedSource<T>, T> {
```
chamikaramj: Ditto.

chamikaramj:
Looking at the error seems like a default (negative) timestamp value was set somehow ? I'm not sure.

Failures are bit concerning though. Does this change break backwards compatibility somehow ?

chamikaramj:
cc: @kennknowles

sdks/java/core/src/main/java/org/apache/beam/sdk/io/ReadAllViaFileBasedSource.java
iemejia: Yes that's exactly what I thought. I mean I over refactored in this PR I can just let things as they were, where this is done just at the `runners/core-construction-java` level so no need to do that.

sdks/java/core/src/main/java/org/apache/beam/sdk/io/ReadAllViaFileBasedSource.java
iemejia: Wondering the same if I should better let this as it was, because the reuse we do here is pretty low, however is where I noticed the difference between `JavaReadViaImpulse` and `ReadAllViajFileBasedSource`.

iemejia:
I agree that the failures are weird, but maybe they are happening because we were hiding the case by not adding the timestamp when outputting. In any case it seems we should be consistent in both classes where we do the similar action: `JavaReadViaImpulse` and `ReadAllViaFileBasedSource`. Do we take into account the timestamp or do we ignore it but probably is not good to mix both behaviors (that's why I did all the other refactorings, but they are not mandatory for the core aspect of this iissue).

kennknowles:
The stacktrace you posted was not the root cause. It is helpful to know the stacktrace of the IllegalArgumentException. The assertion is likely preventing output of data that would just be droppable anyhow. The error message does not seem quite right, but I bet there is danger here.

iemejia:
To be sure that this is not a side effect of the refactors I am going to close this and do the minimum viable PR to fix the issue in PR #8741
Let s move the discussion there.
I will bring the other 'minor' clean ups in a different PR afterwards.

PR #8692, opened 2019-05-27
 Refactor default bundle size into a constant in TextIO

iemejia:
Make more readable the default bundle how for JavaReadViaImpulse too.
These changes are mostly for consistency around the codebase.

R: @aromanenko-dev


iemejia:
Merged as green. Thanks!

PR #8715, opened 2019-05-29
[BEAM-7442][BEAM-5650] Read sequentially from bounded sources in UnboundedSourceWrapper

mxm:
Bounded sources are converted to unbounded sources using the
BoundedToUnboundedSourceAdapter. This applies the streaming semantics for
unbounded sources, which is to read round-robin from all sources after
splitting. This doesn't work particularly well with a large number of files and
leads to a large number of open file handles / connections.

Instead, we can simple read the bounded sources sequentially.

--- | --- | --- | --- | ---

### DIFF
--- a/runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/io/UnboundedSourceWrapper.java
+++ b/runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/io/UnboundedSourceWrapper.java
@@ -22,6 +22,7 @@ import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import org.apache.beam.runners.core.construction.SerializablePipelineOptions;
+import org.apache.beam.runners.core.construction.UnboundedReadFromBoundedSource;
 import org.apache.beam.runners.flink.FlinkPipelineOptions;
 import org.apache.beam.runners.flink.metrics.FlinkMetricContainer;
 import org.apache.beam.runners.flink.metrics.ReaderInvocationUtil;
@@ -69,6 +70,13 @@ public class UnboundedSourceWrapper<OutputT, CheckpointMarkT extends UnboundedSo
   private final String stepName;
   /** Keep the options so that we can initialize the localReaders. */
   private final SerializablePipelineOptions serializedOptions;
+
+  /**
+   * We are processing bounded data and should read from the sources sequentially instead of reading
+   * round-robin from all the sources. In case of file sources this avoids having too many open
+   * files/connections at once.
+   */
+  private final boolean isConvertedBoundedSource;
 
   /** For snapshot and restore. */
   private final KvCoder<? extends UnboundedSource<OutputT, CheckpointMarkT>, CheckpointMarkT>
@@ -134,6 +142,8 @@ public class UnboundedSourceWrapper<OutputT, CheckpointMarkT extends UnboundedSo
       throws Exception {
     this.stepName = stepName;
     this.serializedOptions = new SerializablePipelineOptions(pipelineOptions);
+    this.isConvertedBoundedSource =
+        source instanceof UnboundedReadFromBoundedSource.BoundedToUnboundedSourceAdapter;
 
     if (source.requiresDeduping()) {
       LOG.warn("Source {} requires deduping but Flink runner doesn't support this yet.", source);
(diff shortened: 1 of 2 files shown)


mxm:
@JozoVilcek Could you have a look?

runners/flink/src/test/java/org/apache/beam/runners/flink/translation/wrappers/streaming/io/UnboundedSourceWrapperTest.java
```diff
@@ -707,4 +780,77 @@ public StreamStatus getStreamStatus() {
       return currentStreamStatus;
     }
   }
+
+  private static class BoundedIntegerSource extends OffsetBasedSource<Integer> {
```
JozoVilcek: Hm,`CountingSource.upTo()` is deprecated. Otherwise it might be possible to reuse it here. Just a note

runners/flink/src/test/java/org/apache/beam/runners/flink/translation/wrappers/streaming/io/UnboundedSourceWrapperTest.java
```diff
@@ -690,6 +701,68 @@ private static void testSourceDoesNotShutdown(boolean shouldHaveReaders) throws
...
+            public Object getCheckpointLock() {
+              return new Object();
+            }
+
+            @Override
+            public void close() {}
+          });
+
+      // the source is effectively split into two parts
+      assertThat(integers.size(), is(500));
```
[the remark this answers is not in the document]
mxm: If you look further up, it sets the number of parallel sub tasks to 2 and the initial splitting is performed with a parallelism of 4. So that makes 2 instances for this partition which is run in the test.

Good point on the output ordering. I've improved the output check. Please have a look.

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/io/UnboundedSourceWrapper.java
```diff
@@ -217,34 +227,33 @@ public void run(SourceContext<WindowedValue<ValueWithRecordId<OutputT>>> ctx) th
...
-      }
-
+    } else if (isConvertedBoundedSource) {
       setNextWatermarkTimer(this.runtimeContext);
 
-      while (isRunning) {
-        boolean dataAvailable;
-        synchronized (ctx.getCheckpointLock()) {
-          dataAvailable = readerInvoker.invokeAdvance(reader);
+      // We read sequentially from all bounded sources
```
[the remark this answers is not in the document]
mxm: It doesn't make a difference because the watermark is always the minimum across all readers. In case of the `UnboundedReadFromBoundedSource` adapter, all input is read from all sources and only then the watermark is progressed.

See: https://github.com/apache/beam/commit/11d9ec5ebff4820b36db4b6ea4df7a0f79115ddd#diff-6bd5346c8d8943cdb40aa5cf5b1731bdR354

runners/flink/src/test/java/org/apache/beam/runners/flink/translation/wrappers/streaming/io/UnboundedSourceWrapperTest.java
mxm: There is also `CountingSource.createUnboundedFrom` which is package-private, but it looks like all the public methods are deprecated.

runners/flink/src/test/java/org/apache/beam/runners/flink/translation/wrappers/streaming/io/UnboundedSourceWrapperTest.java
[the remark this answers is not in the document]
JozoVilcek: Aha, I missed `Mockito.when(runtimeContextMock.getIndexOfThisSubtask()).thenReturn(0)` .. I thought it is going to process both partitions. Thanks

runners/flink/src/test/java/org/apache/beam/runners/flink/translation/wrappers/streaming/io/UnboundedSourceWrapperTest.java
JozoVilcek: I think it is deprecated for the sake of `GenerateSequence` which works on `PTransform` level. It uses `CountingSource` internally. Maybe it would make sense to be able to use also `source` level from this. But maybe there are not so many sources like this in tests

mxm: Looking at all the other test which use `CountingSource.upTo()` I think it is probably ok to depend on this method, instead of essentially duplicating code. I've removed the source.

mxm:
Thanks for the review!

mxm:
Unrelated failure in Java PreCommit: ```org.apache.beam.runners.dataflow.worker.fn.control.TimerReceiverTest.testSingleTimerScheduling```

PR #8718, opened 2019-05-29
[BEAM-7450] Support unbounded reads with HCatalogIO

jhalaria:
This PR adds the ability to add unbounded amounts of data using HCatalogIO

1.  It extends PollFn which is responsible for emitting new partition indices when they become available
2. A reader which will read the partition that was jus emitted.

R: @iemejia @lukecwik 
CC: @jkff  (just for info or glad to receive feedback if you are interested).

--- | --- | --- | --- | ---


jhalaria:
@iemejia - Please review when you get a chance. Thank you.

apilloud:
Test failure looks real.
```

java.lang.IllegalArgumentException: org.apache.beam.sdk.io.hcatalog.PartitionPoller, @ProcessElement processElement(ProcessContext, PartitionRangeTracker): Has tracker type PartitionRangeTracker, but the DoFn's tracker type must be of type RestrictionTracker.
	at org.apache.beam.sdk.transforms.reflect.DoFnSignatures$ErrorReporter.throwIllegalArgument(DoFnSignatures.java:1495)
	at org.apache.beam.sdk.transforms.reflect.DoFnSignatures$ErrorReporter.checkArgument(DoFnSignatures.java:1500)
	at org.apache.beam.sdk.transforms.reflect.DoFnSignatures.verifySplittableMethods(DoFnSignatures.java:595)
	at org.apache.beam.sdk.transforms.reflect.DoFnSignatures.parseSignature(DoFnSignatures.java:474)
	at org.apache.beam.sdk.transforms.reflect.DoFnSignatures.lambda$getSignature$0(DoFnSignatures.java:140)
	at java.util.HashMap.computeIfAbsent(HashMap.java:1127)
	at org.apache.beam.sdk.transforms.reflect.DoFnSignatures.getSignature(DoFnSignatures.java:140)
	at org.apache.beam.sdk.transforms.ParDo.validate(ParDo.java:547)
	at org.apache.beam.sdk.transforms.ParDo.of(ParDo.java:397)
	at org.apache.beam.sdk.io.hcatalog.HCatalogUnboundedReader.expand(HCatalogUnboundedReader.java:110)
	at org.apache.beam.sdk.io.hcatalog.HCatalogUnboundedReader.expand(HCatalogUnboundedReader.java:77)
	at org.apache.beam.sdk.Pipeline.applyInternal(Pipeline.java:537)
	at org.apache.beam.sdk.Pipeline.applyTransform(Pipeline.java:471)
	at org.apache.beam.sdk.values.PCollection.apply(PCollection.java:357)
	at org.apache.beam.sdk.io.hcatalog.HCatalogIOTest.testWriteThenUnboundedReadSuccess(HCatalogIOTest.java:220)
```

iemejia:
I still need to check in more details the Readers and the intention. Not sure if I understood well but I have the impression that the Bounded could be eventually a replacement of HCatalogIO current Read. Is this the case? If so probably it would be easier to divide this PR in two (one that introduces the Bounded SDF version) and another for the Unbounded but well maybe I have to do a second round to get this correctly.

jhalaria:
@iemejia - Thank you so much for your feedback.

- I have addressed all of your recommendations from above.
- Removed the extra tracker and used offsetRangeTracker instead.
- Unit test coverage for the PartitionRangeTracker.
- Refined the public API to be HcatalogIO.unbounded()
- This PR only adds an unbounded reads capability. In order for the unit tests to not continuously poll for new partitions, I had to add a flag to make the unbounded behave as bounded.
- I will have a subsequent PR that unifies the two readers into one and the caller can decide which one to use based a flag. If we like this approach, then we can potentially have just one `reads` API. However the only down side of doing this would be adding certain checks that allows reading a table even without partitions which is a requirement for the unbounded reader to work correctly.


Please have another look at the PR and let me know your feedback. Thanks.

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/HCatRecordReaderFn.java
```diff
@@ -0,0 +1,204 @@
...
+    // pass the 'desired' split count as an hint to the API
+    Map<String, String> configProps = new HashMap<>(spec.getConfigProperties());
+    configProps.put(
+        HCatConstants.HCAT_DESIRED_PARTITION_NUM_SPLITS, String.valueOf(desiredSplitCount));
+    return DataTransferFactory.getHCatReader(entity, configProps).prepareRead();
+  }
+
+  private long getFileSizeForPartitions(PartitionWrapper partitionWrapper) throws Exception {
+    IMetaStoreClient client = null;
+    Configuration conf = new Configuration();
```
iemejia: Can you extract the code that creates a new HiveMetastoreClient into a new static method in a class called `HCatalogUtils` and use it here and in `PartitionPollerFn` too. Extra points if you change it in `HCatalogIO.getEstimatedSizeBytes` too.

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/HCatalogIO.java
```diff
@@ -123,8 +132,132 @@ public static Read read() {
...
+          readSpec.getWatermarkPartitionColumn() != null,
+          "withsWatermarkPartitionColumn() is required");
+      checkArgument(
+          readSpec.getWatermarkTimestampConverter() != null,
+          "withWatermarkTimestampConverter() is required");
+      checkArgument(readSpec.getPartitionCols() != null, "withPartitionCols() is required");
+      checkArgument(
+          readSpec.getPartitionComparator() != null, "withPartitionComparator() is required");
+      return input
+          .apply("ConvertToReadRequest", Create.of(new ArrayList<>(Arrays.asList(readSpec))))
```
iemejia: `.apply("ConvertToReadRequest", Create.of(readSpec))` simpler = better.

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/PartitionPollerFn.java
```diff
@@ -0,0 +1,146 @@
...
+          forSort.indexOf(tracker.currentRestriction().getLastCompletedPartition());
+      trueStart = indexOfLastCompletedPartition + 1;
+    } else {
+      trueStart = 0;
+    }
+
+    for (int i = trueStart; i < forSort.size(); i++) {
+      if (tracker.tryClaim(forSort.get(i))) {
+        c.output(
+            new HCatRecordReaderFn.PartitionWrapper(
```
iemejia: I am not sure that we need the `PartitionWrapper` class. Can't we just add the missing part of the `Read`, the partition as an attribute of the readRequest (with package private access) and then just pass the readRequest (`Read`) object all around?

jhalaria: Good point. After the refactor, since everything is now in Read(s), this can go away.

jhalaria:
@iemejia - I think once this PR gets merged, the next step will clean up the multiple reads method that we currently have. Moving forward there will only be one Read class that can read both bounded and unbounded data depending on the parameters passed. Do you think we can tackle that as the next step or if not what would make it cleaner in this PR? 
When I am extending from Read, I have to add the boilerplate code for the properties inside Read. I thing that we can do in this PR itself is to have only 1 Reads class and then within the expand, call the DoFn's needed for unbounded reads if only the right arguments are set. What do you think?

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/HCatRecordReaderFn.java
```diff
@@ -0,0 +1,165 @@
...
+    // process the splits returned by native API
+    // this could be different from 'desiredSplitCount' calculated above
+    LOG.info(
+        "Splitting: estimated size {}, desired split count {}, actual split count",
+        estimatedSizeBytes,
+        desiredSplitCount);
+    return new OffsetRange(0, readerContext.numSplits());
+  }
+
+  @NewTracker
```
jkff: This is unnecessary to declare, the default implementation will do the same thing.

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/HCatRecordReaderFn.java
```diff
@@ -0,0 +1,165 @@
...
+        readRequest.getWatermarkTimestampConverter().apply(partitionWaterMark);
+
+    int desiredSplitCount = 1;
+    long estimatedSizeBytes = getFileSizeForPartitions(processContext.element());
+    if (estimatedSizeBytes > 0) {
+      desiredSplitCount = (int) Math.ceil((double) estimatedSizeBytes / Integer.MAX_VALUE);
+    }
+    ReaderContext readerContext = getReaderContext(processContext.element(), desiredSplitCount);
+
+    for (long i = tracker.currentRestriction().getFrom(); tracker.tryClaim(i); i++) {
```
jkff: Hmm why does this one need to be an SDF rather than simply a composition of two DoFn's with a reshuffle in between? (where the first DoFn would emit (Read, int) pairs where the int is 0..numSplits) Is there so many splits that it is prohibitive?

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/HCatalogIO.java
```diff
@@ -123,12 +129,147 @@ public static Read read() {
     return new AutoValue_HCatalogIO_Read.Builder().setDatabase(DEFAULT_DATABASE).build();
   }
 
+  /**
+   * Read unbounded data from Hive. This essentially means that as new partitions show up, we start
+   * reading data for that partition.
+   */
+  public static UnboundedRead unbounded() {
```
jkff: Are you sure you need a new class for this, can you just add a boolean parameter to Read "isUnbounded" and add an if() statement in expand?

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/PartitionPollerFn.java
```diff
@@ -0,0 +1,141 @@
...
+import org.apache.beam.vendor.guava.v20_0.com.google.common.collect.ImmutableList;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hive.metastore.IMetaStoreClient;
+import org.apache.hadoop.hive.metastore.api.Partition;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Unbounded poller to listen for new partitions. */
+@UnboundedPerElement
+class PartitionPollerFn extends DoFn<Read, Read> {
```
jkff: Can this one be replaced with a Watch transform, whose PollFn returns a complete list of current partitions? it scales quite well (if I remember correctly, e.g. a million elements was not a problem, as long as you encode them efficiently), and seems like it would save you a huge amount of code in this PR.

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/HCatalogIO.java
iemejia: That was probably an ill advise of my part (there was  a totally different approach, not the read based one too). @jhalaria I think this is probablysimilar in principle to what we called the unification, maybe worth to tackle it for this PR.

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/PartitionPollerFn.java
iemejia: Just curiosity question @jkff, does this mean then that we won't need an Unbounded SDF then? How do you tackle the case when new partitions appear?
(KafkaIO has an open ticket on detecting new partitions so I was wondering if the approach you propose could work there too).

jkff: Yes, it does mean that we don't need an SDF here at all; and yes, this would likely work for KafkaIO too. Watch would poll for the current list of partitions, and emit new ones as they appear without re-emitting the already known ones, just like when we watch for new files.

It also means that the current PR would not be introducing any new SDFs, but instead merely reusing an existing one (Watch) - which is a very good thing and in line with SDF's position as enabling more composable/reusable IO APIs as opposed to people having to reimplement similar features in different IOs.

iemejia: Wow, that's super awesome, thanks for bringing this idea @jkff !

iemejia: @jhalaria For further clarity what @jkff is suggesting is to use `Watch` to follow a pattern like [`FileIO.Match.continously`](https://github.com/apache/beam/blob/2798d626507c7f5c8699096b3a188200ac64a299/sdks/java/core/src/main/java/org/apache/beam/sdk/io/FileIO.java#L627) does to poll and discover the new partitions.

jhalaria:
Hey @jkff , Thank you for your suggestion. It would definitely make this a lot simpler. I just wanted to make sure that I understand your overall solution correctly:

- Add a simple Watch transform that outputs all partitions List<Partitions> whenever it gets invoked. Signature of Transform 
```java 
PartitionWatcher extends PollFn<HCatalogIO.Read, List<Partition>> ```
- A DoFn that will invoke HcatalogIO.read with a filter condition added that can read the partitions.

**Comment by jkff** [comment]
@jhalaria Yes, I'm suggesting that you replace ParDo.of(new PartitionPollFn) with Watch.growthOf(new PartitionPollFn), and change PartitionPollFn to be a Watch.PollFn instead of a DoFn.

(Note that this will also make your tests much simpler: now you only need to verify that your PollFn correctly returns the results of a single poll - the rest is already tested in Watch)

**Comment by jhalaria** [comment]
@jkff  @iemejia - This is much cleaner after implementing @jkff suggestions. Should be ready for review now.

File: sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/HCatalogIO.java [hunk 8]
```diff
@@ -186,6 +216,26 @@ public Read withFilter(String filter) {
       return toBuilder().setFilter(filter).build();
     }
 
+    /** Sets the duration after which the PartitionPollerFn resumes polling for new partitions. */
```
jkff: This documentation ideally should be phrased in terms that make sense to a client, rather than reference implementation details. E.g. "If specified, poll for new partitions at this periodicity. Then the returned PCollection will be unbounded. If not specified, (what happens then?)"

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/HCatalogIO.java
```diff
@@ -186,6 +216,26 @@ public Read withFilter(String filter) {
...
+    public Read withPollingInterval(Duration pollingInterval) {
+      return toBuilder().setPollingInterval(pollingInterval).build();
+    }
+
+    /** Specifies the name of the column which contains the watermark values. */
```
jkff: This comment looks wrong. I'm generally not sure what this property is for.

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/HCatalogIO.java
```diff
@@ -196,11 +246,31 @@ Read withContext(ReaderContext context) {
...
     public PCollection<HCatRecord> expand(PBegin input) {
       checkArgument(getTable() != null, "withTable() is required");
       checkArgument(getConfigProperties() != null, "withConfigProperties() is required");
-
-      return input.apply(org.apache.beam.sdk.io.Read.from(new BoundedHCatalogSource(this)));
+      if (!getShouldTreatUnboundedAsBounded()) {
+        checkArgument(getPollingInterval() != null, "withPollingInterval() is required");
+        final Read readSpec =
+            new AutoValue_HCatalogIO_Read.Builder()
+                .setDatabase(DEFAULT_DATABASE)
```
jkff: Should this really always use DEFAULT_DATABASE?

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/HCatalogIO.java
```diff
@@ -196,11 +246,31 @@ Read withContext(ReaderContext context) {
...
+            .apply(
+                "WatchForNewPartitions",
+                Watch.growthOf(new PartitionPollerFn(getConfigProperties()))
+                    .withPollInterval(getPollingInterval()))
+            .apply(Values.create())
+            .apply("PartitionSplitter", ParDo.of(new PartitionSplitterFn()))
+            .apply(Reshuffle.of())
+            .apply("PartitionReader", ParDo.of(new PartitionReaderFn()));
+      } else {
+        // Treat as Bounded
```
jkff: If the only difference is that either we poll for new partitions or we don't, then we don't need shouldTreatUnboundedAsBounded, and should simply check whether pollingInterval is specified.

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/PartitionPollerFn.java
```diff
@@ -0,0 +1,71 @@
...
+import org.apache.beam.sdk.io.hcatalog.HCatalogIO.Read;
+import org.apache.beam.sdk.transforms.DoFn.Setup;
+import org.apache.beam.sdk.transforms.DoFn.Teardown;
+import org.apache.beam.sdk.transforms.Watch.Growth.PollFn;
+import org.apache.beam.sdk.transforms.Watch.Growth.PollResult;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hive.metastore.IMetaStoreClient;
+import org.joda.time.Instant;
+
+/** Unbounded poller to listen for new partitions. */
```
jkff: Boundedness or unboundedness do not apply to a poll function, it simply returns the current state of the thing being polled.

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/PartitionPollerFn.java
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  private List<Read> getPartitions(Read read) throws Exception {
+    return metaStoreClient.listPartitions(read.getDatabase(), read.getTable(), Short.MAX_VALUE)
+        .stream()
+        .map(partition -> read.withPartitionToRead(partition))
+        .collect(Collectors.toList());
+  }
+
+  @Setup
```
jkff: Poll fns don't have setup or teardown, you need to put this into apply() itself.

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/PartitionSplitterFn.java
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.beam.sdk.io.hcatalog;
+
+import org.apache.beam.sdk.io.hcatalog.HCatalogIO.Read;
+import org.apache.beam.sdk.transforms.DoFn;
+import org.apache.beam.sdk.values.KV;
+
+/** Unbounded reader for doing reads from Hcat. */
```
jkff: This comment looks wrong.

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/PartitionSplitterFn.java
```diff
@@ -0,0 +1,36 @@
...
+class PartitionSplitterFn extends DoFn<Read, KV<Read, Integer>> {
+
+  /** Splits partition into key value pairs. */
+  @ProcessElement
+  @SuppressWarnings("unused")
```
jkff: What's unused here?

sdks/java/io/hcatalog/src/test/java/org/apache/beam/sdk/io/hcatalog/HCatalogIOTest.java
```diff
@@ -163,6 +179,55 @@ public void processElement(ProcessContext c) {
...
+    final PCollection<HCatRecord> data =
+        readAfterWritePipeline
+            .apply(
+                "ReadData",
+                HCatalogIO.read()
+                    .withConfigProperties(getConfigPropertiesAsMap(service.getHiveConf()))
+                    .withDatabase(TEST_DATABASE)
+                    .withPartitionCols(partitions)
+                    .withTable(TEST_TABLE)
+                    .withPollingInterval(Duration.millis(15000)))
```
jkff: Does this test terminate? I think your code sets up Watch.growthOf() in a way that doesn't terminate..

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/HCatalogUtils.java
```diff
@@ -0,0 +1,79 @@
...
+      conf.set(entry.getKey(), entry.getValue());
+    }
+    return HCatUtil.getHiveConf(conf);
+  }
+
+  static int getSplitCount(Read readRequest) throws Exception {
+    int desiredSplitCount = 1;
+    long estimatedSizeBytes = HCatalogUtils.getFileSizeForPartitions(readRequest);
+    if (estimatedSizeBytes > 0) {
+      desiredSplitCount = (int) Math.ceil((double) estimatedSizeBytes / Integer.MAX_VALUE);
```
jkff: What's the motivation for this logic?

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/PartitionPollerFn.java
```diff
@@ -0,0 +1,71 @@
...
+import org.apache.beam.sdk.transforms.DoFn.Setup;
+import org.apache.beam.sdk.transforms.DoFn.Teardown;
+import org.apache.beam.sdk.transforms.Watch.Growth.PollFn;
+import org.apache.beam.sdk.transforms.Watch.Growth.PollResult;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hive.metastore.IMetaStoreClient;
+import org.joda.time.Instant;
+
+/** Unbounded poller to listen for new partitions. */
+public class PartitionPollerFn extends PollFn<Read, Read> {
```
jkff: The output of a PollFn should be something that can be efficiently and deterministically encoded (inefficiency will limit scalability; non-deterministic encoding will make Watch produce wrong results). I think in your case it could simply be an Integer for the partition index?

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/HCatalogIO.java
jhalaria: Removed. Older property. Not needed.

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/HCatalogIO.java
jhalaria: I didn't really need to do this here. Removed this logic of creating a spec from here. However, the current version of HcatalogIO does use `DEFAULT_DATABASE`. I can change it to not do that if we want.

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/PartitionSplitterFn.java
jhalaria: Intellij complains about processElement as unused. It was more to make intellij happy. Will remove it.

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/HCatalogUtils.java
jhalaria: While reading a partition, we calculate the size of the partition, so that we can set the value for HCatConstants.HCAT_DESIRED_PARTITION_NUM_SPLITS

sdks/java/io/hcatalog/src/test/java/org/apache/beam/sdk/io/hcatalog/HCatalogIOTest.java
jhalaria: It didn't earlier. Added the ability to specify a termination condition

jhalaria:
@jkff - Thank you very much for your feedback. I have addressed all suggestions. Please let me know if this looks good now.

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/PartitionReaderFn.java
```diff
@@ -0,0 +1,115 @@
...
+            .build();
+    // pass the 'desired' split count as an hint to the API
+    Map<String, String> configProps = new HashMap<>(readRequest.getConfigProperties());
+    configProps.put(
+        HCatConstants.HCAT_DESIRED_PARTITION_NUM_SPLITS, String.valueOf(desiredSplitCount));
+    return DataTransferFactory.getHCatReader(entity, configProps).prepareRead();
+  }
+
+  @ProcessElement
+  @SuppressWarnings("unused")
```
iemejia: remove `@SuppressWarnings("unused")` everywhere

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/HCatalogIO.java
```diff
@@ -163,6 +176,12 @@ private HCatalogIO() {}
 
       abstract Builder setContext(ReaderContext context);
 
+      abstract Builder setPollingInterval(Duration pollingInterval);
+
+      abstract Builder setPartitionCols(ImmutableList<String> partitionCols);
```
iemejia: Please make this a `List` to not have guava dependencies in the public API

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/HCatalogIO.java
```diff
@@ -186,6 +205,28 @@ public Read withFilter(String filter) {
...
+     * If specified, polling for new partitions will happen at this periodicity. The returned
+     * PCollection will be unbounded. However if a withTerminationCondition is set along with
+     * pollingInterval, polling will stop after the termination condition has been met.
+     */
+    public Read withPollingInterval(Duration pollingInterval) {
+      return toBuilder().setPollingInterval(pollingInterval).build();
+    }
+
+    /** Set the names of the columns that are partitions. */
+    public Read withPartitionCols(ImmutableList<String> partitionCols) {
```
iemejia: same as above for List

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/PartitionPollerFn.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.stream.IntStream;
+import org.apache.beam.sdk.io.hcatalog.HCatalogIO.Read;
+import org.apache.beam.sdk.transforms.Watch.Growth.PollFn;
+import org.apache.beam.sdk.transforms.Watch.Growth.PollResult;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hive.metastore.IMetaStoreClient;
+import org.joda.time.Instant;
+
+/** Return the list of current partitions present. */
+public class PartitionPollerFn extends PollFn<Read, Integer> {
```
iemejia: Make package private

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/PartitionReaderFn.java
```diff
@@ -0,0 +1,115 @@
...
+    final List<Partition> partitions =
+        metaStoreClient.listPartitions(
+            readRequest.getDatabase(), readRequest.getTable(), Short.MAX_VALUE);
+    final Partition partition = partitions.get(partitionIndexToRead);
+    checkArgument(
+        partition != null, "Unable to find a partition to read at index " + partitionIndexToRead);
+
+    final int desiredSplitCount = HCatalogUtils.getSplitCount(readRequest, partition);
+    final List<String> values = partition.getValues();
+    final ImmutableList<String> partitionCols = readRequest.getPartitionCols();
```
iemejia: What happens if the user does not define partitionCols as it was before. Maybe worth to make it an empty list as default when building the Read.

jhalaria:
@iemejia - Feedback addressed.

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/HCatalogIO.java
```diff
@@ -196,11 +237,27 @@ Read withContext(ReaderContext context) {
...
     public PCollection<HCatRecord> expand(PBegin input) {
       checkArgument(getTable() != null, "withTable() is required");
       checkArgument(getConfigProperties() != null, "withConfigProperties() is required");
-
-      return input.apply(org.apache.beam.sdk.io.Read.from(new BoundedHCatalogSource(this)));
+      Watch.Growth<Read, Integer, Integer> growthFn;
+      if (getPollingInterval() != null) {
+        growthFn = Watch.growthOf(new PartitionPollerFn()).withPollInterval(getPollingInterval());
+        if (getTerminationCondition() != null) {
+          growthFn =
```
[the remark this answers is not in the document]
iemejia: Sorry I was maybe not clear, I am talking about not repeating agan the code `Watch.growthOf(new PartitionPollerFn()).withPollInterval(getPollingInterval())` because that's already done just before in the definition of `growthFn` and instead do: `growthFn = growthFn.withTerminationPerInput(getTerminationCondition());`.

It is pretty minor but nice to fix.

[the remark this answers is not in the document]
jhalaria: Ah. I see. Done

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/HCatalogIO.java
```diff
@@ -196,11 +253,24 @@ Read withContext(ReaderContext context) {
...
+      if (getPollingInterval() != null) {
+        growthFn = Watch.growthOf(new PartitionPollerFn()).withPollInterval(getPollingInterval());
+        if (getTerminationCondition() != null) {
+          growthFn = growthFn.withTerminationPerInput(getTerminationCondition());
+        }
+        return input
+            .apply("ConvertToReadRequest", Create.of(this))
+            .apply("WatchForNewPartitions", growthFn)
+            .apply("PartitionReader", ParDo.of(new PartitionReaderFn(getConfigProperties())));
+      } else {
```
jkff: Maybe add a check that getTerminationCondition() == null.

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/HCatalogUtils.java
jkff: Sorry, I mean, why is the correct value for desiredSplitCount = estimatedSizeBytes/Integer.MAX_VALUE? Is it cause we want every split to be of size Integer.MAX_VALUE? If so, why? That seems pretty large.

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/PartitionPollerFn.java
```diff
@@ -0,0 +1,56 @@
...
+      metaStoreClient.close();
+    }
+    return pollResult;
+  }
+
+  private List<Integer> getPartitionIndices(Read read) throws Exception {
+    return IntStream.range(
+            0,
+            metaStoreClient
+                .listPartitions(read.getDatabase(), read.getTable(), Short.MAX_VALUE)
```
jkff: What does Short.MAX_VALUE stand for here, and is it in any way related to Integer.MAX_VALUE above?

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/PartitionReaderFn.java
```diff
@@ -0,0 +1,111 @@
...
+        HCatConstants.HCAT_DESIRED_PARTITION_NUM_SPLITS, String.valueOf(desiredSplitCount));
+    return DataTransferFactory.getHCatReader(entity, configProps).prepareRead();
+  }
+
+  @ProcessElement
+  public void processElement(ProcessContext c) throws Exception {
+    final Read readRequest = c.element().getKey();
+    final Integer partitionIndexToRead = c.element().getValue();
+    ReaderContext readerContext = getReaderContext(readRequest, partitionIndexToRead);
+    for (int i = 0; i < readerContext.numSplits(); i++) {
```
jkff: Why do we need a nested loop here: is it possible to just read the whole partition without "splitting" it?

If you want to read the splits of a partition in parallel, then you need to split this into two DoFn's with a reshuffle in between.

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/PartitionPollerFn.java
jhalaria: listPartitions requires maximumPartitions that can be listed while making that API call. https://hive.apache.org/javadocs/r2.2.0/api/org/apache/hadoop/hive/metastore/HiveMetaStoreClient.html#listPartitions-java.lang.String-java.lang.String-short-

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/HCatalogUtils.java
jhalaria: Ah. I see. I fixed it to have the desired bundle size has 128 MB as default. This is only a hint to the DataTransfer API(s). The actual number of splits is won't be exactly what we specify.

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/PartitionReaderFn.java
jhalaria: - As far as I know Data Transfer API(s) do not provide a way to read the whole partition.
- As far as splitting the reads into multiple DoFn's is concerned, I tried doing that but ran into an  issue which needs a fix with hive. [described below]
- Here is what I tried:

1. Pass in <ReaderContext, SplitToRead> as an output of the first DoFn.
2. Create a 2nd DoFn which reads actual data given a context and a split.
3. Added a reshuffle between the 2 DoFn(s)

This results in a `NullPointer Exception` because HCatTableInfo is transient in the ReaderContext and we don't have a way to deserialize it. I will create a PR with hcatalog to fix that. HCatTableInfo doesn't have to be transient since it is serializable.

jhalaria:
@jkff - Addressed feedback.

jkff:
LGTM, I'll let @iemejia merge though, since I'm not currently actively involved in Beam.

jhalaria:
@jkff @iemejia - Thank you very much for all your suggestions in making this PR a lot cleaner.

sdks/java/io/hcatalog/src/main/java/org/apache/beam/sdk/io/hcatalog/PartitionReaderFn.java
iemejia: Can you please fill the JIRA for this. Thanks!

iemejia:
Merged manually to squash the commits.

iemejia:
Thanks a lot @jhalaria not only for the interesting use case but for being so eager to address the suggestions.
Thanks also @jkff it is really a pleasure to see you around giving useful feedback as before even if you are not active in the project anymore.

PR #8725, opened 2019-05-30
[BEAM-6674] Add schema support to JdbcIO read

charithe:
Adds support for directly reading `Row` objects from a JDBC datasource by automatically converting the database schema to a Beam schema.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcIO.java
+++ b/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcIO.java
@@ -35,7 +35,9 @@ import org.apache.beam.sdk.annotations.Experimental;
 import org.apache.beam.sdk.coders.Coder;
 import org.apache.beam.sdk.coders.RowCoder;
 import org.apache.beam.sdk.options.ValueProvider;
+import org.apache.beam.sdk.schemas.NoSuchSchemaException;
 import org.apache.beam.sdk.schemas.Schema;
+import org.apache.beam.sdk.schemas.SchemaRegistry;
 import org.apache.beam.sdk.transforms.Create;
 import org.apache.beam.sdk.transforms.DoFn;
 import org.apache.beam.sdk.transforms.Filter;
@@ -57,6 +59,7 @@ import org.apache.beam.sdk.values.PCollection;
 import org.apache.beam.sdk.values.PCollectionView;
 import org.apache.beam.sdk.values.PDone;
 import org.apache.beam.sdk.values.Row;
+import org.apache.beam.sdk.values.TypeDescriptor;
 import org.apache.commons.dbcp2.BasicDataSource;
 import org.apache.commons.dbcp2.DataSourceConnectionFactory;
 import org.apache.commons.dbcp2.PoolableConnectionFactory;
@@ -798,6 +801,16 @@ public class JdbcIO {
 
       if (getOutputParallelization()) {
         output = output.apply(new Reparallelize<>());
+      }
+
+      try {
+        TypeDescriptor<OutputT> typeDesc = getCoder().getEncodedTypeDescriptor();
+        SchemaRegistry registry = input.getPipeline().getSchemaRegistry();
+        Schema schema = registry.getSchema(typeDesc);
+        output.setSchema(
+            schema, registry.getToRowFunction(typeDesc), registry.getFromRowFunction(typeDesc));
+      } catch (NoSuchSchemaException e) {
+        // ignore
       }
 
       return output;
--- a/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcIO.java
+++ b/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcIO.java
@@ -33,7 +33,9 @@ import javax.annotation.Nullable;
 import javax.sql.DataSource;
 import org.apache.beam.sdk.annotations.Experimental;
 import org.apache.beam.sdk.coders.Coder;
+import org.apache.beam.sdk.coders.RowCoder;
 import org.apache.beam.sdk.options.ValueProvider;
+import org.apache.beam.sdk.schemas.Schema;
 import org.apache.beam.sdk.transforms.Create;
 import org.apache.beam.sdk.transforms.DoFn;
 import org.apache.beam.sdk.transforms.Filter;
@@ -54,6 +56,7 @@ import org.apache.beam.sdk.values.PBegin;
 import org.apache.beam.sdk.values.PCollection;
 import org.apache.beam.sdk.values.PCollectionView;
 import org.apache.beam.sdk.values.PDone;
+import org.apache.beam.sdk.values.Row;
 import org.apache.commons.dbcp2.BasicDataSource;
 import org.apache.commons.dbcp2.DataSourceConnectionFactory;
 import org.apache.commons.dbcp2.PoolableConnectionFactory;
@@ -183,6 +186,15 @@ public class JdbcIO {
    */
   public static <T> Read<T> read() {
     return new AutoValue_JdbcIO_Read.Builder<T>()
+        .setFetchSize(DEFAULT_FETCH_SIZE)
+        .setOutputParallelization(true)
+        .build();
+  }
+
+  /** Read Beam {@link Row}s from a JDBC data source. */
+  @Experimental(Experimental.Kind.SCHEMAS)
+  public static ReadRows readRows() {
+    return new AutoValue_JdbcIO_ReadRows.Builder()
         .setFetchSize(DEFAULT_FETCH_SIZE)
         .setOutputParallelization(true)
         .build();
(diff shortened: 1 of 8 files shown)


charithe:
R: @reuvenlax 
R: @jbonofre

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcIO.java
```diff
@@ -188,6 +191,15 @@
...
   }
 
+  /** Read Beam {@link Row}s from a JDBC data source. */
+  @Experimental(Experimental.Kind.SCHEMAS)
+  public static ReadRows readRows() {
+    return new AutoValue_JdbcIO_ReadRows.Builder()
+        .setFetchSize(DEFAULT_FETCH_SIZE)
+        .setOutputParallelization(true)
+        .build();
+  }
```
reuvenlax: This shouldn't be limited to just reading Row objects, as any type can have a schema. So you should be able to have a readWithSchema(Class<T> clazz) and get back a PCollection<T> (if the SchemaRegistry contains a schema for clazz, and if that schema matches the one inferred at the source). readRows is still useful but can be a wrapper around this.

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcIO.java
```diff
@@ -391,6 +403,123 @@ DataSource buildDatasource() {
     void setParameters(PreparedStatement preparedStatement) throws Exception;
   }
 
+  /** Implementation of {@link #readRows()}. */
+  @AutoValue
+  @Experimental(Experimental.Kind.SCHEMAS)
+  public abstract static class ReadRows extends PTransform<PBegin, PCollection<Row>> {
```
reuvenlax: can we share the part that's common with Read? I'm worried that that people will end up adding features to one codepath and not the other.

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcIO.java
charithe: Do you prefer an explicit `readWithSchema` method or do you prefer it to be implicit?. That is:

- Make the JDBC row mapper optional
- If the user provides a JDBC row mapper, then:
    - If the output type of the row mapper exists in the registry, attach that schema to the output PCollection
    - Otherwise, the output PCollection will not have a schema
- If the user does not provide a JDBC row mapper, then:
    - If the output type exists in the schema registry and is compatible with the inferred schema from the database, generate our own row mapper to map the resultset and attach the schema to the output PCollection
    - Otherwise, throw an exception

reuvenlax:
I think we can make it implicit in this case.

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcIO.java
reuvenlax: I think we can make it implicit in this case.

charithe: I looked into this a little bit more and I think I may have misunderstood a few things.

I had the idea that if I get the `TypeDescriptor` of the PCollection, I could query the schema registry to get the necessary information to attach the schema. However, this doesn't seem to work in practice. Of course, I could be doing something wrong there but I came to the realisation that if it was indeed a viable solution, then it would have been already implemented at the core as it would provide a generic way to implicitly attach schemas to any PCollection that contains suitable element types.

Now I am not quite sure what benefit a `readWithSchema(clazz)` method would have. If the user does need schema support, they could just as well call `setSchema` on the PCollection themselves, can't they? We can reduce the boilerplate involved but that is not specific to the JDBC IO module and should probably go in to the core schema module as a utility function.

Since the user is responsible for defining how to map the ResultSet to a type of their choosing, they could potentially do some field transforms during the mapping and produce an output type whose schema doesn't match up with the schema inferred from the raw data. Therefore I feel like comparing the two schemas will not be very useful either.

What are your thoughts?

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcIO.java
charithe: Do you have any suggestions on how to do this? I wasn't happy with the duplication either but I couldn't see a nice way to share state between AutoValue classes. Both the existing `Read` and `ReadAll` classes are pretty much identical to each other as well so I assumed that was a known issue.

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/SchemaUtil.java
```diff
@@ -0,0 +1,321 @@
...
+                  .put(Schema.TypeName.STRING, ResultSet::getString)
+                  .build());
+
+  private static final ResultSetFieldExtractor DATE_EXTRACTOR = createDateExtractor();
+  private static final ResultSetFieldExtractor TIME_EXTRACTOR = createTimeExtractor();
+  private static final ResultSetFieldExtractor TIMESTAMP_EXTRACTOR = createTimestampExtractor();
+
+  /*
+   * Mapping of primitive JDBC types to Beam schema types.
+   * TODO: Add support for BLOB and CLOB types
```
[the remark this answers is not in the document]
charithe: Done: https://issues.apache.org/jira/browse/BEAM-7485

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcIO.java
reuvenlax: I was thinking that this would allow us to not have to create a separate fork of the Read PTransform. Just keep one Read<T>, which happens to support schemas (as long as T is in the SchemaRegistry); yes the user can call setSchema themselves, however forcing the user to do a lookup in the SchemaRegistry themselves in ugly, and we should make that implicit. Then readRows will just return Read<Row>, so we should not need a separate class.

Does that work?

charithe: So I assumed that it would be easy to obtain the schema for `T` and implicitly attach it to the PCollection. However, this appears to be difficult to do reliably unless the user explicitly provides a reference to `Class<T>` as you correctly suggested. 

Given that the user has to make a conscious effort to enable schema support by providing a reference to `Class<T>`, wouldn't it make sense to introduce a generic utility function such as `PCollections.withSchema(PCollection<T>, Class<T>)` that would work for any kind of PCollection and not just the output of JDBC IO?

reuvenlax:
What you suggested already works via Schema inference. If you have
PCollection<T> and T has a registered schema, then Beam will infer it
(assuming that T is not erased for some reason). However to convert the
Jdbc class to T requires knowing the actual type in the IO.

charithe:
I modified the code to attach the schema to PCollection if the type is registered in the registry (https://github.com/apache/beam/pull/8725/commits/046acd0c7f29e90beffb2fcf9a631f3a6bdc5da1). Is that what you were expecting?

Regarding the `readRows` method I introduced which automatically generates a row mapper based on the inferred schema, I am struggling to figure out a nice way of exposing it without some duplication of the `Read` class. 
- Reusing the `Read` class by making the row mapper and coder optional would be confusing for users because they have to remember to explicitly create a `Read<Row>` instance and be aware of the things they shouldn't set in that instance.
- Extending the `Read` class won't work because the visibility of `withRowMapper` and `withCoder` methods cannot be downgraded.
- As only row mapper and coder are optional, there are still quite few other arguments that the user needs to pass. So making `readRows` accept them all in the argument list would be unfriendly as well.

I may be over-thinking this part. Do you have a suggestion on how you would design the API in this case?

reuvenlax:
inferBeamSchema should be called in the path where you lookup the schema registry as well I think. In that scenario you want to make sure that the schema is compatible.

charithe:
If the user provides their own row mapper to convert the resultset to some type `T`, then I don't see why the schema of `T` should be compatible with the inferred schema of the resultset. For example, the query might have separate fields to hold `first_name` and `last_name` but during the mapping the user might choose to combine those two fields to create a single field named `full_name`.

reuvenlax:
Good point, I agree.

charithe:
Thanks Reuven. I think the only other point remaining is the duplication of code between `Read` and `ReadRows`. As I mentioned in a comment above, I am struggling to find a good solution for that problem. Do you have any suggestions on how to address that problem?

reuvenlax:
I will think about the code duplication a bit more, but I don't think it needs to be resolved for this PR. Could you file a JIRA ticket for that, and I will merge this PR.

reuvenlax:
lgtm

terekete:
Hey all -  I was wondering if there is an example or docs on how to use the the new implementation to read rows using the Row Coder and Row Mapper and auto-infer the beam schema ?

PR #8741, opened 2019-06-02
[BEAM-7428] Output the timestamp on elements in ReadAllViaFileBasedSource

iemejia:
R: @chamikaramj @kennknowles

### DIFF
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/io/ReadAllViaFileBasedSource.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/io/ReadAllViaFileBasedSource.java
@@ -106,7 +106,7 @@ public class ReadAllViaFileBasedSource<T>
               .createForSubrangeOfFile(file.getMetadata(), range.getFrom(), range.getTo())
               .createReader(c.getPipelineOptions())) {
         for (boolean more = reader.start(); more; more = reader.advance()) {
-          c.output(reader.getCurrent());
+          c.outputWithTimestamp(reader.getCurrent(), reader.getCurrentTimestamp());
         }
       }
     }


iemejia:
It seems the failure is the same, and the link with Watch/continuous tests is there.

kennknowles:
Sorry for the delay. Taking another look to refresh myself.

kennknowles:
So I understand why the output timestamp is -inf. That is just how BoundedSource works; it always outputs everything at -inf so that it is safe to move forward to anywhere you want.

So you can't run a bounded source and use those timestamps if the input element has a timestamp later than -inf. But you can just output at the same timestamp as the input if you want, I guess. I think this is an area of SDF that I talked about with Eugene and the conclusion was that we might need per-key watermarks or something crazy. I don't think it got solved.

iemejia:
Does this mean that to have a correct behavior we should make  [`BoundedReader.getCurrentTimestamp`](https://github.com/apache/beam/blob/c85f0f51634e65fe332e8bbe18fbf0689eb44d0f/sdks/java/core/src/main/java/org/apache/beam/sdk/io/BoundedSource.java#L393-L397) `final` to avoid that it gets overridden and someone outputs anything different of -INF as a timestamp? [Of course this is a breaking change but at least in theory] or we should just ignore timestamps of `BoundedReader` to make the change have less impact?].

What made me discover this issue was to see the different behaviour of the  [`JavaReadViaImpulse` wrapping of the Source API](https://github.com/apache/beam/blob/c85f0f51634e65fe332e8bbe18fbf0689eb44d0f/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/JavaReadViaImpulse.java#L139), maybe we should not output with timestamps there to be consistent everywhere?

Pinging @jkff in case he is around and have something to say too.

kennknowles:
Oh, I think what I said was not quite right. I think `BoundedSource.getCurrentTimestamp()` can be allowed to output other timestamps. But generally they don't bother since it is easy to use WithTimestamps on a PCollection where the watermark is -inf. In the SDF case you just don't have the same situation at all.

chamikaramj:
I'm not sure where the difference in behavior comes from but is it possible that input timestamps in two cases are different somehow (-inf vs current time) ?

@jkff or @lukecwik might have more context.

lukecwik:
@iemejia 
How do you resolve the case where you want to track the timestamp of the file as the output timestamp instead of the boundedsources timestamp. For example, the [Watch](https://github.com/apache/beam/blob/9428cc8c6fbffe7190c923850acbd98474210ad7/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/Watch.java#L754) transform sets the timestamp on the output. It would make sense that a user may want to use the timestamp there instead of the timestamp stored within the boundedsource. I think in most cases your change is the change that people want.

@chamikaramj 
It is unlikely that they are different for most sources but the [Watch](https://github.com/apache/beam/blob/9428cc8c6fbffe7190c923850acbd98474210ad7/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/Watch.java#L754) transform does meaningfully set it.

jkff:
I think we discussed this before, and concluded that reader.getCurrentTimestamp() ought to return a timestamp greater than timestamp of the element being read, which is consistent with watermark logic everywhere in Beam (including SDF): you can output forward in time but not backward.

Most BoundedSource's indeed output at -INF, which is inconsistent with the above. On the other hand, you could argue that they do so because they assume their source itself has a timestamp of -INF, and what they really mean is "output at the timestamp of my source"?

lukecwik:
@jkff, that makes sense when it is a regular DoFn but SplittableDoFns will be able to hold the output watermark independently of the input watermark. ReadAllViaFileBasedSource is a DoFn right now but I believe it should be treated as a SplittableDoFn because the DoFn is much closer to a BoundedSource -> SDF wrapper then a regular DoFn.

jkff:
@lukecwik What do you mean by "should be treated as a SplittableDoFn"?

lukecwik:
@jkff I would have wrote ReadAllViaFileBasedSource as a SplittableDoFn and not as regular DoFn. So in our discussion about what we should choose as the output time, I would have considered the fact that we can hold the output watermark and hence could choose either the timestamp of the element or the timestamp the source is emitting.

jkff:
I'm not sure I follow. Sounds like you're talking about ProcessContext.updateWatermark()? It is true that this API is only implemented for SDFs (even though technically we could support it for regular DoFns too), but I don't see how it answers the question of this PR, which timestamp should we output with. That API still does not make outputting backwards in time be permitted.

Perhaps the correct thing to do is to make BoundedSource.getCurrentTimestamp be `@Nullable`, where null means "I don't know the timestamp". Then if it's null then use the source's timestamp, else use the provided timestamp. That seems better than the current treatment, where BoundedSource by default pretends that all its elements are actually infinitely old.

lukecwik:
Yes, I'm talking about ProcessContext.updateWatermark().

I believe you should be able to output backwards in time and depending on
what the watermark is, the data can be considered late and possibly dropped
(alternative suggested in https://issues.apache.org/jira/browse/BEAM-644).
Since the SDF can control the watermark, the SDF could hold the watermark
at -inf and output at arbitrary timestamps. Note that this isn't practical
in streaming pipelines with an unbounded input PCollection unless the
timestamps the BoundedSource is reporting is correlated to the watermark.
So I believe that:
Your suggestion to have getCurrentTimestamp return "unknown" makes a lot of
sense and that we should always use the timestamp of the BoundedSource and
fallback to the element timestamp if it is unknown.

But what should we be updating the watermark to?
* smallest timestamp in the bounded source?
* largest timestamp in the bounded source?
* ...?

jkff:
But there's no API for holding the watermark: on the contrary, updateWatermark() allows the watermark to progress *faster* than it otherwise would, not slower - e.g. it allows the watermark to progress gradually as parts of the restriction are processed, as opposed to having it stay at (element timestamp) and then jump to +INF only when the whole restriction is done.

In the current PR, I think we shouldn't deal with watermark at all. Just make getCurrentTimestamp able to return "unknown" and fall back to element timestamp if unknown.

iemejia:
I think I got lost in this discussion because it seems to refer discussions I am not aware of. Can someone please help me summarize what is going on here so I can better document it and/or try to change the changes (if this is because an issue). Thanks.

lukecwik:
I haven't been able to get back to this yet but have been meaning to
explain what I mean.

lukecwik:
In the Beam model, the watermark propagates forward since the input watermark >= output watermark of each transform. The input watermark is a bound saying that all data that is before me will be considered late while the output watermark says that all data that is output before me will be considered late. So lets say T1 provides data to T2, then
```
... >= input_watermark(T1) >= output_watermark(T1) >= input_watermark(T2) >= ouptut_watermark(T2) >= ...
```
Note that the input/output watermarks are computed by the runner and aren't ever exposed to the SDK. Whether something is late or ontime is figured out as part of the GroupByKey. The runner also automatically advances the output watermark when all the data for a given input watermark has been consumed (watermark holds which is a special timer that isn't meant to be exposed to SDKs and will be modeled properly with the resolution of https://issues.apache.org/jira/browse/BEAM-2535 does allow the SDK to hold back the watermark).

This seems all great but what about the roots of the pipeline. This is where the `UnboundedSource` has the ability to report what it thinks the watermark should be and the runner computes the watermark by taking the min across all UnboundedSource instances for a particular root.

Another property of the Beam model is that if you output elements with a timestamp that is before the output watermark, the data will be classified as late (and depending on the trigger, may be dropped).

Combining these two properties of the Beam model, we get to the issue of what timestamp we should output with.

My concern is that the proposed solution of using getCurrentTimestamp and if it is unknown fallback to using the element timestamp can lead to data being marked late by default in the case where ReadAllViaFileBasedSource is used in a streaming pipeline.

The reason why I suggest to model this as an SDF is that I believe SDFs should be able to hold the output watermark just like an UnboundedSource by invoking ProcessContext.updateWatermark() which would then allow for ReadAllViaFileBasedSource to output data without any of it being marked late.

Eugene believes that ProcessContext.updateWatermark() should only be used to advance the output watermark faster then the input and to not be able to hold the output watermark and hence the data could be marked as late.

jkff:
I see UnboundedSource, regular DoFn and SDF as the same thing and I think they should behave identically w.r.t. watermarks (doesn't mean they should all support the same capabilities, but there should be no contradictions between the capabilities they *do* support).

All three take an input and produce outputs.

* In all three, produced outputs should be considered late if they're before the output watermark.
* In all three, by default the output watermark equals timestamp of the input element (this is not entirely accurate but I don't remember enough to state this entirely accurately) representing the principle that you can't output back in time. For the purposes of this statement, the "input element" of UnboundedSource has timestamp -INF.
* We've considered cases where you *do* want to output back in time by a limited amount, and there was some JIRA about adding a watermark shift primitive, but we didn't get around to it. We kinda have DoFn.allowedLateness for that.
* In some cases, "timestamp of the input element" is too slow, that's why UnboundedSource and SDF allow you to advance the watermark faster. Regular DoFn's don't allow you to do that because they output sufficiently few things that it doesn't matter. I don't see SDF's updateWatermark() as a watermark hold - we use a watermark hold under the hood to signal to the runner that processing of the element isn't actually done, but if there was some way to signal that differently (e.g. if a runner didn't use timers to schedule continuations) then a hold would not be necessary and updateWatermark() would only serve its true purpose of advancing the output watermark before the element is done.

lukecwik:
The trouble I have is that advancing the output watermark for the specific element+restriction doesn't really provide much since the runner needs to be able to maintain that the `output watermark <= input watermark` since it is the only thing that knows what future element+restriction may be produced. In the case when the input to a SplittableDoFn is bounded, once all the elements+restrictions are being processed, the runner can then use the min watermark of element+restrictions as the output watermark.

lukecwik:
Sorry for taking so long but I think in this case we should always output `max(element timestamp, reader timestamp)` and not update the BoundedSource API to return `unknown`.

This way ReadAllViaFileBasedSource will output the current reader timestamp in the bounded case since the input will be `-INF` and in the streaming case the user will have to make sure that the input timestamp is always <= the timestamp of all the records stored in the BoundedSource.

lukecwik:
@iemejia Was I able to give you the appropriate feedback on how to proceed with this change or do you have additional questions?

PR #8767, opened 2019-06-05
[BEAM-7493] Use specified archivesBaseName for project dependencies

kennknowles:
Some projects choose a custom archive name, but we depend on a computed incorrect name.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy
+++ b/buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy
@@ -87,6 +87,9 @@ class BeamModulePlugin implements Plugin<Project> {
     /** Controls whether the dependency analysis plugin is enabled. */
     boolean enableStrictDependencies = false
 
+    /** Override the default "beam-" + `dash separated path` archivesBaseName. */
+    String archivesBaseName = null;
+
     /**
      * List of additional lint warnings to disable.
      * In addition, defaultLintSuppressions defined below
@@ -142,6 +145,9 @@ class BeamModulePlugin implements Plugin<Project> {
      * By default we exclude any class underneath the org.apache.beam namespace.
      */
     List<String> shadowJarValidationExcludes = ["org/apache/beam/**"]
+
+    /** Override the default "beam-" + `dash separated path` archivesBaseName. */
+    String archivesBaseName = null;
   }
 
   // A class defining the set of configurable properties for createJavaExamplesArchetypeValidationTask
@@ -299,8 +305,8 @@ class BeamModulePlugin implements Plugin<Project> {
     return project.hasProperty('isRelease')
   }
 
-  def archivesBaseName(Project p) {
-    'beam' + p.path.replace(':', '-')
+  def defaultArchivesBaseName(Project p) {
+    return 'beam' + p.path.replace(':', '-')
   }
 
   void apply(Project project) {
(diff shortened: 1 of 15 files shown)


kennknowles:
@angoenka @robertwb @tweise @boyuanzz 

Failure across both portability runs:

```
:sdks:java:container:docker FAILED
--
ADD failed: stat /var/lib/docker/tmp/docker-builder527487348/target/beam-sdks-java-harness.jar: no such file or directory
```

pinging since you might have a quick answer to what has to happen for this archive name

kennknowles:
Yes, it seems that project already has the property set to `harness` not `beam-sdks-java-harness`.

kennknowles:
I see; it is a built-in thing that we are deliberately overriding. That makes this a bit more work but still probably OK. We just have two places with different opinions about how to override and need them to do things in different ways.

lukecwik:
The Docker container boot.go expects a specific file name for the jar. The harness jar artifact is renamed here to give it that stable name: https://github.com/apache/beam/blob/e6104a2fcb2bd8e25e06d06428cb7e6223d30083/sdks/java/container/build.gradle#L56

We could avoid this issue in the future if there was a way to use a project property that represented the jar path instead of pattern matching.

kennknowles:
OK I've turned this into a hack. Confirmed that the built jar from the harness has the expected name and the nexmark pom has the expected contents.

lukecwik:
Wouldn't it be better to just update the rename rule instead of insert the hack?

kennknowles:
Actually pretty much every jar was wrong with my initial change, because everything already has `archivesBaseName` set by gradle. My preference would be to decouple the archivesBaseName from the publishing configuration. But if we really want convention over configuration then we just use a different property to override it in the modules that want to set it. That way the BeamModulePlugin can always override the gradle defaults, and the modules can override the BeamModulePlugin defaults.

lukecwik:
Ping me when this is ready for review.

kennknowles:

kennknowles:
Run Python PreCommit

kennknowles:
This is ready for review.

kennknowles:
(As a hack - demonstrated that it is green. If you want to block on doing a non-hack then I welcome the feedback.)

lukecwik:
The suggestion would be to have applyJavaNature take an archivesBaseName and if none is specified then to setup the default. This way we aren't guessing the intent based upon whether the name contains "beam" or not.

kennknowles:
OK I can go ahead with that

buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy
```diff
@@ -1485,9 +1497,14 @@ class BeamModulePlugin implements Plugin<Project> {
     project.ext.applyPortabilityNature = {
       PortabilityNatureConfiguration configuration = it ? it as PortabilityNatureConfiguration : new PortabilityNatureConfiguration()
 
+      if (configuration.archivesBaseName) {
```
lukecwik: nit: Shouldn't this be automatically handled by applyJavaNature already?

kennknowles: Yes, it is redundant here. I wanted to localize this line to where the PortabilityConfiguration is processed. My thinking was to keep it somewhat independent of the behavior of applyJavaNature, or the fact that applyPortabilityNature depends on applyJavaNature. So the rest of either of those methods can change but this will not regress.

PR #8775, opened 2019-06-06
[BEAM-7478] Detect class path from the "java.class.path" property

mxm:
The current approach to assume the classloader is a UrlClassloader is
fundamentally flawed because:

- it relies on an implementation detail (cast to UrlClassloader)
- UrlClassloader does not exist in Java >= 9
- UrlClassloader#getUrls() returns jars belonging to the JRE which leads to
  errors like the following:

```
Caused by: java.lang.IllegalAccessError: class sun.reflect.GeneratedSerializationConstructorAccessor70 cannot access its superclass sun.reflect.SerializationConstructorAccessorImpl
```

I've tested the approach here with a remote Flink cluster and got rid of the
above error message.

Apart from fixing the issue present, the changes will also make the jar
detection code compatible with Java >= 9.

--- | --- | --- | --- | ---

### DIFF
--- a/runners/core-construction-java/src/test/java/org/apache/beam/runners/core/construction/PipelineResourcesTest.java
+++ b/runners/core-construction-java/src/test/java/org/apache/beam/runners/core/construction/PipelineResourcesTest.java
@@ -23,56 +23,43 @@ import static org.junit.Assert.assertThrows;
 
 import java.io.File;
 import java.io.IOException;
-import java.net.URL;
-import java.net.URLClassLoader;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import org.apache.beam.vendor.guava.v20_0.com.google.common.collect.ImmutableList;
+import org.junit.AfterClass;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
 import org.junit.rules.TemporaryFolder;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
-import org.mockito.Mockito;
 
 /** Tests for PipelineResources. */
 @RunWith(JUnit4.class)
 public class PipelineResourcesTest {
 
+  private static final String classPath = System.getProperty("java.class.path");
+
   @Rule public transient TemporaryFolder tmpFolder = new TemporaryFolder();
   @Rule public transient ExpectedException thrown = ExpectedException.none();
+
+  @AfterClass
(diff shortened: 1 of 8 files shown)


runners/core-construction-java/src/test/java/org/apache/beam/runners/core/construction/PipelineResourcesTest.java
```diff
@@ -23,56 +23,54 @@
...
 import org.junit.rules.TemporaryFolder;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
-import org.mockito.Mockito;
 
 /** Tests for PipelineResources. */
 @RunWith(JUnit4.class)
 public class PipelineResourcesTest {
 
+  private static final String classPath = System.getProperty("java.class.path");
```
lukecwik: Add test rule [RestoreSystemProperties](https://github.com/apache/beam/blob/master/sdks/java/core/src/test/java/org/apache/beam/sdk/testing/RestoreSystemProperties.java)

mxm: It's part of the test scope in java-core. Maybe move the rule to `test-utils`?

mxm:
Sounds reasonable to continue using the `UrlClassloader` approach but have the System property as a fallback. I'm not sure either if application servers set the class path property, but it seems to work reliably for deployments which start from the `java` command.

I'm a bit worried though that I'm seeing the `IllegalAccessError` with the UrlClassloader. Do you think we could come up with a way to filter out JRE components? Shipping those libraries does not seem right.

runners/core-construction-java/src/test/java/org/apache/beam/runners/core/construction/PipelineResourcesTest.java
lukecwik: Sounds like a good idea.

It would be good if we could move as many of the files in https://github.com/apache/beam/tree/master/sdks/java/core/src/test/java/org/apache/beam/sdk/testing over to it.

lukecwik:
Can you provide a stack trace for the IllegalAccessError?

mxm:
@lukecwik Full stack trace is in the JIRA issue: https://issues.apache.org/jira/browse/BEAM-7478

lukecwik:
The stack trace didn't give me any insights on how to make this code better then what I had suggested before.

PR #8778, opened 2019-06-06
Importing hamcrest dependency at matcher runtime

pabloem:
This prevents us from importing hamcrest dependency every time we run Beam.

### DIFF
--- a/sdks/python/apache_beam/testing/util.py
+++ b/sdks/python/apache_beam/testing/util.py
@@ -25,9 +25,6 @@ import io
 import tempfile
 from builtins import object
 
-from hamcrest.core import assert_that as hamcrest_assert
-from hamcrest.library.collection import contains_inanyorder
-
 from apache_beam import pvalue
 from apache_beam.transforms import window
 from apache_beam.transforms.core import Create
@@ -154,6 +151,8 @@ def matches_all(expected):
       the elements of a single PCollection.
   """
   def _matches(actual):
+    from hamcrest.core import assert_that as hamcrest_assert
+    from hamcrest.library.collection import contains_inanyorder
     expected_list = list(expected)
 
     hamcrest_assert(actual, contains_inanyorder(*expected_list))


pabloem:
r: @aaltay

sdks/python/apache_beam/testing/util.py
```diff
@@ -154,6 +151,8 @@ def matches_all(expected):
       the elements of a single PCollection.
   """
   def _matches(actual):
+    from hamcrest.core import assert_that as hamcrest_assert
```
aaltay: Would this break users if they use assert and they are not aware that they need to pre-install hamcrest as an extra dependency?

pabloem: yes. This would break them if they use this specific assertion, and did not install beam's test dependencies via `pip install apache-beam[test]`, or install hamcrest directly via `pip install pyhamcrest`

pabloem: The same would happen if users use other Beam utilities such as `bigquery_matcher.py`, or `pipeline_verifiers.py`

aaltay: `pipeline_verifiers.py` is a good argument. (Which btw. does not do dynamic loading)

Should we say that, if you are using testing/* module you need to install [test] extra packages? And also change pipeline_verifiers.py to use dynamic loading as well?

aaltay:
Just to avoid blocking people we can go ahead with this change. - LGTM in that sense.

But I would like us to actually figure out what a user needs to do use testing/*. Until this point the story for [test] extra package was for running unit tests of beam. Which most users does not need and it has heavy dependencies like pandas and numpy. This change shifts things to, require using hamcrest for using testing/*. (Actually it was already like this as you have shown with pipeline_verifiers.)

Perhaps we can add hamcrest as a core dependency.

pabloem:
Thanks. I'll work on an extra change to improve documentation - but yes, worth considering what we should do. Pipeline verifiers is tricky to do dynamic loading as the classes defined in it depend on hamcrest.

aaltay:
Thank you. Please file a JIRA so that we can track.

pabloem:
Filed https://issues.apache.org/jira/browse/BEAM-7509

PR #8795, opened 2019-06-07
[BEAM-7510] Fixing fileio tests checking JSON serialization (flaky)

pabloem:
The test was performing checking of `file.read() == json.dumps(someobject)` - which would be flaky due to different ordering of keys in a dictionary when serializing to JSON.

What it does now is `json.loads(file.read()) == someobject`


pabloem:
@tvalentyn it was a silly issue with non-deterministic json serialization.

PR #8797, opened 2019-06-07
[BEAM-7511] Fixes the bug in KafkaTable Initialization.

riazela:
Initialized configUpdates to an empty HashMap in the constructor and removed unused methods. 
------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/kafka/BeamKafkaTable.java
+++ b/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/kafka/BeamKafkaTable.java
@@ -46,6 +46,10 @@ public abstract class BeamKafkaTable extends BaseBeamTable {
   private List<TopicPartition> topicPartitions;
   private Map<String, Object> configUpdates;
 
+  protected BeamKafkaTable(Schema beamSchema) {
+    super(beamSchema);
+  }
+
   public BeamKafkaTable(Schema beamSchema, String bootstrapServers, List<String> topics) {
     super(beamSchema);
     this.bootstrapServers = bootstrapServers;
--- a/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/kafka/BeamKafkaTable.java
+++ b/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/kafka/BeamKafkaTable.java
@@ -53,6 +53,18 @@ public abstract class BeamKafkaTable extends BaseBeamTable {
     this.configUpdates = new HashMap<>();
   }
 
+  public BeamKafkaTable(
+      Schema beamSchema, List<TopicPartition> topicPartitions, String bootstrapServers) {
+    super(beamSchema);
+    this.bootstrapServers = bootstrapServers;
+    this.topicPartitions = topicPartitions;
+  }
+
+  public BeamKafkaTable updateConsumerProperties(Map<String, Object> configUpdates) {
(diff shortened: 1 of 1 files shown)


akedin:
R: @XuMingmin @apilloud @amaliujia

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/kafka/BeamKafkaTable.java
```diff
@@ -45,26 +46,11 @@
...
-  }
-
-  public BeamKafkaTable(
-      Schema beamSchema, List<TopicPartition> topicPartitions, String bootstrapServers) {
-    super(beamSchema);
-    this.bootstrapServers = bootstrapServers;
-    this.topicPartitions = topicPartitions;
-  }
-
-  public BeamKafkaTable updateConsumerProperties(Map<String, Object> configUpdates) {
```
apilloud: This method is public, is it really unused? This takes away the ability to set the `withConsumerConfigUpdates` field. If it is really unused lets delete `configUpdates` it all together.

akedin: The only way to call it is by calling `buildBeamSqlTable()` on a `KafkaTableProvider`, then casting it to correct class, and then calling this method. I cannot find any code in Beam that does it. And no code should ever be doing that in the first place. My assumption is it was used in tests similar to `getBootstrapServers()` in the same class.

+1 to removing the `configUpdates`

apilloud: I think it is quite possible that there exists external (possibly not opensource) code that extends `KafkaTableProvider` and overloads `buildBeamSqlTable()` for the purposes of adding features not in Beam. I've written things like this in the past.

Even if that is the case I don't think supporting manipulating a single field is a reasonable approach. We should still delete this. (I think it would be reasonable to move the construction of the KafkaIO reader and writer into seperate functions that can be overloaded, but that doesn't need to be part of this change.)

XuMingmin: I do have my own `KafkaTableProvider` to handle customized auth settings. Instead of delete the methods in `KafkaTableProvider`, I would prefer to support it in `KafkaTableProvider`.

riazela: I returned the deleted methods back so that we can support it later in `KafkaTableProvider`.

akedin: @XuMingmin thank you for pointing it out. I think exposing these methods in `KafkaTableProvider` is probably out of scope of this PR. These methods also would not be used in the existing code. If you rely on these in some customized version, can you contribute some of it back with tests and maybe examples? So that it doesn't get deleted by accident?

XuMingmin: surely, create https://issues.apache.org/jira/browse/BEAM-7522 to track.

akedin: Thanks. Will merge this PR then if there are no further concerns

PR #8801, opened 2019-06-07
Make temp_location an attribute of the StandardOptions class

ostrokach:
Currently, the `--temp_location` argument is parsed by the `GoogleCloudOptions` class. However, `--temp_location` or a similar argument may be useful for other runners as well.

For example, in the case of `DirectRunner`, `temp_location` may be the place where the runner stores its temporary files, if it ever supports out of process shuffles, combines, etc. In the case of `InteractiveRunner`, `temp_location` may be a good default location for the runner to store its cache files. In the case of the `WriteToFiles` transform, setting `temp_location` to `p.options.view_as(GoogleCloudOptions).temp_location` makes it appear as if `temp_location` has to be on GCS, which from my understanding, is not the case unless we are using the `DataflowRunner`?

CC: @ivant @ananvay 

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/options/pipeline_options.py
+++ b/sdks/python/apache_beam/options/pipeline_options.py
@@ -590,7 +590,7 @@ class GoogleCloudOptions(PipelineOptions):
       super(GoogleCloudOptions, self).__setattr__(name, value)
 
   @deprecated(
-      since='2.16.0',
+      since='2.17.0',
       custom_message=(
           'GoogleCloudOptions.temp_location is deprecated since %since%. '
           'Please use StandardOptions.temp_location or '
@@ -602,7 +602,7 @@ class GoogleCloudOptions(PipelineOptions):
       return self.view_as(StandardOptions).temp_location
 
   @deprecated(
-      since='2.16.0',
+      since='2.17.0',
       custom_message=(
           'GoogleCloudOptions.temp_location is deprecated since %since%. '
           'Please use StandardOptions.temp_location or '
--- a/sdks/python/apache_beam/options/pipeline_options.py
+++ b/sdks/python/apache_beam/options/pipeline_options.py
@@ -593,7 +593,8 @@ class GoogleCloudOptions(PipelineOptions):
       since='2.16.0',
       custom_message=(
           'GoogleCloudOptions.temp_location is deprecated since %since%. '
-          'Use GoogleCloudOptions.gcp_temp_location instead.'))
+          'Please use StandardOptions.temp_location or '
+          'GoogleCloudOptions.gcp_temp_location, as appropriate.'))
   def _get_temp_location(self):
     if self.gcp_temp_location is not None:
       return self.gcp_temp_location
@@ -604,7 +605,8 @@ class GoogleCloudOptions(PipelineOptions):
       since='2.16.0',
       custom_message=(
           'GoogleCloudOptions.temp_location is deprecated since %since%. '
-          'Use GoogleCloudOptions.gcp_temp_location instead.'))
+          'Please use StandardOptions.temp_location or '
+          'GoogleCloudOptions.gcp_temp_location, as appropriate.'))
   def _set_temp_location(self, temp_location):
     self.gcp_temp_location = temp_location
 
--- a/sdks/python/apache_beam/options/pipeline_options.py
+++ b/sdks/python/apache_beam/options/pipeline_options.py
@@ -41,6 +41,7 @@ from apache_beam.options.value_provider import StaticValueProvider
 from apache_beam.options.value_provider import ValueProvider
 from apache_beam.transforms.display import HasDisplayData
 from apache_beam.utils import processes
+from apache_beam.utils.annotations import deprecated
 
 __all__ = [
     'PipelineOptions',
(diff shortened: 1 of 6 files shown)


ostrokach:
retest this please

aaltay:
Should we have a gcp_temp_location which defaults to temp_location (as in Java : https://github.com/apache/beam/blob/39dab79b7bd56a47c0f2a02033ab4ca3bd4a67e2/sdks/java/extensions/google-cloud-platform-core/src/main/java/org/apache/beam/sdk/extensions/gcp/options/GcpOptions.java#L245)

IIRC @udim had some taught on this.

ostrokach:
retest this please

ostrokach:
retest this please

ostrokach:
retest this please

ostrokach:
@aaltay @udim I updated the PR to include both `--temp_location` and `--gcp_temp_location`. However, I am not sure if the division of concerns between those to options is the same as it is in the Java SDK?

I would think that most people would continue to use `--temp_location` even when using GCP, unless they want to separate their Dataflow temp files from the SDK temp files?

aaltay:
Another issues with this change is that, it is backward incompatible to move an option from its current class. This will break user pipelines that are setting these options in their code.

Unless there is a workaround we may want to delay this until beam 3.0.

@ostrokach are you blocked on this change?

robertwb:
Is it possible to have this option settable/gettable from both places? In
the end, they're just wrappers around a common dictionary.

ostrokach:
@aaltay I am not blocked on this change; I just figured that it is strange to use one `GoogleCloudOptions` for things that might not be GCloud-related, like cache location. But there are things around it, like having a separate `--cache_location` argument for example.

@robertwb The `PipelineOptions` class [collects the arguments defined by every subclass](https://github.com/apache/beam/blob/7fe54a0e178acf6957a797ac17edc4ec74e4bd42/sdks/python/apache_beam/options/pipeline_options.py#L179) into a single `_BeamArgumentParser`, which is basically just an `argparse.ArgumentParser`. We could over-write the [`add_argument`](https://github.com/python/cpython/blob/3.7/Lib/argparse.py#L1322) method of `_BeamArgumentParser` in order to skip duplicate arguments, but there does not seem to be a clean API for doing this without relying on "private" attributes, and then there are edge cases where maybe the type or the destination of the new argument is different from that of the old, and those cases may end up being hard to debug.

robertwb:
We could also just add a property to GoogleCloudOptions that gets/sets the
value in the underlying dict (possibly with a deprecation warning) for
backwards compatibility. It would only be parsed by StandardOptions.

aaltay:
@robertwb suggestion sounds good. We can change it so that it will be

StandardOptions.temp_location 
GcpOptions.gcp_temp_location (defaulting to StandardOptions.temp_location)

Also, how does this options, their defaults, and which ones are being required works in Java SDK?

ostrokach:
Run Portable_Python PreCommit

ostrokach:
@aaltay @robertwb I updated the PR so that:

- Move `temp_location` attribute to `StandardOptions`.
- Calling `GoogleCloudOptions.temp_location` falls back to `StandardOptoins.temp_locatioin` and emits a deprecation warning.
- `GoogleCloudOptions.gcp_temp_location` is used for Dataflow related tasks requiring a GCS bucket. If `GoogleCloudOptions.gcp_temp_location` is not provided, we fall back to `StandardOptoins.temp_locatioin`.

sdks/python/apache_beam/options/pipeline_options.py
```diff
@@ -500,13 +506,52 @@ def _add_argparse_args(cls, parser):
...
+    if name in ["temp_location"]:
+      self._set_temp_location(value)
+    else:
+      super(GoogleCloudOptions, self).__setattr__(name, value)
+
+  @deprecated(
+      since='2.16.0',
+      custom_message=(
+          'GoogleCloudOptions.temp_location is deprecated since %since%. '
+          'Use GoogleCloudOptions.gcp_temp_location instead.'))
```
robertwb: Possibly also suggest StandardOptions.temp_location?

sdks/python/apache_beam/options/pipeline_options.py
```diff
@@ -500,13 +506,52 @@ def _add_argparse_args(cls, parser):
                         choices=['COST_OPTIMIZED', 'SPEED_OPTIMIZED'],
                         help='Set the Flexible Resource Scheduling mode')
 
+  def __getattr__(self, name):
```
robertwb: Suggestion: consider using `@property` here.

sdks/python/apache_beam/options/pipeline_options.py
ostrokach: I updated the message to read:

```
          'GoogleCloudOptions.temp_location is deprecated since %since%. '
          'Please use StandardOptions.temp_location or '
          'GoogleCloudOptions.gcp_temp_location, as appropriate.'))
```

Does that sound better?

sdks/python/apache_beam/options/pipeline_options.py
ostrokach: That was my first thought as well, but I couldn't figure out how to get it to work.

The problem is that attribute access of the `GoogleCloudOptions` class ends up being performed using `PipelineOptions.__getattr__` and `PipelineOptions.__setattr__`, and those only check if the attribute is set in the (child) classes `_add_argparse_args(cls, parser)` method, not if the (child) class actually has that attribute.

Since we can't have both `StandardOptions._add_argparse_args` and `GoogleCloudOptions._add_argparse_args` define the `temp_location` setting, I ended up patching the `GoogleCloudOptions.__getattr__` and `GoogleCloudOptions.__setattr__` methods to add an additional check for the `temp_location` attribute (and then call the parent `PipelineOptions.__getattr__` / `PipelineOptions.__setattr__`).

ostrokach:
retest this please

ostrokach:
Run Portable_Python PreCommit

robertwb:
@ostrokach Can you rebase this on the latest master?

ostrokach:
retest this please

ostrokach:
retest this please

PR #8822, opened 2019-06-11
[BEAM-7513] Adding Row Count for Bigquery Table

riazela:
[BEAM-7513] Methods are added to support BigQueryTable row count. Also, BeamCalciteTable can return Statistics so that calcite uses this row count for the optimization.
Integration tests are added for BigQueryTable row count method. 

R: @akedin 

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/BeamSqlTable.java
+++ b/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/BeamSqlTable.java
@@ -17,6 +17,8 @@
  */
 package org.apache.beam.sdk.extensions.sql;
 
+import org.apache.beam.sdk.extensions.sql.impl.BeamRowCountStatistics;
+import org.apache.beam.sdk.options.PipelineOptions;
 import org.apache.beam.sdk.schemas.Schema;
 import org.apache.beam.sdk.values.PBegin;
 import org.apache.beam.sdk.values.PCollection;
@@ -36,4 +38,9 @@ public interface BeamSqlTable {
 
   /** Get the schema info of the table. */
   Schema getSchema();
+
+  /** Estimates the number of rows or returns null if there is no estimation. */
+  default BeamRowCountStatistics getRowCount(PipelineOptions options) {
+    return BeamRowCountStatistics.UNKNOWN;
+  }
 }
--- a/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/BeamCalciteTable.java
+++ b/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/BeamCalciteTable.java
@@ -21,9 +21,11 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import org.apache.beam.sdk.extensions.sql.BeamSqlTable;
+import org.apache.beam.sdk.extensions.sql.impl.rel.BeamEnumerableConverter;
 import org.apache.beam.sdk.extensions.sql.impl.rel.BeamIOSinkRel;
 import org.apache.beam.sdk.extensions.sql.impl.rel.BeamIOSourceRel;
 import org.apache.beam.sdk.extensions.sql.impl.utils.CalciteUtils;
+import org.apache.beam.vendor.guava.v20_0.com.google.common.collect.ImmutableList;
 import org.apache.beam.vendor.guava.v20_0.com.google.common.collect.ImmutableMap;
 import org.apache.calcite.adapter.java.AbstractQueryableTable;
 import org.apache.calcite.linq4j.QueryProvider;
@@ -38,6 +40,8 @@ import org.apache.calcite.rel.type.RelDataTypeFactory;
 import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.schema.ModifiableTable;
 import org.apache.calcite.schema.SchemaPlus;
+import org.apache.calcite.schema.Statistic;
+import org.apache.calcite.schema.Statistics;
 import org.apache.calcite.schema.TranslatableTable;
 
 /** Adapter from {@link BeamSqlTable} to a calcite Table. */
@@ -59,6 +63,25 @@ public class BeamCalciteTable extends AbstractQueryableTable
   @Override
   public RelDataType getRowType(RelDataTypeFactory typeFactory) {
     return CalciteUtils.toCalciteRowType(this.beamTable.getSchema(), typeFactory);
+  }
+
+  @Override
+  public Statistic getStatistic() {
+    /*
+     Changing class loader is required for the JDBC path. It is similar to what done in
+     {@link BeamEnumerableConverter#toRowList} and {@link BeamEnumerableConverter#toEnumerable }.
+    */
+    final ClassLoader originalClassLoader = Thread.currentThread().getContextClassLoader();
+    try {
+      Thread.currentThread().setContextClassLoader(BeamEnumerableConverter.class.getClassLoader());
+      BeamRowCountStatistics beamStatistics =
+          beamTable.getRowCount(BeamEnumerableConverter.createPipelineOptions(pipelineOptions));
+      return beamStatistics.isUnknown()
+          ? Statistics.UNKNOWN
+          : Statistics.of(beamStatistics.getRowCount().doubleValue(), ImmutableList.of());
+    } finally {
+      Thread.currentThread().setContextClassLoader(originalClassLoader);
+    }
   }
 
   @Override
--- a/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/BeamSqlEnv.java
+++ b/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/BeamSqlEnv.java
@@ -38,6 +38,8 @@ import org.apache.beam.sdk.extensions.sql.meta.provider.ReadOnlyTableProvider;
 import org.apache.beam.sdk.extensions.sql.meta.provider.TableProvider;
 import org.apache.beam.sdk.extensions.sql.meta.provider.UdfUdafProvider;
 import org.apache.beam.sdk.extensions.sql.meta.store.InMemoryMetaStore;
+import org.apache.beam.sdk.options.PipelineOptions;
+import org.apache.beam.sdk.options.PipelineOptionsFactory;
 import org.apache.beam.sdk.transforms.Combine.CombineFn;
 import org.apache.beam.sdk.transforms.SerializableFunction;
 import org.apache.beam.vendor.guava.v20_0.com.google.common.base.Strings;
@@ -66,14 +68,26 @@ public class BeamSqlEnv {
     return new BeamSqlEnvBuilder(tableProvider);
   }
 
+  /**
+   * This method creates {@link org.apache.beam.sdk.extensions.sql.impl.BeamSqlEnv} using empty
+   * Pipeline Options. It should only be used in tests.
+   */
   public static BeamSqlEnv readOnly(String tableType, Map<String, BeamSqlTable> tables) {
     return withTableProvider(new ReadOnlyTableProvider(tableType, tables));
   }
 
+  /**
+   * This method creates {@link org.apache.beam.sdk.extensions.sql.impl.BeamSqlEnv} using empty
+   * Pipeline Options. It should only be used in tests.
+   */
   public static BeamSqlEnv withTableProvider(TableProvider tableProvider) {
-    return builder(tableProvider).build();
+    return builder(tableProvider).setPipelineOptions(PipelineOptionsFactory.create()).build();
   }
 
+  /**
+   * This method creates {@link org.apache.beam.sdk.extensions.sql.impl.BeamSqlEnv} using empty *
+   * Pipeline Options. It should only be used in tests.
+   */
   public static BeamSqlEnv inMemory(TableProvider... tableProviders) {
     InMemoryMetaStore inMemoryMetaStore = new InMemoryMetaStore();
     for (TableProvider tableProvider : tableProviders) {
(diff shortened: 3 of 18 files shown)


sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/BeamSqlTable.java
```diff
@@ -36,4 +37,9 @@
 
   /** Get the schema info of the table. */
   Schema getSchema();
+
+  /** Estimates the number of rows or returns null if there is no estimation. */
+  default Double getRowCount(PipelineOptions options) {
```
akedin: Would it make sense to create our own version of `Statistic` for extensibility (only with row count for now)?

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/bigquery/BigQueryTable.java
```diff
@@ -39,13 +45,36 @@
...
 
   BigQueryTable(Table table, BigQueryUtils.ConversionOptions options) {
     super(table.getSchema());
     this.conversionOptions = options;
     this.bqLocation = table.getLocation();
   }
 
+  @Override
+  public Double getRowCount(PipelineOptions options) {
+    if (tableSize == -1.) {
```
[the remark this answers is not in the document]
akedin: This is also a magic constant that means "table size value has not been set". This is hard to realized without reading the whole code. The reason is because with `Double` you don't have a distinction between "value not set", "unknown", or any other value that is not an actual row count. If you had a custom class, e.g. `BeamTableStatistic`, you could have a default value there similar to `Statistic.UNKNOWN`. This way this code would look like:

```
if (statistic == null) {
  statistic = getRowCountFromBQ(...)
}
return statistic;
```

and would help readability a bit if you move the logic into a separate method:

```
private static BeamTableStatistic getRowCountFromBQ(PipelineOptions o, String bqLocation) {
   try {
      BigInteger rowCount = BigQueryHelpers.getNumRows(...);
      return BeamTableStatistic.of(rowCount);
   }  catch (...) {
      return BeamTableStatistic.unknown();
   }
}
```

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/BeamCalciteTable.java
```diff
@@ -61,6 +68,39 @@ public RelDataType getRowType(RelDataTypeFactory typeFactory) {
...
   }
 
+  @Override
+  public Statistic getStatistic() {
+    return new Statistic() {
+      @Override
+      public Double getRowCount() {
+        Double rc =
+            beamTable.getRowCount(BeamEnumerableConverter.createPipelineOptions(pipelineOptions));
+        rc = (rc == null) ? Statistics.UNKNOWN.getRowCount() : rc;
```
akedin: If you had a custom statistic class, this would look something like:
```
return statistic.isUnknown()
  ? Statistics.UNKNOWN
  : Statistics.of(statistic.getRowCount(), ImmutabelList.of());
```

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/BeamSqlTable.java
akedin: Also why Double?

sdks/java/extensions/sql/src/test/java/org/apache/beam/sdk/extensions/sql/meta/provider/bigquery/BigQueryRowCountIT.java
```diff
@@ -0,0 +1,114 @@
...
+
+    BeamSqlRelUtils.toPCollection(pipeline, sqlEnv.parseQuery(insertStatement));
+    pipeline.run().waitUntilFinish(Duration.standardMinutes(5));
+
+    sqlTable = provider.buildBeamSqlTable(table);
+    Double size1 = sqlTable.getRowCount(TestPipeline.testingPipelineOptions());
+
+    assertNotNull(size0);
+    assertNotNull(size1);
+    assertEquals(1, size1 - size0, 0.1);
```
akedin: can you skip the `size0` logic? You have a separate test that verifies that initial size is zero.

akedin: also add couple of rows as well?

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/BeamSqlTable.java
riazela: The reason for double is because of calcite Statistic. It returns Double for row count.

akedin: I suggest we use `long` or something else that is exact for now. It will be straightforward to go to approximate types if we need to, but not other way around

akedin:
R: @apilloud @amaliujia @kennknowles @XuMingmin

amaliujia:
run `./gradlew spotlessApply` locally to fix `Spotless` check.

riazela:
It seems I was running spotlessApply only for the SQL package. I will fix it and push it.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/BeamCalciteTable.java
```diff
@@ -61,6 +65,15 @@ public RelDataType getRowType(RelDataTypeFactory typeFactory) {
     return CalciteUtils.toCalciteRowType(this.beamTable.getSchema(), typeFactory);
   }
 
+  @Override
+  public Statistic getStatistic() {
+    BeamRowCountStatistics beamStatistics =
+        beamTable.getRowCount(BeamEnumerableConverter.createPipelineOptions(pipelineOptions));
```
apilloud: This `pipelineOptions` is really `sqlShellOptions`. It is only populated in the JDBC path. For this path to work, you need to wrap this whole function in the proper class loader. Here is an example: https://github.com/apache/beam/blob/0d613080918675c5f511a20a6bb6bd3558875398/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamEnumerableConverter.java#L110

For the `SqlTransform.query` path you need to plumb through the pipeline options. When you do that it would be good to rename this field to clarify when each works.

See slide 15 and 16 for a block diagram of how these two paths work: https://docs.google.com/presentation/d/1ezeaxu7x4y6gnijkuQExl1XYQNWOYUjzsLAJcGhy7tQ/edit#slide=id.g33b013e8f5_1_23

riazela: Thank you for the comment and sending the slides.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/BeamCalciteTable.java
```diff
@@ -67,11 +67,17 @@ public RelDataType getRowType(RelDataTypeFactory typeFactory) {
...
   @Override
   public Statistic getStatistic() {
-    BeamRowCountStatistics beamStatistics =
-        beamTable.getRowCount(BeamEnumerableConverter.createPipelineOptions(pipelineOptions));
-    return beamStatistics.isUnknown()
-        ? Statistics.UNKNOWN
-        : Statistics.of(beamStatistics.getRowCount().doubleValue(), ImmutableList.of());
+    final ClassLoader originalClassLoader = Thread.currentThread().getContextClassLoader();
+    try {
+      Thread.currentThread().setContextClassLoader(BeamEnumerableConverter.class.getClassLoader());
```
akedin: Can you add a comment that points to `BeamEnumerableConverter` and explains that this is required for JDBC path?

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/BeamSqlEnv.java
```diff
@@ -133,6 +136,7 @@ private BeamSqlEnvBuilder(TableProvider tableProvider) {
       functionSet = new HashSet<>();
       autoLoadUdfs = false;
       autoLoadBuiltinFunctions = false;
+      pipelineOptions = PipelineOptionsFactory.create();
```
akedin: I think we should force that pipeline options are set. I would set it to null here and add a `checkNotNull(pipelineOptions)` in the `build()` method

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/BeamSqlEnv.java
```diff
@@ -194,14 +198,18 @@ public BeamSqlEnvBuilder setQueryPlannerClassName(String name) {
       return this;
     }
 
+    public void setPipelineOptions(PipelineOptions pipelineOptions) {
```
akedin: make it part of the builder? I.e. return `this` similar to other methods here, so that it can be used in a fluent call chain

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/BeamCalciteTable.java
riazela: Sure. I'll add it.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/JdbcDriver.java
```diff
@@ -141,17 +144,34 @@ public Connection connect(String url, Properties info) throws SQLException {
...
+      connection.setPipelineOptionsMap(getOptionsMap(options));
       return connection;
     } catch (SQLException e) {
       throw new RuntimeException(e);
     }
   }
+
+  /** Converts {@link PipelineOptions} to its map format. */
+  private static Map<String, String> getOptionsMap(PipelineOptions options) {
+    ObjectMapper objectMapper = new ObjectMapper();
```
akedin: make it a static constant?

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/JdbcDriver.java
```diff
@@ -141,17 +144,34 @@ public Connection connect(String url, Properties info) throws SQLException {
...
+    Map map = objectMapper.convertValue(options, Map.class);
+    map = (Map) map.get("options");
+    if (map == null) {
+      map = new HashMap();
+    }
+    Map<String, String> map2 = new HashMap<>();
+    for (Object entry : map.entrySet()) {
+      Map.Entry ent = (Map.Entry) entry;
+      map2.put(ent.getKey().toString(), ent.getValue().toString());
```
akedin: `getValue().toString()` is probably wrong. I think if you have a complex object there, it would be converted to a `Map<>` by the `ObjectMapper` and then `toString()` would return something like `{key=value}` which is not a valid JSON, so we won't be able to parse it back. What has to be done is we should use the same mapper to convert the value to JSON string instead, probably something along these lines:

```java
optionsMap = map.entrySet()
 .stream()
 .collect(toMap(MapEntry::getKey, e-> OBJECT_MAPPER.writeValueAsString(e.getValue())))
```

sdks/java/extensions/sql/src/test/java/org/apache/beam/sdk/extensions/sql/meta/provider/bigquery/BigQueryTestTableProvider.java
```diff
@@ -0,0 +1,68 @@
...
+  @Nullable
+  @Override
+  public Table getTable(String tableName) {
+    return tableSpecMap.get(tableName);
+  }
+
+  @Override
+  public BeamSqlTable buildBeamSqlTable(Table table) {
+    BeamSqlTable t = beamSqlTableMap.get(table.getLocation());
+    if (t == null) {
```
akedin: I think that guard-statement style here would look better, i.e. `if (t != null) { return null; } ` and no `else`. Reduces the nesting and makes easier to find matching braces.

akedin:
run java examples dataflow precommit

akedin:
run java_examples_dataflow precommit

sdks/java/extensions/sql/build.gradle
```diff
@@ -85,6 +85,7 @@ dependencies {
   shadow project(path: ":sdks:java:core", configuration: "shadow")
   shadow project(":sdks:java:extensions:join-library")
   shadow library.java.slf4j_api
+  shadow "com.fasterxml.jackson.core:jackson-core:2.9.9"
```
akedin: you can use `library.java.jackson_core`. It's defined [here](https://github.com/apache/beam/blob/a71bfda77df36aa1531f01533c372233cfba0dd9/buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy#L496)

akedin:
run sql postcommit

sdks/java/extensions/sql/build.gradle
akedin: ah, and please squash the commits

akedin:
run sql postcommit

akedin:
run sql postcommit

akedin:
run sql postcommit

akedin:
run sql postcommit

akedin:
run sql postcommit

akedin:
run sql postcommit

akedin:
run sql postcommit

lukecwik:
This is causing several postcommit failures for Dataflow/Flink/Spark/Direct runners.
Filed https://issues.apache.org/jira/browse/BEAM-7583 with details.

PR #8826, opened 2019-06-11
[BEAM-5148] Implement MongoDB IO for Python SDK

y1chi:

[BEAM-5148] Implement MongoDB IO for Python SDK
------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/io/mongodbio.py
+++ b/sdks/python/apache_beam/io/mongodbio.py
@@ -1,0 +1,364 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+"""This module implements IO classes to read and write data on MongoDB.
+
+
+Read from MongoDB
+-----------------
+:class:`ReadFromMongoDB` is a ``PTransform`` that reads from a configured
+MongoDB source and returns a ``PCollection`` of dict representing MongoDB
+documents.
+To configure MongoDB source, the URI to connect to MongoDB server, database
+name, collection name needs to be provided.
+
+Example usage::
+
+  pipeline | ReadFromMongoDB(uri='mongodb://localhost:27017',
+                             db='testdb',
+                             coll='input')
+
+
+Write to MongoDB:
+-----------------
+:class:`WriteToMongoDB` is a ``PTransform`` that writes MongoDB documents to
+configured sink, and the write is conducted through a mongodb bulk_write of
+``ReplaceOne`` operations. If the document's _id field already existed in the
+MongoDB collection, it results in an overwrite, otherwise, a new document
+will be inserted.
+
+Example usage::
+
+  pipeline | WriteToMongoDB(uri='mongodb://localhost:27017',
+                            db='testdb',
+                            coll='output',
+                            batch_size=10)
+
+
+No backward compatibility guarantees. Everything in this module is experimental.
+"""
+
+from __future__ import absolute_import
+
+from bson import objectid
+from pymongo import MongoClient
+from pymongo import ReplaceOne
+
+import apache_beam as beam
+from apache_beam.io import iobase
+from apache_beam.io.range_trackers import OffsetRangeTracker
+from apache_beam.transforms import DoFn
+from apache_beam.transforms import PTransform
+from apache_beam.transforms import Reshuffle
+from apache_beam.utils.annotations import experimental
+
+__all__ = ['ReadFromMongoDB', 'WriteToMongoDB']
+
+
+@experimental()
+class ReadFromMongoDB(PTransform):
+  """A ``PTransfrom`` to read MongoDB documents into a ``PCollection``.
+  """
+
+  def __init__(self,
+               uri='mongodb://localhost:27017',
+               db=None,
+               coll=None,
+               filter=None,
+               projection=None,
+               extra_client_params=None):
+    """Initialize a :class:`ReadFromMongoDB`
+
+    Args:
+      uri (str): The MongoDB connection string following the URI format
+      db (str): The MongoDB database name
+      coll (str): The MongoDB collection name
+      filter: A `bson.SON
+        <https://api.mongodb.com/python/current/api/bson/son.html>`_ object
+        specifying elements which must be present for a document to be included
+        in the result set
+      projection: A list of field names that should be returned in the result
+        set or a dict specifying the fields to include or exclude
+      extra_client_params(dict): Optional `MongoClient
+        <https://api.mongodb.com/python/current/api/pymongo/mongo_client.html>`_
+        parameters
+
+    Returns:
+      :class:`~apache_beam.transforms.ptransform.PTransform`
+
+    """
+    if extra_client_params is None:
+      extra_client_params = {}
+    if not isinstance(db, str):
+      raise ValueError('ReadFromMongDB db param must be specified as a string')
+    if not isinstance(coll, str):
+      raise ValueError('ReadFromMongDB coll param must be specified as a '
+                       'string')
+    self._mongo_source = _BoundedMongoSource(
+        uri=uri,
+        db=db,
+        coll=coll,
+        filter=filter,
+        projection=projection,
+        extra_client_params=extra_client_params)
+
+  def expand(self, pcoll):
+    return pcoll | iobase.Read(self._mongo_source)
+
+
+class _BoundedMongoSource(iobase.BoundedSource):
+  def __init__(self,
+               uri=None,
+               db=None,
+               coll=None,
+               filter=None,
+               projection=None,
+               extra_client_params=None):
+    if extra_client_params is None:
+      extra_client_params = {}
+    if filter is None:
+      filter = {}
+    self.uri = uri
+    self.db = db
+    self.coll = coll
+    self.filter = filter
+    self.projection = projection
+    self.spec = extra_client_params
+    self.doc_count = self._get_document_count()
+    self.avg_doc_size = self._get_avg_document_size()
+    self.client = None
+
+  def estimate_size(self):
+    return self.avg_doc_size * self.doc_count
+
+  def split(self, desired_bundle_size, start_position=None, stop_position=None):
+    # use document cursor index as the start and stop positions
+    if start_position is None:
+      start_position = 0
+    if stop_position is None:
+      stop_position = self.doc_count
+
+    # get an estimate on how many documents should be included in a split batch
+    desired_bundle_count = desired_bundle_size // self.avg_doc_size
+
+    bundle_start = start_position
+    while bundle_start < stop_position:
+      bundle_end = min(stop_position, bundle_start + desired_bundle_count)
+      yield iobase.SourceBundle(weight=bundle_end - bundle_start,
+                                source=self,
+                                start_position=bundle_start,
+                                stop_position=bundle_end)
+      bundle_start = bundle_end
+
+  def get_range_tracker(self, start_position, stop_position):
+    if start_position is None:
+      start_position = 0
+    if stop_position is None:
+      stop_position = self.doc_count
+    return OffsetRangeTracker(start_position, stop_position)
+
+  def read(self, range_tracker):
+    with MongoClient(self.uri, **self.spec) as client:
+      # docs is a MongoDB Cursor
+      docs = client[self.db][self.coll].find(
+          filter=self.filter, projection=self.projection
+      )[range_tracker.start_position():range_tracker.stop_position()]
+      for index in range(range_tracker.start_position(),
+                         range_tracker.stop_position()):
+        if not range_tracker.try_claim(index):
+          return
+        yield docs[index - range_tracker.start_position()]
+
+  def display_data(self):
+    res = super(_BoundedMongoSource, self).display_data()
+    res['uri'] = self.uri
+    res['database'] = self.db
+    res['collection'] = self.coll
+    res['filter'] = self.filter
+    res['project'] = self.projection
+    res['mongo_client_spec'] = self.spec
+    return res
+
+  def _get_avg_document_size(self):
+    with MongoClient(self.uri, **self.spec) as client:
+      size = client[self.db].command('collstats', self.coll).get('avgObjSize')
+      if size is None or size <= 0:
+        raise ValueError(
+            'Collection %s not found or average doc size is '
+            'incorrect', self.coll)
+      return size
+
+  def _get_document_count(self):
+    with MongoClient(self.uri, **self.spec) as client:
+      return max(client[self.db][self.coll].count_documents(self.filter), 0)
+
+
+@experimental()
+class WriteToMongoDB(PTransform):
+  """WriteToMongoDB is a ``PTransform`` that writes a ``PCollection`` of
+  mongodb document to the configured MongoDB server.
+
+  In order to make the document writes idempotent so that the bundles are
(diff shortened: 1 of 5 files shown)


y1chi:
Tested the basic functionalities with local docker setup.

y1chi:
retest this please

y1chi:
R: @chamikaramj  
Hi cham would you mind take a look on this initial implementation of mongodb io when you have time?

y1chi:
R: @ihji

sdks/python/apache_beam/io/mongodbio_test.py
```diff
@@ -0,0 +1,171 @@
...
+from apache_beam.io.mongodbio import _BoundedMongoSource
+from apache_beam.io.mongodbio import _GenerateObjectIdFn
+from apache_beam.io.mongodbio import _MongoSink
+from apache_beam.io.mongodbio import _WriteMongoFn
+from apache_beam.testing.test_pipeline import TestPipeline
+from apache_beam.testing.util import assert_that
+from apache_beam.testing.util import equal_to
+
+
+class Test_BoundedMongoSource(unittest.TestCase):
```
chamikaramj: class MongSourceTest

sdks/python/apache_beam/io/mongodbio_test.py
```diff
@@ -0,0 +1,171 @@
...
+
+  @mock.patch('apache_beam.io.mongodbio.MongoClient')
+  def test_get_document_count(self, mock_client):
+    mock_client.return_value.__enter__.return_value.__getitem__ \
+      .return_value.__getitem__.return_value.count_documents.return_value = 10
+
+    self.assertEqual(10, self.mongo_source._get_document_count())
+
+
+class TestReadFromMongoDB(unittest.TestCase):
```
chamikaramj: ReadFromMongDBTest (following convention of other tests)

sdks/python/apache_beam/io/mongodbio_test.py
```diff
@@ -0,0 +1,171 @@
...
+      }, {
+          'x': 2,
+          '_id': 123
+      }])
+                | "Generate ID" >> beam.ParDo(_GenerateObjectIdFn())
+                | "Check" >> beam.Map(lambda x: '_id' in x))
+      assert_that(output, equal_to([True] * 2))
+
+
+class Test_WriteMongoFn(unittest.TestCase):
```
chamikaramj: Ditto here and below (regarding test naming)

sdks/python/apache_beam/io/mongodbio_test.py
```diff
@@ -0,0 +1,171 @@
...
+  @mock.patch('apache_beam.io.mongodbio._MongoSink')
+  def test_process(self, mock_sink):
+    docs = [{'x': 1}, {'x': 2}, {'x': 3}]
```
chamikaramj: Consider adding an exhaustive test for dynamic work rebalancing similar to following (not entirely sure if this is possible with mocks).
https://github.com/apache/beam/blob/master/sdks/python/apache_beam/io/textio_test.py#L386

sdks/python/apache_beam/io/mongodbio.py
```diff
@@ -0,0 +1,292 @@
...
+
+  def get_range_tracker(self, start_position, stop_position):
+    if start_position is None:
+      start_position = 0
+    if stop_position is None:
+      stop_position = self.doc_count
+    return OffsetRangeTracker(start_position, stop_position)
+
+  def read(self, range_tracker):
+    if range_tracker.try_claim(range_tracker.start_position()):
```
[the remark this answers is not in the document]
y1chi: Thanks for clarifying this, I was confused about the try_claim document that says "This method should be called **if and only if** the record is at a split point. ". Is every element considered as a split poit?

sdks/python/apache_beam/io/mongodbio.py
```diff
@@ -0,0 +1,292 @@
...
+      uri (str): The MongoDB connection string following the URI format
+      db (str): The MongoDB database name
+      coll (str): The MongoDB collection name
+      filter: A `bson.SON
+        <https://api.mongodb.com/python/current/api/bson/son.html>`_ object
+        specifying elements which must be present for a document to be included
+        in the result set
+      projection: A list of field names that should be returned in the result
+        set or a dict specifying the fields to include or exclude
+      **kwargs: Optional `MongoClient
```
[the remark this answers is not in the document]
y1chi: make sense

sdks/python/apache_beam/io/mongodbio.py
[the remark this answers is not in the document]
chamikaramj: Depends on the source. For this source we can split at any document (count), right ? In that case we'll have to claim every integer. This is similar to CountingSource.
https://github.com/apache/beam/blob/38166b03d07a27e99633ba09e392a1258f9432ef/sdks/python/apache_beam/io/utils.py#L50

Also, we have to return if any record cannot be claimed.

sdks/python/apache_beam/io/mongodbio.py
```diff
@@ -0,0 +1,292 @@
...
+
+
+class _GenerateObjectIdFn(DoFn):
+  def process(self, element, *args, **kwargs):
+    # if _id field already exist we keep it as it is, otherwise the ptransform
+    # generates a new _id field to achieve idempotent write to mongodb.
+    result = element.copy()
+
+    if '_id' not in element:
+      result['_id'] = objectid.ObjectId()
```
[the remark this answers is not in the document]
y1chi: it's basically same function that mongodb server uses to generates id. From the document it is highly likely unique but not guaranteed in a distributed fashion I guess. However it is still suggested way to add _id in https://docs.mongodb.com/manual/reference/bson-types/#objectid.

sdks/python/apache_beam/io/mongodbio.py
```diff
@@ -0,0 +1,292 @@
...
+           | Reshuffle() \
+           | beam.ParDo(_WriteMongoFn(self._uri, self._db, self._coll,
+                                      self._batch_size, **self._spec))
+
+
+class _GenerateObjectIdFn(DoFn):
+  def process(self, element, *args, **kwargs):
+    # if _id field already exist we keep it as it is, otherwise the ptransform
+    # generates a new _id field to achieve idempotent write to mongodb.
+    result = element.copy()
```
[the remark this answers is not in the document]
y1chi: forgot the cleanup some old code.

sdks/python/apache_beam/io/mongodbio.py
```diff
@@ -0,0 +1,292 @@
...
+      return client[self.db][self.coll].count_documents(self.filter)
+
+
+@experimental()
+class WriteToMongoDB(PTransform):
+  def __init__(self,
+               uri='mongodb://localhost:27017',
+               db=None,
+               coll=None,
+               batch_size=1,
```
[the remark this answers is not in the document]
y1chi: changed to 100

sdks/python/apache_beam/io/mongodbio.py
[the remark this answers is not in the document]
y1chi: forgot to cleanup stale code.

sdks/python/apache_beam/io/mongodbio.py
```diff
@@ -0,0 +1,292 @@
...
+        set or a dict specifying the fields to include or exclude
+      **kwargs: Optional `MongoClient
+        <https://api.mongodb.com/python/current/api/pymongo/mongo_client.html>`_
+        parameters as keyword arguments
+
+    Returns:
+      :class:`~apache_beam.transforms.ptransform.PTransform`
+
+    """
+    self._mongo_source = _BoundedMongoSource(uri=uri,
```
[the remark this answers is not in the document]
y1chi: I added validation for db and coll, the rest are probably better handled by mongo client.

sdks/python/apache_beam/io/mongodbio.py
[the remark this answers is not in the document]
y1chi: changed to 100

sdks/python/apache_beam/io/mongodbio.py
```diff
@@ -0,0 +1,292 @@
...
+  def __init__(self,
+               uri='mongodb://localhost:27017',
+               db=None,
+               coll=None,
+               batch_size=1,
+               **kwargs):
+    """
+
+    Args:
+      uri (str): The MongoDB connection string following the URI format
```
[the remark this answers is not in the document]
y1chi: done

sdks/python/apache_beam/io/mongodbio.py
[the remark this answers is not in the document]
y1chi: done

sdks/python/apache_beam/io/mongodbio_test.py
y1chi: This appears to be a source test? I've added one accordingly to MongoSourceTest

y1chi:
Run Portable_Python PreCommit

y1chi:
Run Portable_Python PreCommit

y1chi:
@chamikaramj I think I addressed all the comments, would you mind take another look?

sdks/python/apache_beam/io/mongodbio.py
```diff
@@ -0,0 +1,358 @@
...
+    with MongoClient(self.uri, **self.spec) as client:
+      # docs is a MongoDB Cursor
+      docs = client[self.db][self.coll].find(
+          filter=self.filter, projection=self.projection
+      )[range_tracker.start_position():range_tracker.stop_position()]
+      for index in range(range_tracker.start_position(),
+                         range_tracker.stop_position()):
+        if not range_tracker.try_claim(index):
+          return
+        yield docs[index - range_tracker.start_position()]
```
chamikaramj: I think you should be yielding docs[index] here.
Otherwise same index will be included multiple times in the read iterator.

chamikaramj: Also, probably tests have to be adjusted to capture this (assuming this was a bug).

sdks/python/apache_beam/io/mongodbio.py
```diff
@@ -0,0 +1,358 @@
...
+    res['database'] = self.db
+    res['collection'] = self.coll
+    res['filter'] = self.filter
+    res['project'] = self.projection
+    res['mongo_client_spec'] = self.spec
+    return res
+
+  def _get_avg_document_size(self):
+    with MongoClient(self.uri, **self.spec) as client:
+      return client[self.db].command('collstats', self.coll)['avgObjSize']
```
chamikaramj: Will we be always get a valid size here ? 

Otherwise we should validate and pick a suitable default (for example if service return a negative value for some reason).

sdks/python/apache_beam/io/mongodbio.py
```diff
@@ -0,0 +1,358 @@
...
+
+  def _get_document_count(self):
+    with MongoClient(self.uri, **self.spec) as client:
+      return client[self.db][self.coll].count_documents(self.filter)
```
chamikaramj: Ditto about validating the value from service.

sdks/python/apache_beam/io/mongodbio.py
```diff
@@ -0,0 +1,358 @@
...
+  retry-able without creating duplicates, the PTransform added 2 transformations
+  before final write stage:
+  a ``GenerateId`` transform and a ``Reshuffle`` transform.::
+
+                  -----------------------------------------------
+    Pipeline -->  |GenerateId --> Reshuffle --> WriteToMongoSink|
+                  -----------------------------------------------
+                                  (WriteToMongoDB)
+
+  The ``GenerateId`` transform adds a random *_id* field to the documents if
```
chamikaramj: random and unique

sdks/python/apache_beam/io/mongodbio.py
```diff
@@ -0,0 +1,358 @@
...
+  they don't already have one, it uses the same format as MongoDB default. The
+  ``Reshuffle`` transform makes sure that no fusion happens between
+  ``GenerateId`` and the final write stage transform, so that when the
```
chamikaramj: , so that the set of documents and their unique IDs are not regenerated if final write step is retried due to a failure. This prevents duplicate writes of the same document with different unique IDs.

sdks/python/apache_beam/io/mongodbio_test.py
```diff
@@ -0,0 +1,187 @@
...
+  def setUp(self, mock_size, mock_count):
+    mock_size.return_value = 10
+    mock_count.return_value = 5
+    self.mongo_source = _BoundedMongoSource('mongodb://test', 'testdb',
+                                            'testcoll')
+
+  def test_estimate_size(self):
+    self.assertEqual(self.mongo_source.estimate_size(), 50)
+
+  def test_split(self):
```
chamikaramj: You should be able to use assert_sources_equal_reference_source() here.
https://github.com/apache/beam/blob/master/sdks/python/apache_beam/io/source_test_utils.py#L116

sdks/python/apache_beam/io/mongodbio_test.py
chamikaramj: Yes. Thanks.

sdks/python/apache_beam/io/mongodbio_test.py
```diff
@@ -0,0 +1,187 @@
...
+          2, mock_sink.return_value.__enter__.return_value.write.call_count)
+
+  def test_display_data(self):
+    data = _WriteMongoFn(batch_size=10).display_data()
+    self.assertEqual(10, data['batch_size'])
+
+
+class MongoSinkTest(unittest.TestCase):
+  @mock.patch('apache_beam.io.mongodbio.MongoClient')
+  def test_write(self, mock_client):
```
chamikaramj: Also add a tests to confirm that (1)we accept unique IDs that are already available if set (2) if unique IDs are not set we generate them before writing.

sdks/python/apache_beam/io/mongodbio.py
y1chi: I guess this is what I intended here, the `docs` is a mongodb cursor that can iterate through documents between start_position and stop_position. And individual element is accessed through the offset. 
for example if `docs` cover documents from position 30 to 60, then docs[0] represents the 30th document in mongodb, and docs[1] is the 31th.  
Using yield docs[index] could be causing index error since the index here represents the overall position.

sdks/python/apache_beam/io/mongodbio.py
y1chi: it may be hard to pick a suitable value, and could cause unwanted side effect, I can validate the value and throw exception if the value is unexpected.

sdks/python/apache_beam/io/mongodbio_test.py
y1chi: done

sdks/python/apache_beam/io/mongodbio.py
chamikaramj: Thanks. You are correct. I read this wrong.

sdks/python/apache_beam/io/mongodbio_test.py
y1chi: Added the cases in WriteToMongoDBTest.

y1chi:
Run Python PreCommit

y1chi:
Run RAT PreCommit

y1chi:
> LGTM. Thanks.
> 
> Please squash commits for merging according to https://beam.apache.org/contribute/

Thanks for reviewing!

chamikaramj:
Thanks. Merging.

ovaltzer:
hi there, quick question:
for some reason, after using this code in 2.14.0 we fail to run pipelines due to dependency issue with 
`    from bson.py3compat import abc, string_type, PY3, text_type
ImportError: cannot import name abc
`
do you have any idea what is the problem (we use python 2.7) ?

aaltay:
@ovaltzer - This seems to be related https://github.com/py-bson/bson/issues/82

Do you have a `bson` installed in your environment? If I am reading the above issue correctly `pymongo` has its own copy of `bson` and not compatible with the one installed from pypi.

/cc: @chamikaramj

ovaltzer:
Hey yes. Sorry I forgot to update but this was the issue. We needed to change our setup files to use pymongo. For some reason this this became mandatory dependency as part of 2.14

aaltay:
It is a good idea to update the documents.In addition we can try to do `bson` import optionally here (https://github.com/apache/beam/blob/master/sdks/python/apache_beam/io/mongodbio.py#L58) so at least user of just `bson` but not mongo db would not be affected of this issue.

PR #8853, opened 2019-06-13
Update portable schema representation and java SchemaTranslation

TheNeuralBit:
Also adds tests in SchemaTranslationTest.

Things that are *not* currently included in this PR:
* LogicalType registration and resolution by URN. We cannot decode a Schema with a logical type.
* Representing datetime and decimal as logical types (see [BEAM-7554](https://issues.apache.org/jira/browse/BEAM-7554)). Instead they are still represented as primitive types in Java's `Schema.FieldType` and they are mapped to the appropriate URNs when converting to/from the proto representation.

--- | --- | --- | --- | ---

R: @reuvenlax, @robertwb, @kennknowles

### DIFF
--- a/model/pipeline/src/main/proto/beam_runner_api.proto
+++ b/model/pipeline/src/main/proto/beam_runner_api.proto
@@ -648,19 +648,29 @@ message StandardCoders {
 // Experimental: A representation of a Beam Schema.
 message Schema {
   enum AtomicType {
-    BYTE = 0;
-    INT16 = 1;
-    INT32 = 2;
-    INT64 = 3;
-    FLOAT = 4;
-    DOUBLE = 5;
-    STRING = 6;
-    BOOLEAN = 7;
-    BYTES = 8;
+    UNSPECIFIED = 0;
+    BYTE = 1;
+    INT16 = 2;
+    INT32 = 3;
+    INT64 = 4;
+    FLOAT = 5;
+    DOUBLE = 6;
+    STRING = 7;
+    BOOLEAN = 8;
+    BYTES = 9;
   }
 
   message ArrayType {
     FieldType element_type = 1;
+  }
+
+  message MapType {
+    FieldType key_type = 1;
+    FieldType value_type = 2;
+  }
+
+  message RowType {
+    Schema schema = 1;
   }
 
   message LogicalType {
@@ -669,18 +679,13 @@ message Schema {
     FieldType representation = 3;
   }
 
-  message MapType {
-    FieldType key_type = 1;
-    FieldType value_type = 2;
-  }
-
   message FieldType {
     bool nullable = 1;
     oneof type_info {
       AtomicType atomic_type = 2;
       ArrayType array_type = 3;
       MapType map_type = 4;
-      Schema row_type = 5;
+      RowType row_type = 5;
       LogicalType logical_type = 6;
     }
   }
--- a/model/pipeline/src/main/proto/beam_runner_api.proto
+++ b/model/pipeline/src/main/proto/beam_runner_api.proto
@@ -647,29 +647,26 @@ message StandardCoders {
 
 // Experimental: A representation of a Beam Schema.
 message Schema {
-  enum TypeName {
+  enum AtomicType {
     BYTE = 0;
     INT16 = 1;
     INT32 = 2;
     INT64 = 3;
-    DECIMAL = 4;
-    FLOAT = 5;
-    DOUBLE = 6;
-    STRING = 7;
-    DATETIME = 8;
-    BOOLEAN = 9;
-    BYTES = 10;
-    ARRAY = 11;
-    MAP = 13;
-    ROW = 14;
-    LOGICAL_TYPE = 15;
+    FLOAT = 4;
+    DOUBLE = 5;
+    STRING = 6;
+    BOOLEAN = 7;
+    BYTES = 8;
+  }
+
+  message ArrayType {
+    FieldType element_type = 1;
   }
 
   message LogicalType {
-    string id = 1;
+    string urn = 1;
     string args = 2;
-    FieldType base_type = 3;
-    bytes serialized_class = 4;
+    FieldType representation = 3;
   }
 
   message MapType {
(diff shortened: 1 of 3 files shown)


model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -647,29 +647,26 @@ message StandardCoders {
 
 // Experimental: A representation of a Beam Schema.
 message Schema {
-  enum TypeName {
+  enum AtomicType {
```
kennknowles: Isn't 0 in an enum supposed to be reserved for unknown? It is wise, because of defaulting in proto libs.

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -678,12 +675,12 @@ message Schema {
...
-    TypeName type_name = 1;
-    bool nullable = 2;
+    bool nullable = 1;
     oneof type_info {
-      FieldType collection_element_type = 3;
+      AtomicType atomic_type = 2;
+      ArrayType array_type = 3;
       MapType map_type = 4;
-      Schema row_schema = 5;
+      Schema row_type = 5;
```
kennknowles: Proto best practice I _think_ is to go ahead and have a `RowType` message with one field. It has overhead, yes.

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/SchemaTranslation.java
```diff
@@ -25,33 +25,29 @@
...
+          .put(TypeName.INT32, RunnerApi.Schema.AtomicType.INT32)
+          .put(TypeName.INT64, RunnerApi.Schema.AtomicType.INT64)
+          .put(TypeName.FLOAT, RunnerApi.Schema.AtomicType.FLOAT)
+          .put(TypeName.DOUBLE, RunnerApi.Schema.AtomicType.DOUBLE)
+          .put(TypeName.STRING, RunnerApi.Schema.AtomicType.STRING)
+          .put(TypeName.BOOLEAN, RunnerApi.Schema.AtomicType.BOOLEAN)
+          .put(TypeName.BYTES, RunnerApi.Schema.AtomicType.BYTES)
           .build();
 
+  private static final String URN_BEAM_LOGICAL_DATETIME = "urn:beam:logical:datetime";
```
kennknowles: Two stylistic nits:

1. We tend to omit the `urn`, don't we? While it does make a valid URI out of the thing, it seems a bit silly.
2. I would leave out `logical` but put in something like `type` or `schema_type` or `fieldtype` to namespace.

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/SchemaTranslation.java
```diff
@@ -25,33 +25,29 @@
...
-          .put(TypeName.STRING, RunnerApi.Schema.TypeName.STRING)
-          .put(TypeName.DATETIME, RunnerApi.Schema.TypeName.DATETIME)
-          .put(TypeName.BOOLEAN, RunnerApi.Schema.TypeName.BOOLEAN)
-          .put(TypeName.BYTES, RunnerApi.Schema.TypeName.BYTES)
-          .put(TypeName.ARRAY, RunnerApi.Schema.TypeName.ARRAY)
-          .put(TypeName.MAP, RunnerApi.Schema.TypeName.MAP)
-          .put(TypeName.ROW, RunnerApi.Schema.TypeName.ROW)
-          .put(TypeName.LOGICAL_TYPE, RunnerApi.Schema.TypeName.LOGICAL_TYPE)
+
+  private static final BiMap<TypeName, RunnerApi.Schema.AtomicType> ATOMIC_TYPE_MAPPING =
```
kennknowles: TBH I find a `switch` clearer than a map lookup, and it takes the same amount of code space. Not for this PR, in which you are just editing the existing structure not restructuring.

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/SchemaTranslation.java
```diff
@@ -139,32 +150,43 @@ private static Field fieldFromProto(RunnerApi.Schema.Field protoField) {
   }
 
   private static FieldType fieldTypeFromProto(RunnerApi.Schema.FieldType protoFieldType) {
-    TypeName typeName = TYPE_NAME_MAPPING.inverse().get(protoFieldType.getTypeName());
     FieldType fieldType;
-    switch (typeName) {
-      case ROW:
-        fieldType = FieldType.row(fromProto(protoFieldType.getRowSchema()));
+    switch (protoFieldType.getTypeInfoCase()) {
```
kennknowles: Another not-for-this PR comment that this would be cleaner with the `switch` in a function so the branches could all `return`.

model/pipeline/src/main/proto/beam_runner_api.proto
lukecwik: Yes, to my knowledge specifically adding an UNSPECIFIED with a value of 0 will make this clearer.
For example: https://github.com/apache/beam/blob/eb5a8c2cbd973972491dedef23aff98e450a9f5a/model/pipeline/src/main/proto/beam_runner_api.proto#L451

lukecwik:
Is the beam_runner_api.proto the right place to put all the schema stuff?

model/pipeline/src/main/proto/beam_runner_api.proto
TheNeuralBit: Done.

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/SchemaTranslation.java
TheNeuralBit: Done. I went with fieldtype

model/pipeline/src/main/proto/beam_runner_api.proto
TheNeuralBit: Done. Agreed this is much cleaner

TheNeuralBit:
I guess I don't have a strong opinion, I was just updating it in place. Do you think it should get it's own schema.proto file?

kennknowles:
I think it would be great to have a separate `schema.proto` file. I wouldn't block merging on this. I would definitely like that to be a separate commit if you do add it here. Moving + editing in one commit would be bad form IMO.

TheNeuralBit:
Agreed. I can follow-up with PR(s) for that move and the other code cleanup suggestions.

robertwb:
LGTM too. Thanks.

PR #8859, opened 2019-06-14
[BEAM-7018] Added Regex transform for PythonSDK

mszb:

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/transforms/util.py
+++ b/sdks/python/apache_beam/transforms/util.py
@@ -24,9 +24,11 @@ from __future__ import division
 import collections
 import contextlib
 import random
+import re
 import time
 import typing
 import warnings
+from builtins import filter
 from builtins import object
 from builtins import range
 from builtins import zip
@@ -69,6 +71,7 @@ __all__ = [
     'Distinct',
     'Keys',
     'KvSwap',
+    'Regex',
     'Reify',
     'RemoveDuplicates',
     'Reshuffle',
@@ -865,3 +868,222 @@ class Reify(object):
 
     def expand(self, pcoll):
       return pcoll | ParDo(self.add_window_info)
+
+
+class Regex(object):
+  """
+  PTransform  to use Regular Expression to process the elements in a
+  PCollection.
+  """
+
+  ALL = "__regex_all_groups"
+
+  @staticmethod
+  def _regex_compile(regex):
+    """Return re.compile if the regex has a string value"""
+    if isinstance(regex, str):
+      regex = re.compile(regex)
+    return regex
+
+  @staticmethod
+  @typehints.with_input_types(str)
+  @typehints.with_output_types(str)
+  @ptransform_fn
+  def matches(pcoll, regex, group=0):
+    """
+    Returns the matches (group 0 by default) if zero or more characters at the
+    beginning of string match the regular expression. To match the entire
+    string, add "$" sign at the end of regex expression.
+
+    Group can be integer value or a string value.
+
+    Args:
+      regex: the regular expression string or (re.compile) pattern.
+      group: (optional) name/number of the group, it can be integer or a string
+        value. Defaults to 0, meaning the entire matched string will be
+        returned.
+    """
+    regex = Regex._regex_compile(regex)
+
+    def _process(element):
+      m = regex.match(element)
+      if m:
+        yield m.group(group)
+    return pcoll | FlatMap(_process)
+
+  @staticmethod
+  @typehints.with_input_types(str)
+  @typehints.with_output_types(typing.List[str])
+  @ptransform_fn
+  def all_matches(pcoll, regex):
+    """
+    Returns all matches (groups) if zero or more characters at the beginning
+    of string match the regular expression.
+
+    Args:
+      regex: the regular expression string or (re.compile) pattern.
+    """
+    regex = Regex._regex_compile(regex)
+
+    def _process(element):
+      m = regex.match(element)
+      if m:
+        yield [m.group(ix) for ix in range(m.lastindex + 1)]
+
+    return pcoll | FlatMap(_process)
+
+  @staticmethod
+  @typehints.with_input_types(str)
+  @typehints.with_output_types(typing.Tuple[str, str])
+  @ptransform_fn
+  def matches_kv(pcoll, regex, keyGroup, valueGroup=0):
+    """
+    Returns the KV pairs if the string matches the regular expression, deriving
+    the key & value from the specified group of the regular expression.
+
+    Args:
+      regex: the regular expression string or (re.compile) pattern.
+      keyGroup: The Regex group to use as the key. Can be int or str.
+      valueGroup: (optional) Regex group to use the value. Can be int or str.
+        The default value "0" returns entire matched string.
+    """
+    regex = Regex._regex_compile(regex)
+
+    def _process(element):
+      match = regex.match(element)
+      if match:
+        yield (match.group(keyGroup), match.group(valueGroup))
+    return pcoll | FlatMap(_process)
+
+  @staticmethod
+  @typehints.with_input_types(str)
+  @typehints.with_output_types(str)
+  @ptransform_fn
+  def find(pcoll, regex, group=0):
+    """
+    Returns the matches if a portion of the line matches the Regex. Returns
+    the entire group (group 0 by default). Group can be integer value or a
+    string value.
+
+    Args:
+      regex: the regular expression string or (re.compile) pattern.
+      group: (optional) name of the group, it can be integer or a string value.
+    """
+    regex = Regex._regex_compile(regex)
+
+    def _process(element):
+      r = regex.search(element)
+      if r:
+        yield r.group(group)
+    return pcoll | FlatMap(_process)
+
+  @staticmethod
+  @typehints.with_input_types(str)
+  @typehints.with_output_types(typing.Union[typing.List[str],
+                                            typing.Tuple[str, str]])
+  @ptransform_fn
+  def find_all(pcoll, regex, group=0, outputEmpty=True):
+    """
+    Returns the matches if a portion of the line matches the Regex. By default,
+    list of group 0 will return with empty items. To get all groups, pass the
+    `Regex.ALL` flag in the `group` parameter which returns all the groups in
+    the tuple format.
+
+    Args:
+      regex: the regular expression string or (re.compile) pattern.
+      group: (optional) name of the group, it can be integer or a string value.
+      outputEmpty: (optional) Should empty be output. True to output empties
+        and false if not.
+    """
+    regex = Regex._regex_compile(regex)
+
+    def _process(element):
+      matches = regex.finditer(element)
+      if group == Regex.ALL:
+        yield [(m.group(), m.groups()[0]) for m in matches if outputEmpty
+               or m.groups()[0]]
+      else:
+        yield [m.group(group) for m in matches if outputEmpty or m.group(group)]
+    return pcoll | FlatMap(_process)
+
+  @staticmethod
+  @typehints.with_input_types(str)
+  @typehints.with_output_types(typing.Tuple[str, str])
+  @ptransform_fn
+  def find_kv(pcoll, regex, keyGroup, valueGroup=0):
+    """
+    Returns the matches if a portion of the line matches the Regex. Returns the
+    specified groups as the key and value pair.
+
+    Args:
+      regex: the regular expression string or (re.compile) pattern.
+      keyGroup: The Regex group to use as the key. Can be int or str.
+      valueGroup: (optional) Regex group to use the value. Can be int or str.
+        The default value "0" returns entire matched string.
+    """
+    regex = Regex._regex_compile(regex)
+
+    def _process(element):
+      matches = regex.finditer(element)
+      if matches:
+        for match in matches:
+          yield (match.group(keyGroup), match.group(valueGroup))
+
+    return pcoll | FlatMap(_process)
+
+  @staticmethod
+  @typehints.with_input_types(str)
+  @typehints.with_output_types(str)
+  @ptransform_fn
+  def replace_all(pcoll, regex, replacement):
+    """
+    Returns the matches if a portion of the line  matches the regex and
+    replaces all matches with the replacement string.
+
+    Args:
+      regex: the regular expression string or (re.compile) pattern.
+      replacement: the string to be substituted for each match.
+    """
+    regex = Regex._regex_compile(regex)
+    return pcoll | Map(lambda elem: regex.sub(replacement, elem))
+
+  @staticmethod
+  @typehints.with_input_types(str)
+  @typehints.with_output_types(str)
+  @ptransform_fn
+  def replace_first(pcoll, regex, replacement):
+    """
+    Returns the matches if a portion of the line matches the regex and replaces
+    the first match with the replacement string.
+
+    Args:
+      regex: the regular expression string or (re.compile) pattern.
+      replacement: the string to be substituted for each match.
+    """
+    regex = Regex._regex_compile(regex)
+    return pcoll | Map(lambda elem: regex.sub(replacement, elem, 1))
+
+  @staticmethod
+  @typehints.with_input_types(str)
+  @typehints.with_output_types(typing.List[str])
+  @ptransform_fn
+  def split(pcoll, regex, outputEmpty=False):
+    """
+    Returns the list string which was splitted on the basis of regular
+    expression. It will not output empty items (by defaults).
+
+    Args:
+      regex: the regular expression string or (re.compile) pattern.
+      outputEmpty: (optional) Should empty be output. True to output empties
+          and false if not.
+    """
+    regex = Regex._regex_compile(regex)
+    outputEmpty = bool(outputEmpty)
+
+    def _process(element):
+      r = regex.split(element)
+      if r and not outputEmpty:
+        r = list(filter(None, r))
+      yield r
+
+    return pcoll | FlatMap(_process)
--- a/sdks/python/apache_beam/transforms/util_test.py
+++ b/sdks/python/apache_beam/transforms/util_test.py
@@ -24,6 +24,7 @@ import itertools
 import logging
 import math
 import random
+import re
 import time
 import unittest
 from builtins import object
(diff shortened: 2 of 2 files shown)


robertwb:
Thanks. These look generally useful. I wonder, however, if they should be in their own module rather than in a Regex class in the util module (which is generally a java-ism because everything must be in a class). Alternatively one could have a Regex module with methods like matches that returns the relevant PTransforms. 

The other thought that came to me is that this could probably be simplified a lot using lambdas, instead of creating a new PTransform and DoFn class each time. E.g.

```
class Regex(object):
  def matches(regex, group=0):
    regex = _regex_compile(regex)  # Do this once at construction time
    def maybe_match(element):
      m = regex.match(element)
      if m:
        yield m.group(group)
    return beam.FlatMap(maybe_match)
```

If this is a common pattern, we could have a utility function

```
  def match_objects(regex):
    regex = _regex_compile(regex)
    def maybe_match(element):
      m = regex.match(element)
      if m:
        yield m
    return beam.FlatMap(maybe_match)

  def matches(regex, group):
    return match_object(regex) | beam.Map(lambda m: m.group(group))
```

mszb:
Hi @robertwb. Thanks for review PR and for your suggestions. I believe making the separate the regex module is not what something I've seen in the py sdk. Because other modules (ex: Reify) is also available as a utility in py sdk. Moreover, regex transform is not something I think can be called as a module, so my thoughts util.py is a good place for these kinds of transforms 

And for the other part, I think you are right, using lambda will definitely improve the code in terms of readability.
let me make some changes.

mszb:
Run Python PreCommit

mszb:
retest this please

mszb:
Run Python PreCommit

mszb:
retest this please

mszb:
Run Python PreCommit

mszb:
Run Portable_Python PreCommit

mszb:
@robertwb I've made the changes you mentioned. Can you please review this PR.
Thanks.

mszb:
Run Python PreCommit

mszb:
R: @aaltay
R: @robertwb

aaltay:
@robertwb do you want to finish reviewing this?

mszb:
@aaltay I think @robertwb is unavailable. Can you assign someone else to review this PR?
Thanks!

mszb:
R: @aaltay 
R: @robertwb

sdks/python/apache_beam/transforms/util.py
```diff
@@ -864,3 +867,225 @@ def add_window_info(element, timestamp=DoFn.TimestampParam,
 
     def expand(self, pcoll):
       return pcoll | ParDo(self.add_window_info)
+
+
+@ptransform_fn
+def matches_all_object(pcoll, regex, group=None):
```
aaltay: Do we expect these to be public interfaces?

If not, can we prefix them with `_` and move them under Regex.

aaltay:
@mszb Both Robert and I were out of office for a bit. The change looks good to me. I will give @robertwb another day if he has any comments. After that we can merge it.

sdks/python/apache_beam/transforms/util.py
mszb: Nope, it should be private. Let me fix this.

sdks/python/apache_beam/transforms/util.py
```diff
@@ -864,3 +867,225 @@ def add_window_info(element, timestamp=DoFn.TimestampParam,
...
+  @staticmethod
+  def _regex_compile(regex):
+    """Return re.compile if the regex has a string value"""
+    if isinstance(regex, str):
+      regex = re.compile(regex)
+    return regex
+
+  @staticmethod
+  @ptransform_fn
+  def _matches_all_object(pcoll, regex, group=None):
```
robertwb: The group parameter seems to be ignored. Omit?

sdks/python/apache_beam/transforms/util.py
```diff
@@ -864,3 +867,225 @@ def add_window_info(element, timestamp=DoFn.TimestampParam,
...
+    """
+    PTransform to find the matches as per the regex.
```
robertwb: Comment in the docstring what the output should be. An example could be very helpful.

sdks/python/apache_beam/transforms/util.py
```diff
@@ -864,3 +867,225 @@ def add_window_info(element, timestamp=DoFn.TimestampParam,
...
+    """
+    def _process(element):
+      m = regex.finditer(element)
+      results = list()
+      for _, match in enumerate(m, start=1):
```
robertwb: Why the enumerate?

sdks/python/apache_beam/transforms/util.py
```diff
@@ -864,3 +867,225 @@ def add_window_info(element, timestamp=DoFn.TimestampParam,
...
+                                            valueGroup=valueGroup)
+
+  @staticmethod
+  @typehints.with_input_types(str)
+  @typehints.with_output_types(str)
+  @ptransform_fn
+  def find(pcoll, regex, group=None):
+    """
+    Returns the matches if a portion of the line matches the Regex. Returns
+    the entire line (group 0 by default). Group can be integer value and a
```
robertwb: s/the entire line/the entire group/g

sdks/python/apache_beam/transforms/util.py
```diff
@@ -864,3 +867,225 @@ def add_window_info(element, timestamp=DoFn.TimestampParam,
...
+    string value.
+
+    Args:
+      regex: the regular expression string or (re.compile) pattern.
+      group: (optional) name of the group, it can be integer or a string value.
+    """
+    regex = Regex._regex_compile(regex)
+    group = group or 0
```
robertwb: Why not just set 0 as the default value above?

sdks/python/apache_beam/transforms/util.py
```diff
@@ -864,3 +867,225 @@ def add_window_info(element, timestamp=DoFn.TimestampParam,
...
+
+    return pcoll | FlatMap(_process)
+
+  @staticmethod
+  @typehints.with_input_types(str)
+  @typehints.with_output_types(str)
+  @ptransform_fn
+  def matches(pcoll, regex, group=None):
+    """
+    Returns the matches if the entire line matched the Regex. Returns the
```
robertwb: As implemented, this returns whether the element starts with the given match, not whether it matches the entire line.

sdks/python/apache_beam/transforms/util.py
```diff
@@ -864,3 +867,225 @@ def add_window_info(element, timestamp=DoFn.TimestampParam,
...
+
+    def _process(element):
+      m = regex.match(element)
+      if m:
+        yield m.group(group)
+    return pcoll | FlatMap(_process)
+
+  @staticmethod
+  @typehints.with_input_types(str)
+  @typehints.with_output_types(str)
```
robertwb: Regex._matches_all_object returns a PCollection of lists (of groups). I'm not, however, sure that's what one wants. 

It seems the most primitive operation would be to have a transform that returns the (non-trivial) match objects themselves. Everything could then be implemented on top of that (e.g. extracting various groups). 

There's also a question of whether, when doing a "match all", a PCollection of lists (on for each matching element) would be best to return, or a flattened pcollection of all matches from all elements (less information, so I'd defer to the former or let it be optional).

sdks/python/apache_beam/transforms/util.py
```diff
@@ -864,3 +867,225 @@ def add_window_info(element, timestamp=DoFn.TimestampParam,
...
+      r = regex.search(element)
+      if r:
+        yield r.group(group)
+    return pcoll | FlatMap(_process)
+
+  @staticmethod
+  @typehints.with_input_types(str)
+  @typehints.with_output_types(str)
+  @ptransform_fn
+  def find_all(pcoll, regex):
```
robertwb: What's the difference between this and all_matches?

sdks/python/apache_beam/transforms/util.py
```diff
@@ -864,3 +867,225 @@ def add_window_info(element, timestamp=DoFn.TimestampParam,
...
+    """
+    Returns the matches if a portion of the line matches the Regex. Returns all
+    the groups as a List of string.
+
+    Args:
+      regex: the regular expression string or (re.compile) pattern.
+      keyGroup: The Regex group to use as the key. Can be int or str.
```
robertwb: These arguments don't look right.

sdks/python/apache_beam/transforms/util.py
```diff
@@ -864,3 +867,225 @@ def add_window_info(element, timestamp=DoFn.TimestampParam,
...
+      valueGroup: The Regex group to use the value. Can be int or str.
+    """
+    regex = Regex._regex_compile(regex)
+    return pcoll | Regex._matches_all_object(regex=regex)
+
+  @staticmethod
+  @typehints.with_input_types(str)
+  @typehints.with_output_types(typehints.KV[K, V])
+  @ptransform_fn
+  def find_kv(pcoll, regex, keyGroup, valueGroup):
```
robertwb: What's the difference between this and _matches_kv_object and matches_kv?

sdks/python/apache_beam/transforms/util.py
```diff
@@ -864,3 +867,225 @@ def add_window_info(element, timestamp=DoFn.TimestampParam,
...
+      regex: the regular expression string or (re.compile) pattern.
+      keyGroup: The Regex group to use as the key. Can be int or str.
+      valueGroup: The Regex group to use the value. Can be int or str.
+    """
+    regex = Regex._regex_compile(regex)
+    return pcoll | Regex._matches_all_object(regex=regex)
+
+  @staticmethod
+  @typehints.with_input_types(str)
+  @typehints.with_output_types(typehints.KV[K, V])
```
robertwb: These hints don't look right. Shouldn't the input be a kv, and the output of known type?

sdks/python/apache_beam/transforms/util.py
```diff
@@ -864,3 +867,225 @@ def add_window_info(element, timestamp=DoFn.TimestampParam,
...
+  @ptransform_fn
+  def matches_kv(pcoll, regex, keyGroup, valueGroup):
+    """
+    Returns the matches if the entire line matches the Regex. Returns the
```
robertwb: "Returns the matches" doesn't seem right. I'd say it maps the string into a kv pair if the string matches a regular expression, deriving the key and value from the specified group of the regular expression.

sdks/python/apache_beam/transforms/util.py
mszb: yes, you are right, let me remove it.

sdks/python/apache_beam/transforms/util.py
mszb: Okay, its a good idea. I am on it.

sdks/python/apache_beam/transforms/util.py
mszb: just used my old snippet which also uses the index, in our case indexes are not required, I will update this as well.

sdks/python/apache_beam/transforms/util.py
mszb: Yes, you are right because re.match start matching the regex from the beginning of the string. I'll update the docs.

sdks/python/apache_beam/transforms/util.py
mszb: `find_all()` should return results if the portion of the line matches the regular expression while `all_matches()` returns if the entire line matches the expression.
Fixed the code by using `re.finditer` in the `find_all()` and `re.match` in the `all_matches()`

sdks/python/apache_beam/transforms/util.py
mszb: Previous used `_matches_kv_object()` for both `matches_kv()` & `find_kv()`, that's why i created a private method to reuse the transform for both cases. But that approach is not correct and I've made some changes. It should be fine now!

sdks/python/apache_beam/transforms/util.py
mszb: It takes the string and returns the KV pair of it as defined in the regex. So I think hints are fine... Or I am missing something!

sdks/python/apache_beam/transforms/util.py
mszb: Done

sdks/python/apache_beam/transforms/util.py
mszb: Done

sdks/python/apache_beam/transforms/util.py
mszb: Done

sdks/python/apache_beam/transforms/util.py
mszb: Done

sdks/python/apache_beam/transforms/util.py
mszb: Done

mszb:
Run Python PreCommit

mszb:
Run Python PreCommit

mszb:
Run Python PreCommit

sdks/python/apache_beam/transforms/util.py
mszb: As per the java implementation, matches_kv return the list of KV (of groups). 
https://github.com/apache/beam/blob/master/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/Regex.java#L558 
I've updated the code little bit and found we no longer require `Regex._matches_all_object` as a separate private method so I moved the functionality in the `Regex.matches_kv` and removed the `Regex._matches_all_object` method.
(Previously I was using `Regex._matches_all_object` for both (`matches_kv` & `find_kv`), and now after the changes, I am using `re.match` for `matches_kv` and `re.finditer` for `find_kv`.)

And the third point regarding the return of the "match all". I think (the first approach you mentioned) PCollection of lists (containing the matching element of each item) is more relevant.  Because (as in the java api) "match all" applies to a single item of a collection, not to the entire collection.

mszb:
Run Python PreCommit

mszb:
Run Python PreCommit

aaltay:
@robertwb do you have additional comments?

sdks/python/apache_beam/transforms/util.py
robertwb: As as I read this code, for every input element, it returns the concatenated list of all groups of all matches it found, which feels odd. E.g. consider the regex "a(b*)" applied to "abb ax abbb". This will result in a single output ["abb", "bb", "a", "", "abbb", "bbb"].  

On the other hand, the Java implementation would only return ["abb", "bb"] (which isn't clear from its docs, and definitely doesn't match the name FindAll let alone re.findall). 

I think what makes sense here is to provide a findall() that is basically Map(lambda line: re.findall(regex, line)) and finditer that uses finditer, and the latter returns match objects which can be used in a follow-up lambda to extract groups, etc. rather than a flat list of indexed groups that would be more difficult to consume.

sdks/python/apache_beam/transforms/util.py
```diff
@@ -864,3 +867,223 @@ def add_window_info(element, timestamp=DoFn.TimestampParam,
...
+    Args:
+      regex: the regular expression string or (re.compile) pattern.
+      replacement: the string to be substituted for each match.
+    """
+    regex = Regex._regex_compile(regex)
+
+    def _process(element):
+      yield regex.sub(replacement, element)
+
+    return pcoll | FlatMap(_process)
```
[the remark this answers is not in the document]
mszb: Done.

sdks/python/apache_beam/transforms/util.py
```diff
@@ -864,3 +867,223 @@ def add_window_info(element, timestamp=DoFn.TimestampParam,
...
+          and false if not.
+    """
+    regex = Regex._regex_compile(regex)
+    outputEmpty = bool(outputEmpty)
+
+    def _process(element):
+      r = regex.split(element)
+      if r and not outputEmpty:
+        r = list(filter(None, r))
+      yield r
```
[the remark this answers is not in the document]
mszb: Fix the type hints to `typehints.List[str]`. I think it's better to get results of each collection item rather then flattens them all. Your thoughts?

sdks/python/apache_beam/transforms/util.py
```diff
@@ -864,3 +867,223 @@ def add_window_info(element, timestamp=DoFn.TimestampParam,
...
+      regex: the regular expression string or (re.compile) pattern.
+    """
+    regex = Regex._regex_compile(regex)
+
+    def _process(element):
+      m = regex.match(element)
+      results = list()
+
+      if m:
+        results.append(m.group())
```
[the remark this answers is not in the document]
mszb: Done.

mszb:

sdks/python/apache_beam/transforms/util.py
mszb: So what I understand from your feedback is:
1. for the 'Regex.find_all()', we need to change `re.finditer` to `re.findall` and return the result as a list.
2. create a new method `Regex.finditer()` which returns "match objects" instances.. 

The issue with the 'MatchObject' is we can not pickle it.
Showing error `PicklingError: Can't pickle <built-in method match of _sre.SRE_Pattern`
So either we can make a custom match object class and return it or uses KV pairs? Where K as the string of group and V as a tuple containing all groups?
For example:
lets say sting is "abb ax abbb" & regex expression is "a(b*)", so the results would be like:
`[
   KV("abb", ('bb',)), 
   KV("a", ('',)), 
   KV("abbb", ('bbb',)
]`
or maybe we can include group 0 in the values as well.
`[
   KV("abb", ('abb', 'bb',)), 
   KV("a", ('ab', '')), 
   KV("abbb", ('abbb', 'bbb')
]`

Your thoughts?

robertwb:
On Wed, Jul 24, 2019 at 6:30 PM Shoaib Zafar <notifications@github.com>
wrote:

> *@mszb* commented on this pull request.
> ------------------------------
>
> In sdks/python/apache_beam/transforms/util.py
> <https://github.com/apache/beam/pull/8859#discussion_r306906519>:
>
> > +      group: (optional) name of the group, it can be integer or a string value.
> +    """
> +    regex = Regex._regex_compile(regex)
> +    group = group or 0
> +
> +    def _process(element):
> +      r = regex.search(element)
> +      if r:
> +        yield r.group(group)
> +    return pcoll | FlatMap(_process)
> +
> +  @staticmethod
> +  @typehints.with_input_types(str)
> +  @typehints.with_output_types(str)
> +  @ptransform_fn
> +  def find_all(pcoll, regex):
>
> So what I understand from your feedback is:
>
>    1. for the 'Regex.find_all()', we need to change re.finditer to
>    re.findall and return the result as a list.
>    2. create a new method Regex.finditer() which returns "match objects"
>    instances..
>
> The issue with the 'MatchObject' is we can not pickle it.
> Showing error PicklingError: Can't pickle <built-in method match of
> _sre.SRE_Pattern
>
Oh, that is really unfortunate :(.

> So either we can make a custom match object class and return it
>
This seems nicest from an API perspective, but I'm worried about the
maintenance in faithfully emulating the match object over time.

> or uses KV pairs? Where K as the string of group and V as a tuple
> containing all groups?
> For example:
> lets say sting is "abb ax abbb" & regex expression is "a(b*)", so the
> results would be like:
> [ KV("abb", ('bb',)), KV("a", ('',)), KV("abbb", ('bbb',) ]
> or maybe we can include group 0 in the values as well.
> [ KV("abb", ('abb', 'bb',)), KV("a", ('ab', '')), KV("abbb", ('abbb',
> 'bbb') ]
>
In Python there's no such thing as a KV, it's just a 2-tuple.

Here's what I would do: many other transforms have an (optional) group
parameter, defaulting to 0. I would do that for find_all as well, so the
result would be simply

    ["abb", "a", "abbb"]

You could also have a special value (say, a constant like Regex.ALL) that
would return all the groups, i.e.

    [('a', 'abb'), ('a', ''), ('abbb', 'bbb')]

mszb:
Thanks for the feedback @robertwb. This approach seems good, I'll update the code!
Only one questions though, What I think, we create `Regex.find_all` which just Map `re.findall` and put all approach you mentioned above in the `Regex.find_iter` method!

Because `re.findall` returns a list of all the groups (`re.findall("a(b*)", "abb ax abbb")` >> `['bb', '', 'bbb']`) whereas in the above example we are going to return a list of group(0) `["abb", "a", "abbb"]`. Your thoughts?

robertwb:
Someone using Regex.find_iter might expect match objects, just as in
Python, so I'd avoid that naming.

You might need to use re.finditer to implement Regex.find_all though, due
to Regex.findall's variant signature.

mszb:
okay.

mszb:
Run Python PreCommit

mszb:
Hi @robertwb. I've made the changes you mentioned. Hope you find it well.

sdks/python/apache_beam/transforms/util_test.py
```diff
@@ -648,68 +648,70 @@ def test_find_empty(self):
...
+      test1_expected = [['abb', 'a', 'abbb'], ['ab', 'ab', 'ab']]
+      test2_expected = [['b', 'b', 'b'], ['bb', '', 'bbb']]
+      test3_expected = [['b', 'b', 'b'], ['bb', 'bbb']]
+      test4_expected = [['b', 'b', 'b'], ['bb', '', 'bbb']]
+      test5_expected = [[('ab', 'b'), ('ab', 'b'), ('ab', 'b')],
+                        [('abb', 'bb'), ('a', ''), ('abbb', 'bbb')]]
+      test6_expected = [[('ab', 'b'), ('ab', 'b'), ('ab', 'b')],
+                        [('abb', 'bb'), ('abbb', 'bbb')]]
+
+      assert_that(test1, equal_to(test1_expected), label='assert test 1')
```
robertwb: I would still put the test expectations up next to the defines, and label them in a meaningful way. Probably getting rid of the intermediates would make this easier to read, e.g.

```
assert_that(
    pcol | 'group 1' >> util.Regex.find_all("a(b*)", group=1),
    equal_to([['b', 'b', 'b'], ['bb', '', 'bbb']]),
    label='CheckGroup1')

assert_that(
    pcol | 'group 1 non empty' >> util.Regex.find_all("a(b*)", group=1, outputEmpty=False),
    equal_to([['b', 'b', 'b'], ['bb','bbb']]),
    label='CheckGroup1NoEmpty')

...
```

mszb: Sure, let me update them and put some meaningful labels.

As always, your feedback is very helpful. Thanks!

mszb:
Hi @aaltay. Robert has approved the PR. Do you have anything to add?

aaltay:
Nothing to add. Merging.

PR #8870, opened 2019-06-15
Fixing file naming for windows

pabloem:
cc: @chamikaramj 

I need to test this on windows before merging.

Colons are not acceptable in file names for windows: https://docs.microsoft.com/en-us/windows/desktop/FileIO/naming-a-file

open question: By default, fileio will give names with colons to files. Should we not do this, since it will be by-default broken in windows?


pabloem:
Run Python PreCommit

chamikaramj:
Yeah, I think it makes sense to pick a naming convention that works for windows as well. Also, let's make sure that Java and Python SDKs have the same behavior in this regard.

chamikaramj:
LGTM for getting this PR in for fixing tests.

PR #8872, opened 2019-06-17
[BEAM-3645] add ParallelBundleManager

Hannah-Jiang:
Add ParallelBundleManager class.
This class splits _GroupingBuffer into N pieces to reduce data size and improve processing performance.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/runners/portability/fn_api_runner.py
+++ b/sdks/python/apache_beam/runners/portability/fn_api_runner.py
@@ -124,8 +124,9 @@ class BeamFnControlServicer(beam_fn_api_pb2_grpc.BeamFnControlServicer):
       future = None
     else:
       if not item.instruction_id:
-        self._uid_counter += 1
-        item.instruction_id = 'control_%s' % self._uid_counter
+        with self._lock:
+          self._uid_counter += 1
+          item.instruction_id = 'control_%s' % self._uid_counter
       future = ControlFuture(item.instruction_id)
       self._futures_by_id[item.instruction_id] = future
     self._push_queue.put(item)
@@ -716,11 +717,11 @@ class FnApiRunner(runner.PipelineRunner):
           if other_input not in deferred_inputs:
             deferred_inputs[other_input] = _ListBuffer([])
         # TODO(robertwb): merge results
-        bundle_manager._skip_registration = True
         # We cannot split deferred_input until we include residual_roots to
         # merged results. Without residual_roots, pipeline stops earlier and we
         # may miss some data.
         bundle_manager._num_workers = 1
+        bundle_manager._skip_registration = True
         last_result, splits = bundle_manager.process_bundle(
             deferred_inputs, data_output)
         last_sent = deferred_inputs
@@ -1352,6 +1353,7 @@ class BundleManager(object):
   """
 
   _uid_counter = 0
+  _lock = threading.Lock()
 
   def __init__(
       self, controller, get_buffer, get_input_coder_impl, bundle_descriptor,
@@ -1472,8 +1474,9 @@ class BundleManager(object):
 
   def process_bundle(self, inputs, expected_outputs):
     # Unique id for the instruction processing this bundle.
-    BundleManager._uid_counter += 1
-    process_bundle_id = 'bundle_%s' % BundleManager._uid_counter
+    with BundleManager._lock:
+      BundleManager._uid_counter += 1
+      process_bundle_id = 'bundle_%s' % BundleManager._uid_counter
 
     # Register the bundle descriptor, if needed - noop if already registered.
     registration_future = self._register_bundle_descriptor()
@@ -1555,7 +1558,7 @@ class ParallelBundleManager(BundleManager):
           self._controller, self._get_buffer, self._get_input_coder_impl,
           self._bundle_descriptor, self._progress_frequency,
           self._registered).process_bundle(part, expected_outputs),
-                                               list(filter(None, part_inputs))):
+                                               part_inputs):
 
         split_result_list += split_result
         if merged_result is None:
--- a/sdks/python/apache_beam/runners/portability/fn_api_runner.py
+++ b/sdks/python/apache_beam/runners/portability/fn_api_runner.py
@@ -139,19 +139,34 @@ class BeamFnControlServicer(beam_fn_api_pb2_grpc.BeamFnControlServicer):
       self._state = self.DONE_STATE
 
 
+class _PartitionBuffer(object):
+  """This class is created to support partition(n) function."""
+  def __init__(self, inputs):
+    self._inputs = inputs
+
+  def partition(self, n):
+    v = list(self._inputs.values())[0]
+    if isinstance(v, list):
+      return [self._inputs]
+    elif isinstance(v, _GroupingBuffer):
+      partitions = []
+      for name, input in self._inputs.items():
+        for part in input.partition(n):
+          partitions.append({name : part})
+      return partitions
+    else:
+      raise NotImplementedError(type(self._inputs))
+
+
 class _GroupingBuffer(object):
   """Used to accumulate groupded (shuffled) results."""
-  def __init__(self, pre_grouped_coder, post_grouped_coder, windowing,
-               num_workers=1):
+  def __init__(self, pre_grouped_coder, post_grouped_coder, windowing):
     self._key_coder = pre_grouped_coder.key_coder()
     self._pre_grouped_coder = pre_grouped_coder
     self._post_grouped_coder = post_grouped_coder
     self._table = collections.defaultdict(list)
     self._windowing = windowing
-    self._grouped_output = None
-    self._lock = threading.Lock()
-    self._iter_flag = 0
-    self._num_workers = num_workers
+    self._grouped_output = []
 
   def append(self, elements_data):
     if self._grouped_output:
@@ -169,61 +184,45 @@ class _GroupingBuffer(object):
           value if is_trivial_windowing
           else windowed_key_value.with_value(value))
 
-  def __iter__(self):
-    # use self._iter_flag to indicate _grouped_output process status.
-    # 0: not started, 1: started, 2: finished, 99: failed
-    if self._iter_flag > 0:
-      while self._iter_flag == 1:
-        time.sleep(0.2)
-
-      if self._iter_flag == 2:
-        return iter(self._grouped_output)
-      elif self._iter_flag == 99:
-        raise RuntimeError('Failed when reading from buffer.')
+  def _output_ready(self):
+    if len(self._grouped_output) == 0:
+      if self._windowing.is_default():
+        globally_window = GlobalWindows.windowed_value(None).with_value
+        windowed_key_values = lambda key, values: [
+            globally_window((key, values))]
       else:
-        raise RuntimeError('Unhandled error occurred when reading from buffer.')
+        # TODO(pabloem, BEAM-7514): Trigger driver needs access to the clock
+        #   note that this only comes through if windowing is default - but what
+        #   about having multiple firings on the global window.
+        #   May need to revise.
+        trigger_driver = trigger.create_trigger_driver(self._windowing, True)
+        windowed_key_values = trigger_driver.process_entire_key
 
-    with self._lock:
-      try:
-        self._iter_flag = 1
-        self._grouped_output = []
-        num_workers = min(self._num_workers, len(self._table))
-        output_stream_list = []
-        for _ in range(num_workers):
-          output_stream_list.append(create_OutputStream())
+      key_coder_impl = self._key_coder.get_impl()
+      coder_impl = self._post_grouped_coder.get_impl()
+      for encoded_key, windowed_values in self._table.items():
+        key = key_coder_impl.decode(encoded_key)
+        output_stream = create_OutputStream()
+        for wkvs in windowed_key_values(key, windowed_values):
+          coder_impl.encode_to_stream(wkvs, output_stream, True)
 
-        if self._windowing.is_default():
-          globally_window = GlobalWindows.windowed_value(None).with_value
-          windowed_key_values = lambda key, values: [
-              globally_window((key, values))]
-        else:
-          # TODO(pabloem, BEAM-7514): Trigger driver needs access to the clock
-          #   note that this only comes through if windowing is default - but
-          #   what about having multiple firings on the global window.
-          #   May need to revise.
-          trigger_driver = trigger.create_trigger_driver(self._windowing, True)
-          windowed_key_values = trigger_driver.process_entire_key
+        self._grouped_output.append(output_stream.get())
+      self._table = None
 
-        coder_impl = self._post_grouped_coder.get_impl()
-        key_coder_impl = self._key_coder.get_impl()
+  def __iter__(self):
+    self._output_ready()
+    return iter(self._grouped_output)
 
-        for idx, (encoded_key, windowed_values) in \
-            enumerate(self._table.items()):
-          key = key_coder_impl.decode(encoded_key)
-          for wkvs in windowed_key_values(key, windowed_values):
-            coder_impl.encode_to_stream(wkvs,
-                                        output_stream_list[idx % num_workers],
-                                        True)
+  def partition(self, n):
+    self._output_ready()
 
-        for output_stream in output_stream_list:
-          self._grouped_output.append(b''.join([output_stream.get()]))
+    n = min(n, len(self._grouped_output))
+    partitions = [[] for _ in range(n)]
 
-        self._iter_flag = 2
-        return iter(self._grouped_output)
-      except Exception as e:
-        self._iter_flag = 99
-        raise RuntimeError('Failed to generate an iterator from buffer. %s' % e)
+    for idx, out in enumerate(self._grouped_output):
+      partitions[idx % n].append(out)
 
+    return partitions
 
 
 class _WindowGroupingBuffer(object):
@@ -269,6 +268,7 @@ class _WindowGroupingBuffer(object):
 
 
 class FnApiRunner(runner.PipelineRunner):
+  _num_workers = 1
 
   def __init__(
       self,
@@ -288,12 +288,12 @@ class FnApiRunner(runner.PipelineRunner):
       provision_info: provisioning info to make available to workers, or None
     """
     super(FnApiRunner, self).__init__()
+    FnApiRunner._num_workers = num_workers
     self._last_uid = -1
     self._default_environment = (
         default_environment
         or beam_runner_api_pb2.Environment(urn=python_urns.EMBEDDED_PYTHON))
     self._bundle_repeat = bundle_repeat
-    self._num_workers = num_workers
     self._progress_frequency = None
     self._profiler_factory = None
     self._use_state_iterables = use_state_iterables
@@ -322,8 +322,8 @@ class FnApiRunner(runner.PipelineRunner):
     pipeline.visit(DataflowRunner.group_by_key_input_visitor())
     self._bundle_repeat = self._bundle_repeat or options.view_as(
         pipeline_options.DirectOptions).direct_runner_bundle_repeat
-    self._num_workers = max(options.view_as(
-        pipeline_options.DirectOptions).direct_num_workers, self._num_workers)
+    FnApiRunner._num_workers = max(FnApiRunner._num_workers, options.view_as(
+        pipeline_options.DirectOptions).direct_num_workers)
     self._profiler_factory = profiler.Profile.factory_from_options(
         options.view_as(pipeline_options.ProfilingOptions))
 
@@ -457,15 +457,14 @@ class FnApiRunner(runner.PipelineRunner):
                                              process_bundle_descriptor,
                                              data_input,
                                              data_output,
-                                             get_input_coder_callable,
-                                             num_workers):
+                                             get_input_coder_callable):
     for k in range(self._bundle_repeat):
       try:
         controller.state.checkpoint()
         ParallelBundleManager(
             controller, lambda pcoll_id: [], get_input_coder_callable,
             process_bundle_descriptor, self._progress_frequency, k
-        ).process_bundle(data_input, data_output, num_workers)
+        ).process_bundle(data_input, data_output)
       finally:
         controller.state.restore()
 
@@ -669,8 +668,7 @@ class FnApiRunner(runner.PipelineRunner):
               pipeline_components
               .pcollections[output_pcoll].windowing_strategy_id]
           pcoll_buffers[buffer_id] = _GroupingBuffer(
-              pre_gbk_coder, post_gbk_coder, windowing_strategy,
-              self._num_workers)
+              pre_gbk_coder, post_gbk_coder, windowing_strategy)
       else:
         # These should be the only two identifiers we produce for now,
         # but special side input writes may go here.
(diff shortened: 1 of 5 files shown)


Hannah-Jiang:
R: @robertwb

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
```diff
@@ -1185,6 +1185,32 @@ def create_pipeline(self):
...
+    raise unittest.SkipTest("Multiworker doesn't support split request.")
+
+class FnApiRunnerTestWithMultiWorkersAndBundleRepeat(FnApiRunnerTest):
+
+  def create_pipeline(self):
+    return beam.Pipeline(
+        runner=fn_api_runner.FnApiRunner(num_workers=2, bundle_repeat=2))
+
+  def test_checkpoint(self):
+    raise unittest.SkipTest("Multiworker doesn't support split request.")
```
Hannah-Jiang: ParallelBundleManager works well with all tests except the ones with split manager. Without fully knowing about how split request works, I'm not sure what issue it is. I would like to confirm if we want to support pipelines with split request in this case, because splitting is another way to chunk data into smaller set. In addition, depends on how split manager is designed, in some cases, split function will not be called. For example, `test_split_half` test only run split when `num_elements == total_num_elements`, however, with ParalleBundleManager, num_elements will not be equal to total_num_elements because it is already chunked to N pieces.

Hannah-Jiang:
Run Python PreCommit

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
Hannah-Jiang: Here is a pattern I observed with `test_checkpoint` test. If we split the buffer to N pieces, there are (N-1) elements repeat twice.
For example, if we split [a,b,c] into 2 pieces, it returns one of [a,a,b,c] or [a,b,b,c] or [a,b,c,c]. If we split [a,b,c] into 3 pieces, it returns one of [a,a,b,b,c] or [a,a,b,c,c] or [a,b,b,c,c]. if we split [a,b,c,d] into 2 pieces, it returns one of [a,a,b,c,d] or [a,b,b,c,d] or [a,b,c,c,d] or [a,b,c,d,d].

Hannah-Jiang: This isn't an issue anymore with a new class with partition() function.

Hannah-Jiang:
Run Portable_Python PreCommit

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -139,6 +139,25 @@ def done(self):
       self._state = self.DONE_STATE
 
 
+class _PartitionBuffer(object):
```
robertwb: Is there any reason this is a class that has a single method rather than just a function?

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -139,6 +139,25 @@ def done(self):
...
+  """This class is created to support partition(n) function."""
+  def __init__(self, inputs):
+    self._inputs = inputs
+
+  def partition(self, n):
+    v = list(self._inputs.values())[0]
```
robertwb: The check, if any, should be inside the loop and per input. This code (implicitly) assumes exactly one element in inputs--we shouldn't do so without explicitly asserting that. Better would be to handle general dicts.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -139,6 +139,25 @@ def done(self):
...
+    if isinstance(v, list):
```
robertwb: I still think it is more natural to use polymorphism rather than if-elif chains for known types. Is there a reason to not do this?

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -139,6 +139,25 @@ def done(self):
...
+      return [self._inputs]
```
robertwb: We will also want to split this case up.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -179,16 +197,33 @@ def __iter__(self):
         #   May need to revise.
         trigger_driver = trigger.create_trigger_driver(self._windowing, True)
         windowed_key_values = trigger_driver.process_entire_key
-      coder_impl = self._post_grouped_coder.get_impl()
+
       key_coder_impl = self._key_coder.get_impl()
+      coder_impl = self._post_grouped_coder.get_impl()
```
robertwb: Why this change?

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -179,16 +197,33 @@ def __iter__(self):
...
       for encoded_key, windowed_values in self._table.items():
         key = key_coder_impl.decode(encoded_key)
+        output_stream = create_OutputStream()
         for wkvs in windowed_key_values(key, windowed_values):
           coder_impl.encode_to_stream(wkvs, output_stream, True)
-      self._grouped_output = [output_stream.get()]
+
+        self._grouped_output.append(output_stream.get())
```
robertwb: Creating a new stream per element is expensive (and will also be expensive on the other side). Best to take n as an input here.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -179,16 +197,33 @@ def __iter__(self):
...
       self._table = None
+
+  def __iter__(self):
```
robertwb: Do we actually use raw `__iter__` anywhere anymore? If not, remove. If so, perhaps implement as iter(self.partition(1)).

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -179,16 +197,33 @@ def __iter__(self):
...
+    self._output_ready()
     return iter(self._grouped_output)
 
+  def partition(self, n):
+    self._output_ready()
+
```
robertwb: Nit: lots of blank lines in this function (and above).

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -233,11 +268,13 @@ def encoded_items(self):
 
 
 class FnApiRunner(runner.PipelineRunner):
+  _num_workers = 1
```
robertwb: This should be an instance, not class-level, variable. Class variables are global state.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -284,6 +322,8 @@ def run_pipeline(self, pipeline, options):
     pipeline.visit(DataflowRunner.group_by_key_input_visitor())
     self._bundle_repeat = self._bundle_repeat or options.view_as(
         pipeline_options.DirectOptions).direct_runner_bundle_repeat
+    FnApiRunner._num_workers = max(FnApiRunner._num_workers, options.view_as(
```
robertwb: Rather than max, it's probably better to have one take precedence. Alternatively, only have one way to set this (e.g. via the option).

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1444,9 +1485,12 @@ def _generate_splits_for_testing(self,
         break
     return split_results
 
-  def process_bundle(self, inputs, expected_outputs):
+  def process_bundle(self, inputs, expected_outputs, parallel_uid_counter=None):
     # Unique id for the instruction processing this bundle.
-    BundleManager._uid_counter += 1
+    if parallel_uid_counter:
```
robertwb: Why this change. It seems we don't need to ever pass the uid counter in externally.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
Hannah-Jiang: I tried to split it, however, `test_pardo_timers` and other tests using `_run_pardo_state_timers()` cannot pass. A pattern I observed is when we have a timer, it has two elements at inputs and should be passed together, otherwise, the pipeline is hanging. How do we work around these cases?

an input example with a timer:
```
[{'Create/Read/Impulse': [b'\x7f\xdf;dZ\x1c\xac\t\x00\x00\x00\x01\x0f\x00'], 'ParDo(TimerDoFn)_timers_to_read_timer/Read': []}]
```
This will split into 
```
[{'Create/Read/Impulse': [b'\x7f\xdf;dZ\x1c\xac\t\x00\x00\x00\x01\x0f\x00']}, {'ParDo(TimerDoFn)_timers_to_read_timer/Read': []}]
```

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -141,13 +141,18 @@ def done(self):
...
 class _GroupingBuffer(object):
   """Used to accumulate groupded (shuffled) results."""
-  def __init__(self, pre_grouped_coder, post_grouped_coder, windowing):
+  def __init__(self, pre_grouped_coder, post_grouped_coder, windowing,
+               num_workers=1):
     self._key_coder = pre_grouped_coder.key_coder()
     self._pre_grouped_coder = pre_grouped_coder
     self._post_grouped_coder = post_grouped_coder
     self._table = collections.defaultdict(list)
+    self._table_count = 0
```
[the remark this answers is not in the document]
Hannah-Jiang: It is not used and removed at the new commit.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -691,13 +737,14 @@ def input_for(ptransform_id, input_id):
           if other_input not in deferred_inputs:
             deferred_inputs[other_input] = []
         # TODO(robertwb): merge results
-        last_result, splits = BundleManager(
+        last_result, splits = ParallelBundleManager(
```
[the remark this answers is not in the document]
Hannah-Jiang: I changed it to always use `ParallelBundleManager`, when num_workers = 1, inputs will not be spitted. Does this sound good?

sdks/python/apache_beam/runners/portability/fn_api_runner.py
Hannah-Jiang: This class is removed now.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
Hannah-Jiang: I added `_ListBuffer` class to support partitioning for list.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
Hannah-Jiang: This is removed now.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
Hannah-Jiang: It turns out that we cannot split inputs with a timer. This is fixed at new commit.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
Hannah-Jiang: Yeah, changed to `partition(n)` at new commit..

sdks/python/apache_beam/runners/portability/fn_api_runner.py
Hannah-Jiang: Changed to support option only.

Hannah-Jiang:
Robert, thanks for your comments, they are very helpful and thoughtful.
I fixed all at new commit, can you please take a look when you have time?
Many thanks.

Hannah-Jiang:
Run Python PreCommit

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -181,14 +196,32 @@ def __iter__(self):
...
       self._table = None
     return iter(self._grouped_output)
 
+  def __iter__(self):
+    """ Since partition() returns a list of list, added this __iter__ to return
+    a list to simplify code when we need to iterate through ALL elements of
+    _GroupingBuffer.
+    """
+    if len(self._grouped_output) == 0:
+      self.partition(1)
```
robertwb: If calling a method just for its side effects, it probably makes sense to have a comment.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -181,14 +196,32 @@ def __iter__(self):
...
+    iter_result = []
+    for output in self._grouped_output:
+      for out in output:
+        iter_result.append(out)
```
robertwb: Lists support + for concatination. Thus you can do

`return iter(sum(self.partition(1), []))`

to sum up the lists into the full list. (The second argument to sum is the "start" element.) You could also use `itertools.chain(*self.partition(1))` (which might be cleaner as there's no need to even write iter() in this case).

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1343,6 +1381,7 @@ def __init__(
     self._controller = controller
     self._get_buffer = get_buffer
     self._get_input_coder_impl = get_input_coder_impl
+    self._num_workers = num_workers
```
robertwb: This argument can now be omitted.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1506,6 +1545,52 @@ def process_bundle(self, inputs, expected_outputs):
     return result, split_results
 
 
+class ParallelBundleManager(BundleManager):
+
+  def _check_inputs_split(self, expected_outputs):
+    # We skip splitting inputs when timer is set, because operations are not
```
robertwb: See if the suggestion below fixes this.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1506,6 +1545,52 @@ def process_bundle(self, inputs, expected_outputs):
...
+    return True
+
+  def process_bundle(self, inputs, expected_outputs, num_workers=None):
+    num_workers = num_workers or self._num_workers
+    param_list = []
+
+    if self._check_inputs_split(expected_outputs):
+      for name, input in inputs.items():
+        for part in input.partition(num_workers):
+          param_list.append(({name : part}, expected_outputs))
```
robertwb: OK, I think this is where the bug above is. Timers are a case where there is more than one set of inputs, but this here adds a new element to param list for every input for every part. 

Instead, I think you need something like

```
partitioned_inputs = [{} for _ in range(num_workers)]
for name, input in inputs.items():
  for ix, part in input.partition(num_workers):
    partitioned_inputs[ix][name] = part
```

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1506,6 +1545,52 @@ def process_bundle(self, inputs, expected_outputs):
...
+    else:
+      param_list.append((inputs, expected_outputs))
+
+    merged_result = None
+    split_result_list = []
+    with futures.ThreadPoolExecutor(max_workers=num_workers) as executor:
+      for result, split_result in executor.map(lambda p: BundleManager(
+          self._controller, self._get_buffer, self._get_input_coder_impl,
+          self._bundle_descriptor, self._progress_frequency,
+          self._registered).process_bundle(*p), param_list):
```
robertwb: As expected_outputs never changes, you could simplify this by doing

`executor.map(lambda part: BundleManager(...).process_bundle(part, expected_outputs), partitioned_inputs)`

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
```diff
@@ -1176,6 +1176,18 @@ def create_pipeline(self):
...
 
+class FnApiRunnerTestWithGrpcAndMultiWorkers(FnApiRunnerTest):
+
+  def create_pipeline(self):
+    from apache_beam.options.pipeline_options import DirectOptions
+    p = beam.Pipeline(
+        runner=fn_api_runner.FnApiRunner(
+            default_environment=beam_runner_api_pb2.Environment(
+                urn=python_urns.EMBEDDED_PYTHON_GRPC)))
+    p._options.view_as(DirectOptions).direct_num_workers = 2
```
robertwb: Generally it's preferable to create a pipelines object and pass it to the Pipeline constructor instead of mutating it after the fact.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
Hannah-Jiang: This worked fantastically!!

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -139,6 +139,18 @@ def done(self):
       self._state = self.DONE_STATE
 
 
+class _ListBuffer(list):
+  """Used to support parititioning of a list."""
+  def partition(self, n):
+    n = min(n, len(self))
```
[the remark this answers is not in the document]
Hannah-Jiang: I think we still need it, because if we return n groups with some empty groups, we will pass additional {target:[]} to `process_bundle` which is not expected. Some tests are hanging without this.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
Hannah-Jiang: This is used at L:1540. Default num_workers is got from self._num_workers, and it can be overwritten if num_worker is passed to `process_bundle`.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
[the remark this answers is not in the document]
robertwb: We already need to be able to support empty groups (e.g. initially all timers are empty, when timers are fired all "normal" inputs are empty, some PCollections are actually empty). If there are hanging tests, we should investigate these. 

In terms of hanging tests, the key is that the worker waits for a data stream for every InputPortOperator it has in the graph. So there must be a set of elements (even if it is empty) to send to each of these. Maybe this invariant is being broken somewhere.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1506,6 +1534,39 @@ def process_bundle(self, inputs, expected_outputs):
     return result, split_results
 
 
+class ParallelBundleManager(BundleManager):
+
+  def process_bundle(self, inputs, expected_outputs, num_workers=None):
```
robertwb: Do we need two separate ways of passing num_workers?

sdks/python/apache_beam/runners/portability/fn_api_runner.py
robertwb: I see. We should capture this in the constructor of ParallelBundleProcessor (which may call its superclass constructor), not here.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
Hannah-Jiang: I added it in case we want to change num_workers for a certain bundle only, instead of changing at BundleManager level. Do we have a case that process multi bundles parallel with different num_workers? I now changed back to support only self._num_workers, I would like to grab your advice if we even need to consider such cases.

robertwb: While I wouldn't rule out finding a reason to support such a thing later, I would say that it's be cleaner to add such support if/once we need it rather than now. It's generally much easier to read/follow code if parameters come from only one place.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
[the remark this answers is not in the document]
Hannah-Jiang: Here is my investigation.
With `n = min(n, len(self)`, we generate following partitioned inputs.
```
[{'Create/Read/Impulse': [b'\x7f\xdf;dZ\x1c\xac\t\x00\x00\x00\x01\x0f\x00'], 'ParDo(BufferDoFn)_timers_to_read_timer/Read': []}, {}]
```
With fixed n, we get following partitioned input.
```
[{'Create/Read/Impulse': [b'\x7f\xdf;dZ\x1c\xac\t\x00\x00\x00\x01\x0f\x00'], 'ParDo(BufferDoFn)_timers_to_read_timer/Read': []}, {'Create/Read/Impulse': []}]
```
With first case, the second bundle is empty, which can be handled automatically. With second case, second bundle has an input, but without timer, so workers are waiting timers to be sent, which is not possible with second bundle and getting stuck. This happens to all tests which uses timer.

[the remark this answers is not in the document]
robertwb: There were multiple occurrences of `n = min(n, len(self)`, did you change all of them? What I'm worried about is (say) one buffer having 2 elements and another 3 elements which we can't handle automatically (as the third dict would not be empty--we'd need to have the lists, even empty lists, for all inputs for all dicts).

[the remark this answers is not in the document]
Hannah-Jiang: I got it. The problem was here.
```
def partition(self, n):
    return [self[k::n] for k in range(n)] if self else [[]]
```
After I change it to return [] *n empty lists, it all worked.
```
def partition(self, n):
    return [self[k::n] for k in range(n)] if self else [[] for _ in range(n)]
```

Thanks very much for your critical thinking!

Hannah-Jiang:
Run Python PreCommit

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -139,6 +139,12 @@ def done(self):
       self._state = self.DONE_STATE
 
 
+class _ListBuffer(list):
+  """Used to support parititioning of a list."""
+  def partition(self, n):
+    return [self[k::n] for k in range(n)] if self else [[] for _ in range(n)]
```
robertwb: You can drop the `if self else [[] for _ in range(n)]`, the slicing will work fine on an empty list as well.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1329,7 +1356,7 @@ class BundleManager(object):
 
   def __init__(
       self, controller, get_buffer, get_input_coder_impl, bundle_descriptor,
-      progress_frequency=None, skip_registration=False):
+      num_workers, progress_frequency=None, skip_registration=False):
```
robertwb: Delete this argument.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1506,6 +1533,46 @@ def process_bundle(self, inputs, expected_outputs):
     return result, split_results
 
 
+class ParallelBundleManager(BundleManager):
+
+  def __init__(
+      self, controller, get_buffer, get_input_coder_impl, bundle_descriptor,
+      num_workers, progress_frequency=None, skip_registration=False):
```
robertwb: Generally it'd be preferable to pass num_workers as the last argument, by keyword, so the signatures between super and subclasses agree.

Hannah-Jiang:
Run Python PreCommit

Hannah-Jiang:
Run Portable_Python PreCommit

Hannah-Jiang:
Run RAT PreCommit

Hannah-Jiang:

Hannah-Jiang:
Run Portable_Python PreCommit

Hannah-Jiang:
Run Python PreCommit

Hannah-Jiang:
Run Python PreCommit

Hannah-Jiang:
Run Python PreCommit

Hannah-Jiang:
Run Python PreCommit

Hannah-Jiang:
Run Python PreCommit

Hannah-Jiang:
Run Python PreCommit

Hannah-Jiang:
Run Python PreCommit

Hannah-Jiang:
Run Python PreCommit

Hannah-Jiang:
Run Python PreCommit

Hannah-Jiang:
Run Python PreCommit

PR #8884, opened 2019-06-18
Add a PCollectionCache ABC and several file-based implementations

ostrokach:
Currently, the Python SDK lacks a clear API for data inter-op between a Python console and a Beam pipeline. During interactive development and debugging of Beam pipelines, this type of inter-op would be tremendously useful. 

Other "big data" processing tools in the Python ecosystem already have this functionality. For example, PySpark has [`createDataFrame`](https://spark.apache.org/docs/latest/api/python/pyspark.sql.html#pyspark.sql.SparkSession.createDataFrame), which allows the creation of an RDD from a Python list or a Pandas DataFrame, and [`collect()`]() and [`toPandas()`](https://spark.apache.org/docs/latest/api/python/pyspark.sql.html#pyspark.sql.DataFrame.toPandas) methods which make it possible to transfer the data to the Python interpreter. Dask has a similar API using [`dask.compute()`](https://docs.dask.org/en/latest/api.html#dask.compute).

Although Beam has a `Create` `PTransform` which makes it possible to create a `PCollection` from a Python list. However, this has limited . Beam lacks a well-defined API for "materializing" a `PCollection` and loading it into the memory of an interactive Python session. 

The following applications of `PCollectionCache` will be implemented in subsequent PRs:

- Add a stream-based implementation (using Google Cloud Pub/Sub) of the `PCollectionCache` API [#8961].
- Update `InteractiveRunner` and `CacheManager` to use `PCollectionCache` objects for storing cache data. This will add streaming pipeline support to `InteractiveRunner` and allow for more flexibility when optimizing the the interactive user experience.

Demos: 

| Notebook | MyBinder | Colab |
| ------------- | ------------- | -------- |
| [01-demo_explicit_caching.ipynb](https://github.com/ostrokach/beam-notebooks/blob/master/notebooks/01-demo_explicit_caching.ipynb) | [Binder](https://mybinder.org/v2/gh/ostrokach/beam-notebooks/master?filepath=notebooks%2F01-demo_explicit_caching.ipynb) | [Colab](https://colab.research.google.com/github/ostrokach/beam-notebooks/blob/master/notebooks/01-demo_explicit_caching.ipynb)
| [01-demo_pipeline_segmentation.ipynb](https://github.com/ostrokach/beam-notebooks/blob/master/notebooks/01-demo_pipeline_segmentation.ipynb) | [Binder](https://mybinder.org/v2/gh/ostrokach/beam-notebooks/master?filepath=notebooks%2F01-demo_pipeline_segmentation.ipynb) | [Colab](https://colab.research.google.com/github/ostrokach/beam-notebooks/blob/master/notebooks/01-demo_pipeline_segmentation.ipynb)
| [01-demo_pipeline_cloning.ipynb](https://github.com/ostrokach/beam-notebooks/blob/master/notebooks/01-demo_pipeline_cloning.ipynb) | [Binder](https://mybinder.org/v2/gh/ostrokach/beam-notebooks/master?filepath=notebooks%2F01-demo_pipeline_cloning.ipynb) | [Colab](https://colab.research.google.com/github/ostrokach/beam-notebooks/blob/master/notebooks/01-demo_pipeline_cloning.ipynb)

CC: @ivant @ananvay  
R: @aaltay @udim 

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/runners/interactive/caching/file_based_cache.py
+++ b/sdks/python/apache_beam/runners/interactive/caching/file_based_cache.py
@@ -1,0 +1,309 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+from __future__ import absolute_import
+
+import urllib
+
+from apache_beam import coders
+from apache_beam.io import textio
+from apache_beam.io import tfrecordio
+from apache_beam.io.filesystems import FileSystems
+from apache_beam.runners.interactive.caching import PCollectionCache
+from apache_beam.testing import datatype_inference
+from apache_beam.transforms import PTransform
+
+try:
+  from weakref import finalize
+except ImportError:
+  from backports.weakref import finalize
+
+try:  # Python 3
+  unquote_to_bytes = urllib.parse.unquote_to_bytes
+  quote = urllib.parse.quote
+except AttributeError:  # Python 2
+  # pylint: disable=deprecated-urllib-function
+  unquote_to_bytes = urllib.unquote
+  quote = urllib.quote
+
+__all__ = [
+    "FileBasedCache",
+    "TextBasedCache",
+    "TFRecordBasedCache",
+    "SafeFastPrimitivesCoder",
+]
+
+
+class FileBasedCache(PCollectionCache):
+
+  def __init__(self, location, mode="error", persist=False, **writer_kwargs):
+    """Initialize a PCollectionCache object.
+
+    Args:
+      location (str): A string indicating the location where the cache should
+          be stored. Typically, this includes the folder and the filename
+          prefix for files that will be created.
+      mode (str): Controls the behavior when one or more files matching location
+          already exit. If "error", an IOError will be raised. If "overwrite",
+          the existing files will be removed.
+      persist (bool): A flag indicating whether the underlying data should be
+          destroyed when the cache instance goes out of scope.
+      writer_kwargs (Dict[str, Any]): A dictionary of key-value pairs
+          to be passed to the underlying writer objects.
+
+    Raises:
+      ~exceptions.ValueError: If the specified mode is not supported.
+      ~exceptions.IOError: If one or more files matching `location` already
+          exist and mode == "error".
+    """
+    self.location = location
+    self.element_type = None
+    self.default_coder = writer_kwargs.pop("coder", None)
+    self._writer_kwargs = writer_kwargs
+    self._num_writes = 0
+    self._persist = persist
+    self._finalizer = self._configure_finalizer(persist)
+
+    # TODO(ostrokach): Implement append mode.
+    if mode not in ['error', 'overwrite']:
+      raise ValueError("'mode' must be set to one of: ['error', 'overwrite'].")
+    exitsting_files = list(glob_files(self.file_pattern))
+    if mode == "error" and exitsting_files:
+      raise IOError("The following cache files already exist: {}.".format(
+          exitsting_files))
+    if mode == "overwrite":
+      self.truncate()
+
+    root, _ = FileSystems.split(self._file_path_prefix)
+    try:
+      FileSystems.mkdirs(root)
+    except IOError:
+      pass
+    # It is possible to read from am empty stream, so it should also be possible
+    # to read from an empty file.
+    FileSystems.create(self._file_path_prefix + ".empty").close()
+
+  @property
+  def persist(self):
+    return self._persist
+
+  @persist.setter
+  def persist(self, persist):
+    if self._persist == persist:
+      return
+    self._persist = persist
+    if self._finalizer:
+      self._finalizer.detach()
+    self._finalizer = self._configure_finalizer(persist)
+
+  @property
+  def timestamp(self):
+    timestamp = 0
+    for path in glob_files(self.file_pattern):
+      timestamp = max(timestamp, FileSystems.last_updated(path))
+    return timestamp
+
+  def reader(self, **kwargs):
+    """Returns a reader ``PTransform`` to be used for reading the contents of
+    the cache into a Beam Pipeline.
+
+    Args:
+      kwargs (Dict[str, Any]): Arguments to be passed to the underlying reader
+          class.
+    """
+    reader_kwargs = self._reader_kwargs.copy()
+    reader_kwargs.update(kwargs)
+    reader = self._reader_class(self.file_pattern, **reader_kwargs)
+    # Keep a reference to the parent object so that cache does not get garbage
+    # collected while the pipeline is running.
+    reader._cache = self
+    return reader
+
+  def writer(self):
+    """Returns a writer object to be used for writing the contents of a
+    PCollection from a Beam Pipeline into the cache.
+    """
+    self._num_writes += 1
+    writer_kwargs = self._writer_kwargs.copy()
+
+    if self.element_type is None:
+      writer = PatchedWriter(self, self._writer_class,
+                             (self._file_path_prefix,), writer_kwargs)
+      return writer
+
+    if ("coder" in self._reader_passthrough_arguments and
+        "coder" not in writer_kwargs):
+      writer_kwargs["coder"] = (
+          self.default_coder if self.default_coder is not None else
+          coders.registry.get_coder(self.element_type))
+    writer = self._writer_class(self._file_path_prefix, **writer_kwargs)
+    # Keep a reference to the parent object so that cache does not get garbage
+    # collected while the pipeline is running.
+    writer._cache = self
+    return writer
+
+  def read(self, **kwargs):
+    """Returns an iterator over the contents of the cache.
+
+    Args:
+      kwargs (Dict[str, Any]): Arguments to be passed to the underlying reader
+          class.
+
+    Returns:
+      An iterator over the elements in the cache.
+    """
+    reader_kwargs = self._reader_kwargs.copy()
+    reader_kwargs.update(kwargs)
+    source = self.reader(**reader_kwargs)._source
+    range_tracker = source.get_range_tracker(None, None)
+    for element in source.read(range_tracker):
+      yield element
+
+  def write(self, elements):
+    """Writes a collection of elements into the cache.
+
+    Args:
+      elements (Iterable[Any]): A collection of elements to be written to cache.
+    """
+    self._num_writes += 1
+    if self.element_type is None:
+      # TODO(ostrokach): We might want to infer the element type from the first
+      # N elements, rather than reading the entire iterator.
+      elements = list(elements)
+      self.element_type = datatype_inference.infer_element_type(elements)
+    sink = self.writer()._sink
+    handle = sink.open(self._file_path_prefix)
+    try:
+      for element in elements:
+        sink.write_record(handle, element)
+    finally:
+      sink.close(handle)
+
+  def truncate(self):
+    """Removes all contents from the cache."""
+    FileSystems.delete(list(glob_files(self.file_pattern)))
+    FileSystems.create(self._file_path_prefix + ".empty").close()
+    self.element_type = None
+
+  def remove(self):
+    """Deletes the cache, including all underlying data.
+
+    The cache should not be used after this method is called.
+    """
+    self._finalizer()
+
+  @property
+  def removed(self):
+    return not self._finalizer.alive
+
+  def __del__(self):
+    self.remove()
+
+  @property
+  def file_pattern(self):
+    return self.location + '**'
+
+  @property
+  def _reader_kwargs(self):
+    reader_kwargs = {
+        k: v for k, v in self._writer_kwargs.items()
+        if k in self._reader_passthrough_arguments
+    }
+    if ("coder" in self._reader_passthrough_arguments and
+        "coder" not in reader_kwargs):
+      if self.default_coder is not None:
+        reader_kwargs["coder"] = self.default_coder
+      elif self.element_type is None:
+        reader_kwargs["coder"] = None
+      else:
+        reader_kwargs["coder"] = coders.registry.get_coder(self.element_type)
+    return reader_kwargs
+
+  @property
+  def _file_path_prefix(self):
+    return self.location + "-{:03d}".format(self._num_writes)
+
+  def _configure_finalizer(self, persist):
+    if persist:
+      return finalize(self, lambda: None)
+    else:
+      return finalize(
+          self, lambda pattern: FileSystems.delete(list(glob_files(pattern))),
+          self.file_pattern)
+
+
+class TextBasedCache(FileBasedCache):
+  """A ``PCollectionCache`` object which uses a text-based file format to store
+  the underlying data.
+  """
+  _reader_class = textio.ReadFromText
+  _writer_class = textio.WriteToText
+  _reader_passthrough_arguments = {"coder", "compression_type"}
+
+  def __init__(self, location, **writer_kwargs):
+    writer_kwargs["coder"] = SafeFastPrimitivesCoder()
+    super(TextBasedCache, self).__init__(location, **writer_kwargs)
+
+
+class TFRecordBasedCache(FileBasedCache):
+  """A ``PCollectionCache`` object which uses the TFRecord file format to store
+  the underlying data.
+  """
+
+  _reader_class = tfrecordio.ReadFromTFRecord
+  _writer_class = tfrecordio.WriteToTFRecord
+  _reader_passthrough_arguments = {"coder", "compression_type"}
+
+
+class PatchedWriter(PTransform):
+  """A wrapper over a write ``PTransform`` which sets the element_type of the
+  parent cache when the writer is expanded into a pipeline.
+  """
+
+  def __init__(self, cache, writer_class, writer_args, writer_kwargs):
+    self._cache = cache
+    self._writer_class = writer_class
+    self._writer_args = writer_args
+    self._writer_kwargs = writer_kwargs
+
+  def expand(self, pcoll):
+    if self._cache.element_type is None:
+      self._cache.element_type = pcoll.element_type
+    writer_kwargs = self._writer_kwargs.copy()
+    if "coder" in self._cache._reader_passthrough_arguments:
+      writer_kwargs["coder"] = (
+          self._cache.default_coder if self._cache.default_coder is not None
+          else coders.registry.get_coder(self._cache.element_type))
+    writer = self._writer_class(*self._writer_args, **writer_kwargs)
+    return pcoll | writer
+
+
+class SafeFastPrimitivesCoder(coders.Coder):
+  """This class add an quote/unquote step to escape special characters."""
+
+  def encode(self, value):
+    return quote(
+        coders.coders.FastPrimitivesCoder().encode(value)).encode('utf-8')
+
+  def decode(self, value):
+    return coders.coders.FastPrimitivesCoder().decode(unquote_to_bytes(value))
+
+
+def glob_files(pattern):
+  match = FileSystems.match([pattern])
+  assert len(match) == 1
+  for metadata in match[0].metadata_list:
+    yield metadata.path
--- a/sdks/python/apache_beam/runners/interactive/caching/file_based_cache.py
+++ b/sdks/python/apache_beam/runners/interactive/caching/file_based_cache.py
@@ -87,6 +87,7 @@ class FileBasedCache(PCollectionCache):
     self._num_writes = 0
     self._persist = persist
     self._finalizer = self._configure_finalizer(persist)
+    self._timestamp = 0
 
     exitsting_files = list(glob_files(self.file_pattern))
     if exitsting_files:
@@ -119,10 +120,9 @@ class FileBasedCache(PCollectionCache):
 
   @property
   def timestamp(self):
-    timestamp = 0
     for path in glob_files(self.file_pattern):
-      timestamp = max(timestamp, FileSystems.last_updated(path))
-    return timestamp
+      self._timestamp = max(self._timestamp, FileSystems.last_updated(path))
+    return self._timestamp
 
   def reader(self, **kwargs):
     """Returns a reader ``PTransform`` to be used for reading the contents of
--- a/sdks/python/apache_beam/runners/interactive/caching/file_based_cache.py
+++ b/sdks/python/apache_beam/runners/interactive/caching/file_based_cache.py
@@ -248,7 +248,12 @@ class FileBasedCache(PCollectionCache):
     return not self._finalizer.alive
 
   def __del__(self):
-    self.delete()
+    try:
+      self.delete()
+    except AttributeError:
+      # Sometimes the destructor can be called before the finalizer is
+      # configured.
+      pass
 
   @property
   def file_pattern(self):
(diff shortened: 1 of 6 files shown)


ostrokach:
retest this please

ostrokach:
retest this please

ostrokach:
retest this please

ostrokach:
retest this please

ostrokach:
Run Portable_Python PreCommit

ostrokach:
retest this please

ostrokach:
retest this please

ostrokach:
retest this please

ostrokach:
retest this please

ostrokach:
retest this please

aaltay:
@ostrokach a few suggestions:
- There is API related changes here. It would be good to make a proposal on dev@ with a design document as a starting point. There is already a similar discussion started. (/cc @KevinGG)
- Parts of the PR could be separated and I believe you already did it with assertArrayCountEqual
- It is possible to run direct runner directly on native pyhton types, (e.g. [1,2,] | PTransform) works today by auto converting the input type. We need to reconcile the change with that behavior.

ostrokach:
Hi @aaltay, just a few notes regarding your comments:
 
> * There is API related changes here. It would be good to make a proposal on dev@ with a design document as a starting point. There is already a similar discussion started. (/cc @KevinGG)

I have talked to @KevinGG to make sure that the changes that are proposed in this PR and in apache/beam#8961 and complimentary to, and would not interfere with, the changes that they are making. But your point is well taken; I will finish a design document and will circulate it on the dev@ mailing list.

> * Parts of the PR could be separated and I believe you already did it with assertArrayCountEqual.

Yes, I have tried to split this PR into independent components. The `assertArrayCountEqual` PR has been merged (apache/beam#9235) and I have rebased this PR accordingly. The `datatype_inference` PR is still outstanding (apache/beam#9281).

> * It is possible to run direct runner directly on native pyhton types, (e.g. [1,2,] | PTransform) works today by auto converting the input type. We need to reconcile the change with that behavior.

As far as I understand, we can already do this with the proposed cache objects?

```python
>> cache = filebasedcache.SafeTextBasedCache("/tmp/beam-test-")
>> [1, 2, 3, "a", "hoho"] | cache.writer()
['/tmp/beam-output-cache9-001-00000-of-00001']
>> list(cache.read())
[1, 2, 3, 'a', 'hoho']
```

ostrokach:
retest this please

ostrokach:
retest this please

ostrokach:
retest this please

sdks/python/apache_beam/runners/interactive/caching/filebasedcache.py
```diff
@@ -0,0 +1,276 @@
...
+    if self._cache.requires_coder:
+      writer_kwargs["coder"] = self._cache.coder
+    writer = self._writer_class(*self._writer_args, **writer_kwargs)
+    return pcoll | writer
+
+
+class SafeFastPrimitivesCoder(coders.Coder):
+  """This class add an quote/unquote step to escape special characters."""
+
+  # pylint: disable=deprecated-urllib-function
```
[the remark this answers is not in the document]
ostrokach: Sorry, forgot to remove this when copy-pasting :).

sdks/python/apache_beam/runners/interactive/caching/filebasedcache.py
```diff
@@ -0,0 +1,276 @@
...
+
+  def encode(self, value):
+    return quote(
```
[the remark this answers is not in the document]
ostrokach: @qinyeli may be able to provide more feedback, since she is the one who first addressed the issue. 

From my understanding, the `apache_bem.coders.StrUtf8Coder` coder supports writing string data only, while we want to be able to serialize any data format. There is `apache_bem.coders.FastPrimitivesCoder`, but if we write the output of `FastPrimitivesCoder.encode(element)` directly to a file, we lose the information about when a given element stops and a new element starts. 

For example:

```python
temp = io.BytesIO()
coder = beam.coders.FastPrimitivesCoder()

data = [{"a": 11}]

for element in data:
    element_encoded = coder.encode(element)
    print("in", element, element_encoded)
    temp.write(element_encoded)
    
temp.seek(0)
for line in temp.read().split(b"\n"):
    print("out", coder.decode(line), line)
```

prints:

```raw
in {'a': 11} b'\x07\x01\x04\x01a\x01\x0b'
out {'a': 11} b'\x07\x01\x04\x01a\x01\x0b'
```

but the following code:

```python
temp = io.BytesIO()
coder = beam.coders.FastPrimitivesCoder()

data = [{"a": 10}]

for element in data:
    element_encoded = coder.encode(element)
    print("in", element, element_encoded)
    temp.write(element_encoded)
    
temp.seek(0)
for line in temp.read().split(b"\n"):
    print("out", coder.decode(line), line)
```

prints:

```raw
in {'a': 10} b'\x07\x01\x04\x01a\x01\n'
out {'a': 0} b'\x07\x01\x04\x01a\x01'
out None b''
```.

`apache_bem.coders.FastPrimitivesCoder` also fails for numpy arrays.

File: sdks/python/apache_beam/runners/interactive/caching/filebasedcache.py [hunk 3]
```diff
@@ -0,0 +1,276 @@
...
+
+class TextBasedCache(FileBasedCache):
+
+  _reader_class = textio.ReadFromText
+  _writer_class = textio.WriteToText
+  _reader_passthrough_arguments = {"coder", "compression_type"}
+  requires_coder = True
+
+  def __init__(self, location, **writer_kwargs):
+    warnings.warn("TextBasedCache is not reliable and should not be used.")
```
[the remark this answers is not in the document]
ostrokach: Yes, I agree.

sdks/python/apache_beam/runners/interactive/caching/filebasedcache.py
```diff
@@ -0,0 +1,276 @@
...
+        if k in self._reader_passthrough_arguments
+    }
+    return reader_kwargs
+
+  @property
+  def _file_path_prefix(self):
+    return self.location + "-{:03d}".format(self._num_writes)
+
+
+class TextBasedCache(FileBasedCache):
```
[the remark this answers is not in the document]
ostrokach: Yes, I agree. While I work on the design document and have it reviewed, I removed from this PR all cache implementations except for `TextBasedCache` and `TFRecordBasedCache`. Those cache formats are already available to `CacheManager`, and the advantages and disadvantages of the two formats should be reasonably clear.

sdks/python/apache_beam/runners/interactive/caching/filebasedcache.py
```diff
@@ -0,0 +1,276 @@
...
+    "TextBasedCache",
+    "SafeTextBasedCache",
+    "TFRecordBasedCache",
+    "ParquetBasedCache",
+    "AvroBasedCache",
+    "SafeFastPrimitivesCoder",
+]
+
+
+class FileBasedCache(PCollectionCache):
```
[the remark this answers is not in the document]
ostrokach: The idea is for these cache objects to be thin wrappers over `WriteToXXX` and `ReadFromXXX` and to be used in a similar way. Therefore, there should be a single cache for all workers, and concurrent writes are discouraged (unless they are by different workers in the same Beam pipeline).

I tried to make the behavior of `FileBasedCache` and `StreamBasedCache` similar, in that repeated and concurrent writes append data to the cache, and repeated reads produce all available data each time.

sdks/python/apache_beam/runners/interactive/caching/filebasedcache.py
```diff
@@ -0,0 +1,276 @@
...
+    FileSystems.create(self._file_path_prefix + ".empty").close()
+
+  @property
+  def persist(self):
+    return self._persist
+
+  @persist.setter
+  def persist(self, persist):
+    self._persist = persist
+    self._finalizer = (lambda: None) if self._persist else finalize(
```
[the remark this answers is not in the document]
ostrokach: Yes, thank you, this was a serious mistake on my part! I fixed the persist logic and added a test for the intended behavior.

The idea behind `cache.persist = True` is to make it possible to persist cache between sessions. With Jupyter notebooks, it is often the case that the user has to restart the kernel, for example if they ran out of memory or if they want to reload a dependency. In order to facilitate this use-case, we may need a way to persist cache between sessions. This could be accomplished by setting `cache.persist = True` for every cache and saving the cache instances to a `shelve`.

sdks/python/apache_beam/runners/interactive/caching/filebasedcache.py
```diff
@@ -0,0 +1,276 @@
...
+        self, lambda pattern: FileSystems.delete(list(glob_files(pattern))),
```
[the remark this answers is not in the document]
ostrokach: Yes, thank you for pointing this out. The pattern should always end with `**` in order to include files inside subfolders.

sdks/python/apache_beam/runners/interactive/caching/filebasedcache.py
```diff
@@ -0,0 +1,276 @@
...
+    for element in source.read(range_tracker):
+      yield element
+
+  def write(self, elements):
+    self._num_writes += 1
+    if self.requires_coder and self.coder is None:
+      # TODO(ostrokach): We might want to infer the element type from the first
+      # N elements, rather than reading the entire iterator.
+      elements = list(elements)
+      element_type = infer_element_type(elements)
```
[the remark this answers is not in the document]
ostrokach: Yes, this sounds reasonable. I have made changes such that `cache.element_type` is set the first time anything is written to the cache. For the coder, we now either use the coder provided by the user (stored inside `cache.default_coder`) or infer the coder from `cache.element_type` when it is needed.

ostrokach:
Hi @udim, thanks a lot for the detailed reviews or the two PRs, and sorry for the delay in responding. I have made changes that, I believe, should address most of your concerns, and I have sent a [design document]() to the Beam Dev mailing list describing the intention of these PRs.

The changes that I made were force-pushed, but you can see the diff between the current version and the version that you reviewed here: https://github.com/ostrokach/beam/commit/2029ada46c5cef6070b9aa5f168938dad980f586.

Thanks!

sdks/python/setup.py
```diff
@@ -104,6 +104,7 @@ def get_version():
 REQUIRED_PACKAGES = [
     'avro>=1.8.1,<2.0.0; python_version < "3.0"',
     'avro-python3>=1.8.1,<2.0.0; python_version >= "3.0"',
+    'backports.weakref>=1.0rc1,<1.1',
```
aaltay: I believe the decision was not to support python 2 for interactive pipelines. Do we need this here?

KevinGG:
> /cc @KevinGG

I'm looking at the PRs:
- https://github.com/apache/beam/pull/8884
- https://github.com/apache/beam/pull/8961
And the design doc from Alexey: 
https://docs.google.com/document/d/1DYWrT6GL_qDCXhRMoxpjinlVAfHeVilK5Mtf8gO6zxQ/

I can share some screenshots of my prototype data visualization with existing cache manager.
dive
screencapture-localhost-8888-lab-2019-09-12-11_07_07
The visualization (one-shot or dynamical plotting) will be independent from boundedness of PCollection. As long as the file based cache supports reading and writing as normal files, everything will work automatically.

udim:
@ostrokach FYI, I'm OOO from tomorrow until Oct 1. I've been busy with trying to fix type hints for the 2.16 release, but hopefully I'll have more time for reviews after I'm back.

sdks/python/apache_beam/runners/interactive/caching/pcollection_cache.py
```diff
@@ -0,0 +1,105 @@
...
+      List[Any]: A list of elements in the PCollections.
+    """
+    raise NotImplementedError
+
+  @abc.abstractmethod
+  def write(self):
+    raise NotImplementedError
+
+  @abc.abstractmethod
+  def truncate(self):
```
KevinGG: Will the `truncate` method mean anything special in some of the implementations other than clearing up the content? Is the naming trying to avoid conflicting with method name `remove`? Otherwise, could `clear` cache be an intuitive name?

ostrokach: My idea was that `truncate` would remove the contents of the cache and return it to a pristine state (similar to `TRUNCATE TABLE` in a database), while `remove` would destroy the cache and all underlying data, making the cache no longer usable (similar to `DROP TABLE` in a database). In the case of file-based caches, one of the distinctions is that the former would keep the folder and metadata of the cache while the latter would destroy those. In the case of pubsub-based-cache, the former would keep the cache topic while the latter would destroy it. 

I am completely open to renaming those methods. Do you think `clear` would be better than `truncate`?

KevinGG: Thanks! I think truncate is a great name with your explanation.

rohdesamuel:
Hey Alexey, I'm also working on Interactive Beam. Are you ready to merge?

ostrokach:
Yes, it should be, especially if it sounds like something that would be helpful.

As a side note, one other place where these caches might be helpful is when people wish to split up their Beam pipeline into a series of smaller pipelines, due to restrictions on Pipeline proto size in Cloud Dataflow. I ran into this a few weeks back and it was nice to have a disposable cache object instead of explicitly writing each part of the pipeline into some file with some coder / schema, etc.

rohdesamuel:
Thanks Alexey! This will be super helpful.

@pabloem can you merge this when you can please?

sdks/python/apache_beam/runners/interactive/caching/file_based_cache.py
```diff
@@ -0,0 +1,311 @@
...
+    self._num_writes += 1
+    writer_kwargs = self._writer_kwargs.copy()
+
+    if self.element_type is None:
+      writer = PatchedWriter(self, self._writer_class,
+                             (self._file_path_prefix,), writer_kwargs)
+      return writer
+
+    if ("coder" in self._reader_passthrough_arguments and
+        "coder" not in writer_kwargs):
```
aaltay: Since _init__ popped coder out of writer_kwargs, this would always be true, right?

sdks/python/apache_beam/runners/interactive/caching/file_based_cache.py
```diff
@@ -0,0 +1,311 @@
...
+    """
+    self._num_writes += 1
+    writer_kwargs = self._writer_kwargs.copy()
+
+    if self.element_type is None:
+      writer = PatchedWriter(self, self._writer_class,
+                             (self._file_path_prefix,), writer_kwargs)
+      return writer
+
+    if ("coder" in self._reader_passthrough_arguments and
```
aaltay: What defines `_reader_passthrough_arguments`? All the subclasses?

sdks/python/apache_beam/runners/interactive/caching/file_based_cache.py
```diff
@@ -0,0 +1,311 @@
...
+
+  def write(self, elements):
+    """Writes a collection of elements into the cache.
+
+    Args:
+      elements (Iterable[Any]): A collection of elements to be written to cache.
+    """
+    self._num_writes += 1
+    if self.element_type is None:
+      # TODO(ostrokach): We might want to infer the element type from the first
```
aaltay: Please reference a JIRA.

sdks/python/apache_beam/runners/interactive/caching/file_based_cache.py
```diff
@@ -0,0 +1,311 @@
...
+    # Keep a reference to the parent object so that cache does not get garbage
+    # collected while the pipeline is running.
+    reader._cache = self
+    return reader
+
+  def writer(self):
+    """Returns a writer object to be used for writing the contents of a
+    PCollection from a Beam Pipeline into the cache.
+    """
+    self._num_writes += 1
```
aaltay: Both writer and write increments this. What is it representing?

sdks/python/apache_beam/runners/interactive/caching/file_based_cache.py
```diff
@@ -0,0 +1,311 @@
...
+    try:
+      for element in elements:
+        sink.write_record(handle, element)
+    finally:
+      sink.close(handle)
+
+  def truncate(self):
+    """Removes all contents from the cache."""
+    FileSystems.delete(list(glob_files(self.file_pattern)))
+    FileSystems.create(self._file_path_prefix + ".empty").close()
```
aaltay: Why are we creating this? For incrementing the timestamp?

sdks/python/apache_beam/runners/interactive/caching/file_based_cache.py
```diff
@@ -0,0 +1,311 @@
...
+      return finalize(
+          self, lambda pattern: FileSystems.delete(list(glob_files(pattern))),
+          self.file_pattern)
+
+
+class TextBasedCache(FileBasedCache):
+  """A ``PCollectionCache`` object which uses a text-based file format to store
+  the underlying data.
+  """
+  _reader_class = textio.ReadFromText
```
aaltay: It might be better to explicitly pass configurations like reader class, writer class, etc to the base classes constructor. Otherwise implementers would not be able to tell without looking into examples what needs to be implemented in subclasses.

sdks/python/apache_beam/runners/interactive/caching/file_based_cache.py
```diff
@@ -0,0 +1,311 @@
...
+
+  def encode(self, value):
+    return quote(
+        coders.coders.FastPrimitivesCoder().encode(value)).encode('utf-8')
+
+  def decode(self, value):
+    return coders.coders.FastPrimitivesCoder().decode(unquote_to_bytes(value))
+
+
+def glob_files(pattern):
```
aaltay: Use FileSystems.match ?

aaltay:
> Thanks Alexey! This will be super helpful.
> 
> @pabloem can you merge this when you can please?

I do not believe reviews were completed on this PR yet for it to be merged.

sdks/python/apache_beam/runners/interactive/caching/pcollection_cache.py
```diff
@@ -0,0 +1,105 @@
...
+    Args:
+      **reader_kwargs: Arguments to pass to the underlying reader class.
+
+    Returns:
+      A source from which we can read a PCollection.
+    """
+    raise NotImplementedError
+
+  @abc.abstractmethod
+  def writer(self):
```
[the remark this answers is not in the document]
ostrokach: My rationale for passing `writer_kwargs` to `__init__` while passing `reader_kwargs` to `read` and `reader` was as follows:

1. We should be able to write to the same cache object repeatedly. One of the main reasons for this is to make the concept of a "cache" transferable for bounded and unbounded PCollections.
2. It does not make sense to write to the same cache object using a different set of arguments. In fact, it is very likely that this will lead to cache corruption (for example, using a different coder or schema). This is why I opted to pass `writer_kwargs` to the constructor rather than to the `writer` method.
3. In (I expect rather rare) cases, the user may want to pass additional arguments to the `read` and `reader` methods. The example I had in mind is someone wishing to read only a subset of the columns from a Parquet file while debugging, but then pass all the columns into the subsequent PTransform. We made this possible by allowing `read` and `reader` to take additional keyword arguments, although I am not 100% happy with the aesthetics of this.
4. In many cases, the arguments passed to the underlying `writer_class` must also be passed to the underlying `reader_class`. The most common examples may be `coder`, `compression_type`, `schema`, etc. In the interim, I decided to control this using the `_reader_passthrough_arguments` attribute, while ultimately it would be nicer if we could infer these from the underlying `reader_class` and `writer_class` objects themselves?

sdks/python/apache_beam/runners/interactive/caching/file_based_cache.py
ostrokach: We are creating an empty file so that the reader PTransforms (such as `ReadFromText`) return an empty PCollection when reading from an empty PCollectionCache. Otherwise, they would raise an `IOError`, which is inconsistent with the behavior observed when reading from an unbounded cache (`ReadFromPubSub` for example).

But I agree that this is hacky. Should I remove the writing of an empty file and raise a JIRA issue regarding this inconsistency instead?

sdks/python/apache_beam/runners/interactive/caching/file_based_cache.py
ostrokach: I am keeping track of `num_writes` mostly in order to create a different file extension for each subsequent write:

```python
  @property
  def _file_path_prefix(self):
    return self.cache_spec + "-{:03d}".format(self._num_writes)
```

Otherwise, I would have no way of making sure that a call to `write` does not overwrite previously-written data.

Does this sound reasonable?

sdks/python/apache_beam/runners/interactive/caching/file_based_cache.py
ostrokach: Sorry, could you elaborate a bit? The `glob_files` function does use `FileSystems.match`, but returns paths as strings rather than metadata objects, etc. This is especially helpful as an argument to `weakref.finalize`.

sdks/python/apache_beam/runners/interactive/caching/file_based_cache.py
```diff
@@ -0,0 +1,311 @@
...
+    if self._finalizer:
+      self._finalizer.detach()
+    self._finalizer = self._configure_finalizer(persist)
+
+  @property
+  def timestamp(self):
+    timestamp = 0
+    for path in glob_files(self.file_pattern):
+      timestamp = max(timestamp, FileSystems.last_updated(path))
+    return timestamp
```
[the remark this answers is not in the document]
ostrokach: No, thank you for picking this up!

I fixed the code so that timestamp does not go back to 0 after a truncation and added a test for this behavior.

sdks/python/apache_beam/runners/interactive/caching/pcollection_cache.py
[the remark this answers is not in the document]
ostrokach: Since the PCollectionCache interface is still experimental, I decided to change `write` and `writer` methods to take on an additional `writer_kwargs` argument. While I do not think that it should be used in most (or any) cases, I understand the confusion that is caused by asymmetry between the `read` and `write` methods (and between the `reader` and `writer` methods).

An alternative would be to remove the `reader_kwargs` argument from the `read` and `reader` methods, and that is also probably something that should be considered?

sdks/python/apache_beam/runners/interactive/caching/file_based_cache.py
ostrokach: The idea behind `_reader_passthrough_arguments` was to specify the set of `writer_kwargs` that should also be passed onto the reader PTransform (such as `coder`, `compression_type`, etc.).

Since this attribute has come up in several of your comments, I understand that it is confusing to the user and probably not something that they should worry about. Therefore, I hard-coded this parameter in the specific cache implementations (e.g. see the `__init__` methods of `TextBasdCache`, `TFRecordBasedCace`, etc.).

sdks/python/apache_beam/runners/interactive/caching/file_based_cache.py
ostrokach: I think point 2. can be addressed by refusing to create a cache if the *directory* specified by `cache_spec` already exists (rather than one or more files matching the `cache_spec + "*"` file pattern).

Point 1 is less important and could probably be addressed down the road by patching the base reader PTransforms.

Does this alternative sound better to you than creating `.empty` files?

sdks/python/apache_beam/runners/interactive/caching/file_based_cache.py
ostrokach: I had not considered doing it this way, but I do see that this approach has some advantages. I changed the `FileBasedCache` class to take `reader_class` and `writer_class` as arguments. Furthermore, I delegated the setting of default arguments to the child classes (e.g. using `functools.partial`), removing the need for the `_writer_kwargs` attribute.

However, in general I still think that the `PCollectionCache` interface should not include `reader_class` and `writer_class` as arguments. This is because, at least in the case of unbounded PCollections, the implementation of caches will be entirely backend-dependent (for example, `PubSubBasedCache` is likely to share little code with `KafkaBasedCache`). In those cases, having `reader_class` and `writer_class` as arguments is likely to cause confusion.

ostrokach:
retest this please

ostrokach:
retest this please

PR #8914, opened 2019-06-20
[Beam-6696] GroupIntoBatches transform for Python SDK

raheelkhan:
This PR adds GroupIntoBatches transform in the Python SDK

------------------------

--- | --- | --- | --- | ---


raheelkhan:
@aaltay

raheelkhan:
@robertwb

sdks/python/apache_beam/transforms/util.py
```diff
@@ -671,6 +679,60 @@ def WithKeys(pcoll, k):
   return pcoll | Map(lambda v: (k, v))
 
 
+@typehints.with_input_types(typehints.KV[K, V])
+class GroupIntoBatches(PTransform):
+  def __init__(self, batch_size):
+    self.batch_size = batch_size
+
+  @staticmethod
+  def of_size(batch_size):
```
aaltay: I do not believe we need `of_size` in Python. Users can directly use GroupIntoBatches().

sdks/python/apache_beam/transforms/util.py
```diff
@@ -671,6 +679,60 @@ def WithKeys(pcoll, k):
...
+    return GroupIntoBatches(batch_size)
+
+  def expand(self, pcoll):
+    input_coder = coders.registry.get_coder(pcoll)
+    if not input_coder.is_kv_coder():
```
aaltay: Do you need to check this? Type hint should ensure that this is of kv type.

sdks/python/apache_beam/transforms/util.py
```diff
@@ -671,6 +679,60 @@ def WithKeys(pcoll, k):
...
+
+
+def _pardo_group_into_batches(batch_size, input_coder):
+  ELEMENT_STATE = BagStateSpec('values', input_coder)
+  COUNT_STATE = CombiningValueStateSpec('count', input_coder, CountCombineFn())
+  EXPIRY_TIMER = TimerSpec('expiry', TimeDomain.WATERMARK)
+
+  class _GroupIntoBatchesDoFn(DoFn):
+
+    def process(self, element,
```
aaltay: State/Timers does not work for all runners in python.

Instead, see this alternative implementation (https://github.com/apache/beam/blob/11aacd2ea2e2b1fec54efc0557e016e2b7f8d944/sdks/python/apache_beam/transforms/util.py#L430) Similar to the `_WindowAwareBatchingDoFn`, you batch elements in process and emit in process and finish_bundle.

aaltay:
I will be out of office next week, adding @udim as a reviewer to follow up.

sdks/python/apache_beam/transforms/util.py
raheelkhan: I'll take a look into this, but the ticket [https://issues.apache.org/jira/browse/BEAM-6696](https://issues.apache.org/jira/browse/BEAM-6696) says that it should use state to batch across bundles.

How should I interpret this ?

Thanks

aaltay: Good point.

We can use state/timers as you did here. Mark it as experimental and add a strongly worded comment or maybe even an actual warning to say that use of this requires state/timer support from the runner.

sdks/python/apache_beam/transforms/util.py
raheelkhan: fixed

sdks/python/apache_beam/transforms/util.py
raheelkhan: fixed

sdks/python/apache_beam/transforms/util.py
raheelkhan: fixed

aaltay: Thank you, resolving this.

sdks/python/apache_beam/transforms/util.py
```diff
@@ -671,6 +681,56 @@ def WithKeys(pcoll, k):
   return pcoll | Map(lambda v: (k, v))
 
 
+@experimental()
+@typehints.with_input_types(typehints.KV[K, V])
+class GroupIntoBatches(PTransform):
```
aaltay: Could you add a pydocs comment to this transform. With, what does it do, what is its expectations, and mentioning that its experimental status?

raheelkhan: Done.

robertwb:
This seems highly redundant with the existing BatchElements; it seems better to update the latter to work well in streaming mode rather than introduce a new transform.

aaltay:
Difference as I understand is BatchElements will batch based on its estimator, works without state and timer APIs. GroupIntoBatches will batch into fixed size, works for streaming and matches the behavior of Java transform with same name.

I agree there is duplication. What is your suggestion to merging the two?

robertwb:
I suggest we update BatchElements to be backed by state, assuming we can do
so without performance degredation. (The estimator is equivalent to fixed
size when min == max.) It would also likely make sense to add auto-sizing
capabilities to Java.

(Another difference is that GroupIntoBatches requires keyed input, and
batches per (and then drops) the key. If we keep the key, we should
probably still emit it as we do for Java.)

aaltay:
That makes sense, but how would that work for runners that does not support state.

I moved your comment to https://issues.apache.org/jira/browse/BEAM-6696 so that we do not lose track here in this PR. (I will be OOO for a few weeks.)

PR #8923, opened 2019-06-21
[BEAM-7613] HadoopFileSystem can work with more than one cluster.

dmvk:
HadoopFileSystem creates underlying FileSystem (one from org.apache.hadoop) instance during it's construction. Single FileSystem instance is tied to a particular cluster (scheme + authority pair). In case we want to talk to another cluster, this fail due to FileSystem#checkPath.

 

This can be fixed by using FileSystem#get(java.net.URI, org.apache.hadoop.conf.Configuration) instead of FileSystem#newInstance(org.apache.hadoop.conf.Configuration)

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/io/hadoop-file-system/src/main/java/org/apache/beam/sdk/io/hdfs/HadoopFileSystem.java
+++ b/sdks/java/io/hadoop-file-system/src/main/java/org/apache/beam/sdk/io/hdfs/HadoopFileSystem.java
@@ -74,14 +74,19 @@ import org.slf4j.LoggerFactory;
  * </ul>
  */
 class HadoopFileSystem extends FileSystem<HadoopResourceId> {
+
   private static final Logger LOG = LoggerFactory.getLogger(HadoopFileSystem.class);
 
   @VisibleForTesting static final String LOG_CREATE_DIRECTORY = "Creating directory %s";
   @VisibleForTesting static final String LOG_DELETING_EXISTING_FILE = "Deleting existing file %s";
-  @VisibleForTesting final org.apache.hadoop.fs.FileSystem fileSystem;
 
-  HadoopFileSystem(Configuration configuration) throws IOException {
-    this.fileSystem = org.apache.hadoop.fs.FileSystem.newInstance(configuration);
+  private final String scheme;
+
+  @VisibleForTesting final Configuration configuration;
+
+  HadoopFileSystem(String scheme, Configuration configuration) {
+    this.scheme = scheme;
+    this.configuration = configuration;
   }
 
   @Override
@@ -97,7 +102,8 @@ class HadoopFileSystem extends FileSystem<HadoopResourceId> {
               matchRecursiveGlob(spec.substring(0, index + 1), spec.substring(index + 1)));
         } else {
           // normal glob
-          final FileStatus[] fileStatuses = fileSystem.globStatus(new Path(spec));
+          final Path path = new Path(spec);
+          final FileStatus[] fileStatuses = path.getFileSystem(configuration).globStatus(path);
           if (fileStatuses != null) {
             for (FileStatus fileStatus : fileStatuses) {
               metadata.add(toMetadata(fileStatus));
@@ -118,10 +124,11 @@ class HadoopFileSystem extends FileSystem<HadoopResourceId> {
 
   private Set<Metadata> matchRecursiveGlob(String directorySpec, String fileSpec)
       throws IOException {
+    final org.apache.hadoop.fs.FileSystem fs = new Path(directorySpec).getFileSystem(configuration);
     Set<Metadata> metadata = new HashSet<>();
     if (directorySpec.contains("*")) {
       // An abstract directory with a wildcard is converted to concrete directories to search.
-      FileStatus[] directoryStatuses = fileSystem.globStatus(new Path(directorySpec));
+      FileStatus[] directoryStatuses = fs.globStatus(new Path(directorySpec));
       for (FileStatus directoryStatus : directoryStatuses) {
         if (directoryStatus.isDirectory()) {
           metadata.addAll(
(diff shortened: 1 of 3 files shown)


sdks/java/io/hadoop-file-system/src/main/java/org/apache/beam/sdk/io/hdfs/HadoopFileSystem.java
```diff
@@ -313,7 +328,7 @@ protected HadoopResourceId matchNewResource(String singleResourceSpec, boolean i
 
   @Override
   protected String getScheme() {
-    return fileSystem.getScheme();
+    return "hdfs";
```
dmvk: I'm little bit concerned about this. We should probably have more registrars for different schemas. At least for `webhdfs`, which is commonly used.

lukecwik: The original intent was for a different scheme to be returned for each different hdfs configuration so a user if they wanted could register hdfs1, hdfs2, webhdfs, ...

All the registration logic is happening based in https://github.com/apache/beam/blob/553c56075b0fdcbc1a3d37d8b708697f1ee1f5de/sdks/java/io/hadoop-file-system/src/main/java/org/apache/beam/sdk/io/hdfs/HadoopFileSystemRegistrar.java#L42 and each HadoopFileSystem instance is mean to be related to a different configuration.

This way when they are writing their pipeline they would write `Read.from("hdfs1://...")`

dmvk: Got the idea. Anyway don't think, that `hdfs\d+` is a valid scheme, that hadoop understands.

If we want to support multiple clusters, using the native protocol, what matters is an authority section of the URL. I think we can introduce some kind of `AuthorityAwareFileSystem`, that would be just a container for the HadoopFileSystems with the same scheme and decide based on the authority part, which filesystem should be used.

There are some things I'm not sure about.
1) What should we do in case there are multiple configs for same scheme + authority? I think failing fast should be the best option here.
2) We are duplicating functionality already provided by Hadoop. This is an equivalent of specifying multiple `nameservice` entries in the same config. Do we want to do this, because it is kind of an unexpected API for an user coming from the Hadoop world?

lukecwik: I don't think we need to provide an `AuthorityAwareFileSystem` in Beam. As long as the HadoopFileSystem knows how to handle multiple authorities by selecting the appropriate configuration for it, we are good.

For the other schemes that HDFS currently supports such as webhdfs, ftp, ..., they should all be registered as separate HadoopFileSystem instances.

If Hadoop starts to support arbitrary schemes, the hdfs1, hdfs2, ... is meant to represent the case where you want to access the same authority but with different settings which is why I agree with failing fast when multiple configurations provide the same scheme+authority.

dmvk: If we want to do it this way, it would be a good idea to allow user to set only a single configuration, which would be a breaking change to the API, but we don't allow user to use it anyway right now.

```
    checkArgument(
        configurations.size() <= 1,
        String.format(
            "The %s currently only supports at most a single Hadoop configuration.",
            HadoopFileSystemRegistrar.class.getSimpleName()));
```

This way we wouldn't have to deal with multiple `fs.defaultFS` definitions and other possible config option conflicts.

It is probably less flexible, but we should have either this, or force user to specify a config per cluster.

WDYT?

lukecwik: I think it is ok if there are multiple configurations and we choose which one based upon authority within a single HadoopFileSystem instance. Any settings that a user puts down that are conflicting are up to them to resolve. We or Hadoop can provide helpful errors if they do this but it is really up to them to have valid hadoop configurations.

davidak09:
I can confirm that this solution works in our scenario. :+1: 

What we do:
- we have a job which writes to 2 HDFS outputs
- the first main output is on `hdfs://cluster1`
- the second output is on `hdfs://cluster2`
- both clusters are defined in a single hadoop configuration, `hdfs://cluster1` is defined as hadoop property `fs.defaultFS`

Without this fix we get an exception when we try to write to HDFS on `cluster2`:
```
org.apache.beam.sdk.util.UserCodeException: java.lang.IllegalArgumentException: Wrong FS: hdfs://cluster2/some_path, expected: hdfs://cluster1
        at org.apache.beam.sdk.util.UserCodeException.wrap(UserCodeException.java:34)
        at org.apache.beam.sdk.io.WriteFiles$WriteUnshardedTempFilesFn$DoFnInvoker.invokeProcessElement(Unknown Source)
        at org.apache.beam.runners.core.SimpleDoFnRunner.invokeProcessElement(SimpleDoFnRunner.java:213)
        at org.apache.beam.runners.core.SimpleDoFnRunner.processElement(SimpleDoFnRunner.java:175)
        ...
```
With this fix everything works just fine.

tvalentyn:
What are the next steps on this PR?

dmvk:
@tvalentyn this is still valid, will take a look at it later this week

dmvk:
I've changed HadoopFileSystemRegistrar to work as follows:

- Resolve scheme from fs.defaultFS (compatible with original implementation)
- In case dfs.nameservices are configured, we imply HA hadoop cluster and also register hdfs and webhdfs schemes

This should be backward compatible and also support multiple clusters. We leave other settings up to the user to validate.

WDYT?

dmvk:
Maybe we can log warning in case defaultFS has suspicious scheme (eg. `file://`, `s3://`, ...)

lukecwik:
Run Java PreCommit

PR #8731, opened 2019-05-31
[BEAM-7416] MetricsReader in Python should save only supported metrics

kkucharc:
Fix that currently MetricsReader is saving all metrics, not only those which are required for load tests (if happens that user collects some other metrics too).

------------------------

Thank you for your contribution! Follow this checklist to help us incorporate your contribution quickly and easily:

 - [x] [**Choose reviewer(s)**](https://beam.apache.org/contribute/#make-your-change) and mention them in a comment (`R: @username`).
 - [x] Format the pull request title like `[BEAM-XXX] Fixes bug in ApproximateQuantiles`, where you replace `BEAM-XXX` with the appropriate JIRA issue, if applicable. This will automatically link the pull request to the issue.
 - [ ] If this contribution is large, please file an Apache [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

Lang | SDK | Apex | Dataflow | Flink | Gearpump | Samza | Spark
--- | --- | --- | --- | --- | --- | --- | ---
Go | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go/lastCompletedBuild/) | --- | --- | --- | --- | --- | ---
Java | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Apex/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Dataflow/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Flink/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Flink_Batch/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Flink_Streaming/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Gearpump/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Samza/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Spark/lastCompletedBuild/)
[...]

PR #8941, opened 2019-06-25
[BEAM-7528] Save load test metrics according to distribution name

kkucharc:
In case there are collected different metrics than the one triggered by load tests, the runtime metric was not correct calculated. All other metrics are saved as well.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/testing/load_tests/load_test_metrics_utils.py
+++ b/sdks/python/apache_beam/testing/load_tests/load_test_metrics_utils.py
@@ -31,6 +31,7 @@ Currently it is possible to have following metrics types:
 
 from __future__ import absolute_import
 
+from itertools import groupby
 import logging
 import time
 import uuid
@@ -111,11 +112,15 @@ class MetricsReader(object):
       insert_rows.append(counter_dict)
 
     dists = metrics['distributions']
-    if len(dists) > 0:
-      runtime = RuntimeMetric(dists, submit_timestamp, metric_id)\
-        .as_dict()
-      insert_rows.append(runtime)
-
+    sorted_dists = sorted(dists, key=lambda x: x.key.metric.name)
+    for key, dists in groupby(sorted_dists, lambda x: x.key.metric.name):
+      dists = list(dists)
+      if key == RUNTIME_METRIC:
+        metric = RuntimeMetric(dists, submit_timestamp, metric_id)
+        insert_rows.append(metric.as_dict())
+      else:
+        metrics = DistributionMetrics(dists, submit_timestamp, metric_id, key).values_list
+        insert_rows += metrics
     return insert_rows
 
 
@@ -138,15 +143,33 @@ class Metric(object):
 class CounterMetric(Metric):
   def __init__(self, counter_dict, submit_timestamp, metric_id):
     super(CounterMetric, self).__init__(submit_timestamp, metric_id)
-    self.value = counter_dict.committed
     self.label = str(counter_dict.key.metric.name)
+    self.value = counter_dict.committed
 
+
+class DistributionMetrics(Metric):
+  def __init__(self, dists, submit_timestamp, metric_id, key):
+    super(DistributionMetrics, self).__init__(submit_timestamp, metric_id)
+    self.dist_label = str(key)
+    self.values_list = self._get_list(dists)
+
+  def _get_list(self, dists):
+    labels = ['mean', 'max', 'min', 'sum']
+    list = []
+    for dist in dists:
+      for label in labels:
+        self.label = self.dist_label + '_' +  label
+        self.value = getattr(dist.committed,label)
+        list.append(self.as_dict())
+
+    return list
 
 class RuntimeMetric(Metric):
   def __init__(self, runtime_list, submit_timestamp, metric_id):
     super(RuntimeMetric, self).__init__(submit_timestamp, metric_id)
-    self.value = self._prepare_runtime_metrics(runtime_list)
     self.label = RUNTIME_METRIC
+    self.value = self._prepare_runtime_metrics(runtime_list)
+
 
   def _prepare_runtime_metrics(self, distributions):
     min_values = []


kkucharc:
Run Python Load Tests Smoke

kkucharc:
Run Python Load Tests Smoke

kkucharc:
Run Python PreCommit

kkucharc:
Run Python Load Tests Smoke

kkucharc:
Run Python Load Tests Smoke

kkucharc:
Run Python PreCommit

kkucharc:
Run Python Load Tests Smoke

kkucharc:
Run Python PreCommit

kkucharc:
R: @tvalentyn can you take a look? cc: @pabloem cc: @kamilwu

sdks/python/apache_beam/testing/load_tests/load_test_metrics_utils.py
```diff
@@ -111,11 +112,15 @@ def _prepare_all_metrics(self, metrics):
       insert_rows.append(counter_dict)
 
     dists = metrics['distributions']
-    if len(dists) > 0:
-      runtime = RuntimeMetric(dists, submit_timestamp, metric_id)\
-        .as_dict()
-      insert_rows.append(runtime)
-
+    sorted_dists = sorted(dists, key=lambda x: x.key.metric.name)
```
kamilwu: For a performance reason, it'd be worth to use `attrgetter` from `operator` module as a key parameter. 
The same applies to the instruction below.

tvalentyn: If size of `dists` is not that large, we can as well keep current version as it may be more readable.

sdks/python/apache_beam/testing/load_tests/load_test_metrics_utils.py
```diff
@@ -138,8 +143,25 @@ def as_dict(self):
...
+
+class DistributionMetrics(Metric):
+  def __init__(self, dists, submit_timestamp, metric_id, key):
+    super(DistributionMetrics, self).__init__(submit_timestamp, metric_id)
+    self.dist_label = str(key)
+    self.values_list = self._get_list(dists)
+
+  def _get_list(self, dists):
+    labels = ['mean', 'max', 'min', 'sum']
+    list = []
```
tvalentyn: list is also a type, let's use a different variable name.

sdks/python/apache_beam/testing/load_tests/load_test_metrics_utils.py
```diff
@@ -138,8 +143,25 @@ def as_dict(self):
...
     super(CounterMetric, self).__init__(submit_timestamp, metric_id)
-    self.value = counter_dict.committed
     self.label = str(counter_dict.key.metric.name)
+    self.value = counter_dict.committed
+
+
+class DistributionMetrics(Metric):
+  def __init__(self, dists, submit_timestamp, metric_id, key):
+    super(DistributionMetrics, self).__init__(submit_timestamp, metric_id)
+    self.dist_label = str(key)
```
tvalentyn: Can we expect callers to pass an `str` to this constructor to avoid input type ambiguity and str conversion here?

sdks/python/apache_beam/testing/load_tests/load_test_metrics_utils.py
```diff
@@ -138,8 +143,25 @@ def as_dict(self):
 class CounterMetric(Metric):
   def __init__(self, counter_dict, submit_timestamp, metric_id):
     super(CounterMetric, self).__init__(submit_timestamp, metric_id)
-    self.value = counter_dict.committed
     self.label = str(counter_dict.key.metric.name)
+    self.value = counter_dict.committed
+
+
+class DistributionMetrics(Metric):
+  def __init__(self, dists, submit_timestamp, metric_id, key):
```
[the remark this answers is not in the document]
tvalentyn: also do multiple dists have the same dist label? I also wonder what is the relationship between label in metric_id - is there 1-1 correspondence?

sdks/python/apache_beam/testing/load_tests/load_test_metrics_utils.py
```diff
@@ -138,8 +143,25 @@ def as_dict(self):
...
+    super(DistributionMetrics, self).__init__(submit_timestamp, metric_id)
+    self.dist_label = str(key)
+    self.values_list = self._get_list(dists)
+
+  def _get_list(self, dists):
```
tvalentyn: Consider a more mnemonic name for this helper - a reader of the code would have to look into implementation to understand what this method does.

sdks/python/apache_beam/testing/load_tests/load_test_metrics_utils.py
```diff
@@ -138,8 +143,25 @@ def as_dict(self):
 class CounterMetric(Metric):
   def __init__(self, counter_dict, submit_timestamp, metric_id):
     super(CounterMetric, self).__init__(submit_timestamp, metric_id)
-    self.value = counter_dict.committed
     self.label = str(counter_dict.key.metric.name)
+    self.value = counter_dict.committed
+
+
+class DistributionMetrics(Metric):
```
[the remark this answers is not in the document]
kkucharc: You are right, naming here isn't very accurate. To give more context:
Metrics such as counters or the runtime metric are just one value. Runtime is evaluated from multiple distribution metrics, but distribution metric has 4 values inside (sum, mean, max, min). There was situation when metrics weren't introduced exactly in test but in the pipelines in project (it happened in tfx project) and I wasn't sure what should be saved in those cases, so I decided that we should save everything.
Also it's not a dumb question :)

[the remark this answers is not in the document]
kkucharc: [Here](https://github.com/apache/beam/pull/8731) was decision about saving all metrics (not only specified by user in test).

kkucharc:
Run Python Load Tests Combine Dataflow Batch

sdks/python/apache_beam/testing/load_tests/load_test_metrics_utils.py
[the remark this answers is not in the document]
kkucharc: Renaming it sounds reasonable. I'm changing it this way.

Your comment made me think that I may be wrong by assuming there is no problem if someone would add measurements to the same distribution several times during one pipeline. In BQ they would be redundantly saved with same `id`, `label` and `timestamp` but with different `value`. It would be good to differ them somehow to know how a saved distribution looks like.

sdks/python/apache_beam/testing/load_tests/load_test_metrics_utils.py
kkucharc: You are right. Thank you.

sdks/python/apache_beam/testing/load_tests/load_test_metrics_utils.py
kkucharc: Sure, makes sense.

sdks/python/apache_beam/testing/load_tests/load_test_metrics_utils.py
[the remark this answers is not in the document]
tvalentyn: Have you considered saving median instead of mean? a single outlier may be affecting mean for the time window it is considered. Evaluating mean (aka 50th percentile) would produce a more reliable signal. One can also consider other percentiles, p90, p99 if necessary.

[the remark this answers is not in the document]
tvalentyn: How do you capture the time window over which you collected an aggregate metric (min/max)?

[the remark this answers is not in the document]
kkucharc: I understand your point, calculating one value might "save" some space in BigQuery. Unfortunately distribution metrics doesn't support with median or other percentiles out of the box (maybe it's a good idea for feature request?) and I am not sure if it's good idea to introduce such logic in util for reading metrics. Mainly because in the situation when we would like save some external (not expected in MetricsReader) metrics which we don't know what mean (but still may be valuable for the others ex. tfx team) calculating median or any aggregates may cause loosing meaningful info (maybe someone who defined those external metrics needs max, min, sum or mean). Do you agree?

As it comes to time windows, also distribution doesn't store such thing. It is possible to retrieve the step name in which metrics would be collected. And this I can add to metric label we save.

This is confusing problem. To sum everything up:
- What to do with metrics that weren't collected in our pipeline but somewhere deeper? It is decided to save them. But I can suggest here that it would help to have pipeline option called `save_external_metrics`. WDYT?
- In what shape we should save metrics that we don't know nothing about them? IMO probably as raw as possible, because we don't know what is useful for user that collects them.

sdks/python/apache_beam/testing/load_tests/load_test_metrics_utils.py
kkucharc: This is good advice Kamil, thank you. But I must agree with Valentyn. Because of rather small arrays we won't get much out of performance by using `attrgetter` and as I checked how it would look like it may be a little bit less readable.

sdks/python/apache_beam/testing/load_tests/load_test_metrics_utils.py
```diff
@@ -111,11 +112,15 @@ def _prepare_all_metrics(self, metrics):
...
     dists = metrics['distributions']
-    if len(dists) > 0:
-      runtime = RuntimeMetric(dists, submit_timestamp, metric_id)\
-        .as_dict()
-      insert_rows.append(runtime)
-
+    sorted_dists = sorted(dists, key=lambda x: x.key.metric.name)
+    for key, dists in groupby(sorted_dists, lambda x: x.key.metric.name):
+      dists = list(dists)
+      if key == RUNTIME_METRIC:
```
tvalentyn: It will be more robust to check group by metric namespace+name, not just name.

sdks/python/apache_beam/testing/load_tests/load_test_metrics_utils.py
tvalentyn: re line 110: let's use brackets for implicit line continuation and remove '\'. From PEP-0008:
> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces

sdks/python/apache_beam/testing/load_tests/load_test_metrics_utils.py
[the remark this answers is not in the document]
tvalentyn: I see, looks like saving median would require changes to beam. Regarding unknown distributions, I see following options:
1) Save sum, count, min, max, collected across all distributions associated with a metric with the same name.  In other words, total sum, total count, min of minimums and max of maximums. 
2) Save `(sum, count, min, max)` for each distribution without aggregation (not sure if useful, but since we are exporting to bigquery, we can run a query later to aggregate something we need).
3) register aggregators with an instance of MetricsReader. Registration will take the name of the metric, and an aggregator instance that recieves a list of all distribution readings, and computes desired output to store into Bigquery. 

I think 2 is most flexible, but perhaps 1) is sufficient and will cover most of the use-cases. Feel free to ask @robertwb or @pabloem for an opinion. 3 seems like overengineering at this point.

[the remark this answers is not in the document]
pabloem: I am okay with either 1 or 2. Do we have full metric names( step+namespace+metric)?

sdks/python/apache_beam/testing/load_tests/load_test_metrics_utils.py
kkucharc: That's right. I changes with all similar cases.

sdks/python/apache_beam/testing/load_tests/load_test_metrics_utils.py
kkucharc: Sure. Makes sense. I had to divide list into 2 (with our namespace and "other" namespaces).

sdks/python/apache_beam/testing/load_tests/load_test_metrics_utils.py
```diff
@@ -138,8 +143,25 @@ def as_dict(self):
 class CounterMetric(Metric):
   def __init__(self, counter_dict, submit_timestamp, metric_id):
     super(CounterMetric, self).__init__(submit_timestamp, metric_id)
-    self.value = counter_dict.committed
     self.label = str(counter_dict.key.metric.name)
+    self.value = counter_dict.committed
```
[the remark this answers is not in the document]
kkucharc: Sure, I'm adding docstrings everywhere

sdks/python/apache_beam/testing/load_tests/load_test_metrics_utils.py
kkucharc: Ok, actually I removed it.

sdks/python/apache_beam/testing/load_tests/load_test_metrics_utils.py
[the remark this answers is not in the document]
kkucharc: Ok, so I changed it into option 2. I changes it also to have name namespace+step(where possible)+metric.

kkucharc:
@tvalentyn I think I applied all your advices from comments. Can you take a look again?

kkucharc:
Valentyn is OOO this week, @pabloem do you mind checking this?

sdks/python/apache_beam/testing/load_tests/load_test_metrics_utils.py
```diff
@@ -75,17 +73,114 @@
...
+    metric_type(str): type of value from distribution metric which will
+      be saved (ex. max, min, mean, sum)
+    submit_timestamp: timestamp when metric is saved
+    dist(object) distribution object from pipeline result
+    metric_id(uuid): id of the current test run
+
+  Returns:
+    dictionary prepared for saving according to schema
+  """
+  return DistributionMetric(dist, submit_timestamp, metric_id,
```
pabloem: We create a generic distribution here, right? Should we tag them as generic somehow? Maybe it's just the fact that it has a different name thatn the distributions we do care about. Just wondering.

PR #7061, opened 2018-11-16
[BEAM-5191] Support for BigQuery clustering

wscheep:
Implemented BigQuery clustering: https://cloud.google.com/bigquery/docs/clustered-tables.

As this is related to BigQuery TimePartitioning, I based my implementation on this commit:
https://github.com/apache/beam/commit/b0e03a33cf0c2c573a2d34d88506e19ebb28c934

As far as I know, there are no integration tests covering time partitioning, so I did not add tests for clustering. If needed I can write some if someone points me in the right direction.

This is my first feature PR, so I'm eager to get some proper feedback.
@jkff, @reuvenlax as you committed & authored time partitioning, can you have a look?

Thanks,
Wout

------------------------

Follow this checklist to help us incorporate your contribution quickly and easily:

 - [ ] Format the pull request title like `[BEAM-XXX] Fixes bug in ApproximateQuantiles`, where you replace `BEAM-XXX` with the appropriate JIRA issue, if applicable. This will automatically link the pull request to the issue.
 - [ ] If this contribution is large, please file an Apache [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

It will help us expedite review of your Pull Request if you tag someone (e.g. `@username`) to look at it.

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

Lang | SDK | Apex | Dataflow | Flink | Gearpump | Samza | Spark
--- | --- | --- | --- | --- | --- | --- | ---
Go | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_GradleBuild/lastCompletedBuild/) | --- | --- | --- | --- | --- | ---
[...]

PR #8945, opened 2019-06-25
[BEAM-5191] Support for BigQuery clustering

jklukas:
This takes the commits from #7061, rebases on master, and adds an `enableClustering` method to allow users to opt in to the updated coder when using dynamic destinations.

This follows the pattern of #6914 where we added a new version of `MetadataCoder`, documenting how a user could opt in and why they might want to.

This should solve the coder compatibility concerns that were the blockers to merging #7061. 

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryIO.java
+++ b/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryIO.java
@@ -25,6 +25,7 @@ import static org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Prec
 import static org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Preconditions.checkState;
 
 import com.google.api.client.json.JsonFactory;
+import com.google.api.services.bigquery.model.Clustering;
 import com.google.api.services.bigquery.model.Job;
 import com.google.api.services.bigquery.model.JobConfigurationQuery;
 import com.google.api.services.bigquery.model.JobReference;
@@ -1529,6 +1530,9 @@ public class BigQueryIO {
     @Nullable
     abstract ValueProvider<String> getJsonTimePartitioning();
 
+    @Nullable
+    abstract Clustering getClustering();
+
     abstract CreateDisposition getCreateDisposition();
 
     abstract WriteDisposition getWriteDisposition();
@@ -1597,6 +1601,8 @@ public class BigQueryIO {
       abstract Builder<T> setJsonSchema(ValueProvider<String> jsonSchema);
 
       abstract Builder<T> setJsonTimePartitioning(ValueProvider<String> jsonTimePartitioning);
+
+      abstract Builder<T> setClustering(Clustering clustering);
 
       abstract Builder<T> setCreateDisposition(CreateDisposition createDisposition);
 
@@ -1732,6 +1738,10 @@ public class BigQueryIO {
     /**
      * Writes to table specified by the specified table function. The table is a function of {@link
      * ValueInSingleWindow}, so can be determined by the value or by the window.
+     *
+     * <p>If the function produces destinations configured with clustering fields, ensure that
+     * {@link #withClustering()} is also set so that the clustering configurations get properly
+     * encoded and decoded.
      */
     public Write<T> to(
         SerializableFunction<ValueInSingleWindow<T>, TableDestination> tableFunction) {
@@ -1739,7 +1749,13 @@ public class BigQueryIO {
       return toBuilder().setTableFunction(tableFunction).build();
     }
 
-    /** Writes to the table and schema specified by the {@link DynamicDestinations} object. */
+    /**
+     * Writes to the table and schema specified by the {@link DynamicDestinations} object.
+     *
+     * <p>If any of the returned destinations are configured with clustering fields, ensure that the
+     * passed {@link DynamicDestinations} object returns {@link TableDestinationCoderV3} when {@link
+     * DynamicDestinations#getDestinationCoder()} is called.
+     */
     public Write<T> to(DynamicDestinations<T, ?> dynamicDestinations) {
       checkArgument(dynamicDestinations != null, "dynamicDestinations can not be null");
       return toBuilder().setDynamicDestinations(dynamicDestinations).build();
@@ -1798,7 +1814,7 @@ public class BigQueryIO {
      * Allows newly created tables to include a {@link TimePartitioning} class. Can only be used
      * when writing to a single table. If {@link #to(SerializableFunction)} or {@link
      * #to(DynamicDestinations)} is used to write dynamic tables, time partitioning can be directly
-     * in the returned {@link TableDestination}.
+     * set in the returned {@link TableDestination}.
      */
     public Write<T> withTimePartitioning(TimePartitioning partitioning) {
       checkArgument(partitioning != null, "partitioning can not be null");
@@ -1820,6 +1836,34 @@ public class BigQueryIO {
     public Write<T> withJsonTimePartitioning(ValueProvider<String> partitioning) {
       checkArgument(partitioning != null, "partitioning can not be null");
       return toBuilder().setJsonTimePartitioning(partitioning).build();
+    }
+
+    /**
+     * Specifies the clustering fields to use when writing to a single output table. Can only be
+     * used when {@link#withTimePartitioning(TimePartitioning)} is set. If {@link
+     * #to(SerializableFunction)} or {@link #to(DynamicDestinations)} is used to write to dynamic
+     * tables, the fields here will be ignored; call {@link #withClustering()} instead.
+     */
+    public Write<T> withClustering(Clustering clustering) {
+      checkArgument(clustering != null, "clustering can not be null");
+      return toBuilder().setClustering(clustering).build();
+    }
+
+    /**
+     * Allows writing to clustered tables when {@link #to(SerializableFunction)} or {@link
+     * #to(DynamicDestinations)} is used. The returned {@link TableDestination} objects should
+     * specify the time partitioning and clustering fields per table. If writing to a single table,
+     * use {@link #withClustering(Clustering)} instead to pass a {@link Clustering} instance that
+     * specifies the static clustering fields to use.
+     *
+     * <p>Setting this option enables use of {@link TableDestinationCoderV3} which encodes
+     * clustering information. The updated coder is compatible with non-clustered tables, so can be
+     * freely set for newly deployed pipelines, but note that pipelines using an older coder must be
+     * drained before setting this option, since {@link TableDestinationCoderV3} will not be able to
+     * read state written with a previous version.
+     */
+    public Write<T> withClustering() {
+      return toBuilder().setClustering(new Clustering()).build();
     }
 
     /** Specifies whether the table should be created if it does not exist. */
@@ -2093,6 +2137,11 @@ public class BigQueryIO {
             "The supplied getTableFunction object can directly set TimePartitioning."
                 + " There is no need to call BigQueryIO.Write.withTimePartitioning.");
       }
+      if (getClustering() != null && getClustering().getFields() != null) {
+        checkArgument(
+            getJsonTimePartitioning() != null,
+            "Clustering fields can only be set when TimePartitioning is set.");
+      }
 
       DynamicDestinations<T, ?> dynamicDestinations = getDynamicDestinations();
       if (dynamicDestinations == null) {
@@ -2101,7 +2150,8 @@ public class BigQueryIO {
               DynamicDestinationsHelpers.ConstantTableDestinations.fromJsonTableRef(
                   getJsonTableRef(), getTableDescription());
         } else if (getTableFunction() != null) {
-          dynamicDestinations = new TableFunctionDestinations<>(getTableFunction());
+          dynamicDestinations =
+              new TableFunctionDestinations<>(getTableFunction(), getClustering() != null);
         }
 
         // Wrap with a DynamicDestinations class that will provide a schema. There might be no
@@ -2122,7 +2172,8 @@ public class BigQueryIO {
           dynamicDestinations =
               new ConstantTimePartitioningDestinations<>(
                   (DynamicDestinations<T, TableDestination>) dynamicDestinations,
-                  getJsonTimePartitioning());
+                  getJsonTimePartitioning(),
+                  StaticValueProvider.of(BigQueryHelpers.toJsonString(getClustering())));
         }
       }
       return expandTyped(input, dynamicDestinations);
--- a/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/CreateTables.java
+++ b/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/CreateTables.java
@@ -130,6 +130,16 @@ public class CreateTables<DestinationT, ElementT>
           dynamicDestinations,
           tableDestination,
           destination);
+      boolean destinationCoderSupportsClustering =
+          !(dynamicDestinations.getDestinationCoder() instanceof TableDestinationCoderV2);
+      checkArgument(
+          tableDestination.getClustering() == null || destinationCoderSupportsClustering,
+          "DynamicDestinations.getTable() may only return destinations with clustering configured"
+              + " if a destination coder is supplied that supports clustering, but %s is configured"
+              + " to use TableDestinationCoderV2. Set withClustering() on BigQueryIO.write() and, "
+              + " if you provided a custom DynamicDestinations instance, override"
+              + " getDestinationCoder() to return TableDestinationCoderV3.",
+          dynamicDestinations);
       TableReference tableReference = tableDestination.getTableReference().clone();
       if (Strings.isNullOrEmpty(tableReference.getProjectId())) {
         tableReference.setProjectId(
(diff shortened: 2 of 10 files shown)


jklukas:
R: @juancho088 @reuvenlax @alexvanboxel @chamikaramj who were reviewers on #7061

Also cc @wscheep who authored #7061 which is still the bulk of the code here.

jklukas:

sdks/java/io/google-cloud-platform/src/test/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryIOWriteTest.java
```diff
@@ -391,6 +392,93 @@ public void testTimePartitioning(BigQueryIO.Write.Method insertMethod) throws Ex
...
+                      return new TableDestination(
+                          tableSpec,
+                          null,
+                          new TimePartitioning().setType("DAY").setField("date"),
+                          new Clustering().setFields(ImmutableList.of("date")));
+                    })
+                .withTestServices(fakeBqServices)
+                .withMethod(BigQueryIO.Write.Method.FILE_LOADS)
+                .withSchema(schema)
+                .enableClustering()
```
jklukas: Notably, this test fails if `enableClustering()` is not called because the default `TableDestinationCoderV2` is used and clustering information is dropped before the table is created. This is exactly the behavior we want for backwards compatibility.

sdks/java/io/google-cloud-platform/src/test/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryTimePartitioningClusteringIT.java
```diff
@@ -0,0 +1,150 @@
...
+  @Test
+  public void testE2EBigQueryClustering() throws Exception {
+    String tableName = "weather_stations_clustered_" + System.currentTimeMillis();
+
+    Pipeline p = Pipeline.create(options);
+
+    p.apply(BigQueryIO.readTableRows().from(options.getInput()))
+        .apply(ParDo.of(new KeepStationNumberAndConvertDate()))
+        .apply(
+            BigQueryIO.writeTableRows()
```
chamikaramj: Can we also add one with dynamic destinations and clustering ?

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryIO.java
```diff
@@ -1793,6 +1814,32 @@ static String getExtractDestinationUri(String extractDestinationDir) {
...
+     * #to(DynamicDestinations)} is used. The returned {@link TableDestination} objects should
+     * specify the clustering fields per table. If writing to a single table, use {@link
+     * #withClustering(Clustering)} instead to specify the clustering fields.
+     *
+     * <p>Setting this option enables use of {@link TableDestinationCoderV3} which encodes
+     * clustering information. Pipelines using an older coder must be drained before setting this
+     * option, since {@link TableDestinationCoderV3} will not be able to read state written with a
+     * previous version.
+     */
+    public Write<T> enableClustering() {
```
chamikaramj: Having these two methods seems to make the API pretty brittle.

How about just having one function withClustering() that optionally takes a clustering object ? In dynamic destinations case optional Clustering object can be skipped/null and the method will behave similar to enableClustering() ?

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/DynamicDestinationsHelpers.java
```diff
@@ -167,7 +175,12 @@ public TableDestination getTable(DestinationT destination) {
     @Override
     Coder<DestinationT> getDestinationCoderWithDefault(CoderRegistry registry)
         throws CannotProvideCoderException {
-      return inner.getDestinationCoderWithDefault(registry);
+      Coder<DestinationT> destinationCoder = getDestinationCoder();
```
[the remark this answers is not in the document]
jklukas: `DynamicDestinations#getDestinationCoderWithDefault` is commented as:

```
  // Gets the destination coder. If the user does not provide one, try to find one in the coder
  // registry. If no coder can be found, throws CannotProvideCoderException.
```

This code is written with potentially multiple layers of delegation, and I think the correct behavior here is to return the first non-delegated implementation of `getDestinationCoder()` that appears as we move down the delegation chain.

I would argue that the existing behavior is incorrect. Currently, if an implementing class defines a custom return value for `getDestinationCoder`, that value is ignored when you call `getDestinationCoderWithDefault`. My expectation is that `getDestinationCoderWithDefault` would always return the same value as `getDestinationCoder` except in the null case, in which `getDestinationCoderWithDefault` would then attempt to look up a coder in the registry.

So the change here is intended to fix broken behavior. 

It's possible that a user has written a custom class that extends DelegatingDynamicDestinations and relies on the incorrect behavior, but it feels unlikely to me.

For the scope of the coders provided here, I don't believe this change affects behavior (the method was already returning TableDestinationCoderV2 in all cases).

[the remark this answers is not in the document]
jklukas: It may actually be better to remove this override altogether. `getDestinationCoderWithDefault` is defined as package-private and probably shouldn't be overriden. Instead, we let `getDestinationCoder` handle delegation, and the base implementation of `getDestinationCoderWithDefault` will make the appropriate `getDestinationCoder` call before falling back to the registry.

sdks/java/io/google-cloud-platform/src/test/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryTimePartitioningClusteringIT.java
```diff
@@ -0,0 +1,150 @@
...
+                .to(String.format("%s.%s", datasetName, tableName))
+                .withTimePartitioning(getTimepartitioning())
+                .withSchema(getSchema())
+                .withCreateDisposition(BigQueryIO.Write.CreateDisposition.CREATE_IF_NEEDED)
+                .withWriteDisposition(BigQueryIO.Write.WriteDisposition.WRITE_TRUNCATE));
+
+    p.run().waitUntilFinish();
```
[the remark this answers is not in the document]
jklukas: I believe these ITs are the work of @wscheep in the previous PR, and I am going to borrow liberally from these examples to add the dynamic destinations ITs.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryIO.java
jklukas: Implemented. `enableClustering()` is now `withClustering()` and it sets `clustering` to a default `Clustering` instance with no fields set, which we use as a flag for enabling clustering on dynamic destinations.

sdks/java/io/google-cloud-platform/src/test/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryTimePartitioningClusteringIT.java
jklukas: Added. It doesn't look like the ITs are run by Jenkins by default and I also can't find that there's a command to kick off BigQueryIO integration tests. Do you have any advice about how we can get results for these?

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/DynamicDestinationsHelpers.java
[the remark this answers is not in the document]
jklukas: I've removed the override getDestinationCoderWithDefault completely.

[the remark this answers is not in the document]
jklukas: > I've removed the override getDestinationCoderWithDefault completely.

Looks like that caused some regressions where destination coder couldn't be found from type. I'm looking into it.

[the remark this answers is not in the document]
jklukas: The override turns out to be necessary because the `DynamicDestinations` instance itself is eventually passed into `extractFromTypeParameters` and the inner class potentially has richer type information than the delegating class.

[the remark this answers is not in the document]
jklukas: I've put the override method back in place, including the modification of first checking if we have a non-null return from `getDestinationCoder()`.

chamikaramj:
Run Java PostCommit

sdks/java/io/google-cloud-platform/src/test/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryTimePartitioningClusteringIT.java
chamikaramj: Post-commit test suite should capture these. Just triggered it.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/DynamicDestinationsHelpers.java
```diff
@@ -165,8 +173,12 @@ public TableDestination getTable(DestinationT destination) {
     }
 
     @Override
-    Coder<DestinationT> getDestinationCoderWithDefault(CoderRegistry registry)
+    public Coder<DestinationT> getDestinationCoderWithDefault(CoderRegistry registry)
```
chamikaramj: Was change of modifier here (to public) intentional ? Wasn't sure from your explanation below.

sdks/java/io/google-cloud-platform/src/test/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryIOWriteTest.java
```diff
@@ -391,6 +392,93 @@ public void testTimePartitioning(BigQueryIO.Write.Method insertMethod) throws Ex
     assertEquals(timePartitioning, table.getTimePartitioning());
   }
 
+  @Test
+  public void testClusteringStreamingInserts() throws Exception {
```
chamikaramj: Probably also add a test that confirms that we fail if withClustering() is set without withTimePartitioning().

chamikaramj:
FYI: I'm OOO till 6/9 so please add another reviewer if you need to get this merged early.

LGTM from my side other than the two comments I mentioned and integration tests passing.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/DynamicDestinationsHelpers.java
jklukas: Unintentional. Will revert.

jklukas:
Run Java PostCommit

sdks/java/io/google-cloud-platform/src/test/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryIOWriteTest.java
jklukas: Added

```
  @Test(expected = IllegalArgumentException.class)
  public void testClusteringThrowsWithoutPartitioning() throws Exception {
```

jklukas:
> LGTM from my side other than the two comments I mentioned and integration tests passing.

Addressed those comments, integration tests passed, and now I've squashed and rebased on master.

> FYI: I'm OOO till 7/9 so please add another reviewer if you need to get this merged early.

I'm also OOO next week, so if there are any further changes that need to be made, I'll be back on 7/15. We have several weeks before the branch cut date for 2.15, so I don't see any point in rushing. 

And thanks for the thoughtful review, @chamikaramj .

jklukas:
Run Java PostCommit

jklukas:
Run Java PostCommit

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/TableDestinationCoderV3.java
```diff
@@ -0,0 +1,64 @@
...
+import org.apache.beam.sdk.coders.AtomicCoder;
+import org.apache.beam.sdk.coders.Coder;
+import org.apache.beam.sdk.coders.NullableCoder;
+import org.apache.beam.sdk.coders.StringUtf8Coder;
+
+/**
+ * A {@link Coder} for {@link TableDestination} that includes time partitioning and clustering
+ * information. Users must opt in to this version of the coder by setting one of the clustering
+ * options on {@link BigQueryIO.Write}, otherwise {@link TableDestinationCoderV2} will be used and
+ * clustering information will be discarded.
```
reuvenlax: if a user forgets to set a clustering option but creates a TableDestination with clustering information, will we warn them or just silently discard?

jklukas:
There is no warning at this point. For the case of loading to an existing
table, the operation will throw exceptions anyway if clustering is
mismatched.  Perhaps there are compile time cases we could warn about.

I will have to think more about when I'm back in office next week if we
want to block on that.

reuvenlax:
A "good enough" solution might be to always print a warning when clustering
is not enabled. We will want to wordsmith the warning so it's not too
frightening, and it's clear that it only applies if trying to specify
clustering.

chamikaramj:
Thanks. I agree that a compile time (or even run-time) failure is better but a warning should be OK if this is not possible for some reason.

I'll wait till this is addressed for the merge.

jklukas:
Run Java PostCommit

jklukas:
Run Java PreCommit

jklukas:

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/CreateTables.java
```diff
@@ -130,6 +130,16 @@ private TableDestination getTableDestination(ProcessContext context, Destination
...
+      boolean destinationCoderSupportsClustering =
+          !(dynamicDestinations.getDestinationCoder() instanceof TableDestinationCoderV2);
+      checkArgument(
+          tableDestination.getClustering() == null || destinationCoderSupportsClustering,
+          "DynamicDestinations.getTable() may only return destinations with clustering configured"
+              + " if a destination coder is supplied that supports clustering, but %s is configured"
+              + " to use TableDestinationCoderV2. Set withClustering() on BigQueryIO.write() and, "
+              + " if you provided a custom DynamicDestinations instance, override"
+              + " getDestinationCoder() to return TableDestinationCoderV3.",
+          dynamicDestinations);
```
jklukas: We cannot know at compile-time whether a custom table function or DynamicDestinations instance will produce any destination with clustering enabled, so we have to check the produced destinations at runtime.

This check will cause a runtime failure for StreamingInserts if a clustered destination is produced without the relevant configuration to use the newer destination coder.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/WriteTables.java
```diff
@@ -165,6 +166,16 @@ public void processElement(ProcessContext c, BoundedWindow window) throws Except
...
+      boolean destinationCoderSupportsClustering =
+          !(dynamicDestinations.getDestinationCoder() instanceof TableDestinationCoderV2);
+      checkArgument(
+          tableDestination.getClustering() == null || destinationCoderSupportsClustering,
+          "DynamicDestinations.getTable() may only return destinations with clustering configured"
+              + " if a destination coder is supplied that supports clustering, but %s is configured"
+              + " to use TableDestinationCoderV2. Set withClustering() on BigQueryIO.write() and, "
+              + " if you provided a custom DynamicDestinations instance, override"
+              + " getDestinationCoder() to return TableDestinationCoderV3.",
+          dynamicDestinations);
```
jklukas: This check will cause a runtime failure for BatchLoads if a clustered destination is produced without the relevant configuration to use the newer destination coder.

jklukas:
R: @chamikaramj 

I've added in runtime checks for both the streaming insert and batch load cases where the transform will fail with an exception that includes instructions about how to configure clustering if a clustered destination is produced without the right coder in place.

Does this look ready to merge?

chamikaramj:
LGTM. Thanks Jeff.

chamikaramj:
Run Java PostCommit

chamikaramj:
Run Dataflow ValidatesRunner

jklukas:
Run Java PostCommit

jklukas:
Run Java PostCommit

jklukas:
Run Java PostCommit

jklukas:
Run Java PostCommit

jklukas:
Run Dataflow ValidatesRunner

jklukas:
@chamikaramj All the tests here are finally passing again after transient issues.

chamikaramj:
Thanks. Merging.

PR #8951, opened 2019-06-26
[BEAM-7545] Adding RowCount to TextTable

riazela:
This change enables row count estimation for text table. It will first read couple of rows from the files in the directory and then using the length of those rows and the size of the files estimates the size of table.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/io/FileIO.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/io/FileIO.java
@@ -729,6 +729,55 @@ public class FileIO {
       builder.add(DisplayData.item("directoryTreatment", getDirectoryTreatment().toString()));
     }
 
+    /**
+     * @return True if metadata is a directory and directory Treatment is SKIP.
+     * @throws java.lang.IllegalArgumentException if metadata is a directory and directoryTreatment
+     *     is Prohibited.
+     * @throws java.lang.UnsupportedOperationException if metadata is a directory and
+     *     directoryTreatment is not SKIP or PROHIBIT.
+     */
+    static boolean shouldSkipDirectory(
+        MatchResult.Metadata metadata, DirectoryTreatment directoryTreatment) {
+      if (metadata.resourceId().isDirectory()) {
+        switch (directoryTreatment) {
+          case SKIP:
+            return true;
+          case PROHIBIT:
+            throw new IllegalArgumentException(
+                "Trying to read " + metadata.resourceId() + " which is a directory");
+
+          default:
+            throw new UnsupportedOperationException(
+                "Unknown DirectoryTreatment: " + directoryTreatment);
+        }
+      }
+
+      return false;
+    }
+
+    /**
+     * Converts metadata to readableFile. Make sure {@link
+     * #shouldSkipDirectory(org.apache.beam.sdk.io.fs.MatchResult.Metadata,
+     * org.apache.beam.sdk.io.FileIO.ReadMatches.DirectoryTreatment)} returns false before using.
+     */
+    static ReadableFile matchToReadableFile(
+        MatchResult.Metadata metadata, Compression compression) {
+
+      compression =
+          (compression == Compression.AUTO)
+              ? Compression.detect(metadata.resourceId().getFilename())
+              : compression;
+      return new ReadableFile(
+          MatchResult.Metadata.builder()
+              .setResourceId(metadata.resourceId())
+              .setSizeBytes(metadata.sizeBytes())
+              .setLastModifiedMillis(metadata.lastModifiedMillis())
+              .setIsReadSeekEfficient(
+                  metadata.isReadSeekEfficient() && compression == Compression.UNCOMPRESSED)
+              .build(),
+          compression);
+    }
+
     private static class ToReadableFileFn extends DoFn<MatchResult.Metadata, ReadableFile> {
       private final ReadMatches spec;
 
@@ -738,36 +787,11 @@ public class FileIO {
 
       @ProcessElement
       public void process(ProcessContext c) {
-        MatchResult.Metadata metadata = c.element();
-        if (metadata.resourceId().isDirectory()) {
-          switch (spec.getDirectoryTreatment()) {
-            case SKIP:
-              return;
-
-            case PROHIBIT:
-              throw new IllegalArgumentException(
-                  "Trying to read " + metadata.resourceId() + " which is a directory");
-
-            default:
-              throw new UnsupportedOperationException(
-                  "Unknown DirectoryTreatment: " + spec.getDirectoryTreatment());
-          }
+        if (shouldSkipDirectory(c.element(), spec.getDirectoryTreatment())) {
+          return;
         }
-
-        Compression compression =
-            (spec.getCompression() == Compression.AUTO)
-                ? Compression.detect(metadata.resourceId().getFilename())
-                : spec.getCompression();
-        c.output(
-            new ReadableFile(
-                MatchResult.Metadata.builder()
-                    .setResourceId(metadata.resourceId())
-                    .setSizeBytes(metadata.sizeBytes())
-                    .setLastModifiedMillis(metadata.lastModifiedMillis())
-                    .setIsReadSeekEfficient(
-                        metadata.isReadSeekEfficient() && compression == Compression.UNCOMPRESSED)
-                    .build(),
-                compression));
+        ReadableFile r = matchToReadableFile(c.element(), spec.getCompression());
+        c.output(r);
       }
     }
   }
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/io/TextRowCountEstimator.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/io/TextRowCountEstimator.java
@@ -1,0 +1,219 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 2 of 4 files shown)


riazela:
R: @akedin

riazela:

riazela:

riazela:

riazela:

riazela:

riazela:

sdks/java/core/src/main/java/org/apache/beam/sdk/io/FileIO.java
```diff
@@ -738,36 +772,12 @@ private ToReadableFileFn(ReadMatches spec) {
...
-              throw new IllegalArgumentException(
-                  "Trying to read " + metadata.resourceId() + " which is a directory");
-
-            default:
-              throw new UnsupportedOperationException(
-                  "Unknown DirectoryTreatment: " + spec.getDirectoryTreatment());
-          }
+        ReadableFile r =
+            matchToReadableFile(c.element(), spec.getDirectoryTreatment(), spec.getCompression());
+        if (r == null) {
```
akedin: Without reading everything it's unclear why match can be a `null`, e.g. there could be an error of some sort (access permission), or no matches or something else. I suggest creating another method along the lines of:

```java
boolean shouldSkipDirectory(metadata) {
  if (... isDirectory()) {
     ..
     case SKIP:
       return true;
     case PROHIBIT:
       throw ...
  }
}
```

This way it's clear that the point is specifically to handle/skip directories:

```java

void process(c) {
  if (shouldSkipDirectory(c.element())) {
    return;
  }
  c.output(toReadableFile(c.element()));
}

File: sdks/java/core/src/main/java/org/apache/beam/sdk/io/TextRowCountEstimator.java [hunk 2]
```diff
@@ -0,0 +1,154 @@
...
+        .sampleAllFiles()
+        .setNumSampledBytes(DEFAULT_NUM_BYTES_LINES)
+        .setCompression(DEFAULT_COMPRESSION)
+        .setDirectoryTreatment(DEFAULT_DIRECTORY_TREATMENT)
+        .setEmptyMatchTreatment(DEFAULT_EMPTY_MATCH_TREATMENT);
+  }
+
+  /** Builder for {@link org.apache.beam.sdk.io.TextRowCountEstimator}. */
+  @AutoValue.Builder
+  public abstract static class Builder {
```
akedin: I think `Builder` usually lives at the end of the file

sdks/java/core/src/main/java/org/apache/beam/sdk/io/TextRowCountEstimator.java
```diff
@@ -0,0 +1,154 @@
...
+      long fileSize = metadata.sizeBytes();
+      numberOfReadFiles += fileSize == 0 ? 0 : 1;
+      totalFileSizes += fileSize;
+    }
+
+    if (totalFileSizes == 0) {
+      return 0L;
+    }
+
+    if (numberOfReadLines == 0) {
```
akedin: Is it safe to only check `numberOfReadLines`? If it's zero, then why do we care about number of files?

sdks/java/core/src/main/java/org/apache/beam/sdk/io/TextRowCountEstimator.java
```diff
@@ -0,0 +1,154 @@
...
+    MatchResult match = FileSystems.match(getFilePattern(), getEmptyMatchTreatment());
+
+    for (MatchResult.Metadata metadata : match.metadata()) {
+
+      if ((!isSamplingAllFiles()) && numberOfReadFiles > getNumSampledFiles()) {
+        break;
+      }
+
+      FileIO.ReadableFile file =
+          FileIO.ReadMatches.matchToReadableFile(
```
akedin: I am not sure if `directoryTreatment` matters here, we probably can just always skip them. It might matter for actual processing but for row count estimation we have to look at the files no matter what kind directory handling we want at run time.

sdks/java/core/src/main/java/org/apache/beam/sdk/io/TextRowCountEstimator.java
```diff
@@ -0,0 +1,154 @@
...
+              getDelimiters());
+      FileBasedSource<String> source =
+          CompressedSource.from(textSource).withCompression(file.getCompression());
+      try (BoundedSource.BoundedReader<String> reader =
+          source
+              .createForSubrangeOfFile(file.getMetadata(), range.getFrom(), range.getTo())
+              .createReader(pipelineOptions)) {
+        for (boolean more = reader.start(); more; more = reader.advance()) {
+          int length = reader.getCurrent().length();
+          linesSize += reader.getCurrent().length() + 1; // for \n at the end of line
```
akedin: Is it ok to ignore `\n\r`? What does `TextIO` think about line ends?

sdks/java/core/src/main/java/org/apache/beam/sdk/io/TextRowCountEstimator.java
```diff
@@ -0,0 +1,154 @@
...
+  public abstract Compression getCompression();
+
+  public abstract EmptyMatchTreatment getEmptyMatchTreatment();
+
+  public abstract FileIO.ReadMatches.DirectoryTreatment getDirectoryTreatment();
+
+  public static TextRowCountEstimator.Builder builder() {
+    return (new AutoValue_TextRowCountEstimator.Builder())
+        .sampleAllFiles()
+        .setNumSampledBytes(DEFAULT_NUM_BYTES_LINES)
```
akedin: Should this be number of characters, not bytes?

sdks/java/core/src/main/java/org/apache/beam/sdk/io/TextRowCountEstimator.java
```diff
@@ -0,0 +1,154 @@
...
+
+    public abstract Builder setCompression(Compression compression);
+
+    public abstract Builder setDelimiters(byte[] delimiters);
+
+    public abstract Builder setFilePattern(String filePattern);
+
+    public abstract Builder setEmptyMatchTreatment(EmptyMatchTreatment emptyMatchTreatment);
+
+    public Builder sampleAllFiles() {
```
akedin: I think it's simpler to make this a separate abstract boolean method, so that you don't have to actually implement this and the getter

sdks/java/core/src/main/java/org/apache/beam/sdk/io/TextRowCountEstimator.java
akedin: Should we check if  `linesSize > getNumSampledBytes()` ?

akedin: Is the length in bytes or chars?

sdks/java/core/src/main/java/org/apache/beam/sdk/io/TextRowCountEstimator.java
```diff
@@ -0,0 +1,154 @@
...
+
+    if (totalFileSizes == 0) {
+      return 0L;
+    }
+
+    if (numberOfReadLines == 0) {
+      return 0L;
+    }
+
+    return totalFileSizes * numberOfReadLines / linesSize;
```
akedin: Can you add a comment here explaining the formula?

I am also thinking whether we should add a configuration for:

 * skipping empty files from calculation;
 * choosing a different statistic, not just mean (e.g. median or any percentile);

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/text/TextTable.java
```diff
@@ -59,6 +68,30 @@ public String getFilePattern() {
...
+    return rowCountStatistics;
+  }
+
+  private static BeamRowCountStatistics getTextRowEstimate(
+      PipelineOptions options, String filePattern) {
+    TextRowCountEstimator textRowCountEstimator =
+        TextRowCountEstimator.builder().setFilePattern(filePattern).build();
+    try {
+      Long rows = textRowCountEstimator.estimateRowCount(options);
+      if (rows != null) {
```
akedin: is it ever `null`?

sdks/java/core/src/main/java/org/apache/beam/sdk/io/TextRowCountEstimator.java
riazela: I think we can remove the previous if clause. Previously I was returning null in case of numberOfReadLines==0, and I forgot to fix this part after changing that.
I was thinking about returning null when we are not sure about the estimate (we haven't read non empty lines) and return 0 when we are sure that the number of lines is 0. I will change this part and push another commit.

sdks/java/core/src/main/java/org/apache/beam/sdk/io/TextRowCountEstimator.java
riazela: I moved it

sdks/java/core/src/main/java/org/apache/beam/sdk/io/TextRowCountEstimator.java
riazela: Yes I think I have messed up the bytes and characters. I should probably change either lineSizes to bytes instead of characters or change this to be number of characters.

sdks/java/core/src/main/java/org/apache/beam/sdk/io/TextRowCountEstimator.java
riazela: Yes. I have set it as a default value to SKIP. So it will just skip the directories. Do you think it is better to remove the setter for that?

riazela:

riazela:
Run Java PostCommit

PR #8954, opened 2019-06-26
Spotless config update to include java files under src directory

akedin:
Restrict spotless java to `src`

Spotless seems to fail when doing `./gradlew :sdks:java:core:spotlessJavaCheck`, it attempts to check the generated sources although they are excluded in the configuration. Changing the configuration to restrict to java files only in `src` instead.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy
+++ b/buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy
@@ -816,13 +816,8 @@ class BeamModulePlugin implements Plugin<Project> {
           project.sourceSets.each { sourceSet ->
             targetFiles += sourceSet.allJava
           }
-          target targetFiles.matching {
-            include '**/*.java'
-            exclude '**/archetype-resources/src/**'
-            exclude '**/build/generated/**'
-            exclude '**/build/generated-src/**'
-            exclude '**/build/generated-*-avro-*/**'
-          }
+
+          target targetFiles.matching { include 'src/*/java/**/*.java' }
         }
       }
 


akedin:
retest this please

akedin:

akedin:
Run Python PreCommit

akedin:

akedin:
R: @adude3141 @mxm @lukecwik

buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy
```diff
@@ -817,11 +818,8 @@ class BeamModulePlugin implements Plugin<Project> {
             targetFiles += sourceSet.allJava
           }
           target targetFiles.matching {
-            include '**/*.java'
+            include '**/src/**/*.java'
```
lukecwik: What prevents us from doing ```src/**/*.java```?

I believe the path is relative to the build.gradle file applying this plugin.

akedin: Nothing prevents us from doing that to my knowledge, updated accordingly. Just wanted to keep it less restrictive to be on the safe side. 'src/**..' seems to work for the cases I tested so it makes more sense.

buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy
```diff
@@ -38,6 +37,8 @@ import org.gradle.api.tasks.javadoc.Javadoc
 import org.gradle.api.tasks.testing.Test
 import org.gradle.testing.jacoco.tasks.JacocoReport
 
+import java.util.concurrent.atomic.AtomicInteger
```
lukecwik: nit: it looks like your IDE import order is incorrect for groovy files.

akedin: probably `src/*/java/**` then, there can potentially be more source sets (e.g. I wonder if it would make sense to model integration tests as source sets and move them out of `test`, or model different flink versions as source sets instead of projects).

akedin: fixed both, seems to work

lukecwik: SGTM on the `src/*/java/**`.

Moving ITs out to a different source set would be one way to do it but this would move them out of the tests artifact that is generated and uploaded to maven and we would have to configure other projects that depend on these tests to possibly depend on both now.

akedin:
retest this please

mxm:
Thanks, this is much simpler.

adude3141:
@mxm Does it still catch spotless errors in flinks common folder? I think, this might be broken again...

mxm:
@adude3141 I was concerned about this as well, but this keeps the source set support. When we added source set support we included all Java files which forced us to exclude the generated ones. This regex includes all Java files under `src/*` which gets rid of the generated ones and conforms with the multiple Flink source sets which use this standard layout. So we're good :)

adude3141:
Ah, thanks for verifying. So it must be some issue on my machine. As I am currently unable to fail spotlessCheck by adding random format errors to runners/flink/src, whereas it fails if doing the same to 1.5 srcs... And I tend to believe this was previously working...

mxm:
Thanks for taking a closer look @adude3141. I only checked out the diff here and figured the changes looked good. Indeed, the spotless configuration is broken in the main Flink source directory under `runners/flink/src` due to this PR. 

It is strange that this only stopped working for one of the two source sets, especially because they have the same source layout. The only solution I see for now is to revert and exclude the generated files for which the PR was opened.

PR #8962, opened 2019-06-28
[BEAM-6675] Generate JDBC statement and preparedStatementSetter automatically when schema is available

JawadHyder:
Added functionality to automatically generate JDBC insert statement and `preparedStatementSetter` when there is schema attached to the `PCollection`. This functionality is triggered when the user has not provided `withStatement` and `withPreparedStatementSetter` during `JdbcIO.write()` operation.

------------------------

--- | --- | --- | --- | ---


JawadHyder:
retest this please

JawadHyder:
R: @reuvenlax 
R: @jbonofre

JawadHyder:
Run Java PreCommit

JawadHyder:

salmanVD:
Run Java PreCommit

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcUtil.java
```diff
@@ -0,0 +1,179 @@
...
+                  return "?";
+                })
+            .collect(Collectors.joining(", "));
+
+    return String.format("INSERT INTO %s(%s) VALUES(%s)", tableName, fieldNames, valuePlaceholder);
+  }
+
+  /** PreparedStatementSetCaller for Schema Field types. * */
+  private static Map<Schema.TypeName, JdbcIO.PreparedStatementSetCaller> typeNamePsSetCallerMap =
+      new EnumMap<Schema.TypeName, JdbcIO.PreparedStatementSetCaller>(
```
amaliujia: Seems missing handling some types in Schema?

amaliujia: O NVM. There is a separate section to handle other types (array, logical types, etc.)

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcUtil.java
```diff
@@ -0,0 +1,179 @@
...
+        };
+      case LOGICAL_TYPE:
+        {
+          String logicalTypeName = fieldType.getLogicalType().getIdentifier();
+          JDBCType jdbcType = JDBCType.valueOf(logicalTypeName);
+          switch (jdbcType) {
+            case DATE:
+              return (element, ps, i, fieldWithIndex) -> {
+                ps.setDate(
+                    i + 1, new Date(element.getDateTime(fieldWithIndex.getIndex()).getMillis()));
```
amaliujia: Question: is it true to set `epoch millis` for Date and Time type?

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcUtil.java
```diff
@@ -0,0 +1,179 @@
...
+            case TIME:
+              return (element, ps, i, fieldWithIndex) -> {
+                ps.setTime(
+                    i + 1, new Time(element.getDateTime(fieldWithIndex.getIndex()).getMillis()));
+              };
+            case TIMESTAMP_WITH_TIMEZONE:
+              return (element, ps, i, fieldWithIndex) -> {
+                ps.setTimestamp(
+                    i + 1,
+                    new Timestamp(element.getDateTime(fieldWithIndex.getIndex()).getMillis()));
```
amaliujia: As this is a `timezoned` type, is time zone handled here?

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcUtil.java
amaliujia: Actually BYTES is not handled?

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcUtil.java
```diff
@@ -0,0 +1,179 @@
...
+    return (element, ps, i, fieldWithIndex) -> {
+      validateLogicalTypeLength(
+          fieldWithIndex.getField(), element.getString(fieldWithIndex.getIndex()).length());
+      ps.setString(i + 1, element.getString(fieldWithIndex.getIndex()));
+    };
+  }
+
+  private static void validateLogicalTypeLength(Schema.Field field, Integer length) {
+    if (field.getType().getTypeName().isLogicalType()
+        && length >= Integer.parseInt(field.getType().getLogicalType().getArgument())) {
```
amaliujia: I think logical type argument is not always set. If the argument is not set as an Integer, maybe we can further check base type?

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcUtil.java
salmanVD: The constructor for Date & Time takes a long value so I don't see any issue setting epoch for date & time as it gives you only Date & Time in return.  Do you have a second thought on this?

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcUtil.java
sehrish-vd: Bytes is handled as well https://github.com/JawadHyder/beam/blob/9bb7280a65a4ea506f36d8597aaee3500633ef5a/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcUtil.java#L96

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcUtil.java
sehrish-vd: @amaliujia the case is if the string length of input field exceeds from the table max limit then exception should be raised, base type will still be same even if the length differs.

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcUtil.java
sehrish-vd: updated - UTC timezone is set as default when saving timestamp with timezone.

sehrish-vd:
Run Java PreCommit

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcUtil.java
salmanVD: The constructor for Date & Time takes a long value so I don't see any issue setting epoch for date & time.  Do you have any other thoughts on this?

amaliujia: I guess here we should obey `java.sql.Date` and `java.sql.Time`'s specs. 

https://docs.oracle.com/javase/7/docs/api/java/sql/Date.html
https://docs.oracle.com/javase/8/docs/api/index.html?java/sql/Time.html

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcUtil.java
amaliujia: If so,  could we check input field to see if it is bytes or string?

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcUtil.java
amaliujia: I am thinking `element.getDateTime` will return a Joda Datetime, which has timezone information already?

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcUtil.java
sehrish-vd: Input fields are checked in the switch case then called createStringCaller() for string for bytes it's createBytesCaller()
https://github.com/JawadHyder/beam/blob/9bb7280a65a4ea506f36d8597aaee3500633ef5a/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcUtil.java#L155
If only type check is enough then wouldn't it truncate the string/bytes if it exceeds the max limit from the table?

amaliujia: O I see. Then that's good enough. I wasn't sure how this function is called. E.g. If this function is called to verify other types like Date, Time, etc.

sehrish-vd: No these methods are called for string and bytes due to variable length. So one thing, if the argument is not set should we throw exception (the field details like argument is fetched from table schema so I don't think it can be null) or let it proceed with the insertion?

amaliujia: Argument is an optional parameter. So I think if it is not set, we can proceed to insertion.

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcUtil.java
sehrish-vd: As per my understanding we should use these constructors instead of epoch
Time - Time(int hour, int minute, int second)
Date - Date(int year, int month, int day)
but both are deprecated as per docs, are you referring to the same thing @amaliujia ?

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcUtil.java
sehrish-vd: timestamp is the moment in time I don't think timezone is needed here because when user is going to read data from DB (timestamp field) then it needs to be converted at that point.
Please check this link https://stackoverflow.com/questions/15994450/joda-datetime-to-timestamp-conversion/15994657#15994657
can you share something related to your point? may be I am missing something here

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcUtil.java
sehrish-vd: So there are cases:
1. If argument is not set - bypass the check and proceed with insertion (can we use a warning message here as data might truncate due to length mismatch?)
2. If argument is set and it is not integer - do same as above
3. If argument is set and it is integer - throw exception if length mismatch

Let's finalize these cases or any other case you think might possible or we can raise separate jira for this and remove checks from here?

sehrish-vd:
@amaliujia can you please review the comments again?

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcUtil.java
amaliujia: Something similar, for example:
```
The date components should be set to the "zero epoch" value of January 1, 1970 and should not be accessed.
```

I am not sure if current construction for `TIME` is doing this.

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcUtil.java
amaliujia: There are two TIMESTAMP: `TIMESTAMP without TIMEZONE`(sometimes I saw it's called DATETIME), `TIMESTAMP with TIMEZONE`.

In this implementation, it's processing a `TIMESTAMP with TIMEZONE` type, which definitely has a timezone information that matters. I think here are two concepts matter (which is also covered by your link): an absolute point in time and an offset. Offset (specified by timezone) should be lost for `TIMESTAMP with TIMEZONE`


I find this SO question is useful: https://stackoverflow.com/questions/14070572/is-java-sql-timestamp-timezone-specific

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcUtil.java
amaliujia: Sounds good to all of these. It would be backward-compatible to allow bypassing checking in 1 and 2 such that existing types won't break if setting an argument.

amaliujia:
Thanks @sehrish-vd  and @JawadHyder for working on this PR! Responded to comments!

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcUtil.java
sehrish-vd: done

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcUtil.java
sehrish-vd: used setTimestamp with calendar instance for timezone
done

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcUtil.java
sehrish-vd: updated for both date and time objects
done

sehrish-vd:
Run Java PreCommit

amaliujia:
LGTM

lgajowy:
Run Java JdbcIO Performance Test

amaliujia:
Thanks @lgajowy!

PR #8964, opened 2019-06-28
[BEAM-6740] Add PTransformTranslator for Combine.Globally

iemejia:
It seems that we missed the payloads for Combine.Globally that can have specific translations as is the case of the spark runner.

R: @lukecwik

### DIFF
--- a/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/CombineTranslation.java
+++ b/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/CombineTranslation.java
@@ -37,7 +37,6 @@ import org.apache.beam.sdk.runners.AppliedPTransform;
 import org.apache.beam.sdk.transforms.Combine;
 import org.apache.beam.sdk.transforms.CombineFnBase.GlobalCombineFn;
 import org.apache.beam.sdk.transforms.PTransform;
-import org.apache.beam.sdk.util.AppliedCombineFn;
 import org.apache.beam.sdk.util.SerializableUtils;
 import org.apache.beam.sdk.values.KV;
 import org.apache.beam.sdk.values.PCollection;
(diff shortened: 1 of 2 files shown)


iemejia:
Run Spark ValidatesRunner

iemejia:
Run Portable_Python PreCommit

iemejia:
Run Spark ValidatesRunner

iemejia:
Run Java PreCommit

iemejia:
Run Java PreCommit

iemejia:
PR is finally green. Can you PTAL @lukecwik

iemejia:
Run Spark ValidatesRunner

iemejia:
Run Spark ValidatesRunner

iemejia:
Run Spark ValidatesRunner

iemejia:
Run Spark ValidatesRunner

iemejia:
Sorry for the possible spam on all my changes to this PR. I did some extra refinements to remove duplicated code. It is now ready for review. PTAL @lukecwik

lukecwik:
This change looks good on the translation side but `COMBINE_GLOBALLY` doesn't have the corresponding definition on how to execute it since `COMBINE_PER_KEY` has the following [combine components](https://github.com/apache/beam/blob/e65c176a9f34e45d408281e1101a2ae54cef0f6c/model/pipeline/src/main/proto/beam_runner_api.proto#L253) defined: `COMBINE_PER_KEY_PRECOMBINE`, `COMBINE_PER_KEY_MERGE_ACCUMULATORS`, `COMBINE_PER_KEY_EXTRACT_OUTPUTS` and `COMBINE_GROUPED_VALUES` which all work in a keyed context.

How do you want the expansion of combine globally to look during execution and what additional well known URNs should we be defining?

youngoli:
Yeah, the existing doc doesn't touch on `Combine.Globally` at all, but it's essentially just `Combine.PerKey` with a global key. Once you've grouped all elements under that global key I think you could just use `COMBINE_GROUPED_VALUES`. (From what I remember combiner lifting isn't really useful with a global combine but theoretically you could use the lifted version of the expansion too.)

So I'd say a `Combine.Globally` can just expand to `GroupByGlobalKey` -> `Combine.GroupedValues`, where `GroupByGlobalKey` is any transform that groups all elements in an iterator under some arbitrary key. I don't know if a transform like that exists in the Java SDK already, but it shouldn't be hard to implement.

iemejia:
To be honest I was not even aware of these Combine components. This PR and #9026 come from a regression that was accidentally introduced into the Spark runner when we moved it into URN based translation. Spark runner specializes both `Combine.Globally` and `Combine.GroupedValues` but both translations were never called because there were not `PTransformTranslator`s  associated for both of them. So the goal of these PRs is first to fix that. However I am curious on any extra feedback for things we can improve for the respective cases.

PR #8977, opened 2019-07-01
[BEAM-7668] Add GetPipeline method to gRPC JobService

chadrik:
This adds a `GetPipeline` method to the gRPC `JobService` and implements it for `InMemoryJobService`.   The method is quite similar to the existing `GetState` method in design/behavior. 

As described in the Jira issue, I see this feature as a useful first step towards making the Job API capable of serving as a backend to a web app for submitting, viewing, and monitoring Beam jobs.  The next step after this PR would be to add a `GetJobs` method that sends a summary of all of the active jobs and their ids.  

Remaining issues:

- go: I have not implemented anything for the go sdk yet.  I'm still trying to figure out which version of `protoc` I need to use with `go generate`:  3.6 and 3.7 produced significant diffs, with a lot of style changes, so I assume it's something older.  Any help here would be greatly appreciated.
- tests:  I've looked at the related tests for Java and Python and they don't seem to be testing the rpc service endpoints. The python tests in `portable_runner_test` seem to be largely starting the service and shutting it down in different ways, without actually running any test methods.  Basically what I'm saying is I don't see a clear framework for testing RPC methods in any language, but feel free to point me in the right direction. 

--- | --- | --- | --- | ---

### DIFF
--- a/runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/InMemoryJobService.java
+++ b/runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/InMemoryJobService.java
@@ -26,6 +26,8 @@ import org.apache.beam.model.jobmanagement.v1.JobApi.CancelJobRequest;
 import org.apache.beam.model.jobmanagement.v1.JobApi.CancelJobResponse;
 import org.apache.beam.model.jobmanagement.v1.JobApi.DescribePipelineOptionsRequest;
 import org.apache.beam.model.jobmanagement.v1.JobApi.DescribePipelineOptionsResponse;
+import org.apache.beam.model.jobmanagement.v1.JobApi.GetJobPipelineRequest;
+import org.apache.beam.model.jobmanagement.v1.JobApi.GetJobPipelineResponse;
 import org.apache.beam.model.jobmanagement.v1.JobApi.GetJobStateRequest;
 import org.apache.beam.model.jobmanagement.v1.JobApi.GetJobStateResponse;
 import org.apache.beam.model.jobmanagement.v1.JobApi.JobMessage;
@@ -38,6 +40,7 @@ import org.apache.beam.model.jobmanagement.v1.JobApi.RunJobRequest;
 import org.apache.beam.model.jobmanagement.v1.JobApi.RunJobResponse;
 import org.apache.beam.model.jobmanagement.v1.JobServiceGrpc;
 import org.apache.beam.model.pipeline.v1.Endpoints;
+import org.apache.beam.model.pipeline.v1.RunnerApi;
 import org.apache.beam.runners.core.construction.graph.PipelineValidator;
 import org.apache.beam.runners.fnexecution.FnService;
 import org.apache.beam.sdk.fn.stream.SynchronizedStreamObserver;
@@ -224,6 +227,26 @@ public class InMemoryJobService extends JobServiceGrpc.JobServiceImplBase implem
       JobInvocation invocation = getInvocation(invocationId);
       JobState.Enum state = invocation.getState();
       GetJobStateResponse response = GetJobStateResponse.newBuilder().setState(state).build();
+      responseObserver.onNext(response);
+      responseObserver.onCompleted();
+    } catch (Exception e) {
+      String errMessage =
+          String.format("Encountered Unexpected Exception for Invocation %s", invocationId);
+      LOG.error(errMessage, e);
+      responseObserver.onError(Status.INTERNAL.withCause(e).asException());
+    }
+  }
+
+  @Override
+  public void getPipeline(
+      GetJobPipelineRequest request, StreamObserver<GetJobPipelineResponse> responseObserver) {
+    LOG.trace("{} {}", GetJobPipelineRequest.class.getSimpleName(), request);
+    String invocationId = request.getJobId();
+    try {
+      JobInvocation invocation = getInvocation(invocationId);
+      RunnerApi.Pipeline pipeline = invocation.getPipeline();
+      GetJobPipelineResponse response =
+          GetJobPipelineResponse.newBuilder().setPipeline(pipeline).build();
       responseObserver.onNext(response);
       responseObserver.onCompleted();
     } catch (Exception e) {
--- a/sdks/python/apache_beam/runners/portability/local_job_service.py
+++ b/sdks/python/apache_beam/runners/portability/local_job_service.py
@@ -131,6 +131,10 @@ class LocalJobServicer(beam_job_api_pb2_grpc.JobServiceServicer):
     return beam_job_api_pb2.GetJobStateResponse(
         state=self._jobs[request.job_id].state)
 
+  def GetPipeline(self, request, context=None):
+    return beam_job_api_pb2.GetJobPipelineResponse(
+        state=self._jobs[request.job_id]._pipeline_proto)
+
   def Cancel(self, request, context=None):
     self._jobs[request.job_id].cancel()
     return beam_job_api_pb2.CancelJobRequest(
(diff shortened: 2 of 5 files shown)


chadrik:
R: @herohde 
R: @lukecwik

chadrik:
R: @lukecwik
R: @angoenka

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/InMemoryJobService.java
```diff
@@ -234,6 +237,26 @@ public void getState(
...
+    try {
+      JobInvocation invocation = getInvocation(invocationId);
+      RunnerApi.Pipeline pipeline = invocation.getPipeline();
+      GetJobPipelineResponse response =
+          GetJobPipelineResponse.newBuilder().setPipeline(pipeline).build();
+      responseObserver.onNext(response);
+      responseObserver.onCompleted();
+    } catch (Exception e) {
+      String errMessage =
+          String.format("Encountered Unexpected Exception for Invocation %s", invocationId);
```
lukecwik:
```suggestion
          String.format("Encountered unexpected exception for invocation %s", invocationId);
```

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/InMemoryJobService.java
```diff
@@ -234,6 +237,26 @@ public void getState(
...
+      LOG.error(errMessage, e);
```
lukecwik: nit: SLF4J allows you to do
```
LOG.error(Encountered unexpected exception for invocation {}", invocationId, e); 
```
instead of using `String.format`

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/InMemoryJobService.java
```diff
@@ -234,6 +237,26 @@ public void getState(
...
   }
 
+  @Override
+  public void getPipeline(
+      GetJobPipelineRequest request, StreamObserver<GetJobPipelineResponse> responseObserver) {
+    LOG.trace("{} {}", GetJobPipelineRequest.class.getSimpleName(), request);
+    String invocationId = request.getJobId();
+    try {
+      JobInvocation invocation = getInvocation(invocationId);
+      RunnerApi.Pipeline pipeline = invocation.getPipeline();
```
lukecwik: We aren't handling the NOT_FOUND case as described in the API contract in the proto.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/InMemoryJobService.java
chadrik: Note that this capitalization is present in 6 other messages in this file.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/InMemoryJobService.java
chadrik: Again, this pattern is repeated in 5 other places in this file.  I was just trying to stay consistent.  

If you don't like the style used in the file, I would recommend addressing them all at once in another PR.

lukecwik: Sounds good to me.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/InMemoryJobService.java
chadrik: Again, I copied the API contract from `getState`, and handled the error case in exactly the same way, so if it's not correct for `getPipeline`, that's because it's not correct for `getState` or the other 4 methods that use this same pattern. 

In case of error it's responding with the exception:

```java
catch (Exception e) {
  ...
  responseObserver.onError(Status.INTERNAL.withCause(e).asException());
}
```

I'm new to Java, so I just assumed this was responding with `NOT_FOUND` in the case that `getInvocation(invocationId)` failed.

lukecwik: We are handling the exception but are doing it incorrectly since getInvocation() throws a Status.NOT_FOUND exception but since we are catching it and then rethrowing it as an `INTERNAL` exception we are losing the `NOT_FOUND`.

The logic needs to be updated to be:
```
try {
  ... do stuff ...
} catch (StatusException e) {
  responseObserver.onError(e);
} catch (Exception e) {
  ... original exception handling that converts the status to INTERNAL ...
}
```

chadrik: Ok, I'll make a separate issue in Jira for this, since it's technically a breaking change, and I'll follow up with another PR with your suggestion, and the style fixes (if you don't mind me lumping them together).

lukecwik: I would consider it a bugfix since contractually we said we were going to do X and did Y and not a breaking change.

I'm fine with you lumping it in or not.

chadrik:
Added tests.

chadrik:
I _think_ this PR is ready to go.   I'm not sure if the Python PreCommit test failure is real or not.  I clicked on details but Jenkins says there are no failures.

lukecwik:
Run Python PreCommit

sdks/python/apache_beam/runners/portability/local_job_service.py
```diff
@@ -131,6 +131,10 @@ def GetState(self, request, context=None):
     return beam_job_api_pb2.GetJobStateResponse(
         state=self._jobs[request.job_id].state)
 
+  def GetPipeline(self, request, context=None):
+    return beam_job_api_pb2.GetJobPipelineResponse(
+        state=self._jobs[request.job_id]._pipeline_proto)
```
lukecwik:
```suggestion
        pipeline=self._jobs[request.job_id]._pipeline_proto)
```

lukecwik:
I merged this with a fixup addressing my last comment: https://github.com/apache/beam/commit/05dba6a5b9fb7e1753262f8eed5854c8e19cd794

chadrik:
thanks!

chadrik:
I created an issue for the proto contract: https://issues.apache.org/jira/browse/BEAM-7720

chadrik:
Note, I just realized that I did not update the go stubs...   Is there a reason that the generated code is committed to the repo for go, but not for python and java?

lukecwik:
It may be because of how Go modules have dependencies on other Go modules and since they are statically compiled it may require those Go modules to perform the proto generation on our behalf to perform the compilation but I'm not exactly certain. @lostluck any details you can provide here?

lostluck:
Nit: beam isn't yet using Go Modules. Go calls the unit of dependency packages.

Conventionally, the go tooling doesn't do generation at compile time. The main reason is there are ample tools to do that externally to the go tooling (eg. Gradle, bazel, make, etc) that any end user can set that up should they desire it.
Further, generated code is for the package authors, not the end user, and should be commitable.

The impedance mismatch is that beam is a mid-level library. So the proto code needs to be generated with `go generate`, and committed, because of the above conventions. Eg. Beam devs may use Gradle, but end go programmers will probably not, not can we mandate it.

We haven't been rigorous about what version of protoc and the go proto packages to use, hence the piles of differences. As long as the external proto API doesn't change in a backwards incompatible way and the integration tests against flink and spark work, I'm not bothered about what the generated code looks like until we can ensure people can use consistent versions to generate the code, with Go Modules.

PR #8979, opened 2019-07-01
[BEAM-3645] add multiplexing for python FnApiRunner

Hannah-Jiang:
This PR is the second part of BEAM-3645. This PR adds round robin and multiplext part.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/runners/direct/direct_runner_test.py
+++ b/sdks/python/apache_beam/runners/direct/direct_runner_test.py
@@ -45,13 +45,12 @@ class DirectPipelineResultTest(unittest.TestCase):
   def test_waiting_on_result_stops_executor_threads(self):
     pre_test_threads = set(t.ident for t in threading.enumerate())
 
-    for runner in ['DirectRunner', 'BundleBasedDirectRunner',
+    for runner in ['DirectRunner', #'BundleBasedDirectRunner',
                    'SwitchingDirectRunner']:
       pipeline = test_pipeline.TestPipeline(runner=runner)
       _ = (pipeline | beam.Create([{'foo': 'bar'}]))
       result = pipeline.run()
       result.wait_until_finish()
-
       post_test_threads = set(t.ident for t in threading.enumerate())
       new_threads = post_test_threads - pre_test_threads
       if len(new_threads) > 0:
--- a/sdks/python/apache_beam/runners/direct/direct_runner_test.py
+++ b/sdks/python/apache_beam/runners/direct/direct_runner_test.py
@@ -17,7 +17,9 @@
 
 from __future__ import absolute_import
 
+import logging
 import threading
+import time
 import unittest
 
 import hamcrest as hc
@@ -52,6 +54,15 @@ class DirectPipelineResultTest(unittest.TestCase):
 
       post_test_threads = set(t.ident for t in threading.enumerate())
       new_threads = post_test_threads - pre_test_threads
+      if len(new_threads) > 0:
+        for i in range(1, 6):
+          logging.info('%s: Wait %s second(s) for thread(s) to close.' %
+                       (runner, i))
+          time.sleep(i)
+          post_test_threads = set(t.ident for t in threading.enumerate())
+          new_threads = post_test_threads - pre_test_threads
+          if len(new_threads) == 0:
+            break
       self.assertEqual(len(new_threads), 0)
 
   def test_direct_runner_metrics(self):
@@ -122,4 +133,5 @@ class DirectPipelineResultTest(unittest.TestCase):
 
 
 if __name__ == '__main__':
+  logging.getLogger().setLevel(logging.INFO)
   unittest.main()
--- a/sdks/python/apache_beam/runners/direct/direct_runner_test.py
+++ b/sdks/python/apache_beam/runners/direct/direct_runner_test.py
@@ -17,9 +17,7 @@
 
 from __future__ import absolute_import
 
-import logging
 import threading
-import time
 import unittest
 
 import hamcrest as hc
@@ -45,23 +43,15 @@ class DirectPipelineResultTest(unittest.TestCase):
   def test_waiting_on_result_stops_executor_threads(self):
     pre_test_threads = set(t.ident for t in threading.enumerate())
 
-    for runner in ['DirectRunner', #'BundleBasedDirectRunner',
+    for runner in ['DirectRunner', 'BundleBasedDirectRunner',
                    'SwitchingDirectRunner']:
       pipeline = test_pipeline.TestPipeline(runner=runner)
       _ = (pipeline | beam.Create([{'foo': 'bar'}]))
       result = pipeline.run()
       result.wait_until_finish()
+
       post_test_threads = set(t.ident for t in threading.enumerate())
       new_threads = post_test_threads - pre_test_threads
-      if len(new_threads) > 0:
-        for i in range(1, 6):
-          logging.info('%s: Wait %s second(s) for thread(s) to close.' %
-                       (runner, i))
-          time.sleep(i)
-          post_test_threads = set(t.ident for t in threading.enumerate())
-          new_threads = post_test_threads - pre_test_threads
-          if len(new_threads) == 0:
-            break
       self.assertEqual(len(new_threads), 0)
 
   def test_direct_runner_metrics(self):
@@ -132,5 +122,4 @@ class DirectPipelineResultTest(unittest.TestCase):
 
 
 if __name__ == '__main__':
-  logging.getLogger().setLevel(logging.INFO)
   unittest.main()
--- a/sdks/python/apache_beam/runners/portability/fn_api_runner.py
+++ b/sdks/python/apache_beam/runners/portability/fn_api_runner.py
@@ -1138,8 +1138,7 @@ class GrpcServer(object):
     self.state = state
     self.provision_info = provision_info
     self.control_server = grpc.server(
-        futures.ThreadPoolExecutor(max_workers=10,
-                                   thread_name_prefix='grpc_control_server'))
+        futures.ThreadPoolExecutor(max_workers=10))
     self.control_port = self.control_server.add_insecure_port('[::]:0')
     self.control_address = 'localhost:%s' % self.control_port
 
@@ -1149,14 +1148,12 @@ class GrpcServer(object):
     no_max_message_sizes = [("grpc.max_receive_message_length", -1),
                             ("grpc.max_send_message_length", -1)]
     self.data_server = grpc.server(
-        futures.ThreadPoolExecutor(max_workers=10,
-                                   thread_name_prefix='grpc_data_server'),
+        futures.ThreadPoolExecutor(max_workers=10),
         options=no_max_message_sizes)
     self.data_port = self.data_server.add_insecure_port('[::]:0')
 
     self.state_server = grpc.server(
-        futures.ThreadPoolExecutor(max_workers=10,
-                                   thread_name_prefix='grpc_state_server'),
+        futures.ThreadPoolExecutor(max_workers=10),
         options=no_max_message_sizes)
     self.state_port = self.state_server.add_insecure_port('[::]:0')
 
@@ -1191,8 +1188,7 @@ class GrpcServer(object):
         self.state_server)
 
     self.logging_server = grpc.server(
-        futures.ThreadPoolExecutor(max_workers=2,
-                                   thread_name_prefix='grpc_logging_server'),
+        futures.ThreadPoolExecutor(max_workers=2),
         options=no_max_message_sizes)
     self.logging_port = self.logging_server.add_insecure_port('[::]:0')
     beam_fn_api_pb2_grpc.add_BeamFnLoggingServicer_to_server(
@@ -1659,9 +1655,7 @@ class ParallelBundleManager(BundleManager):
 
     merged_result = None
     split_result_list = []
-    with futures.ThreadPoolExecutor(
-        max_workers=self._num_workers,
-        thread_name_prefix='parallel_bundle_processor') as executor:
+    with futures.ThreadPoolExecutor(max_workers=self._num_workers) as executor:
       for result, split_result in executor.map(lambda part: BundleManager(
           self._worker_handler_list, self._get_buffer,
           self._get_input_coder_impl, self._bundle_descriptor,
--- a/sdks/python/apache_beam/runners/portability/fn_api_runner.py
+++ b/sdks/python/apache_beam/runners/portability/fn_api_runner.py
@@ -90,11 +90,10 @@ class ControlConnection(object):
   def __init__(self):
     self._push_queue = queue.Queue()
     self._input = None
-    self._input_is_set = False
     self._futures_by_id = dict()
     self._read_thread = threading.Thread(
         name='beam_control_read', target=self._read)
-    self._state = BeamFnControlServicer.STARTED_STATE
+    self._state = BeamFnControlServicer.UNSTARTED_STATE
 
   def _read(self):
     for data in self._input:
@@ -117,11 +116,12 @@ class ControlConnection(object):
     return self._push_queue.get()
 
   def set_input(self, input):
-    if self._input_is_set:
-      raise RuntimeError('input is already set.')
-    self._input = input
-    self._read_thread.start()
-    self._input_is_set = True
+    with ControlConnection._lock:
+      if self._input:
+        raise RuntimeError('input is already set.')
+      self._input = input
+      self._read_thread.start()
+      self._state = BeamFnControlServicer.STARTED_STATE
 
   def close(self):
     if self._state == BeamFnControlServicer.STARTED_STATE:
@@ -131,7 +131,7 @@ class ControlConnection(object):
 
 
 class BeamFnControlServicer(beam_fn_api_pb2_grpc.BeamFnControlServicer):
-  """Implementation of BeamFnControlServicer for a single client."""
+  """Implementation of BeamFnControlServicer for clients."""
 
   UNSTARTED_STATE = 'unstarted'
   STARTED_STATE = 'started'
@@ -148,6 +148,7 @@ class BeamFnControlServicer(beam_fn_api_pb2_grpc.BeamFnControlServicer):
     self._req_sent = collections.defaultdict(int)
     self._req_worker_mapping = {}
     self._log_req = logging.getLogger().getEffectiveLevel() <= logging.DEBUG
+    self.connections_by_worker_id = collections.defaultdict(ControlConnection)
 
   def Control(self, iterator, context):
     with self._lock:
@@ -161,8 +162,7 @@ class BeamFnControlServicer(beam_fn_api_pb2_grpc.BeamFnControlServicer):
       raise RuntimeError('All workers communicate through gRPC should have '
                          'worker_id. Received None.')
 
-    conn_handler = GrpcServer.get_control_conn_handler()
-    control_conn = conn_handler[worker_id]
+    control_conn = self.connections_by_worker_id[worker_id]
     control_conn.set_input(iterator)
 
     while True:
@@ -174,9 +174,7 @@ class BeamFnControlServicer(beam_fn_api_pb2_grpc.BeamFnControlServicer):
         self._req_sent[to_push.instruction_id] += 1
 
   def done(self):
-    with self._lock:
-      self._state = self.DONE_STATE
-
+    self._state = self.DONE_STATE
     logging.debug('Runner: Requests sent by runner: %s',
                   [(str(req), cnt) for req, cnt in self._req_sent.items()])
     logging.debug('Runner: Requests multiplexing info: %s',
@@ -995,7 +993,7 @@ class WorkerHandler(object):
   """
 
   _registered_environments = {}
-  _worker_id = -1
+  _worker_id_counter = -1
 
   def __init__(
       self, control_handler, data_plane_handler, state, provision_info):
@@ -1012,13 +1010,11 @@ class WorkerHandler(object):
     self.state = state
     self.provision_info = provision_info
 
-  def get_worker_id(self):
-    WorkerHandler._worker_id += 1
-    return 'worker_%s' % WorkerHandler._worker_id
+    WorkerHandler._worker_id_counter += 1
+    self.worker_id = 'worker_%s' % WorkerHandler._worker_id_counter
 
   def close(self):
     self.stop_worker()
-    WorkerHandler._worker_id = -1
 
   def start_worker(self):
     raise NotImplementedError
@@ -1059,7 +1055,7 @@ class EmbeddedWorkerHandler(WorkerHandler):
     super(EmbeddedWorkerHandler, self).__init__(
         self, data_plane.InMemoryDataChannel(), state, provision_info)
     self.control_conn = self
-    self.data_conn = self.data_plane_handler.data_conn
+    self.data_conn = self.data_plane_handler
 
     self.worker = sdk_worker.SdkWorker(
         sdk_worker.BundleProcessorCache(
@@ -1068,7 +1064,6 @@ class EmbeddedWorkerHandler(WorkerHandler):
                 self.data_plane_handler.inverse()),
             {}))
     self._uid_counter = 0
-    self.worker_id = self.get_worker_id()
 
   def push(self, request, unused_worker_id=None):
     if not request.instruction_id:
@@ -1130,8 +1125,7 @@ class BasicProvisionService(
 class GrpcServer(object):
 
   _DEFAULT_SHUTDOWN_TIMEOUT_SECS = 5
-  _cached_grpc_server = None
-  instance_count = 0
+  _shared_grpc_server = None
   _lock = threading.Lock()
 
   def __init__(self, state, provision_info):
@@ -1141,6 +1135,7 @@ class GrpcServer(object):
         futures.ThreadPoolExecutor(max_workers=10))
     self.control_port = self.control_server.add_insecure_port('[::]:0')
     self.control_address = 'localhost:%s' % self.control_port
+    GrpcServer._shared_grpc_server = self
 
     # Options to have no limits (-1) on the size of the messages
     # received or sent over the data plane. The actual buffer size
@@ -1160,7 +1155,6 @@ class GrpcServer(object):
     self.control_handler = BeamFnControlServicer()
     beam_fn_api_pb2_grpc.add_BeamFnControlServicer_to_server(
         self.control_handler, self.control_server)
-    self.control_conn_handler = collections.defaultdict(ControlConnection)
 
     # If we have provision info, serve these off the control port as well.
     if self.provision_info:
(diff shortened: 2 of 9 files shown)


Hannah-Jiang:
Run Python PreCommit

Hannah-Jiang:
Run Portable_Python PreCommit

Hannah-Jiang:
R: @robertwb

Hannah-Jiang:
kindly ping.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1155,17 +1324,24 @@ def logging_api_service_descriptor(self):
...
-        self.data_server.stop(self._DEFAULT_SHUTDOWN_TIMEOUT_SECS),
-        self.state_server.stop(self._DEFAULT_SHUTDOWN_TIMEOUT_SECS),
-        self.logging_server.stop(self._DEFAULT_SHUTDOWN_TIMEOUT_SECS)
-    ]
-    for w in to_wait:
-      w.wait()
-    super(GrpcWorkerHandler, self).close()
+    # Wait all workers are done and close GRPC server.
+    WorkerHandler._worker_cnt -= 1
+    if WorkerHandler._worker_cnt == 0:
```
robertwb: Seems like this logic belongs in GrpcServer(). 

Maybe have a static GrpcServer.get_instance() method that does the caching and incrementing (rather than a static member on GrpcWorkerHandler) together with a release_instance() method that's called in GrpcWorkerHandler.close() (which may cause it to shut down). Right now one class is responsible for incrementing, another for decrementing, making it harder to follow.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1142,6 +1278,39 @@ def __init__(self, state, provision_info):
...
+    self.control_handler = self.grpc_server.control_handler
+
+    self.data_server = self.grpc_server.data_server
+    self.data_plane_handler = self.grpc_server.data_plane_handler
+    self.data_port = self.grpc_server.data_port
+
+    self.state_server = self.grpc_server.state_server
+    self.state_port = self.grpc_server.state_port
+
+    self.logging_server = self.grpc_server.logging_server
```
robertwb: Do we need a handle to all these servers?

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -995,9 +1099,37 @@ def create(cls, environment, state, provision_info):
...
+  """ This class handles multi worker handlers and round robin tasks to
+  workers."""
+
+  def __init__(self):
+    self._worker_handlers = {}
+
+  def _add_worker_handler(self, worker_handler):
+    self._worker_handlers[worker_handler._worker_id] = worker_handler
+
+  def _get_worker_handler(self, worker_id=None):
```
robertwb: I'd split this into two methods, rather than having a variant return type based on whether worker_id is set.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -936,15 +1006,43 @@ def close(self):
       pass
 
 
+class WorkerInfo(object):
```
robertwb: I think this class can go away, its parts merged into either WorkerHandler and ControlConnection (above).

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1289,26 +1466,30 @@ def __init__(self, environments, job_provision_info=None):
     self._cached_handlers = {}
     self._state = FnApiRunner.StateServicer() # rename?
 
-  def get_worker_handler(self, environment_id):
+  def get_worker_handler(self, environment_id, num_workers):
```
robertwb: This is plural now, right?

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1289,26 +1466,30 @@ def __init__(self, environments, job_provision_info=None):
...
 
-    worker_handler = self._cached_handlers.get(environment_id)
-    if worker_handler is None:
-      worker_handler = self._cached_handlers[
-          environment_id] = WorkerHandler.create(
-              environment, self._state, self._job_provision_info)
-      worker_handler.start_worker()
-    return worker_handler
+    worker_handler_list = self._cached_handlers.get(environment_id)
+    if worker_handler_list is None:
```
robertwb: What if it's the wrong size?

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -995,9 +1099,37 @@ def create(cls, environment, state, provision_info):
         provision_info)
 
 
+class MultiWorkerWorkerHandler(object):
```
robertwb: Is there a reason for this class to be distinct from WorkerHandlerManager?

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -92,34 +92,78 @@ class BeamFnControlServicer(beam_fn_api_pb2_grpc.BeamFnControlServicer):
...
   def __init__(self):
-    self._push_queue = queue.Queue()
     self._futures_by_id = dict()
-    self._read_thread = threading.Thread(
-        name='beam_control_read', target=self._read)
+    self._read_thread = None
     self._uid_counter = 0
     self._state = self.UNSTARTED_STATE
     self._lock = threading.Lock()
+    self._multi_worker_worker_handler = None
```
robertwb: I think we don't need to have this here anymore with the other proposed changes.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1480,42 +1664,54 @@ def process_bundle(self, inputs, expected_outputs):
...
     # Actually start the bundle.
     process_bundle_req = beam_fn_api_pb2.InstructionRequest(
         instruction_id=process_bundle_id,
         process_bundle=beam_fn_api_pb2.ProcessBundleRequest(
             process_bundle_descriptor_reference=self._bundle_descriptor.id))
-    result_future = self._controller.control_handler.push(process_bundle_req)
+    result_future = self._worker_handler.control_handler.push(
+        process_bundle_req, self._worker_handler._worker_id)
+
+    if not split_manager and isinstance(self._worker_handler,
```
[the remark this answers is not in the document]
Hannah-Jiang: What I observed here is, with non-GRPC worker handler, if we send request before sending data, bundle is treated as empty and processed before data is sent so it always return nothing. With GRPC worker handlers, request should be sent before data, because we need to multiplexing data according to the request. That's why I split it into two parts, but I don't think this is ideal way to handle this case.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -92,34 +92,78 @@ class BeamFnControlServicer(beam_fn_api_pb2_grpc.BeamFnControlServicer):
...
-    for data in self._inputs:
-      self._futures_by_id.pop(data.instruction_id).set(data)
+  def _read(self, worker):
+    for data in worker._get_input():
+      with self._lock:
+        if data.instruction_id in self._futures_by_id:
+          self._futures_by_id.pop(data.instruction_id).set(data)
+      response_type = data.WhichOneof('response')
+      if response_type == 'process_bundle':
+        worker._change_work_load(-1)
```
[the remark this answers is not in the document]
Hannah-Jiang: work_load is used at `_get_available_worker_handler()`, without this, do we want to randomly assign workers?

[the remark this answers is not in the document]
robertwb: We could, e.g., round robin with a counter. Essentially, we create N workers, use N workers, and are then done with N workers, so we can keep this logic nice and local.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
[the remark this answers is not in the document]
robertwb: Using the ideas above, we should avoid the need to send the request before the data in the GRPC case. (It is true that split_manager is incompatible with EmbeddedWorkerHandler, wouldn't hurt to assert this.)

(Nit: it's not clear that `isinstance(GrpcWorkerHandler)` and `isinstance(EmbeddedWorkerHandler)` cover both cases, e.g. someone could add a third type without knowing they had to fix the logic here. If we continued with this, it'd be better to do `isinstance(GrpcWorkerHandler)` vs. `not isinstance(GrpcWorkerHandler)`)

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1142,6 +1278,39 @@ def __init__(self, state, provision_info):
...
 
+
+class GrpcWorkerHandler(WorkerHandler):
+  """An grpc based worker_handler for fn API control, state and data planes."""
+
+  _DEFAULT_SHUTDOWN_TIMEOUT_SECS = 5
+  _cached_grpc_server = None
+
+  def __init__(self, state, provision_info):
+    if not GrpcWorkerHandler._cached_grpc_server:
```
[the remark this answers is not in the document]
Hannah-Jiang: This one is called within a for loop, do we need locking here?

Hannah-Jiang:
Thanks Robert for reviewing my PR, as always, your comments are very helpful.
I fixed all comments except one above which I didn't resolve.
Can you please take a look? Thank you.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1165,7 +1251,54 @@ def close(self):
...
     for w in to_wait:
       w.wait()
-    super(GrpcWorkerHandler, self).close()
+
+
+class GrpcWorkerHandler(WorkerHandler):
+  """An grpc based worker_handler for fn API control, state and data planes."""
+
+  _lock = threading.Lock()
+  control_conn_handler = ConnectionHandler()
```
robertwb: As before, avoid global state.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1142,17 +1230,15 @@ def __init__(self, state, provision_info):
...
 
-  def data_api_service_descriptor(self):
-    return endpoints_pb2.ApiServiceDescriptor(
-        url='localhost:%s' % self.data_port)
-
-  def state_api_service_descriptor(self):
-    return endpoints_pb2.ApiServiceDescriptor(
-        url='localhost:%s' % self.state_port)
+  @classmethod
+  def get_instance(cls, state, provision_info):
```
robertwb: This method and `release_instance` should do the reference counting and decision whether to "actually" close and deallocate the server, not the callers. That keeps it more local and easier to reason about.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1067,10 +1155,10 @@ def GetProvisionInfo(self, request, context=None):
         info=self._info)
 
 
-class GrpcWorkerHandler(WorkerHandler):
-  """An grpc based controller for fn API control, state and data planes."""
+class GrpcServer(object):
 
   _DEFAULT_SHUTDOWN_TIMEOUT_SECS = 5
+  _cached_grpc_server = None
```
robertwb: This is the one singleton global instance we should have. The fact that it's a singleton is just an optimization, hidden in its own `get_instance` method (which could return new instances each time, or some other strategy, if desired). All other state should be referenced off of its instance state (transitively).

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1165,7 +1251,54 @@ def close(self):
...
       w.wait()
-    super(GrpcWorkerHandler, self).close()
+
+
+class GrpcWorkerHandler(WorkerHandler):
+  """An grpc based worker_handler for fn API control, state and data planes."""
+
+  _lock = threading.Lock()
+  control_conn_handler = ConnectionHandler()
+  data_conn_handler = data_plane.ConnectionHandler()
```
robertwb: Get the data connection from the grpc server instance, not a global.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1142,6 +1278,39 @@ def __init__(self, state, provision_info):
...
+      self.grpc_server = GrpcWorkerHandler._cached_grpc_server = \
+        GrpcServer(state, provision_info)
+    else:
+      self.grpc_server = GrpcWorkerHandler._cached_grpc_server
+
+    self.state = state
+
+    self.control_server = self.grpc_server.control_server
+    self.control_address = self.grpc_server.control_address
+    self.control_handler = self.grpc_server.control_handler
```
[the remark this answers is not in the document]
robertwb: GrpcWorkerHandler doesn't need a connection to the control handler, just the control connection to its worker(id). Similarly for the data channel(s).

sdks/python/apache_beam/runners/worker/data_plane.py
```diff
@@ -300,42 +300,35 @@ def __init__(self, data_stub):
...
+      conn_handler.add(self._WRITES_FINISHED)
     self._closed = True
+    self._conn_handler.close()
 
   def output_stream(self, instruction_id, transform_id):
     def add_to_send_queue(data):
       if data:
-        worker_id = self._task_worker_mapping[instruction_id]
-        self._to_send_per_worker[worker_id].put(
+        worker_id = 'worker_%s' % (int(instruction_id.split('_')[1]) %
```
robertwb: It seems brittle to depend on the way workers are named here. Instead, split things up as with the Control connection, putting the bulk of the logic (including methods like `output_stream` and `_write_outputs`) onto the `ChannelConnection` connection object that already only holds a connection to the right worker.

sdks/python/apache_beam/runners/worker/data_plane.py
```diff
@@ -300,42 +300,35 @@ def __init__(self, data_stub):
...
-class GrpcServerMultiPlexDataChannel(GrpcServerDataChannel):
-  """A DataChannel wrapping the server side of a BeamFnData connection,
-  implementing data multiplexing."""
-
-  def __init__(self, task_worker_mapping):
+  def __init__(self):
     super(GrpcServerDataChannel, self).__init__()
-    self._to_send_per_worker = collections.defaultdict(queue.Queue)
-    self._task_worker_mapping = task_worker_mapping
+    self._conn_handler = ConnectionHandler()
```
robertwb: As above, a defaultdict should suffice.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
Hannah-Jiang: I didn't quite get it, can you please explain a little more?

Hannah-Jiang:
I made some changes, but there are still some places use global variables. In addition, it gets stuck with GrpcWorkerHandler, and happens more frequently if increase num of workers. I guess I missed lock somewhere. I am trying to merge this PR by 7/31, which is cut date for beam 2.15, so would like it to be reviewed while I am investigating the lock issue.

sdks/python/apache_beam/runners/worker/data_plane.py
```diff
@@ -69,22 +69,78 @@ def close(self):
...
+    self.data_conn._start_reader(data_stub.Data(
+        self.data_conn._write_outputs()))
+
+
+class GrpcServerDataChannel(
+    beam_fn_api_pb2_grpc.BeamFnDataServicer, DataChannel):
+  """A DataChannel wrapping the server side of a BeamFnData connection."""
+
+  def __init__(self):
+    self.data_conn = GrpcDataChannelConnection()
```
robertwb: I'm not following this--nothing has connected to the service at this point. 

I'm sorry I wasn't clear enough (this kind of thing is much more easily explained in person at a whiteboard...) Here the notion of Channel is already equivalent to Connection. 

If you look at https://github.com/apache/beam/blob/release-2.14.0/sdks/python/apache_beam/runners/worker/data_plane.py#L300 we have GrpcServerDataChannel inheriting from both beam_fn_api_pb2_grpc.BeamFnDataServicer and _GrpcDataChannel, because these used to be in a 1:1 relationship. What we need to do is split them apart, such that the server class `GrpcServerDataChannel` (rename to BeamFnDataServicer or similar) holds a [default]dict of `_GrpcDataChannel`s indexed by worker id, and does the appropriate thread starting on each call to `Data()`. Consumer would then go from the data servicer to the channel associated with the worker they care about, rather than assume they are one and the same.

I think this should be a much smaller change.

sdks/python/apache_beam/runners/worker/data_plane.py
```diff
@@ -69,22 +69,78 @@ def close(self):
...
+
+  def Data(self, elements_iterator, context):
+    # py27 doesn't support recursive import at module level
```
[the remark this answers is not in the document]
Hannah-Jiang: circular imports work fine with py3, at least with py36. I haven't tested with py35 and py37.
But we don't need it any more after creating `connections_by_worker_id` at channel rather than at runner side.

sdks/python/apache_beam/runners/direct/direct_runner_test.py
```diff
@@ -52,6 +54,15 @@ def test_waiting_on_result_stops_executor_threads(self):
...
       new_threads = post_test_threads - pre_test_threads
+      if len(new_threads) > 0:
+        for i in range(1,6):
+          logging.info('%s: Wait %s second(s) for thread(s) to close.' %
+                       (runner, i))
+          time.sleep(i)
+          post_test_threads = set(t.ident for t in threading.enumerate())
+          new_threads = post_test_threads - pre_test_threads
+          if len(new_threads) == 0:
+            break
```
Hannah-Jiang: When I test at local, 10% of time (sample size: 4,000) threads are closed after waiting 1 second with `DirectRunner` and `SwitchingDirectRunner`. 90% of time threads are closed immediately with all runners.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1010,7 +1065,7 @@ def __init__(self, unused_payload, state, provision_info):
             {}))
     self._uid_counter = 0
 
-  def push(self, request):
+  def push(self, request, unused_worker_id=None):
```
robertwb: We should be able to delete this unused argument everywhere.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1067,10 +1122,11 @@ def GetProvisionInfo(self, request, context=None):
         info=self._info)
 
 
-class GrpcWorkerHandler(WorkerHandler):
-  """An grpc based controller for fn API control, state and data planes."""
+class GrpcServer(object):
 
   _DEFAULT_SHUTDOWN_TIMEOUT_SECS = 5
+  _shared_grpc_server = None
```
robertwb: Remove. (I think the lock is unused as well.)

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1286,30 +1370,35 @@ class WorkerHandlerManager(object):
...
-
-    worker_handler = self._cached_handlers.get(environment_id)
-    if worker_handler is None:
-      worker_handler = self._cached_handlers[
-          environment_id] = WorkerHandler.create(
-              environment, self._state, self._job_provision_info)
-      worker_handler.start_worker()
-    return worker_handler
+    worker_handler_list = self._cached_handlers[environment_id]
+    if not worker_handler_list or len(worker_handler_list) < num_workers:
```
robertwb: `not worker_handler_list` will be subsumed by `len(worker_handler_list) < num_workers`.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1142,6 +1199,40 @@ def __init__(self, state, provision_info):
...
+    for w in to_wait:
+      w.wait()
+    GrpcServer._shared_grpc_server = None
+
+
+class GrpcWorkerHandler(WorkerHandler):
+  """An grpc based worker_handler for fn API control, state and data planes."""
+
+  def __init__(self, state, provision_info):
+    self.grpc_server = GrpcServer._shared_grpc_server
```
robertwb: The GrpcServer should be passed through WorkerHandler.create(...) to this constructor, not implicitly via a global (which may get crossed if there is more than one pipeline).

Hannah-Jiang:
The reason threads are not all killed immediately for direct runners for 10% of time is we are starting grpc servers and close it with direct runners. Threads created for grpc servers at FnApi side are all closed immediately, but it seems like a thread at grpc library is closed a little late. If confirmed that if we don't start grpc servers for direct runner, then all thread are closed immediately, so it would be resolved when we don't create grpc servers for direct embedded runners.

A side note about thread name is, we can add thread_name_prefix for ThreadPoolExecutor. This is supported by concurrent lib from v3.6. However, both Python2 and Python3 are using [concurrent v3.2](https://docs.python.org/3/library/concurrent.futures.html), so even if we add it, thread names are in default format for threads created through ThreadPoolExecutor. However, it is still a good practice to add it in case concurrent lib version is updated in future and we will have correct thread names for free.

sdks/python/apache_beam/runners/direct/direct_runner_test.py
```diff
@@ -43,15 +45,23 @@ class DirectPipelineResultTest(unittest.TestCase):
   def test_waiting_on_result_stops_executor_threads(self):
     pre_test_threads = set(t.ident for t in threading.enumerate())
 
-    for runner in ['DirectRunner', 'BundleBasedDirectRunner',
+    for runner in ['DirectRunner', #'BundleBasedDirectRunner',
```
robertwb: Don't forget to revert this line.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -92,52 +141,46 @@ class BeamFnControlServicer(beam_fn_api_pb2_grpc.BeamFnControlServicer):
...
-    self._inputs = iterator
-    # Note: We only support one client for now.
-    self._read_thread.start()
+
+    worker_id = dict(context.invocation_metadata()).get('worker_id')
+    if not worker_id:
+      raise RuntimeError('All workers communicate through gRPC should have '
+                         'worker_id. Received None.')
+
+    control_conn = self.connections_by_worker_id[worker_id]
```
robertwb: I just realized that defaultdicts may not be threadsafe, which could cause a rare but hard-to-track-down bug here. 

We should make this private and move all direct references to this dict to a method that grabs a lock before accessing it. (Similarly for the data channel.)

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -961,6 +1010,9 @@ def __init__(
     self.state = state
     self.provision_info = provision_info
 
+    WorkerHandler._worker_id_counter += 1
```
robertwb: This needs to be locked.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -987,21 +1039,26 @@ def wrapper(constructor):
     return wrapper
 
   @classmethod
-  def create(cls, environment, state, provision_info):
+  def create(cls, environment, state, provision_info, grpc_server=None):
```
robertwb: Any reason to let grpc_server be omitted? Might be better to be an error to not explicitly provide it.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1142,6 +1202,38 @@ def __init__(self, state, provision_info):
...
+    self.data_port = grpc_server.data_port
+    self.data_conn = grpc_server.data_plane_handler \
+      .connections_by_worker_id[self.worker_id]
+
+    self.state_port = grpc_server.state_port
+    self.logging_port = grpc_server.logging_port
+
   def data_api_service_descriptor(self):
     return endpoints_pb2.ApiServiceDescriptor(
         url='localhost:%s' % self.data_port)
```
robertwb: Maybe write this as `url='localhost:%s' % self.grpc_server.data_port` rather than storing `data_port` on `self`. Similarly for the others.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1286,30 +1375,36 @@ class WorkerHandlerManager(object):
   def __init__(self, environments, job_provision_info=None):
     self._environments = environments
     self._job_provision_info = job_provision_info
-    self._cached_handlers = {}
+    self._cached_handlers = collections.defaultdict(list)
     self._state = FnApiRunner.StateServicer() # rename?
+    # TODO(hannahjiang): create a grpc_server only for grpc env.
```
robertwb: We have to resolve this TODO before we merge, as by far the most common environment (e.g. for almost all our tests) is embedded.

Hannah-Jiang: I am curious when we have multiple environments for a pipeline?

Hannah-Jiang:
> The reason threads are not all killed immediately for direct runners for 10% of time is we are starting grpc servers and close it with direct runners. Threads created for grpc servers at FnApi side are all closed immediately, but it seems like a thread at grpc library is closed a little late. If confirmed that if we don't start grpc servers for direct runner, then all thread are closed immediately (sample size is 3,000.), so it would be resolved when we don't create grpc servers for direct embedded runners.
> 
> A side note about thread name is, we can add thread_name_prefix for ThreadPoolExecutor, which is supported by concurrent lib from v3.6. However, threads names created at grpc lib are in the default format. (Thread-N).

I noticed a problem with passing `thread_name_prefix` to `ThreadPoolExecutor`. `thread_name_prefix` works fine with py27, py36 and py37, but py35 throws out `got an unexpected keyword argument 'thread_name_prefix'` error everywhere. If we want to pass thread_name_prefix`, we need to add a py version check everywhere, can we skip it for now or it's still better to add it?

robertwb:
Looks good to me, pending tests, lint, etc. are happy. 

For now, lets skip adding thread_name_prefix due to lack of support on 3.5. It would probably still be a good thing to have (e.g. providing our own constructor that conditionally passes this argument down) but not needed as part of this PR.

Hannah-Jiang:

sdks/python/apache_beam/runners/portability/fn_api_runner.py
robertwb: Multi-language. (E.g. python might run embedded, go and java via docker.)

Hannah-Jiang:
All comments are resolved and it passed all tests. Is it possible to merge to master if it looks good?
Thanks.

yifanzou:
@robertwb Would you please take another look on this PR when you have a moment? This is a blocking issue of 2.15 release. We have to decide whether roll it out in 2.15 or postpone to the next release. Thanks.

PR #8828, opened 2019-06-11
[BEAM-7534] add --mountTempDir option for Docker container

ihji:
add --mountTempDir option for easier data sharing with Docker container. This will enable us to test FnApi features without GCS connectivity.

------------------------

Thank you for your contribution! Follow this checklist to help us incorporate your contribution quickly and easily:

 - [ ] [**Choose reviewer(s)**](https://beam.apache.org/contribute/#make-your-change) and mention them in a comment (`R: @username`).
 - [ ] Format the pull request title like `[BEAM-XXX] Fixes bug in ApproximateQuantiles`, where you replace `BEAM-XXX` with the appropriate JIRA issue, if applicable. This will automatically link the pull request to the issue.
 - [ ] If this contribution is large, please file an Apache [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

Lang | SDK | Apex | Dataflow | Flink | Gearpump | Samza | Spark
--- | --- | --- | --- | --- | --- | --- | ---
Go | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go/lastCompletedBuild/) | --- | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_VR_Flink/lastCompletedBuild/) | --- | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_VR_Spark/lastCompletedBuild/)
[...]

PR #8982, opened 2019-07-02
[BEAM-5440] Pass docker run options to SDK harness containers

sambvfx:
This adds support for passing docker run options through to the SDK harness container. The goal was to support mounting volumes, but is generalized to capture other `docker run [OPTIONS]` the user may desire. (`--user`?)

After reviewing the related [jira issue](https://issues.apache.org/jira/browse/BEAM-5440), I chose the route of modifying the `DockerPayload` proto, and did not add additional `PortableOptions` flags. Instead these options are being parsed from the existing `PortableOptions.environment_config` within the SDK. I found how the `PortableOptions.environment_type` and `PortableOptions.environment_config` work fairly unintuitive and adding this additional esoteric behavior was acceptable for now.

```
--environment_config "-v /tmp/beam_test:/tmp/beam_test {container_name}"
```

With these changes, I've successfully executed a simple beam graph with a flink runner (via the python sdk) that mounts a volume and and touches a file. These changes should not break existing pipelines that simply omit any of the docker options.

Assuming this is the desired approach, remaining issues would be:

- Port changes to other SDKS
- Add concept of `DockerPayload.options` to other parts of the core code. e.g. [Environments.createDockerEnvironment](https://github.com/apache/beam/blob/master/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/Environments.java#L117).
- Better tests? I've added a simple test in `portability_runner_test.py` confirming the parsing/construction of the DockerPayload proto.
- I also updated `DockerSdkWorkerHandler`, but as far as I can tell this isn't being used anywhere?

Any additional guidance would be appreciated!

--- | --- | --- | --- | ---

### DIFF
--- a/runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/environment/DockerEnvironmentFactory.java
+++ b/runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/environment/DockerEnvironmentFactory.java
@@ -22,6 +22,7 @@ import static org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.More
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.time.Duration;
+import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -116,6 +117,7 @@ public class DockerEnvironmentFactory implements EnvironmentFactory {
 
     // Prepare docker invocation.
     String containerImage = dockerPayload.getContainerImage();
+    String dockerOptions = dockerPayload.getOptions();
     // TODO: https://issues.apache.org/jira/browse/BEAM-4148 The default service address will not
     // work for Docker for Mac.
     String loggingEndpoint = loggingServiceServer.getApiServiceDescriptor().getUrl();
@@ -134,6 +136,10 @@ public class DockerEnvironmentFactory implements EnvironmentFactory {
 
     if (!retainDockerContainer) {
       dockerArgsBuilder.add("--rm");
+    }
+
+    if (!dockerOptions.isEmpty()) {
+      dockerArgsBuilder.addAll(Arrays.asList(dockerOptions.split("\\s+")));
     }
 
     List<String> args =
--- a/sdks/python/apache_beam/runners/portability/fn_api_runner.py
+++ b/sdks/python/apache_beam/runners/portability/fn_api_runner.py
@@ -32,6 +32,7 @@ import sys
 import threading
 import time
 import uuid
+import shlex
 from builtins import object
 from concurrent import futures
 
@@ -1329,6 +1330,7 @@ class DockerSdkWorkerHandler(GrpcWorkerHandler):
     super(DockerSdkWorkerHandler, self).__init__(state, provision_info,
                                                  grpc_server)
     self._container_image = payload.container_image
+    self._options = payload.options
     self._container_id = None
 
   def start_worker(self):
@@ -1337,22 +1339,36 @@ class DockerSdkWorkerHandler(GrpcWorkerHandler):
         subprocess.check_call(['docker', 'pull', self._container_image])
       except Exception:
         logging.info('Unable to pull image %s' % self._container_image)
-      self._container_id = subprocess.check_output(
-          ['docker',
-           'run',
-           '-d',
-           # TODO:  credentials
-           '--network=host',
-           self._container_image,
-           '--id=%s' % self.worker_id,
-           '--logging_endpoint=%s' % self.logging_api_service_descriptor().url,
-           '--control_endpoint=%s' % self.control_address,
-           '--artifact_endpoint=%s' % self.control_address,
-           '--provision_endpoint=%s' % self.control_address,
-          ]).strip()
-      while True:
-        logging.info('Waiting for docker to start up...')
-        status = subprocess.check_output([
+      cmd = [
+          'docker',
+          'run',
+          '-d',
+          # TODO:  credentials
+          '--network=host',
+      ]
+      if self._options:
+        cmd.extend(shlex.split(self._options))
+      cmd.extend([
+          self._container_image,
+          '--id=%s' % self.worker_id,
+          '--logging_endpoint=%s' % self.logging_api_service_descriptor().url,
+          '--control_endpoint=%s' % self.control_address,
+          '--artifact_endpoint=%s' % self.control_address,
+          '--provision_endpoint=%s' % self.control_address,
+      ])
+      self._container_id = subprocess.check_output(cmd).strip()
+    while True:
+      logging.info('Waiting for docker to start up...')
+      status = subprocess.check_output([
+          'docker',
+          'inspect',
+          '-f',
+          '{{.State.Status}}',
+          self._container_id]).strip()
+      if status == 'running':
+        break
+      elif status in ('dead', 'exited'):
+        subprocess.call([
             'docker',
             'inspect',
             '-f',
--- a/sdks/python/apache_beam/runners/portability/portable_runner.py
+++ b/sdks/python/apache_beam/runners/portability/portable_runner.py
@@ -121,13 +121,17 @@ class PortableRunner(runner.PipelineRunner):
             'Unknown environment type: %s' % portable_options.environment_type)
 
     if environment_urn == common_urns.environments.DOCKER.urn:
-      docker_image = (
-          portable_options.environment_config
-          or PortableRunner.default_docker_image())
+      if not portable_options.environment_config:
+        docker_image = PortableRunner.default_docker_image()
+        docker_options = ''
+      else:
+        docker_options, _, docker_image = portable_options.environment_config \
+          .strip().rpartition(' ')
       return beam_runner_api_pb2.Environment(
           urn=common_urns.environments.DOCKER.urn,
           payload=beam_runner_api_pb2.DockerPayload(
-              container_image=docker_image
+              container_image=docker_image,
+              options=docker_options
           ).SerializeToString())
     elif environment_urn == common_urns.environments.PROCESS.urn:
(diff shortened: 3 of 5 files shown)


sambvfx:
R: @lukecwik
R: @angoenka
R: @pabloem
R: @charlesccychen

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1235,26 +1235,32 @@ class DockerSdkWorkerHandler(GrpcWorkerHandler):
...
+      cmd.extend(self._options.split(' '))
+    cmd.extend([
+      self._container_image,
+      '--id=%s' % uuid.uuid4(),
+      '--logging_endpoint=%s' % self.logging_api_service_descriptor().url,
+      '--control_endpoint=%s' % self.control_address,
+      '--artifact_endpoint=%s' % self.control_address,
+      '--provision_endpoint=%s' % self.control_address,
+    ])
+    self._container_id = subprocess.check_output(cmd).strip()
```
mxm: As far as I can see this is just reformating? Please consider reverting since this is unrelated.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/environment/DockerEnvironmentFactory.java
```diff
@@ -136,6 +138,10 @@ public RemoteEnvironment createEnvironment(Environment environment) throws Excep
       dockerArgsBuilder.add("--rm");
     }
 
+    if (!dockerOptions.isEmpty()) {
+      dockerArgsBuilder.addAll(Arrays.asList(dockerOptions.split("\\s+")));
```
mxm: Can we make the options JSON encoded?

sdks/python/apache_beam/runners/portability/portable_runner.py
```diff
@@ -128,13 +128,17 @@ def _create_environment(options):
...
     if environment_urn == common_urns.environments.DOCKER.urn:
-      docker_image = (
-          portable_options.environment_config
-          or PortableRunner.default_docker_image())
+      if not portable_options.environment_config:
+        docker_image = PortableRunner.default_docker_image()
+        docker_options = ''
+      else:
+        docker_options, _, docker_image = portable_options.environment_config \
+          .strip().rpartition(' ')
```
mxm: I believe this should decode a JSON config string here.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
sambvfx: The reformatting was necessary to add the optional `DockerPayload.options` into the docker run command.

```python
    if self._options:
      cmd.extend(self._options.split(' '))
```

I didn't see anywhere this class was being used... I'm happy to revert if we don't want/need this functionality here.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/environment/DockerEnvironmentFactory.java
sambvfx: Thanks for taking a look @mxm! 

I'm not sure I understand the desire to introduce JSON here. I admit this java string parsing is not very robust. Is the motivation to move the docker option shell string parsing into the SDKs and send along an encoded list of strings?

For reference the docker options could be anything acceptable to pass into `docker run [OPTIONS]`.

```
'-v /Volumes/mounts/foo:/Volumes/mounts/foo:ro'
'--user sambvfx'
'--env "ENV_VAR=bar"'
```

If this was python I would likely parse the string with [shlex.split](https://docs.python.org/2.7/library/shlex.html#shlex.split) - is there an equivalent in Java for safely parsing shell-style strings?

ihji: +1 for JSON encoded options (like process environment options):
https://github.com/apache/beam/blob/master/sdks/python/apache_beam/options/pipeline_options.py#L805

sambvfx: Are you suggesting we attempt to keep the flags in sync with our own json structure? i.e. 

```python
{
    'volumes': ['/Volumes/mnt/foo:/Volumes/mnt/foo', '/Volumes/mnt/bar:/Volumes/mnt/bar'], 
    'user': 'sambvfx',
    'env': {'<Environment variables 1>': '<ENV_VAL>'},
}
```

I don't think that design is very maintainable. There are nearly 100 flags for `docker run [OPTIONS]`. This JSON structure would have to be understood and properly unpacked by the `DockerEnvironmentFactory` - increasing the complexity of the code there.

The JSON approach would also make it nearly impossible to provide as argparse flags to the SDK. Which currently looks like this:

```
python -m mymodule --runner PortableRunner --job_endpoint localhost:8099 --environment_type DOCKER --environment_config "-v /Volumes/mnt/foo:/Volumes/mnt/foo -v /Volumes/mnt/bar:/Volumes/mnt/bar --user sambvfx MY_CONTAINER_NAME"
```

ihji: I didn't mean we should keep all the flags in sync. I'm simply suggesting (don't know how @mxm thinks but at least for me) that it would be better if we could split the container URL and the docker run options i.e.
```
{
    'container_name': 'MY_CONTAINER_NAME',
    'run_options': '-v /Volumes/mnt/foo:/Volumes/mnt/foo -v /Volumes/mnt/bar:/Volumes/mnt/bar --user sambvfx'
}
```

sambvfx: I'm happy with your proposed change as it mirrors the `DockerPayload` proto. I was confused because this json request was attached to the java code, but this would only exist within the SDK before we unpack it into the protobuf.

It's worth noting the downsides of this would be:
- backwards incompatible
- difficult to provide json as an argparse argument (within the SDK)

Both are inconsequential to me.

chadrik:
I think we are going to want to do this in a backward compatible way. Any
suggestions?

lukecwik:
In this PR is that we assume that `docker run` is the only way to start a docker container which means that users who pass in `run_options` will only work with runners that use `docker run` to launch the containers. Otherwise the runners will need to either reject jobs with run options or interpret all the options found and either ignore them if they aren't important or map them onto something meaningful that they use themselves.

Also, does `docker run` maintain backwards compatibility for their flags or will users who run their pipeline locally then have issues when deploying them on a cluster based runner with a potentially different version of docker (is there a concept of min docker version needed?)?

mxm:
I agree that we shouldn't be passing Docker options directly to the Docker command. Instead like in #8828 we should have an option for mounting a directory which can then be translated by Beam for the Docker version available. Adding this flag to the Docker environment, like done in this PR, looks like a feasible approach. Best to consult the mailing list though.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/environment/DockerEnvironmentFactory.java
mxm: Sorry for the delay here. The JSON part here is only for configuration from the SDK side, also called environment config. The Proto changes should remain as-is.

It is ok to make this backwards-incompatible change now. For the next release we will publish containers and then the environment config format should be consistent. Also, the JSON structure will make it much easier to be backwards-compatible in future versions. If you want to remain backwards-compatibility, you could also check if we have a JSON string or not and still support the old parsing, but IMHO not needed.

chadrik: In reviewing this I'm a bit concerned about the splitting on whitespace issue that @sambvfx pointed out.  It's going to break for someone eventually with something like:

```json
{
    "container_name": "MY_CONTAINER_NAME",
    "run_options": "-v \"/Volumes/mnt/terrible folder name\""
}
```

This stack overflow post on the subject does not have any easy solutions: https://stackoverflow.com/questions/1082953/shlex-alternative-for-java

Our options are:

1. deal with this later.  it's a solvable problem but it does not _necessarily_ need to be solved now
2. try to write a `shlex.split` java port now as part of this PR.  @mxm  do you know of any open source options for this?
3. use a list of strings for the run options in the docker payload, which obviously makes this even harder for a user to type the json equivalent, but makes things easier to code.  The json would look like this: 
   ```json
   {
       "container_name": "MY_CONTAINER_NAME",
       "run_options": ["-v",  "/Volumes/mnt/terrible folder name"]
   }
   ```
4. same as 3, but as a convenience, use `shlex.split()` in python to prep the run_options if they are given as a string

mxm: +1 for a list of options to avoid problems with the splitting of whitespace-containing paths. Note that we could add an easier to use wrapper for the Docker option which would then simply add those options to the environment config. Given that the configuration is for advanced use cases anyway, the proposed solution looks fine.

aromanenko-dev:
What is a status of this PR/Jira? What are the next steps to move it forward? I believe that passing the user docker options is a very useful feature.

lukecwik:
> What is a status of this PR/Jira? What are the next steps to move it forward? I believe that passing the user docker options is a very useful feature.

There was a concern about exposing all docker run options since it would likely be incompatible with cluster managers that use docker.

The discussion was in https://lists.apache.org/thread.html/fa3fede5d2967daab676d3869d2fc94f505517504c3a849db789ef54%40%3Cdev.beam.apache.org%3E

aromanenko-dev:
Thanks @lukecwik for details.

PR #8983, opened 2019-07-02
[BEAM-7640] Rename the package name for amazon-web-services2 from aws to aws2

iemejia:
This is needed to avoid name collisions at debug time with the
amazon-web-services module. Also it allows users to migrate
progressively to the new module and even be able to mix both if
required.

R: @aromanenko-dev

### DIFF
--- a/sdks/java/io/amazon-web-services2/src/main/java/org/apache/beam/sdk/io/aws2/options/AwsModule.java
+++ b/sdks/java/io/amazon-web-services2/src/main/java/org/apache/beam/sdk/io/aws2/options/AwsModule.java
@@ -15,7 +15,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.apache.beam.sdk.io.aws.options;
+package org.apache.beam.sdk.io.aws2.options;
 
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import com.fasterxml.jackson.core.JsonGenerator;
--- a/sdks/java/io/amazon-web-services2/src/main/java/org/apache/beam/sdk/io/aws2/options/AwsOptions.java
+++ b/sdks/java/io/amazon-web-services2/src/main/java/org/apache/beam/sdk/io/aws2/options/AwsOptions.java
@@ -15,7 +15,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.apache.beam.sdk.io.aws.options;
+package org.apache.beam.sdk.io.aws2.options;
 
 import org.apache.beam.sdk.annotations.Experimental;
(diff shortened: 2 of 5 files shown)


iemejia:
Run Java PreCommit

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/s3/DefaultS3ClientBuilderFactory.java
```diff
@@ -19,8 +19,8 @@
 
 import com.amazonaws.client.builder.AwsClientBuilder;
 import com.amazonaws.services.s3.AmazonS3ClientBuilder;
-import org.apache.beam.sdk.io.aws.options.S3ClientBuilderFactory;
-import org.apache.beam.sdk.io.aws.options.S3Options;
+import org.apache.beam.sdk.io.aws2.options.S3ClientBuilderFactory;
```
aromanenko-dev: Why do we need to change the package name for the classes from the current (v1) `amazon-web-services` module?

sdks/java/io/amazon-web-services2/src/main/java/org/apache/beam/sdk/io/aws2/options/AwsModule.java
```diff
@@ -15,7 +15,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.apache.beam.sdk.io.aws.options;
+package org.apache.beam.sdk.io.aws2.options;
```
aromanenko-dev: Don't we need to change a file path of this package (to `sdks/java/io/amazon-web-services2/src/main/java/org/apache/beam/sdk/io/aws2/...`) as well to reflect a package name?

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/s3/DefaultS3ClientBuilderFactory.java
iemejia:
```
This is needed to avoid name collisions at debug time with the
amazon-web-services module. Also it allows users to migrate
progressively to the new module and even be able to mix both if
required.
```

iemejia: oh wait this is a mess you are right

iemejia:
Hi I haven't noticed I had screwed the PR because of an IntelliJ error, It should be ok now. Please re review @aromanenko-dev

PR #8994, opened 2019-07-03
[BEAM-7711] Add DATETIME as a logical type in BeamSQL

amaliujia:
```
DATETIME as a type represents a year, month, day, hour, minute, second, and subsecond(millis)

it ranges from 0001-01-01 00:00:00 to 9999-12-31 23:59:59.999.
```

Current idea is use `Instant` to save DATETIME because DATE and TIME are already represented in this way. What really matters is how does functions work on this type, which shall be left for function implementations.

This is an **experimental** idea.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/utils/CalciteUtils.java
+++ b/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/utils/CalciteUtils.java
@@ -25,8 +25,6 @@ import org.apache.beam.sdk.schemas.LogicalTypes.PassThroughLogicalType;
 import org.apache.beam.sdk.schemas.Schema;
 import org.apache.beam.sdk.schemas.Schema.FieldType;
 import org.apache.beam.sdk.schemas.Schema.TypeName;
-import org.apache.beam.vendor.guava.v20_0.com.google.common.collect.BiMap;
-import org.apache.beam.vendor.guava.v20_0.com.google.common.collect.ImmutableBiMap;
 import org.apache.beam.vendor.guava.v20_0.com.google.common.collect.ImmutableMap;
 import org.apache.calcite.avatica.util.ByteString;
 import org.apache.calcite.rel.type.RelDataType;
@@ -57,6 +55,15 @@ public class CalciteUtils {
     public static final String IDENTIFIER = "SqlTimeType";
 
     public TimeType() {
+      super(IDENTIFIER, "", FieldType.DATETIME);
+    }
+  }
+
+  /** A LogicalType corresponding to DATETIME. */
+  public static class DateTimeType extends PassThroughLogicalType<Instant> {
+    public static final String IDENTIFIER = "SqlDateTimeType";
+
+    public DateTimeType() {
       super(IDENTIFIER, "", FieldType.DATETIME);
     }
   }
@@ -99,7 +106,8 @@ public class CalciteUtils {
       return logicalId.equals(DateType.IDENTIFIER)
           || logicalId.equals(TimeType.IDENTIFIER)
           || logicalId.equals(TimeWithLocalTzType.IDENTIFIER)
-          || logicalId.equals(TimestampWithLocalTzType.IDENTIFIER);
+          || logicalId.equals(TimestampWithLocalTzType.IDENTIFIER)
+          || logicalId.equals(DateTimeType.IDENTIFIER);
     }
     return false;
   }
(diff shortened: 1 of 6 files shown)


amaliujia:
R: @kennknowles @apilloud

amaliujia:
Run Java PreCommit

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/utils/CalciteUtils.java
```diff
@@ -154,6 +163,7 @@ public static boolean isStringType(FieldType fieldType) {
           .put(TIME_WITH_LOCAL_TZ, SqlTypeName.TIME_WITH_LOCAL_TIME_ZONE)
           .put(TIMESTAMP, SqlTypeName.TIMESTAMP)
           .put(TIMESTAMP_WITH_LOCAL_TZ, SqlTypeName.TIMESTAMP_WITH_LOCAL_TIME_ZONE)
+          .put(DATETIME, SqlTypeName.TIMESTAMP)
```
amaliujia: The challenging here is how to choose the SqlTypeName in Calcite for DateTime. As there is not such SqlTypeName exist for DateTime, a workaround is use `TIMESTAMP_WITH_LOCAL_TIME_ZONE` for DateTime.

apilloud: This seems to be the crux of the issue. This came from BigQuery? If I'm reading the docs correctly BigQuery `DATETIME` is what the SQL standard calls `TIMESTAMP [WITHOUT TIME ZONE]` (and Calcite calls `TIMESTAMP`). BigQuery `TIMESTAMP` is what the SQL standard calls `TIMESTAMP WITH TIME ZONE` (and Calcite calls [`TIMESTAMP_WITH_TIMEZONE`](https://github.com/apache/calcite/blob/9721283bd0ce46a337f51a3691585cca8003e399/core/src/main/java/org/apache/calcite/sql/type/SqlTypeFamily.java#L104) and maps to `TIMESTAMP`).

I think the mapping to `TIMESTAMP` is correct here but is this actually just `TIMESTAMP`? What is the benefit we get by adding an extra type?

amaliujia: Because if we have both `DATETIME`(`TIMESTAMP[WITHOUT TIME ZONE]`) and `TIMESTAMP`(`TIMESTAMP WITH TIME ZONE`) mapped to `TIMESTAMP`, we won't be able to differentiate these two types in Calcite logical plans(For example, schema is created from RowType in BeamCalcRel: https://github.com/apache/beam/blob/master/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamCalcRel.java#L122)

apilloud: If you map to `TIMESTAMP_WITH_LOCAL_TIME_ZONE` as a workaround, you end up with the same problem just with different data types. It seems that differentiating between the two types is not possible without a change to calcite.

apilloud: (Or a change to how we interact with calcite.)

amaliujia: For sure ideally the change should happen to Calcite (add the corresponding type for example).

As a workaround, what I was really thinking is  use `TIMESTAMP_WITH_LOCAL_TIME_ZONE` for `DateTime` and we actually disable `TIMESTAMP_WITH_LOCAL_TIME_ZONE` as a type for BeamSQL right now. If you think this is not acceptable, then I guess we have to make a change to Calcite.

kennknowles: What are the differences between DATETIME and TIMESTAMP WITHOUT TIME ZONE, so that they need to be distinguished?

amaliujia: I cannot find a precious definition of `TIMESTMAP WITHOUT TIME ZONE` from Calcite so I don't really know the difference.

However Calcite maps both `TIMESTMAP WITHOUT TIME ZONE` and ``TIMESTMAP WITH TIME ZONE` to `TIMESTAMP`

amaliujia: The problem is Beam ZetaSQL will only see Calcite Schema. If we treat `DATETIME` as Calcite's `TIMESTAMP` (thus mixed `DATETIME` with `TIMESTAMP WITH TIME ZONE`), it means Beam ZetaSQL loses the information of `DATETIME` type in plans.

kennknowles: I've done some reading and I agree that BQ DATETIME is the same as TIMESTAMP WITHOUT TIME ZONE. The conversion is lossless except for precision on the subsecond part. BQ TIMESTAMP is the same as TIMESTAMP WITH TIME ZONE.

The current mapping here is incorrect.

amaliujia: Yes. This mapping is not correct.

The discussion here is what mapping we should use for DATETIME. If you check SqlTypename, you will find there is no TIMESTAMP without TIME ZONE (or equivalent type) exist.

kennknowles: I believe that `TIMESTAMP` is already without time zone. See https://issues.apache.org/jira/browse/CALCITE-2394 for some discussion I have had about some of this.

amaliujia: It is interesting then. I haven't read the relevant JIRA but at least sounds like at least one of our mapping is wrong, either DATETIME or TIMESTAMP WITH TIME ZONE.

amaliujia:
There is still an open question for type mapping. I will clsoe this PR as I am not working on it so other people can pick it up.

PR #8998, opened 2019-07-03
Make OffsetRangeTracker.try_split() try best to split when suggested split is invalid

boyuanzz:
When OffsetRangeTracker.try_split() cannot perform split at suggested position, try best to split as soon as possible, aka, at current_processing_position + 1.

R: @lukecwik

### DIFF
--- a/sdks/python/apache_beam/io/range_trackers.py
+++ b/sdks/python/apache_beam/io/range_trackers.py
@@ -133,29 +133,44 @@ class OffsetRangeTracker(iobase.RangeTracker):
       self._validate_record_start(record_start, False)
       self._last_record_start = record_start
 
+  def _validate_split_offset(self, split_offset):
+    if split_offset <= self._last_record_start:
+      logging.debug(
+          'Split position %s at %r is not valid: '
+          'already past proposed stop offset', split_offset, self)
+      return False
+    if (split_offset < self.start_position()
+          or split_offset >= self.stop_position()):
+      logging.debug(
+          'Split position %s at %s is not valid: '
+          'proposed split position out of range', split_offset, self)
+      return False
+    return True
+
   def try_split(self, split_offset):
     assert isinstance(split_offset, (int, long))
     with self._lock:
       if self._stop_offset == OffsetRangeTracker.OFFSET_INFINITY:
-        logging.debug('refusing to split %r at %d: stop position unspecified',
+        logging.debug('Refusing to split %r at %d: stop position unspecified',
                       self, split_offset)
         return
+
       if self._last_record_start == -1:
         logging.debug('Refusing to split %r at %d: unstarted', self,
                       split_offset)
         return
 
-      if split_offset <= self._last_record_start:
-        logging.debug(
-            'Refusing to split %r at %d: already past proposed stop offset',
-            self, split_offset)
-        return
-      if (split_offset < self.start_position()
-          or split_offset >= self.stop_position()):
-        logging.debug(
-            'Refusing to split %r at %d: proposed split position out of range',
-            self, split_offset)
-        return
+      if not self._validate_split_offset(split_offset):
+        split_offset = self._last_attempted_record_start + 1
+        logging.debug('Suggested split position is not a valid split position,'
+                      'try to split as soon as possible at position %d.',
+                      split_offset)
+        # Adjusted split may reach to the end of the range.
+        if split_offset >= self._stop_offset:
+          logging.debug('Adjusted split has reached the end of the range.'
+                        'Cannot perform any split for range [%d, %d)',
+                        self._start_offset, self._stop_offset)
+          return
 
       logging.debug('Agreeing to split %r at %d', self, split_offset)
 
--- a/sdks/python/apache_beam/io/range_trackers.py
+++ b/sdks/python/apache_beam/io/range_trackers.py
@@ -140,7 +140,7 @@ class OffsetRangeTracker(iobase.RangeTracker):
           'already past proposed stop offset', split_offset, self)
       return False
     if (split_offset < self.start_position()
-          or split_offset >= self.stop_position()):
+        or split_offset >= self.stop_position()):
       logging.debug(
           'Split position %s at %s is not valid: '
           'proposed split position out of range', split_offset, self)
--- a/sdks/python/apache_beam/io/concat_source_test.py
+++ b/sdks/python/apache_beam/io/concat_source_test.py
@@ -94,7 +94,7 @@ class ConcatSourceTest(unittest.TestCase):
   def test_range_source(self):
     source_test_utils.assert_split_at_fraction_exhaustive(RangeSource(0, 10, 3))
 
-  def test_conact_source(self):
+  def test_concat_source(self):
     source = ConcatSource([RangeSource(0, 4),
                            RangeSource(4, 8),
                            RangeSource(8, 12),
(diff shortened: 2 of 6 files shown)


boyuanzz:
Run Portable_Python PreCommit

sdks/python/apache_beam/io/range_trackers.py
```diff
@@ -115,29 +115,42 @@ def set_current_position(self, record_start):
...
-            'Refusing to split %r at %d: already past proposed stop offset',
-            self, split_offset)
-        return
-      if (split_offset < self.start_position()
-          or split_offset >= self.stop_position()):
-        logging.debug(
-            'Refusing to split %r at %d: proposed split position out of range',
-            self, split_offset)
-        return
+      if not self._validate_split_offset(split_offset):
```
lukecwik: Refactoring and re-using this method twice will make the logging confusing since it will seem that we choose a bad split point and then choose another bad split point if we are at the end of the range. Can we restructure the if statements?

boyuanzz: Done in commit https://github.com/apache/beam/pull/8998/commits/8ddec14de1a16f7665975732ad917b97bd8a862a.

boyuanzz:
commit https://github.com/apache/beam/pull/8998/commits/8c84d575fa5576f7e95fdeb53d9f288709cd5cfb and https://github.com/apache/beam/pull/8998/commits/dbc9aa20544edb8d1531b18c224e40216dfcf62e fixes broken UTs, but needs to wait for https://github.com/apache/beam/pull/9011 checked in.

sdks/python/apache_beam/io/range_trackers.py
```diff
@@ -115,29 +115,44 @@ def set_current_position(self, record_start):
...
+        split_offset = self._last_record_start + 1
```
chamikaramj: Why are we updating the variable split_offset here ? This looks confusing since we are not doing anything with the update split_offset (other than logging).

sdks/python/apache_beam/io/range_trackers.py
```diff
@@ -115,29 +115,44 @@ def set_current_position(self, record_start):
...
+        logging.debug('Suggested split position is not a valid split position,'
+                      'try to split as soon as possible at position %d.',
```
chamikaramj: This can be done at any position after current split_offset right ? (not just split_offset + 1)

sdks/python/apache_beam/io/range_trackers.py
```diff
@@ -115,29 +115,44 @@ def set_current_position(self, record_start):
...
+                      split_offset)
+        # Adjusted split may reach to the end of the range.
+        if split_offset >= self._stop_offset:
+          logging.debug('Adjusted split has reached the end of the range.'
```
chamikaramj: Not sure how this log is useful either ? Split will be done by the system not the user (latter is the intended observer of the logs).

sdks/python/apache_beam/io/range_trackers.py
chamikaramj: Ah, I see now that we try to re-split at the updated position. But is "self._last_record_start + 1" what you want ? This split will also fail unless last record was 1 byte, no ? 

Why don't we just rely on the caller to do a better split instead of trying to do two splits in the same try_split() call ?

boyuanzz: > Ah, I see now that we try to re-split at the updated position.

Just for clarification, we are not trying to split twice. The adjust split only happens when the suggested split is not valid.

> But is "self._last_record_start + 1" what you want ? 

The idea here is, if the suggested split is invalid, the OffsetRangeTracker should split as soon as possible. self._last_record_start + 1 is next record start pos, right?

> This split will also fail unless last record was 1 byte, no ?

Based on how OffsetRangeTracker implemented, I think the adjusted split can only fail when self_.last_record_start has reach to the end of the range. Could you please explain more why you think the last record should be 1 byte?

> Why don't we just rely on the caller to do a better split instead of trying to do two splits in the same try_split() call ?

If the caller refers to SDK, the sdk cannot now what's the next position in the range_tracker,  since it depends on the implementation details of one range tracker. 
If the caller refers to a runner, the runner can only suggest a split. However, a runner can also maintain a split logic likes, asking for split first, if sdk refuses to split, then asking for checkpoint. But this is kind of out scope of this PR.

sdks/python/apache_beam/io/range_trackers.py
boyuanzz: right, and I chose to split as soon as possible, which has a high possibility to perform a valid split : )

sdks/python/apache_beam/io/range_trackers.py
boyuanzz: For me, when sdk refuses to split, I would like to know why, which may help optimize either suggested split position and adjusted one.

boyuanzz:
> commit [8c84d57](https://github.com/apache/beam/commit/8c84d575fa5576f7e95fdeb53d9f288709cd5cfb) and [dbc9aa2](https://github.com/apache/beam/commit/dbc9aa20544edb8d1531b18c224e40216dfcf62e) fixes broken UTs, but needs to wait for #9011 checked in.

#9011 closed after discussion with Cham. Have a new PR https://github.com/apache/beam/pull/9058 to address current issue.

boyuanzz:
R: @chamikaramj 
Could you please also help review this PR? Within https://github.com/apache/beam/pull/9058 checked in, the adjusted split point is changed to `last_attempted_record_start + 1`. 
There is an additional change in [source_test_utils_test.py](https://github.com/apache/beam/pull/8998/files#diff-e54774691ec4135689d2c64aa3622fff), which forces one more try_claim when reading the whole records. 
Thanks for your help!

boyuanzz:
Run Portable_Python PreCommit

boyuanzz:

boyuanzz:
kindly pinged.

sdks/python/apache_beam/io/range_trackers.py
```diff
@@ -133,29 +133,44 @@ def set_current_position(self, record_start):
...
-            'Refusing to split %r at %d: proposed split position out of range',
-            self, split_offset)
-        return
+      if not self._validate_split_offset(split_offset):
+        split_offset = self._last_attempted_record_start + 1
+        logging.debug('Suggested split position is not a valid split position,'
+                      'try to split as soon as possible at position %d.',
+                      split_offset)
+        # Adjusted split may reach to the end of the range.
+        if split_offset >= self._stop_offset:
```
chamikaramj: Can we just do a self._validate_split_offset(split_offset) again instead of this check ?

sdks/python/apache_beam/io/range_trackers.py
```diff
@@ -133,29 +133,44 @@ def set_current_position(self, record_start):
...
-            self, split_offset)
-        return
-      if (split_offset < self.start_position()
-          or split_offset >= self.stop_position()):
-        logging.debug(
-            'Refusing to split %r at %d: proposed split position out of range',
-            self, split_offset)
-        return
+      if not self._validate_split_offset(split_offset):
+        split_offset = self._last_attempted_record_start + 1
```
chamikaramj: I'm bit worried about the change of behavior due to this. This might end up creating lot more splits that we do today. But then again this might be OK since backend already asked the SDK to split. Did you run any large scale tests (for example, that use textio) to determine if you see any difference in performance characteristics ?

Also, what's the motivation for this change ? Do you think performance will improve since we re-split at an earlier position ? If so I think this should be backed by performance benchmarks that verifies that this change actually helps.

chamikaramj: Also, please add a JIRA.

PR #6845, opened 2018-10-26
[BEAM-5866] Fix `Row#equals`

kanterov:
This pull request defines two properties that any `Coder` implementation should satisfy. There are two coders that don't properly implement `structuralValue`: `MapCoder` and `RowCoder`.

`RowCoder#consistentWithEquals` is defined as always true. It isn't true in case schema has field of type `Map<BYTES, ?>`. However, such field type makes little sense. One of the ways to fix it would be to reject such schemas. For now, this test is ignored.

------------------------

Follow this checklist to help us incorporate your contribution quickly and easily:

 - [x] Format the pull request title like `[BEAM-XXX] Fixes bug in ApproximateQuantiles`, where you replace `BEAM-XXX` with the appropriate JIRA issue, if applicable. This will automatically link the pull request to the issue.
 - [ ] If this contribution is large, please file an Apache [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

It will help us expedite review of your Pull Request if you tag someone (e.g. `@username`) to look at it.

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

Lang | SDK | Apex | Dataflow | Flink | Gearpump | Samza | Spark
--- | --- | --- | --- | --- | --- | --- | ---
Go | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_GradleBuild/lastCompletedBuild/) | --- | --- | --- | --- | --- | ---
[...]

PR #9009, opened 2019-07-08
[BEAM-7699] Propagate null check/comparison to deepEquals

honeyAndSw:
This is trivial improvement of `deepEqualsForMap`.
Looking at https://github.com/apache/beam/pull/6845 and https://github.com/apache/beam/pull/8368, equality check can be simpler as checking null values will be done in next recursion.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/values/Row.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/values/Row.java
@@ -422,14 +422,8 @@ public abstract class Row implements Serializable {
         V value = e.getValue();
         V otherValue = b.get(key);
 
-        if (value == null) {
-          if (otherValue != null || !b.containsKey(key)) {
-            return false;
-          }
-        } else {
-          if (!deepEquals(value, otherValue, valueType)) {
-            return false;
-          }
+        if (!deepEquals(value, otherValue, valueType)) {
+          return false;
         }
       }
 
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/values/Row.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/values/Row.java
@@ -422,7 +422,7 @@ public abstract class Row implements Serializable {
         V value = e.getValue();
         V otherValue = b.get(key);
 
-        if (!deepEquals(value, otherValue, valueType)) {
+        if (!b.containsKey(key) || !deepEquals(value, otherValue, valueType)) {
           return false;
         }
       }
--- a/sdks/java/core/src/test/java/org/apache/beam/sdk/values/RowTest.java
+++ b/sdks/java/core/src/test/java/org/apache/beam/sdk/values/RowTest.java
@@ -287,16 +287,32 @@ public class RowTest {
 
   @Test
   public void testCreateAndCompareMapWithDifferentKey() {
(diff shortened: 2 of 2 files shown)


sdks/java/core/src/main/java/org/apache/beam/sdk/values/Row.java
```diff
@@ -422,14 +422,8 @@ static int deepHashCode(Object a, Schema.FieldType fieldType) {
...
 
-        if (value == null) {
-          if (otherValue != null || !b.containsKey(key)) {
-            return false;
-          }
-        } else {
-          if (!deepEquals(value, otherValue, valueType)) {
-            return false;
-          }
+        if (!deepEquals(value, otherValue, valueType)) {
```
lukecwik: This won't be right if maps allow for null values. For example, if we have `a = { "s": null }` and `b = { }` then you need the `containsKey` check.

honeyAndSw: Right, I think `a = {"s": null}` and `b = {"t": null}` would be better example as they need to have the same size first

lukecwik: Would you like to add a test for this case so that future similar changes which would break this would be caught?

sdks/java/core/src/main/java/org/apache/beam/sdk/values/Row.java
```diff
@@ -422,14 +422,8 @@ static int deepHashCode(Object a, Schema.FieldType fieldType) {
...
 
-        if (value == null) {
-          if (otherValue != null || !b.containsKey(key)) {
-            return false;
-          }
-        } else {
-          if (!deepEquals(value, otherValue, valueType)) {
-            return false;
-          }
+        if (!b.containsKey(key) || !deepEquals(value, otherValue, valueType)) {
```
lukecwik: This is now more expensive because we now do two map lookups for all values while we used to only do two map lookups when the value was null.

sdks/java/core/src/main/java/org/apache/beam/sdk/values/Row.java
honeyAndSw: `RowTest#testCreateAndCompareNullableMapWithDifferentKey` covers this case, what do you think?

sdks/java/core/src/main/java/org/apache/beam/sdk/values/Row.java
honeyAndSw: Right, I'm thinking about reverting this change and leaving tests only

vagofrank:
Hi! what's the status for this change?

lukecwik:
@TheNeuralBit is this a change your interested in taking over and finishing?

TheNeuralBit:
Yeah I could do that. 
Based on the last comment it looks like the only action is to revert the actual change and make the PR only add tests?

lukecwik:
> Yeah I could do that.
> Based on the last comment it looks like the only action is to revert the actual change and make the PR only add tests?

To my knowledge yes.

TheNeuralBit:
Thanks stalebot, I'd still like to look at this.

PR #9011, opened 2019-07-08
Make LineSource claim the right position not the left position

boyuanzz:
Assuming there is a LineSource like:
'11\n' '222\n', '333\n'], when trying to claim the second element '222\n', the claim position in previous implementation is 3 (0 + len('11\n')), with my change, the claimed position will be 6 (0 + len('11\n') + len('222\n') - 1).
The reason to do this is to avoid the situation like, when reading to the end of range, the range_tracker doesn't claim till the end position.

### DIFF
--- a/sdks/python/apache_beam/io/filebasedsource_test.py
+++ b/sdks/python/apache_beam/io/filebasedsource_test.py
@@ -52,19 +52,22 @@ from apache_beam.transforms.display_test import DisplayDataItemMatcher
 
 class LineSource(FileBasedSource):
 
+  def _find_first_record(self, file, start):
+    file.seek(start - 1)
+    while file.read(1) != b'\n':
+      start -= 1
+      file.seek(start-1)
+    file.seek(start)
+    return start
+
   def read_records(self, file_name, range_tracker):
     f = self.open_file(file_name)
     try:
       start = range_tracker.start_position()
       if start > 0:
-        # Any line that starts after 'start' does not belong to the current
-        # bundle. Seeking to (start - 1) and skipping a line moves the current
-        # position to the starting position of the first line that belongs to
-        # the current bundle.
-        start -= 1
-        f.seek(start)
-        line = f.readline()
-        start += len(line)
+        # Any line that starts after 'start' should be included in current
+        # bundle.
+        start = self._find_first_record(f, start)
       line = f.readline()
       current = start + len(line) - 1
       while line:
@@ -696,7 +699,7 @@ class TestSingleFileSource(unittest.TestCase):
     source = SingleFileSource(fbs, file_name, 0, 10 * 6)
     range_tracker = source.get_range_tracker(0, 20)
     read_data = [value for value in source.read(range_tracker)]
-    self.assertCountEqual(expected_data[:4], read_data)
+    self.assertCountEqual(expected_data[:3], read_data)
 
   def test_read_range_at_end(self):
     fbs = LineSource('dummy_pattern', validate=False)
@@ -707,7 +710,7 @@ class TestSingleFileSource(unittest.TestCase):
     source = SingleFileSource(fbs, file_name, 0, 10 * 6)
     range_tracker = source.get_range_tracker(40, 60)
     read_data = [value for value in source.read(range_tracker)]
-    self.assertCountEqual(expected_data[-3:], read_data)
+    self.assertCountEqual(expected_data[-4:], read_data)
 
   def test_read_range_at_middle(self):
     fbs = LineSource('dummy_pattern', validate=False)
@@ -718,7 +721,7 @@ class TestSingleFileSource(unittest.TestCase):
     source = SingleFileSource(fbs, file_name, 0, 10 * 6)
     range_tracker = source.get_range_tracker(20, 40)
     read_data = [value for value in source.read(range_tracker)]
-    self.assertCountEqual(expected_data[4:7], read_data)
+    self.assertCountEqual(expected_data[3:6], read_data)
 
   def test_produces_splits_desiredsize_large_than_size(self):
     fbs = LineSource('dummy_pattern', validate=False)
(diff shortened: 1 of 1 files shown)


boyuanzz:
R: @chamikaramj , could you please help review this PR? The motivation to do this change is to make filebasedsource_test.LineSource be capable with updated OffsetRangeTracker.try_split in PR https://github.com/apache/beam/pull/8998 .

chamikaramj:
I'm bit confused by this change. Seems like https://github.com/apache/beam/pull/8998 tries to make the same try_split() call perform splitting twice (BTW left some comments there) ? Why do we need to change claiming to last byte position of a line to achieve that ?

I think this change makes the range tracker hard to understand which can result in bugs. Also will we have to do similar changes to other sources that use OffsetRangeTracker (for example, all file-based sources that are widely used by users) after https://github.com/apache/beam/pull/8998 ?

chamikaramj:
Also, seems like this breaks the specification of the range tracker interface: https://github.com/apache/beam/blob/master/sdks/python/apache_beam/io/iobase.py#L257

boyuanzz:
> I'm bit confused by this change. Seems like #8998 tries to make the same try_split() call perform splitting twice (BTW left some comments there) ? Why do we need to change claiming to last byte position of a line to achieve that ?
> 
Even with #8998, the split is only performed once either suggested split or adjusted split is valid.
Then reason why needs this PR is, imaging we have a 8 bytes record in range [0, 8) and range_tracker.try_claim(0) is called. At this point, _last_record_start = 0 and #8998 will compute 1 as split_offset, which is valid based on how OffsetRangeTracker implemented. But actually the range [0, 8) has been read based on how LineSource implemented.

Further more, I think it's a bug  of LineSource. Imaging when LineSource has finished reading last record, the range_tracker is not claimed to the end of range, which is not allowed in Java implementation with check_done: https://github.com/apache/beam/blob/b953645ed6db837d24284d7fe1fe091e7309f821/sdks/java/core/src/main/java/org/apache/beam/sdk/io/range/OffsetRangeTracker.java#L115

> I think this change makes the range tracker hard to understand which can result in bugs.
>
I think the worst case for #8998 is even the adjusted split is not valid, which is as worse as original split logic. 

> Also will we have to do similar changes to other sources that use OffsetRangeTracker (for example, all file-based sources that are widely used by users) after #8998 ?
>
I don't think so. The reason why I change LineSource is because we use LineSource to run UT. And if looking other block-like source like TextIO and AvroIO, they also claim the right position:
TextIO: https://github.com/apache/beam/blob/c01b0121b23ea19640ba3d8169f21d76de66440d/sdks/python/apache_beam/io/textio.py#L185
AvroIO: https://github.com/apache/beam/blob/0ac265b59a679804d105cc05c8a17f0c31402030/sdks/python/apache_beam/io/avroio.py#L429

boyuanzz:
> Also, seems like this breaks the specification of the range tracker interface: https://github.com/apache/beam/blob/master/sdks/python/apache_beam/io/iobase.py#L257

Could you please explain a little bit more about this specification? I don't really understand what it means...Sorry for the inconvenience and thanks in advance!

lukecwik:
The big difference between the Java implementation and this implementation is that we keep track of both the last successfully claimed position in addition to the last attempted position. The important difference is that the last attempted position is moved >= end of range after returning the last record and when the reader tries to get the next element.

boyuanzz:
> The big difference between the Java implementation and this implementation is that we keep track of both the last successfully claimed position in addition to the last attempted position. The important difference is that the last attempted position is moved >= end of range after returning the last record and when the reader tries to get the next element.

@chamikaramj Do you think it's better for python to have the same logic as java? We may also want to introduce check_done() logic to range_tracker. If so, I can create JIRA issue and keep tracking them. What do you think?

chamikaramj:
Yes, Java logic sounds better. So prob. better to go that route.

> Could you please explain a little bit more about this specification? I don't really understand what it means...Sorry for the inconvenience and thanks in advance!

" * All records returned by a source '[A, B)' must have starting positions in this range."

IIUC, with this change, if a record that starts at (A-x) ends at A. That record will be successfully claimed for position range  '[A, B)' which violates above statement.

boyuanzz:
> IIUC, with this change, if a record that starts at (A-x) ends at A. That record will be successfully claimed for position range '[A, B)' which violates above statement.

I feel like it depends on how we define the ```record_start```.  Imaging a record [A, A+x), in this PR I changed the ```record_start``` from A to A+x-1, which doesn't break that statement right?

lukecwik:
The reason why the record start is usually where the line starts and not where the line ends is that seeks are usually inefficient when going backwards and thus it is easier to seek to A and then read forwards to find the start of the first line instead of reading backwards to find the beginning of the line from A (with a special case for when A is 0).

> > IIUC, with this change, if a record that starts at (A-x) ends at A. That record will be successfully claimed for position range '[A, B)' which violates above statement.
> 
> I feel like it depends on how we define the `record_start`. Imaging a record [A, A+x), in this PR I changed the `record_start` from A to A+x-1, which doesn't break that statement right?

chamikaramj:
In addition to what Luke said, we should also try to keep our position based sources consistent. I think currently for most position based sources starting position (of record/block) is taken to be position that we claim through the range tracker.

boyuanzz:
Thanks, Luke and Cham. Sorry for the confusion of record_start_pos. I think the major point of this PR is when reading  a record [start_pos, start_pos + len(record)), the range_tracker should claim start_pos + len(record) - 1 rather than start_pos.

For the comment ```All records returned by a source '[A, B)' must have starting positions in this range.```, my understanding is, if I have claimed range [start_pos, start_pos + len(record)), the record should be in this range, which is also true within this PR. What do you think?

In terms of source consistency, at least avro and text io always claim next_record_start.

boyuanzz:
Discussed with Cham offline. Agreed on source should claim the start_pos of record. Added lastAttemptedOffset to OffsetRangeTracker should work for this case. Closed this PR and will have a separate PR to add lastAttemptedOffset. Thanks Luke and Cham!

PR #9015, opened 2019-07-09
[BEAM-7709] Re-use node for explicit flattens

lostluck:
Have all the links for a given Flatten PTransform use the same instance. Subsequent calls for these links will hit the cached version.

Simple problem with a simple fix. Adding a test would require adding a full test suite (this particular  filebarely has one at present), and I've not yet reconstructed a Git development environment since my previous laptop died, which makes that trickier from the web interface.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/go/pkg/beam/core/runtime/exec/translate.go
+++ b/sdks/go/pkg/beam/core/runtime/exec/translate.go
@@ -492,6 +492,11 @@ func (b *builder) makeLink(from string, id linkID) (Node, error) {
 
 	case graphx.URNFlatten:
 		u = &Flatten{UID: b.idgen.New(), N: len(transform.Inputs), Out: out[0]}
+	
+		// Use the same flatten instance for all the inputs links to this transform.
+		for i := 0; i < len(transform.Inputs); i++ {
+			b.links[linkID{id.to, i}] = u
+		}
 
 	case urnDataSink:
 		port, cid, err := unmarshalPort(payload)


lostluck:
R: @herohde

lostluck:
Run Go PostCommit

lostluck:
R: @youngoli @lukecwik 
(it looks like Henning is out this week, so no time to pull them in.)

lukecwik:
It is valid to have multiple copies of downstream DoFns but it adds a lot of complexity to do it right and is unlikely to be performant. (The complexity comes from things like user state/timers expectations that readers/writers have a consistent view per key).

lostluck:
Thanks!
Agreed. That would have been fine and wouldn't have triggered the bug I saw. In particular, subsequent links were creating new *flatten* nodes which meant there was no guarding each of them forwarding the startbundle request to a single downstream DoFn instance. The Go SDK treats each PTransform Node as a state machine to ensure the Bundle Lifecycle is followed.

youngoli:
Just so I understand this change, when handling a link between a transform with a Flatten URN and one of its inputs: Before it would just create a Flatten instance and set it as the link, now it also uses the same Flatten for every other input to the PTransform, so that when those links get put into this function later it exits early at line 320?

lostluck:
That's right! The "shortcut" on line 320 avoids repeating work in case something else is using that PCollection. Good ol' [memoization](https://en.wikipedia.org/wiki/Memoization).
Before it would, get a link from a PCollection, and just create a flatten *per* PCollection. As Luke mentions, this is totally fine if the runner "split" different upstream DoFns into different bundle executions, which naturally leads to multiple instances of the downstream DoFn. That's fine*. While it's unnecessary to have the flatten unit in that instance, it's not wrong.
However, if the two pcollections are in the *same* bundle execution, and there's only a single downstream consumer for the flatten, having multiple flatten nodes like the code was doing before is wrong, since the SDK side flatten has one job essentially: to ensure that the downstream StartBundles and FinishBundles are only ever called once per bundle. Naturally, this doesn't work with multiple instances of the same flatten node.
So the solution is to ensure all the inputs to a given Flatten Unit are mapped to a the same instance of that flatten unit.

*except as, Luke mentioned, for stateful processing, where we'll need to be sure if we're creating multiple instances of the same DoFn, that they're passed the same "state holder".

PR #9016, opened 2019-07-09
Reformat CamelCase function naming style to underscore style for synthe…

boyuanzz:
Happened to find there are some functions use CamelCase, which seems like not an python convention to me. Feel free to close this PR if I'm wrong : )
R: @pabloem 
CC: @laschmidt


pabloem:
Agree : )

PR #9019, opened 2019-07-09
[BEAM-7696] Prepare files to stage also in local master of spark runner.

yanlin-Lynn:
The result of PipelineResources.detectClassPathResourcesToStage may contain directory, which will cause exception, when running unit test of SparkRunner. See the attached picture file in [BEAM-7696](https://issues.apache.org/jira/browse/BEAM-7696) for detail of exception in such case. Even though the exception does not make SparkContext stop when running in local master, it's better not to include non-jars files in classpath.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/runners/spark/src/main/java/org/apache/beam/runners/spark/SparkPipelineOptions.java
+++ b/runners/spark/src/main/java/org/apache/beam/runners/spark/SparkPipelineOptions.java
@@ -17,7 +17,9 @@
  */
 package org.apache.beam.runners.spark;
 
+import java.io.File;
 import java.util.List;
+import java.util.stream.Collectors;
 import org.apache.beam.runners.core.construction.PipelineResources;
 import org.apache.beam.sdk.annotations.Experimental;
 import org.apache.beam.sdk.options.ApplicationNameOptions;
@@ -150,17 +152,24 @@ public interface SparkPipelineOptions
   void setCacheDisabled(boolean value);
 
   /**
-   * Local configurations work in the same JVM and have no problems with improperly formatted files
-   * on classpath (eg. directories with .class files or empty directories). Prepare files for
-   * staging only when using remote cluster (passing the master address explicitly).
+   * Classpath contains non jar files (eg. directories with .class files or empty directories) will
+   * cause exception in running log. Though the {@link org.apache.spark.SparkContext} can handle
+   * this when running in local master, it's better not to include non-jars files in classpath.
    */
-  static void prepareFilesToStageForRemoteClusterExecution(SparkPipelineOptions options) {
-    if (!options.getSparkMaster().matches("local\\[?\\d*\\]?")) {
-      options.setFilesToStage(
-          PipelineResources.prepareFilesForStaging(
-              options.getFilesToStage(),
-              MoreObjects.firstNonNull(
-                  options.getTempLocation(), System.getProperty("java.io.tmpdir"))));
-    }
+  static void prepareFilesToStage(SparkPipelineOptions options) {
+    List<String> filesToStage =
+        options.getFilesToStage().stream()
+            .map(File::new)
+            .filter(File::exists)
+            .map(
+                file -> {
+                  return file.getAbsolutePath();
+                })
+            .collect(Collectors.toList());
+    options.setFilesToStage(
+        PipelineResources.prepareFilesForStaging(
+            filesToStage,
+            MoreObjects.firstNonNull(
+                options.getTempLocation(), System.getProperty("java.io.tmpdir"))));
   }
 }
--- a/runners/spark/src/main/java/org/apache/beam/runners/spark/SparkPipelineRunner.java
+++ b/runners/spark/src/main/java/org/apache/beam/runners/spark/SparkPipelineRunner.java
@@ -18,7 +18,7 @@
 package org.apache.beam.runners.spark;
 
 import static org.apache.beam.runners.core.construction.PipelineResources.detectClassPathResourcesToStage;
-import static org.apache.beam.runners.spark.SparkPipelineOptions.prepareFilesToStageForRemoteClusterExecution;
+import static org.apache.beam.runners.spark.SparkPipelineOptions.prepareFilesToStage;
 
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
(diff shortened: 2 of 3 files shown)


iemejia:
Run Spark ValidatesRunner

iemejia:

iemejia:
R: @ibzib 
Kyle can you please help me with this review. Thx.

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/PipelineResources.java
```diff
@@ -61,7 +61,10 @@
     List<String> files = new ArrayList<>();
     for (URL url : ((URLClassLoader) classLoader).getURLs()) {
       try {
-        files.add(new File(url.toURI()).getAbsolutePath());
+        File file = new File(url.toURI());
+        if (file.exists()) {
```
ibzib: Since SparkContext will reject all directories [1], it is probably best to exclude directories too.

```suggestion
        if (file.exists() && !file.isDirectory()) {
```

[1] https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/SparkContext.scala#L1796-L1802

yanlin-Lynn:
> Hi Yanlin, thanks for the PR.
> 
> Since `PipelineResources.java` is in the `core-construction` package, it is shared by all Beam runners. While adding directories just causes Spark to throw an exception, directories might be acceptable or even desirable on different runners (I don't actually know--just being cautious here). So it would be better to limit this change to the `runners.spark` package (maybe by calling `PipelineResources.detectClassPathResourcesToStage` as-is, then removing directories in `SparkPipelineOptions.prepareFilesToStage`).

Yeah. I aggre with you. And in this patch, I did not exclude directories.

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/PipelineResources.java
yanlin-Lynn: Actually we cannot exclude directories. Excluding directories will cause flink runner related unit test case fail.

ibzib:
I suggest changing only the Spark runner code. That way, we can exclude nonexistent files and directories, without having to worry about the other runners. (While it may seem strange that other runners would ever use paths to resources that do not exist at the time of calling `detectClassPathResourcesToStage`, it's hard to prove that it couldn't happen, so I think it's better to just limit the scope of this change.)

yanlin-Lynn:
> I suggest changing only the Spark runner code. That way, we can exclude nonexistent files and existing directories, without having to worry about the other runners. (While it may seem strange that other runners would ever use paths to resources that do not exist at the time of calling `detectClassPathResourcesToStage`, it's hard to prove that it couldn't happen, so I think it's better to just limit the scope of this change.)

OK. This is a safer way to do this.

iemejia:
Run Spark ValidatesRunner

runners/spark/src/main/java/org/apache/beam/runners/spark/SparkPipelineOptions.java
```diff
@@ -150,17 +152,24 @@ public String create(PipelineOptions options) {
...
-          PipelineResources.prepareFilesForStaging(
-              options.getFilesToStage(),
-              MoreObjects.firstNonNull(
-                  options.getTempLocation(), System.getProperty("java.io.tmpdir"))));
-    }
+  static void prepareFilesToStage(SparkPipelineOptions options) {
+    List<String> filesToStage =
+        options.getFilesToStage().stream()
+            .map(File::new)
+            .filter(File::exists)
```
ibzib: What about directories?

yanlin-Lynn: The [PipelineResources.prepareFilesForStaging](https://github.com/apache/beam/blob/master/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/PipelineResources.java#L82) will package directories into jar file. That's why call this method before create a SparkContext.

ibzib: Okay, I missed that.

It looks like `prepareFilesForStaging` also throws `IllegalStateException` if a file doesn't exist. It might be better to throw the exception to alert the user, rather than silently filtering out a missing dependency and have an error show up later.

ibzib: Oh, I guess there are always nonexistent files. Weird. Well, this is fine then.

iemejia:
Thanks @yanlin-Lynn !

lukecwik:
I suggest that we re-use the code in Dataflow and to automatically convert directory classpath entries into jars:
https://github.com/apache/beam/blob/3e97543d53cf30b7ac072e225d358e8436784220/runners/google-cloud-dataflow-java/src/main/java/org/apache/beam/runners/dataflow/util/PackageUtil.java#L256

It will likely require some refactoring to make it available to more runners.

yanlin-Lynn:
> I suggest that we re-use the code in Dataflow and to automatically convert directory classpath entries into jars:
> https://github.com/apache/beam/blob/3e97543d53cf30b7ac072e225d358e8436784220/runners/google-cloud-dataflow-java/src/main/java/org/apache/beam/runners/dataflow/util/PackageUtil.java#L256
> 
> It will likely require some refactoring to make it available to more runners.

Currently, there exist `detectClassPathResourcesToStage` and `prepareFilesForStaging` in [PipelineResources](https://github.com/apache/beam/blob/master/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/PipelineResources.java#L51). Do you mean to do the packaging of directories inside `detectClassPathResourcesToStage`, so no need to call `prepareFilesForStaging` manually?

lukecwik:
> > I suggest that we re-use the code in Dataflow and to automatically convert directory classpath entries into jars:
> > https://github.com/apache/beam/blob/3e97543d53cf30b7ac072e225d358e8436784220/runners/google-cloud-dataflow-java/src/main/java/org/apache/beam/runners/dataflow/util/PackageUtil.java#L256
> > 
> > It will likely require some refactoring to make it available to more runners.
> 
> Currently, there exist `detectClassPathResourcesToStage` and `prepareFilesForStaging` in [PipelineResources](https://github.com/apache/beam/blob/master/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/PipelineResources.java#L51). Do you mean to do the packaging of directories inside `detectClassPathResourcesToStage`, so no need to call `prepareFilesForStaging` manually?

I thought the issue was that the SparkRunner couldn't handle directories but I misread the conversation above and its related to non-existent files so my comment doesn't make any sense and no change is needed/wanted.

PR #9023, opened 2019-07-09
RestrictionTracker.current_watermark returns None

boyuanzz:
RestrictionTracker.current_watermark returns None by default and treats it as MIN_TIMESTAMP.

### DIFF
--- a/model/fn-execution/src/main/proto/beam_fn_api.proto
+++ b/model/fn-execution/src/main/proto/beam_fn_api.proto
@@ -194,7 +194,8 @@ message BundleApplication {
   // are produced by this PTransform into each of its output PCollections
   // when invoked with this application.
   //
-  // If there is no watermark reported, use MIN_TIMESTAMP by default.
+  // If there is no watermark reported from RestrictionTracker, the sdk should
+  // report MIN_TIMESTAMP by default.
   map<string, google.protobuf.Timestamp> output_watermarks = 4;
 
   // (Required) Whether this application potentially produces an unbounded
--- a/model/fn-execution/src/main/proto/beam_fn_api.proto
+++ b/model/fn-execution/src/main/proto/beam_fn_api.proto
@@ -194,8 +194,8 @@ message BundleApplication {
   // are produced by this PTransform into each of its output PCollections
   // when invoked with this application.
   //
-  // If there is no watermark reported from RestrictionTracker, the sdk should
-  // report MIN_TIMESTAMP by default.
+  // If there is no watermark reported from RestrictionTracker, the runner will
+  // use MIN_TIMESTAMP by default.
   map<string, google.protobuf.Timestamp> output_watermarks = 4;
 
   // (Required) Whether this application potentially produces an unbounded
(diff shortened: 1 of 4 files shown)


sdks/python/apache_beam/runners/common.py
```diff
@@ -704,13 +704,17 @@ def try_split(self, fraction):
         restriction_provider = self.signature.get_restriction_provider()
         primary_size = restriction_provider.restriction_size(element, primary)
         residual_size = restriction_provider.restriction_size(element, residual)
+        # If no current watermark reported, use MIN_TIMESTAMP by default.
+        current_watermark = (restriction_tracker.current_watermark()
```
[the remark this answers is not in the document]
boyuanzz: Provide a temporary workaround in commit https://github.com/apache/beam/pull/9023/commits/e46ea6271f97e02afdfdd9c4a172771e51e229b9

model/fn-execution/src/main/proto/beam_fn_api.proto
```diff
@@ -193,6 +193,8 @@ message BundleApplication {
   // value represents a lower bound on the timestamps of elements that
   // are produced by this PTransform into each of its output PCollections
   // when invoked with this application.
+  //
+  // If there is no watermark reported, use MIN_TIMESTAMP by default.
```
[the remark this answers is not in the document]
boyuanzz: In my implementation, the sdk picks ```MIN_TIMESTAMPS``` to report to runner if ```RestrictionTracker``` returns ```None```. I rephrased the comment a little bit in commit: https://github.com/apache/beam/pull/9023/commits/404fb66c1cb700cb19a5310e85cee917550a2e71.

boyuanzz:
Run Python PreCommit

boyuanzz:
@lukecwik Could you please take another look?

boyuanzz:
R: @robertwb

model/fn-execution/src/main/proto/beam_fn_api.proto
```diff
@@ -193,6 +193,9 @@ message BundleApplication {
   // value represents a lower bound on the timestamps of elements that
   // are produced by this PTransform into each of its output PCollections
   // when invoked with this application.
+  //
+  // If there is no watermark reported from RestrictionTracker, the sdk should
+  // report MIN_TIMESTAMP by default.
```
robertwb: Are we sure we want to make this required? (In particular, when a watermark is provided, it may make sense to check that it is >= the input watermark for non-late data, which would be violated if we provide MIN_TIMESTAMP as a default value.

boyuanzz: It depends on how runner treats MIN_TIMESTAMP. On the other hand, it's also possible that user implementation returns MIN_TIMESTAMP right?

robertwb: Correct, if we fill this in we can't distinguish between the user setting MIN_TIMESTAMP and the framework defaulting to it at any point downstream. Better to let the "unset" propagate all the way tot the runner.

boyuanzz: Changed in commit https://github.com/apache/beam/pull/9023/commits/127ab262ae26b0906bd009875a01b6ea232b06ab

robertwb:
The Java and Go failures are irrelevant, all that changed there was the comments in the proto file.

PR #7005, opened 2018-11-12
[BEAM-6037] Make Spark runner pipeline translation based on URNs

iemejia:
Baby steps towards portable translation, at least this brings an homogeneous start.

R: @angoenka 
CC: @lukecwik

PR #9026, opened 2019-07-09
[BEAM-7653] Add PTransformTranslator for Combine.GroupedValues

iemejia:
This is a follow up to #8995 but for the case of a runner that wants to have a specific translation of `Combine.GroupedValues`. Notice that this is the case of the Spark runner and this PR is to address a regression on it (since the runner moved to URN based translation instead of class based it is not calling the specialized translation).
R: @lukecwik

### DIFF
--- a/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/CombineTranslation.java
+++ b/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/CombineTranslation.java
@@ -18,6 +18,7 @@
 package org.apache.beam.runners.core.construction;
 
 import static org.apache.beam.runners.core.construction.PTransformTranslation.COMBINE_GLOBALLY_TRANSFORM_URN;
+import static org.apache.beam.runners.core.construction.PTransformTranslation.COMBINE_GROUPED_VALUES_TRANSFORM_URN;
 import static org.apache.beam.runners.core.construction.PTransformTranslation.COMBINE_PER_KEY_TRANSFORM_URN;
 
 import com.google.auto.service.AutoService;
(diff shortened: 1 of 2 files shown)


iemejia:
Run Spark ValidatesRunner

iemejia:
Notice that the extra fix in the test of the Spark runner shows that the validation is now correct because it calls a spark specific map instead of Beam's composite translation.

lukecwik:
Is there a reason why the spark runner is intercepting and overriding combine grouped values and not combine globally/perKey and providing appropriate expansions for them?

lukecwik:
@youngoli FYI

iemejia:
Spark Runner actually translates the three cases: `Combine.PerKey`, `Combine.Globally and `Combine.GroupedValues`. But since there were not translators for the URN based case the translation was defaulting to the one of `Combine.PerKey`. This PR intends to make the missitn `Combine.GroupedValues` to work as before.
For ref #7005 was the PR where we switched from class based translation to URN based.

lukecwik:
I understand that #7005 was where the translation didn't handle these other cases and I'm just trying to learn how Spark is optimizing Combine.GroupedValues.

iemejia:
It is relatively simple [it just 'combines' (reduces) the value of the iterable part and immediately produce the WindowedValue of this with a direct map](https://github.com/apache/beam/blob/14d8c3d64c0eb42744e02b3f4fe9890f59c4e2f1/runners/spark/src/main/java/org/apache/beam/runners/spark/translation/TransformTranslator.java#L188-L196), it is probably just a bit more lightweight than the normal translation because of reduced use of Beam constructs.

lukecwik:
I see, this is for translation for when we execute pipelines without using the portable execution and where we can deserialize the users combinefn.

iemejia:
Yes, a 'simpler' case. Thanks Luke.

PR #9031, opened 2019-07-09
[BEAM-7715] Mark user facing APIs related to External transforms as @Experimental

iemejia:
R: @chamikaramj 
CC: @mxm

### DIFF
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/expansion/ExternalTransformRegistrar.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/expansion/ExternalTransformRegistrar.java
@@ -18,12 +18,14 @@
 package org.apache.beam.sdk.expansion;
 
 import java.util.Map;
+import org.apache.beam.sdk.annotations.Experimental;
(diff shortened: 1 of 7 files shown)


chamikaramj:
LGTM. Thanks.

iemejia:
No, this is quite hard to get agreed on because it can be a mix of criteria from the maintainers + probably API stability (has the API changed or being improved in the last n releases). I think it is worth to discuss and create some guidelines for this at the mailing list but probably impossible to find a solution that fits everyone about removing Experimental annotations.

In any case for the case of this PR probably we all agree that marking it is as `@Experimental` until it gets more mature so we can refine the API if needed.

mxm:
Retest this please

iemejia:
I could not annotate properly the python transforms because there seems to be an issue on annotating python classes. For the moment I just added the Experimental message in the documentation for the python transforms (as I noticed that other python classes do) and filled [BEAM-7757](https://issues.apache.org/jira/browse/BEAM-7757) so this gets reviewed.

PR #9032, opened 2019-07-10
[BEAM-7386] Bi-Temporal Join

rezarokni:
The extensions / time series is intended to be a incubation area for utilities that assist with working with time series data. 

This sample will take two streams left stream and right stream. It will match the left stream to the nearest right stream based on their timestamp. Nearest left stream is either <= to the right stream timestamp.

------------------------

--- | --- | --- | --- | ---


rezarokni:
@reuvenlax , @kennknowles

sdks/java/extensions/timeseries/src/test/java/org/apache/beam/sdk/extensions/timeseries/joins/BiTemporalCoderTests.java
```diff
@@ -0,0 +1,148 @@
...
+      Arrays.asList(
+          BiTemporalJoinResult.of()
+              .setLeftData(KV.of("key", -1), Instant.ofEpochMilli(Long.MAX_VALUE))
+              .setRightData(KV.of("key", 0), Instant.ofEpochMilli(Long.MAX_VALUE)));
+
+  /**
+   * Generated data to check that the wire format has not changed. To regenerate, see {@link
+   * org.apache.beam.sdk.coders.PrintBase64Encodings}.
+   */
+  private static final List<String> TEST_ENCODINGS =
```
rezarokni: TODO: Missing Test.

rezarokni:
@ruwang

kennknowles:
I believe you are looking for @amaliujia

amaliujia:
Thank Kenn! Will take a look soon.

sdks/java/extensions/timeseries/src/main/java/org/apache/beam/sdk/extensions/timeseries/joins/BiStreamJoinResultCoder.java
```diff
@@ -0,0 +1,121 @@
...
+      Coder<K> keyCoder, Coder<V1> leftCoder, Coder<V2> rightCoder) {
+
+    return new BiStreamJoinResultCoder<>(keyCoder, leftCoder, rightCoder);
+  }
+
+  @Override
+  public void encode(BiTemporalJoinResult<K, V1, V2> value, OutputStream outStream)
+      throws IOException {
+
+    if (value == null) {
```
amaliujia: key, left and right coders are already `NullableCoder` which tolerates `NULL`?

sdks/java/extensions/timeseries/src/main/java/org/apache/beam/sdk/extensions/timeseries/joins/BiStreamJoinResultCoder.java
```diff
@@ -0,0 +1,121 @@
...
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.Arrays;
+import java.util.List;
+import org.apache.beam.sdk.coders.*;
+import org.apache.beam.sdk.values.TimestampedValue;
+import org.apache.beam.sdk.values.TypeDescriptor;
+import org.apache.beam.sdk.values.TypeParameter;
+
+public class BiStreamJoinResultCoder<K, V1, V2>
```
amaliujia: `BiStreamJoinResultCoder` and `BiTemporalJoinResultCoder` are duplicates?

sdks/java/extensions/timeseries/src/test/java/org/apache/beam/sdk/extensions/timeseries/joins/BiTemporalCacheTestIT.java
```diff
@@ -0,0 +1,150 @@
...
+import org.apache.beam.sdk.values.TupleTag;
+import org.joda.time.Duration;
+import org.joda.time.Instant;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class BiTemporalCacheTestIT implements Serializable {
```
amaliujia: How does this IT run? Will `./gradlew test` trigger it?

amaliujia:
Retest this please

amaliujia:
I am not seeing Java precommit tests are triggered. I am less familiar on how does that is enabled. Guessing have to set at here: https://github.com/apache/beam/blob/master/build.gradle#L132

kennknowles:
Precommits are back

sdks/java/extensions/timeseries/build.gradle
```diff
@@ -0,0 +1,49 @@
...
+
+dependencies {
+  compile library.java.guava
+  compile project(path: ":sdks:java:core", configuration: "shadow")
+  // Needed to run the Example.
+  compile project(path: ":runners:direct-java")
+  testCompile library.java.hamcrest_core
+  testCompile library.java.hamcrest_library
+  testCompile library.java.junit
+  //testRuntimeOnly project(path: ":runners:direct-java")
```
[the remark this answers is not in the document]
rezarokni: Done

sdks/java/extensions/timeseries/src/main/java/org/apache/beam/sdk/extensions/timeseries/joins/BiStreamJoinResultCoder.java
rezarokni: Fixed.

sdks/java/extensions/timeseries/src/test/java/org/apache/beam/sdk/extensions/timeseries/joins/BiTemporalCacheTestIT.java
rezarokni: @lukecwik What is the best way to setup IT tests, does the framework just pick these up based on RequiresRunner?

sdks/java/extensions/timeseries/src/main/java/org/apache/beam/sdk/extensions/timeseries/joins/BiStreamJoinResultCoder.java
rezarokni: BiStreamJoinResultCoder file was left over artifact from rename of class. Thanx for catching.

sdks/java/extensions/timeseries/src/test/java/org/apache/beam/sdk/extensions/timeseries/joins/BiTemporalCacheTestIT.java
rezarokni: I think I found  the way this is done. Changed the gradle.build to accommodate,. it.

rezarokni:
@ruwang Any views on the main stream class :-)

rezarokni:
@amaliujia Would you be able to have a look at the main stream class this week?

amaliujia:
@rezarokni 

Was seeing not all checks succeeded so thought you worked on making those checks pass. But just clicked into logs and looked like they are not relevant, so will take a closer look on main stream class soon.

rezarokni:
> @rezarokni
> 
> Was seeing not all checks succeeded so thought you worked on making those checks pass. But just clicked into logs and looked like they are not relevant, so will take a closer look on main stream class soon.

There are some check style changes that need to be made, if OK I will fixup with your next comments.

amaliujia:
@rezarokni

You have to find a committer as well on this PR to review/merge it.

rezarokni:
@amaliujia  thanx!
@kennknowles would you have time for a review please?

sdks/java/extensions/timeseries/src/main/java/org/apache/beam/sdk/extensions/timeseries/joins/BiTemporalStreams.java
```diff
@@ -0,0 +1,462 @@
...
+  }
+
+  @Experimental
+  public static class BiTemporalJoin<K, V1, V2>
+      extends PTransform<KeyedPCollectionTuple<K>, PCollection<BiTemporalJoinResult<K, V1, V2>>> {
+
+    private static final Logger LOG = LoggerFactory.getLogger(BiTemporalStreams.class);
+
+    // Sets the limit at which point the processed left stream values are garbage collected
+    static int GC_LIMIT = 1000;
```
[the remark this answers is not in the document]
rezarokni: OOO at the moment, but the current experimental implementation does not deal with late data ( I think I had a note in the java doc). When I am back I will add a check to throw illegal state if late data is not set to ZERO.  Until we have retractions the best way that we can deal with late data is to deadletter it.

The GC_LIMIT ( which should have a better name, will fix that when I am back) is actually to reduce the amount of processing needed in sorting the left hand side. As the whole list needs to be sorted across each finish bundle, but once an element is processed it is no longer needed. Once we have Sorted Map State most of the code for this class can be deleted, as there will be no need for the cache that we have to use at the moment.  :-)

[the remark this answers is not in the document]
rezarokni: With regards to Je-ik first comment, 
@amaliujia @kennknowles had a quick glance at the StatefullDoFnRunner, it uses a Timer to do GC. I dont think all runners use Timers for this operation, but as je-ik mentioned some will do ... If I recall there is some issues with how multiple timers will interact with each other, when a Timer resets itself. Should this be standardized, as I think right now this could potentially mean things will work on some runners and not on others?

sdks/java/extensions/timeseries/src/main/java/org/apache/beam/sdk/extensions/timeseries/joins/BiTemporalStreams.java
```diff
@@ -0,0 +1,462 @@
...
+        leftCache = null;
+        rightCache = null;
+      }
+
+      /**
+       * The processTimer checks if there are left stream values beyound the last processed date, if
+       * there is then it will check for a match and output as well as increment the processed date.
+       */
+      @OnTimer("processTimer")
+      public void OnTimer(
```
rezarokni: Variable name clean up missed removal of Quote and reversed left / right side .

sdks/java/extensions/timeseries/src/main/java/org/apache/beam/sdk/extensions/timeseries/joins/BiTemporalStreams.java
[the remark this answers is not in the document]
je-ik: @rezarokni I have managed to (hopefully finally) fix the issue with timers. Related PR is #9190, I cherry picked the two commits into PR rezarokni#2 and the failing tests are green on my local machine, with the exception of `BiTemporalCacheTest.cacheTest`, which is still failing (consistently). Don't know what is the issue with that, but I hope it is no longer related to the timer ordering. Could you validate that?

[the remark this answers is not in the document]
rezarokni: Thanx for looking into this Jan, just back from OOO will look at this over the next week.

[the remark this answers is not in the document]
rezarokni: Looks like one of the keys is no longer getting processed. Thats the symptom, digging into the cause.

sdks/java/extensions/timeseries/src/main/java/org/apache/beam/sdk/extensions/timeseries/joins/BiTemporalStreams.java
```diff
@@ -0,0 +1,464 @@
...
+      public <V> List<TimestampedValue<V>> createAndLoadSortedList(
+          BoundedWindow window, BagState<TimestampedValue<V>> stream) {
+
+        List<TimestampedValue<V>> sortedBundleList = new ArrayList<>();
+        final List<TimestampedValue<V>> data = sortedBundleList;
+        stream.read().forEach(t -> data.add(t));
+        data.sort(Comparator.comparing(t -> t.getTimestamp()));
+        sortedBundleList = data;
+
+        LOG.debug("Catch miss for maxTime {} ", window.maxTimestamp());
```
je-ik:
```suggestion
        LOG.debug("Cache miss for maxTime {} ", window.maxTimestamp());
```

sdks/java/extensions/timeseries/src/main/java/org/apache/beam/sdk/extensions/timeseries/joins/BiTemporalStreams.java
```diff
@@ -0,0 +1,464 @@
...
+@Experimental
+public class BiTemporalStreams {
+
+  public static <K, V1, V2> BiTemporalJoin<K, V1, V2> join(
+      TupleTag<V1> leftTag, TupleTag<V2> rightTag, Duration window) {
+    return new BiTemporalJoin<>(leftTag, rightTag, window);
+  }
+
+  @Experimental
+  public static class BiTemporalJoin<K, V1, V2>
```
je-ik: Should this be called `BiTemporalLeftJoin`? It seems not to work equally in both directions.

sdks/java/extensions/timeseries/src/main/java/org/apache/beam/sdk/extensions/timeseries/joins/BiTemporalStreams.java
```diff
@@ -0,0 +1,464 @@
...
+
+    private BiTemporalJoin(TupleTag<V1> leftTag, TupleTag<V2> rightTag, Duration window) {
+      this.leftTag = leftTag;
+      this.rightTag = rightTag;
+      this.window = window;
+    }
+
+    public static <K, V1, V2> BiTemporalJoin create(
+        TupleTag<V1> leftTag, TupleTag<V2> rightTag, Duration window) {
+      return new BiTemporalJoin<K, V1, V2>(leftTag, rightTag, window);
```
je-ik:
```suggestion
      return new BiTemporalJoin<>(leftTag, rightTag, window);
```

sdks/java/extensions/timeseries/src/main/java/org/apache/beam/sdk/extensions/timeseries/joins/BiTemporalStreams.java
```diff
@@ -0,0 +1,464 @@
...
+    // Sets the limit at which point the processed left stream values are garbage collected
+    static int LEFT_STREAM_GC_LIMIT = 1000;
+
+    Coder<V1> leftCoder;
+    Coder<V2> rightCoder;
+
+    TupleTag<V1> leftTag;
+    TupleTag<V2> rightTag;
+
+    Duration window;
```
je-ik: Hm, I'm wondering, does this have to work in FixedWindows? I think there should be no blocker for this to work on GlobalWindows as well. Even for any other windows, provided the windowFn is the same on both sides, right?

sdks/java/extensions/timeseries/src/main/java/org/apache/beam/sdk/extensions/timeseries/joins/BiTemporalStreams.java
```diff
@@ -0,0 +1,464 @@
...
+        if (input.getValue().getRightData() != null) {
+          TimestampedValue<V2> rightData = input.getValue().getRightData();
+          rightStream.add(rightData);
+        }
+
+        if (input.getValue().getLeftData() != null) {
+          TimestampedValue<V1> rightStreamValue = input.getValue().getLeftData();
+          leftStream.add(rightStreamValue);
+
+          long processTimerTimestamp = Optional.ofNullable(timerTimestamp.read()).orElse(0L);
```
je-ik: Timestamp 0 seems no to be ideal default value, because it can (in theory) be valid timestamp that was actually set by an element. Maybe `BoundedWindow.TIMESTAMP_MIN_VALUE.getMillis()` would be better. And maybe instead of `long` the state could hold the `Instant`? `InstantCoder` should be efficient enough.

sdks/java/extensions/timeseries/src/main/java/org/apache/beam/sdk/extensions/timeseries/joins/BiTemporalStreams.java
```diff
@@ -0,0 +1,464 @@
...
+
+        Iterator<TimestampedValue<V1>> leftStreamValueIterator = leftStreamValueList.iterator();
+
+        long lastProcessedLeftStreamTimestamp =
+            Optional.ofNullable(lastProcessedLeftStreamTimestampState.read()).orElse(0L);
+
+        // Index position of leftStreamValue queue
+        int prevIdx = 0;
+
+        // Procssed Trade count
```
je-ik:
```suggestion
        // Count of processed left elements
```

sdks/java/extensions/timeseries/src/main/java/org/apache/beam/sdk/extensions/timeseries/joins/BiTemporalStreams.java
```diff
@@ -0,0 +1,464 @@
...
+              BiTemporalJoinResult<K, V1, V2> m =
+                  createMatch(dataKey.read(), subList, leftStreamValue, idx);
+
+              prevIdx = prevIdx + idx;
+
+              output.outputWithTimestamp(m, leftStreamValue.getTimestamp());
+              lastProcessedLeftStreamTimestamp = leftStreamValue.getTimestamp().getMillis();
+            }
+          }
+          // Increase the processed Trade Counter to check for GC LIMIT
```
je-ik:
```suggestion
          // Increase the processed left element counter to check for GC LIMIT
```

sdks/java/extensions/timeseries/src/main/java/org/apache/beam/sdk/extensions/timeseries/joins/BiTemporalStreams.java
```diff
@@ -0,0 +1,464 @@
...
+          ++processedLeftStreamCount;
+        }
+
+        // Check if there is more work left in the leftStreamValue queue
+        // Create timer to fire if there is
+        // If the number of old left stream values in the queue is @ LEFT_STREAM_GC_LIMIT then do GC
+        // If no work left delete the Trade Queue
```
je-ik:
```suggestion
        // If no work left, delete the left queue
```

sdks/java/extensions/timeseries/src/main/java/org/apache/beam/sdk/extensions/timeseries/joins/BiTemporalStreams.java
rezarokni: The potential for OOM is not just in batch it can also happen in stream, dependent on the data size. For the datasets that we first explored this use case with however largemem workers was able to deal with the data volumes even with very large windows. 

One solution to reduce the chance of a OOM, would be to wrap an outer sharding layer to this transform. This would shard up the right hand data into smaller time segments. And the left hand side would be sent to all-possible time shards on the right hand side. With multiple matches evaluated in downstream transform. 

All of that could also be made transparent to the user of course, but I had not yet gone down that route in this @experimental class. It would be good to see if folks find it useful and to get back data on expected data sizes before adding in the complexity. Also I understand folks are looking into a potential sorted map state, which would allow for the whole of this cache code to be deleted :-)

je-ik: It is true, that it can OOM on stream, but the chances are pretty m uch lower, than on stream. Essentially, watermark would have to stop being forwarded for long enough. But - even if this happens, checkpointing and recovery mechanisms of streaming engines would eventually make the watermark move and the code would have a chance to complete. Once this fails on batch, it will fail forever. And there is practically no reason for that, because on batch the input can be sorted, and watermarks can be updated gradually (instead of having two states - min and max). That would be way more efficient.

rezarokni:
Dataflow Runner and Flink Runner all pass the tests, as does the current DirectRunner . Will add some more tests to this PR to confirm and reopen so we can start the library extension.

PR #8035, opened 2019-03-12
[BEAM-6810] Disable CalcRemoveRule to fix trivial projections

kanterov:
Trivial programs project precisely their input fields, without dropping
or re-ordering them. It turns out that CalcRemoveRule eliminates them,
even if there is a field aliasing. It results in an unexpected
resulting schema of SQL query.

It isn't clear whether it's an issue in Calcite or Beam rules, as a
workaround we disable CalcRemoveRule.

------------------------

Thank you for your contribution! Follow this checklist to help us incorporate your contribution quickly and easily:

 - [x] [**Choose reviewer(s)**](https://beam.apache.org/contribute/#make-your-change) and mention them in a comment (`R: @username`).
 - [x] Format the pull request title like `[BEAM-XXX] Fixes bug in ApproximateQuantiles`, where you replace `BEAM-XXX` with the appropriate JIRA issue, if applicable. This will automatically link the pull request to the issue.
 - [x] If this contribution is large, please file an Apache [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

Lang | SDK | Apex | Dataflow | Flink | Gearpump | Samza | Spark
--- | --- | --- | --- | --- | --- | --- | ---
Go | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go/lastCompletedBuild/) | --- | --- | --- | --- | --- | ---
[...]

PR #6139, opened 2018-08-03
[BEAM-5072][SQL]Remove RelCollationTraitDef configuration in BeamQueryPlanner

amaliujia:
Right now we configure `RelCollationTraitDef` in `BeamQueryPlanner` but seems like we don't use/implement it. However this configuration causes `SortRemoveRule` to apply sort to SortRel's input and then drop the SorRel (which does not really sort the SortRel's input). 

Remove `RelCollationTraitDef`and disable `SortRemoveRule` for now.

If don't remove CollationTrait, one example is:
```
INFO: SQLPlan>
BeamIOSinkRel(table=[[beam, ORDER_ID_TABLE]], operation=[INSERT], flattened=[true])
  LogicalSort(sort0=[$0], dir0=[ASC])
    LogicalProject(order_id=[$0])
      BeamIOSourceRel(table=[[beam, ORDER_DETAILS]])

Aug 03, 2018 1:07:24 PM org.apache.beam.sdk.extensions.sql.impl.BeamQueryPlanner convertToBeamRel
INFO: BEAMPlan>
BeamIOSinkRel(table=[[beam, ORDER_ID_TABLE]], operation=[INSERT], flattened=[true])
  BeamCalcRel(expr#0..3=[{inputs}], order_id=[$t0])
    BeamIOSourceRel(table=[[beam, ORDER_DETAILS]])
```

------------------------

Follow this checklist to help us incorporate your contribution quickly and easily:

 - [ ] Format the pull request title like `[BEAM-XXX] Fixes bug in ApproximateQuantiles`, where you replace `BEAM-XXX` with the appropriate JIRA issue, if applicable. This will automatically link the pull request to the issue.
 - [ ] If this contribution is large, please file an Apache [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

It will help us expedite review of your Pull Request if you tag someone (e.g. `@username`) to look at it.

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

Lang | SDK | Apex | Dataflow | Flink | Gearpump | Samza | Spark
--- | --- | --- | --- | --- | --- | --- | ---
Go | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_GradleBuild/lastCompletedBuild/) | --- | --- | --- | --- | --- | ---
[...]

PR #9040, opened 2019-07-10
[BEAM-7545] Reordering Beam Joins

riazela:
This adds rules to reorder joins. Calcite's rules may produce cross joins, another version of them are implemented to prevent this.

R: @akedin 

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/planner/BeamRuleSets.java
+++ b/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/planner/BeamRuleSets.java
@@ -25,6 +25,8 @@ import org.apache.beam.sdk.extensions.sql.impl.rule.BeamBasicAggregationRule;
 import org.apache.beam.sdk.extensions.sql.impl.rule.BeamCalcRule;
 import org.apache.beam.sdk.extensions.sql.impl.rule.BeamEnumerableConverterRule;
 import org.apache.beam.sdk.extensions.sql.impl.rule.BeamIntersectRule;
+import org.apache.beam.sdk.extensions.sql.impl.rule.BeamJoinAssociateRule;
+import org.apache.beam.sdk.extensions.sql.impl.rule.BeamJoinPushThroughJoinRule;
 import org.apache.beam.sdk.extensions.sql.impl.rule.BeamJoinRule;
 import org.apache.beam.sdk.extensions.sql.impl.rule.BeamMinusRule;
 import org.apache.beam.sdk.extensions.sql.impl.rule.BeamSortRule;
@@ -46,6 +48,7 @@ import org.apache.calcite.rel.rules.FilterJoinRule;
 import org.apache.calcite.rel.rules.FilterProjectTransposeRule;
 import org.apache.calcite.rel.rules.FilterSetOpTransposeRule;
 import org.apache.calcite.rel.rules.FilterToCalcRule;
+import org.apache.calcite.rel.rules.JoinCommuteRule;
 import org.apache.calcite.rel.rules.JoinPushExpressionsRule;
 import org.apache.calcite.rel.rules.ProjectCalcMergeRule;
 import org.apache.calcite.rel.rules.ProjectFilterTransposeRule;
@@ -103,6 +106,10 @@ public class BeamRuleSets {
 
           // join rules
           JoinPushExpressionsRule.INSTANCE,
+          JoinCommuteRule.INSTANCE,
+          BeamJoinAssociateRule.INSTANCE,
+          BeamJoinPushThroughJoinRule.RIGHT,
+          BeamJoinPushThroughJoinRule.LEFT,
 
           // remove union with only a single child
           UnionEliminatorRule.INSTANCE,
--- a/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rule/BeamJoinAssociateRule.java
+++ b/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rule/BeamJoinAssociateRule.java
@@ -1,0 +1,52 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.beam.sdk.extensions.sql.impl.rule;
+
+import org.apache.beam.sdk.extensions.sql.impl.rel.BeamJoinRel;
+import org.apache.calcite.plan.RelOptRuleCall;
+import org.apache.calcite.rel.core.Join;
+import org.apache.calcite.rel.core.RelFactories;
+import org.apache.calcite.rel.rules.JoinAssociateRule;
+import org.apache.calcite.tools.RelBuilderFactory;
+
+/**
+ * This is very similar to {@link org.apache.calcite.rel.rules.JoinAssociateRule}. It only checks if
+ * the resulting condition is supported before transforming.
+ */
+public class BeamJoinAssociateRule extends JoinAssociateRule {
+
+  public static final org.apache.beam.sdk.extensions.sql.impl.rule.BeamJoinAssociateRule INSTANCE =
+      new org.apache.beam.sdk.extensions.sql.impl.rule.BeamJoinAssociateRule(
+          RelFactories.LOGICAL_BUILDER);
+
+  private BeamJoinAssociateRule(RelBuilderFactory relBuilderFactory) {
+    super(relBuilderFactory);
+  }
+
+  @Override
+  public void onMatch(final RelOptRuleCall call) {
+    super.onMatch(
+        new JoinRelOptRuleCall(
+            call,
+            rel -> {
+              Join topJoin = (Join) rel;
+              Join bottomJoin = (Join) ((Join) rel).getRight();
+              return BeamJoinRel.isJoinLegal(topJoin) && BeamJoinRel.isJoinLegal(bottomJoin);
+            }));
+  }
+}
--- a/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rule/BeamJoinPushThroughJoinRule.java
+++ b/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rule/BeamJoinPushThroughJoinRule.java
@@ -1,0 +1,70 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.beam.sdk.extensions.sql.impl.rule;
+
+import org.apache.beam.sdk.extensions.sql.impl.rel.BeamJoinRel;
+import org.apache.calcite.plan.RelOptRule;
+import org.apache.calcite.plan.RelOptRuleCall;
+import org.apache.calcite.rel.core.Join;
+import org.apache.calcite.rel.core.RelFactories;
+import org.apache.calcite.rel.logical.LogicalJoin;
+import org.apache.calcite.rel.rules.JoinPushThroughJoinRule;
+import org.apache.calcite.tools.RelBuilderFactory;
+
+/**
+ * This is exactly similar to {@link org.apache.calcite.rel.rules.JoinPushThroughJoinRule}. It only
+ * checks if the condition of the new bottom join is supported.
+ */
+public class BeamJoinPushThroughJoinRule extends JoinPushThroughJoinRule {
+  /** Instance of the rule that works on logical joins only, and pushes to the right. */
+  public static final RelOptRule RIGHT =
+      new BeamJoinPushThroughJoinRule(
+          "BeamJoinPushThroughJoinRule:right",
+          true,
+          LogicalJoin.class,
+          RelFactories.LOGICAL_BUILDER);
+
+  /** Instance of the rule that works on logical joins only, and pushes to the left. */
+  public static final RelOptRule LEFT =
+      new BeamJoinPushThroughJoinRule(
+          "BeamJoinPushThroughJoinRule:left",
+          false,
+          LogicalJoin.class,
+          RelFactories.LOGICAL_BUILDER);
+
+  /** Creates a JoinPushThroughJoinRule. */
+  private BeamJoinPushThroughJoinRule(
+      String description,
+      boolean right,
+      Class<? extends Join> clazz,
+      RelBuilderFactory relBuilderFactory) {
+    super(description, right, clazz, relBuilderFactory);
+  }
+
+  @Override
+  public void onMatch(RelOptRuleCall call) {
+    super.onMatch(
+        new JoinRelOptRuleCall(
+            call,
+            rel -> {
+              Join topJoin = (Join) rel.getInput(0);
+              Join bottomJoin = (Join) ((Join) rel.getInput(0)).getLeft();
+              return BeamJoinRel.isJoinLegal(topJoin) && BeamJoinRel.isJoinLegal(bottomJoin);
+            }));
+  }
+}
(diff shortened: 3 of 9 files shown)


akedin:
retest this please

riazela:
run java post_commit

riazela:
Run Java PostCommit

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/planner/BeamRuleSets.java
```diff
@@ -103,6 +106,10 @@
 
           // join rules
           JoinPushExpressionsRule.INSTANCE,
+          JoinCommuteRule.INSTANCE,
```
amaliujia: Is there a good way to test the effect of adding such rules. We have seen examples [1] and [2] that adds rules that causes issues.

[1]: https://github.com/apache/beam/pull/8035
[2]: https://github.com/apache/beam/pull/6139

amaliujia: cc @kanterov as well who also had relevant experience.

riazela:
run java postcommit

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/planner/BeamRuleSets.java
riazela: I have tested the effect of adding these rules in terms of performance gain for some of the queries in https://docs.google.com/document/d/1DM_bcfFbIoc_vEoqQxhC7AvHBUDVCAwToC8TYGukkII/edit#

However, in terms of testing against bugs, all I can say is it is passing our current tests; therefore, if it can create illegal plans for some query, that means we are not covering it in our tests and it should be added. I am not sure how we can test this against all possible scenarios and queries. I think one way to improve our test cases is checking other systems' tests (Such as Calcite's test cases) and add them to ours. I will appreciate if you have any suggestion for that

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rule/JoinRelOptRuleCall.java
```diff
@@ -0,0 +1,101 @@
...
+    return originalCall.getParents();
+  }
+
+  @Override
+  public RelBuilder builder() {
+    return originalCall.builder();
+  }
+
+  @Override
+  public void transformTo(RelNode rel, Map<RelNode, RelNode> equiv) {
```
akedin: nit: can you move this and `JoinChecker` to the top and add a comment that these are the only two custom things, and everything else is delegated to the `originalCall`?

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rule/BeamJoinPushThroughJoinRule.java
```diff
@@ -0,0 +1,70 @@
...
+      boolean right,
+      Class<? extends Join> clazz,
+      RelBuilderFactory relBuilderFactory) {
+    super(description, right, clazz, relBuilderFactory);
+  }
+
+  @Override
+  public void onMatch(RelOptRuleCall call) {
+    super.onMatch(
+        new JoinRelOptRuleCall(
```
akedin: just an idea to think about, don't know if it makes sense:

* implement the check `BeamJoinRel.isJoinLegal(topJoin) && BeamJoinRel.isJoinLegal(bottomJoin)` in `JoinRelOptRuleCall`;
* create an interface like `JoinInputNodesMatcher` or something, which returns the top and the bottom join from the node, e.g.:

```java
interface JoinInputNodesMatcher  implements ... {
  Pair<Join, Join> getTopAndBottomJoin(RelOptRuleCall call);

  default void onMatch(RelOptRuleCall call) {
     super.onMatch(new JoinRelOptRuleCall(this));
  }
}
```
* make the `BeamJoinPushThroughJoinRule` implement that interface;

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/planner/BeamRuleSets.java
amaliujia: But it's less common to have a change without any test?

I am thinking a test that you can enable rules and then check output plan.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rule/BeamJoinAssociateRule.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+public class BeamJoinAssociateRule extends JoinAssociateRule {
+  // ~ Static fields/initializers ---------------------------------------------
+
+  /** The singleton. */
+  public static final org.apache.beam.sdk.extensions.sql.impl.rule.BeamJoinAssociateRule INSTANCE =
+      new org.apache.beam.sdk.extensions.sql.impl.rule.BeamJoinAssociateRule(
+          RelFactories.LOGICAL_BUILDER);
+
+  // ~ Constructors -----------------------------------------------------------
```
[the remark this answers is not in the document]
riazela: Removed it.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/planner/BeamRuleSets.java
akedin: In general we should have tests. However in this case the inputs/outputs are not changing depending on the inputs. We should make sure that we have a couple of tests that we know trigger these rules but we cannot enforce it, i.e. I don't know how we can ensure that we even hit the rule by looking at the results. If we don't know what a rule breaks, can we come up with a meaningful test for it?

I also don't think we should check the plans. It will be too brittle. Rules can be enabled and disabled independently and probably interact in hard-to-predict ways. If VolcanoPlanner (or in general Calcite stuff) is tested, then we should not re-test it and accept that it applies the rules correctly. We add some rules that are applied the same way as all other rules and inherit from them. In this case the rules behave the same way as built-in Calcite rules, and reject the plans that we already don't support in the join rel.

It feels like the only issue can be is the applicability of the reordering operation itself. Looking at what join operations we support and how it works (e.g. window is inherited for non-merging windows, or the input has to be explicitly re-windowed for merging windows, and that we only allow the default trigger that has itself as a continuation trigger), it looks safe and from the top of the head I cannot come up with a test case that we accept but that would break with the rule / without the rule.

Probably a unit test of the join rejection would make sense but I don't think it's  really valuable as we literally call the same code that we already do in the existing join rel and that probably already have tests (this should be confirmed) that trigger that logic.

amaliujia: Give me some time that let me check the details of PR and how a useful test could be.

Generally if we are making changes that we cannot test, we should postponing such changes until we figure out how to test it.

I think planner rules can be tested because I see Calcite and Flink have tests around optimization rules.

amaliujia: Because this PR is implementing reordering joins, a useful test would be a test, in which a three-way join is reordered. As join reordering is already measured in https://docs.google.com/document/d/1DM_bcfFbIoc_vEoqQxhC7AvHBUDVCAwToC8TYGukkII/edit#, wouldn't it be straightforward to have a similar test? Without such a test, how do we even know if join reordering is working?

In terms of checking output plans, Flink has been doing many tests on Calcite optimization rules (see [here](https://github.com/apache/flink/blob/master/flink-table/flink-table-planner-blink/src/test/scala/org/apache/flink/table/plan/rules/logical/RewriteMultiJoinConditionRuleTest.scala) and [here](https://github.com/apache/flink/tree/master/flink-table/flink-table-planner-blink/src/test/resources/org/apache/flink/table/plan)). Flink's practice has shown that verifying output plan is deterministic and stable.

The basic idea is if you want to test an optimization, only enable relevant rules in test case(so rules are hit will be known) I can see by Flink's way, you can test rules even if rules can be disabled and enabled independently:

In [BeamRuleSets.java](https://github.com/apache/beam/blob/master/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/planner/BeamRuleSets.java)
```
static BeamJoinReorderingRelSet = 
{JoinCommuteRule.INSTANCE, 
JoinAssociateRule.INSTANCE}
```

In JoinReorderingTest.java
```
FrameworkConfig testConfig = createTestConfig(BeamJoinReorderingRelSet)
Planner testPlanner = Frameworks.getPlanner(testConfig);

// setup input tables

expected_plan = PlanLoader.load(testcase.class)
verifyPlan(testPlanner.getPlan(sql), expected_plan)
```

By doing so, if a relevant rule is disable, it will break existing join reordering tests, which has good value for us. It also justifies this PR is doing join reordering.

Because we are starting the effort to have more optimization rules in BeamSQL, Flink's practice on testing is a great example that we can learn and apply to BeamSQL to maintain our codebase's health.

riazela: I just added a test to check if the join of three tables are reordered in the expected way.

sdks/java/extensions/sql/src/test/java/org/apache/beam/sdk/extensions/sql/impl/rule/JoinReorderingTest.java
```diff
@@ -0,0 +1,156 @@
...
+    createThreeTables(env, tableProvider);
+
+    BeamRelNode parsedQuery =
+        env.parseQuery(
+            "select * from large_table "
+                + " JOIN medium_table on large_table.medium_key = medium_table.large_key "
+                + " JOIN small_table on medium_table.small_key = small_table.medium_key ");
+    assertLargeIsTop(parsedQuery);
+
+    parsedQuery =
```
akedin: It's better to have many small tests, the goal is to be able to identify the problem quickly looking at the failure. I would split this into tests like `testSmallTableOnTheLeft()`, `testSmallTableOnTheRight()` etc

sdks/java/extensions/sql/src/test/java/org/apache/beam/sdk/extensions/sql/impl/rule/JoinReorderingTest.java
```diff
@@ -0,0 +1,156 @@
...
+import org.apache.beam.sdk.extensions.sql.impl.rel.BeamRelNode;
+import org.apache.beam.sdk.extensions.sql.meta.provider.test.TestTableProvider;
+import org.apache.beam.sdk.values.Row;
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.core.Join;
+import org.junit.Assert;
+import org.junit.Test;
+
+/**
+ * This test ensures that we are reordering joins and get a plan similar to Join(large,Join(small,
```
akedin: I think the critical part is to only include the rules you need as Rui suggested (re-ordering rule in this case), otherwise you don't know for sure why the large join is at the top. And we should also include the opposite tests, that without the rule the order is the same as in original query.

amaliujia: +1

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/planner/BeamRuleSets.java
akedin: There are few separate kinds of tests relevant here:

* tests of Beam/Beam SQL:
  * we test that the system works correctly;
  * we verify that expected inputs produce correct results;
     * as long as everything works correctly and produces correct data, everything is good;
  * if we add a new optimization rule that doesn't affect the behavior+correctness of the system, no tests in this category should fail;

* tests of specific rules. They test that a concrete rule works as expected. To test it we need to make sure that:
  * without the rule the behavior doesn't happen (no reorders);
  * with the rule the behavior happens (reorder does happen);
  * this specific rule is applied and caused the change, and no other rule could have caused this effect:
    * this potentially can happen if there are multiple interdependent rules that trigger in presence of each other. Whether this is a correct and intended situation or not we should be isolating this case;
  * if someone adds another rule to the bigger "production" set, this suite should not break, it only tests the specific rule;
  * if we decide to not use the rule and exclude it from the "production" rule set, this test should not break;

* "integration" test, or system configuration test. We ensure that the system is configured the certain way:
   * we can look at what rules are enabled and fail if the enabled rule set is different from what we expect;
   * we can test that the larger "production" ruleset has some behavior, e.g. that all rules combined have the effect of ordering enabled, so that if someone disables it with another rule, it breaks;

I am not sure if we currently have a straightforward way to setup the tests that exercise only one rule. E.g. to test a concrete rule we probably need to setup the parser and make it work without all rules except the ones we need, probably even without Beam physical rules (which we also don't test directly). If we are able to get something like this from Calcite/Flink, we should do it.

sdks/java/extensions/sql/src/test/java/org/apache/beam/sdk/extensions/sql/impl/rule/JoinReorderingTest.java
riazela: I added two more types of tests. One is testing if each of the rules alone acts correctly (by once converting a query without that and once converting it with the rule) and the other tests if the entire system without reordering rules does not reorder joins.

However, I still do not think it is a good idea to test the entire system without the join reordering rules to check if they are not producing optimal join order. (testSystemNotReorderingWithoutRules())

We expect the system to reorder the joins. But I am not sure why it should be an expected behavior of the system not to reorder joins without these specific rules. (We are not checking this for many other rules and features in the system. For instance we are not checking if the filter will be pushed down of the join in case of removal of FilterJoinRule. And I think this can also be applied to many "subsets of rules" that do some sort of behavior when applied together.) 

Furthermore, this test will break if someone adds a rule that reorder some certain types of joins or reorder all joins in some other way and I am not sure why we should prevent them from doing that as long as the system is still producing our expected optimized order.

sdks/java/extensions/sql/src/test/java/org/apache/beam/sdk/extensions/sql/impl/rule/JoinReorderingTest.java
```diff
@@ -0,0 +1,462 @@
...
+    assertTopTableInJoins(parsedQuery, "large_table");
+  }
+
+  @Test
+  public void testSystemNotReorderingWithoutRules() {
+    TestTableProvider tableProvider = new TestTableProvider();
+    createThreeTables(tableProvider);
+    ImmutableList.Builder<RelOptRule> ruleSetBuilder = ImmutableList.<RelOptRule>builder();
+    for (RuleSet set : BeamRuleSets.getRuleSets()) {
+      for (RelOptRule rule : set) {
```
akedin: nit: I think it can be expressed in a declarative way, something like:
```java
set
 .stream()
 .flatMap(set -> set.stream()) //whatever the right way to convert the ruleset to stream of rules
 .filter(rule -> ...)
 .collect(toList());
```
This way it's stateless, has less moving parts, and less nesting

sdks/java/extensions/sql/src/test/java/org/apache/beam/sdk/extensions/sql/impl/rule/JoinReorderingTest.java
akedin: It makes sense. The behavior of the system is our design decision. One way to document our design is to write tests. Until now there were no tests around these things and I don't think we had a clear idea of what's happening and didn't have such contracts defined around rules. Now that we do it's fine to go one way or another if it's conscious and explicit. In this case the actual behavior is not going to change no matter how we decide to approach the "rules integration" part and I doubt that anytime soon we will face these kinds of issues. And if we do, we can re-assess the decisions at that point if needed. I don't have good reasons to lean towards one way or another at this point.

sdks/java/extensions/sql/src/test/java/org/apache/beam/sdk/extensions/sql/impl/rule/JoinReorderingTest.java
riazela: Yes, thank you. I just changed it.

akedin:

sdks/java/extensions/sql/src/test/java/org/apache/beam/sdk/extensions/sql/impl/rule/JoinReorderingTest.java
amaliujia: Agree with Anton that changes and designs are documented by tests. As we are at an early stage of having optimizations, what we have now is definitely not perfect and we can refine them as time goes.

amaliujia:
LGTM

tvalentyn:
I think this may be breaking Java precommit tests:

```
16:23:09 > Task :sdks:java:extensions:sql:compileTestJava
16:23:09 /home/jenkins/jenkins-slave/workspace/beam_PreCommit_Java_Phrase@2/src/sdks/java/extensions/sql/src/test/java/org/apache/beam/sdk/extensions/sql/impl/rule/JoinReorderingTest.java:35: error: package org.apache.beam.vendor.guava.v20_0.com.google.common.collect does not exist
16:23:09 import org.apache.beam.vendor.guava.v20_0.com.google.common.collect.ImmutableList;
16:23:09                                                                    ^
16:23:09 /home/jenkins/jenkins-slave/workspace/beam_PreCommit_Java_Phrase@2/src/sdks/java/extensions/sql/src/test/java/org/apache/beam/sdk/extensions/sql/impl/rule/JoinReorderingTest.java:36: error: package org.apache.beam.vendor.guava.v20_0.com.google.common.collect does not exist
16:23:09 import org.apache.beam.vendor.guava.v20_0.com.google.common.collect.ImmutableMap;
16:23:09                                                                    ^
16:23:09 /home/jenkins/jenkins-slave/workspace/beam_PreCommit_Java_Phrase@2/src/sdks/java/extensions/sql/src/test/java/org/apache/beam/sdk/extensions/sql/impl/rule/JoinReorderingTest.java:367: error: cannot find symbol
16:23:09   private final ImmutableMap<String, Table> tables;
```

tvalentyn:
Opened https://issues.apache.org/jira/browse/BEAM-7784

PR #9049, opened 2019-07-12
[BEAM-7680] Removed flaky assertion and separated unit tests in synthetic source test

kkucharc:
Synthetic step unit tests was split into separate test cases. Removed flaky assertion which was depending on time.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/testing/synthetic_pipeline_test.py
+++ b/sdks/python/apache_beam/testing/synthetic_pipeline_test.py
@@ -57,7 +57,7 @@ class SyntheticPipelineTest(unittest.TestCase):
 
   # pylint: disable=expression-not-assigned
 
-  def test_count_elements_after_synthetic_step(self):
+  def test_synthetic_step_multiplies_output_elements_count(self):
     with beam.Pipeline() as p:
       pcoll = p | beam.Create(list(range(10))) | beam.ParDo(
           synthetic_pipeline.SyntheticStep(0, 0, 10))
@@ -71,10 +71,9 @@ class SyntheticPipelineTest(unittest.TestCase):
           synthetic_pipeline.SyntheticStep(0, 0.5, 10))
 
     elapsed = time.time() - start
-
     self.assertGreaterEqual(elapsed, 0.5, elapsed)
 
-  def test_count_elements_after_synthetic_sdf_step(self):
+  def test_synthetic_sdf_step_multiplies_output_elements_count(self):
     with beam.Pipeline() as p:
       pcoll = p | beam.Create(list(range(10))) | beam.ParDo(
           synthetic_pipeline.get_synthetic_sdf_step(0, 0, 10))
@@ -88,7 +87,6 @@ class SyntheticPipelineTest(unittest.TestCase):
           synthetic_pipeline.get_synthetic_sdf_step(0, 0.5, 10))
 
     elapsed = time.time() - start
-
     self.assertGreaterEqual(elapsed, 0.5, elapsed)
 
   def test_synthetic_step_split_provider(self):
--- a/sdks/python/apache_beam/testing/synthetic_pipeline_test.py
+++ b/sdks/python/apache_beam/testing/synthetic_pipeline_test.py
@@ -60,7 +60,7 @@ class SyntheticPipelineTest(unittest.TestCase):
   def test_count_elements_after_synthetic_step(self):
     with beam.Pipeline() as p:
       pcoll = p | beam.Create(list(range(10))) | beam.ParDo(
-          synthetic_pipeline.SyntheticStep(0, 0.5, 10))
+          synthetic_pipeline.SyntheticStep(0, 0, 10))
       assert_that(
           pcoll | beam.combiners.Count.Globally(), equal_to([100]))
 
@@ -72,12 +72,12 @@ class SyntheticPipelineTest(unittest.TestCase):
 
     elapsed = time.time() - start
 
-    self.assertTrue(elapsed >= 0.5, elapsed)
+    self.assertGreaterEqual(elapsed, 0.5, elapsed)
 
   def test_count_elements_after_synthetic_sdf_step(self):
     with beam.Pipeline() as p:
       pcoll = p | beam.Create(list(range(10))) | beam.ParDo(
-          synthetic_pipeline.get_synthetic_sdf_step(0, 0.5, 10))
+          synthetic_pipeline.get_synthetic_sdf_step(0, 0, 10))
       assert_that(
           pcoll | beam.combiners.Count.Globally(), equal_to([100]))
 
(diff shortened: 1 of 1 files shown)


sdks/python/apache_beam/testing/synthetic_pipeline_test.py
```diff
@@ -57,29 +57,39 @@ class SyntheticPipelineTest(unittest.TestCase):
...
 
+  def test_minimal_runtime_with_synthetic_step_delay(self):
+    start = time.time()
+    with beam.Pipeline() as p:
+      p | beam.Create(list(range(10))) | beam.ParDo(
+          synthetic_pipeline.SyntheticStep(0, 0.5, 10))
+
     elapsed = time.time() - start
-    # TODO(chamikaramj): Fix the flaky time based bounds.
-    self.assertTrue(0.5 <= elapsed <= 3, elapsed)
```
kkucharc: @chamikaramj I assumed that if we want to check if here if delay is set we can only measure bottom limit. WDYT?

sdks/python/apache_beam/testing/synthetic_pipeline_test.py
```diff
@@ -57,29 +57,39 @@ class SyntheticPipelineTest(unittest.TestCase):
...
-  def test_synthetic_step(self):
-    start = time.time()
+  def test_count_elements_after_synthetic_step(self):
     with beam.Pipeline() as p:
       pcoll = p | beam.Create(list(range(10))) | beam.ParDo(
           synthetic_pipeline.SyntheticStep(0, 0.5, 10))
       assert_that(
           pcoll | beam.combiners.Count.Globally(), equal_to([100]))
 
+  def test_minimal_runtime_with_synthetic_step_delay(self):
```
kkucharc: @chamikaramj I also decided to split it into two separate test cases to be give more meaningful names (hope it's meaningful enough).

kkucharc:
R: @chamikaramj R: @laschmidt can you take a look? cc: @pabloem @kamilwu

sdks/python/apache_beam/testing/synthetic_pipeline_test.py
```diff
@@ -57,29 +57,39 @@ class SyntheticPipelineTest(unittest.TestCase):
 
   # pylint: disable=expression-not-assigned
 
-  def test_synthetic_step(self):
-    start = time.time()
+  def test_count_elements_after_synthetic_step(self):
```
kamilwu: The best test names describe what the test does without the need for looking into the code. I suggest `test_synthetic_step_yields_10_elements_per_input_record`, as well as `test_synthetic_sdf_step_yields_10_elements_per_input_record` in the similar test below. WDYT?

sdks/python/apache_beam/testing/synthetic_pipeline_test.py
```diff
@@ -57,29 +57,39 @@ class SyntheticPipelineTest(unittest.TestCase):
...
+      p | beam.Create(list(range(10))) | beam.ParDo(
+          synthetic_pipeline.SyntheticStep(0, 0.5, 10))
+
     elapsed = time.time() - start
-    # TODO(chamikaramj): Fix the flaky time based bounds.
-    self.assertTrue(0.5 <= elapsed <= 3, elapsed)
 
-  def test_synthetic_sdf_step(self):
-    start = time.time()
+    self.assertTrue(elapsed >= 0.5, elapsed)
```
kamilwu: `self.assertGreaterEqual` will produce more meaningful message in case of failure.

kamilwu:
Thanks @kkucharc! I left two comments.

kkucharc:
Run Python PreCommit

sdks/python/apache_beam/testing/synthetic_pipeline_test.py
kkucharc: Fair point. Well, to be honest my `description what test does without looking into the code` was more about the state that is tested/expected so if the count is correct. But this is just different approach. We would be more precise and say `test_synthetic_step_returns_10_elements_per_10_input_record` what actually explains what is deeper in `SyntheticStep` implementation or just assume that we expect 100 records at the end and if someone wants more details will read the test: `test_synthetic_step_returns_100_elements`.

sdks/python/apache_beam/testing/synthetic_pipeline_test.py
kkucharc: Thanks, changing it

sdks/python/apache_beam/testing/synthetic_pipeline_test.py
kamilwu: Imagine that you have to test a completely different behaviour. You add a new test which covers it and asserts that the number of output elements is equal to 20 (or anything else). Then the test `test_synthetic_step_returns_20_elements` would be not distinguishable from the `test_synthetic_step_returns_100_elements` without investigating the code. And we want to avoid this.

kkucharc: To be honest I am not really sure if the numbers here are essential, but more if the count is just correct (that's why I suggested the original name). I can agree, if there will be similar test that will also return 100 elements, it would have to be distinguished somehow. So what do you think about compromise where we focus what is tested and we skip numbers, because those aren't important in this case? By the rule `test[Feature being tested]` we can have something like: `test_synthetic_step_multiplies_output_elements_count`.

sdks/python/apache_beam/testing/synthetic_pipeline_test.py
```diff
@@ -57,29 +57,39 @@ class SyntheticPipelineTest(unittest.TestCase):
 
   # pylint: disable=expression-not-assigned
 
-  def test_synthetic_step(self):
-    start = time.time()
+  def test_count_elements_after_synthetic_step(self):
     with beam.Pipeline() as p:
       pcoll = p | beam.Create(list(range(10))) | beam.ParDo(
           synthetic_pipeline.SyntheticStep(0, 0.5, 10))
```
kkucharc: Also I just realized that `delay` is here no longer tested in this testcase, so I'm removing it from here

sdks/python/apache_beam/testing/synthetic_pipeline_test.py
kamilwu: Allright. Looks good for me

robertwb:
@kamilwu any further thoughts on this?

kkucharc:
Run Portable_Python PreCommit

pabloem:

kkucharc:
Run Python PreCommit

kamilwu:
@robertwb No, looks good for me.

robertwb:
Run Python PreCommit

pabloem:
Run Python PreCommit

pabloem:
Thanks all!

PR #9056, opened 2019-07-12
[BEAM-7746] Add python type hints

chadrik:
EDIT:  jump to the latest conversation here: https://github.com/apache/beam/pull/9056#issuecomment-546732744

--- 

This is an early look at type annotations for the beam source.  It is _not_ attempting to address the issue of porting beam runtime typing from its own internal typehints module to the official tying module, so perhaps I should open a new ticket just for type hinting the source.

EDIT: done.  https://issues.apache.org/jira/browse/BEAM-7746

There's a lot left to do, including getting this hooked up to tests, but I wanted to get it out there so that A) I can get feedback before I get too far along, and B) in case anyone wants to join forces with me.

Some points for discussion / debate:

## type annotation style

In python 2.7, we're limited to type comments, and it can be quite difficult to keep them under the 80 character limit, since they're not very flexible. I'm following the lead of typing-heavy projects like mypy and using the following style for functions with more than one or two arguments:

```python
  def __init__(self,
               evaluation_context, # type: EvaluationContext
               applied_ptransform,  # type: AppliedPTransform
               input_committed_bundle,
               side_inputs
               ):
    # type: (...) -> None
```

## typing module import

Some projects establish a convention to import only the typing objects required for the current module, e.g.:

```python
from typing import Iterable, Tuple
```

Since beam has its own internal `typehints` module I'm importing the `typing` module in the same fashion to avoid conflicts/confusion.  However, this approach has the downside of making some type comments a lot longer and less legible, e.g. `typing.Union[typing.Iterable[str], str]`, which may make it impossible to stay below 80 characters in some cases.

## type checking pipeline construction

The type hints are useful first-and-foremost for developers of beam, but they should also prove useful for consumers of the beam API, i.e. developers of beam pipelines.   Though it's a secondary concern at the moment, I'm trying to ensure that the use case of pipeline creation is covered, so that mypy will generate proper errors for invalid combinations of transforms prior to submission, but due to a [bug/weakness](https://github.com/python/mypy/issues/6933) this is harder than expected.  I'm making progress writing a mypy plugin to patch up the problem, but it's not done yet.  Using the results of this process for runtime typing (e.g. to lookup coders) is out of scope for this PR.

## str vs unicode

Currently I'm glossing over this problem.  The _correct_ way to annotate strings for both python 2 and 3 is to use `typing.Text` and `typing.AnyStr`, but it's a huge pain in the ass, in particular because it's often unclear where unicode is valid or invalid (I've also run into inaccuracies in the python typeshed wrt strings).  In my own projects, I just annotate everything as `str` and handle the few cases where mypy complains about `unicode` by using `typing.cast(str, whatever)`, since the problem will get a lot simpler when we switch to python3, and `bytes` becomes the exception.  I'm open to suggestions here. 

---

I'll probably think of more later, but I'm out of time and want to get this out the door!

--- | --- | --- | --- | ---

### DIFF
--- a/buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy
+++ b/buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy
@@ -1768,7 +1768,7 @@ class BeamModulePlugin implements Plugin<Project> {
           project.exec { commandLine virtualenvCmd }
           project.exec {
             executable 'sh'
-            args '-c', ". ${project.ext.envdir}/bin/activate && pip install --retries 10 --upgrade tox==3.11.1 grpcio-tools==1.3.5"
+            args '-c', ". ${project.ext.envdir}/bin/activate && pip install --retries 10 --upgrade tox==3.11.1 -r ${project.rootDir}/sdks/python/build-requirements.txt"
           }
         }
         // Gradle will delete outputs whenever it thinks they are stale. Putting a
--- a/sdks/python/.pylintrc
+++ b/sdks/python/.pylintrc
@@ -143,6 +143,7 @@ disable =
   unnecessary-lambda,
   unnecessary-pass,
   unneeded-not,
+  unsubscriptable-object,
   unused-argument,
   unused-wildcard-import,
   useless-object-inheritance,
@@ -178,6 +179,7 @@ indent-after-paren=4
 ignore-long-lines=(?x)
   (^\s*(import|from)\s
    |^\s*(\#\ )?<?(https?|ftp):\/\/[^\s\/$.?#].[^\s]*>?$
+   |# type:
    )
 
 [VARIABLES]
--- a/sdks/python/apache_beam/io/gcp/datastore/v1/query_splitter.py
+++ b/sdks/python/apache_beam/io/gcp/datastore/v1/query_splitter.py
@@ -37,7 +37,7 @@ try:
                            PropertyFilter.GREATER_THAN,
                            PropertyFilter.GREATER_THAN_OR_EQUAL]
 except ImportError:
-  UNSUPPORTED_OPERATORS = None
+  UNSUPPORTED_OPERATORS = None  # type: ignore
 # pylint: enable=wrong-import-order, wrong-import-position
 
 
--- a/sdks/python/apache_beam/io/gcp/pubsub.py
+++ b/sdks/python/apache_beam/io/gcp/pubsub.py
@@ -26,6 +26,8 @@ from __future__ import absolute_import
 
 import re
 from builtins import object
+from typing import Any
+from typing import Optional
 
 from future.utils import iteritems
 from past.builtins import unicode
@@ -87,6 +89,7 @@ class PubsubMessage(object):
 
   @staticmethod
   def _from_proto_str(proto_msg):
+    # type: (bytes) -> PubsubMessage
     """Construct from serialized form of ``PubsubMessage``.
 
     Args:
@@ -121,6 +124,7 @@ class PubsubMessage(object):
 
   @staticmethod
   def _from_message(msg):
+    # type: (Any) -> PubsubMessage
     """Construct from ``google.cloud.pubsub_v1.subscriber.message.Message``.
 
     https://googleapis.github.io/google-cloud-python/latest/pubsub/subscriber/api/message.html
@@ -134,8 +138,14 @@ class ReadFromPubSub(PTransform):
   """A ``PTransform`` for reading from Cloud Pub/Sub."""
   # Implementation note: This ``PTransform`` is overridden by Directrunner.
 
-  def __init__(self, topic=None, subscription=None, id_label=None,
-               with_attributes=False, timestamp_attribute=None):
+  def __init__(self,
+               topic=None,  # type: Optional[str]
+               subscription=None,  # type: Optional[str]
+               id_label=None,  # type: Optional[str]
+               with_attributes=False,  # type: bool
+               timestamp_attribute=None  # type: Optional[str]
+              ):
+    # type: (...) -> None
     """Initializes ``ReadFromPubSub``.
 
     Args:
@@ -242,8 +252,13 @@ class WriteToPubSub(PTransform):
   """A ``PTransform`` for writing messages to Cloud Pub/Sub."""
   # Implementation note: This ``PTransform`` is overridden by Directrunner.
 
-  def __init__(self, topic, with_attributes=False, id_label=None,
-               timestamp_attribute=None):
+  def __init__(self,
+               topic,  # type: str
+               with_attributes=False,  # type: bool
+               id_label=None,  # type: Optional[str]
+               timestamp_attribute=None  # type: Optional[str]
+              ):
+    # type: (...) -> None
     """Initializes ``WriteToPubSub``.
 
     Args:
@@ -267,6 +282,7 @@ class WriteToPubSub(PTransform):
 
   @staticmethod
   def to_proto_str(element):
+    # type: (PubsubMessage) -> bytes
     if not isinstance(element, PubsubMessage):
       raise TypeError('Unexpected element. Type: %s (expected: PubsubMessage), '
                       'value: %r' % (type(element), element))
@@ -327,8 +343,13 @@ class _PubSubSource(dataflow_io.NativeSource):
       fetches ``PubsubMessage`` protobufs.
   """
 
-  def __init__(self, topic=None, subscription=None, id_label=None,
-               with_attributes=False, timestamp_attribute=None):
+  def __init__(self,
+               topic=None,  # type: Optional[str]
+               subscription=None,  # type: Optional[str]
+               id_label=None,  # type: Optional[str]
+               with_attributes=False,  # type: bool
+               timestamp_attribute=None  # type: Optional[str]
+              ):
     self.coder = coders.BytesCoder()
     self.full_topic = topic
     self.full_subscription = subscription
@@ -385,7 +406,12 @@ class _PubSubSink(dataflow_io.NativeSink):
   This ``NativeSource`` is overridden by a native Pubsub implementation.
   """
 
-  def __init__(self, topic, id_label, with_attributes, timestamp_attribute):
+  def __init__(self,
+               topic,  # type: str
+               id_label,  # type: Optional[str]
+               with_attributes,  # type: bool
+               timestamp_attribute  # type: Optional[str]
+              ):
     self.coder = coders.BytesCoder()
     self.full_topic = topic
     self.id_label = id_label
--- a/sdks/python/apache_beam/io/iobase.py
+++ b/sdks/python/apache_beam/io/iobase.py
@@ -45,6 +45,7 @@ from typing import Iterable
 from typing import Iterator
 from typing import Optional
 from typing import Tuple
+from typing import TypeVar
 
 from apache_beam import coders
 from apache_beam import pvalue
@@ -70,6 +71,7 @@ if TYPE_CHECKING:
 __all__ = ['BoundedSource', 'RangeTracker', 'Read', 'RestrictionTracker',
            'Sink', 'Write', 'Writer']
 
+U = TypeVar('U')
 
 # Encapsulates information about a bundle of a source generated when method
 # BoundedSource.split() is invoked.
@@ -896,14 +898,17 @@ class Read(ptransform.PTransform):
         return source.split(
             self.get_desired_chunk_size(self.source.estimate_size()))
 
+      def read_splits(split):
+        return split.source.read(
+            split.source.get_range_tracker(
+                split.start_position, split.stop_position))
+
       return (
           pbegin
           | core.Impulse()
           | 'Split' >> core.FlatMap(split_source)
           | util.Reshuffle()
-          | 'ReadSplits' >> core.FlatMap(lambda split: split.source.read(
-              split.source.get_range_tracker(
-                  split.start_position, split.stop_position))))
+          | 'ReadSplits' >> core.FlatMap(read_splits))
     else:
       # Treat Read itself as a primitive.
       return pvalue.PCollection(self.pipeline,
@@ -1015,10 +1020,16 @@ class WriteImpl(ptransform.PTransform):
     do_once = pcoll.pipeline | 'DoOnce' >> core.Create([None])
     init_result_coll = do_once | 'InitializeWrite' >> core.Map(
         lambda _, sink: sink.initialize_write(), self.sink)
-    if getattr(self.sink, 'num_shards', 0):
-      min_shards = self.sink.num_shards
+
+    def addkey(x):
+      # type: (U) -> Tuple[Any, U]
+      return (None, x)
+
+    num_shards = getattr(self.sink, 'num_shards', 0)
+    if num_shards:
+      min_shards = num_shards
       if min_shards == 1:
-        keyed_pcoll = pcoll | core.Map(lambda x: (None, x))
+        keyed_pcoll = pcoll | core.Map(addkey)
       else:
         keyed_pcoll = pcoll | core.ParDo(_RoundRobinKeyFn(min_shards))
       write_result_coll = (keyed_pcoll
@@ -1028,15 +1039,19 @@ class WriteImpl(ptransform.PTransform):
                                _WriteKeyedBundleDoFn(self.sink),
                                AsSingleton(init_result_coll)))
     else:
+      def getvalue(x):
+        # type: (Tuple[Any, Iterable[U]]) -> Iterable[U]
+        return x[1]
+
       min_shards = 1
       write_result_coll = (pcoll
                            | 'WriteBundles' >>
                            core.ParDo(_WriteBundleDoFn(self.sink),
                                       AsSingleton(init_result_coll))
-                           | 'Pair' >> core.Map(lambda x: (None, x))
+                           | 'Pair' >> core.Map(addkey)
                            | core.WindowInto(window.GlobalWindows())
                            | core.GroupByKey()
-                           | 'Extract' >> core.FlatMap(lambda x: x[1]))
+                           | 'Extract' >> core.FlatMap(getvalue))
     # PreFinalize should run before FinalizeWrite, and the two should not be
     # fused.
     pre_finalize_coll = do_once | 'PreFinalize' >> core.FlatMap(
--- a/sdks/python/apache_beam/io/iobase.py
+++ b/sdks/python/apache_beam/io/iobase.py
@@ -920,10 +920,13 @@ class Read(ptransform.PTransform):
 
   def _infer_output_coder(self, input_type=None, input_coder=None):
     # type: (...) -> Optional[coders.Coder]
+    from apache_beam.runners.dataflow.native_io import iobase as dataflow_io
     if isinstance(self.source, BoundedSource):
       return self.source.default_output_coder()
-    else:
+    elif isinstance(self.source, dataflow_io.NativeSource):
       return self.source.coder
+    else:
+      return None
 
   def display_data(self):
     return {'source': DisplayDataItem(self.source.__class__,
--- a/sdks/python/apache_beam/io/iobase.py
+++ b/sdks/python/apache_beam/io/iobase.py
@@ -39,6 +39,12 @@ import uuid
 from builtins import object
 from builtins import range
 from collections import namedtuple
+from typing import TYPE_CHECKING
+from typing import Any
+from typing import Iterable
+from typing import Iterator
+from typing import Optional
+from typing import Tuple
 
 from apache_beam import coders
 from apache_beam import pvalue
@@ -55,6 +61,11 @@ from apache_beam.transforms.display import HasDisplayData
 from apache_beam.utils import timestamp
 from apache_beam.utils import urns
 from apache_beam.utils.windowed_value import WindowedValue
+
+if TYPE_CHECKING:
+  from apache_beam.io import restriction_trackers
+  from apache_beam.runners.pipeline_context import PipelineContext
+  from apache_beam.utils.timestamp import Timestamp
 
 __all__ = ['BoundedSource', 'RangeTracker', 'Read', 'RestrictionTracker',
            'Sink', 'Write', 'Writer']
@@ -85,6 +96,10 @@ class SourceBase(HasDisplayData, urns.RunnerApiFn):
   """Base class for all sources that can be passed to beam.io.Read(...).
   """
   urns.RunnerApiFn.register_pickle_urn(python_urns.PICKLED_SOURCE)
+
+  def is_bounded(self):
+    # type: () -> bool
+    raise NotImplementedError
 
 
 class BoundedSource(SourceBase):
@@ -124,6 +139,7 @@ class BoundedSource(SourceBase):
   """
 
   def estimate_size(self):
+    # type: () -> Optional[int]
     """Estimates the size of source in bytes.
 
     An estimate of the total size (in bytes) of the data that would be read
@@ -136,7 +152,12 @@ class BoundedSource(SourceBase):
     """
     raise NotImplementedError
 
-  def split(self, desired_bundle_size, start_position=None, stop_position=None):
+  def split(self,
+            desired_bundle_size,  # type: int
+            start_position=None,  # type: Optional[int]
+            stop_position=None,  # type: Optional[int]
+           ):
+    # type: (...) -> Iterator[SourceBundle]
     """Splits the source into a set of bundles.
 
     Bundles should be approximately of size ``desired_bundle_size`` bytes.
@@ -153,7 +174,11 @@ class BoundedSource(SourceBase):
     """
     raise NotImplementedError
 
-  def get_range_tracker(self, start_position, stop_position):
+  def get_range_tracker(self,
+                        start_position,  # type: Optional[int]
+                        stop_position,  # type: Optional[int]
+                       ):
+    # type: (...) -> RangeTracker
     """Returns a RangeTracker for a given position range.
 
     Framework may invoke ``read()`` method with the RangeTracker object returned
@@ -837,6 +862,7 @@ class Read(ptransform.PTransform):
   """A transform that reads a PCollection."""
 
   def __init__(self, source):
+    # type: (SourceBase) -> None
     """Initializes a Read transform.
 
     Args:
@@ -884,9 +910,11 @@ class Read(ptransform.PTransform):
                                 is_bounded=self.source.is_bounded())
 
   def get_windowing(self, unused_inputs):
+    # type: (...) -> core.Windowing
     return core.Windowing(window.GlobalWindows())
 
   def _infer_output_coder(self, input_type=None, input_coder=None):
+    # type: (...) -> Optional[coders.Coder]
     if isinstance(self.source, BoundedSource):
       return self.source.default_output_coder()
     else:
@@ -898,6 +926,7 @@ class Read(ptransform.PTransform):
             'source_dd': self.source}
 
   def to_runner_api_parameter(self, context):
+    # type: (PipelineContext) -> Tuple[str, beam_runner_api_pb2.ReadPayload]
     return (common_urns.deprecated_primitives.READ.urn,
             beam_runner_api_pb2.ReadPayload(
                 source=self.source.to_runner_api(context),
@@ -907,6 +936,7 @@ class Read(ptransform.PTransform):
 
   @staticmethod
   def from_runner_api_parameter(parameter, context):
+    # type: (beam_runner_api_pb2.ReadPayload, PipelineContext) -> Read
     return Read(SourceBase.from_runner_api(parameter.source, context))
 
 
@@ -977,6 +1007,7 @@ class WriteImpl(ptransform.PTransform):
   """Implements the writing of custom sinks."""
 
   def __init__(self, sink):
+    # type: (Sink) -> None
     super(WriteImpl, self).__init__()
     self.sink = sink
 
@@ -1089,6 +1120,7 @@ def _finalize_write(unused_element, sink, init_result, write_results,
 class _RoundRobinKeyFn(core.DoFn):
 
   def __init__(self, count):
+    # type: (int) -> None
     self.count = count
 
   def start_bundle(self):
@@ -1134,6 +1166,7 @@ class RestrictionTracker(object):
     raise NotImplementedError
 
   def current_progress(self):
+    # type: () -> RestrictionProgress
     """Returns a RestrictionProgress object representing the current progress.
     """
     raise NotImplementedError
@@ -1275,6 +1308,7 @@ class RestrictionTracker(object):
     raise NotImplementedError
 
   def deferred_status(self):
+    # type: () -> Optional[Tuple[restriction_trackers.OffsetRange, Timestamp]]
     """ Returns deferred_residual with deferred_watermark.
 
     TODO(BEAM-7472): Remove defer_status() once SDF.process() uses
@@ -1357,6 +1391,7 @@ class _SDFBoundedSourceWrapper(ptransform.PTransform):
       self._weight = restriction.weight
 
     def current_progress(self):
+      # type: () -> RestrictionProgress
       return RestrictionProgress(
           fraction=self._delegate_range_tracker.fraction_consumed())
 
--- a/sdks/python/apache_beam/transforms/ptransform.py
+++ b/sdks/python/apache_beam/transforms/ptransform.py
@@ -317,7 +317,7 @@ class PTransform(WithTypeHints, HasDisplayData):
 
   def __init__(self, label=None):
     super(PTransform, self).__init__()
-    self.label = label
+    self.label = label  # type: ignore # https://github.com/python/mypy/issues/3004
 
   @property
   def label(self):
--- a/sdks/python/apache_beam/transforms/ptransform.py
+++ b/sdks/python/apache_beam/transforms/ptransform.py
@@ -358,11 +358,12 @@ class PTransform(WithTypeHints, HasDisplayData):
     # type: () -> str
     return self.__class__.__name__
 
-  def with_input_types(self, input_type_hint):
+  def with_input_types(self, *input_type_hints, **unused_kwargs):
+    # type: (PTransformT, *Any, **Any) -> PTransformT
     """Annotates the input type of a :class:`PTransform` with a type-hint.
 
     Args:
-      input_type_hint (type): An instance of an allowed built-in type, a custom
+      input_type_hints (type): An instance of an allowed built-in type, a custom
         class, or an instance of a
         :class:`~apache_beam.typehints.typehints.TypeConstraint`.
 
@@ -377,8 +378,16 @@ class PTransform(WithTypeHints, HasDisplayData):
       :class:`PTransform` object. This allows chaining type-hinting related
       methods.
     """
+    if len(input_type_hints) != 1:
+      raise TypeCheckError(
+          'PTransform %s expects a single input type hint. Got %d)' % (
+              self.__class__, len(input_type_hints)))
+    if unused_kwargs:
+      raise TypeCheckError(
+          'PTransform %s does not take any keyword type hints. Got %s)' % (
+              self.__class__, unused_kwargs))
     input_type_hint = native_type_compatibility.convert_to_beam_type(
-        input_type_hint)
+        input_type_hints[0])
     validate_composite_type_param(input_type_hint,
                                   'Type hints for a PTransform')
     return super(PTransform, self).with_input_types(input_type_hint)
--- a/sdks/python/apache_beam/transforms/ptransform.py
+++ b/sdks/python/apache_beam/transforms/ptransform.py
@@ -48,6 +48,17 @@ from builtins import object
 from builtins import zip
 from functools import reduce
 from functools import wraps
+from typing import TYPE_CHECKING
+from typing import Any
+from typing import Callable
+from typing import Dict
+from typing import Optional
+from typing import Sequence
+from typing import Tuple
+from typing import Type
+from typing import TypeVar
+from typing import Union
+from typing import overload
 
 from google.protobuf import message
 
@@ -68,11 +79,24 @@ from apache_beam.typehints.trivial_inference import instance_to_type
 from apache_beam.typehints.typehints import validate_composite_type_param
 from apache_beam.utils import proto_utils
 
+if TYPE_CHECKING:
+  from apache_beam import coders
+  from apache_beam.pipeline import Pipeline
+  from apache_beam.runners.pipeline_context import PipelineContext
+  from apache_beam.transforms.core import Windowing
+  from apache_beam.portability.api import beam_runner_api_pb2
+
 __all__ = [
     'PTransform',
     'ptransform_fn',
     'label_from_callable',
     ]
+
+T = TypeVar('T')
+PTransformT = TypeVar('PTransformT', bound='PTransform')
+ConstructorFn = Callable[
+    [Optional[Any], 'PipelineContext'],
+    Any]
 
 
 class _PValueishTransform(object):
@@ -307,27 +331,31 @@ class PTransform(WithTypeHints, HasDisplayData):
   with input as an argument.
   """
   # By default, transforms don't have any side inputs.
-  side_inputs = ()
+  side_inputs = ()  # type: Sequence[pvalue.AsSideInput]
 
   # Used for nullary transforms.
-  pipeline = None
+  pipeline = None  # type: Optional[Pipeline]
 
   # Default is unset.
-  _user_label = None
+  _user_label = None  # type: Optional[str]
 
   def __init__(self, label=None):
+    # type: (Optional[str]) -> None
     super(PTransform, self).__init__()
     self.label = label  # type: ignore # https://github.com/python/mypy/issues/3004
 
   @property
   def label(self):
+    # type: () -> str
     return self._user_label or self.default_label()
 
   @label.setter
   def label(self, value):
+    # type: (Optional[str]) -> None
     self._user_label = value
 
   def default_label(self):
+    # type: () -> str
     return self.__class__.__name__
 
   def with_input_types(self, input_type_hint):
@@ -409,6 +437,7 @@ class PTransform(WithTypeHints, HasDisplayData):
                 pvalue_.element_type))
 
   def _infer_output_coder(self, input_type=None, input_coder=None):
+    # type: (...) -> Optional[coders.Coder]
     """Returns the output coder to use for output of this transform.
 
     Note: this API is experimental and is subject to change; please do not rely
@@ -454,12 +483,14 @@ class PTransform(WithTypeHints, HasDisplayData):
         ' side_inputs=%s' % str(self.side_inputs) if self.side_inputs else '')
 
   def _check_pcollection(self, pcoll):
+    # type: (pvalue.PCollection) -> None
     if not isinstance(pcoll, pvalue.PCollection):
       raise error.TransformError('Expecting a PCollection argument.')
     if not pcoll.pipeline:
       raise error.TransformError('PCollection not part of a pipeline.')
 
   def get_windowing(self, inputs):
+    # type: (Any) -> Windowing
     """Returns the window function to be associated with transform's output.
 
     By default most transforms just return the windowing function associated
@@ -555,7 +586,45 @@ class PTransform(WithTypeHints, HasDisplayData):
     else:
       return input_dict
 
-  _known_urns = {}
+  _known_urns = {}  # type: Dict[str, Tuple[Optional[type], ConstructorFn]]
+
+  @classmethod
+  @overload
+  def register_urn(cls,
+                   urn,  # type: str
+                   parameter_type,  # type: Type[T]
+                  ):
+    # type: (...) -> Callable[[Union[type, Callable[[T, PipelineContext], Any]]], Callable[[T, PipelineContext], Any]]
+    pass
+
+  @classmethod
+  @overload
+  def register_urn(cls,
+                   urn,  # type: str
+                   parameter_type,  # type: None
+                  ):
+    # type: (...) -> Callable[[Union[type, Callable[[bytes, PipelineContext], Any]]], Callable[[bytes, PipelineContext], Any]]
+    pass
+
+  @classmethod
+  @overload
+  def register_urn(cls,
+                   urn,  # type: str
+                   parameter_type,  # type: Type[T]
+                   constructor  # type: Callable[[T, PipelineContext], Any]
+                  ):
+    # type: (...) -> None
+    pass
+
+  @classmethod
+  @overload
+  def register_urn(cls,
+                   urn,  # type: str
+                   parameter_type,  # type: None
+                   constructor  # type: Callable[[bytes, PipelineContext], Any]
+                  ):
+    # type: (...) -> None
+    pass
 
   @classmethod
   def register_urn(cls, urn, parameter_type, constructor=None):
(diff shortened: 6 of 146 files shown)


chadrik:
R: @aaltay 
R: @charlesccychen

charlesccychen:
CC: @udim

sdks/python/apache_beam/examples/wordcount.py
```diff
@@ -44,7 +52,8 @@ def __init__(self):
         self.__class__, 'word_len_dist')
     self.empty_line_counter = Metrics.counter(self.__class__, 'empty_lines')
 
-  def process(self, element):
+  def process(self, element, *args, **kwargs):
```
robertwb: Why are the *args and **kwargs parameters added here?

sdks/python/apache_beam/examples/wordcount.py
```diff
@@ -85,31 +98,66 @@ def run(argv=None):
...
+  read = 'read' >> _read
+  reveal_type(read)
+
   # Read the text file[pattern] into a PCollection.
-  lines = p | 'read' >> ReadFromText(known_args.input)
+  lines = p | read
+  reveal_type(lines)  # PCollection[unicode*]
+
+  def make_ones(x):
+    # type: (T) -> Tuple[T, int]
```
robertwb: Can this not be inferred? Similarly below.

sdks/python/apache_beam/examples/wordcount.py
```diff
@@ -65,8 +74,12 @@ def process(self, element):
       self.word_lengths_dist.update(len(w))
     return words
 
+  def to_runner_api_parameter(self, unused_context):
+    pass
+
 
 def run(argv=None):
+  # type: (Optional[Sequence[str]]) -> None
```
robertwb: One of the benefits of gradual typing is that types are optional. It's unclear that this hint (among others) provides a net positive value to (to merit its verbosity).

sdks/python/apache_beam/io/gcp/datastore/v1/query_splitter.py
```diff
@@ -37,7 +37,7 @@
                            PropertyFilter.GREATER_THAN,
                            PropertyFilter.GREATER_THAN_OR_EQUAL]
 except ImportError:
-  UNSUPPORTED_OPERATORS = None
+  UNSUPPORTED_OPERATORS = None  # type: ignore
```
robertwb: Are all of these due to trying to suppress warnings?

sdks/python/apache_beam/io/gcp/pubsub.py
```diff
@@ -134,8 +136,13 @@ class ReadFromPubSub(PTransform):
   """A ``PTransform`` for reading from Cloud Pub/Sub."""
   # Implementation note: This ``PTransform`` is overridden by Directrunner.
 
-  def __init__(self, topic=None, subscription=None, id_label=None,
-               with_attributes=False, timestamp_attribute=None):
+  def __init__(self,
+               topic=None,  # type: typing.Optional[str]
```
robertwb: If we keep these, I wonder if it'd be easier to read if the comments were all aligned.

sdks/python/apache_beam/io/iobase.py
```diff
@@ -972,14 +979,15 @@ def expand(self, pcoll):
                        'or be a PTransform. Received : %r' % self.sink)
 
 
-class WriteImpl(ptransform.PTransform):
+class WriteImpl(ptransform.PTransform[InT, None]):
```
robertwb: Or should this be a PTransform[pvalue.PCollection[InT], None]?

sdks/python/apache_beam/examples/wordcount.py
robertwb: (And I realize this is a WIP, but we should be very careful about changing the wordcount (and other examples) to keep then concise as an introduction to the Beam APIs.

chadrik: Sorry, I should have clarified that the changes to wordcount really amount to tests, primarily of the mypy plugin.  Ultimately, we'll want a separate example to demonstrate typing.  I'll get this reverted soon.

sdks/python/apache_beam/io/iobase.py
chadrik: IMHO, that would be redundant.  It won't change the behavior of the type checks, and it would add verbosity to _a lot_ of class definitions, since there are numerous `PTransforms` throughout the code.

sdks/python/apache_beam/io/gcp/datastore/v1/query_splitter.py
chadrik: Correct.  I can add a comment to the code clarify.  

The problem is that if we mark `UNSUPPORTED_OPERATORS` as `Optional` then mypy will generate numerous errors in this module, because the code assumes the import succeeded and that `UNSUPPORTED_OPERATORS` is _not_ `None`.   

An alternative would be to set it to `[]`.

sdks/python/apache_beam/examples/wordcount.py
chadrik: Can you explain what you mean?   

Let me try to answer what I think you're asking about.

First, even for a simple function like `make_ones`, mypy does not guess at the arguments and return type.  

Next, as a user annotating pipelines I have a few of choices  (all of them are valid):

1. annotate functions, and let mypy infer the types of the `PTransforms` they are passed to
2. annotate `PTransforms`
3. annotate both, as a way to guard against unintentional incompatibilities

Option 2 would look like this:

```python
  def make_ones(x):
    return (x, 1)

  makemap = beam.Map(make_ones)  # type: beam.Map[T, Tuple[T, int]]
```

To be clear, in option 2, mypy is not using the type of `beam.Map` to infer the type of the `make_ones`, it's simply assuming it's correct, since in our thought experiment it's unannotated.  In other words, the following would not generate an error because `make_ones` is untyped:

```python
  def make_ones(x):
    return (x + 1, 1)  # x can only be a numeric type!

  makemap = beam.Map(make_ones)  # type: beam.Map[str, Tuple[str, int]]
```

This module is very messy right now, so I apologize for any confusion that it's generated.

sdks/python/apache_beam/io/gcp/pubsub.py
chadrik: I'll leave that up to the Beam Team, but we're still going to have the line length limit to contend with.

sdks/python/apache_beam/io/iobase.py
chadrik: oh, also, it will require an import of the `apache_beam.pvalue` module into all of those modules as well.  you won't be able to protect it inside a `typing.TYPE_CHECKING` block since it's used in the class definition.

chadrik:
> typing module import: Given that we're deprecating the old type hints, I think we should take a sweep through the codebase and replace all non-qualified typehints with their typing equivalents (when possible) or qualified ones (where not). This would allow us to use the common convention of importing from the typing module directly.

Sorry, I don't completely follow this.  IIUC, we can't replace `apache_beam.typehints` with `typing` until beam knows how to deal with `typing` internally, right?

chadrik:
> type checking pipeline construction I think the goal of JIRA-7060 is specifically to make type annotations useful to pipeline developers. Making tools like mypy work on the beam codebase, and user pipelines in general, is another (worthy) goal.

I made a new ticket for the latter, and changed the subject of this PR:  https://issues.apache.org/jira/browse/BEAM-7746

sdks/python/apache_beam/examples/wordcount.py
robertwb: Sounds good. Thanks for the clarification.

sdks/python/apache_beam/examples/wordcount.py
robertwb: That's really unfortunate... I'd have expected more of mypy :(. In that case our own inference (which does handle this) is all the more important.

sdks/python/apache_beam/io/iobase.py
robertwb: But some transforms take and return other things than PCollections. Maybe if T is shorthand for PCollection[T] but that seems ambiguous.

sdks/python/apache_beam/typehints/decorators.py
```diff
@@ -193,7 +200,7 @@ def __repr__(self):
         self.input_types, self.output_types)
 
 
-class WithTypeHints(object):
+class WithTypeHints(Generic[InT, OutT]):
```
udim: I'm afraid that if we add a second typing system here will create confusion. If I write a PTransform, do I specify an input type to Generic[] or do I use a @with_input_type decorator? Which one is enforced?

sdks/python/setup.py
```diff
@@ -25,19 +25,71 @@
...
+  user_options = []
+
+  def initialize_options(self):
+    """Abstract method that is required to be overwritten"""
+
+  def finalize_options(self):
+    """Abstract method that is required to be overwritten"""
+
+  def get_project_path(self, include_dists=[]):
+      # with_2to3 = six.PY3 and getattr(self.distribution, 'use_2to3', False)
```
udim: leftover comments?

sdks/python/apache_beam/transforms/ptransform.py
```diff
@@ -465,56 +484,70 @@ def get_windowing(self, inputs):
...
       return _ChainedPTransform(self, right)
     return NotImplemented
 
-  def __ror__(self, left, label=None):
-    """Used to apply this PTransform to non-PValues, e.g., a tuple."""
-    pvalueish, pvalues = self._extract_input_pvalues(left)
-    pipelines = [v.pipeline for v in pvalues if isinstance(v, pvalue.PValue)]
-    if pvalues and not pipelines:
-      deferred = False
+  if not typing.TYPE_CHECKING:
```
udim: I wish this was a decorator instead of adding an indent.

mypy.ini
```diff
@@ -0,0 +1,19 @@
+[mypy]
+python_version = 2.7
```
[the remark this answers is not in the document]
chadrik: mypy needs python 3 to run, but this flag is for the version of the code that it's _checking_.  
 when the code is ready we'll want to run this in 3.x mode.

sdks/python/apache_beam/io/gcp/pubsub.py
```diff
@@ -121,6 +122,7 @@ def _to_proto_str(self):
 
   @staticmethod
   def _from_message(msg):
+    # type: () -> PubsubMessage
```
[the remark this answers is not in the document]
chadrik: you're right.  that's a mistake.

sdks/python/apache_beam/typehints/decorators.py
```diff
@@ -100,13 +101,19 @@ def foo((a, b)):
...
 __all__ = [
     'with_input_types',
     'with_output_types',
     'WithTypeHints',
     'TypeCheckError',
 ]
 
+InT = TypeVar('InT')
+OutT = TypeVar('OutT')
+WithTypeHintsT = TypeVar('WithTypeHintsT', bound='WithTypeHints')
```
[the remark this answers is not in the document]
chadrik: it tells mypy that any type filling type var `WithTypeHintsT` must be `WithTypeHints` or a sub-class thereof.

sdks/python/apache_beam/typehints/decorators.py
chadrik: True.  I think the somewhat unsatisfactory answer is that for the time being you need both: one for runtime checking and the other for static checking, until such a time as they can become the same.  I think trying to do that all in one PR is going to be too much.

off the top of my head, the order this should probably be done is:

1. support runtime type hints using `typing` module instead of `typehints`
2. add static type hints to the beam code and begin enforcing it using mypy (this PR)
3. support static validation of user pipelines (mypy plugin, etc)
4. support runtime validations based on `typing` annotations

There are a lot of "ifs" surrounding step 4.  We may need to get to python3-only first to avoid the pitfalls of type comments.  We may find that step 3 makes it less important.

sdks/python/setup.py
chadrik: yeah, this command class is based off of several in the distutils source, and it was unclear to me if we'd need to support this other case, so it's there as a reminder to figure that out. 

To be crystal clear, this PR is very much a WIP.  Just wanted to get eyes on it before I get too far ahead of myself.

sdks/python/apache_beam/transforms/ptransform.py
chadrik: interesting idea, but how do you decorate an import statement?

sdks/python/apache_beam/typehints/decorators.py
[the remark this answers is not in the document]
chadrik: practically speaking, this is less about enforcing correctness (as in "make sure that a user always passes a `WithTypeHints` instance"), and more about avoiding errors due to mypy not knowing what type it's dealing with (e.g. trying to access a method will error unless mypy is aware that the object  is a `WithTypeHints` instance).

sdks/python/apache_beam/io/iobase.py
```diff
@@ -972,14 +979,15 @@ def expand(self, pcoll):
...
   """Implements the writing of custom sinks."""
 
   def __init__(self, sink):
     super(WriteImpl, self).__init__()
     self.sink = sink
 
   def expand(self, pcoll):
+    # type: (pvalue.PCollection[InT]) -> pvalue.PCollection[None]
```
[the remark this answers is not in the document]
chadrik: the plugin only deals with the `__or__` and `__ror__` methods.

sdks/python/apache_beam/transforms/ptransform.py
udim: I don't think you can decorate an import, but this is a method.

chadrik: ah, sorry, I missed the context here.  yeah, that would be a nice feature.  would be worth bringing up at the mypy github repo.

note that this change is to accommodate analyzing user pipelines via the mypy plugin, which  I'm now thinking would be best to separate into another PR.   With some more hacking, I might ultimately be able to avoid this bit of ugliness.

chadrik:
I'm really curious to know what the general consensus is on this PR, implementation details aside.  

Do you all like the idea of adding type annotations?  

If I can get some subset of the current package passing, would you be willing to merge something like this in?  

What do you see as the major blockers?

sdks/python/apache_beam/io/gcp/datastore/v1/query_splitter.py
robertwb: Is it possible to opt certain modules out, if they become to noisy (and adding all the suppression is noisy as well)?

sdks/python/apache_beam/io/iobase.py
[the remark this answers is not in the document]
robertwb: But would it insure that the `__[r]or__` method agrees with the generic type of the PTransform which agrees with this method?

sdks/python/apache_beam/typehints/decorators.py
robertwb: (1) is already mostly there, for the user-facing API at least. How much of (2) is required for (3) and how invasive it is seems to still be an open question, which this PR is giving good insight to. (Long term, it'd also be nice to have (2) for development as well.) Given that mypy can't even do `beam.Map(lambda x: (x, 1))` I think we'll have to rely a lot on (4). 

All will be much nicer once we can have python-3 style annotations in the code itself. We can't do that for our own codebase until Python 2 is LTS only, but user code could start using it much sooner.

chadrik: > How much of (2) is required for (3) and how invasive it is seems to still be an open question, which this PR is giving good insight to. 

agreed.

>  Given that mypy can't even do beam.Map(lambda x: (x, 1)) I think we'll have to rely a lot on (4).

I think people who want to use static typing for their pipelines will have to stop using lambdas.  So there will definitely be a split between those who are willing to make that adjustment and those who want to continue relying on the current interface for dynamic typing. 

> All will be much nicer once we can have python-3 style annotations in the code itself. We can't do that for our own codebase until Python 2 is LTS only, but user code could start using it much sooner.

When we get to that stage, we can use this tool to convert comments to annotations:  https://github.com/ilevkivskyi/com2ann

chadrik:
> Beam knows how to convert from typing to its internal representation (some fixes at #9109).

Awesome.  Getting that merged in will make this PR a lot easier.

sdks/python/apache_beam/io/iobase.py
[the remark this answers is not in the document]
chadrik: yes, I believe it could be used to achieve that.

sdks/python/apache_beam/io/gcp/datastore/v1/query_splitter.py
chadrik: Yes, you can exclude errors from specific modules and packages.  That said, it's still a lot of work to get even a restricted set of modules passing, and if you restrict the set of checked modules too much then you undermine the value of type checking.  I've written a [mypy runner](https://github.com/chadrik/mypy-runner) that let's you focus on a specific set of errors, so that you can tackle problems one error _type_ at a time, rather than one file at a time.  We're using this where I work and it's been quite successful.

robertwb:
> > Beam knows how to convert from typing to its internal representation (some fixes at #9109).
> 
> Awesome. Getting that merged in will make this PR a lot easier.

It's in.

chadrik:
> It's in.

It's out :(  

I see it got reverted:  8afee953445e14988ef9be755d80fd402043e861

Did it end up causing a regression?

aaltay:
> > It's in.
> 
> It's out :(
> 
> I see it got reverted: [8afee95](https://github.com/apache/beam/commit/8afee953445e14988ef9be755d80fd402043e861)
> 
> Did it end up causing a regression?

There is a JIRA link in the revert PR: https://issues.apache.org/jira/browse/BEAM-7798 -- It was causing test failures.

udim:
Back in: https://github.com/apache/beam/pull/9125

chadrik:
I left a note on that PR about `transforms.core` not being converted yet.

Separate question:

Are the following two examples functionally equivalent in beam?

```python
coll = pipe | core.Map(lambda _, x: x.do_stuff(), self.thing)
```

```python
coll = pipe | core.Map(lambda _, x=self.thing: x.do_stuff())
```

I would hope that they are, but I could see there being some difference in how the transforms get serialized.

The reason I ask is that it's going to be tricky to deal with the shifting signatures of methods like `DoFn.process()` and `PTransform.expand()`, since mypy can be [a bit picky about this](https://github.com/python/mypy/issues/1237).  I'm trying to work out some strategies for addressing this, and this is one.

udim:
> Are the following two examples functionally equivalent in beam?
> 
> ```python
> coll = pipe | core.Map(lambda _, x: x.do_stuff(), self.thing)
> ```
> 
> ```python
> coll = pipe | core.Map(lambda _, x=self.thing: x.do_stuff())
> ```
> 
Here's my attempt at recreating what Beam does:
```
>>> import dill
>>> o = object()
>>> l = lambda x=o: x
>>> l_loaded = dill.loads(dill.dumps(l))
>>> l() == l_loaded()
False
>>> l() == l_loaded(o)
True
```
Passing self.thing as a default argument causes it to get pickled with the lambda and makes it a distinct object after unpickling.
I'm not sure thought what pickling the enclosing scope would do.

chadrik:
> Passing self.thing as a default argument causes it to get pickled with the lambda and makes it a distinct object after unpickling.

yes, any object will be a distinct object after unpickling:

```
>>> import dill
>>> o = object()
>>> o_loaded = dill.loads(dill.dumps(o))
>>> o == o_loaded
False
>>> id(o)
4402307584
>>> id(o_loaded)
4405359896
```

But this is just an implementation detail of `object`.  `object.__eq__` is likely doing an `id()` check, and this test simply demonstrates that you should not use `object` with pickle/dill if equality checks are important to your code.  

Luckily, it is possible for distinctly different objects to behave the same (otherwise beam wouldn't work at all):

```
>>> import dill
>>> o = [1, 2]
>>> o_loaded = dill.loads(dill.dumps(o))
>>> o == o_loaded
True
4402307585
>>> id(o_loaded)
4405359892
```

Anyway, IIUC, the function and the args passed to `DoFn` will _both_ be serialized using dill, so the argument object in question should undergo a serialization/deserialization transformation either way.  My question was more about the details of the beam internals, esp as it might pertain to portability, grpc model, and urns etc that I would not be familiar with as a beam newbie.

chadrik:
I'm getting very close to getting this to pass, but I'm waiting on `transforms.core` to be converted from typehints to typing.  Any updates on that?

sdks/python/apache_beam/io/gcp/pubsub.py
chadrik: fixed

robertwb:
https://github.com/apache/beam/pull/9179 awaiting your review for core.py

Also, yes, the two examples are the same, unless some.thing is a deferred side input (e.g. pvalue.AsIterable(pcoll) in which only the first is valid (but an argument could be made we should accept the second as well).

chadrik:
This PR is getting close.  It represents an enormous amount of effort, and spans a substantial part of the code base, so I'd like to start progressing towards getting this merged.  Rebasing against master has been pretty painless so far, but I'm afraid that could change soon.

I've broken the PR up into bite-sized commits to help tell a story about each set of changes.  I've also dropped the more complex changes -- the mypy plugin and generics -- and I'll deal with those in a future PR.

Other than questions and comments you might have about the content of the PR, the main issues that I need input on are primarily about style.

## Line length

It's very difficult to keep type comments to 80 characters.  Not only is there more information to describe, the type comments themselves cannot span more than one line.  This will improve when we get to python 3.6 and annotations are an official part of the syntax, since they can be defined over multiple lines just like normal python objects (which they are).  

There are a handful of practices that will minimize our line length, but even in concert they won't work 100% of the time. Here are the main ones:

- Use `from typing import Foo` vs `import typing`.  This is the single most impactful thing we can do.  It also greatly improves legibility.  Compare these two options:  
  - `typing.Optional[typing.Tuple[typing.Dict[str, str], float]]`
  - `Optional[Tuple[Dict[str, str], float]]`
- Use type aliases.  e.g.  `AwesomeType = Optional[Tuple[Dict[str, str], float]]`.  I prefer to use this sparingly, only when there's a complex type shared in many places.  Here's why:
  - quite often we can use duck-typing to reduce the requirement for certain functions. e.g. there might be a function where `Optional[Tuple[Mapping[str, str], typing.SupportsFloat]]` would do instead of `AwesomeType`.
  - I don't like to have to constantly refer to another location to see the type
- Change the way that we style functions so that they provide more room for annotations.  e.g. consider these options:
    ```python
    def really_long_function_name(arg1, # type: Optional[Tuple[Dict[str, str], float]]
                                  arg2, # type: int
                                 ):
      # type: (...) -> Tuple[str, float]
      code
    ```

    ```python
    def really_long_function_name(
      arg1, # type: Optional[Tuple[Dict[str, str], float]]
      arg2, # type: int
    ):
      # type: (...) -> Tuple[str, float]
      code
    ```
    The beam code seems to favor the former over the latter, though I see both present.  We should decide what our policy will be.  In this PR, I've determined the style on a case-by-case basis, mostly favoring  the former. 

When all else fails we can use `# pylint: disable=line-too-long` *after* the type comment.  I've added these to `apache_beam.pipeline` so you can see an example.

It's a larger conversation, but it might be worth discussing increasing the line length.  Many type-annotated projects have increased their line length to 99 or more characters.  This is a big change, that would involve a lot of debate. 

## Unused module imports

Pylint is not able to properly track type annotations used within type comments (which is the majority), and so generates spurious errors about unused imports for most of the typing classes.  Newer versions of pylint can track annotations within comments, but only for variable annotations and not for function annotations, so it's not a complete solution.  Even if we think there is a benefit to a partial solution, it will take some work to get to the newer version of pylint because it's python3-only.

The solution I'm proposing for now is to simply ignore the problem, by using the following pattern when importing types:

```
# pylint: disable=unused-import
from typing import TYPE_CHECKING
from typing import Any
from typing import Callable
from typing import Container
from typing import DefaultDict
from typing import Dict
from typing import Iterable
from typing import Iterator
from typing import List
# pylint: enable=unused-import
```

That will leave it up to developers to get right for now, and when we get to a pure python3 code-base we can fix anything that's gone astray.

---

Let me know what you think.  I'm excited to get this into the code base.  If you open this branch up in an IDE that is properly setup for type annotations, you'll see that the annotations are inspectable and greatly improve your ability to comprehend and traverse the code.  They take a bit to get used to, but even the detractors on my team are now converts.   Overall, adding typing to the Beam code base was actually very straight-forward, relatively speaking, because the code is already well structured and implicitly precise about the types that are expected (in other words, I discovered there are very few `Union` types required).  That said, I found a number of docstrings that were wrong about the types expected (I still need to go back and correct a bunch), so that's the beauty of type annotations:  if you get it wrong you'll get an error!

chadrik:
I created an issue for the pylint bug: https://github.com/PyCQA/pylint/issues/3112

I don't expect a quick resolution.

chadrik:

chadrik:
run seed job

chadrik:
Run PythonLint PreCommit

chadrik:
Run PythonLint PreCommit

chadrik:
Run PythonLint PreCommit

chadrik:
Ok, I have what I think are satisfactory answers to all of the problems that I encountered except for the 4 issues below that I need input on. 

@robertwb can you help me find the answers to these, please!

---

```
apache_beam/io/iobase.py:924: error: "SourceBase" has no attribute "coder"  [attr-defined]
```

I can't find any sub-classes of `SourceBase` that have a coder attribute.  Is this safe to remove or is this a dataflow thing?

---

```
apache_beam/runners/worker/statesampler_slow.py:77: error: "StateSampler" has no attribute "_states_by_name"  [attr-defined]
```

`statesampler_slow.StateSampler` does not have `_states_by_name` attribute, but `statesampler.StateSampler` does.  I could add this attribute to `statesampler_slow.StateSampler`, but I don't think it would be used. The more straightforward solution may be to edit `statesampler_slow.StateSampler.reset()` to do nothing.  Right now I think it would error if it ran.

---
```
apache_beam/runners/portability/fn_api_runner_transforms.py:280: error: Invalid index type "Optional[str]" for "MutableMapping[str, Environment]"; expected type "str"  [index]
```

It's unclear to me whether `Stage.environment` is meant to be `str` or `Optional[str]`.  The way that it's initialized it _could_ be `Optional[str]`:

```python
class Stage(object):
  """A set of Transforms that can be sent to the worker for processing."""
  def __init__(self,
               name,  # type: str
               transforms,  # type: List[beam_runner_api_pb2.PTransform]
               downstream_side_inputs=None,  # type: Optional[FrozenSet[str]]
               must_follow=frozenset(),  # type: FrozenSet[Stage]
               parent=None,  # type: Optional[Stage]
               environment=None,  # type: Optional[str]
               forced_root=False
              ):
    ...

    if environment is None:
      environment = functools.reduce(
          self._merge_environments,
          (self._extract_environment(t) for t in transforms))
    self.environment = environment

  ...

  @staticmethod
  def _extract_environment(transform):
    # type: (beam_runner_api_pb2.PTransform) -> Optional[str]
    if transform.spec.urn in PAR_DO_URNS:
      pardo_payload = proto_utils.parse_Bytes(
          transform.spec.payload, beam_runner_api_pb2.ParDoPayload)
      return pardo_payload.do_fn.environment_id
    elif transform.spec.urn in COMBINE_URNS:
      combine_payload = proto_utils.parse_Bytes(
          transform.spec.payload, beam_runner_api_pb2.CombinePayload)
      return combine_payload.combine_fn.environment_id
    else:
      return None
```

In practice will there will always be at least one ParDo or Combine per stage?  If so we should be asserting that `self.environment is not None` in `Stage.__init__`.  

Alternately, we could assert that `self.environment is not None` just before this call in `executable_stage_transform`.  

The bottom line is that there are currently no guarantees in the code that `self.environment` is not None at this point, and if it is, it will be an error.

---

```
apache_beam/runners/portability/fn_api_runner.py:933: error: "ParallelBundleManager" has no attribute "_skip_registration"  [attr-defined]
```

I can't find anywhere in the code that refers to `_skip_registration`.  Is this safe to remove?

robertwb:
> ```
> apache_beam/io/iobase.py:924: error: "SourceBase" has no attribute "coder"  [attr-defined]
> ```
> 
> I can't find any sub-classes of `SourceBase` that have a coder attribute. Is this safe to remove or is this a dataflow thing?

Good question. I think this is used by some Dataflow-only subclasses of NativeSource. Go ahead and add a coder @property there raising a NotImplementedException to suppress the warning. Hopefully we can just get rid of NativeSource soon. 

> ```
> apache_beam/runners/worker/statesampler_slow.py:77: error: "StateSampler" has no attribute "_states_by_name"  [attr-defined]
> ```
> 
> `statesampler_slow.StateSampler` does not have `_states_by_name` attribute, but `statesampler.StateSampler` does. I could add this attribute to `statesampler_slow.StateSampler`, but I don't think it would be used. The more straightforward solution may be to edit `statesampler_slow.StateSampler.reset()` to do nothing. Right now I think it would error if it ran.

Yes, just delete this for now. I'm surprised it's never called. 

> ```
> apache_beam/runners/portability/fn_api_runner_transforms.py:280: error: Invalid index type "Optional[str]" for "MutableMapping[str, Environment]"; expected type "str"  [index]
> ```
> 
> It's unclear to me whether `Stage.environment` is meant to be `str` or `Optional[str]`. The way that it's initialized it _could_ be `Optional[str]`:
> ...
> In practice will there will always be at least one ParDo or Combine per stage? If so we should be asserting that `self.environment is not None` in `Stage.__init__`.

In general, this should be the case, but I can't rule out, for example, a Flatten or similar becoming isolated into its own stage, even after fusion. I think we have to assume it can be None when stages are initialized. 

> Alternately, we could assert that `self.environment is not None` just before this call in `executable_stage_transform`.
> 
> The bottom line is that there are currently no guarantees in the code that `self.environment` is not None at this point, and if it is, it will be an error.

I'm fine with an assert there, though it seems noisy compared to the error that we'd just be getting at the next point. 

> ```
> apache_beam/runners/portability/fn_api_runner.py:933: error: "ParallelBundleManager" has no attribute "_skip_registration"  [attr-defined]
> ```
> 
> I can't find anywhere in the code that refers to `_skip_registration`. Is this safe to remove?

Hmm... Looks like this is an alias for registered that got mixed up in the multi-process refactor. https://github.com/apache/beam/blob/8d2997f8d7ad84649b8ecb2f7e2ca2eceb91b6d0/sdks/python/apache_beam/runners/portability/fn_api_runner.py#L1909 Can you file a JIRA to follow up on this?

chadrik:
Thanks for the feedback.

> I'm fine with an assert there, though it seems noisy compared to the error that we'd just be getting at the next point.

Asserts are a pretty common pattern with typing.  It's a way to acknowledge "this _could_ fail here given the type constraints".  Otherwise, mypy doesn't know what errors we're willing to tolerate.  The alternative is to add a `# type: ignore` comment but the assert provides a way to be more precise (both for those reading the code and receiving a potential error at runtime) and also affect all of the code below it at that scope (otherwise it could require many `# type: ignore` comments).  

> Can you file a JIRA to follow up on this?

Will do.

chadrik:
> Can you file a JIRA to follow up on this?

https://issues.apache.org/jira/browse/BEAM-8486

I put a fix for this in this PR, because I was curious to see if the tests would pass.  Even if they do, do you think I should just ignore the problem in this PR (with a TODO referencing the Jira) until we can do a deeper investigation?

robertwb:
OK, an assert is fine. And, yes, ignore/suppress the registration issue for this CL with a TODO(JIRA) for future follow-up.

chadrik:
Run Python PreCommit

chadrik:
Run Python PreCommit

chadrik:

chadrik:

chadrik:
Alright everyone, I'm happy with where this PR is now, and I think it's time to start figuring out how to get it merged.

This PR is admittedly huge, but the daunting scope is mitigated by the fact that the intent of the PR is to _avoid making any behavioral changes_.  I recommend reviewing this PR one commit at a time, as I've tried to organize it into bite-sized chunks, and I've segregated changes with possible unintentional runtime side-effects and made note of them for extra scrutiny.  

An additional consideration with respect to the size of this PR is that there is a critical mass required before mypy can be enabled in CI, so if we were to break up this PR, we wouldn't be able to enable the mypy lint until all of the changes are finally merged (the alternative is hundreds `# type: ignore` comments throughout the code, which I'd rather not do).  Once this PR goes in, it'll be easy to make smaller follow-ups. 

I'm really excited about this PR and I think you all are going to find the code base much easier to navigate and grok once it's in (I already provide this branch to others on the team who are trying to learn beam), not to mention the deep structural analysis that you get during testing (it's helped me to identify a number of real and potential bugs already, which you'll see in my commits).  

One final consideration: there _will_ be a need for training and mentoring on static type analysis, I have some internal docs on the subject which I'm happy to share, and you can always drag me into a PR to help resolve issues.

TheNeuralBit:
Thanks for doing this @chadrik! I'm very excited about this change :)

One quick question: when we switch to python 3, is there a tool that will convert all the type comments to pep 484 type hints?

robertwb:
Thanks for the huge amount of work that's gone into this PR, and I'm looking forwards to the benefits it'll bring to the codebase. 

In terms of getting this reviewed, I understand that mypy can't be enabled [1] until it's all in, but I think it's preferable to get individual commits merged and in incrementally than review commits incrementally and then try to get them all in at once in a single PR. In my spot checking it seems these commits are relatively orthogonal--is that basically the case? Perhaps a reviewer could create PRs cherry-picking commits that they are able to review in one sitting and we could get them in that way? Any other ideas? 

[1] Maybe we could enable it, but not make it a blocker, in which case we could track progress and at least manually track progress/slippage until we enforce it.

chadrik:
> when we switch to python 3, is there a tool that will convert all the type comments to pep 484 type hints?

Yes.  https://github.com/ilevkivskyi/com2ann.  I think we will need to manually fixup a bunch of style issues after running that, because when the comments become annotations they will exceed the line limit.   

Side note: I recently experimented with enabling pylint's auto-fix features (like spotlessApply), but it won't work unless we're willing to change Beams' rather "unique" style.  IIRC, the main deal-breaker is Beams' mixture of 2-space and 4-space tabs depending on context (it's been a headache for me to setup properly in PyCharm as well.  I have something that's close, but not perfect).

> In terms of getting this reviewed, I understand that mypy can't be enabled [1] until it's all in, but I think it's preferable to get individual commits merged and in incrementally than review commits incrementally and then try to get them all in at once in a single PR. In my spot checking it seems these commits are relatively orthogonal--is that basically the case? Perhaps a reviewer could create PRs cherry-picking commits that they are able to review in one sitting and we could get them in that way? Any other ideas?

I like the idea of adding the lint job but not failing until we've gotten everything in. 

There are some commits that depend on others, but the relationship is not always obvious, even to me.  

How about this plan: we first merge in a PR that contains just type comments, so theoretically no chance of runtime changes.  This will be ignore comments, base typing pass, and the mypy lint job with failure disabled  (basically this PR through to https://github.com/apache/beam/pull/9056/commits/c25792c83785bce1ed4b01dd531ee1404b899438).  At this point I think it would be safe for reviewers to cherry-pick the more topical commits that remain, in any order they choose.

robertwb:
Sounds good. That's still quite the big commit, but should be straightforward to review. (I'm not seeing that commit in the list though.)

robertwb:
We also need to get buy-in on the list as this will be yet another "lint" that contributors will have to make happy (though generally a good thing, but it might require some education).

chadrik:
First PR is here: https://github.com/apache/beam/pull/9915

chadrik:
Type annotations gained syntax support in 2 passes:  function annotations
in 3.5 and variable annotations in 3.6.  So yes, once we get to python 3.6
there will no longer be a need for type comments.  Luckly, there is a tool
for automatically converting comments to annotations.

sdks/python/build-requirements.txt
```diff
@@ -0,0 +1,20 @@
...
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+# TODO(BEAM-5414): latest grpcio-tools incompatible with latest protobuf 3.6.1.
+grpcio-tools>=1.3.5,<=1.14.2
+mypy-protobuf==1.12
+future==0.16.0
```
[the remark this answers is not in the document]
chadrik: > This file needs a description. Which dependencies should be included here?

Good point.  These are the requirements for building an sdist for beam.  They happen to all be dependencies of `gen_protos.py`, but conceptually this file could contain other build deps like cython.  I  chose to omit that since it's a bit more complicated since  it's only used on linux and only for certain tests.

> Why is future listed here? It's already in setup.py. 

`gen_protos.py` uses it to futurize the results of `protoc`.  The  packages in `install_requires` are not available at the time that `gen_protos` runs. 

> Why is mypy-protobuf here? Only the mypy command from setup.py seems to need it, so it could be listed under REQUIRED_TEST_PACKAGES there.

Also required for `gen_protos`

buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy
```diff
@@ -1768,7 +1768,7 @@ class BeamModulePlugin implements Plugin<Project> {
           project.exec { commandLine virtualenvCmd }
           project.exec {
             executable 'sh'
-            args '-c', ". ${project.ext.envdir}/bin/activate && pip install --retries 10 --upgrade tox==3.11.1 grpcio-tools==1.3.5"
+            args '-c', ". ${project.ext.envdir}/bin/activate && pip install --retries 10 --upgrade tox==3.11.1 -r ${project.rootDir}/sdks/python/build-requirements.txt"
```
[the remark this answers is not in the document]
chadrik: I touched on this in your other comment, but the big picture here is that python lacks a *good solution for specifying build-time requirements:  i.e. deps required to run `setup.py`.  

Entire ecosystems have been built up around solving this: 

- https://poetry.eustace.io/
- https://snarky.ca/clarifying-pep-518/
- https://github.com/pypa/pip/issues/2381

One note from the last link:

> Pip 10 and later support build dependencies, but they have to be specified in a pyproject.toml file.

To use that we'd have to change `python setup.py sdist` to use `pip`, which makes me nervous.  Certainly more than I'd want to change in this PR.

@robertwb may have some other input.   I noticed from some comments there was an earlier attempt to use the `setup_requires` field of `setup.py`, and then it was dismantled.  IIRC, `setup_requires` has a lot of crippling issues.

In my experience a `build-requirements.txt` is by far the simplest solution.

sdks/python/tox.ini
```diff
@@ -160,16 +160,17 @@ commands =
 [testenv:py27-lint]
 # Checks for py2 syntax errors
 deps =
+  -r build-requirements.txt
```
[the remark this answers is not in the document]
chadrik: This is required if we want `tox` to work on its own, and I really really do.  Using gradle to run tox tasks adds a lot of overhead and I kind of hate it.  gradle creates a complete virtualenv just to build an sdist, then it creates another virtualenv to run tox (and IIRC it does this every time).  If I run tox directly it creates a single virtualenv for both sdist and the tox env and it reuses that virtualenv between runs.  Gradle also just seems slow to start up.  Also, I can't pass individual tests to run when using gradle.  Let's just say that as a python developer, the gradle experience is the thing I like least about the Beam project.

Yes, by adding the gen_proto requirements to `build-requirements.txt` and installing that in  the `setupVirtualenv` task I'm making that overhead worse, but I certainly didn't start that problem:  the code is currently installing `tox` every time that the `sdist` task is run, which is pointless.  In other words, `setupVirtualenv` is currently trying to be the one virtualenv used by all tasks, but different tasks have different requirements.  

I could add a feature to `setupVirtualenv` to provide a requirements file, which would let us cut out `tox` from the `sdist` task which would probably end up in a net overall win, because tox has a lot of deps.  Should I do that here or in a followup PR?

I've been meaning to bring up some of these issues on the list.  I assume the reason for the two step test process (sdist + tox, each with its own virtualenv) is to ensure that tox is using an sdist built in the same way every time, using python2, just as it would be for distribution?

sdks/python/gen_protos.py
```diff
@@ -49,6 +49,164 @@
 ]
 
 
+def generate_urn_files(log, out_dir):
+  """
+  Create a pyi stub for common_urns
```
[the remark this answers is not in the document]
chadrik: good catch!

sdks/python/.pylintrc
```diff
@@ -143,6 +143,7 @@ disable =
   unnecessary-lambda,
   unnecessary-pass,
   unneeded-not,
+  unsubscriptable-object,
```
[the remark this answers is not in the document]
chadrik:
```python
class _ListBuffer(List[bytes]):
  """Used to support parititioning of a list."""
  def partition(self, n):
    # type: (int) -> List[List[bytes]]
    return [self[k::n] for k in range(n)]    # <------ ERROR HERE
```

```python
  def progress_metrics(self):
    # type: () -> beam_fn_api_pb2.Metrics.PTransform
    with self.lock:
      metrics = super(SdfProcessSizedElements, self).progress_metrics()
      current_element_progress = self.current_element_progress()
    if current_element_progress:
      assert self.input_info is not None
      metrics.active_elements.measured.input_element_counts[
          self.input_info[1]] = 1    # <------ ERROR HERE
      metrics.active_elements.fraction_remaining = (
          current_element_progress.fraction_remaining)
    return metrics
```

Generally speaking, there are a handful of checks that overlap between mypy and pylint where pylint does a worse job.  In the second example above `self.input_info` is an `Optional[Tuple]` that's initialized to `None`, so it thinks it's unsubscriptable, but it's initialized to a tuple elsewhere and there's a line just above the error that says `assert self.input_info is not None`.  These are all subtleties that mypy understands but pylint does not.

> Perhaps a newer version of pylint could help?

I'm already using the latest and greatest (one of the commits here ticks the version up from 2.4.2 to 2.4.3)

sdks/python/build-requirements.txt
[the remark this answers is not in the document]
chadrik: > Why is mypy-protobuf here?

To clarify, `mypy-protobuf` is required at the time that `gen_protos` runs (it's used by `protoc`), but it's not technically required by anything other than `py37-mypy` job.  However, we'll want to include the .pyi files in sdist so that end users can get their benefits in PyCharm, mypy, etc.  Which reminds me that I need to update setup.py to include pyi files in `package_data`!

robertwb:
> Yeah, it's a bit pedantic. By default mypy assigns `Any` to an unannotated type, but in the case of `__init__` nothing would break or error if it were omitted. However, it's good for people to get into the habit of fully annotating their functions so that eventually we can reach complete coverage by enabling `--disallow-untyped-defs`:
> 
> ```
>   --disallow-untyped-defs   Disallow defining functions without type
>                             annotations or with incomplete type annotations
>                             (inverse: --allow-untyped-defs)
> ```

Hmm... Especially once we move to Python 3, writing the return type part of `__init__(arg: str, arg2: iint) -> None` seems like we're crossing the point from types being helpful to types being noise. One of the main points of gradual typing is that they can be added *where useful* and omitted where not, and I am firmly in the camp that unneeded symbols in your code are a net negative. But possibly such calls are premature at this point.

chadrik:
> Hmm... Especially once we move to Python 3, writing the return type part of __init__(arg: str, arg2: iint) -> None seems like we're crossing the point from types being helpful to types being noise.

Enabling `--disallow-untyped-defs` is always something we can choose _not_ to do, if we don't feel that the pros outweigh the cons.

sdks/python/tox.ini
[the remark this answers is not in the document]
aaltay: `tox` runs outside of gradle. (At least for me). What is the issue there?

> I assume the reason for the two step test process (sdist + tox, each with its own virtualenv) is to ensure that tox is using an sdist built in the same way every time, using python2, just as it would be for distribution?

sdist step is used for other things (for building artifacts for release). That is the reason for it to exists on its own.

tox, I do not believe we spent time on configuring it to reuse existing virtualenv. If that is an option we can try that.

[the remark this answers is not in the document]
chadrik: > tox runs outside of gradle. (At least for me). What is the issue there?

When tox is run on jenkins, it is invoked via gradle.  The tox task is defined in BeamModulePlugin.groovy like this:

```groovy
      project.ext.toxTask = { name, tox_env ->
        project.tasks.create(name) {
          dependsOn 'setupVirtualenv'
          dependsOn ':sdks:python:sdist'

          doLast {
            // Python source directory is also tox execution workspace, We want
            // to isolate them per tox suite to avoid conflict when running
            // multiple tox suites in parallel.
            project.copy { from project.pythonSdkDeps; into copiedSrcRoot }

            def copiedPyRoot = "${copiedSrcRoot}/sdks/python"
            def distTarBall = "${pythonRootDir}/build/apache-beam.tar.gz"
            project.exec {
              executable 'sh'
              args '-c', ". ${project.ext.envdir}/bin/activate && cd ${copiedPyRoot} && scripts/run_tox.sh $tox_env $distTarBall"
            }
          }
          inputs.files project.pythonSdkDeps
          outputs.files project.fileTree(dir: "${pythonRootDir}/target/.tox/${tox_env}/log/")
        }
      }
```

and it's used here:

https://github.com/apache/beam/blob/master/sdks/python/test-suites/tox/py2/build.gradle

> sdist step is used for other things (for building artifacts for release). That is the reason for it to exists on its own.

I'm not saying that `sdist` task shouldn't exist, I'm saying it adds overhead to use the `sdist` task to generate the tarball for the tox task, instead of just letting tox generate its own sdist tarball, which it is designed to do and is the default behavior (Also, there's the copying of source files which adds more overhead and it's unclear why this exists since tox is designed to create isolated virtualenvs per task environment).

In the definition of `toxTask` above, you see that it depends on `setupVirtualenv` and also on `sdist`, and `sdist` itself depends on `setupVirtualenv`.   That's why we end up running  `setupVirtualenv` twice for each tox task run, and then tox creates yet another virtualenv to actually run the task.  That's 3 virtualenvs per tox job!   I've used travisCI to setup a number of tox-based CI jobs and the typical workflow is to ensure that tox is installed *on the worker image* and then theres only one virtualenv created per job, the one created by tox. 

Anyway, long story short the `build-requirements.txt` line is required for the tox tasks to work.

chadrik:
Hello reviewers,
Thanks for the feedback so far!  There are a few questions that I've responded to that are now in your court.  I'd love to get these addressed soon so that I can minimize the merge conflicts that are piling up. 

Once I have those answers your feedback will be integrated into the smaller "part 1" PR here: https://github.com/apache/beam/pull/9915

thanks!

sdks/python/tox.ini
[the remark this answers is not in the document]
udim: I agree that the gradle config is not the most efficient. The base requirements for running Python tests on gradle are pip, virtualenv, and 4 versions of the python interpreter. Tox is installed by setupVirtualenv. :sdks:python:sdist is run only once (IIUC) for all toxTasks, since running more than one sdist concurrently is racey and makes our tests flakey.
CC: @markflyhigh who set this up.

Thinking about this now, I have a hunch that the tarball in use doesn't ever include Cythonized sources (since they are created by sdist when the cython package is preset).

Source files are copied due to isolation requirements. The Cythonization process adds files which are automatically picked up if present. Perhaps a better way to do this is to exclude C files from sdist for the non-cython envs, but this is the way we do it currently.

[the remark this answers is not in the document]
chadrik: So we agree there's room for improvement.  The question for the moment is, is my current solution ok?

chadrik:
Keeping this as the master PR for typing.  #10367 is currently under review.

PR #9077, opened 2019-07-16
[BEAM-7499] Workaround for tricky Reify testing issue

pabloem:
r: @ttanay 

Hi Tanay! I've seen issues with this test in some environments that are fixed by doing this. My guess is that there's some odd codepath where Create(...)'d elements need to go through a ptransform before having all their attributes (timestamp, window) populated properly?

I'm not making any sense, but please review this change : )

### DIFF
--- a/sdks/python/apache_beam/transforms/util_test.py
+++ b/sdks/python/apache_beam/transforms/util_test.py
@@ -575,6 +575,12 @@ class ReifyTest(unittest.TestCase):
                                   [GlobalWindow()])]
     with TestPipeline() as p:
       pc = p | beam.Create(l)
+      # Map(lambda x: x) PTransform is added after Create here, because when
+      # a PCollection of WindowedValues is created with Create PTransform,
+      # the windows are not assigned to it. Adding a Map forces the
+      # PCollection to go through a DoFn so that the PCollection consists of
+      # the elements with timestamps assigned to them instead of a PCollection
+      # of WindowedValue(element, timestamp, window).
       pc = pc | beam.Map(lambda x: x)
       reified_pc = pc | util.Reify.Window()
       assert_that(reified_pc, equal_to(expected), reify_windows=True)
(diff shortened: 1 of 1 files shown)


Ardagan:
I suggest we actually disable test or look for root cause of this. I'd avoid such workaround since it hides problem.

pabloem:
The test only fails when testing internally at Google with blaze - I have not been able to reproduce the issue in any other environment (jenkins, my local machine with a virtualenv, etc).

I don't want to deactivate a test that only fails internally at Google : /

sdks/python/apache_beam/transforms/util_test.py
```diff
@@ -575,6 +575,7 @@ def test_window(self):
                                   [GlobalWindow()])]
     with TestPipeline() as p:
       pc = p | beam.Create(l)
+      pc = pc | beam.Map(lambda x: x)
```
ttanay: Hi @pabloem 
There is another test(`test_window_in_value`) that creates a PCollection from `WindowedValue`s.
It is used there so that typehint can be evaluated properly. Using the hack enforces it to extract the KV from the WindowedValue.

Can you please add a comment here mentioning that the hack's use here, is different from its use in the other test?

pabloem:
Run Python PreCommit

sdks/python/apache_beam/transforms/util_test.py
ttanay: Otherwise, LGTM :)

pabloem:
Thanks @ttanay

pabloem:
@Ardagan it seems that what happens here is that `beam.Create` does not attribute the windows and timestamps to the elements - these are only added after they go through a ptransform (in this case the no-op Map).

Ardagan:
> @Ardagan it seems that what happens here is that `beam.Create` does not attribute the windows and timestamps to the elements - these are only added after they go through a ptransform (in this case the no-op Map).

Should we consider this a bug? This looks really ambiguous to me.

pabloem:
Run Python PreCommit

pabloem:
I've filed https://jira.apache.org/jira/browse/BEAM-7759
It contains a discussion of what is the problem, and I'll add suggestions to fix it.

akedin:
run python precommit

pabloem:
Run Python PreCommit

pabloem:
Run Python PreCommit

PR #9082, opened 2019-07-16
Add UT to check SdfWrapper overrides ReadFromText

boyuanzz:
Add one more unit test case to verify SdfBoundedSourceWrapper also overrides. And set "beam_fn_api" when initializing a pipeline if using FnApiRunner.

### DIFF
--- a/sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
+++ b/sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
@@ -39,8 +39,10 @@ from tenacity import retry
 from tenacity import stop_after_attempt
 
 import apache_beam as beam
+from apache_beam import pvalue
 from apache_beam.io import restriction_trackers
 from apache_beam.io.concat_source_test import RangeSource
+from apache_beam.io.textio_test import write_data
 from apache_beam.metrics import monitoring_infos
 from apache_beam.metrics.execution import MetricKey
 from apache_beam.metrics.execution import MetricsEnvironment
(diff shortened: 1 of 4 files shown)


boyuanzz:
Run Python PreCommit

boyuanzz:
Run Python PreCommit

boyuanzz:
Run Python PreCommit

tvalentyn:
Gentle reminder @robertwb and @pabloem

sdks/python/apache_beam/runners/portability/flink_runner_test.py
```diff
@@ -262,6 +262,12 @@ def test_sdf_with_sdf_initiated_checkpointing(self):
     def test_sdf_synthetic_source(self):
       raise unittest.SkipTest("BEAM-2939")
 
+    def test_sdf_wrapper_overrides_read(self):
```
robertwb: Perhaps these tests should simply be moved out of this class (which has many subclasses) rather than skipped here.

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
```diff
@@ -519,33 +521,46 @@ def process(
...
 
   @mock.patch('apache_beam.io.iobase._SDFBoundedSourceWrapper.expand')
   def test_sdf_wrapper_overrides_read(self, sdf_wrapper_mock_expand):
     def _fake_wrapper_expand(pbegin):
+      assert isinstance(pbegin, pvalue.PBegin)
+      return (pbegin
+              | beam.Create(['1']))
+
+    sdf_wrapper_mock_expand.side_effect = _fake_wrapper_expand
+    self._run_sdf_wrapper_pipeline(beam.io.Read(RangeSource(0, 4)), ['1'])
```
robertwb: This test looks really wrong... if we're going to use a mock perhaps have the element be 'mocked_data' or something obvious like that.

boyuanzz: Thanks for helping review this. I agree that the data should be `mockd_data`. Beside this, anything else looks wrong?

sdks/python/apache_beam/runners/portability/flink_runner_test.py
boyuanzz: If flink supports SDF and want to use SdfBoundedSourceWrapper, flink runner is expected to replace Read transforms with SdfBoundedSourceWrapper.

boyuanzz:
> Could you clarify what coverage was missing that this is hoping to cover?

These 2 tests are testing with the flag 'use_sdf_bounded_source' and  'beam_fn_api' enable, the actual `Read` / 'ReadFromText' transform is replaced by 'SdfBoundedSourceWrapper' properly. This replacement happens at specific runner level. Currently the replacement is supported by FnApiRunner and DataflowRunner.

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
robertwb: No, just that.

robertwb:
OK, if these are simply testing that the substitution happens, let's move them out of this test class and into another that is not subclassed by a bunch of other classes to ensure that they implement the spec correctly. (In a sense, this class is all validates runner tests.) 

IMHO, we should also be doing this replacement ubiquitously, not require every runner to do so.

boyuanzz:
Run Python PreCommit

boyuanzz:
Passing Python PreCommit is blocked by https://issues.apache.org/jira/browse/BEAM-8150

pabloem:
I guess we're still blocked on BEAM-8150?

boyuanzz:
> I guess we're still blocked on BEAM-8150?

Yes.

PR #9089, opened 2019-07-17
[BEAM-7755] adding support for nested rows and arrays in BigQuery to Beam row conversion

snallapa:
Example implementation of repeated record conversion from BigQuery to beam row through avro. Sorry this is my first contribution so I am a bit new to this. I wanted to put this up as this worked on a test table with a REPEATED RECORD that I tried on and maybe just put it on the radar? 
R: @lukecwik
R: @chamikaramj

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryUtils.java
+++ b/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryUtils.java
@@ -314,7 +314,15 @@ public class BigQueryUtils {
   public static Row toBeamRow(GenericRecord record, Schema schema, ConversionOptions options) {
     List<Object> valuesInOrder =
         schema.getFields().stream()
-            .map(field -> convertAvroFormat(field.getType(), record.get(field.getName()), options))
+            .map(
+                field -> {
+                  try {
+                    return convertAvroFormat(field.getType(), record.get(field.getName()), options);
+                  } catch (Exception cause) {
+                    throw new IllegalArgumentException(
+                        "Error converting field " + field + ": " + cause.getMessage(), cause);
+                  }
+                })
             .collect(toList());
 
     return Row.withSchema(schema).addValues(valuesInOrder).build();
--- a/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryUtils.java
+++ b/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryUtils.java
@@ -505,7 +505,7 @@ public class BigQueryUtils {
       case STRING:
         return convertAvroPrimitiveTypes(beamFieldTypeName, avroValue);
       case ARRAY:
-        return convertAvroArray(beamField, avroValue);
+        return convertAvroArray(beamField, avroValue, options);
       case LOGICAL_TYPE:
         String identifier = beamField.getType().getLogicalType().getIdentifier();
         if (SQL_DATE_TIME_TYPES.contains(identifier)) {
@@ -553,14 +553,19 @@ public class BigQueryUtils {
     return new Instant((long) value / 1000);
   }
 
-  private static Object convertAvroArray(Field beamField, Object value) {
+  private static Object convertAvroArray(Field beamField, Object value, BigQueryUtils.ConversionOptions options) {
     // Check whether the type of array element is equal.
     List<Object> values = (List<Object>) value;
     List<Object> ret = new ArrayList();
     for (Object v : values) {
-      ret.add(
-          convertAvroPrimitiveTypes(
-              beamField.getType().getCollectionElementType().getTypeName(), v));
+      FieldType arrayElementType = beamField.getType().getCollectionElementType();
+      TypeName elementTypeName = arrayElementType.getName();
+      if (elementTypeName.equals(TypeName.ROW)) {
+        GenericData.Record record = (GenericData.Record) v;
+        ret.add(toBeamRow(record, arrayElementType.getRowSchema(), options))
+      } else {
+        ret.add(convertAvroPrimitiveTypes(elementTypeName, v));
+      }
     }
     return (Object) ret;
   }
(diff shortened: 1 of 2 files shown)


amaliujia:
Could you add a test, which will better illustrate what this change is about?

I think you are trying to handle the case of `ROW<ROW<ROW<..>>>`

snallapa:
@amaliujia hey I added a test. I will clean it up later but I wanted to have it done to show you the problem. It is the case you are talking about and that test fails without the change I made. I also want to point out that all the tests in that file there use a certain toBeamRow method and the one I found this issue on did not have tests for. Is the toBeamRow method that goes through Avro not recommended?

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryUtils.java
```diff
@@ -553,14 +554,20 @@ private static ReadableInstant truncateToMillis(Object value) {
...
+  private static Object convertAvroArray(
+      Field beamField, Object value, BigQueryUtils.ConversionOptions options) {
     // Check whether the type of array element is equal.
     List<Object> values = (List<Object>) value;
     List<Object> ret = new ArrayList();
     for (Object v : values) {
-      ret.add(
-          convertAvroPrimitiveTypes(
-              beamField.getType().getCollectionElementType().getTypeName(), v));
+      FieldType arrayElementType = beamField.getType().getCollectionElementType();
```
kennknowles: This, and a little more below, can move out of the loop. Actually you can move it out of the method entirely, I think. You don't need to pass the whole `Field` but it should only depend on the type, right?

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryUtils.java
```diff
@@ -553,14 +554,20 @@ private static ReadableInstant truncateToMillis(Object value) {
...
+      TypeName elementTypeName = arrayElementType.getTypeName();
+      if (elementTypeName.equals(TypeName.ROW)) {
+        GenericData.Record record = (GenericData.Record) v;
+        ret.add(toBeamRow(record, arrayElementType.getRowSchema(), options));
+      } else {
+        ret.add(convertAvroPrimitiveTypes(elementTypeName, v));
```
kennknowles: I think this `if` / `else` piece should just be a recursive call into `convertAvroFormat`. You will need to update it to only need the `FieldType` and not the whole `Field`. This should be easy, I think. I just read the code and it does not seem to actually need the whole `Field`.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryUtils.java
amaliujia: +1

values actually share the same type because it is an ARRAY. At least collection element type can be gotten out of loop

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryUtils.java
amaliujia: big +1.

Using `convertAvroFormat` will make sure all other types (not limited to ROW or primitive) can be handled. It that case this is changed to handle `ARRAY<T>`, not limited to `ARRAY<ROW> or ARRAY<primitive type>`

sdks/java/io/google-cloud-platform/src/test/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryUtilsTest.java
```diff
@@ -422,4 +424,30 @@ public void testToBeamRow_array_row() {
     Row beamRow = BigQueryUtils.toBeamRow(ARRAY_ROW_TYPE, BQ_ARRAY_ROW_ROW);
     assertEquals(ARRAY_ROW_ROW, beamRow);
   }
+
+  @Test
+  public void testToBeamRow_repeated_row() {
```
amaliujia: Thanks for adding this test. It's much more clear now on the change. So this change extends a general support for `ARRAY<T>`.

Optional, it would be great if you could also test `ARRAY<ARRAY<T>>`

snallapa:
Thanks to both of you for the feedback! I updated my branch with some changes and I added one more test as well.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryUtils.java
snallapa: The only spot that needed the entire Field was using its name, so unfortunately if we encounter an error we can't inform the user what Field name caused that error (but we can at least tell them the type)

sdks/java/io/google-cloud-platform/src/test/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryUtilsTest.java
snallapa: Great idea! I added that test as well

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryUtils.java
```diff
@@ -313,7 +314,7 @@ public TableRow apply(T input) {
   public static Row toBeamRow(GenericRecord record, Schema schema, ConversionOptions options) {
     List<Object> valuesInOrder =
         schema.getFields().stream()
-            .map(field -> convertAvroFormat(field, record.get(field.getName()), options))
+            .map(field -> convertAvroFormat(field.getType(), record.get(field.getName()), options))
```
kennknowles: Here is where you can give an error message with the field name.

```java
  .map(field -> {
     try {
       return convertAvroFormat(field.getType(), record.get(field.getName()), options))
     } catch (Exception cause) {
       throw new IllegalArgumentException("Error converting field " + field + ": " cause.getMessage(), cause)
     }
   })
```

snallapa: ah that makes sense. thank you I added that as well!

snallapa:
thanks!! @kennknowles

PR #9090, opened 2019-07-17
[BEAM -7741] Implement SetState for Python SDK

rakeshcusat:
SetState was implemented in Java sdk, to ensure the feature parity SetState is also implemented in Python SDK.

------------------------

--- | --- | --- | --- | ---


rakeshcusat:
R: @robertwb @pabloem

rakeshcusat:
Run Portable_Python PreCommit

rakeshcusat:
R: @angoenka

rakeshcusat:

sdks/python/apache_beam/transforms/userstate_test.py
```diff
@@ -415,6 +422,82 @@ def clear_values(self, bag_state=beam.DoFn.StateParam(BAG_STATE)):
...
 
+  def test_simple_set_stateful_dofn(self):
+    class SimpleTestSetStatefulDoFn(DoFn):
+      BUFFER_STATE = SetStateSpec('buffer', BytesCoder())
+      EXPIRY_TIMER = TimerSpec('expiry', TimeDomain.WATERMARK)
+
+      def process(self, element, buffer=DoFn.StateParam(BUFFER_STATE),
+                  timer1=DoFn.TimerParam(EXPIRY_TIMER)):
+        unused_key, value = element
+        buffer.add(str(value).encode('latin1'))
```
robertwb: Can we just use StringCoder and avoid the explicit encoding? Or if the input is integers, just store the int itself?

sdks/python/apache_beam/transforms/userstate_test.py
```diff
@@ -415,6 +422,82 @@ def clear_values(self, bag_state=beam.DoFn.StateParam(BAG_STATE)):
...
+        timer1.set(20)
+
+      @on_timer(EXPIRY_TIMER)
+      def expiry_callback(self, buffer=DoFn.StateParam(BUFFER_STATE)):
+        yield b''.join(sorted(buffer.read()))
```
robertwb: Just yield sorted(buffer.read()).

sdks/python/apache_beam/transforms/userstate_test.py
```diff
@@ -415,6 +422,82 @@ def clear_values(self, bag_state=beam.DoFn.StateParam(BAG_STATE)):
...
+        value = element[1]
+        set_state.add(value)
+        clear_timer.set(100)
+        emit_timer.set(1000)
+
+      @on_timer(EMIT_TIMER)
+      def emit_values(self, bag_state=beam.DoFn.StateParam(SET_STATE)):
+        for value in bag_state.read():
+          yield value
+        yield 'extra'
```
robertwb: This shouldn't be needed.

sdks/python/apache_beam/transforms/userstate_test.py
```diff
@@ -415,6 +422,82 @@ def clear_values(self, bag_state=beam.DoFn.StateParam(BAG_STATE)):
...
+      test_stream = (TestStream()
+                     .advance_watermark_to(10)
+                     .add_elements([1, 2, 3])
+                     .add_elements([2])
+                     .advance_watermark_to(24))
+      (p
+       | test_stream
+       | beam.Map(lambda x: ('mykey', x))
+       | beam.ParDo(SimpleTestSetStatefulDoFn())
+       | beam.ParDo(self.record_dofn()))
```
robertwb: Is there a reason assert_that doesn't work here?

sdks/python/apache_beam/transforms/userstate_test.py
```diff
@@ -415,6 +422,82 @@ def clear_values(self, bag_state=beam.DoFn.StateParam(BAG_STATE)):
...
+
+      @on_timer(EMIT_TIMER)
+      def emit_values(self, bag_state=beam.DoFn.StateParam(SET_STATE)):
+        for value in bag_state.read():
+          yield value
+        yield 'extra'
+
+      @on_timer(CLEAR_TIMER)
+      def clear_values(self, bag_state=beam.DoFn.StateParam(SET_STATE)):
+        bag_state.clear()
```
robertwb: Maybe do an add, clear, add to make sure it really works.

sdks/python/apache_beam/transforms/userstate_test.py
```diff
@@ -415,6 +422,82 @@ def clear_values(self, bag_state=beam.DoFn.StateParam(BAG_STATE)):
...
+      _ = (p
+           | test_stream
+           | beam.ParDo(SetStateClearingStatefulDoFn())
+           | beam.ParDo(self.record_dofn()))
+
+    self.assertEqual(
+        ['extra'],
+        StatefulDoFnOnDirectRunnerTest.all_records)
+
   def test_stateful_dofn_nonkeyed_input(self):
```
robertwb: Can you add tests without using TestStream to ensure that the FnApiRunner paths also work?

pabloem:
Run Python PreCommit

sdks/python/apache_beam/transforms/userstate_test.py
rakeshcusat: I haven't tried the `assert_that` here. I took reference from already existing test cases. I will update the test cases.

sdks/python/apache_beam/transforms/userstate_test.py
rakeshcusat: @robertwb do you have an example of the test case?

robertwb: It'd be pretty much the same as the existing test, but using Create rather than TestStream. Create just emits all elements at MIN_TIMESTAMP, and then you can set timers to perform future actions.

sdks/python/apache_beam/transforms/userstate_test.py
```diff
@@ -415,6 +422,82 @@ def clear_values(self, bag_state=beam.DoFn.StateParam(BAG_STATE)):
...
+        emit_timer.set(1000)
+
+      @on_timer(EMIT_TIMER)
+      def emit_values(self, bag_state=beam.DoFn.StateParam(SET_STATE)):
+        for value in bag_state.read():
+          yield value
+        yield 'extra'
+
+      @on_timer(CLEAR_TIMER)
+      def clear_values(self, bag_state=beam.DoFn.StateParam(SET_STATE)):
```
robertwb: s/bag_state/set_state/g

sdks/python/apache_beam/transforms/userstate_test.py
rakeshcusat: `assert_that` doesn't work in case of state and timer. Here is the jira ticket filed long back: https://issues.apache.org/jira/browse/BEAM-5295 I can take care of this in a separate ticket.

sdks/python/apache_beam/transforms/userstate_test.py
rakeshcusat: Added some cases to make sure FnApiRunner path is also executed

rakeshcusat:
R: @robertwb Can you take another look? I have taken care of your comments.

rakeshcusat:
Run Portable_Python PreCommit

sdks/python/apache_beam/runners/worker/bundle_processor.py
```diff
@@ -379,6 +379,39 @@ def _commit(self):
...
+  def __init__(self, state_handler, state_key, value_coder):
+    self._state_handler = state_handler
+    self._state_key = state_key
+    self._value_coder = value_coder
+    self._cleared = False
+    self._added_elements = set()
+
+  def read(self):
+    return _ConcatIterable(
+        {} if self._cleared else _StateBackedIterable(
```
robertwb: {} is a dict, not a set. Also, should we be converting this to a frozen set here?

sdks/python/apache_beam/runners/worker/bundle_processor.py
```diff
@@ -379,6 +379,39 @@ def _commit(self):
...
+    self._added_elements.add(value)
+
+  def clear(self):
+    self._cleared = True
+    self._added_elements = set()
+
+  def _commit(self):
+    if self._cleared:
+      self._state_handler.blocking_clear(self._state_key)
+    if self._added_elements:
```
robertwb: One downside of this way of doing things is that if we add the same element many times from different bundles, the underlying bag state keeps growing. Perhaps we should (periodically?) "compact" by reading everything in, eeduplicating via set(), and then writing it all back out. (A TODO is fine here if you don't want to implement it now.)

sdks/python/apache_beam/transforms/userstate_test.py
```diff
@@ -424,18 +426,18 @@ def clear_values(self, bag_state=beam.DoFn.StateParam(BAG_STATE)):
 
   def test_simple_set_stateful_dofn(self):
     class SimpleTestSetStatefulDoFn(DoFn):
-      BUFFER_STATE = SetStateSpec('buffer', BytesCoder())
+      BUFFER_STATE = SetStateSpec('buffer', FastPrimitivesCoder())
```
robertwb: Use the more specific VarIntCoder here instead.

sdks/python/apache_beam/transforms/userstate_test.py
```diff
@@ -474,14 +476,14 @@ def process(self,
...
+        for value in set_state.read():
           yield value
-        yield 'extra'
 
       @on_timer(CLEAR_TIMER)
-      def clear_values(self, bag_state=beam.DoFn.StateParam(SET_STATE)):
-        bag_state.clear()
+      def clear_values(self, set_state=beam.DoFn.StateParam(SET_STATE)):
+        set_state.clear()
+        set_state.add('value2')
```
robertwb: Add an int, something "different" like 100, for consistency.

sdks/python/apache_beam/transforms/userstate_test.py
```diff
@@ -494,9 +496,33 @@ def clear_values(self, bag_state=beam.DoFn.StateParam(SET_STATE)):
            | beam.ParDo(SetStateClearingStatefulDoFn())
            | beam.ParDo(self.record_dofn()))
 
-    self.assertEqual(
-        ['extra'],
-        StatefulDoFnOnDirectRunnerTest.all_records)
+    self.assertEqual(['value2'], StatefulDoFnOnDirectRunnerTest.all_records)
+
+  def test_stateful_set_state_fn_runner(self):
```
robertwb: test_stateful_set_state_portably

sdks/python/apache_beam/transforms/userstate_test.py
```diff
@@ -494,9 +496,33 @@ def clear_values(self, bag_state=beam.DoFn.StateParam(SET_STATE)):
...
+                  set_state=beam.DoFn.StateParam(SET_STATE)):
+        _, value = element
+        aggregated_value = 0
+        set_state.add(value)
+        for saved_value in set_state.read():
+          aggregated_value += saved_value
+        yield aggregated_value
+
+    p = TestPipeline()
+    values = p | beam.Create([('key', 1), ('key', 2), ('key', 3), ('key', 4)])
```
robertwb: Duplicates?

sdks/python/apache_beam/transforms/userstate_test.py
```diff
@@ -494,9 +496,33 @@ def clear_values(self, bag_state=beam.DoFn.StateParam(SET_STATE)):
...
+
+      def process(self,
+                  element,
+                  set_state=beam.DoFn.StateParam(SET_STATE)):
+        _, value = element
+        aggregated_value = 0
+        set_state.add(value)
+        for saved_value in set_state.read():
+          aggregated_value += saved_value
+        yield aggregated_value
```
robertwb: Maybe set a timer and test clear as well?

sdks/python/apache_beam/transforms/userstate_test.py
rakeshcusat: I can pass some different string value instead.

sdks/python/apache_beam/transforms/userstate_test.py
rakeshcusat: Sure, I will add it.

sdks/python/apache_beam/transforms/userstate_test.py
rakeshcusat: I tried to add this test case: 
```
  def test_stateful_set_state_clean_portably(self):

    class SetStateClearingStatefulDoFn(beam.DoFn):

      SET_STATE = SetStateSpec('buffer', VarIntCoder())
      EMIT_TIMER = TimerSpec('emit_timer', TimeDomain.WATERMARK)
      CLEAR_TIMER = TimerSpec('clear_timer', TimeDomain.WATERMARK)

      def process(self,
                  element,
                  set_state=beam.DoFn.StateParam(SET_STATE),
                  emit_timer=beam.DoFn.TimerParam(EMIT_TIMER),
                  clear_timer=beam.DoFn.TimerParam(CLEAR_TIMER)):
        _, value = element
        set_state.add(value)
        if value == 2:
          clear_timer.set(3)
        elif value == 3:
          emit_timer.set(4)

      @on_timer(EMIT_TIMER)
      def emit_values(self, set_state=beam.DoFn.StateParam(SET_STATE)):
          yield sorted(set_state.read())

      @on_timer(CLEAR_TIMER)
      def clear_values(self, set_state=beam.DoFn.StateParam(SET_STATE)):
        set_state.clear()
        set_state.add(100)

    p = TestPipeline()
    values = p | beam.Create(range(1, 5))
    actual_values = (values
                     | beam.Map(lambda t: window.TimestampedValue(t, t))
                     | beam.WindowInto(window.FixedWindows(1))
                     | beam.ParDo(SetStateClearingStatefulDoFn()))

    assert_that(actual_values, equal_to([100]))

    result = p.run()
    result.wait_until_finish()
```
but it didn't work it complained that timer is not supported.
```
(beam) ➜  python git:(BEAM-7741-implement-setstate) ✗ python -m unittest apache_beam.transforms.userstate_test.StatefulDoFnOnDirectRunnerTest.test_stateful_set_state_clean_portably
WARNING:root:Key coder FastPrimitivesCoder for transform <ParDo(PTransform) label=[ParDo(SetStateClearingStatefulDoFn)]> with stateful DoFn may not be deterministic. This may cause incorrect behavior for complex key types. Consider adding an input type hint for this transform.
E
ERROR: test_stateful_set_state_clean_portably (apache_beam.transforms.userstate_test.StatefulDoFnOnDirectRunnerTest)
  File "apache_beam/transforms/userstate_test.py", line 567, in test_stateful_set_state_clean_portably
    result = p.run()
  File "apache_beam/testing/test_pipeline.py", line 107, in run
    else test_runner_api))
  File "apache_beam/pipeline.py", line 406, in run
    self._options).run(False)
  File "apache_beam/pipeline.py", line 419, in run
    return self.runner.run_pipeline(self, self._options)
  File "apache_beam/runners/direct/direct_runner.py", line 128, in run_pipeline
    return runner.run_pipeline(pipeline, options)
  File "apache_beam/runners/portability/fn_api_runner.py", line 319, in run_pipeline
    default_environment=self._default_environment))
  File "apache_beam/runners/portability/fn_api_runner.py", line 323, in run_via_runner_api
    stage_context, stages = self.create_stages(pipeline_proto)
  File "apache_beam/runners/portability/fn_api_runner.py", line 385, in create_stages
    use_state_iterables=self._use_state_iterables)
  File "apache_beam/runners/portability/fn_api_runner_transforms.py", line 489, in create_and_optimize_stages
    stages = list(phase(stages, pipeline_context))
  File "apache_beam/runners/portability/fn_api_runner_transforms.py", line 1168, in inject_timer_pcollections
    raise NotImplementedError('Timers and side inputs.')
NotImplementedError: Timers and side inputs.

Ran 1 test in 0.207s

FAILED (errors=1)
```
This is a little bit confusing for me because I have used Timer on Flink runner and it works fine. This could be a bug on the portable runner side.

sdks/python/apache_beam/runners/worker/bundle_processor.py
rakeshcusat: I still need to take care of this comment

sdks/python/apache_beam/runners/worker/bundle_processor.py
rakeshcusat: I would prefer to have the compaction logic but I will do some research on my end before making the final decision.

rakeshcusat: @robertwb Check the latest implementation of `read()` method. I think this is what you recommended. Let me know if you want me to implement it in different way.

sdks/python/apache_beam/transforms/userstate_test.py
rakeshcusat: I still have to debug this, I might need some help to debug this.

robertwb: Hmm... for some reason it looks like it's assuming you're using a timer and a side input in the same DoFn.

rakeshcusat: @robertwb I spent some time debugging this. It seems like FnApiRunner is running into issues when there are two timers uses. 
```  def test_two_timers_portably(self):

    class TwoTimersDoFn(beam.DoFn):

      EMIT_TIMER = TimerSpec('emit_timer', TimeDomain.WATERMARK)
      CLEAR_TIMER = TimerSpec('clear_timer', TimeDomain.WATERMARK)

      def process(self,
                  element,
                  emit_timer=beam.DoFn.TimerParam(EMIT_TIMER),
                  clear_timer=beam.DoFn.TimerParam(CLEAR_TIMER)
                  ):
        _, value = element
        if value == 2:
          clear_timer.set(3)
          pass
        elif value == 3:
          emit_timer.set(4)
          pass

      @on_timer(EMIT_TIMER)
      def emit_values(self):
        yield 'emit1'

      @on_timer(CLEAR_TIMER)
      def clear_values(self):
        yield 'clear1'

    p = TestPipeline()
    values = p | beam.Create([('key', 1),
                              ('key', 2),
                              ('key', 3),
                              ('key', 4),
                              ('key', 5),
                              ('key', 6)])
    actual_values = (values
                     | beam.Map(lambda t: window.TimestampedValue(t, t[1]))
                     | beam.WindowInto(window.FixedWindows(1))
                     | beam.ParDo(TwoTimersDoFn()))

    assert_that(actual_values, equal_to([100]))

    result = p.run()
    result.wait_until_finish()```
I tested it on the master branch and it failed consistently and I get the same error message. I couldn't figure out the root cause but it is caused by two timers for sure. It works fine when there is only one timer used.
I feel I should file a separate jira ticket to address this issue, thoughts?

**Comment by rakeshcusat** [comment]
@robertwb can you take a another look of this PR? also check this comment: https://github.com/apache/beam/pull/9090/files#r309342482

File: sdks/python/apache_beam/runners/worker/bundle_processor.py (hunk 9 above)
Comment (robertwb) [reply sdks/python/apache_beam/runners/worker/bundle_processor.py]: Looks good, thanks.

File: sdks/python/apache_beam/transforms/userstate.py [hunk 15]
```diff
@@ -60,6 +60,21 @@ def to_runner_api(self, context):
             element_coder_id=context.coders.get_id(self.coder)))
 
 
+class SetStateSpec(StateSpec):
+  """Specification for a user DoFn Set State cell"""
+
+  def __init__(self, name, coder):
+    assert isinstance(name, str)
```
robertwb: TypeError would be a better thing to raise here.

sdks/python/apache_beam/transforms/userstate_test.py
robertwb: Thanks for the analysis. Filing a JIRA for the multiple-timer issue sounds like the right thing to do here.

sdks/python/apache_beam/transforms/userstate_test.py
```diff
@@ -415,6 +423,151 @@ def clear_values(self, bag_state=beam.DoFn.StateParam(BAG_STATE)):
...
+      EMIT_TIMER = TimerSpec('emit_timer', TimeDomain.WATERMARK)
+
+      def process(self,
+                  element,
+                  set_state=beam.DoFn.StateParam(SET_STATE),
+                  emit_timer=beam.DoFn.TimerParam(EMIT_TIMER)):
+        _, value = element
+        set_state.add(value)
+
+        if value == 5:
```
robertwb: We can't, in general, depend on elements being received in order. Perhaps trigger based on the size of the set?

sdks/python/apache_beam/transforms/userstate.py
rakeshcusat: I can change this. But other StateSpec are doing the same way. Probably change their assert in a separate PR.

sdks/python/apache_beam/transforms/userstate_test.py
rakeshcusat: I checked there is already Jira ticket filed by @tweise (https://issues.apache.org/jira/browse/BEAM-7074)

sdks/python/apache_beam/transforms/userstate_test.py
rakeshcusat: that's good idea. I will change the test case.

rakeshcusat:
@robertwb I have taken care of your comments. Can you approve this now?

rakeshcusat:
R: @robertwb PTAL

PR #9098, opened 2019-07-18
[7746] Create a more user friendly external transform API

chadrik:
As a python developer I'd like to have a more opinionated API for implementing external transforms so that A) it can guide me toward more standardized solutions and B) reduce the amount of boilerplate code that I need to write. 

Improvements:

- Create a constant for the expansion service address, to avoid copying it into the `__init__` args of every external transform
- Move the urn to the class-level to enforce consistency (which enables us to reduce boilerplate) and also for the sake of correctness: the value does not vary per instance
- Reduce boilerplate:
  - Provide a helper to create a `ConfigValue` with encoded value
    - Automatically determine the list of coder urns from a compound coder
    - Handle utf-8 encoding, if requested
  - Create the `ExternalTransform` for the user based on the configuration dictionary that they provide

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/coders/coders.py
+++ b/sdks/python/apache_beam/coders/coders.py
@@ -386,8 +386,14 @@ class FastCoder(Coder):
 class BytesCoder(FastCoder):
   """Byte string coder."""
 
+  def __init__(self, encode_utf8=False):
+    """
+    :param encode_utf8: encode the string value as utf-8.
+    """
+    self.encode_utf8 = encode_utf8
+
   def _create_impl(self):
-    return coder_impl.BytesCoderImpl()
+    return coder_impl.BytesCoderImpl(encode_utf8=self.encode_utf8)
 
   def is_deterministic(self):
     return True
--- a/sdks/python/apache_beam/io/external/kafka.py
+++ b/sdks/python/apache_beam/io/external/kafka.py
@@ -37,18 +37,14 @@
 
 from __future__ import absolute_import
 
-from apache_beam import ExternalTransform
-from apache_beam import pvalue
+from apache_beam.transforms.external import External, ExternalRead
 from apache_beam.coders import BytesCoder
 from apache_beam.coders import IterableCoder
 from apache_beam.coders import TupleCoder
 from apache_beam.coders.coders import LengthPrefixCoder
-from apache_beam.portability.api.external_transforms_pb2 import ConfigValue
-from apache_beam.portability.api.external_transforms_pb2 import ExternalConfigurationPayload
-from apache_beam.transforms import ptransform
 
 
-class ReadFromKafka(ptransform.PTransform):
+class ReadFromKafka(ExternalRead):
   """
     An external PTransform which reads from Kafka and returns a KV pair for
     each item in the specified Kafka topics. If no Kafka Deserializer for
@@ -64,11 +60,13 @@ class ReadFromKafka(ptransform.PTransform):
   byte_array_deserializer = 'org.apache.kafka.common.serialization.' \
                             'ByteArrayDeserializer'
 
+  _urn = 'beam:external:java:kafka:read:v1'
+
   def __init__(self, consumer_config,
                topics,
                key_deserializer=byte_array_deserializer,
                value_deserializer=byte_array_deserializer,
-               expansion_service='localhost:8097'):
+               expansion_service=None):
     """
     Initializes a read operation from Kafka.
 
@@ -88,38 +86,29 @@ class ReadFromKafka(ptransform.PTransform):
                                serialization.ByteArrayDeserializer'.
     :param expansion_service: The address (host:port) of the ExpansionService.
     """
-    super(ReadFromKafka, self).__init__()
-    self._urn = 'beam:external:java:kafka:read:v1'
+    super(ReadFromKafka, self).__init__(expansion_service)
     self.consumer_config = consumer_config
     self.topics = topics
     self.key_deserializer = key_deserializer
     self.value_deserializer = value_deserializer
-    self.expansion_service = expansion_service
 
-  def expand(self, pbegin):
-    if not isinstance(pbegin, pvalue.PBegin):
-      raise Exception("ReadFromKafka must be a root transform")
-
-    args = {
+  def get_config_args(self):
+    str_coder = LengthPrefixCoder(BytesCoder(encode_utf8=True))
+    str_list_coder = IterableCoder(str_coder)
+    map_coder = IterableCoder(TupleCoder([str_coder, str_coder]))
+    return {
         'consumer_config':
-            _encode_map(self.consumer_config),
+            self.config_value(self.consumer_config.items(), map_coder),
         'topics':
-            _encode_list(self.topics),
+            self.config_value(self.topics, str_list_coder),
         'key_deserializer':
-            _encode_str(self.key_deserializer),
+            self.config_value(self.key_deserializer, str_coder),
         'value_deserializer':
-            _encode_str(self.value_deserializer),
+            self.config_value(self.value_deserializer, str_coder),
     }
 
-    payload = ExternalConfigurationPayload(configuration=args)
-    return pbegin.apply(
-        ExternalTransform(
-            self._urn,
-            payload.SerializeToString(),
-            self.expansion_service))
 
-
-class WriteToKafka(ptransform.PTransform):
+class WriteToKafka(External):
   """
     An external PTransform which writes KV data to a specified Kafka topic.
     If no Kafka Serializer for key/value is provided, then key/value are
@@ -132,11 +121,13 @@ class WriteToKafka(ptransform.PTransform):
   byte_array_serializer = 'org.apache.kafka.common.serialization.' \
                           'ByteArraySerializer'
 
+  _urn = 'beam:external:java:kafka:write:v1'
+
   def __init__(self, producer_config,
                topic,
                key_serializer=byte_array_serializer,
                value_serializer=byte_array_serializer,
-               expansion_service='localhost:8097'):
+               expansion_service=None):
     """
     Initializes a write operation to Kafka.
 
@@ -156,62 +147,24 @@ class WriteToKafka(ptransform.PTransform):
                                serialization.ByteArraySerializer'.
     :param expansion_service: The address (host:port) of the ExpansionService.
     """
-    super(WriteToKafka, self).__init__()
-    self._urn = 'beam:external:java:kafka:write:v1'
+    super(WriteToKafka, self).__init__(expansion_service)
     self.producer_config = producer_config
     self.topic = topic
     self.key_serializer = key_serializer
     self.value_serializer = value_serializer
-    self.expansion_service = expansion_service
 
-  def expand(self, pvalue):
-    args = {
+  def get_config_args(self):
+    str_coder = LengthPrefixCoder(BytesCoder())
+    map_coder = IterableCoder(TupleCoder(
+                [LengthPrefixCoder(BytesCoder()),
+                 LengthPrefixCoder(BytesCoder())]))
+    return {
         'producer_config':
-            _encode_map(self.producer_config),
+            self.config_value(self.producer_config.items(), map_coder),
         'topic':
-            _encode_str(self.topic),
+            self.config_value(self.topic, str_coder),
         'key_serializer':
-            _encode_str(self.key_serializer),
+            self.config_value(self.key_serializer, str_coder),
         'value_serializer':
-            _encode_str(self.value_serializer),
+            self.config_value(self.value_serializer, str_coder),
     }
-
-    payload = ExternalConfigurationPayload(configuration=args)
-    return pvalue.apply(
-        ExternalTransform(
-            self._urn,
-            payload.SerializeToString(),
-            self.expansion_service))
-
-
-def _encode_map(dict_obj):
-  kv_list = [(key.encode('utf-8'), val.encode('utf-8'))
-             for key, val in dict_obj.items()]
-  coder = IterableCoder(TupleCoder(
-      [LengthPrefixCoder(BytesCoder()), LengthPrefixCoder(BytesCoder())]))
-  coder_urns = ['beam:coder:iterable:v1',
-                'beam:coder:kv:v1',
-                'beam:coder:bytes:v1',
-                'beam:coder:bytes:v1']
-  return ConfigValue(
-      coder_urn=coder_urns,
-      payload=coder.encode(kv_list))
-
-
-def _encode_list(list_obj):
-  encoded_list = [val.encode('utf-8') for val in list_obj]
-  coder = IterableCoder(LengthPrefixCoder(BytesCoder()))
-  coder_urns = ['beam:coder:iterable:v1',
-                'beam:coder:bytes:v1']
-  return ConfigValue(
-      coder_urn=coder_urns,
-      payload=coder.encode(encoded_list))
-
-
-def _encode_str(str_obj):
-  encoded_str = str_obj.encode('utf-8')
-  coder = LengthPrefixCoder(BytesCoder())
-  coder_urns = ['beam:coder:bytes:v1']
-  return ConfigValue(
-      coder_urn=coder_urns,
-      payload=coder.encode(encoded_str))
--- a/sdks/python/apache_beam/transforms/external.py
+++ b/sdks/python/apache_beam/transforms/external.py
@@ -31,6 +31,8 @@ from apache_beam.portability import common_urns
 from apache_beam.portability.api import beam_expansion_api_pb2
 from apache_beam.portability.api import beam_runner_api_pb2
 from apache_beam.runners import pipeline_context
+from apache_beam.portability.api.external_transforms_pb2 import ConfigValue
+from apache_beam.portability.api.external_transforms_pb2 import ExternalConfigurationPayload
 from apache_beam.transforms import ptransform
 
 # Protect against environments where grpc is not available.
@@ -41,6 +43,63 @@ try:
 except ImportError:
   grpc = None
 # pylint: enable=wrong-import-order, wrong-import-position, ungrouped-imports
+
+DEFAULT_EXPANSION_SERVICE = 'localhost:8097'
+FILTERED_CODERS = {
+  'beam:coder:length_prefix:v1',
+}
+
+
+def iter_urns(coder, context=None):
+  yield coder.to_runner_api_parameter(context)[0]
+  for child in coder._get_component_coders():
+    for urn in iter_urns(child, context):
+      yield urn
+
+
+class External(ptransform.PTransform):
+  """
+  Base class for transforms that produce external transforms.
+  """
+  _urn = None
+
+  def __init__(self, expansion_service=None):
+    self.expansion_service = expansion_service or DEFAULT_EXPANSION_SERVICE
+
+  def get_config_args(self):
+    """
+    :return: dictionary of encoded configuration values to pass to
+             ExternalConfigurationPayload
+    """
+    raise NotImplementedError
+
+  @classmethod
+  def config_value(cls, obj, coder):
+    """
+    Helper to create a ConfigValue with an encoded value.
+    """
+    return ConfigValue(
+      coder_urn=[urn for urn in iter_urns(coder)
+                 if urn not in FILTERED_CODERS],
+      payload=coder.encode(obj))
+
+  def expand(self, pbegin):
+    args = self.get_config_args()
+
+    payload = ExternalConfigurationPayload(configuration=args)
+    return pbegin.apply(
+        ExternalTransform(
+            self._urn,
+            payload.SerializeToString(),
+            self.expansion_service))
+
+
+class ExternalRead(ptransform.PTransform):
+  def expand(self, pbegin):
+    if not isinstance(pbegin, pvalue.PBegin):
+      raise Exception("%s must be a root transform" % self.__class__)
+
+    return super(ExternalRead, self).expand(pbegin)
 
 
 class ExternalTransform(ptransform.PTransform):
--- a/sdks/python/apache_beam/coders/coder_impl.py
+++ b/sdks/python/apache_beam/coders/coder_impl.py
@@ -426,7 +426,12 @@ class BytesCoderImpl(CoderImpl):
 
   A coder for bytes/str objects."""
 
+  def __init__(self, encode_utf8=False):
+    self.encode_utf8 = encode_utf8
+
   def encode_to_stream(self, value, out, nested):
+    if self.encode_utf8:
+      value = value.encode('utf-8')
     out.write(value, nested)
 
   def decode_from_stream(self, in_stream, nested):
--- a/sdks/python/apache_beam/io/external/generate_sequence.py
+++ b/sdks/python/apache_beam/io/external/generate_sequence.py
@@ -17,15 +17,11 @@
 
 from __future__ import absolute_import
 
-from apache_beam import ExternalTransform
-from apache_beam import pvalue
+from apache_beam.transforms.external import ExternalRead
 from apache_beam.coders import VarIntCoder
-from apache_beam.portability.api.external_transforms_pb2 import ConfigValue
-from apache_beam.portability.api.external_transforms_pb2 import ExternalConfigurationPayload
-from apache_beam.transforms import ptransform
 
 
-class GenerateSequence(ptransform.PTransform):
+class GenerateSequence(ExternalRead):
   """
     An external PTransform which provides a bounded or unbounded stream of
     integers.
@@ -49,46 +45,27 @@ class GenerateSequence(ptransform.PTransform):
 
     Experimental; no backwards compatibility guarantees.
   """
+  _urn = 'beam:external:java:generate_sequence:v1'
 
   def __init__(self, start, stop=None,
                elements_per_period=None, max_read_time=None,
-               expansion_service='localhost:8097'):
-    super(GenerateSequence, self).__init__()
-    self._urn = 'beam:external:java:generate_sequence:v1'
+               expansion_service=None):
+    super(GenerateSequence, self).__init__(expansion_service)
     self.start = start
     self.stop = stop
     self.elements_per_period = elements_per_period
     self.max_read_time = max_read_time
-    self.expansion_service = expansion_service
 
-  def expand(self, pbegin):
-    if not isinstance(pbegin, pvalue.PBegin):
-      raise Exception("GenerateSequence must be a root transform")
-
+  def get_config_args(self):
     coder = VarIntCoder()
-    coder_urn = ['beam:coder:varint:v1']
     args = {
         'start':
-        ConfigValue(
-            coder_urn=coder_urn,
-            payload=coder.encode(self.start))
+        self.config_value(self.start, coder)
     }
     if self.stop:
-      args['stop'] = ConfigValue(
-          coder_urn=coder_urn,
-          payload=coder.encode(self.stop))
+      args['stop'] = self.config_value(self.stop, coder)
     if self.elements_per_period:
-      args['elements_per_period'] = ConfigValue(
-          coder_urn=coder_urn,
-          payload=coder.encode(self.elements_per_period))
+      args['elements_per_period'] = self.config_value(self.elements_per_period, coder)
     if self.max_read_time:
-      args['max_read_time'] = ConfigValue(
-          coder_urn=coder_urn,
-          payload=coder.encode(self.max_read_time))
-
-    payload = ExternalConfigurationPayload(configuration=args)
-    return pbegin.apply(
-        ExternalTransform(
-            self._urn,
-            payload.SerializeToString(),
-            self.expansion_service))
+      args['max_read_time'] = self.config_value(self.max_read_time, coder)
+    return args


chadrik:
R: @robertwb 
R: @mxm 
R: @lukecwik

sdks/python/apache_beam/coders/coders.py
```diff
@@ -386,8 +386,14 @@ def _create_impl(self):
 class BytesCoder(FastCoder):
   """Byte string coder."""
 
+  def __init__(self, encode_utf8=False):
```
robertwb: We shouldn't be adding a parameter to BytesCoder--if we want utf-8 encoding StrUtf8Coder should be used instead.

sdks/python/apache_beam/transforms/external.py
```diff
@@ -42,6 +44,63 @@
...
+    return pbegin.apply(
+        ExternalTransform(
+            self._urn,
+            payload.SerializeToString(),
+            self.expansion_service))
+
+
+class ExternalRead(ptransform.PTransform):
+  def expand(self, pbegin):
+    if not isinstance(pbegin, pvalue.PBegin):
```
robertwb: I'm not sure if this check is worth having a separate class, especially as we move to more modular IO that does not necessarily have to be at the root.

sdks/python/apache_beam/transforms/external.py
```diff
@@ -42,6 +44,63 @@
...
+
+
+def iter_urns(coder, context=None):
+  yield coder.to_runner_api_parameter(context)[0]
+  for child in coder._get_component_coders():
+    for urn in iter_urns(child, context):
+      yield urn
+
+
+class External(ptransform.PTransform):
```
robertwb: Naming nit: it seems odd to have a transform named External and another named ExternalTransform. Maybe call this ExternalWrapperTransform? Or maybe, given that this is still all relatively experimental, this should all be folded into the same class?

sdks/python/apache_beam/transforms/external.py
```diff
@@ -42,6 +44,63 @@
   grpc = None
 # pylint: enable=wrong-import-order, wrong-import-position, ungrouped-imports
 
+DEFAULT_EXPANSION_SERVICE = 'localhost:8097'
```
robertwb: +1 to moving this here rather than as a default parameter everywhere, though it still doesn't feel right to have it be a global.

sdks/python/apache_beam/transforms/external.py
```diff
@@ -42,6 +44,63 @@
...
+    raise NotImplementedError
+
+  @classmethod
+  def config_value(cls, obj, coder):
+    """
+    Helper to create a ConfigValue with an encoded value.
+    """
+    return ConfigValue(
+      coder_urn=[urn for urn in iter_urns(coder)
+                 if urn not in FILTERED_CODERS],
```
robertwb: I don't think we can safely filter out certain coders, as that changes the encoding.

sdks/python/apache_beam/transforms/external.py
```diff
@@ -42,6 +44,63 @@
...
+    """
+    raise NotImplementedError
+
+  @classmethod
+  def config_value(cls, obj, coder):
+    """
+    Helper to create a ConfigValue with an encoded value.
+    """
+    return ConfigValue(
+      coder_urn=[urn for urn in iter_urns(coder)
```
robertwb: @mxm So is this just a flattened list of all components? What about coders with payloads? Wasn't the plan instead to reference coder ids from the outer pipeline components?

robertwb: Or was this just a stopgap until we had schemas anyway? (Probably not worth tackling in this PR.)

sdks/python/apache_beam/transforms/external.py
```diff
@@ -42,6 +44,63 @@
...
+
+  def get_config_args(self):
+    """
+    :return: dictionary of encoded configuration values to pass to
+             ExternalConfigurationPayload
+    """
+    raise NotImplementedError
+
+  @classmethod
+  def config_value(cls, obj, coder):
```
robertwb: In terms of ease of use, we could make coder an optional argument and infer it based on the type of obj. Even better, we could let  get_config_args return a dict whose values need not be pre-wrapped and just do the wrapping here.

sdks/python/apache_beam/coders/coders.py
chadrik: Ah, good point.  That may change things on the Java side though, since previously this code was using the bytes coder.

sdks/python/apache_beam/transforms/external.py
chadrik: I agree.  I left it there to match the previous behavior, but I'd love to drop down to one class.

sdks/python/apache_beam/transforms/external.py
chadrik: Also agree here.  I struggled with the same questions.  

Another option is to rename the current `ExternalTransform` to `_ExternalTransform`, since that's part of the API we'll be hiding away from the end user, and I'd rather keep the short-and-sweet name for the public.

I'm also fine with changing the existing `ExternalTransform` to become the base class, but it'll require a lot more adjustments to the tests.

sdks/python/apache_beam/transforms/external.py
chadrik: Glad you mentioned it :)

I'm confused about why the `LengthPrefixCoder` urn is not included in the `ConfigValue .coder_urns` list in the original code.  The original list or the kafka `consumer_config` param was `['beam:coder:iterable:v1', 'beam:coder:kv:v1', 'beam:coder:bytes:v1', 'beam:coder:bytes:v1']`.  I was trying to keep the end result the same, and I assumed that there were coders which were "write-only", and thus did not need to be transmitted to Java.

sdks/python/apache_beam/transforms/external.py
chadrik: I agree, we can make this even easier, and we should.  This suggestion got me thinking: a lightweight schema class could be a nice way to do this.  Here's an example using [`attrs`](https://github.com/python-attrs/attrs):

```python
@attrs
ReadFromKafkaSchema(object):
    consumer_config = attrib(type=Iterable[Tuple[str, str]])
    topics = attrib(type=Iterable[str])
    key_deserializer = attrib(type=str)
    value_deserializer = attrib(type=str)
    
class ReadFromKafka(ExternalTransform):
    schema = ReadFromKafkaSchema
    urn = 'beam:external:java:kafka:read:v1'
```

That's all the would be required from a user.  We would automatically handle converting `**kwargs` to a `ExternalConfigurationPayload` on `__init__`.

I like your suggestion of determining the coder automatically from the input type, but I think it might be too automagic.  An explicit schema provides more protection, and it's also a nice simple stand-in for a protobuf object.  I assume we don't want to use those here because it would require each new external transform to add to the beam model.

You may be wary of adding `attrs` as a dependency for this, but I've seen it used in a number of projects, since it answers such a fundamental problem in python.  It was the inspiration for the dataclass feature in python 3.7, but it continues to work on python2.  It also supports type analsys via mypy.   `attrs` is also an ideal library for creating data classes to use with beam transforms, since it's more structured than a dict, can enforce immutability, can be statically analyzed, and can be inspected to automatically serialize.  See [cattrs](https://cattrs.readthedocs.io/en/latest/) for an example of the latter.

sdks/python/apache_beam/transforms/external.py
mxm: We do not need the length prefix coder here, do we? That's because we get the encoded bytes via Protobuf and do not have to reason about the length.

sdks/python/apache_beam/transforms/external.py
mxm: I also think `External` would be nice. IMHO Consolidating the two classes would only be confusing. Note that in Java we use `External` for `ExternalTransform`. A bit of a naming mess.

What about `ExternalTransformBase`? 

Or could we use a builder pattern instead?

sdks/python/apache_beam/transforms/external.py
mxm: Would also favor removing this class. It shouldn't matter if this is a read or not. Perhaps we could encode this in the `External` class whether the transform requires an Impulse or not?

sdks/python/apache_beam/coders/coders.py
mxm: The String utf8 coder was added after the initial cross-language configuration version. For TODO see https://jira.apache.org/jira/browse/BEAM-7244

sdks/python/apache_beam/transforms/external.py
robertwb: Once the portable schema stuff is wrapped up, hopefully we can migrate to that as a more standard format.

sdks/python/apache_beam/transforms/external.py
robertwb: But we do need the lengths if it is, say, KV<LengthPrefixedX, LenthPrefixedY>. What's more, we're getting the already encoded value here (which is different than if we had tried to filter them out before doing the encoding. 

Are you seeing instances of LengthPrefixCoder that need filtering out? If so, this deserves further investigation. Otherwise, let's just drop this.

sdks/python/apache_beam/transforms/external.py
robertwb: I like attrs, but I wouldn't want to force its use. The core value in this class is the mapping from dict + urn to an external transform application, and the users can decide how to define the dict.

sdks/python/apache_beam/coders/coders.py
chadrik: removed changes to `BytesCoder` and replaced usage with `StrUtf8Coder`.

sdks/python/apache_beam/transforms/external.py
chadrik: I removed this filter.  

Note that between this change and the `StrUtf8Coder` change, the list of coder urns produced by `ReadFromKafka` has changed from this:

```
 coder_urn: "beam:coder:iterable:v1"
 coder_urn: "beam:coder:kv:v1"
 coder_urn: "beam:coder:bytes:v1"
 coder_urn: "beam:coder:bytes:v1"
```

to this:

```
 coder_urn: "beam:coder:iterable:v1"
 coder_urn: "beam:coder:kv:v1"
 coder_urn: "beam:coder:length_prefix:v1"
 coder_urn: "beam:coder:string_utf8:v1"
 coder_urn: "beam:coder:length_prefix:v1"
 coder_urn: "beam:coder:string_utf8:v1"
```

Likewise for `WriteToKafka`,

@mxm  can you please confirm whether this will require adjustments to the Kafka Java code?

sdks/python/apache_beam/transforms/external.py
```diff
@@ -42,6 +44,63 @@
...
+    self.expansion_service = expansion_service or DEFAULT_EXPANSION_SERVICE
+
+  def get_config_args(self):
+    """
+    :return: dictionary of encoded configuration values to pass to
+             ExternalConfigurationPayload
+    """
+    raise NotImplementedError
+
+  @classmethod
```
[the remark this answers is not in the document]
chadrik: Simply because `self` was not required.  It could be also be a `staticmethod`.

sdks/python/apache_beam/transforms/external.py
chadrik: Removed `ExternalRead`.

sdks/python/apache_beam/transforms/external.py
chadrik: I've pushed what I think is a good compromise.  

Here's a brief example:

```python
class ReadFromKafka(External):
  byte_array_deserializer = 'org.apache.kafka.common.serialization.' \
                            'ByteArrayDeserializer'

  _urn = 'beam:external:java:kafka:read:v1'

  _schema = {
    'consumer_config': typehints.Iterable[typehints.KV[unicode, unicode]],
    'topics': typehints.Iterable[unicode],
    'key_deserializer': unicode,
    'value_deserializer': unicode,
  }

  def __init__(self, 
               consumer_config,
               topics,
               key_deserializer=byte_array_deserializer,
               value_deserializer=byte_array_deserializer,
               expansion_service=None):
    super(ReadFromKafka, self).__init__(expansion_service)
    self.consumer_config = list(consumer_config.items())
    self.topics = topics
    self.key_deserializer = key_deserializer
    self.value_deserializer = value_deserializer
```

The new `_schema` class attribute guides the extraction and encoding of the configuration payload.  

What I really like about this is that it sets us up for an even easier workflow when we go python3-only, in which case, we can drop the schema and instead read the typehints from the `__init__` annotations:

```python
class ReadFromKafka(External):
  byte_array_deserializer = 'org.apache.kafka.common.serialization.' \
                            'ByteArrayDeserializer'

  _urn = 'beam:external:java:kafka:read:v1'

  def __init__(self, 
               consumer_config: Iterable[Tuple[str, str]],
               topics: Iterable[str],
               key_deserializer: str = byte_array_deserializer,
               value_deserializer: str = byte_array_deserializer,
               expansion_service: Optional[str] = None):
    super(ReadFromKafka, self).__init__(expansion_service)
    self.consumer_config = list(consumer_config.items())
    self.topics = topics
    self.key_deserializer = key_deserializer
    self.value_deserializer = value_deserializer
```

It's still possible for classes to override various methods on the `External` base class to accomplish this however they like.  For example, if they want to do their own config extraction without relying on a schema, they can override `get_config()`, to control the encoding, they can override ``encode_config()` or if they want to do the whole thing, they can override `get_encoded_config()`. 

I provided an example of overriding `get_config()` on `GenerateSequence`.

--- 

There's one (known) problem left: now that it's inferring the coder from typehints I don't know when it's appropriate to automatically insert the `LengthPrefixCoder`.  Any suggestions?

sdks/python/apache_beam/transforms/external.py
chadrik: I'm going to come back to the naming question at the end.  For now I want to focus on the design.

sdks/python/apache_beam/transforms/external.py
chadrik: Oh, and by the way, in python3 , attrs copies the annotations into the `__init__` that it generates, so _if_ a user wanted to use attrs or dataclasses, this would also work (but would not be required):

```python
@attr.s
class ReadFromKafka(External):
  byte_array_deserializer = 'org.apache.kafka.common.serialization.' \
                            'ByteArrayDeserializer'

  _urn = 'beam:external:java:kafka:read:v1'

  consumer_config = attrib(type=Iterable[Tuple[str, str]])
  topics = attrib(type=Iterable[str])
  key_deserializer = attrib(type=str)
  value_deserializer = attrib(type=str)
  expansion_service = attrib(type=Optional[str])
```

sdks/python/apache_beam/io/external/kafka.py
```diff
@@ -64,11 +59,20 @@ class ReadFromKafka(ptransform.PTransform):
   byte_array_deserializer = 'org.apache.kafka.common.serialization.' \
                             'ByteArrayDeserializer'
 
+  _urn = 'beam:external:java:kafka:read:v1'
+
+  _schema = {
+    'consumer_config': typehints.Iterable[typehints.KV[unicode, unicode]],
```
robertwb: Let's use typing rather than typehints wherever we can. We should modify get_coder to accept this via native_type_conversion.convert_to_beam_type.

sdks/python/apache_beam/transforms/external.py
```diff
@@ -42,6 +46,97 @@
...
+  def encode_config(cls, config):
+    schema = cls._schema or {}
+    result = {}
+    for k, v in config.items():
+      typehint = schema.get(k)
+      if v is None and (
+          typehint is None or not isinstance(typehint, typehints.Optional)):
+        # make it easy for user to filter None by default
+        continue
+      result[k] = cls.config_value(v, typehint)
```
robertwb: Do we want to strip the optional wrapping here (e.g. so an Optional[int] is encoded just as a raw into or omitted)?

sdks/python/apache_beam/transforms/external.py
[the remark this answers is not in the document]
mxm: Static would be more intuitive here, but fine as-is.

sdks/python/apache_beam/transforms/external.py
mxm: Point taken about the length prefix being required. 

>@mxm can you please confirm whether this will require adjustments to the Kafka Java code?

For the length prefix to be supported, we will add a detection here to properly initialize the coder's component: https://github.com/apache/beam/blob/4e40b53198870b2af1b178bda8ffa326a60833dc/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/expansion/ExpansionService.java#L210 Like already mentioned the encoding scheme is flawed and should very soon be replaced by the Components structure which would hold the element coder instead.

chadrik: > For the length prefix to be supported, we will add a detection here to properly initialize the coder's component:

There's one more wrinkle here:  now that the design has been changed to automatically detect the coder from the type hint, the `LengthPrefixCoder` is not being used at all.

e.g., given this type hint:

```
Iterable[Tuple[unicode, unicode]]
```

We generate this list of coders:

```
 coder_urn: "beam:coder:iterable:v1"
 coder_urn: "beam:coder:kv:v1"
 coder_urn: "beam:coder:string_utf8:v1"
 coder_urn: "beam:coder:string_utf8:v1"
```

How important is the `LengthPrefixCoder` in this case?  Can we drop it?  

@robertwb  If `LengthPrefixCoder` is required, do you have any thoughts on an interface for inserting this automatically?  i.e. Is there anything about a typehint that could indicate when it should automatically be used, or is it always an explicit choice by the user?

sdks/python/apache_beam/transforms/external.py
```diff
@@ -42,6 +46,97 @@
...
+    config = {}
+    missing = []
+    for key in self._schema:
+      try:
+        config[key] = getattr(self, key)
+      except AttributeError:
+        missing.append(key)
+    return config
+
+  def get_encoded_config(self):
```
chamikaramj: Do you expect user's to override (a subset of) all these methods ? Can we reduce the public surface of this class by just keeping get_encoded_config() (for maximum flexibility) and get_schema() (instead of cls._schema) public and making other public methods (get_config, encode_config, config_value) start with _ ?

sdks/python/apache_beam/transforms/external.py
chamikaramj: +1 for simplifying. I prefer folding everything to a single class named 'External'. I don't see the need to have two base classes if that is going to be overridden anyways. Also this will fix the naming discrepancy between Java and Python.

sdks/python/apache_beam/transforms/external.py
chadrik: >  Do you expect user's to override (a subset of) all these methods ?

The idea is to make the `_schema` design as clean and easy to use as possible, so that most users only need to implement a schema.  But I still wanted to provide the flexibility to override methods to modify the behavior for advanced users, if desired.  

> Can we reduce the public surface of this class by just keeping get_encoded_config() (for maximum flexibility) and get_schema() (instead of cls._schema) public and making other public methods (get_config, encode_config, config_value) start with _ ?

Sure, that works for me.  

My ultimate goal is the `dataclasses` design that I outlined above, and it should be very easy to write a `get_schema()` implementation that derives the schema dict from a `dataclass`-based `ExternalTransform`.

sdks/python/apache_beam/transforms/external.py
chadrik: I can do this, but just to be clear, `External` would become an abstract base class in that case, in order to port over the changes I've made to simplify the interface.

So this:

```python
class ExternalTransform(ptransform.PTransform):
  _namespace_counter = 0
  _namespace = threading.local()

  _EXPANDED_TRANSFORM_UNIQUE_NAME = 'root'
  _IMPULSE_PREFIX = 'impulse'

  def __init__(self, urn, payload, endpoint):
    if grpc is None and isinstance(endpoint, str):
      raise NotImplementedError('Grpc required for external transforms.')
    self._urn = urn
    self._payload = payload
    self._endpoint = endpoint
    self._namespace = self._fresh_namespace()

  [...]
```

Would become roughly:

```python
class External(ptransform.PTransform):
  _urn = None
  _namespace_counter = 0
  _namespace = threading.local()

  _EXPANDED_TRANSFORM_UNIQUE_NAME = 'root'
  _IMPULSE_PREFIX = 'impulse'

  def __init__(self, endpoint):
    if grpc is None and isinstance(endpoint, str):
      raise NotImplementedError('Grpc required for external transforms.')
    if self._urn is None:
      raise NotImplementedError('External sub classes must provide a _urn attribute.')
    self._endpoint = endpoint
    self._namespace = self._fresh_namespace()
    self._schema = self.get_schema()

  [...]
```

Such that we can write a subclass like:

```python
class ReadFromKafka(External):
  byte_array_deserializer = 'org.apache.kafka.common.serialization.' \
                            'ByteArrayDeserializer'

  _urn = 'beam:external:java:kafka:read:v1'

  def __init__(self, 
               consumer_config,
               topics,
               key_deserializer=byte_array_deserializer,
               value_deserializer=byte_array_deserializer,
               expansion_service=None):
    super(ReadFromKafka, self).__init__(expansion_service)
    self.consumer_config = list(consumer_config.items())
    self.topics = topics
    self.key_deserializer = key_deserializer
    self.value_deserializer = value_deserializer

  def get_schema(self):
    return {
    'consumer_config': typehints.Iterable[typehints.KV[unicode, unicode]],
    'topics': typehints.Iterable[unicode],
    'key_deserializer': unicode,
    'value_deserializer': unicode,
  }

```

This will change the tests a bit. 

Does that all sound good?

mxm: For the sake of keeping separation of concerns and readability of the code, I would not advise merging the two. One class is responsible for **resolving an External transform via the expansion server**, the other class is responsible for **building a type of payload**. Merging the two will make it harder to add other ways to build payload for external transforms, e.g. add an explicit Proto message for certain transforms.

The abstract class would make this less worse, but the Beam pattern is to let a transform expand. I think we will benefit from the decoupling. If we want to fix the naming issues, we can just rename the classes like already suggested.

So just to recap:

Java: There is only the raw `External` which is equivalent to `ExternalTransform` in Python
Python: There is `External` (Payload building) and `ExternalTransform` (Resolution via Expansion Service)

How about renaming External to `ExternalWithSchema`? To consolidate with Java we could then also rename `ExternalTransform` to `External`.

sdks/python/apache_beam/transforms/external.py
chadrik: Addressed in my latest push.

sdks/python/apache_beam/io/external/kafka.py
chadrik: Fixed.

sdks/python/apache_beam/transforms/external.py
chadrik: To clarify my `dataclasses` comment: that's for a future PR, and it won't require `dataclasses`.

sdks/python/apache_beam/transforms/external.py
chadrik: I think I agree with @mxm.  I'm still a bit new to beam, but I think combining them could add quite a lot of clutter to a class that's meant to be subclassed by users. 

As for names, if we care more about a nice python interface, rather than parity with Java, I might go with:

- rename raw `ExternalTransform` to `_External`: since most users will not directly interact with this
- rename payload building `External` to `ExternalTransform`: since this has the cleanest most self-explanatory name

If we care about parity with Java:

- rename raw `ExternalTransform` to `External`
- rename payload building `External` to `ExternalTransformInterface`

sdks/python/apache_beam/transforms/external.py
chadrik: My question about the `LengthPrefixCoder` is still outstanding, but I have one more for @mxm:  

I realized that I don't understand how the name mapping works for the payloads: e.g. how does  `"elements_per_period": 20` in our python-based payload end up calling `GenerateSequence.Builder.setElementsPerPeriod(20)` in Java?  I don't see any name mapping happening in Java, I don't think it can be based on order (dicts in python are not ordered, plus we omit optional values), so I'm a bit confused.

robertwb: We should not worry about LengthPrefixCoder at all. It's mostly for actors that need to pass records through without interpreting them. Eventually this can all be replaced by schemas anyway, so not something to worry to much about here.

robertwb:
OK, taking a step back and thinking about this some more, there are three essential pieces of data that the user must provide for this transform: the urn and the configuration parameters, and how to encode the configuration parameters. This suggests the methods `get_urn`, `get_config`, and `get_schema` which are invoked by the base class. 

For simplicity, as urns are typically static, we could provide a default `get_urn` that returns `self._urn`. 

`get_config` and `get_schema` are a bit more interesting as there is non-trivial redundancy between them. If `get_schema` is not implemented, we can infer it from the types in `get_config`. If `get_config` is not implemented, we can infer it from the identically-named self attributes in `get_schema`. I propose we let the user implement either (or both). 

Additionally, `get_schema` could be inferred for data classes, or those derived from cattrs. (Should this be done implicitly due to subclassing, or explicitly via a class decorator?)

This is related to the further magic/ease-of-use, which is allowing a single _schema attribute rather than implementing `get_schemea`, and also auto-generating `__init__` methods from it. 

Does this sound right?

sdks/python/apache_beam/transforms/external.py
chadrik: > We should not worry about LengthPrefixCoder at all.

Sounds good.

> Eventually this can all be replaced by schemas anyway, so not something to worry to much about here.

Can you point me at a document or issue for this?  I'd love to learn more.

sdks/python/apache_beam/transforms/external.py
chadrik: yes, I think so.  I didn't realize that Union/Optional caused coders to fall back to FastPrimitive, which is kind of a bummer.

chadrik:
> Does this sound right?

Spot on, as usual.  A few bits comments below:

> For simplicity, as urns are typically static, we could provide a default get_urn that returns self._urn.

Note, we might want `get_urn` to be a classmethod. 

> get_config and get_schema are a bit more interesting as there is non-trivial redundancy between them. If get_schema is not implemented, we can infer it from the types in get_config. If get_config is not implemented, we can infer it from the identically-named self attributes in get_schema. I propose we let the user implement either (or both).

I'm leaning toward requiring `get_schema` and _not_ deriving schema from the config, because I think consistency of the schema is important, especially for the author of the native transform who would (likely) be expecting each key to use the same coder every time.  If we derive the coder from the value, then it could change for a given config key depending on how the external transform is initialized (`x=0` vs `x=0.5`) .  The schema lets us be consistent and concrete.  

> Additionally, get_schema could be inferred for data classes, or those derived from cattrs. (Should this be done implicitly due to subclassing, or explicitly via a class decorator?)

I'd like to avoid the sub-class requirement.  More info below.

> This is related to the further magic/ease-of-use, which is allowing a single _schema attribute rather than implementing get_schemea, and also auto-generating __init__ methods from it.

yes, basically.  let me clarify this a little bit more.

A user should be able to write a complete class including schema using type annotations, like this:

```python
class ReadFromKafka(External):
  _urn = 'beam:external:java:kafka:read:v1'

  def __init__(self,
                     consumer_config: Iterable[Tuple[str, str]],
                     topics: Iterable[str],
                     key_deserializer: str,
                     value_deserializer: str,
                     expansion_service: Optional[str] = None):
    super(ReadFromKafka, self).__init__(expansion_service)
    self.consumer_config = list(consumer_config.items())
    self.topics = topics
    self.key_deserializer = key_deserializer
    self.value_deserializer = value_deserializer
```

And `attrs` / `dataclasses` can be used as a shortcut for deriving the `__init__`:

```python
@dataclass
class ReadFromKafka(External):
  _urn = 'beam:external:java:kafka:read:v1'

  consumer_config: Iterable[Tuple[str, str]]
  topics: Iterable[str]
  key_deserializer: str
  value_deserializer: str
  expansion_service: Optional[str] = field(default=None)
```

`expansion_service` is obviously the odd-man-out here, so I'll give some thought to how this could be simplified. 

The thing I wanted to make clear is that I don't think `get_schema` would need special support for attrs/dataclasses in this case:  both approaches outlined above will add type `__annotations__` to `__init__`, from which we can derive our schema.

sdks/python/apache_beam/transforms/external.py
chadrik: I played around with this some more.  The current design seems to *require* the LengthPrefixCoder.  If I don't wrap the str/bytes coder with length-prefix I get this error:

```
Caused by: org.apache.beam.sdk.coders.CoderException: java.io.EOFException: reached end of stream after reading 38 bytes; 112 bytes expected
```

@mxm  can you confirm that the current design requires the LengthPrefixCoder for strings?

@robertwb I saw the schema / row-coder PR at #9188.  Is this the schema support you were referring to?  Is anyone assigned to porting external transforms to using schema coders?  Luckily I think most of the discussion here about the design of the interface remains valid even after that's complete.

sdks/python/apache_beam/transforms/external.py
chamikaramj: I guess a part of my question was whether we need an inheritance hierarchy to do this or can payload building be a utility/delegate that is not necessarily a transform. Having a similarly named two transforms could be confusing to users. How about adding an ExternalTransformPayloadBuilder that can be sub-classed by users (with a default implementation) that is used as a utility by the transform "External" or "ExternalTransform " that all external transforms have to extend ?

In general we try to keep SDKs consistent but this doesn't have to come at the cost of making Python API not nice.

sdks/python/apache_beam/transforms/external.py
chamikaramj: Thanks. Looks updated public API good.

sdks/python/apache_beam/transforms/external.py
robertwb: Coders also have the notion of nested/unnested. It happens that LengthPrefix(UnnestedBytes) == NestedBytes. Perhaps this is part of the issue here?

Yes, that's the schema/row stuff. I was just commenting that this particular part is going to change (no, no one's on it yet) so whatever works here is fine.

robertwb:
On may have transforms with no config params, or whose parameter is a single string. Requiring a schema in this case is extra boilerplate (though arguably good practice, but in the spirit of PEP 484 very optional). To pass a wrong value (e.g. a float when an int is required) would be a runtime type error (like one would expect). 

A default implementation that looks at the `__annotations__` of `__init__`, if present, allow succinct creation of such classes as mentioned.

sdks/python/apache_beam/transforms/external.py
chadrik: > I guess a part of my question was whether we need an inheritance hierarchy to do this or can payload building be a utility/delegate that is not necessarily a transform. 

I'll break my choice down into a set of user stories / requirements.

**User story 1: As a developer writing a pipeline, using an external transform in a pipeline should _feel_ no different than using a native transform.**

For example, currently we're using `ReadFromPubSub`, but it's not supported on Flink.   Upgrading the following code to an external transform should be as easy as replacing the import:

```python
from apache_beam.io.gcp.pubsub import ReadFromPubSub
pipe = beam.Pipeline(options=pipeline_options)
pipe  | 'PubSubInflow' >> ReadFromPubSub(
    topic='projects/foo/topics/output', 
    with_attributes=True)
```

**User story 2: As a developer writing an external transform in python, it should be as easy as providing a schema and a URN.**   

I've given a few examples of this in this PR, using both type annotations and dataclasses.

---

I think inheritance is the best approach to meeting both of these requirements.  If we can agree on these two requirements, then we just need to determine whether it is also a requirement to have a lower-level concrete class like `ExternalTransform`.   If it is, then we need 2 classes, otherwise, we can use 1.

Unless anyone objects, I think I'll make an adjustment to the PR to combine the functionality of `External` into `ExternalTransform` so that we can see what the design would look like with a single abstract class.

chamikaramj: User story 1:...
Completely agree.
Minor point: Python ReadFromPubSub is not supported in Flink since it's a native transform (fully implemented in Dataflow runner). But yeah, using Java custom PubSub source from Python as a cross-language transform will give us something that works for all runners.

User story 2:...
Also agree. But we have to figure out what is the write API to provide config, schema, URN.
Agreeing with previous comment from @mxm, we have two separate features related to cross-language transforms that we would like to keep in separate classes.
(1) An composite transform that is expandable through the expansion service.
(2) An abstraction for building the Payload for (1).

Looking at these two (1) is surely a Beam transform, but looking at (2) it sounds like something else. So what I'm proposing is.

```python
class External(ptransform.PTransform):
  // We can call this External or ExternalTransform. This is the class that performs expansion.

  // URN is expected to be set by subclasses.
  urn = None
  
 
  def get_payload_builder():
     raise NotImplementedError

class PayloadBuilder:
  def get_conded_config():
    pass

class SchemaBasedPayloadBuilder(PayloadBuilder): 
  def get_schema():
    pass

```
In other words, we can maintain an inheritance hierarchy for payload building. But this doesn't have to be an inheritance hierarchy of Beam transforms.

chadrik: > Minor point: Python ReadFromPubSub is not supported in Flink since it's a native transform (fully implemented in Dataflow runner). But yeah, using Java custom PubSub source from Python as a cross-language transform will give us something that works for all runners.

full disclosure: I'm currently working on making `PubSubIO` externalizable, which is what lead me down this path in the first place.

> In other words, we can maintain an inheritance hierarchy for payload building. But this doesn't have to be an inheritance hierarchy of Beam transforms.

yes, but unless I'm missing something, we can't accomplish user story 1 with this approach. 
 Consider this hypothetical example:

```python
from apache_beam.io.external.gcp.pubsub import ReadFromPubSubPayloadBuilder
pipe = beam.Pipeline(options=pipeline_options)
# ERROR: payload builder is not a transform!
pipe  | 'PubSubInflow' >> ReadFromPubSubPayloadBuilder(
    topic='projects/foo/topics/output', 
    with_attributes=True)
```

chamikaramj: Actually, what I meant was something like.

```python
class ReadFromPubSubPayloadBuilder(SchemaBasedPayloadBuilder):
  def __init__(self, topic):
    self._topic = topic

  def get_schema(self):
    // Implementation

# We can call this class just ReadFromPubSub in case we want to make cross-language version the default.
class ReadFromPubSubExternal(ExternalTransform):
 
  URN = 'urn:external:pubsub'

  def __init__ (self, topic):
    self._payload_builder = ReadFromPubSubPayloadBuilder(topic)

  def get_payload_builder(self):
    return self._payload_builder:

  def expand(self, input):
    # optionally build a composite transform by combining ExternalTransform.expand() with other transforms.
```

#### User pipeline
```python
from apache_beam.io.external.gcp.pubsub import ReadFromPubSubExternal
pipe = beam.Pipeline(options=pipeline_options)
pipe  | 'PubSubInflow' >> ReadFromPubSubExternal(
    topic='projects/foo/topics/output', 
    with_attributes=True)
```

chadrik: @chamikaramj  Ah, I get it now.    

In your example, probably the payload builder will need to be instantiated with `self` (the `ExternalTransform` instance) to get the complete context, but yeah, should be doable.  I'll play around with that idea to see if I can some up with something that meets my UX requirements for User Story 2.

chamikaramj: Yeah, prob. `SchemaBasedPayloadBuilder` should be instantiated with self (transform). I'd say leave `PayloadBuilder` mostly unimplemented for someone who wishes to build the payload in a completely different way.

chadrik: I pushed a change that implements this new design.  I think it's a big improvement. 

One last design question:  should `ExternalTransform.__init__` accept a payload builder, or a payload?  It would simplify the interface for writers of external transforms a little bit, but removing the need to call `.build()`.

Here's a brief overview of the main payload builders that are available (in order of increasing brevity):

## NamedTuple

```python

ReadFromKafkaSchema = typing.NamedTuple(
  'ReadFromKafkaSchema',
  [
    ('consumer_config', typing.List[typing.Tuple[unicode, unicode]]),
    ('topics', typing.List[unicode]),
    ('key_deserializer', unicode),
    ('value_deserializer', unicode),
  ]
)


class ReadFromKafka(ExternalTransform):

  byte_array_deserializer = 'org.apache.kafka.common.serialization.' \
                            'ByteArrayDeserializer'

  URN = 'beam:external:java:kafka:read:v1'

  def __init__(self, consumer_config,
               topics,
               key_deserializer=byte_array_deserializer,
               value_deserializer=byte_array_deserializer,
               expansion_service=None):
    super(ReadFromKafka, self).__init__(
      self.URN,
      NamedTupleBasedPayloadBuilder(
        ReadFromKafkaSchema(
          consumer_config=list(consumer_config.items()),
          topics=topics,
          key_deserializer=key_deserializer,
          value_deserializer=value_deserializer,
        )
      ).build(),
      expansion_service
    )
```

## Annotations

```python
class ReadFromKafka(ExternalTransform):
  byte_array_deserializer = 'org.apache.kafka.common.serialization.' \
                            'ByteArrayDeserializer'

  URN = 'beam:external:java:kafka:read:v1'

  def __init__(self, consumer_config: typing.List[typing.Tuple[unicode, unicode]],
               topics: typing.List[unicode],
               key_deserializer: unicode = byte_array_deserializer,
               value_deserializer: unicode = byte_array_deserializer,
               expansion_service=None):

    super(ReadFromKafka, self).__init__(
      self.URN,
      AnnotationBasedPayloadBuilder(
        self,
        consumer_config=consumer_config,
        topics=topics,
        key_deserializer=key_deserializer,
        value_deserializer=value_deserializer,
      ).build(),
      expansion_service
    )
```

## Dataclasses

We should be able to simplify this even further with a class decorator or base class to implement `__post_init__`.

```python
@dataclass
class ReadFromKafka(ExternalTransform):
  byte_array_deserializer = 'org.apache.kafka.common.serialization.' \
                            'ByteArrayDeserializer'

  URN = 'beam:external:java:kafka:read:v1'

  consumer_config: typing.List[typing.Tuple[unicode, unicode]]
  topics: typing.List[unicode]
  key_deserializer: unicode = byte_array_deserializer
  value_deserializer: unicode = byte_array_deserializer
  expansion_service: InitVar[typing.Optional[str]] = None

  def __post_init__(self, expansion_service):
    super(ReadFromKafka, self).__init__(
      self.URN,
      DataclassBasedPayloadBuilder(self).build(),
      expansion_service
    )
```

chadrik: I updated `ExternalTransform` to accept either a payload or a builder.

chadrik:
Run Python PreCommit

sdks/python/apache_beam/transforms/external.py
```diff
@@ -42,6 +46,103 @@
...
+        continue
+      result[k] = cls._config_value(v, typehint)
+    return result
+
+  def build(self):
+    args = self._encode_config(self._values, self._schema)
+    return ExternalConfigurationPayload(configuration=args)
+
+
+class NamedTupleBasedPayloadBuilder(SchemaBasedPayloadBuilder):
```
chamikaramj: Please add unit tests for each of these payload builder types.

sdks/python/apache_beam/transforms/external.py
```diff
@@ -42,6 +46,103 @@
...
+      tuple_instance._field_types, tuple_instance._asdict())
+
+
+class AnnotationBasedPayloadBuilder(SchemaBasedPayloadBuilder):
+  """
+  Build a payload based on an external transform's type annotations.
+
+  Supported in python 3 only.
+  """
+  def __init__(self, transform, **values):
```
chamikaramj: Please add a pydocs and define parameters (here and in other places)

sdks/python/apache_beam/transforms/external.py
```diff
@@ -42,6 +46,103 @@
...
+    self._values = values
+    self._schema = schema or {}
+
+  @classmethod
+  def _encode_config(cls, config, schema):
+    result = {}
+    for k, v in config.items():
+      typehint = schema.get(k)
+      if v is None and (
+          typehint is None or not isinstance(typehint, typehints.Optional)):
```
chamikaramj: So seems like 'None' here means, used default value in remote SDK, right ? We should clarify that in documentation.

chadrik: The behavior is a bit undefined at this point, since we only have `GenerateSequence` as an example.  In that case, the field is null on the Java side as well.  

The reason we skip the fields if they're none here is ultimately because we don't have a cross-platform coder for none/null or a cross-platform way of declaring a type as optional/nullable.  So the behavior would be better defined if we _could_ serialize None and send that value to Java.

sdks/python/apache_beam/io/external/kafka.py
```diff
@@ -37,18 +37,25 @@
...
-
-class ReadFromKafka(ptransform.PTransform):
+import typing
+
+from past.builtins import unicode
+
+from apache_beam.transforms.external import ExternalTransform, NamedTupleBasedPayloadBuilder
+
+
+ReadFromKafkaSchema = typing.NamedTuple(
```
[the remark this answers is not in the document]
chadrik: btw, I used the name "schema" since ultimately we will switch to using the upcoming schema support when serializing this payload.  Not sure if that changes your mind on the naming at all.

sdks/python/apache_beam/transforms/external.py
mxm: @chadrik Sorry for the late reply. Vacations and open-source work do not always go together well :) Actually, for the StringUtf8Coder, the length should be added automatically in the nested context, no need to add the length prefix.

>how does "elements_per_period": 20 in our python-based payload end up calling GenerateSequence.Builder.setElementsPerPeriod(20) in Java?

It is a simple mapping scheme which we also use for the pipeline options. It converts `snake_case` to `camelCase` and then looks up the setter in the Java configuration class.

>@mxm can you confirm that the current design requires the LengthPrefixCoder for strings?

I don't think so, even with the byte encoding in master, the length prefix is added automatically by the `ByteArrayCoder`.

mxm: Perhaps paste the full stack trace here if you are still seeing problems.

chadrik:
@mxm  I'll start working on the serialization tests for this today.  

I haven't tested this recently, but I'm pretty sure that deserialization in Java still does not work because of the LengthPrefix error I described earlier (the Java side seems to expect LengthPrefix, and the python side no longer adds it, since it's difficult to know when to do that automatically).  I'll get you the full stack trace soon.

chadrik:
Pushed a lot of updates to this.  

I fixed up a number of issues and created a new `ImplicitSchemaPayloadBuilder` which covers the use case of determining coders from the provided values, though I'm still not 100% convinced that this is a good idea.  It's fine for python-to-python, but it's going to be brittle if it's intended for expansion via other languages, where expectations are more rigid. 

The tests fail because #9344 is required for the implicit schema generation, so it'd be good to get that PR merged soon.

@mxm this is ready for you to look at why the string decoding is failing on the java side.

robertwb:
Run Python PreCommit

robertwb:
Run RAT PreCommit

robertwb:

chadrik:
Now that all the dependent MRs are in, I can finally demonstrate the coder problem that I was running into.  I'll do that tomorrow.

chadrik:
@mxm I could use your help now!

With the latest from this branch, I'm using the following to expand an external KafkIO xform:

```python
import apache_beam as beam
from apache_beam.io.external.kafka import ReadFromKafka, WriteToKafka

def main(pipeline_options, args):

    pipe = beam.Pipeline(options=pipeline_options)

    (
        pipe
        | 'PubSubInflow' >> ReadFromKafka(
            consumer_config={'foo': 'bar'},
            topics=['this', 'that'])
    )

    result = pipe.run()

    try:
        result.wait_until_finish()
    except KeyboardInterrupt:
        pass
```

And I get this error in the expansion service:

```
[grpc-default-executor-0] INFO org.apache.beam.runners.core.construction.expansion.ExpansionService - Expanding 'root' with URN 'beam:external:java:kafka:read:v1'
Sep 05, 2019 11:33:38 PM org.apache.beam.vendor.grpc.v1p21p0.io.grpc.internal.SerializingExecutor run
SEVERE: Exception while executing runnable org.apache.beam.vendor.grpc.v1p21p0.io.grpc.internal.ServerImpl$JumpToApplicationThreadServerStreamListener$1HalfClosed@31d95dcf
java.lang.RuntimeException: Failed to build transform beam:external:java:kafka:read:v1 from spec urn: "beam:external:java:kafka:read:v1"
payload: "\nl\n\020key_deserializer\022X\n\031beam:coder:string_utf8:v1\022;org.apache.kafka.common.serialization.ByteArrayDeserializer\nn\n\022value_deserializer\022X\n\031beam:coder:string_utf8:v1\022;org.apache.kafka.common.serialization.ByteArrayDeserializer\n\201\001\n\017consumer_config\022n\n\026beam:coder:iterable:v1\n\020beam:coder:kv:v1\n\031beam:coder:string_utf8:v1\n\031beam:coder:string_utf8:v1\022\f\000\000\000\001\003foo\003bar\nM\n\006topics\022C\n\026beam:coder:iterable:v1\n\031beam:coder:string_utf8:v1\022\016\000\000\000\002\004this\004that"

        at org.apache.beam.runners.core.construction.expansion.ExpansionService$ExternalTransformRegistrarLoader.lambda$knownTransforms$0(ExpansionService.java:116)
        at org.apache.beam.runners.core.construction.expansion.ExpansionService$TransformProvider.apply(ExpansionService.java:290)
        at org.apache.beam.runners.core.construction.expansion.ExpansionService.expand(ExpansionService.java:335)
        at org.apache.beam.runners.core.construction.expansion.ExpansionService.expand(ExpansionService.java:372)
        at org.apache.beam.model.expansion.v1.ExpansionServiceGrpc$MethodHandlers.invoke(ExpansionServiceGrpc.java:224)
        at org.apache.beam.vendor.grpc.v1p21p0.io.grpc.stub.ServerCalls$UnaryServerCallHandler$UnaryServerCallListener.onHalfClose(ServerCalls.java:171)
        at org.apache.beam.vendor.grpc.v1p21p0.io.grpc.internal.ServerCallImpl$ServerStreamListenerImpl.halfClosed(ServerCallImpl.java:322)
        at org.apache.beam.vendor.grpc.v1p21p0.io.grpc.internal.ServerImpl$JumpToApplicationThreadServerStreamListener$1HalfClosed.runInContext(ServerImpl.java:762)
        at org.apache.beam.vendor.grpc.v1p21p0.io.grpc.internal.ContextRunnable.run(ContextRunnable.java:37)
        at org.apache.beam.vendor.grpc.v1p21p0.io.grpc.internal.SerializingExecutor.run(SerializingExecutor.java:123)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
        at java.lang.Thread.run(Thread.java:745)
Caused by: org.apache.beam.sdk.coders.CoderException: java.io.EOFException: reached end of stream after reading 58 bytes; 111 bytes expected
        at org.apache.beam.sdk.coders.StringUtf8Coder.decode(StringUtf8Coder.java:104)
        at org.apache.beam.sdk.coders.StringUtf8Coder.decode(StringUtf8Coder.java:90)
        at org.apache.beam.sdk.coders.StringUtf8Coder.decode(StringUtf8Coder.java:37)
        at org.apache.beam.runners.core.construction.expansion.ExpansionService$ExternalTransformRegistrarLoader.populateConfiguration(ExpansionService.java:186)
        at org.apache.beam.runners.core.construction.expansion.ExpansionService$ExternalTransformRegistrarLoader.translate(ExpansionService.java:134)
        at org.apache.beam.runners.core.construction.expansion.ExpansionService$ExternalTransformRegistrarLoader.lambda$knownTransforms$0(ExpansionService.java:113)
        ... 12 more
Caused by: java.io.EOFException: reached end of stream after reading 58 bytes; 111 bytes expected
        at org.apache.beam.vendor.guava.v26_0_jre.com.google.common.io.ByteStreams.readFully(ByteStreams.java:780)
        at org.apache.beam.vendor.guava.v26_0_jre.com.google.common.io.ByteStreams.readFully(ByteStreams.java:762)
        at org.apache.beam.sdk.coders.StringUtf8Coder.readString(StringUtf8Coder.java:60)
        at org.apache.beam.sdk.coders.StringUtf8Coder.decode(StringUtf8Coder.java:100)
        ... 17 more
```

mxm:
@chadrik Fixing one thing at a time would have been easier :) You changed the encoding for strings from raw bytes to using the StringUtf8Coder. Let me have a look.

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
```diff
@@ -1405,24 +1398,24 @@ public void populateDisplayData(DisplayData.Builder builder) {
...
 
         // All byte arrays are UTF-8 encoded strings
-        private Iterable<KV<byte[], byte[]>> producerConfig;
-        private byte[] topic;
-        private byte[] keySerializer;
-        private byte[] valueSerializer;
+        private Iterable<KV<String, String>> producerConfig;
+        private String topic;
+        private String keySerializer;
+        private String valueSerializer;
```
mxm: If you change the types here this will affect the lookup of the configuration fields in ExpansionService. The lookup is performed based on the returned type of the coder. The test still uses the bytes coder, so this will attempt to lookup a `byte[]` field.

mxm: This is for the KafkaIOExternal test which I see failing.

chadrik:
Forget about the tests for a moment, I don’t think they are relevant to the problem I’m seeing. I will fix them shortly.  Try the example code that I sent. 

The original Kafka code was manually handling conversion between utf8 strings and bytes in both python and java, instead of using the proper string coders.  Now that the coders are based on types/schemas in python we are forced to use the correct coder (using bytes in the schema would not have been appropriate because it would have have handled utf8 properly). So fixing this in python forced us to fix it in Java.  

If you think it would be helpful I can make a separate PR to change KafkaIO to use StringUtf8Coder.

sdks/python/apache_beam/transforms/external_test.py
```diff
@@ -48,6 +59,152 @@
...
+      'optional_integer': None,
+  }
+
+  args = {
+      'integer_example': ConfigValue(
+          coder_urn=['beam:coder:varint:v1'],
+          payload=VarIntCoder().encode(values['integer_example'])),
+      'string_example': ConfigValue(
+          coder_urn=['beam:coder:string_utf8:v1'],
+          payload=StrUtf8Coder().encode(values['string_example'])),
```
mxm: This is not sufficient because the Python `StrUtf8Coder` does not add a length prefix while the Java one does. Maybe that is actually a bug. The old version was using a length prefix coder before the bytes coder.

mxm: See https://github.com/apache/beam/blob/9678149872de2799ea1643f834f2bec88d346af8/sdks/python/apache_beam/coders/coders.py#L326

chamikaramj: Difference between Python and Java UTF-8 coders was discussed before in dev list: https://lists.apache.org/thread.html/a3a0d9e7c4196bb6be14ba0bec103209317dcb98e781560eb3ccd48c@%3Cdev.beam.apache.org%3E

Unfortunately I don't believe we resolved this though.

robertwb: This is the nested vs. unnested issue back to bite us again. 

Here we should be using the nested context consistently. For Python, write `coder.get_impl().encode_nested(value)`. Probably would be worth adding an `encode_nested` method to coder itself.

chadrik: @robertwb  Can you provide a bit more explanation for the beam-newbs in the audience (me), please?  Does this just ensure lengths prefixed for applicable types?

robertwb: Basically, ever coder actually represents two concrete encodings, a "nested" one for use in an unbounded input stream (e.g. as a stream of many elements, or within a composite element type like list), and a "outer" one for use when the end-of-record must be provided explicitly (e.g. when writing to a text file, were the newline delimits strings and prefixing each line with a string length would be bad). (In Java this is reified in the second Context argument of the encode/decode functions. For Python it's only visible in the impl layer). 

For some coders, nested and unnested are identical (e.g. varints or doubles, where the element length is implicit in the encoding). For others, a length prefix is added (e.g. utf8 strings and bytes). For others still, it gets pushed down (e.g. for kv coders, the key is always encoded nested, and the value is encoded in the outer context). 

We're trying to move to use the "nested" one everywhere, as this is really confusing (see the linked thread).

mxm: I think we should be good here and elsewhere to use the nested encoding. On the Java side, we need to change the `decode` call in ExpansionServer to use the `NESTED` parameter. The old code here did not need that because it simply wrapped everything in an additional LengthPrefixCoder, which `encode_nested` will essentially do as well. So either reverting back to adding a wrapping LengthPrefixCoder or using encode_nested should work. Note that the latter also requires the additional NESTED parameter on the Java side in ExpansionServer.

chadrik:
Run Java PreCommit

chadrik:
Run Java PreCommit

chadrik:
woohoo!

chadrik:
@robertwb  there's a commit in this PR that makes it possible to specify a test's minimum major _and minor_ version for python.  I use it in this PR to ensure that the dataclasses tests, which rely on syntax changes introduced in python 3.6, and the dataclasses module introduced in python 3.7, are not loaded and run by the python 3.5 tests.  To do so I just extended the regexes a bit.  Let me know if you think this should be in a separate PR.

robertwb:
In this particular case, can you just guard the test from running with a runtime check of sys.version, rather than the boilerplate to exclude tests in the testing scripts? (The py3 stuff was needed for exercising new syntax). 

After that, could you squash to more logical commits, and then I think it's ready to merge. Thanks again for doing this!

mxm:
Thanks you for your persistent work on the PR @chadrik. Will be great to merge this! :)

chadrik:
> In this particular case, can you just guard the test from running with a runtime check of sys.version, rather than the boilerplate to exclude tests in the testing scripts? (The py3 stuff was needed for exercising new syntax).

I can't do a runtime check because this _is_ a syntax change, introduced in python 3.6 (variable annotations, PEP-526) so the py35 tests fail with a syntax error during load.  I figured this solution was simpler (though uglier) than trying to lobby the mailing to drop python 3.5 support :)

robertwb:
Oh, I missed that new syntax...

I would be a fan of dropping 3.5 support, it's security-fix-only,
source-release-only by this point, and might be easier to pull now than
after we have a bunch of Python 3 users. However, I agree that's bigger
than this PR, so it's fine as is.

aaltay:
I will revert this, postcommit tests are failing: https://issues.apache.org/jira/browse/BEAM-8229

aaltay:
Filed an issues for having py 3.x tests as part of precommits: https://issues.apache.org/jira/browse/BEAM-8230

This seems like a simple issue, feel free to fix forward if that can happen quicker than a rollback.

chadrik:
There's a misunderstanding here.  Pre-commit _does_ run the python 3.x tests, but it runs many of the tests on python 3.5.  This PR introduced tests that can only run on python 3.6 or higher, and I had to do some extra work to the beam test framework to make that possible:  https://github.com/apache/beam/pull/9098/commits/0c31f7c8b1c108f7df4cbedfaf9da0ce30092ce0

I appears that same work to properly exclude/include tests at the major + minor version needs to be done for post-commit tests.  I didn't do that simply because I didn't know it could be a problem.  I assumed everything went through tox.

aaltay:
> I appears that same work to properly exclude/include tests at the major + minor version needs to be done for post-commit tests. 

@chadrik I agree this seems be the issue.

robertwb:
Sorry about that. 

This PR has a lot going on in it. Chad, could you pull out the bulk of this PR (minus the data classes, and associated test and test framework changes) and we could get that latter part in via a subsequent PR?

PR #9130, opened 2019-07-23
[BEAM-7802] Expose a method to make an Avro-based PCollection into an Schema-based one

iemejia:
This is just my first PR around Schema/SQL. I found that it was not straight forward to transform an Avro-based PCollection into a Beam schema one so did this. It also does some minor clean ups in AvroUtils and the SQL example.

R: @reuvenlax @kanterov


iemejia:

reuvenlax:
Very cool!

iemejia:
Rebased to fix a merge issue PTAL when you have some time. Notice that there are other Avro / Schema / SQL improvements as part of this but all are isolated in its own commits to ease the review.

sdks/java/core/src/main/java/org/apache/beam/sdk/io/AvroUtils.java
```diff
@@ -1,40 +0,0 @@
...
-package org.apache.beam.sdk.io;
-
-import java.io.Serializable;
-import org.apache.avro.Schema;
-import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Function;
-import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Supplier;
-import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Suppliers;
-
-/** Helpers for working with Avro. */
-class AvroUtils {
```
kanterov: There is one more similar class in `AvroCoder`, but I guess changing it will introduce incompatible change into `AvroCoder` and break streaming pipelines, so let's keep it for now.

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/utils/AvroUtils.java
```diff
@@ -127,7 +127,7 @@ public static FixedBytesField withSize(int size) {
 
     /** Create a {@link FixedBytesField} from a Beam {@link FieldType}. */
     @Nullable
-    public static FixedBytesField fromBeamFieldType(FieldType fieldType) {
+    static FixedBytesField fromBeamFieldType(FieldType fieldType) {
```
kanterov: I see that you want to hide internal APIs, but I'm not sure it worth it at the moment. Everything in this file is an unstable internal API, and it's used outside of Beam codebase to workaround existing limitations. At Spotify, we use a few of these methods, and it seems there are other contributors following this pattern (for instance, https://github.com/apache/beam/commit/172b563fdd36019d3284139417808681314ac364#diff-d7153fa055ebd85f76700232cb4e5cce). I would say for now it's completely fine to keep everything public as is, class is annotated with `@Experimental`

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/utils/AvroUtils.java
```diff
@@ -309,6 +310,18 @@ public static GenericRecord toGenericRecord(
     return g -> toGenericRecord(g, avroSchema);
   }
 
+  /** Transform an existing Avro-based PCollection input into an Schema-based PCollection. */
+  public static <T> PCollection<T> asSchemaPCollection(
+      PCollection<T> pc, Class<T> clazz, @Nullable org.apache.avro.Schema schema) {
+    if (!pc.hasSchema()) {
+      Schema beamSchema = getSchema(clazz, schema);
+      if (beamSchema != null) {
+        pc.setSchema(beamSchema, getToRowFunction(clazz, schema), getFromRowFunction(clazz));
```
kanterov: `pc.setSchema` is equivalent to `pc.setCoder(SchemaCoder.of(...))`

I was thinking, what if we just create a static method to create `SchemaCoder` instead, then user-facing API would be like:
```
KafkaIO.read(MyRecord.class)
  .setCoder(AvroUtils.schemaCoder(MyRecord.class))
```

or, it can be done implicitly by registering classes in advance:

```
p.getSchemaRegistry().registerSchemaProvider(MyRecord.class, new AvroRecordSchema());
```

or for every `SpecificRecord` (didn't try this, but it should work as well):
```
p.getSchemaRegistry().registerSchemaProvider(SpecificRecord.class, new AvroRecordSchema());
```

sdks/java/core/src/main/java/org/apache/beam/sdk/io/AvroUtils.java
iemejia: Ok just for curiosity which class is it?

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/utils/AvroUtils.java
iemejia: Well I think because `@Experimental` is applied to the full class we will probably end up exposing more stuff than we want when we remove the annotation. keeping it with the lowest possible access level is probably wiser and will still allow to evolve access if needed (the other way around won't work). But I can change it if you prefer.

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/utils/AvroUtils.java
iemejia: Yes good idea, that looks cleaner I will add the `schemaCoder` method to AvroUtils, it should cover all the mentioned cases. will ping you back when done. Thanks for the review ideas.

iemejia:
Did suggested updates PTAL @kanterov I added `schemaCoder` methods to match those of `AvroCoder.of` and one additional one to do it from an existing `AvroCoder` that one is relatively simple from the others and I can remove it if you prefer, but I ended up rewriting these two lines multiple times.

iemejia:
I was starting to think if we should somehow make every AvroCoder PCollection a SchemaCoder for user friendliness, but I have my doubts on taking this 'implicit' approach, maybe worried about breaking backwards incompatibility, but somehow it makes sense too, mmm... hard to decide/know.

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/utils/AvroUtils.java
kanterov: I agree, but at this stage, it doesn't seem practical for me, because by reducing access we are going to break existing code and make it problematic to upgrade it to the next Beam version.

kanterov:
Changing AvroCoder will definitely break compatibility, especially streaming pipelines reading from PubSub or Kafka. In addition, SchemaCoder for Avro isn't as good (yet) as AvroCoder. As an example, it would serialize enums as strings, that is very inefficient when shuffling data. Another source of problems is that it doesn't support all Avro features. I believe once it matures we it could be the default, but we aren't there. In any case, I think it's a good exercise to think where we want to put SchemaCoder and how we are going to evolve AvroCoder, so, probably we should start a threat on dev@.

The code looks good. I agree and support your motivation on making fewer things private, but I don't find it practical to break it now given that we know for sure that there are codebases relying on it being public to avoid limitations of existing APIs, so I propose to postpone this before things stabilize.

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/utils/AvroUtils.java
iemejia: I am going to revert the access commit then, but I still don't see the point. The complete `AvroUtils` class is `@Experimental` which means we are in the perfect moment to do refine those aspects. We will have a harder time to do so in the future but well I suppose that's not a big deal.

kanterov: If you ask me, I would rather make the whole class protected (or delete it), hiding everything, and all user-facing functionality will move to AvroCoder

sdks/java/core/src/main/java/org/apache/beam/sdk/io/AvroUtils.java
kanterov: `AvroCoder.SerializableSchemaSupplier`

kanterov: But we can't change it, or it will break Java serialization.

kanterov: Actually, I don't know how it works now, but it uses classes from vendored guava, that changes namespace each time we change guava version. I'm wondering if we already broke it without noticing.

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/utils/AvroUtils.java
iemejia: Yes you are right but that class does a lot other magic too. We should probably decide on that later on, but I think is important because these are user friendly fixes and the current uber class hides multiple things.

sdks/java/core/src/main/java/org/apache/beam/sdk/io/AvroUtils.java
kanterov: I asked on dev, but probably we already broke it a couple of times

sdks/java/core/src/main/java/org/apache/beam/sdk/io/AvroIO.java
```diff
@@ -186,6 +188,49 @@
  * scalability. Note that it may decrease performance if the filepattern matches only a small number
  * of files.
  *
+ * <h3>Inferring Beam schemas from Avro files</h3>
+ *
+ * <p>If you want to use SQL or schema based operations on an Avro-based PCollection. You must
```
RyanSkraba:
```suggestion
 * <p>If you want to use SQL or schema based operations on an Avro-based PCollection, you must
```

sdks/java/core/src/main/java/org/apache/beam/sdk/io/AvroIO.java
```diff
@@ -186,6 +188,49 @@
...
+ *
+ * <pre>{@code
+ * PCollection<AvroAutoGenClass> records =
+ *     p.apply(AvroIO.read(...).from(...).withBeamSchemas(true);
+ * }</pre>
+ *
+ * <h3>Inferring Beam schemas from Avro PCollections</h3>
+ *
+ * <p>If you created an Avro-based PCollection by other means e.g. reading records from Kafka or as
+ * the output of another PTransform. You may be interested on making your PCollection schema-aware
```
RyanSkraba:
```suggestion
 * the output of another PTransform, you may be interested on making your PCollection schema-aware
```

sdks/java/core/src/main/java/org/apache/beam/sdk/io/AvroIO.java
```diff
@@ -875,6 +926,13 @@ public void populateDisplayData(DisplayData.Builder builder) {
           recordClass,
           schemaSupplier.get());
     }
+
+    private static class JsonToSchema implements Function<String, Schema>, Serializable {
```
RyanSkraba: AVRO-1852 to the rescue!  Just for reference: schemas are directly serializable from Avro 1.9.1+ (sharing the happiness!)

sdks/java/core/src/main/java/org/apache/beam/sdk/io/AvroIO.java
iemejia: done

sdks/java/core/src/main/java/org/apache/beam/sdk/io/AvroIO.java
iemejia: done

sdks/java/core/src/main/java/org/apache/beam/sdk/io/AvroIO.java
```diff
@@ -186,6 +188,49 @@
...
+ * <p>If you created an Avro-based PCollection by other means e.g. reading records from Kafka or as
+ * the output of another PTransform. You may be interested on making your PCollection schema-aware
+ * so you can use the Schema-based APIs or Beam's SqlTransform.
+ *
+ * <p>If you are using Avro specific records (generated classes from an Avro schema), you can
+ * register a schema provider for the specific Avro class to make any PCollection of these objects
+ * schema-aware.
+ *
+ * <pre>{@code
+ * pipeline.getSchemaRegistry().registerSchemaProvider(AvroAutoGenClass.class, new AvroRecordSchema());
```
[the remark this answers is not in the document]
iemejia: good one, definitely simpler

sdks/java/core/src/main/java/org/apache/beam/sdk/io/AvroIO.java
iemejia: :) yaaayy !

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/utils/AvroUtils.java
iemejia: I am letting things as they were and resolving this one.

iemejia:
@kanterov I restored the access modifiers and let everything as suggested.

I think we should merge this as it is and open the discussion in dev@ since the classes are still experimental we can still adapt the changes we conclude from the discussion and ‘experimental’ users can benefit of having this available in the meantime. WDYT?

kanterov:
@iemejia great, agree, please feel free to merge when PreCommit check passes

iemejia:
Run Java PreCommit

iemejia:
Run Java PreCommit

iemejia:
Merging since it is green now. Thanks for the review @kanterov and @RyanSkraba

PR #9131, opened 2019-07-23
[BEAM-7794] DynamoDBIOTest is blocking forever

iemejia:
This just disables the blocking tests mentioned in the ML

R: @akedin 
CC: @cmachgodaddy


iemejia:
Anton can you check this works in your machine, we disabled only the tests but it may be that the issue is in the `@BeforeClass` / `@AfterClass` methods, if that's the case I will do a PR to disable the full test.

akedin:
Will check in a second, thank you for taking a look

akedin:
Ignored these tests on my machine, it still hangs with the same symptoms. Ignored the whole file - passes, so it does look like it comes from `@Before/After/Rule`.

Unrelated to this change, I also see `:sdks:java:io:amazon-web-services:spotbugsMain` failing to connect to a gradle process (all other `spotbugsMains` seems to have finished), then times out and fails the build.

iemejia:
Arrghh I accidentally pushed to the release branch the updated version of the cherry pick of this PR. Sorry.
Can you help me confirm it does not get stucked anymore @akedin. Thanks

akedin:
Works on release branch for me now. Thank you!

PR #9139, opened 2019-07-24
[BEAM-7794] Ignore blocking DynamoDBIOtest class

iemejia:
R: @jbonofre

### DIFF
--- a/sdks/java/io/amazon-web-services/src/test/java/org/apache/beam/sdk/io/aws/dynamodb/DynamoDBIOTest.java
+++ b/sdks/java/io/amazon-web-services/src/test/java/org/apache/beam/sdk/io/aws/dynamodb/DynamoDBIOTest.java
@@ -48,6 +48,7 @@ import org.junit.rules.ExpectedException;
 import org.mockito.Mockito;
 
 /** Test Coverage for the IO. */
+@Ignore("[BEAM-7794] DynamoDBIOTest is blocking forever")
 public class DynamoDBIOTest implements Serializable {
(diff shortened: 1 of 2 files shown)


sdks/java/io/amazon-web-services/src/test/java/org/apache/beam/sdk/io/aws/dynamodb/DynamoDBIOTest.java
```diff
@@ -48,6 +48,7 @@
 import org.mockito.Mockito;
 
 /** Test Coverage for the IO. */
+@Ignore
```
jbonofre: Maybe worth to add comment of the reason for the ignore, with something like `@Ignore("reason is ...")`

jbonofre:
I'm waiting for Jenkins before merging.

iemejia:

sdks/java/io/amazon-web-services/src/test/java/org/apache/beam/sdk/io/aws/dynamodb/DynamoDBIOTest.java
iemejia: Good point, updated now.

iemejia:

iemejia:
Error is unrelated to this PR, I am merging eagerly because I need to cherry pick it in the 2.14 branch. Thanks JB !

jbonofre:
+1, thanks !

PR #9145, opened 2019-07-24
[BEAM-1009] Update multiple modules to use Mockito 2

iemejia:
I think it is better to start doing this incrementally otherwise we will probably never get it fully.
Notice that I did not choose a more recent version of mockito because I got a conflict with [this issue](https://github.com/mockito/mockito/issues/1606).

Notice also that I did not go straight to Mockito 3 because Powermock have issues with it too, but we can tackle that in the future (since API compatibility is higher between 2 and 3, that will be easier).

R: @lukecwik


iemejia:

iemejia:

iemejia:

iemejia:

iemejia:

tvalentyn:
@lukecwik was OOO recently and might have missed this. Consider resolving the conflict and bumping this up.

lukecwik:
I'm back from vacation and taking a look.

lukecwik:
FYI, #9338 is updating to Mockito 3

PR #7842, opened 2019-02-14
[BEAM-5731] Only accept key operator in Python 3.

robertwb:
This is in line with Python's change to its sorting operations.

------------------------

Thank you for your contribution! Follow this checklist to help us incorporate your contribution quickly and easily:

 - [ ] [**Choose reviewer(s)**](https://beam.apache.org/contribute/#make-your-change) and mention them in a comment (`R: @username`).
 - [ ] Format the pull request title like `[BEAM-XXX] Fixes bug in ApproximateQuantiles`, where you replace `BEAM-XXX` with the appropriate JIRA issue, if applicable. This will automatically link the pull request to the issue.
 - [ ] If this contribution is large, please file an Apache [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

Lang | SDK | Apex | Dataflow | Flink | Gearpump | Samza | Spark
--- | --- | --- | --- | --- | --- | --- | ---
Go | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go/lastCompletedBuild/) | --- | --- | --- | --- | --- | ---
Java | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Apex/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Dataflow/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Flink/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Flink_Batch/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Flink_Streaming/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Gearpump/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Samza/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Spark/lastCompletedBuild/)
[...]

PR #9153, opened 2019-07-25
[BEAM-6694] Added Approximate Quantile Transfrom on Python SDK

mszb:
Added PTransforms for getting an idea of a PCollection's data distribution using approximate N-tiles (e.g. quartiles, percentiles, etc.), globally & per-key.

It offers the same API as its Java counterpart 
https://github.com/apache/beam/blob/master/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/ApproximateQuantiles.java

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/transforms/stats.py
+++ b/sdks/python/apache_beam/transforms/stats.py
@@ -221,7 +221,7 @@ class ApproximateUniqueCombineFn(CombineFn):
     except Exception as e:
       raise RuntimeError("Runtime exception: %s", e)
 
-  # created an issue https://issues.apache.org/jira/browse/BEAM-7285 to speep up
+  # created an issue https://issues.apache.org/jira/browse/BEAM-7285 to speed up
   # merge process.
   def merge_accumulators(self, accumulators, *args, **kwargs):
     merged_accumulator = self.create_accumulator()
@@ -489,7 +489,6 @@ class ApproximateQuantilesCombineFn(CombineFn):
                      _QuantileBuffer(elements=qs.unbuffered_elements))
       qs.unbuffered_elements = []
       self._collapse_if_needed(qs)
-    return qs
 
   def _offset(self, newWeight):
     """
@@ -587,8 +586,8 @@ class ApproximateQuantilesCombineFn(CombineFn):
       quantile_state.min_val = element
     elif self._comparator(element, quantile_state.max_val) > 0:
       quantile_state.max_val = element
-    qs = self._add_unbuffered(quantile_state, elem=element)
-    return qs
+    self._add_unbuffered(quantile_state, elem=element)
+    return quantile_state
 
   def merge_accumulators(self, accumulators):
     """Merges all the accumulators (quantile state) as one."""
@@ -604,7 +603,7 @@ class ApproximateQuantilesCombineFn(CombineFn):
         qs.max_val = accumulator.max_val
 
       for unbuffered_element in accumulator.unbuffered_elements:
-        qs = self._add_unbuffered(qs, unbuffered_element)
+        self._add_unbuffered(qs, unbuffered_element)
 
       qs.buffers.extend(accumulator.buffers)
     self._collapse_if_needed(qs)
--- a/sdks/python/apache_beam/transforms/stats.py
+++ b/sdks/python/apache_beam/transforms/stats.py
@@ -21,6 +21,7 @@ from __future__ import absolute_import
 from __future__ import division
 
 import heapq
+import itertools
 import math
 import sys
 import typing
@@ -29,9 +30,11 @@ from builtins import round
 from apache_beam import coders
 from apache_beam import typehints
 from apache_beam.transforms.core import *
+from apache_beam.transforms.display import DisplayDataItem
 from apache_beam.transforms.ptransform import PTransform
 
 __all__ = [
+    'ApproximateQuantiles',
     'ApproximateUnique',
 ]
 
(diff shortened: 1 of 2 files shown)


mszb:
retest this please

aaltay:
R: @Hannah-Jiang could you make this first pass review here since you worked on the similar  ApproximateUnique transform.

mszb:

sdks/python/apache_beam/transforms/combiners_test.py
```diff
@@ -481,5 +484,315 @@ def test_with_input_types_decorator_violation(self):
         _ = pc | beam.CombineGlobally(self.fn)
 
 
+class ApproximateQuantilesTest(unittest.TestCase):
```
Hannah-Jiang: This class should go `stats_test.py`.

sdks/python/apache_beam/transforms/combiners.py
```diff
@@ -924,3 +928,405 @@ def merge_accumulators(self, accumulators):
...
+      self.reverse = reverse
+
+    def expand(self, pcoll):
+      return pcoll | core.CombinePerKey(ApproximateQuantilesCombineFn.create(
+          num_quantiles=self.num_quantiles, compare=self.compare,
+          key=self.key, reverse=self.reverse
+      ))
+
+    def display_data(self):
+      return {
```
Hannah-Jiang: We return the same format for Globally and PerKey, can we share the format?

sdks/python/apache_beam/transforms/combiners.py
```diff
@@ -924,3 +928,405 @@ def merge_accumulators(self, accumulators):
...
+      }
+
+
+class _QuantileBuffer(object):
+  """A single buffer in the sense of the referenced algorithm."""
+
+  def __init__(self, elements, level=None, weight=None):
+    self.elements = elements
+    self.level = level or 0
+    self.weight = weight or 1
```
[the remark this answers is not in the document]
mszb: Done!

sdks/python/apache_beam/transforms/combiners.py
```diff
@@ -924,3 +928,405 @@ def merge_accumulators(self, accumulators):
...
+        converted to key function as Python 3 does not support cmp.
+      key: (optional) Key is  a mapping of elements to a comparable key, similar
+        to the key argument of Python's sorting methods.
+    """
+
+    def __init__(self, num_quantiles, compare=None, key=None, reverse=False):
+      self.num_quantiles = num_quantiles
+      self.compare = compare
+      self.key = key
+      self.reverse = reverse
```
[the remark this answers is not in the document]
mszb: Done!

sdks/python/apache_beam/transforms/combiners.py
```diff
@@ -924,3 +928,405 @@ def merge_accumulators(self, accumulators):
 
   def extract_output(self, accumulator):
     return accumulator[0]
+
+
+class ApproximateQuantiles(object):
```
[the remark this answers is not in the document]
mszb: Done!

sdks/python/apache_beam/transforms/combiners_test.py
mszb: Done!

sdks/python/apache_beam/transforms/combiners.py
mszb: Done!

mszb:
Hi @Hannah-Jiang. Thanks for reviewing my PR and the feedbacks. I've made some changes you mentioned. Can you please take a look? Thank you.

mszb:
retest this please

mszb:

Hannah-Jiang:
This looks good.
A minor point is can we add more test cases with different compare functions?
For example, compare with absolute values, compare with string, pass a customized function to compare (instead of a simple a,b comparison). Now we compare strings by length, but I think in real practice, there are many cases users want to compare by string values or pass a customized compare functions.

mszb:
Hi @Hannah-Jiang. Thanks for the feedback. There is a test case `test_alternate_comparator()` in which user get the results as per the length of the string. But I can see your point, let me add some more.

sdks/python/apache_beam/transforms/stats.py
```diff
@@ -234,3 +239,419 @@ def extract_output(accumulator):
...
+  """
+
+  @staticmethod
+  def _display_data(num_quantiles, compare, key, reverse):
+    return {
+        'num_quantiles': DisplayDataItem(num_quantiles, label="Quantile Count"),
+        'compare': DisplayDataItem(compare.__class__,
+                                   label='Record Comparer FN'),
+        'key': DisplayDataItem(key.__class__, label='Record Comparer Key'),
+        'reverse': DisplayDataItem(reverse.__class__, label='Is reversed')
```
aaltay: reverse is a boolean. We would like to capture whether it is True or False, instead of it being a type bool.

Similarly for compare, __class__ will be of type None or function and that is less helpful. (For that data could be compare.__name__ for example if it is not None.)

Similarly for key,  we should try to have some meaningful thing to display to users.

sdks/python/apache_beam/transforms/stats.py
```diff
@@ -234,3 +239,419 @@ def extract_output(accumulator):
...
+          num_quantiles=self._num_quantiles, compare=self._compare,
+          key=self._key, reverse=self._reverse))
+
+    def display_data(self):
+      return ApproximateQuantiles._display_data(
+          num_quantiles=self._num_quantiles, compare=self._compare,
+          key=self._key, reverse=self._reverse)
+
+  @typehints.with_input_types(typing.Tuple[K, V])
+  @typehints.with_output_types(typing.List[typing.Tuple[K, V]])
```
aaltay: Should this be: `[typing.Tuple[K, typing.List[V]]` instead?

Or at least docstring seems to imply that.

sdks/python/apache_beam/transforms/stats.py
```diff
@@ -234,3 +239,419 @@ def extract_output(accumulator):
...
+          key=self._key, reverse=self._reverse))
+
+    def display_data(self):
+      return ApproximateQuantiles._display_data(
+          num_quantiles=self._num_quantiles, compare=self._compare,
+          key=self._key, reverse=self._reverse)
+
+
+class _QuantileBuffer(object):
+  """A single buffer in the sense of the referenced algorithm."""
```
aaltay: Is there a link we can add to the referenced algorithm?

sdks/python/apache_beam/transforms/stats.py
```diff
@@ -234,3 +239,419 @@ def extract_output(accumulator):
...
+    # into new, full buffers and then take them into account when computing the
+    # final output.
+    self.unbuffered_elements = unbuffered_elements
+
+  def is_empty(self):
+    """Check if the buffered & unbuffered elements are empty or not."""
+    return not self.unbuffered_elements and not self.buffers
+
+
+class ApproximateQuantilesCombineFn(CombineFn):
```
aaltay: Should this start with an underscore? Do expect people to use it directly?

sdks/python/apache_beam/transforms/stats.py
```diff
@@ -234,3 +239,419 @@ def extract_output(accumulator):
...
+  # non-optimal. The impact is logarithmic with respect to this value, so this
+  # default should be fine for most uses.
+  _MAX_NUM_ELEMENTS = 1e9
+  _qs = None  # Refers to the _QuantileState
+
+  # TODO(mszb): For Python 3, remove compare and only keep key.
+  def __init__(self, num_quantiles, buffer_size, num_buffers, compare=None,
+               key=None, reverse=False):
+    if compare:
+      warnings.warn('Compare is not available in Python 3, use key instead.')
```
aaltay: Should this also be an exception if we are running in py3 environment?

sdks/python/apache_beam/transforms/stats.py
```diff
@@ -234,3 +239,419 @@ def extract_output(accumulator):
...
+  # The cost (in time and space) to compute quantiles to a given accuracy is a
+  # function of the total number of elements in the data set. If an estimate is
+  # not known or specified, we use this as an upper bound. If this is too low,
+  # errors may exceed the requested tolerance; if too high, efficiency may be
+  # non-optimal. The impact is logarithmic with respect to this value, so this
+  # default should be fine for most uses.
+  _MAX_NUM_ELEMENTS = 1e9
+  _qs = None  # Refers to the _QuantileState
+
+  # TODO(mszb): For Python 3, remove compare and only keep key.
```
aaltay: Should we not introduce compare at all if this is only useful for python 2? Also from the use below, where we use cmp_to_key to do the mapping, is not py2 and py3 compatible?

/cc @tvalentyn to help with py3 questions.

sdks/python/apache_beam/transforms/stats.py
```diff
@@ -234,3 +239,419 @@ def extract_output(accumulator):
...
+      key = lambda elem: cmp_to_key(self._comparator)(elem)
+
+    self._compare = compare
+    self._num_quantiles = num_quantiles
+    self._buffer_size = buffer_size
+    self._num_buffers = num_buffers
+    self._key = key
+
+  @classmethod
+  def create(cls, num_quantiles, epsilon=None, max_num_elements=None,
```
aaltay: We do not seem to be exposing epsilon and max_num_elements from the transform. How often these parameters would be used? Is defaults good enough for most cases?

sdks/python/apache_beam/transforms/stats.py
```diff
@@ -234,3 +239,419 @@ def extract_output(accumulator):
...
+    elif self._comparator(element, quantile_state.min_val) < 0:
+      quantile_state.min_val = element
+    elif self._comparator(element, quantile_state.max_val) > 0:
+      quantile_state.max_val = element
+    self._add_unbuffered(elem=element)
+    return self._qs
+
+  def merge_accumulators(self, accumulators):
+    """Merges all the accumulators (quantile state) as one."""
+    self._qs = self.create_accumulator()
```
aaltay: merge_accumulators is only allowed to mutate `accumulators[0]`. It should not mutate itself unless that is the case (See: https://github.com/apache/beam/blob/master/sdks/python/apache_beam/transforms/core.py#L736)

Safest option would be to change `self._qs` to `qs`.

sdks/python/apache_beam/transforms/stats_test.py
```diff
@@ -345,5 +353,313 @@ def test_approximate_unique_globally_by_error_with_skewed_data(self):
...
+    expected_items = [
+        DisplayDataItemMatcher('num_quantiles', aq._num_quantiles),
+        DisplayDataItemMatcher('compare', aq._compare.__class__),
+        DisplayDataItemMatcher('key', aq._key.__class__),
+        DisplayDataItemMatcher('reverse', aq._reverse.__class__)
+    ]
+    hc.assert_that(data.items, hc.contains_inanyorder(*expected_items))
+
+
+def _build_quantilebuffer_test_data():
```
aaltay: Is there a bigger test that we can run to see if we produce right results in a larger data set. There is quite a bit of a math here and it would be good if we can test it.

sdks/python/apache_beam/transforms/stats_test.py
```diff
@@ -345,5 +353,313 @@ def test_approximate_unique_globally_by_error_with_skewed_data(self):
     pipeline.run()
 
 
+class ApproximateQuantilesTest(unittest.TestCase):
```
aaltay: Is it possible to run a test (even a one off) on Dataflow to see if this is working end to end?

sdks/python/apache_beam/transforms/stats.py
tvalentyn: On Python 3 we should make the interface consistent with sorted() builtin on Python 3 from the beginning, so that we don't have to deprecate `compare` later. that is, we should allow only `key` and `reverse`, and drop `compare` on Python 3. This will be consistent with `Top()` combiners which underwent a similar change in https://github.com/apache/beam/pull/7842. Given that this is a new transform, I suggest we drop compare across the board, since it will ease migration to Python 3. 
For context, we plan to drop Python 2 support in Beam starting from 2020.

sdks/python/apache_beam/transforms/stats.py
mszb: Yes, the document reference is mentioned on the `ApproximateQuantilesCombineFn` docstring.
`
[MRL98] Manku, Rajagopalan & Lindsay, "Approximate Medians and other
Quantiles in One Pass and with Limited Memory", Proc. 1998 ACM SIGMOD,
Vol 27, No 2, p 426-435, June 1998.
http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.6.6513&rep=rep1&type=pdf
`

I'll update this doc & clearly mention here where to find the document reference.

sdks/python/apache_beam/transforms/stats.py
mszb: This class can be used directly if someone wants to use this with custom error bound rate (`epsilon`) or change the size of `max_input_elements`. I think it's good to have if the user can customize these parameters for some cases if they like.

sdks/python/apache_beam/transforms/stats.py
mszb: Yes, because we want to encourage them to use `key` instead.

sdks/python/apache_beam/transforms/stats.py
mszb: Default values are good enough for most of the cases. But there might be the cases where a user wants to customize the maximum input size and epsilon.

sdks/python/apache_beam/transforms/stats_test.py
mszb: The test data is taken from **Munro-Paterson Algorithm** *(http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.6.6513&rep=rep1&type=pdf)* In which we create buffer size from 96 to 15,259 max.

By the way, it does not create any buffer, just calculates the sizes!

sdks/python/apache_beam/transforms/stats.py
mszb: So @aaltay, as @tvalentyn suggest, should drop the compare support?

sdks/python/apache_beam/transforms/stats_test.py
mszb: I ran an example on dataflow and its working as expected.
https://console.cloud.google.com/dataflow/jobsDetail/locations/us-central1/jobs/2019-08-01_05_54_51-13352737957197087204?project=apache-beam-testing

sdks/python/apache_beam/transforms/stats.py
mszb: Done!

sdks/python/apache_beam/transforms/stats.py
aaltay: Let's drop compare support and go with @tvalentyn's suggestion.

sdks/python/apache_beam/transforms/stats_test.py
aaltay: Great!

sdks/python/apache_beam/transforms/stats.py
aaltay: Sounds good.

sdks/python/apache_beam/transforms/stats.py
aaltay: Sounds good.

sdks/python/apache_beam/transforms/stats_test.py
aaltay: Let's add a large one, that does not require a new test data file. How about something like 256-512?

mszb: I didn't get your point which says, _"that does not require a new **test data file**"_  & ... _"**256-512**"_.

Can you please elaborate?

aaltay: Sorry for the confusion. I meant:
* Let's use a buffer size between 256-512. Pick any number that works for you in that range. How about 512 if that works?
* I wanted to keep the size small enough, so that it can be added right here in the stats_test.py file. Otherwise, we will probably need to have a separate data file to keep the test information. I wanted to avoid that case.

mszb: Hi Ahmet.
As discussed, the `_build_quantilebuffer_test_data` method returns a list containing 25 items (which are the test scenarios):
```
[
[0.1, 100000, 11, 98], 
[0.1, 1000000, 14, 123], 
[0.1, 10000000, 17, 153], 
[0.1, 100000000, 21, 96], 
[0.1, 1000000000, 24, 120], 
[0.05, 100000, 11, 98], 
[0.05, 1000000, 14, 123], 
[0.05, 10000000, 17, 153], 
[0.05, 100000000, 20, 191], 
[0.05, 1000000000, 23, 239], 

[0.01, 100000, 9, 391], 

[0.01, 1000000, 11, 977], 
[0.01, 10000000, 14, 1221], 
[0.01, 100000000, 17, 1526], 
[0.01, 1000000000, 21, 954], 
[0.005, 100000, 8, 782], 
[0.005, 1000000, 11, 977], 
[0.005, 10000000, 14, 1221], 
[0.005, 100000000, 17, 1526], 
[0.005, 1000000000, 20, 1908], 
[0.001, 100000, 6, 3125], 
[0.001, 1000000, 9, 3907], 
[0.001, 10000000, 11, 9766], 
[0.001, 100000000, 14, 12208], 
[0.001, 1000000000, 17, 15259]
]
```

where each tuple holds epsilon, maxInputSize, numBuffers, bufferSize in that order. (These tuples are passed into test_efficiency and test_correctness as a parameterized.expand)

The range of values of bufferSize is 98 (smallest) to 15259 (largest).

The buffer size range you mentioned (256-512) is also represented in this generated test dataset.

And the `ApproximateQuantilesBufferTest` takes around 80 milliseconds to complete on my machine!

sdks/python/apache_beam/transforms/stats.py
mszb: Done!

sdks/python/apache_beam/transforms/stats.py
mszb: Done!

sdks/python/apache_beam/transforms/stats_test.py
```diff
@@ -345,5 +353,273 @@ def test_approximate_unique_globally_by_error_with_skewed_data(self):
...
+    while k < num_of_quantiles - 1:
+      expected = (size - 1) * k / (num_of_quantiles - 1)
+      quantiles.append([expected - absoluteError, expected + absoluteError])
+      k = k + 1
+    quantiles.append(size - 1)
+    return quantiles
+
+  def test_quantiles_globaly(self):
+    with TestPipeline() as p:
+      pc = p | Create(range(101))
```
aaltay: Let's change this to an actual list instead of a range object. The reason,  if in the future we convert this to a validates runner test, we would need to do the same `from builtins import range` in the workers.

Suggestion:
`pc = p | Create([x for x in range(101)])`

sdks/python/apache_beam/transforms/stats_test.py
```diff
@@ -345,5 +353,273 @@ def test_approximate_unique_globally_by_error_with_skewed_data(self):
...
+      quantiles = pc | beam.ApproximateQuantiles.Globally(5)
+      assert_that(quantiles, equal_to([[0, 25, 50, 75, 100]]))
+
+  def test_quantiles_globaly_reversed(self):
+    with TestPipeline() as p:
+      data = range(101)
+      pc = p | Create(data)
```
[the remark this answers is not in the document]
mszb: Yes, you are right. I've updated the test cases.

sdks/python/apache_beam/transforms/stats_test.py
mszb: I've changed it to `pc = p | Create(list(range(101)))`. As `list` will be a bit faster as compared to list comprehension which loops through each element!

PR #9156, opened 2019-07-25
[BEAM-7495] Add fine-grained progress reporting

aryann:
The current progress reporting code reports changes in progress once each server
response is consumed. With this change, we interpolate the progress between the
last response consumed and the current one, causing progress to go up with every
call to #advance.

Progress reporting allows Google Cloud Dataflow to make more intelligent
decisions during dynamic worker rebalancing.

------------------------

--- | --- | --- | --- | ---


aryann:
R: @chamikaramj

FYI: @kmjung @mmladenovski

kmjung:
You might want to look at how progress estimation is implemented for block-based sources:

https://github.com/apache/beam/blob/64c62b1e42a84746e7aa97dddfa1fce95919a6b2/sdks/java/core/src/main/java/org/apache/beam/sdk/io/BlockBasedSource.java

kmjung:
I don't see any correctness issues. My only question would be whether you might want to move the computation of the exact fractional position into getFractionConsumed so that it isn't (necessarily) done for every row.

aryann:
> You might want to look at how progress estimation is implemented for block-based sources:
> 
> https://github.com/apache/beam/blob/64c62b1e42a84746e7aa97dddfa1fce95919a6b2/sdks/java/core/src/main/java/org/apache/beam/sdk/io/BlockBasedSource.java
> I don't see any correctness issues. My only question would be whether you might want to move the computation of the exact fractional position into getFractionConsumed so that it isn't (necessarily) done for every row.

Thanks for the pointer to the other source. Our source doesn't have an isStarted or isDone. I don't think we should add those methods just to localize the computation in getFractionConsumed because that's more coordination that we have to get right. Additionally, we have to update a number of fields in readNextRecord, like num rows read from response, anyway, so I think the code is clearer if the computation of the fractional value is next to the code that updates those fields. Otherwise, you'd have to hop between isStarted, isDone, readNextRow, and getFractionConsumed to understand how the value is computed.

aryann:

kmjung:
LGTM.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryStorageStreamSource.java
```diff
@@ -210,20 +218,28 @@ private synchronized boolean readNextRecord() throws IOException {
...
 
       record = datumReader.read(record, decoder);
       current = parseFn.apply(new SchemaAndRecord(record, tableSchema));
+
+      rowsReadFromCurrentResponse++;
+      fractionConsumed =
+          fractionConsumedFromPreviousResponse
+              + (fractionConsumedFromCurrentResponse - fractionConsumedFromPreviousResponse)
+                  * rowsReadFromCurrentResponse
+                  * 1.0
```
chamikaramj: Please add a comment describing this.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryStorageStreamSource.java
```diff
@@ -210,20 +218,28 @@ private synchronized boolean readNextRecord() throws IOException {
...
+                  / rowCountFromCurrentResponse;
```
chamikaramj: Should this (rowCountFromCurrentResponse) be totalRowCountOfCurrentResponse ?

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryStorageStreamSource.java
```diff
@@ -210,20 +218,28 @@ private synchronized boolean readNextRecord() throws IOException {
...
+        fractionConsumedFromPreviousResponse = fractionConsumedFromCurrentResponse;
         ReadRowsResponse nextResponse = responseIterator.next();
         decoder =
             DecoderFactory.get()
                 .binaryDecoder(
                     nextResponse.getAvroRows().getSerializedBinaryRows().toByteArray(), decoder);
-        fractionConsumedFromLastResponse = getFractionConsumed(nextResponse);
+        rowsReadFromCurrentResponse = 0L;
+        rowCountFromCurrentResponse = nextResponse.getAvroRows().getRowCount();
+        fractionConsumedFromCurrentResponse = getFractionConsumed(nextResponse);
```
chamikaramj: I think we need to adjust variable names here. Current naming makes it pretty confusing. Should this "nextResponse" be named "currentResponse" since it result in "fractionConsumedFromCurrentResponse" ?

aryann: Done.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryStorageStreamSource.java
aryann: That is a clearer name. Done.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryStorageStreamSource.java
aryann: Done.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryStorageStreamSource.java
```diff
@@ -210,20 +218,28 @@ private synchronized boolean readNextRecord() throws IOException {
...
-        // consequential.
-        fractionConsumed = fractionConsumedFromLastResponse;
+        fractionConsumedFromPreviousResponse = fractionConsumedFromCurrentResponse;
         ReadRowsResponse nextResponse = responseIterator.next();
         decoder =
             DecoderFactory.get()
                 .binaryDecoder(
                     nextResponse.getAvroRows().getSerializedBinaryRows().toByteArray(), decoder);
-        fractionConsumedFromLastResponse = getFractionConsumed(nextResponse);
+        rowsReadFromCurrentResponse = 0L;
```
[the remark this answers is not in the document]
aryann: Done.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryStorageStreamSource.java
```diff
@@ -165,8 +165,13 @@ public String toString() {
     private GenericRecord record;
     private T current;
     private long currentOffset;
+
+    // Values used for progress reporting.
     private double fractionConsumed;
-    private double fractionConsumedFromLastResponse;
+    private double fractionConsumedFromPreviousResponse;
+    private double fractionConsumedFromCurrentResponse;
+    private long rowsReadFromCurrentResponse;
```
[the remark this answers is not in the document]
aryann: I purposefully chose long variable names to avoid additional comments.

If you don't feel that the names are sufficiently self-describing, I'm happy to add more comments. Please let me know!

aryann:
@chamikaramj, please take another look!

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryStorageStreamSource.java
```diff
@@ -210,20 +218,35 @@ private synchronized boolean readNextRecord() throws IOException {
...
                 .binaryDecoder(
-                    nextResponse.getAvroRows().getSerializedBinaryRows().toByteArray(), decoder);
-        fractionConsumedFromLastResponse = getFractionConsumed(nextResponse);
+                    currentResponse.getAvroRows().getSerializedBinaryRows().toByteArray(), decoder);
+
+        // Since we now have a new response, reset the row counter for the current response.
+        rowsReadFromCurrentResponse = 0L;
+
+        totalRowCountFromCurrentResponse = currentResponse.getAvroRows().getRowCount();
+        fractionConsumedFromCurrentResponse = getFractionConsumed(currentResponse);
```
chamikaramj: Please add asserts to make sure that "fractionConsumedFromCurrentResponse > fractionConsumedFromPreviousResponse" and that "0 <= fractionConsumedFromCurrentResponse <= 1"

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryStorageStreamSource.java
```diff
@@ -210,20 +218,35 @@ private synchronized boolean readNextRecord() throws IOException {
...
+      // consumed value from the previous server response (or zero if we're consuming the first
+      // response) and by interpolating the fractional value in the current response based on how
+      // many rows have been consumed.
+      rowsReadFromCurrentResponse++;
+      fractionConsumed =
+          fractionConsumedFromPreviousResponse
+              + (fractionConsumedFromCurrentResponse - fractionConsumedFromPreviousResponse)
+                  * rowsReadFromCurrentResponse
+                  * 1.0
+                  / totalRowCountFromCurrentResponse;
```
chamikaramj: What if totalRowCountFromCurrentResponse == 0 ? Please add an assert if this can never happen.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryStorageStreamSource.java
aryann: Done.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryStorageStreamSource.java
aryann: It should never be 0. Added precondition check above.

aryann:
@chamikaramj, please take another look!

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryStorageStreamSource.java
```diff
@@ -210,20 +219,50 @@ private synchronized boolean readNextRecord() throws IOException {
...
 
       record = datumReader.read(record, decoder);
       current = parseFn.apply(new SchemaAndRecord(record, tableSchema));
+
+      // Updates the fraction consumed value. This value is calculated by summing the fraction
+      // consumed value from the previous server response (or zero if we're consuming the first
+      // response) and by interpolating the fractional value in the current response based on how
+      // many rows have been consumed.
+      rowsReadFromCurrentResponse++;
+      fractionConsumed =
```
chamikaramj: Sorry one more comment. 
In practice, readNextRecord() will get called much more often that getFractionConsumed(). Latter is called by a secondary progress reporting thread (for Dataflow). So can we please move this computation to the getFractionConsumed() method ?

aryann: I purposefully put this computation in readNextRecord(). Ken pointed this out in an earlier comment.

This is clearer because (1) the computation is close to the code that updates the variables involved; and (2) moving this to getFractionConsumed() makes computing the 1.0 for end of stream a little trickier because you'll either need another field to signal the end of the stream or you fiddle with the existing vars so the computation leads to 1.0--both of these schemes make the code a little harder to follow than what we have now (i.e., "fractionConsumed = 1d").

I don't think we should worry about whether this is efficient or not. This is a pretty cheap numerical computation. I suspect the AVRO parsing for each row is going to dominate the amount of time spent (... in the case where no I/O takes place).

chamikaramj: Ok. Sg.

chamikaramj:
LGTM. Thanks.

chamikaramj:
Run Java PreCommit

kmjung:
Please hold off on merging this for now. It sounds like Aryan is concerned
that there may be an issue related to splitting.

chamikaramj:
Ok. Lemme know when that's addressed.

aryann:
@chamikaramj thank you for your patience. I've fixed the bug that I spotted related to splits. There is a unit test that covers this with a comment explaining the reasoning behind the numbers. Please take another look.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryStorageStreamSource.java
```diff
@@ -210,20 +219,50 @@ private synchronized boolean readNextRecord() throws IOException {
...
+          fractionConsumedFromPreviousResponse
+              + (fractionConsumedFromCurrentResponse - fractionConsumedFromPreviousResponse)
```
chamikaramj: So here, you compute the fraction consumed by adding to the previous fraction consumed. I guess this is where the bug was ? (sorry, don't have a clear diff due to the fixup).

Why do we calculate fraction consuming by adding to the previous fraction consumed ? Previous response will be off in case of a split (as you probably found out). Won't it be more accurate to determine the fraction consumed by always considering current values ? ("total consumed so far" / current size). Does the API limit this calculation somehow ? (if so please add a clear comment about that).

It's hard for me to figure out whether this is correct or not without knowing what the value returned by API call in getFractionConsumed() below. So please add a comment there as well and if possible and add a link to the a BQ public API doc that gives a description about the value returned from that call.

chamikaramj:
BTW I noticed that we do not document Method DIRECT_READ as experimental at following location.
https://github.com/apache/beam/blob/master/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryIO.java#L632

I think we should clearly document it as experimental there till it's stable. Unfortunately experimental annotation will not work for this option. How about renaming it to EXPERIMENTAL_DIRECT_READ ?

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryStorageStreamSource.java
aryann: > So here, you compute the fraction consumed by adding to the previous fraction consumed. I guess this is where the bug was ? (sorry, don't have a clear diff due to the fixup).

This section has not changed. There was never a bug here.

> 
> Why do we calculate fraction consuming by adding to the previous fraction consumed ? 

We are not adding to the previous fractionConsumed. We are adding to fractionConsumedFromPreviousResponse, which is the value returned from the last response. This value means how much progress you have made assuming you have consumed all rows in that response. If you have a new response and you have consumed zero rows then your progress == fractionConsumedFromPreviousResponse. Otherwise, it's an interpolation between fractionConsumedFromPreviousResponse and fractionConsumedFromCurrentResponse.

> Previous response will be off in case of a split (as you probably found out). Won't it be more accurate to determine the fraction consumed by always considering current values ? ("total consumed so far" / current size). Does the API limit this calculation somehow ? (if so please add a clear comment about that).

The calculation will not be off in the case of a split. That was the bug that I addressed in the last commit. I probably should not have amended the commit, but added a separate commit. Apologies for the lack of proper diffs. The fix is on line 365:

fractionConsumedFromCurrentResponse = fractionConsumed;

On the next call to readNextRecord(), fractionConsumedFromCurrentResponse becomes fractionConsumedFromPreviousResponse. Using the last fractionConsumed value for fractionConsumedFromPreviousResponse is safe because the following response's consumed value will be larger (the denominator is now smaller even if the internal offset, i.e., numerator, is the same), so we can safely interpolate between those two values. We have an internal end-to-end test that tests this.

> 
> It's hard for me to figure out whether this is correct or not without knowing what the value returned by API call in getFractionConsumed() below. So please add a comment there as well and if possible and add a link to the a BQ public API doc that gives a description about the value returned from that call.

The value is in the range [0.0, 1.0]. Does that answer your question?

aryann:
> BTW I noticed that we do not document Method DIRECT_READ as experimental at following location.
> https://github.com/apache/beam/blob/master/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryIO.java#L632
> 
> I think we should clearly document it as experimental there till it's stable. Unfortunately experimental annotation will not work for this option. How about renaming it to EXPERIMENTAL_DIRECT_READ ?

Assuming no one has a dependency on this identifier, I'm okay with that, but let's not tackle this in this pull request since that's unrelated.

chamikaramj:
LGTM. Thanks.

PR #9158, opened 2019-07-25
[DO NOT MERGE] Test ZetaSQL on Beam

apilloud:
This tests that we can use the ZetaSQL parser on Beam.

This contains no actual code and should not be reviewed or merged!


amaliujia:
Run SQL PostCommit

PR #9168, opened 2019-07-26
[BEAM-7840] Provide MapTuple and FlatMapTuple for Python 3 users.

robertwb:
These are like Map and FlatMap but expand out tuple input elements across
several arguments.  This will be useful as tuple argument unpacking has been
removed in Python 3.  Instead of having to convert

    Map(lambda (k, v): expresion(k, v))

into

    Map(lambda k_v: expression(k_v[0], k_v[1]))

one can now write

    MapTuple(lambda k, v: expression(k, v))

------------------------

--- | --- | --- | --- | ---


robertwb:
R: @tvalentyn I was reminded yesterday of a complaint someone had about migrating to Python 3. Here's a potential solution, let me know what you think. (It'd be good to get this into 2.15 if you think it'd be helpful.)

sdks/python/apache_beam/transforms/core.py
```diff
@@ -1316,6 +1323,146 @@ def Map(fn, *args, **kwargs):  # pylint: disable=invalid-name
...
+  get_type_hints(wrapper).input_types = get_type_hints(fn).input_types
+  output_hint = get_type_hints(fn).simple_output_type(label)
+  if output_hint:
+    get_type_hints(wrapper).set_output_types(typehints.Iterable[output_hint])
+
+  argspec = getfullargspec(fn)
+  num_defaults = len(argspec.defaults or ())
+  if num_defaults < len(args) + len(kwargs):
+    raise TypeError('Side inputs must have defaults for MapTuple.')
+  modified_argspec = inspect.ArgSpec(
```
tvalentyn: The code here is not Py3 compatible (see test failures). Could we reuse helpers in decorators.py or introduce a new helper there instead of calling inspect.* methods directly?
Also, we should keep in mind that we eventually want to support keyword-only arguments.

sdks/python/apache_beam/utils/timestamp.py
```diff
@@ -151,7 +151,10 @@ def __int__(self):
   def __eq__(self, other):
     # Allow comparisons between Duration and Timestamp values.
     if not isinstance(other, Duration):
-      other = Timestamp.of(other)
+      try:
+        other = Timestamp.of(other)
+      except TypeError:
+        return NotImplemented
```
tvalentyn: Is  `assert_that(pcoll | 'StaticSides' >> beam.core.MapTuple(fn, 's1', 's2'),` from the test case evaluated differently without returning `NotImplemented` here?

sdks/python/apache_beam/examples/snippets/snippets.py
```diff
@@ -446,7 +446,7 @@ def examples_wordcount_minimal(renames):
       # [END examples_wordcount_minimal_count]
 
       # [START examples_wordcount_minimal_map]
-      | beam.Map(lambda word_count: '%s: %s' % (word_count[0], word_count[1]))
```
tvalentyn: I saw such pattern in quite a few places so this may be a helpful syntactic sugar, as long as it does not make things difficult for us. 
I wouldn't rush this PR to 2.15.0 if we need more time to polish things.

tvalentyn:
nit: s/FlatMap/MapTuple in PR title and add a JIRA.

sdks/python/apache_beam/transforms/core.py
robertwb: I had played around with Python 3, but apparently that didn't make it into my final solution... Fixed.

sdks/python/apache_beam/utils/timestamp.py
robertwb: It's something I noticed during testing--it doesn't affect the passing test, but for comparing to a different type (e.g. None or a string) it does matter.

sdks/python/apache_beam/examples/snippets/snippets.py
robertwb: I don't think it'll make it harder for us, it's similar to the stuff we're already doing to support Map, Filter, etc. As for 2.15, I'm hoping we can start pushing Python 3 even more strongly then, which is why it'd be nice to get this syntactic sugar in.

sdks/python/apache_beam/transforms/core.py
```diff
@@ -1316,6 +1322,142 @@ def Map(fn, *args, **kwargs):  # pylint: disable=invalid-name
...
+  """
+  if not callable(fn):
+    raise TypeError(
+        'MapTuple can be used only with callable objects. '
+        'Received %r instead.' % (fn))
+
+  if _fn_takes_side_inputs(fn):
+    wrapper = lambda x, *args, **kwargs: [fn(*(tuple(x) + args), **kwargs)]
+  else:
+    wrapper = lambda x: [fn(*x)]
```
tvalentyn: If we keep `else` branch, we probably want to cast `x` to tuple as well.

sdks/python/apache_beam/examples/snippets/snippets.py
tvalentyn: Ok, we should tag associated Jira with 'fix_version==2.15.0' in this case.

sdks/python/apache_beam/pipeline_test.py
```diff
@@ -168,6 +168,35 @@ def test_flatmap_builtin(self):
...
+                label='StaticSidesCheck')
+    assert_that(pcoll | 'DynamicSides' >> beam.core.MapTuple(fn, side1, side2),
+                equal_to([('e1', 'e2', MIN_TIMESTAMP, 's1', 's2')]),
+                label='DynamicSidesCheck')
+    assert_that(pcoll | 'MixedSides' >> beam.core.MapTuple(fn, s2=side2),
+                equal_to([('e1', 'e2', MIN_TIMESTAMP, None, 's2')]),
+                label='MixedSidesCheck')
+
+    # FlatMapTuple is similar.
+    assert_that(pcoll | 'FlatMap' >> beam.core.FlatMapTuple(fn, s2=side2),
```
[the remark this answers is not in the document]
robertwb: Done.

sdks/python/apache_beam/transforms/core.py
robertwb: Set.

sdks/python/apache_beam/transforms/core.py
```diff
@@ -1316,6 +1322,142 @@ def Map(fn, *args, **kwargs):  # pylint: disable=invalid-name
...
+    ~exceptions.TypeError: If the **fn** passed as argument is not a callable.
+      Typical error is to pass a :class:`DoFn` instance which is supported only
+      for :class:`ParDo`.
+  """
+  if not callable(fn):
+    raise TypeError(
+        'MapTuple can be used only with callable objects. '
+        'Received %r instead.' % (fn))
+
+  if _fn_takes_side_inputs(fn):
```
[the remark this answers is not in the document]
robertwb: Nice catch. 

Calling a function without varargs and kwargs is over twice as fast (e.g. no need to create empty dictionaries), which is why these branches were introduced. I've fixed the logic.

sdks/python/apache_beam/transforms/core.py
robertwb: The Python interpreter will do that for us. As this is the fast path, probably better avoid the extra cast.

sdks/python/apache_beam/transforms/core.py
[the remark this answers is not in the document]
tvalentyn: Thanks, good to know.

sdks/python/apache_beam/transforms/core.py
```diff
@@ -1431,24 +1432,25 @@ def FlatMapTuple(fn, *args, **kwargs):  # pylint: disable=invalid-name
         'MapTuple can be used only with callable objects. '
```
tvalentyn: s/MapTuple/FlatMapTuple in line 1432

sdks/python/apache_beam/transforms/core.py
```diff
@@ -1316,6 +1322,144 @@ def Map(fn, *args, **kwargs):  # pylint: disable=invalid-name
...
+
+  label = 'MapTuple(%s)' % ptransform.label_from_callable(fn)
+
+  argspec = getfullargspec(fn)
+  num_defaults = len(argspec.defaults or ())
+  if num_defaults < len(args) + len(kwargs):
+    raise TypeError('Side inputs must have defaults for MapTuple.')
+
+  if argspec.defaults or args or kwargs:
+    wrapper = lambda x, *args, **kwargs: [fn(*(tuple(x) + args), **kwargs)]
```
udim: I was looking at this code and wondered why is there a `tuple(x)` not `x`, since x should always be tuple according to the docstring? The tests pass with a plain `x`.

sdks/python/apache_beam/transforms/core.py
```diff
@@ -1316,6 +1322,144 @@ def Map(fn, *args, **kwargs):  # pylint: disable=invalid-name
...
+    raise TypeError(
+        'MapTuple can be used only with callable objects. '
+        'Received %r instead.' % (fn))
+
+  label = 'MapTuple(%s)' % ptransform.label_from_callable(fn)
+
+  argspec = getfullargspec(fn)
+  num_defaults = len(argspec.defaults or ())
+  if num_defaults < len(args) + len(kwargs):
+    raise TypeError('Side inputs must have defaults for MapTuple.')
```
shoyer, line 1371: @robertwb  I'm quite curious about the reason for this restriction, which doesn't exist on `beam.Map`.

Is this just a detail of the current implementation that could potentially removed, or is there a more fundamental reason for it?

Also, would keyword only arguments be OK on Python 3? Based on the implementation, I would assume the answer is

PR #9190, opened 2019-07-30
[BEAM-7520] Fix timer firing order in DirectRunner

je-ik:
Fixes [BEAM-7520].

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/runners/direct-java/src/main/java/org/apache/beam/runners/direct/EvaluationContext.java
+++ b/runners/direct-java/src/main/java/org/apache/beam/runners/direct/EvaluationContext.java
@@ -140,6 +140,7 @@ class EvaluationContext {
    *     null} if the transform that produced the result is a root transform
    * @param completedTimers the timers that were delivered to produce the {@code completedBundle},
    *     or an empty iterable if no timers were delivered
+   * @param pushedBackTimers timers that have been pushed back during processing
    * @param result the result of evaluating the input bundle
    * @return the committed bundles contained within the handled {@code result}
    */
@@ -226,7 +227,11 @@ class EvaluationContext {
   private void fireAvailableCallbacks(AppliedPTransform<?, ?, ?> producingTransform) {
     TransformWatermarks watermarks = watermarkManager.getWatermarks(producingTransform);
     Instant outputWatermark = watermarks.getOutputWatermark();
-    callbackExecutor.fireForWatermark(producingTransform, outputWatermark);
+    try {
+      callbackExecutor.fireForWatermark(producingTransform, outputWatermark);
+    } catch (InterruptedException ex) {
+      Thread.currentThread().interrupt();
+    }
   }
 
   /** Create a {@link UncommittedBundle} for use by a source. */
@@ -369,7 +374,7 @@ class EvaluationContext {
    * <p>This is a destructive operation. Timers will only appear in the result of this method once
    * for each time they are set.
    */
-  public Collection<FiredTimers<AppliedPTransform<?, ?, ?>>> extractFiredTimers() {
+  Collection<FiredTimers<AppliedPTransform<?, ?, ?>>> extractFiredTimers() {
     forceRefresh();
     return watermarkManager.extractFiredTimers();
   }
--- a/runners/direct-java/src/main/java/org/apache/beam/runners/direct/QuiescenceDriver.java
+++ b/runners/direct-java/src/main/java/org/apache/beam/runners/direct/QuiescenceDriver.java
@@ -249,6 +249,7 @@ class QuiescenceDriver implements ExecutionDriver {
    * Exception)}.
    */
   private class TimerIterableCompletionCallback implements CompletionCallback {
+
     private final Iterable<TimerData> timers;
 
     TimerIterableCompletionCallback(Iterable<TimerData> timers) {
@@ -258,8 +259,9 @@ class QuiescenceDriver implements ExecutionDriver {
     @Override
     public final CommittedResult handleResult(
         CommittedBundle<?> inputBundle, TransformResult<?> result) {
-      CommittedResult<AppliedPTransform<?, ?, ?>> committedResult =
-          evaluationContext.handleResult(inputBundle, timers, result);
+
+      final CommittedResult<AppliedPTransform<?, ?, ?>> committedResult;
+      committedResult = evaluationContext.handleResult(inputBundle, timers, result);
       for (CommittedBundle<?> outputBundle : committedResult.getOutputs()) {
         pendingWork.offer(
             WorkUpdate.fromBundle(
--- a/runners/direct-java/src/main/java/org/apache/beam/runners/direct/StatefulParDoEvaluatorFactory.java
+++ b/runners/direct-java/src/main/java/org/apache/beam/runners/direct/StatefulParDoEvaluatorFactory.java
@@ -20,10 +20,14 @@ package org.apache.beam.runners.direct;
 import static org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Preconditions.checkState;
 
 import com.google.auto.value.AutoValue;
+import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.PriorityQueue;
 import org.apache.beam.runners.core.KeyedWorkItem;
 import org.apache.beam.runners.core.KeyedWorkItems;
 import org.apache.beam.runners.core.StateNamespace;
@@ -34,6 +38,7 @@ import org.apache.beam.runners.core.StateTags;
 import org.apache.beam.runners.core.TimerInternals.TimerData;
 import org.apache.beam.runners.direct.DirectExecutionContext.DirectStepContext;
 import org.apache.beam.runners.direct.ParDoMultiOverrideFactory.StatefulParDo;
+import org.apache.beam.runners.direct.WatermarkManager.TimerUpdate;
 import org.apache.beam.runners.local.StructuralKey;
 import org.apache.beam.sdk.coders.Coder;
 import org.apache.beam.sdk.options.PipelineOptions;
@@ -56,6 +61,7 @@ import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.cache.CacheBuild
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.cache.CacheLoader;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.cache.LoadingCache;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.Lists;
+import org.joda.time.Instant;
 
 /** A {@link TransformEvaluatorFactory} for stateful {@link ParDo}. */
 final class StatefulParDoEvaluatorFactory<K, InputT, OutputT> implements TransformEvaluatorFactory {
@@ -232,10 +238,13 @@ final class StatefulParDoEvaluatorFactory<K, InputT, OutputT> implements Transfo
       implements TransformEvaluator<KeyedWorkItem<K, KV<K, InputT>>> {
 
     private final DoFnLifecycleManagerRemovingTransformEvaluator<KV<K, InputT>> delegateEvaluator;
+    private final List<TimerData> pushedBackTimers = new ArrayList<>();
+    private final DirectTimerInternals timerInternals;
 
     public StatefulParDoEvaluator(
         DoFnLifecycleManagerRemovingTransformEvaluator<KV<K, InputT>> delegateEvaluator) {
       this.delegateEvaluator = delegateEvaluator;
+      this.timerInternals = delegateEvaluator.getParDoEvaluator().getStepContext().timerInternals();
     }
 
     @Override
@@ -245,7 +254,12 @@ final class StatefulParDoEvaluatorFactory<K, InputT, OutputT> implements Transfo
         delegateEvaluator.processElement(windowedValue);
       }
 
-      for (TimerData timer : gbkResult.getValue().timersIterable()) {
+      Instant currentInputWatermark = timerInternals.currentInputWatermarkTime();
+      PriorityQueue<TimerData> toBeFiredTimers =
+          new PriorityQueue<>(Comparator.comparing(TimerData::getTimestamp));
+      gbkResult.getValue().timersIterable().forEach(toBeFiredTimers::add);
+      while (!toBeFiredTimers.isEmpty()) {
+        TimerData timer = toBeFiredTimers.poll();
         checkState(
             timer.getNamespace() instanceof WindowNamespace,
             "Expected Timer %s to be in a %s, but got %s",
@@ -255,17 +269,23 @@ final class StatefulParDoEvaluatorFactory<K, InputT, OutputT> implements Transfo
         WindowNamespace<?> windowNamespace = (WindowNamespace) timer.getNamespace();
         BoundedWindow timerWindow = windowNamespace.getWindow();
         delegateEvaluator.onTimer(timer, timerWindow);
+        if (timerInternals.containsUpdateForTimeBefore(currentInputWatermark)) {
+          break;
+        }
       }
+      pushedBackTimers.addAll(toBeFiredTimers);
     }
 
     @Override
     public TransformResult<KeyedWorkItem<K, KV<K, InputT>>> finishBundle() throws Exception {
       TransformResult<KV<K, InputT>> delegateResult = delegateEvaluator.finishBundle();
-
+      TimerUpdate timerUpdate =
+          delegateResult.getTimerUpdate().withPushedBackTimers(pushedBackTimers);
+      pushedBackTimers.clear();
       StepTransformResult.Builder<KeyedWorkItem<K, KV<K, InputT>>> regroupedResult =
           StepTransformResult.<KeyedWorkItem<K, KV<K, InputT>>>withHold(
                   delegateResult.getTransform(), delegateResult.getWatermarkHold())
-              .withTimerUpdate(delegateResult.getTimerUpdate())
+              .withTimerUpdate(timerUpdate)
               .withState(delegateResult.getState())
               .withMetricUpdates(delegateResult.getLogicalMetricUpdates())
               .addOutput(Lists.newArrayList(delegateResult.getOutputBundles()));
--- a/runners/direct-java/src/main/java/org/apache/beam/runners/direct/WatermarkCallbackExecutor.java
+++ b/runners/direct-java/src/main/java/org/apache/beam/runners/direct/WatermarkCallbackExecutor.java
@@ -19,9 +19,12 @@ package org.apache.beam.runners.direct;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.PriorityQueue;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executor;
 import javax.annotation.Nonnull;
 import org.apache.beam.sdk.runners.AppliedPTransform;
@@ -116,14 +119,30 @@ class WatermarkCallbackExecutor {
    * Schedule all pending callbacks that must have produced output by the time of the provided
    * watermark.
    */
-  public void fireForWatermark(AppliedPTransform<?, ?, ?> step, Instant watermark) {
+  public void fireForWatermark(AppliedPTransform<?, ?, ?> step, Instant watermark)
+      throws InterruptedException {
     PriorityQueue<WatermarkCallback> callbackQueue = callbacks.get(step);
     if (callbackQueue == null) {
       return;
     }
     synchronized (callbackQueue) {
+      List<Runnable> toFire = new ArrayList<>();
       while (!callbackQueue.isEmpty() && callbackQueue.peek().shouldFire(watermark)) {
-        executor.execute(callbackQueue.poll().getCallback());
+        toFire.add(callbackQueue.poll().getCallback());
+      }
+      if (!toFire.isEmpty()) {
+        CountDownLatch latch = new CountDownLatch(toFire.size());
+        toFire.forEach(
+            r ->
+                executor.execute(
+                    () -> {
+                      try {
+                        r.run();
+                      } finally {
+                        latch.countDown();
+                      }
+                    }));
+        latch.await();
       }
     }
   }
--- a/runners/direct-java/src/main/java/org/apache/beam/runners/direct/WatermarkManager.java
+++ b/runners/direct-java/src/main/java/org/apache/beam/runners/direct/WatermarkManager.java
@@ -36,11 +36,17 @@ import java.util.NavigableSet;
 import java.util.Objects;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.function.Consumer;
 import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.GuardedBy;
@@ -63,7 +69,9 @@ import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.Comparis
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.HashBasedTable;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.ImmutableList;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.Iterables;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.Lists;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.Ordering;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.Sets;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.SortedMultiset;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.Table;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.TreeMultiset;
@@ -237,13 +245,18 @@ public class WatermarkManager<ExecutableT, CollectionT> {
     // This per-key sorted set allows quick retrieval of timers that should fire for a key
     private final Map<StructuralKey<?>, NavigableSet<TimerData>> objectTimers;
 
-    private AtomicReference<Instant> currentWatermark;
+    private final AtomicReference<Instant> currentWatermark;
+
+    private final Consumer<TimerData> timerUpdateNotification;
 
     public AppliedPTransformInputWatermark(
-        String name, Collection<? extends Watermark> inputWatermarks) {
-      this.name = name;
+        String name,
+        Collection<? extends Watermark> inputWatermarks,
+        Consumer<TimerData> timerUpdateNotification) {
 
+      this.name = name;
       this.inputWatermarks = inputWatermarks;
+
       // The ordering must order elements by timestamp, and must not compare two distinct elements
       // as equal. This is built on the assumption that any element added as a pending element will
       // be consumed without modifications.
@@ -255,7 +268,8 @@ public class WatermarkManager<ExecutableT, CollectionT> {
       this.pendingTimers = TreeMultiset.create();
       this.objectTimers = new HashMap<>();
       this.existingTimers = new HashMap<>();
-      currentWatermark = new AtomicReference<>(BoundedWindow.TIMESTAMP_MIN_VALUE);
+      this.currentWatermark = new AtomicReference<>(BoundedWindow.TIMESTAMP_MIN_VALUE);
+      this.timerUpdateNotification = timerUpdateNotification;
     }
 
     @Override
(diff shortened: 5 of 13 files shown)


je-ik:
Run Direct ValidatesRunner

je-ik:
Run Dataflow ValidatesRunner

je-ik:
Run Flink ValidatesRunner

je-ik:
Run Spark ValidatesRunner

je-ik:
Run SQL PostCommit

je-ik:
R: @kennknowles @iemejia @robertwb This PR is related to previous [BEAM-7785], but solves the original issue [BEAM-7520] more in a more complex way. I'd be happy if you had time to lokk into this. Many thanks!

je-ik:
Run SQL PostCommit

je-ik:
Run Direct ValidatesRunner

je-ik:
Run Dataflow ValidatesRunner

je-ik:
Run Spark ValidatesRunner

je-ik:
Run Flink ValidatesRunner

tvalentyn:
Just a heads-up @kennknowles is out of office for 1 week.

je-ik:
@tvalentyn I know, thanks. This can wait.

je-ik:
@kennknowles would you have time to look into this, please?

reuvenlax:
R: @dpmills

runners/direct-java/src/main/java/org/apache/beam/runners/direct/WatermarkCallbackExecutor.java
```diff
@@ -116,14 +119,30 @@ public void callOnWindowExpiration(
...
       return;
     }
     synchronized (callbackQueue) {
+      List<Runnable> toFire = new ArrayList<>();
       while (!callbackQueue.isEmpty() && callbackQueue.peek().shouldFire(watermark)) {
-        executor.execute(callbackQueue.poll().getCallback());
+        toFire.add(callbackQueue.poll().getCallback());
+      }
+      if (!toFire.isEmpty()) {
+        CountDownLatch latch = new CountDownLatch(toFire.size());
```
kennknowles: Instead of managing the latch yourself, you could gather a list of futures and wait on it. It would be a bit higher level programming style, and not depend on capturing the latch in the closure, etc.

runners/direct-java/src/main/java/org/apache/beam/runners/direct/StatefulParDoEvaluatorFactory.java
```diff
@@ -244,27 +249,40 @@ public void processElement(WindowedValue<KeyedWorkItem<K, KV<K, InputT>>> gbkRes
         delegateEvaluator.processElement(windowedValue);
       }
 
+      Instant lastFired = null;
```
kennknowles: `@Nullable`

runners/direct-java/src/main/java/org/apache/beam/runners/direct/StatefulParDoEvaluatorFactory.java
```diff
@@ -244,27 +249,40 @@ public void processElement(WindowedValue<KeyedWorkItem<K, KV<K, InputT>>> gbkRes
...
+            "lastFired was %s, current %s",
+            lastFired,
+            timer.getTimestamp());
+        if (lastFired != null && lastFired.isBefore(timer.getTimestamp())) {
+          pushedBackTimers.add(timer);
+        } else {
+          lastFired = timer.getTimestamp();
+          WindowNamespace<?> windowNamespace = (WindowNamespace) timer.getNamespace();
+          BoundedWindow timerWindow = windowNamespace.getWindow();
+          delegateEvaluator.onTimer(timer, timerWindow);
```
kennknowles: I'm reading this really quickly, but I think it works only for loops of the same timer, not the general case. And it also leaves the "should fire" implicit only by the fact that it already fired, instead of checking it. I think if you have a unit test where two timer callbacks set each other to "already firing" times you will uncover some issues.

I think there's a simple logic you might have:

1. Add the fired timers in the incoming bundle to a priority queue
2. While any timers to fire, call the first callback
    a. For any timer set by the `@OnTimer` callback, you check if it should fire. If yes, then add to fired timers priority queue. If no, then add it to the committed result.

runners/direct-java/src/main/java/org/apache/beam/runners/direct/WatermarkCallbackExecutor.java
je-ik: If I understand it correctly, then this in unfortunately currently not possible, because the `executor` is not `ExecutorService`, but simply `Executor`, which returns `void`, instead of `Future`. I cannot simply change it to `ExecutorService`, because `EvaluationContext` passes there `MoreExecutors.directExecutor()`, which is not `ExecutorService`.

runners/direct-java/src/main/java/org/apache/beam/runners/direct/StatefulParDoEvaluatorFactory.java
je-ik: :+1:

runners/direct-java/src/main/java/org/apache/beam/runners/direct/StatefulParDoEvaluatorFactory.java
je-ik: I will add the test and check if there are any issues. Regarding the logic you described - generally, yes, that would be the ideal approach. It would also solve some other issues, that still remain, even after this PR - but - it would require a larger refactoring, because there is currently no simple (or simple and apparent to me) way of knowing if `@OnTimer` method actually setup any new timer. The `delegateEvaluator` would have to signal that, and that boils down to `PushbackSideInputDoFnRunner.onTimer` would have to return the new timers. That is API change in core and I wanted to avoid that. But maybe I've overlooked  some other way.

runners/direct-java/src/main/java/org/apache/beam/runners/direct/WatermarkCallbackExecutor.java
je-ik: Hm, my previous comment might suggest, that the change is not needed at all. And maybe it is true. :-) I added this code as part of my "I'm really desperate" period while solving the issue. Just for being sure, that this doesn't break in the future, I'd like to keep it as is, if it is okay.

runners/direct-java/src/main/java/org/apache/beam/runners/direct/StatefulParDoEvaluatorFactory.java
je-ik: I have a failing test case for the case you described. Will work on that. Many thanks!

je-ik: @kennknowles Fixed the test. I'm not sure if it is exactly what you meant (although the new test really uncovered some flaws in the original implementation). Can you have a look [here](https://github.com/apache/beam/pull/9190/files#diff-37ebd0b68ae7d5b951cfca73aba52cc1R689), please?

je-ik:
Run Direct ValidatesRunner

je-ik:
Run Flink ValidatesRunner

je-ik:
Run Spark ValidatesRunner

je-ik:
Run Dataflow ValidatesRunner

je-ik:
Run Direct ValidatesRunner

je-ik:
Run Flink ValidatesRunner

je-ik:
Run Dataflow ValidatesRunner

je-ik:
@kennknowles I added @ValidatesRunner for timer ordering when using `Create` transform instead of `TestStream` (there was a suspicion that DirectRunner handles it somehow wrong, but it seems to pass), but that apparently broke ValidatesRunner of Dataflow. I'm not sure how to proceed, can you give me a hand?

je-ik:
I need to push this forward, because it is prerequisite for more PRs that I have (mostly) ready. But these need to be reliably testable, so sorry if I'm a little impatient. :-)

@kennknowles, if you don't have time for this, would you please delegate this?
Open questions:
 * is the current approach OK? It is tested in terms of not causing deadlocks, performance should be somewhat comparable to current master - I'm aware that the completely correct solution would be to evaluate and add timers to *current* bundle, not rescheduling it for later, but that seems to be major refactor (please, anyone confirm or disprove this statement as it might be wrong)
 * there was observed strange behavior related to `Create` vs. `TestStream` in DirectRunner, but it is not reproducible in this PR. It was observed and described in https://github.com/rezarokni/beam/pull/2#issuecomment-522226493
 * added test for `Create` instead of `TestStream` broke only ValidatesRunner of Dataflow (all other runners seem to pass)

kennknowles:
@reuvenlax who is listed here, is also a good reviewer

Sorry! I missed this because I had a mention but failed to add it to my pending review queue. I'll get to it tomorrow.

je-ik:
Many thanks! :+1:

runners/direct-java/src/main/java/org/apache/beam/runners/direct/CommittedResult.java
```diff
@@ -18,10 +18,10 @@
 package org.apache.beam.runners.direct;
 
 import com.google.auto.value.AutoValue;
+import java.util.Optional;
```
kennknowles: One reason not to use this is that it is not `Serializable` and the Java committee has decided it never should be. I obviously disagree, but until then I don't think `java.util.Optional` should be used.

runners/direct-java/src/main/java/org/apache/beam/runners/direct/CopyOnAccessInMemoryStateInternals.java
```diff
@@ -22,6 +22,7 @@
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.Map;
+import java.util.Optional;
```
kennknowles: ditto

runners/direct-java/src/main/java/org/apache/beam/runners/direct/EvaluationContext.java
```diff
@@ -23,6 +23,7 @@
 import java.util.EnumSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
```
kennknowles: ditto

runners/direct-java/src/main/java/org/apache/beam/runners/direct/ExecutorServiceParallelExecutor.java
```diff
@@ -20,6 +20,7 @@
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Map;
+import java.util.Optional;
```
kennknowles: ditto

runners/direct-java/src/main/java/org/apache/beam/runners/direct/QuiescenceDriver.java
```diff
@@ -22,6 +22,7 @@
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Map;
+import java.util.Optional;
```
kennknowles: ditto

runners/direct-java/src/main/java/org/apache/beam/runners/direct/SideInputContainer.java
```diff
@@ -24,6 +24,7 @@
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
```
kennknowles: ditto

runners/direct-java/src/main/java/org/apache/beam/runners/direct/StatefulParDoEvaluatorFactory.java
```diff
@@ -244,27 +250,40 @@ public void processElement(WindowedValue<KeyedWorkItem<K, KV<K, InputT>>> gbkRes
...
-        WindowNamespace<?> windowNamespace = (WindowNamespace) timer.getNamespace();
-        BoundedWindow timerWindow = windowNamespace.getWindow();
-        delegateEvaluator.onTimer(timer, timerWindow);
+        checkState(
+            lastFired == null || !lastFired.isAfter(timer.getTimestamp()),
+            "lastFired was %s, current %s",
+            lastFired,
+            timer.getTimestamp());
+        if (lastFired != null && lastFired.isBefore(timer.getTimestamp())) {
+          pushedBackTimers.add(timer);
```
kennknowles: Now that I look more closely, doesn't this cause the timers to fire one at a time if they are in order, with all the rest pushed back into the commit result? And only when a timer is moved earlier you go to the other branch and fire multiple timers in the same bundle.

runners/direct-java/src/main/java/org/apache/beam/runners/direct/StatefulParDoEvaluatorFactory.java
kennknowles: I don't think it is quite what I might. Here:

t1.set(50)
t2.set(60)

watermark advances to 100

t1 fires and sets t2 to 40 -- since it is moved earlier it will fire in the same bundle
t2 fires and sets t1 to 30 -- will not fire in the same bundle, but will be committed in the result?

I guess this might get the right result... I was just expecting all the timers that could fire to do it in the same bundle. It is a bit weird that the normal case of timers just coming in order and not setting anything will not fire in one bundle. Did I read that code wrong? This is why I suggested a prio queue, to remove all the logic about `lastFired` and commits. You just fire all the timers, in order (including newly set timers), until the bundle is done. I wonder if there is a use case for always setting a timer for the same time and hoping the bundle will still commit over and over again.

runners/direct-java/src/main/java/org/apache/beam/runners/direct/CommittedResult.java
je-ik: There was a very healthy discussion about usage of `Optional` on flink mailing list: https://lists.apache.org/thread.html/f5f8ce92f94c9be6774340fbd7ae5e4afe07386b6765ad3cfb13aec0@%3Cdev.flink.apache.org%3E
General consensus is that `Optional` should not be used for fields, but only for return values of performance non-critical code. I very much agree with that, if `Optional` is not used as field, its serializability doesn't matter. Moreover, DirectRunner does not rely that heavily on serialization (more than validating, that the user code is serializable). Using standard instead of guava seems to be beneficial to me, without any negative impact.

runners/direct-java/src/main/java/org/apache/beam/runners/direct/StatefulParDoEvaluatorFactory.java
je-ik: Generally yes. The problem is that you cannot fire multiple timers at once, because the outcome of firing a single timer is unknown. It might or might not setup some other timer. If it was to be solved, it seems to me it would require `PushbackSideInputDoFnRunner` (and most likely `DoFnRunner` to return set timers, instead of `void`. The other option would be somehow hack into `WatermarkManager` and be able to retrieve "most recent timers set", os something like that (which seems to be fragile and hackish). The correct solution seems to be a too big refactor, given that the performance penalty for the solution proposed in this PR seems to be low.

runners/direct-java/src/main/java/org/apache/beam/runners/direct/StatefulParDoEvaluatorFactory.java
je-ik: Is it possible (and valid) to set timer for time that is *before* the time another timer was set for? That seems weird and actually can cause exception to be thrown in some cases.

runners/direct-java/src/main/java/org/apache/beam/runners/direct/CommittedResult.java
kennknowles: Yes, that is exactly the logic I am familiar with. We've also had healthy discussions in Beam and seemingly every project. I can't escape null! I share the "null was a mistake, never really made sense in the first place, and should be eliminated from languages" perspective. I believe basic good programming practice to program as though it does not actually exist, except when necessary for compatibility or performance. Using `@Nullable` and `@NonNull` with an extended type checker is a compromise when necessary.

I have not seen any actual evidence for the claim that it causes performance problems. I would easily believe it - It is true that Java is exceptionally inefficient w.r.t. trivial wrappers unless they are well-optimized. But on the thread you linked the only person who brought up real performance said that they could not detect any difference except when _serializing_ it in performance-sensitive paths. So perhaps Java isn't so inefficient at optimizing it away after all.

In any event, please make it a separate change. You can do a bulk change in the codebase if you like. At this point, I am just so tired of this issue which humans are still stuck on that I will let most anything through :-/

je-ik: Yes, I also think that the performance impact of creating one short lived instance should be negligible. I actually tried to measure it, but I didn't get any statistically significant difference in this case. Although, I have seen significant impact in other cases (getting rid of creating useless instance in one place of spark runner gave me about 20% performance gain, but that could have been very specific case). But that is probably not that significant to this PR. :-)

To conclude this, I will drop the commit where I changed the Optional in DirectRunner and start a discuss thread related to making the usage consistent across the code base. Currently there is approx 3:2 usage of java.util vs. guava.

je-ik:
@kennknowles I removed the Optional stuff from this PR, so that it can be focused. I have refactored the code in `StatefulParDoEvaluator` a bit, it now enables firing multiple timers in one bundle (as long as firing any timer does not set or delete any timer that is in the current bundle). If a timer sets timer for time belonging in the current bundle, then all other (not yet fired) timers are pushed back to later bundle. Reason for this is that the logic of handling TimerUpdates lies on `WatermarkManager` and should not be performed by `StatefulParDoEvaluator`. Thanks for the discussion, I didn't dig deeper into the code to see, that I can get quite cleanly instance of `DirectTimerInternals` to get the information that was needed. I think, that now the impact on performance will be truly negligible, because overwhelming majority of cases will fire all timers in bundle that is being processed and only when needed, timers will be pushed back.

je-ik:
Run Java PreCommit

je-ik:
Run Dataflow ValidatesRunner

je-ik:
@kennknowles @reuvenlax The added @ValidatesRunner fails on Dataflow. It looks to me, that it should pass, but cannot tell for sure. Can you check this?

je-ik:
For clarity, it is [this one](https://github.com/apache/beam/pull/9190/files#diff-68ed88cc25171e30985dda4534e68c9bR3184)

je-ik:
@robertwb Would you have time to look into this please?

je-ik:
Run Dataflow ValidatesRunner

je-ik:
The problem with Dataflow validatesRunner was apparently only in timeout (strange though, that the same timeout consistently passed when using `TestStream`, but consistently failed with `Create`). Anyway, I removed the timeout and the test passes.

iemejia:
This PR seems to be almost there. Anything important still missing @kennknowles?

kennknowles:
I haven't had a chance to look again - thanks for the ping.

runners/direct-java/src/main/java/org/apache/beam/runners/direct/StatefulParDoEvaluatorFactory.java
```diff
@@ -254,17 +268,23 @@ public void processElement(WindowedValue<KeyedWorkItem<K, KV<K, InputT>>> gbkRes
         WindowNamespace<?> windowNamespace = (WindowNamespace) timer.getNamespace();
         BoundedWindow timerWindow = windowNamespace.getWindow();
         delegateEvaluator.onTimer(timer, timerWindow);
+        if (timerInternals.containsUpdateForTimeBefore(currentInputWatermark)) {
```
kennknowles: So if I understand this correctly, the `PriorityQueue` causes the timers in the incoming bundle to be fired in timestamp order. That's good. And then if any timer is set that would already be eligible for firing, it ends processing and pushes the rest back. I think this is not optimal, because you could add timer updates to the priority queue and keep firing any eligible timer. The bundle only needs to end when there are no more eligible timers. But on the other hand, I do think it fixes the problem, so happy to merge it.

runners/direct-java/src/main/java/org/apache/beam/runners/direct/WatermarkCallbackExecutor.java
kennknowles: Fair enough :-). Since it passes and solves a problem, let's definitely merge.

runners/direct-java/src/main/java/org/apache/beam/runners/direct/WatermarkManager.java
```diff
@@ -829,6 +864,14 @@ public Instant get() {
   @GuardedBy("refreshLock")
   private final Set<ExecutableT> pendingRefreshes;
 
+  /**
+   * A set of executables with currently extracted timers, that are to be processed. Note that, due
```
kennknowles: Consistency should only be per (key, window) pair.

runners/direct-java/src/main/java/org/apache/beam/runners/direct/WatermarkManager.java
```diff
@@ -1537,6 +1680,21 @@ public boolean equals(Object other) {
           && Objects.equals(this.setTimers, that.setTimers)
           && Objects.equals(this.deletedTimers, that.deletedTimers);
     }
+
+    @Override
+    public String toString() {
+      return "TimerUpdate(key="
```
kennknowles: `MoreObject.toStringHelper` is a great utility for this sort of thing.

runners/direct-java/src/test/java/org/apache/beam/runners/direct/DirectRunnerTest.java
```diff
@@ -616,6 +649,186 @@ public void processElement(ProcessContext c) {
     p.run();
   }
 
+  /**
+   * Test running of {@link Pipeline} which has two {@link POutput POutputs} and finishing the first
+   * one triggers data being fed into the second one.
+   */
+  @Test(timeout = 10000)
+  public void testTwoPOutputsInPipelineWithCascade() throws InterruptedException {
```
kennknowles: This seems like a `@Category(ValidatesRunner.class)` test for use in either `CombineTest` or `GroupByKeyTest`.

runners/direct-java/src/test/java/org/apache/beam/runners/direct/DirectRunnerTest.java
```diff
@@ -616,6 +649,186 @@ public void processElement(ProcessContext c) {
...
+
+    run.waitUntilFinish();
+  }
+
+  @Test(timeout = 5000)
+  public void testTwoTimersSettingEachOther() {
+    Pipeline pipeline = getPipeline();
+    Instant now = new Instant(1500000000000L);
+    Instant end = now.plus(100);
+    PCollection<String> result = pipeline.apply(new TwoTimerTest(now, end));
```
kennknowles: Reading this casually, `TwoTimerTest` is not really clear what is being tested. The code that creates the output and the code that asserts about the expectations are very far from each other. Ideally, you would move the assertion into `TwoTimerTest` somehow. Since `TwoTimerTest` owns the logic, it also owns the expected output.

runners/direct-java/src/test/java/org/apache/beam/runners/direct/DirectRunnerTest.java
```diff
@@ -616,6 +649,186 @@ public void processElement(ProcessContext c) {
...
+    while (start.take() == null) {}
+
+    // and publish messages
+    messages.add(1).add(2).add(3).terminate();
+
+    run.waitUntilFinish();
+  }
+
+  @Test(timeout = 5000)
+  public void testTwoTimersSettingEachOther() {
```
kennknowles: I believe this is a `@Category(ValidatesRunner.class)` test for `ParDoTest`. And same for all state & timers feature tests.

runners/direct-java/src/test/java/org/apache/beam/runners/direct/MultiStepCombineTest.java
```diff
@@ -64,7 +64,7 @@
   private transient KvCoder<String, Long> combinedCoder =
       KvCoder.of(StringUtf8Coder.of(), VarLongCoder.of());
 
-  @Test
+  @Test(timeout = 10000)
```
kennknowles: This is unrelated. It may shrink the code review latency to try to separate little changes and just get them merged.

je-ik: Fixed.

runners/direct-java/src/main/java/org/apache/beam/runners/direct/StatefulParDoEvaluatorFactory.java
je-ik: The problem here is that I don't want to split the logic. `WatermarkManager` is responsible for handling all timer setting/firing/removing, so I didn't want to intrude into the logic from `StatefulParDoEvaluator`. If the current logic is fine, I'd merge it as it is.

runners/direct-java/src/test/java/org/apache/beam/runners/direct/DirectRunnerTest.java
je-ik: I'm afraid that could cause trouble. This test uses a (little hackish) class `StaticQueue`, that will not work well outside single JVM. That probably will not be problem for many runners, that use a local version of distributed runner for tests (spark, flink), but might definitely cause troubles for fully distributed runners (dataflow). Because the tooling in the test works only in local mode, I'd prefer to keep it for DirectRunner only.

runners/direct-java/src/test/java/org/apache/beam/runners/direct/DirectRunnerTest.java
je-ik: Moved to `ParDoTest`.

runners/direct-java/src/main/java/org/apache/beam/runners/direct/WatermarkManager.java
je-ik: :+1:

runners/direct-java/src/main/java/org/apache/beam/runners/direct/WatermarkManager.java
je-ik: That is correct. The problem is that `WatermarkManager` works on level of `PTransform`s. I can push the logic down to `TransformWatermarks`, which is the first level, that is somewhat aware of keys (in `extractFiredTimers`), but that would mean, that timers extracted from `AppliedPTransformInputWatermark` would have to be setup again. That is because the logic requires that extracted timer to be fired, or setup again (in the way that `StatefulParDoEvaluator` pushes them back). I think that this makes the logic far too much complicated given that what we gain is the ability to extract and execute timers for various keys of the same PTransform in parallel. I must emphasize, that this is related only to DirectRunner, which is already pretty slow, because it does many validations and its main purpose is testing. And moreover, I'm not convinced that implementing this optimization in reality would bring and significant performance boost (speculative claim, I didn't do any performance testing). Maybe we can merge it as it is and solve performance issues after if they appear (avoid premature optimization with significant impact on readability and understandability of already somewhat complicated code)?

je-ik:
Run Flink ValidatesRunner

je-ik:
Run Spark ValidatesRunner

je-ik:
Run Dataflow ValidatesRunner

je-ik:
Retest this please

je-ik:
Run Flink ValidatesRunner

je-ik:
Run Spark ValidatesRunner

je-ik:
Run Dataflow ValidatesRunner

je-ik:
Run Samza ValidatesRunner

je-ik:
Retest this please

je-ik:
Run Dataflow ValidatesRunner

je-ik:
Run Flink ValidatesRunner

je-ik:
Run Spark ValidatesRunner

je-ik:
Run Samza ValidatesRunner

je-ik:
@kennknowles Unfortunately, samza (at least) doesn't pass the added @ValidatesRunner tests. What would be the suggested procedure? I can create JIRA to address this, but I'm not sure how to make the test @Ignore in samza. The only option I see would be to add a new @Category and ignore that category. Is there a better way?

je-ik:
Run Jet ValidatesRunner

je-ik:
Run SQL PostCommit

kennknowles:
You can exclude the tests by adding a category and tweaking this list: https://github.com/apache/beam/blob/master/runners/samza/build.gradle#L80

je-ik:
Run Dataflow ValidatesRunner

je-ik:
Run Flink ValidaatesRunner

je-ik:
Run Flink ValidatesRunner

je-ik:

je-ik:
Run Flink ValidatesRunner

je-ik:
Run Flink ValidatesRunner

je-ik:
Run Samza ValidatesRunner

je-ik:
Run Samza ValidatesRunner

je-ik:
Run Flink ValidatesRunner

je-ik:

je-ik:
@kennknowles this seems to be ready from my side, tests passing. WDYT?

kennknowles:
Looks like you need one more "ignore" for the Flink portable ValidatesRunner and Jira for the bug to that runner.

je-ik:
Oops, this looks like an endless story. :-)

je-ik:

je-ik:
Run Dataflow ValidatesRunner

je-ik:
@kennknowles sooo, now it is good, hupefully :crossed_fingers:

runners/direct-java/src/test/java/org/apache/beam/runners/direct/DirectRunnerTest.java
kennknowles: Got it. Makes sense.

kennknowles:
My reading of the commit history is that there are many commits which are not meaningful by themselves. There may be some commits that stand alone. But I will squash these commits when I merge unless you would like to do your own custom squashing into some meaningful small set of commits.

je-ik:
I'll definitely squash the history to some compact isolated commits before merging. Thanks for the approval!

je-ik:
I also plan to create the follow-up JIRAs after the merge and notify dev list.

kennknowles:
Missed one: https://builds.apache.org/job/beam_PostCommit_Java_PVR_Flink_Batch/

je-ik:
Yikes! Thanks, solution in #9869

sdks/java/core/src/test/java/org/apache/beam/sdk/transforms/ParDoTest.java
```diff
@@ -3488,6 +3499,158 @@ public void onTimer(OutputReceiver<String> r) {
...
+    public void testEventTimeTimerOrderingWithCreate() throws Exception {
+      final int numTestElements = 100;
+      final Instant now = new Instant(1500000000000L);
+
+      List<TimestampedValue<KV<String, String>>> elements = new ArrayList<>();
+      for (int i = 0; i < numTestElements; i++) {
+        elements.add(TimestampedValue.of(KV.of("dummy", "" + i), now.plus(i)));
+      }
+
+      testEventTimeTimerOrderingWithInputPTransform(
```
lukecwik, line 3544: The elements in the PCollection produced by the Create transform are considered unordered so the stateful DoFn could see (dummy, 1) before it sees (dummy, 0). Doesn't the DoFn need to be marked with `@RequiresTimeSortedInput` in this case?

je-ik: The purpose of the test is to verify, that timers are fired in timestamp order, not that elements arrive ordered. The test already mimics some of the functionality of `@RequiresTimeSortedInput`, which should solve the issue you mention. This test should be a requirement (`UsesStrictTimerOrdering`) for a runer to be able to support `@RequiresTimeSortedInput`. Have you seen any issues with the test?

PR #9203, opened 2019-07-31
Add "--workerHarnessContainerImage=" to dataflow java11 legacy worker test

boyuanzz:
Without overriding --workerHarnessContainerImage, the pipeline will still use cached worker image, instead of worker jar.

R: @pabloem


pabloem:
Boyuan, sorry, what does this do? It's setting the option to nothing?

boyuanzz:
> Boyuan, sorry, what does this do? It's setting the option to nothing?

Yes, that's in purpose. The dataflow worker jar can be only used when both  `--dataflowWorkerJar=${dataflowLegacyWorkerJar}` and `--workerHarnessContainerImage=` are set up. If only setting up` --dataflowWorkerJar` without setting `--workerHarnessContainerImage` to empty, dataflow artifacts service will still pick up cached worker container.

pabloem:
Okay, thanks Boyuan. Merging.

pabloem:
cc: @mwalenia @lgajowy

PR #9218, opened 2019-08-01
[BEAM-7874], [BEAM-7873] Distributed FnApiRunner bugfixs

Hannah-Jiang:
**[BEAM-7874]**
We hard coded max_workers=10 for grpc servers. In case we pass a direct_num_workers greater than 10, pipeline hangs because not all workers can connect to the runner. Pass num_workers to grpc server instead of hard coding it to 10.

**[BEAM-7873]**
A deadlock would happen with subprocess.Popen() with py2, add a locker to prevent deadlock.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/runners/portability/fn_api_runner.py
+++ b/sdks/python/apache_beam/runners/portability/fn_api_runner.py
@@ -1134,11 +1134,11 @@ class GrpcServer(object):
 
   _DEFAULT_SHUTDOWN_TIMEOUT_SECS = 5
 
-  def __init__(self, state, provision_info):
+  def __init__(self, state, provision_info, num_workers):
     self.state = state
     self.provision_info = provision_info
     self.control_server = grpc.server(
-        futures.ThreadPoolExecutor(max_workers=10))
+        futures.ThreadPoolExecutor(max_workers=num_workers))
     self.control_port = self.control_server.add_insecure_port('[::]:0')
     self.control_address = 'localhost:%s' % self.control_port
 
@@ -1148,12 +1148,12 @@ class GrpcServer(object):
     no_max_message_sizes = [("grpc.max_receive_message_length", -1),
                             ("grpc.max_send_message_length", -1)]
     self.data_server = grpc.server(
-        futures.ThreadPoolExecutor(max_workers=10),
+        futures.ThreadPoolExecutor(max_workers=num_workers),
         options=no_max_message_sizes)
     self.data_port = self.data_server.add_insecure_port('[::]:0')
 
     self.state_server = grpc.server(
-        futures.ThreadPoolExecutor(max_workers=10),
+        futures.ThreadPoolExecutor(max_workers=num_workers),
         options=no_max_message_sizes)
     self.state_port = self.state_server.add_insecure_port('[::]:0')
 
@@ -1386,7 +1386,8 @@ class WorkerHandlerManager(object):
     # assume it's using grpc if environment is not EMBEDDED_PYTHON.
     if environment.urn != python_urns.EMBEDDED_PYTHON and \
         self._grpc_server is None:
-      self._grpc_server = GrpcServer(self._state, self._job_provision_info)
+      self._grpc_server = GrpcServer(
+          self._state, self._job_provision_info, num_workers)
     worker_handler_list = self._cached_handlers[environment_id]
     if len(worker_handler_list) < num_workers:
       for _ in range(len(worker_handler_list), num_workers):
--- a/sdks/python/apache_beam/runners/portability/fn_api_runner.py
+++ b/sdks/python/apache_beam/runners/portability/fn_api_runner.py
@@ -1134,11 +1134,12 @@ class GrpcServer(object):
 
   _DEFAULT_SHUTDOWN_TIMEOUT_SECS = 5
 
-  def __init__(self, state, provision_info, num_workers):
+  def __init__(self, state, provision_info, max_workers):
     self.state = state
     self.provision_info = provision_info
+    self.max_workers = max_workers
     self.control_server = grpc.server(
-        futures.ThreadPoolExecutor(max_workers=num_workers))
+        futures.ThreadPoolExecutor(max_workers=self.max_workers))
     self.control_port = self.control_server.add_insecure_port('[::]:0')
     self.control_address = 'localhost:%s' % self.control_port
 
@@ -1148,12 +1149,12 @@ class GrpcServer(object):
     no_max_message_sizes = [("grpc.max_receive_message_length", -1),
                             ("grpc.max_send_message_length", -1)]
     self.data_server = grpc.server(
-        futures.ThreadPoolExecutor(max_workers=num_workers),
+        futures.ThreadPoolExecutor(max_workers=self.max_workers),
         options=no_max_message_sizes)
     self.data_port = self.data_server.add_insecure_port('[::]:0')
 
     self.state_server = grpc.server(
-        futures.ThreadPoolExecutor(max_workers=num_workers),
+        futures.ThreadPoolExecutor(max_workers=self.max_workers),
         options=no_max_message_sizes)
     self.state_port = self.state_server.add_insecure_port('[::]:0')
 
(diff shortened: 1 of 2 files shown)


Hannah-Jiang:
R: @robertwb 
I just noticed this issue. We want to include this patch to 2.15.
We have one more issue with py2 subprocess lib. If the workaround is reasonable, we want to include that patch to 2.15 as well, so please don't merge this PR yet.

sdks/python/apache_beam/runners/portability/local_job_service.py
```diff
@@ -200,10 +202,13 @@ def run(self):
...
       env_dict['WORKER_ID'] = self._worker_id
 
-    p = subprocess.Popen(
-        self._worker_command_line,
-        shell=True,
-        env=env_dict)
+    # deadlock would happen with py2, so add a lock here.
+    # no problem with py3.
+    with SubprocessSdkWorker._lock:
+      p = subprocess.Popen(
```
Hannah-Jiang: It turns out that deadlock happens with Popen(), rather than with p.wait().
Ran 1500 of wordcount.py with 3 workers and confirmed deadlock doesn't happen.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1319,55 +1319,62 @@ def stop_worker(self):
...
-      if status == 'running':
-        break
-      elif status in ('dead', 'exited'):
-        subprocess.call([
+    with DockerSdkWorkerHandler._lock:
+      try:
+        subprocess.check_call(['docker', 'pull', self._container_image])
+      except Exception:
+        logging.info('Unable to pull image %s' % self._container_image)
+      self._container_id = subprocess.check_output(
```
Hannah-Jiang: I just noticed that DockerSdkWorkerHandler can be improved. It calls `apache_beam.runners.worker.sdk_worker_main` at bootstrap, so  rather than creating multi containers, we only need to pass process # to sdk_worker_main. I am going to work on customized container in Q3, will improve it as part of customized container project.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1134,11 +1134,12 @@ class GrpcServer(object):
 
   _DEFAULT_SHUTDOWN_TIMEOUT_SECS = 5
 
-  def __init__(self, state, provision_info):
+  def __init__(self, state, provision_info, num_workers):
     self.state = state
     self.provision_info = provision_info
+    max_workers = max(10, num_workers)
```
robertwb: How is this safe?

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1319,55 +1320,62 @@ def stop_worker(self):
 @WorkerHandler.register_environment(common_urns.environments.DOCKER.urn,
                                     beam_runner_api_pb2.DockerPayload)
 class DockerSdkWorkerHandler(GrpcWorkerHandler):
+
+  _lock = threading.Lock()
```
robertwb: Add some comments as to why this is needed. Is it to workaround subprocess issues with Python 2? If so, perhaps lift it up as an appropriately named module global (as it should be shared with the SubProcess worker handler below).

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1386,7 +1386,8 @@ def get_worker_handlers(self, environment_id, num_workers):
     # assume it's using grpc if environment is not EMBEDDED_PYTHON.
     if environment.urn != python_urns.EMBEDDED_PYTHON and \
         self._grpc_server is None:
-      self._grpc_server = GrpcServer(self._state, self._job_provision_info)
+      self._grpc_server = GrpcServer(
+          self._state, self._job_provision_info, num_workers)
```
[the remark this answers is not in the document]
Hannah-Jiang: Now I think I understand your comment, hopefully my understanding is correct.
Here is how I decide to throw an error.
keep `max_workers` with GrpcServer(), and whenever we need more threads than `max_workers`, it throws out an error. 
`max_workers = num_workers * len(self._environments)`
I still don't understand why we need to multiply ` len(self._environments)` and how `num_workers` may change at each stage. Isn't if fixed?

sdks/python/apache_beam/runners/portability/fn_api_runner.py
Hannah-Jiang: no guarantee it's safe, instead, I throw out an error when we need more threads than we have.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
Hannah-Jiang: Yes, this is a workaround for subprocess.

I tried to add a function to handle calls to subprocess, but, I was not able to find a good way to handle it. With `DockerSdkWorkerHandler`, we can lock `start_worker()`, but we cannot lock `start_worker()` with `SubprocessSdkWorkerHandler` handler. We should lock `subprocess.Popen()` only with `SubprocessSdkWorkerHandler`, if we lock the whole `start_worker()` function, it waits a worker to finish job, so the pipeline gets stuck.

In fact, I doubt  `DockerSdkWorkerHandler` works with FnApi at the moment. FnApi doesn't stage any artifacts, so docker bootstrap fails and container cannot be created. I will make a change to `DockerSdkWorkerHandler` soon and create only one container and handle multithreading/processing when a job running within container, so we can get rid of lock here for `DockerSdkWorkerHandler`.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
robertwb: So should we simply remove this cap at 10?

sdks/python/apache_beam/runners/portability/fn_api_runner.py
robertwb: > Yes, this is a workaround for subprocess.

Comments would be good for future readers of this code. 

> I tried to add a function to handle calls to subprocess, but, I was not able to find a good way to handle it. With `DockerSdkWorkerHandler`, we can lock `start_worker()`, but we cannot lock `start_worker()` with `SubprocessSdkWorkerHandler` handler. We should lock `subprocess.Popen()` only with `SubprocessSdkWorkerHandler`, if we lock the whole `start_worker()` function, it waits a worker to finish job, so the pipeline gets stuck.

I meant moving the lock out of DockerSdkWorkerHandler and SubprocessSdkWorkerHandler to a common place, e.g. a global _subprocess_lock. 

> In fact, I doubt `DockerSdkWorkerHandler` works with FnApi at the moment. FnApi doesn't stage any artifacts, so docker bootstrap fails and container cannot be created. 

It does work; this is how we run Java pipelines on the FnApiRunner. 

> I will make a change to `DockerSdkWorkerHandler` soon and create only one container and handle multithreading/processing when a job running within container, 

For all SDKs? 

> so we can get rid of lock here for `DockerSdkWorkerHandler`.

The lock is a log against all other calls to subprocess.

Hannah-Jiang:
Hopefully now all the comments are resolved. :)

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1297,6 +1298,11 @@ def stop_worker(self):
     self.worker_thread.join()
 
 
+# deadlock would happen with subprocess lib, so create a global lock to share
```
robertwb: Perhaps state this more declaratively. E.g. 

# The subprocesses module is not threadsafe on Python 2.7. Use this lock to prevent concurrent calls to POpen.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1386,7 +1396,21 @@ def get_worker_handlers(self, environment_id, num_workers):
...
+          num_workers * len(self._environments))
+
+    # each gRPC server is running with fixed number of threads (num_workers
+    # * len(self._environments)), where num_workers is from the first call to
+    # get_worker_handlers(). Assumption here is a worker has a connection to a
+    # gRPC server. In case a stage tries to add more workers
+    # than the threads we have, some workers cannot attached to gRPC and
+    # pipeline will hang, so raise an error here.
+    if self._grpc_server is not None and \
+        num_workers > self._grpc_server.num_threads:
```
robertwb: This is not correct, the condition is `num_workers * len(self._environments) > ...`. Here is where naming would help. 

You can create a variable `max_total_workers = num_workers * len(self._environments)`. Then you can both pass this in and compare against it. 

(While we're at it, I would restructure the if statements here to be a bit clearer. E.g.

```
if environment.urn == python_urns.EMBEDDED_PYTHON:
  pass  # no need for a GrpcServer
elif self._grpc_server is None:
  self._grpc_server = ...
elif max_num_workers > self._grpc_server.max_workers:
  raise ...

Hannah-Jiang: Thanks Robert, again, I learned a lot from you.

PR #9231, opened 2019-08-02
One Problen need help to answer

maskwang520:
Can I change the beam from gradle to maven, because i  need use beam in some other maven project


tvalentyn:
@maskwang520 there were reasons why Beam switched from Maven to Gradle, see [1,2] for previous discussions. Note that Beam publishes artifacts to Maven repositories during each release.

If you have a specific question about using Beam with Maven, I encourage you to ask it on user@beam.apache.org mailing list. 

We should continue the discussion on a mailing list and close this PR. Thank you.
 
[1] https://lists.apache.org/thread.html/bd399ecb17cd211be7c6089b562c09ba9116649c9eabe3b609606a3b@%3Cdev.beam.apache.org%3E
[2] https://lists.apache.org/thread.html/eaeab09978b9d8fc10cd112795d994dfeaa3172c48f6af847db273b2@%3Cdev.beam.apache.org%3E

PR #8370, opened 2019-04-21
[BEAM-3489] add PubSub messageId in PubsubMessage

thinhha:
Make the `recordId` field from `PubsubClient.IncomingMessage` available as a new field called `messageId` in `PubsubMessage`.
Updated the coder for `PubsubMessage` to encode and decode the `messageId`.

------------------------

Thank you for your contribution! Follow this checklist to help us incorporate your contribution quickly and easily:

 - [x] [**Choose reviewer(s)**](https://beam.apache.org/contribute/#make-your-change) and mention them in a comment (`R: @username`).
 - [x] Format the pull request title like `[BEAM-XXX] Fixes bug in ApproximateQuantiles`, where you replace `BEAM-XXX` with the appropriate JIRA issue, if applicable. This will automatically link the pull request to the issue.
 - [ ] If this contribution is large, please file an Apache [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

Lang | SDK | Apex | Dataflow | Flink | Gearpump | Samza | Spark
--- | --- | --- | --- | --- | --- | --- | ---
Go | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go/lastCompletedBuild/) | --- | --- | --- | --- | --- | ---
[...]

PR #9232, opened 2019-08-02
[BEAM-7819] Python - parse PubSub message_id into attributes property

matt-darwin:
According to the documentation, system generated metadata should be in the attributes property of the PubSub message, however this is not the case for message_id which is a seperate property of the protobuf PubSub message.

Parse this and add to the attributes dictionary of the beam PubSub message.

Amend tests to expect message_id k,v to be in the attributes property.

------------------------

 - [X ] [**Choose reviewer(s)**](https://beam.apache.org/contribute/#make-your-change) and mention them in a comment (`R: @username`).
 - [ X] Format the pull request title like `[BEAM-XXX] Fixes bug in ApproximateQuantiles`, where you replace `BEAM-XXX` with the appropriate JIRA issue, if applicable. This will automatically link the pull request to the issue.
 - [ X] If this contribution is large, please file an Apache [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

--- | --- | --- | --- | ---


matt-darwin:
@aaltay @udim

sdks/python/apache_beam/io/gcp/pubsub.py
```diff
@@ -127,6 +127,10 @@ def _from_message(msg):
     """
     # Convert ScalarMapContainer to dict.
     attributes = dict((key, msg.attributes[key]) for key in msg.attributes)
+    # Parse the PubSub message_id and add to attributes
+    if msg.message_id != None:
```
aaltay: Could message_id be None?

matt-darwin: It shouldn't be, was simply being defensive.  I've missed something off here however, and this is only working for the directrunner at the moment as I hadn't amended the _from_protobuf function.  The parsing of the inbound object is slightly different there, so I'm just checking whether this can be done in this manner now, and will amend (and remove the unnecessary check).

matt-darwin: Bit odd, the message_id is returned, but the value currently always seems to be blank, I've tried using the id_label parameter as well to specify message_id, but this still returns nothing.  Will continue to investigate.

lukecwik:
FYI, there is #8370 which is attempting to add the message id to the Beam Java SDK.

udim:
Note that in Dataflow, the Python SDK uses FnAPI to read from Pub/Sub using a Java harness. In other words, the message id might be missing because Java code doesn't provide it somehow. I'll look into it later.

When running locally using DirectRunner, there's a different implementation that replaces `ReadFromPubSub` entirely, which probably needs modification as well. See:
https://github.com/apache/beam/blob/02267999542c2ae0f52d217faa14b32c93b8787d/sdks/python/apache_beam/runners/direct/transform_evaluator.py#L380

matt-darwin:
> Note that in Dataflow, the Python SDK uses FnAPI to read from Pub/Sub using a Java harness. In other words, the message id might be missing because Java code doesn't provide it somehow. I'll look into it later.
> 
I'm afraid my Java is very weak; much worse 


> When running locally using DirectRunner, there's a different implementation that replaces `ReadFromPubSub` entirely, which probably needs modification as well. See:
> https://github.com/apache/beam/blob/02267999542c2ae0f52d217faa14b32c93b8787d/sdks/python/apache_beam/runners/direct/transform_evaluator.py#L380



> FYI, there is #8370 which is attempting to add the message id to the Beam Java SDK.

I'll have a look at building the runner from this change to see if I can then access the message_id from the dataflow runner.

matt-darwin:
> Note that in Dataflow, the Python SDK uses FnAPI to read from Pub/Sub using a Java harness. In other words, the message id might be missing because Java code doesn't provide it somehow. I'll look into it later.
> 
> When running locally using DirectRunner, there's a different implementation that replaces `ReadFromPubSub` entirely, which probably needs modification as well. See:
> https://github.com/apache/beam/blob/02267999542c2ae0f52d217faa14b32c93b8787d/sdks/python/apache_beam/runners/direct/transform_evaluator.py#L380

The directrunner has been working ok with the above changes; I think the issue is on the dataflow runner side.  The directrunner is using the _from_message method, and this is parsing correctly and returning the pubsub message id in my testing so far.

 ```
 def _read_from_pubsub(self, timestamp_attribute):
    from apache_beam.io.gcp.pubsub import PubsubMessage
    from google.cloud import pubsub

    def _get_element(message):
      parsed_message = PubsubMessage._from_message(message)
```

**Comment by matt-darwin** [comment]

**Comment by matt-darwin** [comment]

**Comment by matt-darwin** [comment]
Run Python
Dataflow ValidatesRunner

**Comment by matt-darwin** [comment]

**Comment by matt-darwin** [comment]

File: sdks/python/apache_beam/io/gcp/pubsub.py [hunk 2]
```diff
@@ -127,6 +130,10 @@ def _from_message(msg):
     """
     # Convert ScalarMapContainer to dict.
     attributes = dict((key, msg.attributes[key]) for key in msg.attributes)
+    # Parse the PubSub message_id and add to attributes
+    sysattribs = dict(message_id=msg.message_id)
```
udim: This would override a user-defined 'message_id' attribute. Better to make message_id a member of PubsubMessage.

matt-darwin: Yes I agree.  This was merely done to fit with the existing documentation, however I feel that making the PubSubMessage mimic the protobuf definition makes more sense (and can add the publish_time attribute as well).  In addition, due to the handling of pubsub in dataflow, the message_id is currently not exposed and so is simply an empty attribute at that point. I'm thinking to close this pull request and update the issue with a dependency on the dataflow worker being amended (unfortunately I'm even worse at java than python!)

udim: I investigated the issue yesterday and found the issue was with the Windmill service leaving out message id and publish_time. We can go ahead with this PR, and leave a note saying message_ids will work once Windmill service is updated (point to BEAM-7819 for details).

FYI, I believe PR #8370 will also require the Windmill update for message ids to work on Dataflow (but I'm not sure when Java SDK uses Windmill).

matt-darwin: I've started to look at moving message_id out of the attributes and into it's own attribute, along with publish_time. The latter causes some issues with testing, as it cannot be updated in the protobuf, so the way in which some of the mocked up tests work doesn't quite work.

As such, I'm going to focus on message_id first, & have a think how to deal with the publish_time later.

matt-darwin:
Need to fix up some tests, and the PubSubMessage tests need amending to be in line with those. Haven't had chance to test with the dataflowrunner as yet.

matt-darwin:
Ok, so the PubSub timestamp is actually the google.protobuf.timestamp_pb2.Timestamp type. Have amended everything to expect this for the publish_time attribute

matt-darwin:

matt-darwin:

matt-darwin:
Little bit of a mare there with the Timestamp, went down a bit of a rabbit hole for a while!

All good now, and passing tests ok.

@udim @aaltay good for another pair of eyes now having moved message_id and publish_time to the PubSubMessage itself rather than parsing out to attributes.

sdks/python/apache_beam/io/gcp/pubsub.py
```diff
@@ -61,29 +61,43 @@ class PubsubMessage(object):
     attributes: (dict) Key-value map of str to str, containing both user-defined
       and service generated attributes (such as id_label and
       timestamp_attribute). May be None.
+    message_id: (string) Message_Id as generated by PubSub
+    publish_time: (timestamp) Published time of message as generated by PubSub
+                  as per google.protobuf.timestamp_pb2.Timestamp
+
+    N.B. message_id and publish_time are not yet populated by Windmill service
```
aaltay: Users will probably be not familiar windmill service. Should we rephrase this as:

"TODO(BEAM-7819):  message_id and publish_time are not populated on Dataflow runner"

sdks/python/apache_beam/io/gcp/pubsub_test.py
```diff
@@ -82,25 +83,37 @@ def test_proto_conversion(self):
     self.assertEqual(m_converted.attributes, attributes)
 
   def test_eq(self):
-    a = PubsubMessage(b'abc', {1: 2, 3: 4})
-    b = PubsubMessage(b'abc', {1: 2, 3: 4})
-    c = PubsubMessage(b'abc', {1: 2})
+    publish_time_secs = 1520861821
```
aaltay: Would it be simpler to reduce these 3 lines to a single line as:

```
    publish_time = Timestamp(seconds=1520861821,
                             nanos=234567000)
```

sdks/python/apache_beam/io/gcp/pubsub_test.py
```diff
@@ -82,25 +83,37 @@ def test_proto_conversion(self):
...
+    publish_time_nanos = 234567000
+    publish_time = Timestamp(seconds=publish_time_secs,
+                             nanos=publish_time_nanos)
+    a = PubsubMessage(b'abc', {1: 2, 3: 4}, '1234567890', publish_time)
+    b = PubsubMessage(b'abc', {1: 2, 3: 4}, '1234567890', publish_time)
+    c = PubsubMessage(b'abc', {1: 2}, '1234567890', publish_time)
```
aaltay: Do we need to add additional tests cases to check new equality constraints on different message ids and timestamps?

sdks/python/apache_beam/io/gcp/tests/pubsub_matcher_test.py
```diff
@@ -37,8 +37,15 @@
 except ImportError:
   pubsub = None
 
+try:
+  from google.protobuf.timestamp_pb2 import Timestamp
```
aaltay: Could this move into the same try statement as above? Do we expect protobuf and pubsub to be not installed at the same time?

sdks/python/scripts/run_pylint.sh
```diff
@@ -104,6 +104,7 @@ ISORT_EXCLUDED=(
   "model.py"
   "taxi.py"
   "process_tfma.py"
+  "pubsub_test.py"
```
aaltay: What is the isort problem related to this file?

sdks/python/apache_beam/io/gcp/tests/pubsub_matcher_test.py
matt-darwin: Ah I didn't realise you could have imports in the same try statement.  Fixing this up, which should also sort out the isort issues below (will handle it in the same way on pubsub_tests.py)

sdks/python/scripts/run_pylint.sh
matt-darwin: It was insisting I place the protobuf import prior to the beam imports, but then if I did that, complaining about the grouping with the pubsub import.  However, as mentioned above, I didn't realise you could nest imports in a try/catch, so this should resolve this.

sdks/python/apache_beam/io/gcp/pubsub_test.py
matt-darwin: Yes, I will add to the hash and repr tests as well; a check to ensure that both a and b do not equal d/e with some differing message_id/publish_time data).

sdks/python/apache_beam/io/gcp/pubsub_test.py
matt-darwin: Yes no problem; I'd copied and pasted this from another test where the seconds and nanos were used seperately; but it's not needed here.

sdks/python/apache_beam/io/gcp/pubsub.py
matt-darwin: Ah ok.  WIll update.

udim:
FYI, support for message_id and publish_time in Dataflow should be available in a future update of Dataflow. Until then, those fields will appear unset or blank.

matt-darwin:
> FYI, support for message_id and publish_time in Dataflow should be available in a future update of Dataflow. Until then, those fields will appear unset or blank.

Is that external to the Beam project?

matt-darwin:

udim:
> Is that external to the Beam project?

Yes

matt-darwin:

matt-darwin:
Back to the isort error now: 

```

Running isort...
--
ERROR: /home/jenkins/jenkins-slave/workspace/beam_PreCommit_Python_Commit/src/sdks/python/test-suites/tox/py2/build/srcs/sdks/python/apache_beam/io/gcp/pubsub_test.py Imports are incorrectly sorted.
--- /home/jenkins/jenkins-slave/workspace/beam_PreCommit_Python_Commit/src/sdks/python/test-suites/tox/py2/build/srcs/sdks/python/apache_beam/io/gcp/pubsub_test.py:before	2019-08-17 06:59:14.752504
+++ /home/jenkins/jenkins-slave/workspace/beam_PreCommit_Python_Commit/src/sdks/python/test-suites/tox/py2/build/srcs/sdks/python/apache_beam/io/gcp/pubsub_test.py:after	2019-08-17 07:01:17.704126
@@ -26,6 +26,7 @@
 
import hamcrest as hc
import mock
+from google.protobuf.timestamp_pb2 import Timestamp
 
import apache_beam as beam
from apache_beam.io.gcp.pubsub import PubsubMessage
@@ -52,7 +53,6 @@
from apache_beam.transforms.display_test import DisplayDataItemMatcher
from apache_beam.utils import timestamp
 
-from google.protobuf.timestamp_pb2 import Timestamp
# Protect against environments where the PubSub library is not available.
try:
from google.cloud import pubsub
Command exited with non-zero status 1
```

@aaltay are you happy for the file to be excluded?

matt-darwin:

aaltay:
@matt-darwin - Re: isort - Sure, it is fine to exclude it from isort. Thank you for the explanation.

matt-darwin:
@aaltay and @udim was there anything further you wanted changing on this?

matt-darwin:

aaltay:
@udim do you have additional comments? Can we merge this?

sdks/python/scripts/run_pylint.sh
udim: Can this exclusion be removed now?

matt-darwin: Every time I try to remove it, unfortunately I get a pylint error.

Please see https://github.com/apache/beam/pull/9232#issuecomment-519044616 for details.

aaltay:
The last comment was also addressed. Merging this.

udim:
Post-commits are failing, opened: https://issues.apache.org/jira/browse/BEAM-8153

aaltay:
Thank you @udim. Sent you https://github.com/apache/beam/pull/9479 to revert this PR. We can revert first than address the test issue.

PR #9268, opened 2019-08-05
[BEAM-7738] Add external transform support to PubsubIO

chadrik:
Add support for using PubSubIO as an external transform, so that it works from python on portable runners like Flink.

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/pubsub/PubsubIO.java
+++ b/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/pubsub/PubsubIO.java
@@ -72,7 +72,6 @@ import org.apache.beam.sdk.values.PCollection;
 import org.apache.beam.sdk.values.PDone;
 import org.apache.beam.sdk.values.Row;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.annotations.VisibleForTesting;
-import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Charsets;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.MoreObjects;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.ImmutableMap;
 import org.joda.time.Instant;
@@ -745,22 +744,19 @@ public class PubsubIO {
       @Override
       public PTransform<PBegin, PCollection<T>> buildExternal(External.Configuration config) {
         if (config.topic != null) {
-          StaticValueProvider<String> topic = StaticValueProvider.of(utf8String(config.topic));
+          StaticValueProvider<String> topic = StaticValueProvider.of(config.topic);
           setTopicProvider(NestedValueProvider.of(topic, new TopicTranslator()));
         }
         if (config.subscription != null) {
-          StaticValueProvider<String> subscription =
-              StaticValueProvider.of(utf8String(config.subscription));
+          StaticValueProvider<String> subscription = StaticValueProvider.of(config.subscription);
           setSubscriptionProvider(
               NestedValueProvider.of(subscription, new SubscriptionTranslator()));
         }
         if (config.idAttribute != null) {
-          String idAttribute = utf8String(config.idAttribute);
-          setIdAttribute(idAttribute);
+          setIdAttribute(config.idAttribute);
         }
         if (config.timestampAttribute != null) {
-          String timestampAttribute = utf8String(config.timestampAttribute);
-          setTimestampAttribute(timestampAttribute);
+          setTimestampAttribute(config.timestampAttribute);
         }
         setPubsubClientFactory(FACTORY);
         setNeedsAttributes(config.needsAttributes);
@@ -795,25 +791,25 @@ public class PubsubIO {
       public static class Configuration {
 
         // All byte arrays are UTF-8 encoded strings
-        @Nullable private byte[] topic;
-        @Nullable private byte[] subscription;
-        @Nullable private byte[] idAttribute;
-        @Nullable private byte[] timestampAttribute;
+        @Nullable private String topic;
+        @Nullable private String subscription;
+        @Nullable private String idAttribute;
+        @Nullable private String timestampAttribute;
         private boolean needsAttributes;
 
-        public void setTopic(@Nullable byte[] topic) {
+        public void setTopic(@Nullable String topic) {
           this.topic = topic;
         }
 
-        public void setSubscription(@Nullable byte[] subscription) {
+        public void setSubscription(@Nullable String subscription) {
           this.subscription = subscription;
         }
 
-        public void setIdLabel(@Nullable byte[] idAttribute) {
+        public void setIdLabel(@Nullable String idAttribute) {
           this.idAttribute = idAttribute;
         }
 
-        public void setTimestampAttribute(@Nullable byte[] timestampAttribute) {
+        public void setTimestampAttribute(@Nullable String timestampAttribute) {
           this.timestampAttribute = timestampAttribute;
         }
 
@@ -1064,16 +1060,14 @@ public class PubsubIO {
       @Override
       public PTransform<PCollection<T>, PDone> buildExternal(External.Configuration config) {
         if (config.topic != null) {
-          StaticValueProvider<String> topic = StaticValueProvider.of(utf8String(config.topic));
+          StaticValueProvider<String> topic = StaticValueProvider.of(config.topic);
           setTopicProvider(NestedValueProvider.of(topic, new TopicTranslator()));
         }
         if (config.idAttribute != null) {
-          String idAttribute = utf8String(config.idAttribute);
-          setIdAttribute(idAttribute);
+          setIdAttribute(config.idAttribute);
         }
         if (config.timestampAttribute != null) {
-          String timestampAttribute = utf8String(config.timestampAttribute);
-          setTimestampAttribute(timestampAttribute);
+          setTimestampAttribute(config.timestampAttribute);
         }
         SimpleFunction<T, PubsubMessage> parseFn =
             (SimpleFunction<T, PubsubMessage>) new FormatPayloadFromPubsubMessageProto();
@@ -1098,19 +1092,19 @@ public class PubsubIO {
       public static class Configuration {
 
         // All byte arrays are UTF-8 encoded strings
-        private byte[] topic;
-        @Nullable private byte[] idAttribute;
-        @Nullable private byte[] timestampAttribute;
+        private String topic;
+        @Nullable private String idAttribute;
+        @Nullable private String timestampAttribute;
 
-        public void setTopic(byte[] topic) {
+        public void setTopic(String topic) {
           this.topic = topic;
         }
 
-        public void setIdLabel(@Nullable byte[] idAttribute) {
+        public void setIdLabel(@Nullable String idAttribute) {
           this.idAttribute = idAttribute;
         }
 
-        public void setTimestampAttribute(@Nullable byte[] timestampAttribute) {
+        public void setTimestampAttribute(@Nullable String timestampAttribute) {
           this.timestampAttribute = timestampAttribute;
         }
       }
@@ -1415,9 +1409,5 @@ public class PubsubIO {
     public PubsubMessage apply(PubsubMessage input) {
       return input;
     }
-  }
-
-  private static String utf8String(byte[] bytes) {
-    return new String(bytes, Charsets.UTF_8);
   }
 }
--- a/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/pubsub/PubsubIO.java
+++ b/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/pubsub/PubsubIO.java
@@ -813,8 +813,10 @@ public class PubsubIO {
           this.timestampAttribute = timestampAttribute;
         }
 
-        public void setWithAttributes(Long needsAttributes) {
-          this.needsAttributes = needsAttributes >= 1;
+        public void setWithAttributes(Boolean needsAttributes) {
+          // we must use Boolean instead of boolean because the external payload system
+          // inspects the native type of each coder urn, and BooleanCoder wants Boolean.
+          this.needsAttributes = needsAttributes;
         }
       }
     }
--- a/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/pubsub/PubsubIO.java
+++ b/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/pubsub/PubsubIO.java
@@ -20,7 +20,10 @@ package org.apache.beam.sdk.io.gcp.pubsub;
 import static org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Preconditions.checkState;
 
 import com.google.api.client.util.Clock;
+import com.google.auto.service.AutoService;
 import com.google.auto.value.AutoValue;
+import com.google.protobuf.ByteString;
+import com.google.protobuf.InvalidProtocolBufferException;
 import com.google.protobuf.Message;
 import java.io.IOException;
 import java.io.Serializable;
@@ -39,9 +42,11 @@ import org.apache.beam.sdk.PipelineRunner;
 import org.apache.beam.sdk.annotations.Experimental;
 import org.apache.beam.sdk.annotations.Experimental.Kind;
 import org.apache.beam.sdk.coders.AvroCoder;
+import org.apache.beam.sdk.coders.ByteArrayCoder;
 import org.apache.beam.sdk.coders.Coder;
 import org.apache.beam.sdk.coders.CoderException;
 import org.apache.beam.sdk.coders.StringUtf8Coder;
+import org.apache.beam.sdk.expansion.ExternalTransformRegistrar;
 import org.apache.beam.sdk.extensions.protobuf.ProtoCoder;
 import org.apache.beam.sdk.io.gcp.pubsub.PubsubClient.OutgoingMessage;
 import org.apache.beam.sdk.io.gcp.pubsub.PubsubClient.ProjectPath;
@@ -53,6 +58,7 @@ import org.apache.beam.sdk.options.ValueProvider.StaticValueProvider;
 import org.apache.beam.sdk.schemas.Schema;
 import org.apache.beam.sdk.schemas.utils.AvroUtils;
 import org.apache.beam.sdk.transforms.DoFn;
+import org.apache.beam.sdk.transforms.ExternalTransformBuilder;
 import org.apache.beam.sdk.transforms.MapElements;
 import org.apache.beam.sdk.transforms.PTransform;
 import org.apache.beam.sdk.transforms.ParDo;
@@ -66,6 +72,7 @@ import org.apache.beam.sdk.values.PCollection;
 import org.apache.beam.sdk.values.PDone;
 import org.apache.beam.sdk.values.Row;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.annotations.VisibleForTesting;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Charsets;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.MoreObjects;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.ImmutableMap;
 import org.joda.time.Instant;
@@ -705,7 +712,8 @@ public class PubsubIO {
     abstract Builder<T> toBuilder();
 
     @AutoValue.Builder
-    abstract static class Builder<T> {
+    abstract static class Builder<T>
+        implements ExternalTransformBuilder<External.Configuration, PBegin, PCollection<T>> {
       abstract Builder<T> setTopicProvider(ValueProvider<PubsubTopic> topic);
 
       abstract Builder<T> setPubsubClientFactory(PubsubClient.PubsubClientFactory clientFactory);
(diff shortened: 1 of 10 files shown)


sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/pubsub/PubsubIO.java
```diff
@@ -674,6 +680,85 @@ public String toString() {
...
+          String timestampAttribute = utf8String(config.timestampAttribute);
+          setTimestampAttribute(timestampAttribute);
+        }
+        setNeedsAttributes(config.needsAttributes);
+        setPubsubClientFactory(FACTORY);
+        if (config.needsAttributes) {
+          SimpleFunction<PubsubMessage, T> parseFn =
+              (SimpleFunction<PubsubMessage, T>) new IdentityMessageFn();
+          setParseFn(parseFn);
+          // FIXME: call setCoder(). need to use PubsubMessage proto to be compatible with python
```
chadrik: In the case that an external sdk has requested `needsAttributes`, this transform needs to produce `PubsubMessage` instances rather than byte arrays.  There's a wrinkle here:  python deserializes `PubsubMessage` using protobufs, whereas Java encodes with `PubsubMessageWithAttributesCoder`. 

I believe we want to use protobufs from Java.  Can I get confirmation of that, please?

sdks/python/apache_beam/io/external/gcp/pubsub.py
```diff
@@ -0,0 +1,131 @@
...
+      args['topic'] = _encode_str(self.topic)
+
+    if self.subscription is not None:
+      args['subscription'] = _encode_str(self.subscription)
+
+    if self.id_label is not None:
+      args['id_label'] = _encode_str(self.id_label)
+
+    # FIXME: how do we encode a bool so that Java can decode it?
+    # args['with_attributes'] = _encode_bool(self.with_attributes)
```
chadrik: What's the right way to encode a bool so that Java can decode it correctly?

chadrik:
R: @robertwb
R: @mxm
R: @lukecwik
R: @chamikaramj

sdks/python/apache_beam/io/external/gcp/pubsub.py
chadrik: I encoded as int and handled the cast in java

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/pubsub/PubsubIO.java
chadrik: I serialized the `PubsubMessage` using protobufs.  Since there's no cross-language coder for `PubsubMessage`, and I assumed it would be overreach to add one, I used the bytes coder and then handled converting to and from protobufs in code that lives close to the transforms.

mxm:
@chadrik This is awesome. Thanks a lot for your work! Let's rebase this once #9098 is in. I'll have a look at this very soon.

chadrik:
Thanks @mxm.  This is actually not yet working yet, and I'm getting some mysterious errors deep within flink wrt serialization.  I'm working on writing tests and gathering as much info as I can so that I can report back here.  So far I'm pretty perplexed, but I'm new to Beam, Flink, and Java, so not a big surprise!  Once I have more info, it would be great to have your input.

chadrik:
Here's some info on where I am with this.  I could really use some help to push this over the finish line.

The expansion service runs correctly, sends the expanded transforms back to python, but the job fails inside Java because it's trying to use the incorrect serializer.  There's a good chance that I'm overlooking something very obvious.

Here's the stack trace:

```
2019-08-16 16:04:58,297 INFO  org.apache.flink.runtime.taskmanager.Task                     - Source: PubSubInflow/ExternalTransform(beam:external:java:pubsub:read:v1)/PubsubUnboundedSource/Read(PubsubSource) (1/1) (93e578d8ae737c7502685cd978413a98) switched from RUNNING to FAILED.
java.lang.RuntimeException: org.apache.beam.sdk.io.gcp.pubsub.PubsubMessage cannot be cast to [B
	at org.apache.flink.streaming.runtime.io.RecordWriterOutput.pushToRecordWriter(RecordWriterOutput.java:110)
	at org.apache.flink.streaming.runtime.io.RecordWriterOutput.collect(RecordWriterOutput.java:89)
	at org.apache.flink.streaming.runtime.io.RecordWriterOutput.collect(RecordWriterOutput.java:45)
	at org.apache.flink.streaming.api.operators.AbstractStreamOperator$CountingOutput.collect(AbstractStreamOperator.java:718)
	at org.apache.flink.streaming.api.operators.AbstractStreamOperator$CountingOutput.collect(AbstractStreamOperator.java:696)
	at org.apache.flink.streaming.api.operators.StreamSourceContexts$ManualWatermarkContext.processAndCollect(StreamSourceContexts.java:305)
	at org.apache.flink.streaming.api.operators.StreamSourceContexts$WatermarkContext.collect(StreamSourceContexts.java:394)
	at org.apache.beam.runners.flink.translation.wrappers.streaming.io.UnboundedSourceWrapper.emitElement(UnboundedSourceWrapper.java:342)
	at org.apache.beam.runners.flink.translation.wrappers.streaming.io.UnboundedSourceWrapper.run(UnboundedSourceWrapper.java:268)
	at org.apache.flink.streaming.api.operators.StreamSource.run(StreamSource.java:93)
	at org.apache.flink.streaming.api.operators.StreamSource.run(StreamSource.java:57)
	at org.apache.flink.streaming.runtime.tasks.SourceStreamTask.run(SourceStreamTask.java:97)
	at org.apache.flink.streaming.runtime.tasks.StoppableSourceStreamTask.run(StoppableSourceStreamTask.java:45)
	at org.apache.flink.streaming.runtime.tasks.StreamTask.invoke(StreamTask.java:300)
	at org.apache.flink.runtime.taskmanager.Task.run(Task.java:711)
	at java.lang.Thread.run(Thread.java:745)
Caused by: java.lang.ClassCastException: org.apache.beam.sdk.io.gcp.pubsub.PubsubMessage cannot be cast to [B
	at org.apache.beam.sdk.coders.ByteArrayCoder.encode(ByteArrayCoder.java:41)
	at org.apache.beam.sdk.coders.LengthPrefixCoder.encode(LengthPrefixCoder.java:56)
	at org.apache.beam.sdk.values.ValueWithRecordId$ValueWithRecordIdCoder.encode(ValueWithRecordId.java:105)
	at org.apache.beam.sdk.values.ValueWithRecordId$ValueWithRecordIdCoder.encode(ValueWithRecordId.java:81)
	at org.apache.beam.sdk.util.WindowedValue$FullWindowedValueCoder.encode(WindowedValue.java:578)
	at org.apache.beam.sdk.util.WindowedValue$FullWindowedValueCoder.encode(WindowedValue.java:569)
	at org.apache.beam.sdk.util.WindowedValue$FullWindowedValueCoder.encode(WindowedValue.java:529)
	at org.apache.beam.runners.flink.translation.types.CoderTypeSerializer.serialize(CoderTypeSerializer.java:87)
	at org.apache.flink.streaming.runtime.streamrecord.StreamElementSerializer.serialize(StreamElementSerializer.java:175)
	at org.apache.flink.streaming.runtime.streamrecord.StreamElementSerializer.serialize(StreamElementSerializer.java:46)
	at org.apache.flink.runtime.plugable.SerializationDelegate.write(SerializationDelegate.java:54)
	at org.apache.flink.runtime.io.network.api.serialization.SpanningRecordSerializer.serializeRecord(SpanningRecordSerializer.java:78)
	at org.apache.flink.runtime.io.network.api.writer.RecordWriter.emit(RecordWriter.java:160)
	at org.apache.flink.runtime.io.network.api.writer.RecordWriter.emit(RecordWriter.java:128)
	at org.apache.flink.streaming.runtime.io.RecordWriterOutput.pushToRecordWriter(RecordWriterOutput.java:107)
	... 15 more
```

Here's the serializer that `CoderTypeSerializer.serialize` is using before the exception:

```
WindowedValue$FullWindowedValueCoder(
  ValueWithRecordId$ValueWithRecordIdCoder(
    LengthPrefixCoder(
      ByteArrayCoder)
  ),
  GlobalWindow$Coder
) 
```

Here's the value that's being serialized:

```
TimestampedValueInGlobalWindow{
  value=ValueWithRecordId{
    id=[54, 56, 54, 48, 48, 52, 48, 49, 49, 57, 55, 54, 52, 53, 48], 
    value=org.apache.beam.sdk.io.gcp.pubsub.PubsubMessage@64d3b2a
  }, 
  timestamp=2019-08-16T20:04:29.230Z, 
  pane=PaneInfo.NO_FIRING
}
```

Obviously ByteArrayCoder is not the right serializer for PubsubMessage. 

Note that I get the same error whether `with_attributes` is enabled or disabled.

So why is `ByteArrayCoder` being used?

The graph that's generated after expansion looks right to me.  Here's a table of the primitive transforms and their PCollections (note the last row is a python pardo).

| transform name                                                                     | transform type                                              | out collection | out collection coder             |   |
|------------------------------------------------------------------------------------|-------------------------------------------------------------|----------------|----------------------------------|---|
| external_1root/PubsubUnboundedSource/Read(PubsubSource)                            | beam:transform:read:v1                           | PubsubMessage  | PubsubMessageWithAttributesCoder |   |
| external_1root/PubsubUnboundedSource/PubsubUnboundedSource.Stats/ParMultiDo(Stats) | beam:transform:pardo:v1 of StatsFn                          | PubsubMessage  | PubsubMessageWithAttributesCoder |   |
| external_1root/MapElements/Map/ParMultiDo(Anonymous)                               | beam:transform:pardo:v1 of ParsePayloadAsPubsubMessageProto | byte[]         | BytesCoder                       |   |
| ref_AppliedPTransform_PubSubInflow/Map(_from_proto_str)_4                          | beam:transform:pardo:v1 of _from_proto_str                  | PubsubMessage  | Pickle     |   |

So as far as the Pipeline definition is concerned, it seems like the output of `Read(PubsubSource)` is properly associated with the `PubsubMessageWithAttributesCoder`, but when when the job runs its using `BytesCoder`.  So somehow the wires are getting crossed with `ParMultiDo(Anonymous)`, which uses `BytesCoder`.

Any ideas?

The last thing that's important to note is that I needed `PubsubIO.Read` to output a data type that is compatible with Python, so I wrote a custom parseFn that converts each message to a protobuf byte array.  So the PCollection that bridges the divide between Java and Python uses a BytesCoder, but the transforms on either side do the work of converting to and from protobuf representation of a PubsubMessage.

chadrik:
Ok, I now know how the `PubsubMessageWithAttributesCoder` is getting swapped with a `BytesCoder`, but it turns out it's intentional and I don't know why.

In `FlinkStreamingPortablePipelineTranslator.translateRead` when the coder is initiated it is passed through `LengthPrefixUnknownCoders.addLengthPrefixedCoder` which silently replaces all non-model coders with `LengthPrefixCoder(BytesCoder)`.  This _seems_ like the sort of thing that should print a warning, since I assume that a broken pipeline is a likely outcome.

I'm unclear why this coder swap is necessary.  The java part of this pipeline is a `Read -> ParDo -> ParDo`, shouldn't this segment be able to utilize java-only coders (i.e. non-model coder)?

What's the proper solution here?  The last java ParDo is the one that's ensuring we have a byte array for sending to python, but evidently this needs to be happening in the Read itself?

chadrik:
PubsubIO is officially working on Flink!

Some notes:
- In order for this to work from python, this PR needs to be merged: https://github.com/apache/beam/pull/9426
- This PR is _not_ based on my pending "user-friendly external transform" PR: https://github.com/apache/beam/pull/9098.  There are still some issues that I need to work out with @mxm on that one, so I think this should go first and I'll update that one afterward.

There's one giant caveat:  this actually does not work unless you have this special commit:  https://github.com/chadrik/beam/commit/d12b99084809ec34fcf0be616e94301d3aca4870.  It turns out the same is currently true for KafkaIO external transform support, which came as quite a surprise.  The Jira issue for this problem is here:  https://jira.apache.org/jira/browse/BEAM-7870.  I'd love to see some movement on this.  Happy to help where I can!

chadrik:
I think this PR is ready to merge, but I could use a little help understanding how to resolve the test failures.

Java PreCommit has no test failures, so I'm not sure why it's marked as failed here. Is it because of the warnings?  Jenkins seems to imply that anything over 12 warnings is an error, but I don't see any warnings in files that I changed:
 https://builds.apache.org/job/beam_PreCommit_Java_Commit/7551/

Likewise Python PreCommit list no failures [in jenkins](https://builds.apache.org/job/beam_PreCommit_Python_Commit/8356/testReport/), but is marked as failed here.

And Portable_Python PreCommit just seems to have timed out.

chadrik:
Run Portable_Python PreCommit

chadrik:
Run Python PreCommit

chadrik:
Hi.  Still hoping for a little help bringing this to a close.

mxm:
Build results have expired: https://builds.apache.org/job/beam_PreCommit_Java_Commit/7551/

mxm:
Retest this please

mxm:
@chadrik Does this have to be rebased now that #9098 has been merged?

chadrik:
> Does this have to be rebased now that #9098 has been merged?

It will work as is, but it should be converted to the new API to provide more real-world examples.  I'll do that today.  

If you could help me diagnose the Java test failure it would be much appreciated!

mxm:
Will take a look ASAP, on a low time budget at the moment :)

chamikaramj:
Seems like the Java PreCommit failure is for the new test ?

Stacktrace is:
org.apache.beam.sdk.io.gcp.pubsub.PubsubIOExternalTest > testConstructPubsubRead FAILED
    java.lang.RuntimeException at PubsubIOExternalTest.java:97
        Caused by: java.lang.RuntimeException at PubsubIOExternalTest.java:97
            Caused by: java.lang.reflect.InvocationTargetException at PubsubIOExternalTest.java:97
                Caused by: java.lang.IllegalStateException at PubsubIOExternalTest.java:97

chamikaramj:
Python failure seems to be a docs issue (:sdks:python:test-suites:tox:py2:docs'.)

Could be a flake. Retrying.

chamikaramj:
Run Python PreCommit

mxm:
@chadrik Seems like you fixed the issue. Java tests are passing now: https://builds.apache.org/job/beam_PreCommit_Java_Commit/7889/ The one test failure is unrelated to your changes: 
```
org.apache.beam.sdk.transforms.ParDoLifecycleTest.testTeardownCalledAfterExceptionInStartBundleStateful
```

chadrik:
This is now waiting on the BooleanCoder in python

chadrik:
Run Python PreCommit

chadrik:
Run Java PreCommit

mxm:
Any update here?

chadrik:
We are doing some final testing in production today and then I’ll hopefully
give you the thumbs up.

-chad

chadrik:
Run Python PreCommit

chadrik:
Run Python PreCommit

chadrik:
Run Python PreCommit

chadrik:
Run Python PreCommit

chadrik:
@mxm   This is tested on prem and ready to go!  (with the caveat that the special commit to add dependencies is still required for end users, as is the case with kafka: chadrik@d12b990)

sdks/python/apache_beam/io/external/gcp/pubsub.py
```diff
@@ -0,0 +1,160 @@
...
+        ('subscription', typing.Optional[unicode]),
+        ('id_label', typing.Optional[unicode]),
+        ('with_attributes', bool),
+        ('timestamp_attribute', typing.Optional[unicode]),
+    ]
+)
+
+
+class ReadFromPubSub(beam.PTransform):
+  """An external ``PTransform`` for reading from Cloud Pub/Sub."""
```
mxm: Should we mention here or in the module comments that this experimental and requires the non-standard coders to be in place?

sdks/python/apache_beam/io/external/gcp/pubsub.py
```diff
@@ -0,0 +1,160 @@
...
+        ('id_label', typing.Optional[unicode]),
+        # this is not implemented yet on the Java side:
+        # ('with_attributes', bool),
+        ('timestamp_attribute', typing.Optional[unicode]),
+    ]
+)
+
+
+class WriteToPubSub(beam.PTransform):
+  """An external ``PTransform`` for writing messages to Cloud Pub/Sub."""
```
mxm: Same as above.

sdks/python/apache_beam/io/external/gcp/pubsub.py
mxm: Can be done in a follow-up.

chadrik:
> Perhaps we should even add the additional commits as part of Beam, at least until we have worked around the coders problems?

I think it would be great for this to work out of the box, but I'm not sure what the ramifications of adding those commits would be.   They don't just modify the standard coders, they also change the standard beam java docker containers by adding additional dependencies.  Is it safe to add those even if people aren't using pubsub/kafka?   Is the only downside larger container sizes?

@TheNeuralBit How close are we to having schema coders ready to use?  IIUC, we should be able to register a converter between Row -> PubsubMessage, right?   What's the mechanism for that?  It would be great to put that to use here as soon as the schema coders are ready.

TheNeuralBit:
> How close are we to having schema coders ready to use? IIUC, we should be able to register a converter between Row -> PubsubMessage, right? What's the mechanism for that? It would be great to put that to use here as soon as the schema coders are ready.

@robertwb gave #9188 a LGTM, I just need to get CI passing and I think we can merge it. I'll work on that now.

Are you thinking you'd use beam:coder:row:v1 as the interface for the external transform, and the Java ExternalTransform implementations would handle the conversion of Row to/from PubsubMessage? There's no trivial way to register a converter between Row and PubsubMessage since the latter isn't structured, but of course on the Java side we could have code to serialize the Row to a variety of formats to put in the PubsubMessage payload: Avro, JSON, or the row serialization format itself (although I'm not sure we'd want to encourage using that outside of Beam), would be pretty simple to add. Maybe the format to use could be part of the external transform payload.

chadrik:
> Are you thinking you'd use beam:coder:row:v1 as the interface for the external transform, and the Java ExternalTransform implementations would handle the conversion of Row to/from PubsubMessage? 

There are two places that I see  beam:coder:row:v1 being useful:

1. as a way to declare the construction interface of an external transform, and encode its values.  A schema coder would replace the `configuration` mapping in `pipeline.ExternalConfiguration.ExternalConfigurationPayload` proto.
2. as a coder for structured elements that are exchanged between sdks

Right now I'm mostly thinking about the latter, which is when `PubsubMessage` comes into play.

> There's no trivial way to register a converter between Row and PubsubMessage since the latter isn't structured, 

Maybe I'm thinking about this wrong, but I think the `PubsubMessage` _is_ structured:

```java
public class PubsubMessage {

  private byte[] message;
  private Map<String, String> attributes;
  private String messageId;

  /** Returns the main PubSub message. */
  public byte[] getPayload() {
    return message;
  }

  /** Returns the full map of attributes. This is an unmodifiable map. */
  public Map<String, String> getAttributeMap() {
    return attributes;
  }

  /** Returns the messageId of the message populated by Cloud Pub/Sub. */
  @Nullable
  public String getMessageId() {
    return messageId;
  }
```

I'm not a Java expert by any means, but this seems like a type that would work with AutoValue, we just need to rename `message` to `payload` and `attributes` to `attributeMap`.

What are the requirements for registering a Row converter?

> but of course on the Java side we could have code to serialize the Row to a variety of formats to put in the PubsubMessage payload: Avro, JSON, or the row serialization format itself (although I'm not sure we'd want to encourage using that outside of Beam), would be pretty simple to add. 

I think the payload is not a concern when it comes to portability of external transforms:  it gets encoded/decoded by another transform, not PubsubRead/Write.  We can just assume that's a byte array.

My grasp on the Java side is a bit tenuous, so I'd like for @mxm to confirm or deny what I've written here.

TheNeuralBit:
> Right now I'm mostly thinking about the latter

Agreed, that's what I'm thinking about too.

> Maybe I'm thinking about this wrong, but I think the PubsubMessage is structured:

Ah ok, fair. I was referring specifically to the structure (or lack thereof) of the byte array payload, but you're right the (Python SDK) user can handle creating a byte array themselves, and the row coder can just encode `{byte[] payload, Map<String, String> attributes, String messageId}`

> What are the requirements for registering a Row converter?

### Java
There are a variety of ways to do it. If it were a class that we had control over we could use the [`DefaultSchema`](https://beam.apache.org/releases/javadoc/2.16.0/org/apache/beam/sdk/schemas/DefaultSchema.html) annotation, as long as one of the included SchemaProvider implementations would work (I think JavaBeanSchema is the closest but wouldn't work because PubsubMessage doesn't have setters). I think what we'd want to do here is just implement a [`SchemaProvider`](https://beam.apache.org/releases/javadoc/2.16.0/org/apache/beam/sdk/schemas/SchemaProvider.html) and a [`SchemaProviderRegistrar`](https://beam.apache.org/releases/javadoc/2.16.0/org/apache/beam/sdk/schemas/SchemaProviderRegistrar.html) for `PubsubMessage` and include it in Beam.

@reuvenlax may have a better suggestion.

### Python
With my PR I think it could look like:
```python
# this is py3 syntax for clarity, but we'd probably
# need to use the TypedDict('PubsubMessage', ...) version
class PubsubMessage(TypedDict):
   message: ByteString
   attributes: Mapping[unicode, unicode]
   messageId: unicode

coders.registry.register_coder(PubsubMessage, coders.RowCoder)

pcoll
  | 'make some messages' >> beam.Map(makeMessage).with_output_types(PubsubMessage)
  | 'write to pubsub' >> beam.io.WriteToPubsub(project, topic) # or something
```

chadrik:
>  If it were a class that we had control over we could use the DefaultSchema annotation, as long as one of the included SchemaProvider implementations would work (I think JavaBeanSchema is the closest but wouldn't work because PubsubMessage doesn't have setters).

It may reduce our overall technical debt if we just implement the full set of setters and getters on `PubsubMessage` and use `DefaultSchema`.  It doesn't seem like it would be a bad thing.  Who would have an opinion on that?

TheNeuralBit:
D'oh my bad, we _do_ have control over `PubsubMessage` :man_facepalming: I assumed it was part of the pubsub client library. Yeah I vote we use `DefaultSchema` with either `JavaBeanSchema` or `JavaFieldSchema`, whichever works with fewer changes.

chadrik:
@mxm I marked kafak and pubsub external transforms as external.  I think this is ready to merge.

chadrik:
@TheNeuralBit  On the python side, as with the Java SDK, there is a custom `PubsubMessage` class (take a look in `apache_beam.io.gcp.pubsub`).  The main thing it provides is methods for converting to/from protobuf.  

In your example above, you registered the `TypedDict` sub-class: will there be a similar system in python for converting between a row type to a custom type, like `apache_beam.io.gcp.pubsub.PubsubMessage`?

TheNeuralBit:
> Will there be a similar system in python for converting between a row type to a custom type

It hasn't been designed, but that's definitely something I'd like to support. Would we actually need the `PubsubMessage` protobuf conversion methods for the external transform version of pubsubio/kafkaio, or are you just thinking it would be good to use the same type for consistency? I'm just wondering if we can get away with a simple `TypedDict` sub-class for now

sdks/python/apache_beam/io/external/gcp/pubsub.py
```diff
@@ -0,0 +1,168 @@
...
+
+from past.builtins import unicode
+
+import apache_beam as beam
+from apache_beam.io.gcp import pubsub
+from apache_beam.transforms import Map
+from apache_beam.transforms.external import ExternalTransform
+from apache_beam.transforms.external import NamedTupleBasedPayloadBuilder
+
+ReadFromPubsubSchema = typing.NamedTuple(
```
chamikaramj: Doesn't have to be done in this PR, but in the future it will be great if we just move external PubSub and Kafka transforms to io/gcp/pubsub.py and io/kafka.py and configure/branch based on pipeline options.

sdks/python/apache_beam/io/external/gcp/pubsub.py
```diff
@@ -0,0 +1,168 @@
...
+        ('topic', typing.Optional[unicode]),
+        ('subscription', typing.Optional[unicode]),
+        ('id_label', typing.Optional[unicode]),
+        ('with_attributes', bool),
+        ('timestamp_attribute', typing.Optional[unicode]),
+    ]
+)
+
+
+class ReadFromPubSub(beam.PTransform):
```
chamikaramj: Should we rename to avoid conflicts with ReadFromPubSub/WriteToPubSub in io/gcp/pubsub.py ?

chadrik:
> Would we actually need the PubsubMessage protobuf conversion methods for the external transform version of pubsubio/kafkaio

We would not need to use PubsubMessage _protobuf_ type any more (and thus we wouldn't need the conversion methods), but we may still want the beam PubsubMessage class.  The latter is yielded by  `io.pubsub.ReadFromPubSub`, so if we ant `io.external.pubsub.ReadFromPubSub` to be a drop-in replacement (which I am in favor of) then that means keeping `io.external.pubsub.PubsubMessage` and adding the ability to convert between this and a row-type.

sdks/python/apache_beam/io/external/gcp/pubsub.py
chadrik: I don't love that idea.  The two sets of xforms are do not have the same arguments, in part because of the need for the expansion service endpoint, but also because the Java versions are not guaranteed to be the same as the python versions (though I would like that to be the case).  Even the expansion service argument could somehow be passed through the options, the two sets of xform classes may have different base classes:  ideally the ones in `io.external` inherit from `ExternalTransform` (they currently do not, but they should be able to once we resolve the issue we're working on with @TheNeuralBit )

chadrik:
@TheNeuralBit let's move the conversation about row coders over to https://jira.apache.org/jira/browse/BEAM-7870

This PR is ready to merge!

chamikaramj:
LGTM. Thanks.

Looks like Max already approved so I'll merge.

chadrik:
Whoohoo!

mxm:
Congrats @chadrik! :) Also thanks for your help @chamikaramj and @TheNeuralBit.

PR #9294, opened 2019-08-08
[BEAM-7917] Fix datastore writes failing on retry

sadovnychyi:
When we call batch.begin() the status is set as IN_PROGRESS and then to FINISHED after *any* call to commit(), successful or not.
commit() can only be called on a batch with status IN_PROGRESS -- it will fail with ValueError otherwise.

I cannot say how safe is it to use those private properties of another library from here, but it fixes the issue. We could also re-create batch from scratch (so we don't have to change that private status) -- but not sure if that would cost us any performance. 

https://github.com/googleapis/google-cloud-python/blob/master/datastore/google/cloud/datastore/batch.py
https://issues.apache.org/jira/browse/BEAM-7917

------------------------

--- | --- | --- | --- | ---


sadovnychyi:
R: @tvalentyn

tvalentyn:
R: @udim who is more familiar with this IO.

sdks/python/apache_beam/io/gcp/datastore/v1new/helper.py
```diff
@@ -100,6 +100,10 @@ def write_mutations(batch, throttler, rpc_stats_callback, throttle_delay=1):
 
   try:
     start_time = time.time()
+    if batch._status == batch._FINISHED:
```
udim: This change is unsafe, since it's dealing with the internal state of the client and the behavior may change unexpectedly in the future.
I believe that a better choice is to create a new batch in this method every time.

sadovnychyi: Agree. I think it's a very good idea to propagate this changes to datastoreio.py as well and avoid building `_batch` there at all, and just accumulate `_mutations` or `_batch_mutations` which we would pass to `write_mutations` and build a new batch every time. This will probably involve test changes as well.

Happy to work on it but would like to confirm this idea.

udim: Sounds good

sdks/python/apache_beam/io/gcp/datastore/v1new/datastoreio.py
```diff
@@ -391,6 +392,7 @@ def add_element_to_batch(self, element):
         raise ValueError('Entities to be written to Cloud Datastore must '
                          'have complete keys:\n%s' % client_entity)
       self._batch.put(client_entity)
+      self._batch_mutations.append((helper._WRITE_MUTATION, client_entity))
```
sadovnychyi: We could also pass it only once as another param to write_mutations, and by adding another property to each subclass.

sadovnychyi:
Pushed an update, please take another look.

If we stop constructing batch object in `DatastoreMutateFn` then it's more complicated to calculate the batch size in bytes so far -- we would still have to convert entity/key into protobuf, so performance hit since we are doing it twice (when adding new element and when building a batch). This *could* be an issue when writing a lot of data.

There's also no public API for commiting a bunch of protobufs directly (which would be perfect for us).

So what I come up with is that we are doing everything just like before, but we also keep original non-protobuf entity/key so we can use those to reconstruct the branch *only* on retry. Should be no performance hit unless we are already failing.

(didn't actually test this yet in the wild, will do once we are clear on implementation)

sdks/python/apache_beam/io/gcp/datastore/v1new/helper.py
```diff
@@ -99,6 +105,19 @@ def write_mutations(batch, throttler, rpc_stats_callback, throttle_delay=1):
     rpc_stats_callback(throttled_secs=throttle_delay)
 
   try:
+    try:
+      batch.begin()
+    except ValueError:
+      # batch has already been started previously (and failed),
+      # we need to build it again from scratch (BEAM-7917)
+      batch = get_client(batch.project, batch.namespace).batch()
+      for mutation_type, mutation_element in mutations:
```
sadovnychyi: *Maybe* we could reconstruct it from scratch by using [https://github.com/googleapis/google-cloud-python/blob/master/datastore/google/cloud/datastore/helpers.py#L105](https://github.com/googleapis/google-cloud-python/blob/master/datastore/google/cloud/datastore/helpers.py#L105) (or key_from_protobuf) assuming that we could get those from batch.mutations. Then we don't have to pass them explicitly but we will perform more encoding/decoding when errors occur.

sadovnychyi: Most likely cannot, mutation_pb is different from entities/keys.

sdks/python/apache_beam/io/gcp/datastore/v1new/datastoreio.py
```diff
@@ -340,12 +340,13 @@ def finish_bundle(self):
     def _init_batch(self):
       self._batch_bytes_size = 0
       self._batch = self._client.batch()
-      self._batch.begin()
+      self._batch_mutations = []
 
     def _flush_batch(self):
       # Flush the current batch of mutations to Cloud Datastore.
       latency_ms = helper.write_mutations(
```
udim: Having both _batch_mutations and _batch seems redundant.
I would move `write_mutations` into this class and create the batch object in it.
You could then take advantage of `add_element_to_batch` to populate the batch.

sdks/python/apache_beam/io/gcp/datastore/v1new/datastoreio.py
udim: I think that moving write_mutations into DatastoreMutateFn will let you use add_element_to_batch instead of pairing a mutation_type with each entity (or passing it once).

sdks/python/apache_beam/io/gcp/datastore/v1new/datastoreio.py
udim: Regarding ByteSize(), we could split add_element_to_batch into 2 part (just a suggestion):

```py
    def element_to_client_batch_item(self, element):
      if not isinstance(element, types.Entity):
        raise ValueError('apache_beam.io.gcp.datastore.v1new.datastoreio.Entity'
                         ' expected, got: %s' % type(element))
      if not element.key.project:
        element.key.project = self._project
      client_entity = element.to_client_entity()
      if client_entity.key.is_partial:
        raise ValueError('Entities to be written to Cloud Datastore must '
                         'have complete keys:\n%s' % client_entity)
      return client_entity

    def add_to_batch(self, client_batch_item):
      self._batch.put(client_entity)
```

udim:
> Pushed an update, please take another look.
> 
Made a pass today. Sorry for the delay!

> If we stop constructing batch object in `DatastoreMutateFn` then it's more complicated to calculate the batch size in bytes so far -- we would still have to convert entity/key into protobuf, so performance hit since we are doing it twice (when adding new element and when building a batch). This _could_ be an issue when writing a lot of data.

I believe it is possible to avoid converting and calculating more than once. See my other comments.

> There's also no public API for commiting a bunch of protobufs directly (which would be perfect for us).

Yes, that's the tradeoff with using the new client.

sdks/python/apache_beam/io/gcp/datastore/v1new/datastoreio.py
sadovnychyi: I've made related changes but still couldn't avoid storing batch and raw batch elements separately.
We could skip the rest of elements (over the bytes limit) when building the batch in `write_mutations` and let them be picked up on the next batch (so we won't purge _batch but there might be some elements left from previous batch) -- but it seems to create more confusion.

aaltay:
What are the next steps for this PR?

sadovnychyi:
Another review pass would be helpful.

aaltay:
@udim could you please make another pass when you get a chance?

udim:
run python precommit

sdks/python/apache_beam/io/gcp/datastore/v1new/datastoreio.py
udim: I think your choice is a valid compromise with what to do with the results of `element_to_client_batch_item`. I see 2 options here: 
1. Save the batch items.
2. Discard batch items after calling `ByteSize()` on them.

The first option seems more CPU efficient while the second seems more memory efficient. I don't know which option is faster, but from my limited experience saving CPU at the expense of RAM seems like a good tradeoff.

PR #9296, opened 2019-08-08
[BEAM-7730] Introduce Flink 1.9 Runner

dmvk:
Runner for Flink 1.9. [JIRA](https://issues.apache.org/jira/browse/BEAM-7730).


dmvk:
@sunjincheng121 Thanks for the review! I'll try to finish this one later this week. I think we can not upgrade to 1.8.1 due to:
- https://issues.apache.org/jira/browse/FLINK-13367
- https://issues.apache.org/jira/browse/FLINK-13369

and will have to wait for 1.8.2 release.

Best,
D.

dmvk:
Run Java PreCommit

sunjincheng121:
I see. Thanks @dmvk

dmvk:
@sunjincheng121 I'm still unable to fix the portability test for timers (the one that is `@Ignored` in this PR). It looks up state for a key from a different key-group during `@OnTimer` execution, which results in NPE. I'm still unsure about the cause. When I change paralellism to 1, it works as expected because all the keys are in a single partition.

Flink PR that caused the "regression" from 1.8 is most likely this one https://issues.apache.org/jira/browse/FLINK-12693.

Any ideas?

sunjincheng121:
Thanks @dmvk, I would like to check it, then feedback you! :)

sunjincheng121:
Hi @dmvk，I creat a hotfix to address the problem. (see: https://github.com/apache/beam/pull/9464)

Currently, state keys are encoded as NESTED while the keys in keyselector(KvToByteBufferKeySelector)
are encoded as OUTER. This makes the keyGroupId of the same key are different and  
lead to the test failure. 
The newest flink version makes this bug more visible.

I have check it in my local with your patch. Could you please have look at the changes of the hotfix and double check it whether the hotfix is make sense to you. :)

Thank you!
Jincheng

dmvk:
@sunjincheng121 Thanks for the help! ;) I've changed the flink side coder to NESTED for now, until #9464 discussion gets resolved. I'm unsure about the implications changing context to OUTER would have on other runners.

dmvk:
@sunjincheng121 All checks are finally green ;) Can please you do the final review?

sunjincheng121:
Thanks for the update @dmvk 
LGTM. +1 to mreged.
Best, Jincheng

runners/flink/src/test/java/org/apache/beam/runners/flink/FlinkStreamingTransformTranslatorsTest.java
```diff
@@ -65,12 +64,12 @@ public void readSourceTranslatorBoundedWithMaxParallelism() {
     env.setParallelism(parallelism);
     env.setMaxParallelism(maxParallelism);
 
-    StreamTransformation<?> sourceTransform =
-        applyReadSourceTransform(transform, PCollection.IsBounded.BOUNDED, env);
+    SourceTransformation<?> sourceTransform =
+        (SourceTransformation)
+            applyReadSourceTransform(transform, PCollection.IsBounded.BOUNDED, env);
```
mxm: Why do we need to cast here, change the variable type?

runners/flink/src/test/java/org/apache/beam/runners/flink/translation/wrappers/streaming/DoFnOperatorTest.java
```diff
@@ -1038,7 +1039,7 @@ void pushbackDataCheckpointing(
 
     assertThat(
         stripStreamRecordFromWindowedValue(testHarness.getOutput()),
-        contains(helloElement, worldElement));
+        containsInAnyOrder(helloElement, worldElement));
```
mxm: Why is the order reversed here?

sdks/java/build-tools/src/main/resources/beam/suppressions.xml
```diff
@@ -92,5 +92,6 @@
   <!-- Checkstyle does not correctly detect package files across multiple source directories. -->
   <suppress checks="JavadocPackage" files=".*runners.flink.*CoderTypeSerializer\.java"/>
   <suppress checks="JavadocPackage" files=".*runners.flink.*EncodedTypeSerializer\.java"/>
+  <suppress checks="JavadocPackage" files=".*runners.flink.*BeamStoppableFunction\.java"/>
```
mxm: Please revert.

sdks/java/harness/src/main/java/org/apache/beam/fn/harness/FnApiDoFnRunner.java
```diff
@@ -203,6 +205,7 @@ public void processTimer(
     currentTimer = timer;
     currentTimeDomain = timeDomain;
     try {
+      @SuppressWarnings("unchecked")
```
mxm: Please revert unrelated changes in this file.

runners/flink/1.9/src/main/java/org/apache/beam/runners/flink/translation/types/EncodedValueSerializer.java
```diff
@@ -0,0 +1,121 @@
...
+import org.apache.beam.sdk.coders.Coder;
+import org.apache.flink.api.common.typeutils.TypeSerializer;
+import org.apache.flink.api.common.typeutils.TypeSerializerSchemaCompatibility;
+import org.apache.flink.api.common.typeutils.TypeSerializerSnapshot;
+import org.apache.flink.api.common.typeutils.base.TypeSerializerSingleton;
+import org.apache.flink.core.memory.DataInputView;
+import org.apache.flink.core.memory.DataOutputView;
+
+/** {@link TypeSerializer} for values that were encoded using a {@link Coder}. */
+public final class EncodedValueSerializer extends TypeSerializerSingleton<byte[]> {
```
mxm: Duplicate code, this class already exists twice. Please explain/document.

dmvk:
@tweise Hi Thhomas, there was no upfront discussion as this was first developed in our internal build and now we are in the process of contributing the runner to beam. The runner is made the exactly same way as the runners for previous flink versions. Unfortunately this time, there were many breaking changes on the flink side (due to stream & batch unification effort), so the problem with the current approach became more visible.

There were discussions on ML about removal of 1.5, 1.6 versions, which would IMO eliminate the overhead to a bearable level (these versions are unsupported by the flink community anyway). Most of the classes that we "copy & paste" between versions are due to 1.5 compatibility.

I'll try to bring this to ML to see, if anyone has a better idea how to handle these differences.

Thanks,
D.

runners/flink/1.5/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/io/BeamStoppableFunction.java
```diff
@@ -0,0 +1,29 @@
...
+import org.apache.flink.api.common.functions.StoppableFunction;
+
+/**
+ * Custom StoppableFunction for backward compatibility.
+ *
+ * @see <a
+ *     href="https://github.com/apache/flink/commit/e95b347dda5233f22fb03e408f2aa521ff924996">Flink
+ *     interface removal commit.</a>
+ */
+public interface BeamStoppableFunction extends StoppableFunction {}
```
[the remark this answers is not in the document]
dmvk: Hi Max, this is just an interface for a backward compatibility.

Originally RichParallelSourceFunction was implementing `StoppableFunction`, which was removed as of Flink 1.9., so the following code wouldn't compile:

https://github.com/apache/beam/blob/9043b3e0bc138e63123d3fa55d4b7280eb96a82e/runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkStreamingTransformTranslators.java#L1400

https://github.com/apache/beam/blob/9043b3e0bc138e63123d3fa55d4b7280eb96a82e/runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/io/UnboundedSourceWrapper.java#L365

I didn't want to break this for the previous runner versions, so I introduced this interface that would handle the transition.

Other options would be completely remove the `stop()` implementation from the above mentioned classes.

WDYT?

runners/flink/1.9/src/main/java/org/apache/beam/runners/flink/translation/types/EncodedValueSerializer.java
dmvk: This is due to the current structure of the flink runner. I can not simply reuse the code from the previous runner versions, because if I need to make a change to at least one of the files, it doesn't compile due to class duplicates.

runners/flink/src/test/java/org/apache/beam/runners/flink/FlinkStreamingTransformTranslatorsTest.java
dmvk: Because `SourceTransformation` and `OneInputTrasformation` no longer extend a common interface, so applyReadSourceTransform needs to return `Object` instead.

dmvk:
It seems that 1.5 and 1.6 discussion was never brought to the ML, I've probably recalled it from a private conversation + [BEAM-7962](https://jira.apache.org/jira/browse/BEAM-7962?jql=project%20%3D%20BEAM%20AND%20text%20~%20%22flink%201.5%22).

runners/flink/1.9/src/main/java/org/apache/beam/runners/flink/translation/types/EncodedValueSerializer.java
mxm: If these are the same files as for the 1.8, we can simply use the 1.8 version by adjusting the source path in `build.gradle`:

```gradle
  main_source_dirs = ["$basePath/src/main/java", "$basePath/1.8/src/main/java"]
```

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/io/UnboundedSourceWrapper.java
```diff
@@ -63,7 +62,10 @@
 /** Wrapper for executing {@link UnboundedSource UnboundedSources} as a Flink Source. */
 public class UnboundedSourceWrapper<OutputT, CheckpointMarkT extends UnboundedSource.CheckpointMark>
     extends RichParallelSourceFunction<WindowedValue<ValueWithRecordId<OutputT>>>
-    implements ProcessingTimeCallback, StoppableFunction, CheckpointListener, CheckpointedFunction {
+    implements ProcessingTimeCallback,
+        BeamStoppableFunction,
+        CheckpointListener,
+        CheckpointedFunction {
```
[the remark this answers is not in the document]
mxm: Seems fair to keep this as per the above discussion.

runners/flink/1.9/build.gradle
```diff
@@ -0,0 +1,34 @@
...
+
+def basePath = '..'
+
+/* All properties required for loading the Flink build script */
+project.ext {
+  // Set the version of all Flink-related dependencies here.
+  flink_version = '1.9.0'
+  // Main source directory and Flink version specific code.
+  main_source_dirs = ["$basePath/src/main/java", "./src/main/java"]
+  test_source_dirs = ["$basePath/src/test/java", "./src/test/java"]
```
mxm:
```suggestion
  test_source_dirs = ["$basePath/src/test/java", "$basePath/1.8/src/test/java", "./src/test/java"]
```

This should get rid of the duplicated code.

runners/flink/1.9/build.gradle
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+def basePath = '..'
+
+/* All properties required for loading the Flink build script */
+project.ext {
+  // Set the version of all Flink-related dependencies here.
+  flink_version = '1.9.0'
+  // Main source directory and Flink version specific code.
+  main_source_dirs = ["$basePath/src/main/java", "./src/main/java"]
```
mxm:
```suggestion
  test_source_dirs = ["$basePath/src/main/java", "$basePath/1.8/src/main/java", "./src/main/java"]
```

This should get rid of the duplicated code.

mxm:
IMHO there is no need to block this PR. When the comments are resolved, this looks good to merge. As for your concerns @tweise, there will be a major simplification in many regards (directory structure, code workarounds) once we remove 1.5. Supporting multiple versions does not come for free, especially when interfacing with the internals of a stream processor. That said, the current layout is very practical, we just need to make sure we do not duplicate code if it is not necessary.

tweise:
@mxm Let's continue the discussion on the ML.

tweise:
> It seems that 1.5 and 1.6 discussion was never brought to the ML, I've probably recalled it from a private conversation + BEAM-7962.

I remember that being a topic several times, but don't recall exactly where. But such change **should** be an explicit discussion on the ML. In any case, there seems to be consensus to drop 1.5/1.6.

tweise:
> IMHO there is no need to block this PR.

AFAIK this PR is blocked pending resolution of the key encoding issue, so that it passes tests. Hopefully in parallel we can make progress on trimming down the code duplication also.

mxm:
@dmvk Let us know when you have addressed the comments. Happy to take another look then.

sunjincheng121:
Hi @dmvk @mxm I have left some comments in both #9484 and #9464.

sdks/java/build-tools/src/main/resources/beam/suppressions.xml
dmvk: BeamStoppableFunction is needed, explanation above.

sdks/java/harness/src/main/java/org/apache/beam/fn/harness/FnApiDoFnRunner.java
dmvk: done

runners/flink/1.5/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/io/BeamStoppableFunction.java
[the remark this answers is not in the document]
dmvk: I'd rather create a JIRA to remove this once we get rid of 1.7 runner as some users may still rely on this.

dmvk:
Run Java PreCommit

dmvk:
Run Python PreCommit

sunjincheng121:

runners/flink/1.9/src/test/java/org/apache/beam/runners/flink/streaming/FlinkStateInternalsTest.java
```diff
@@ -0,0 +1,151 @@
...
+import org.apache.flink.runtime.state.memory.MemoryStateBackend;
+import org.apache.flink.runtime.state.ttl.TtlTimeProvider;
+import org.joda.time.Instant;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+/** Tests for {@link FlinkStateInternals}. This is based on {@link StateInternalsTest}. */
+@RunWith(JUnit4.class)
+public class FlinkStateInternalsTest extends StateInternalsTest {
```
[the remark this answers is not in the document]
dmvk: removed

runners/flink/1.9/src/test/java/org/apache/beam/runners/flink/streaming/FlinkBroadcastStateInternalsTest.java
```diff
@@ -0,0 +1,96 @@
...
+import org.junit.runners.JUnit4;
+
+/**
+ * Tests for {@link FlinkBroadcastStateInternals}. This is based on the tests for {@code
+ * StateInternalsTest}.
+ *
+ * <p>Just test value, bag and combining.
+ */
+@RunWith(JUnit4.class)
+public class FlinkBroadcastStateInternalsTest extends StateInternalsTest {
```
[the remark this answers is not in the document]
dmvk: removed

runners/flink/1.9/src/main/java/org/apache/beam/runners/flink/translation/types/CoderTypeSerializer.java
```diff
@@ -0,0 +1,165 @@
...
+import org.apache.flink.api.common.typeutils.TypeSerializerSnapshot;
+import org.apache.flink.core.io.VersionedIOReadableWritable;
+import org.apache.flink.core.memory.DataInputView;
+import org.apache.flink.core.memory.DataOutputView;
+
+/**
+ * Flink {@link org.apache.flink.api.common.typeutils.TypeSerializer} for Beam {@link
+ * org.apache.beam.sdk.coders.Coder Coders}.
+ */
+public class CoderTypeSerializer<T> extends TypeSerializer<T> {
```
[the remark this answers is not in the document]
dmvk: removed

runners/flink/src/test/java/org/apache/beam/runners/flink/translation/wrappers/streaming/DoFnOperatorTest.java
dmvk: I think this is the last issue, will try to investigate today.

runners/flink/flink_runner.gradle
```diff
@@ -142,7 +165,7 @@ class ValidatesRunnerConfig {
 
 def createValidatesRunnerTask(Map m) {
   def config = m as ValidatesRunnerConfig
-  tasks.create(name: config.name, type: Test) {
+  tasks.register(config.name, Test) {
```
mxm: Why this change?

runners/flink/flink_runner.gradle
```diff
@@ -178,8 +201,8 @@ def createValidatesRunnerTask(Map m) {
 createValidatesRunnerTask(name: "validatesRunnerBatch", streaming: false)
 createValidatesRunnerTask(name: "validatesRunnerStreaming", streaming: true)
 
-task validatesRunner {
-  group = "Verification"
+tasks.register('validatesRunner') {
```
mxm: Why this change?

runners/flink/1.8/build.gradle
```diff
@@ -22,11 +22,9 @@ def basePath = '..'
 project.ext {
   // Set the version of all Flink-related dependencies here.
   flink_version = '1.8.2'
-  // Main source directory and Flink version specific code.
-  main_source_dirs = ["$basePath/src/main/java", "./src/main/java"]
-  test_source_dirs = ["$basePath/src/test/java", "./src/test/java"]
-  main_resources_dirs = ["$basePath/src/main/resources"]
-  test_resources_dirs = ["$basePath/src/test/resources"]
```
mxm: Could we keep the ability to override resource dirs?

runners/flink/1.7/build.gradle
```diff
@@ -22,11 +22,9 @@ def basePath = '..'
 project.ext {
   // Set the version of all Flink-related dependencies here.
   flink_version = '1.7.2'
-  // Main source directory and Flink version specific code.
-  main_source_dirs = ["$basePath/src/main/java", "./src/main/java"]
-  test_source_dirs = ["$basePath/src/test/java", "./src/test/java"]
-  main_resources_dirs = ["$basePath/src/main/resources"]
-  test_resources_dirs = ["$basePath/src/test/resources"]
```
mxm: Could we keep the ability to override resource dirs?

runners/flink/1.9/build.gradle
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+def basePath = '..'
+/* All properties required for loading the Flink build script */
+project.ext {
+  // Set the version of all Flink-related dependencies here.
+  flink_version = '1.9.0'
+  // Version specific code overrides.
+  main_source_overrides = ["${basePath}/1.7/src/main/java", "${basePath}/1.8/src/main/java", './src/main/java']
+  test_source_overrides = ["${basePath}/1.7/src/test/java", "${basePath}/1.8/src/test/java", './src/test/java']
```
mxm: Not sure if that it easier to reason about the source, but at least we do not have code duplication. So looks fine for now.

runners/flink/src/test/java/org/apache/beam/runners/flink/translation/wrappers/streaming/DoFnOperatorTest.java
mxm: We can't guarantee the order here for keys in the state backend after restoring from a savepoint. This just so happens to work for older Flink versions. So seems fair to change this.

mxm:
Thanks @dmvk.

mxm:
Retest this please

mxm:
Run ValidatesRunner

mxm:

mxm:

dmvk:
@mxm Thanks! I'll follow-up with docs, build improvement suggestions and 1.9.1 upgrade.

runners/flink/flink_runner.gradle
dmvk: this is recommended approach since 5.1.x as it creates the task lazily

https://docs.gradle.org/current/userguide/task_configuration_avoidance.html

mxm:
Thanks for letting me know.

PR #9313, opened 2019-08-12
[BEAM-5822] Remove non vendored byte-buddy dependency from sdks/java/core

iemejia:
Just fixing some minor things forgotten after #9272

It also enforces that no Beam module uses the unvendored byte-buddy
dependency in the future.

R: @lukecwik 
CC: @vectorijk @kennknowles


iemejia:
@lukecwik One question, What is the reason to have our vendored deps on `sdks/java/core` in `shadow` mode instead of `compile`?

vectorijk:
lgtm

lukecwik:
Thanks for the follow-up on enforcing it.

PR #9316, opened 2019-08-12
[BEAM-7954] Validate beam modules to disable guava imports from vendored gRPC

iemejia:
R: @lukecwik


lukecwik:
The idea is good but there are likely other occurrences as well since the checkstyle in the java precommit is failing. Will wait on an update to fix other offending locations before merging.

Also, we may need to revert the checkstyle check in the future if a gRPC API ever requires its own guava type to be passed into it. For now it doesn't seem to be the case but it might be worthwhile to add a comment to this regard.

iemejia:
Oups my local regex did not work, fixing now. Yes I thought about the eventual case when guava versions won't be happy but I think is better to try to keep it constrained as much as possible. We'll see when it happens.

lukecwik:
Run Portable_Python PreCommit

PR #9327, opened 2019-08-13
Align vendored and non-vendored gRPC versions

iemejia:
R: @lukecwik


iemejia:
Run Portable_Python PreCommit

iemejia:
Mmmm... looks like hadoop was not happy when I tried to align guava's version so better go for the memonte only with gRPC

iemejia:
Run Portable_Python PreCommit

iemejia:
It is green now. PTAL @lukecwik  (sorry to bother you with all these reviews), will shuffle reviewers now that other people are back from holidays.

lukecwik:
Run Java PostCommit

iemejia:
Oups postcommit failed with panic message, probably better to abort this then, I imagine the dependency web for the GCP deps as daunting. Save if you have any suggestion.

lukecwik:
I think updating gRPC will require a significant effort (dealing with the transitive closure of dependency upgrades as well) in updating all users of gRPC to a newer compatible version.

PR #9330, opened 2019-08-13
[BEAM-7969] Report FnAPI counters as deltas in streaming jobs.

Ardagan:
FnAPI counters were not reported in streaming worker.
Streaming Dataflow jobs do not tie counters to workitems,
but report them as deltas on same workitem instead.

This change picks only latest counters per workitem
from FnAPI harness, treats them as updates and reports
to DFE.

[Tested manually at this point]

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/StreamingDataflowWorker.java
+++ b/runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/StreamingDataflowWorker.java
@@ -400,9 +400,13 @@ public class StreamingDataflowWorker {
   private final StreamingDataflowWorkerOptions options;
   private final boolean windmillServiceEnabled;
   private final long clientId;
+
   private final MetricTrackingWindmillServerStub metricTrackingWindmillServer;
   private final CounterSet pendingDeltaCounters = new CounterSet();
   private final CounterSet pendingCumulativeCounters = new CounterSet();
+  private final java.util.concurrent.ConcurrentLinkedQueue<CounterUpdate> pendingMonitoringInfos =
+      new ConcurrentLinkedQueue<>();
+
   // Map from stage name to StageInfo containing metrics container registry and per stage counters.
   private final ConcurrentMap<String, StageInfo> stageInfoMap = new ConcurrentHashMap();
 
@@ -1294,6 +1298,9 @@ public class StreamingDataflowWorker {
       // Blocks while executing work.
       executionState.getWorkExecutor().execute();
 
+      Iterables.addAll(
+          this.pendingMonitoringInfos, executionState.getWorkExecutor().extractMetricUpdates());
+
       commitCallbacks.putAll(executionState.getContext().flushState());
 
       // Release the execution state for another thread to use.
@@ -1852,7 +1859,6 @@ public class StreamingDataflowWorker {
   /** Sends counter updates to Dataflow backend. */
   private void sendWorkerUpdatesToDataflowService(
       CounterSet deltaCounters, CounterSet cumulativeCounters) throws IOException {
-
     // Throttle time is tracked by the windmillServer but is reported to DFE here.
     windmillQuotaThrottling.addValue(windmillServer.getAndResetThrottleTime());
     if (memoryMonitor.isThrashing()) {
@@ -1867,6 +1873,28 @@ public class StreamingDataflowWorker {
           cumulativeCounters.extractUpdates(false, DataflowCounterUpdateExtractor.INSTANCE));
       counterUpdates.addAll(
           deltaCounters.extractModifiedDeltaUpdates(DataflowCounterUpdateExtractor.INSTANCE));
+      if (hasExperiment(options, "beam_fn_api")) {
+        while (!this.pendingMonitoringInfos.isEmpty()) {
+          final CounterUpdate item = this.pendingMonitoringInfos.poll();
+
+          // This change will treat counter as delta.
+          // This is required because we receive cumulative results from FnAPI harness,
+          // while streaming job is expected to receive delta updates to counters on same
+          // WorkItem.
+          if (item.getCumulative()) {
+            item.setCumulative(false);
+          } else {
+            // In current world all counters coming from FnAPI are cumulative.
+            // This is a safety check in case new counter type appears in FnAPI.
+            throw new UnsupportedOperationException(
+                "FnApi counters are expected to provide cumulative values."
+                    + " Please, update convertion to delta logic"
+                    + " if non-cumulative counter type is required.");
+          }
+
+          counterUpdates.add(item);
+        }
+      }
     }
 
     // Handle duplicate counters from different stages. Store all the counters in a multi-map and
@@ -1926,6 +1954,7 @@ public class StreamingDataflowWorker {
             .setWorkItemId(WINDMILL_COUNTER_UPDATE_WORK_ID)
             .setErrors(errors)
             .setCounterUpdates(counterUpdates);
+
     workUnitClient.reportWorkItemStatus(workItemStatus);
 
     // Send any counters appearing more than once in subsequent RPCs:
--- a/sdks/python/apache_beam/runners/dataflow/dataflow_exercise_streaming_metrics_pipeline.py
+++ b/sdks/python/apache_beam/runners/dataflow/dataflow_exercise_streaming_metrics_pipeline.py
@@ -1,0 +1,103 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""A word-counting workflow."""
+
+from __future__ import absolute_import
+
+import time
+
+from apache_beam.metrics import Metrics
+import argparse
+import logging
+import re
+
+from past.builtins import unicode
+
+import apache_beam as beam
+from apache_beam.options.pipeline_options import PipelineOptions
+from apache_beam.options.pipeline_options import SetupOptions
+from apache_beam.options.pipeline_options import StandardOptions
+
+SLEEP_TIME_SECS = 1
+
+
+class StreamingUserMetricsDoFn(beam.DoFn):
+  """Generates several user metric counters."""
+
+  def __init__(self):
+    self.double_message_counter = Metrics.counter(self.__class__,
+                                                  'double_msg_counter_name')
+    self.msg_len_dist_metric = Metrics.distribution(
+        self.__class__, 'msg_len_dist_metric_name')
+
+  def start_bundle(self):
+    time.sleep(SLEEP_TIME_SECS)
+
+  def process(self, element):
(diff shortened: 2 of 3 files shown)


Ardagan:
R: @lukecwik, @ajamato

aaltay:
cc: @angoenka

Ardagan:
@rohdesamuel

runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/StreamingDataflowWorker.java
```diff
@@ -1867,6 +1873,28 @@ private void sendWorkerUpdatesToDataflowService(
           cumulativeCounters.extractUpdates(false, DataflowCounterUpdateExtractor.INSTANCE));
       counterUpdates.addAll(
           deltaCounters.extractModifiedDeltaUpdates(DataflowCounterUpdateExtractor.INSTANCE));
+      if (hasExperiment(options, "beam_fn_api")) {
+        while (!this.pendingMonitoringInfos.isEmpty()) {
+          final CounterUpdate item = this.pendingMonitoringInfos.poll();
```
ajamato: The SDK harness will give you the full value of the counter on each MonitoringInfo. Don't you need to convert it to a delta, since the last reported value? Or does setting cumulative have DFE solve this internally?

You might want to confirm with @edre.

runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/StreamingDataflowWorker.java
```diff
@@ -1867,6 +1873,28 @@ private void sendWorkerUpdatesToDataflowService(
...
+            item.setCumulative(false);
+          } else {
+            // In current world all counters coming from FnAPI are cumulative.
+            // This is a safety check in case new counter type appears in FnAPI.
+            throw new UnsupportedOperationException(
+                "FnApi counters are expected to provide cumulative values."
+                    + " Please, update convertion to delta logic"
+                    + " if non-cumulative counter type is required.");
+          }
+
```
ajamato: This method essentially has 3 inputs correct? pendingMonitoringInfos, deltaCounters and cumulativeCounters.
And one output counterUpdates

Can you write some unit tests which verifies that counterUpdates is populated properly?

runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/StreamingDataflowWorker.java
```diff
@@ -400,9 +400,13 @@ public int getSize() {
   private final StreamingDataflowWorkerOptions options;
   private final boolean windmillServiceEnabled;
   private final long clientId;
+
   private final MetricTrackingWindmillServerStub metricTrackingWindmillServer;
   private final CounterSet pendingDeltaCounters = new CounterSet();
   private final CounterSet pendingCumulativeCounters = new CounterSet();
+  private final java.util.concurrent.ConcurrentLinkedQueue<CounterUpdate> pendingMonitoringInfos =
```
ajamato: can you please add some comments about how pendingMonitoringInfos is concurrently read/written to? i.e. Which threads are running concurrently and using this

runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/StreamingDataflowWorker.java
edre: It's ok to just report the full value as the delta, but only as long as we only report counters when finishing the work item.

Ardagan: Non-cumulative means delta.

runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/StreamingDataflowWorker.java
Ardagan: done.

runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/StreamingDataflowWorker.java
angoenka: Lets also check for item == null to avoid possible race condition as its called at start, via a timer and at stop.

runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/StreamingDataflowWorker.java
```diff
@@ -1867,6 +1873,28 @@ private void sendWorkerUpdatesToDataflowService(
...
+          // This change will treat counter as delta.
+          // This is required because we receive cumulative results from FnAPI harness,
+          // while streaming job is expected to receive delta updates to counters on same
+          // WorkItem.
+          if (item.getCumulative()) {
+            item.setCumulative(false);
+          } else {
+            // In current world all counters coming from FnAPI are cumulative.
+            // This is a safety check in case new counter type appears in FnAPI.
+            throw new UnsupportedOperationException(
```
angoenka: Shall we log the error and move on?

Ardagan: It is better to fail fast in this case. New type of counter appears only when someone modifies SDK, so it comes from beam developer, not user.

runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/StreamingDataflowWorker.java
Ardagan: We are good here since we first initialize class and metric and only then start multiple threads.

Ardagan:
> This map uses `MetricKey` as the key which can do not contain sdkHarness id and can potentially have name collision across multiple sdkHarness.
> We use multiple SdkHarness in case of python SDK which will have this problem.

Can you elaborate on the problem a bit please? IIUC different SDK harnesses will handle different steps, so metric keys will not overlap anyway.

Also I feel fixing that issue better be done in separate PR.

angoenka:
Thanks for the offline discussion.
It seems reasonable to move with the PR and add the integration tests.

aaltay:
@Ardagan do you know what the test failures are?

Ardagan:
> @Ardagan do you know what the test failures are?

Seem to be some libraries import failure. Does work on my machine. Looking for fix.

Ardagan:
Run Java PreCommit

Ardagan:

angoenka:
The error seems to be for matching the metrics now. Is this a different error?

angoenka:
For missing `google.cloud` module, Beam tests do not install GCP modules. https://github.com/apache/beam/blob/master/sdks/python/setup.py#L131

Ardagan:
Testing this in separate PR not to pollute for reviewers. Current summary.
Seems that I should add this test to ValidatesRunner suite, since ValidatesContainer suite doesn't override worker jar.
Regarding pubsub: Regular tests don't import pubsub, so what we do is import within scope of method. This triggers import only when test is triggered.

runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/StreamingDataflowWorker.java
Ardagan: Done

sdks/python/apache_beam/runners/dataflow/dataflow_exercise_streaming_metrics_pipeline.py
```diff
@@ -0,0 +1,100 @@
...
+import apache_beam as beam
+from apache_beam.metrics import Metrics
+from apache_beam.options.pipeline_options import PipelineOptions
+from apache_beam.options.pipeline_options import SetupOptions
+from apache_beam.options.pipeline_options import StandardOptions
+
+SLEEP_TIME_SECS = 1
+
+
+class StreamingUserMetricsDoFn(beam.DoFn):
```
ajamato: Pleaes add some comments to describe what this test does,

runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/StreamingDataflowWorker.java
ajamato: I think you should not fail in this case. What happens if a someone writes a custom SDK to run on Dataflow. It may introduce different metrics, but you should not fail.

An SDK can also be written by someone outside of beam repo entirely, not necessarily written by a beam developer.

https://docs.google.com/document/d/1MtBZYV7NAcfbwyy9Op8STeFNBxtljxgy69FkHMvhTMA/edit
Customization- Custom SDKs+engines can add their own custom metrics and custom monitored state without modifying beam proto files.

sdks/python/apache_beam/runners/dataflow/dataflow_exercise_streaming_metrics_pipeline.py
```diff
@@ -0,0 +1,100 @@
...
+      help=('Input PubSub subscription of the form '
+            '"projects/<PROJECT>/subscriptions/<SUBSCRIPTION>."'))
+  known_args, pipeline_args = parser.parse_known_args(argv)
+
+  pipeline_options = PipelineOptions(pipeline_args)
+  pipeline_options.view_as(SetupOptions).save_main_session = True
+  pipeline_options.view_as(StandardOptions).streaming = True
+  pipeline = beam.Pipeline(options=pipeline_options)
+
+  _ = (pipeline
```
ajamato: Is there any reason why you did not reuse the existing pipeline integration test and run it in streaming mode? Why did you need to write a new pipeline? Rather than reusing code.

sdks/python/apache_beam/runners/dataflow/dataflow_exercise_streaming_metrics_pipeline_test.py
```diff
@@ -0,0 +1,158 @@
...
+                  'experiment': 'beam_fn_api',
+                  'input_subscription': self.input_sub.name,
+                  'output_topic': self.output_topic.name,
+                 }
+
+    argv = self.test_pipeline.get_full_options_as_args(**extra_opts)
+    return dataflow_exercise_streaming_metrics_pipeline.run(argv)
+
+  @attr('IT', 'ValidatesRunner')
+  def test_streaming_pipeline_returns_expected_user_metrics_fnapi_it(self):
```
ajamato: add comment describing that this test does.

sdks/python/apache_beam/runners/dataflow/dataflow_exercise_streaming_metrics_pipeline.py
Ardagan: Reading from files requires side inputs that is not properly supported in FnApi streaming and causes existing pipeline to fail. Reusing your existing pipeline was the first thing I tried to do.

runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/StreamingDataflowWorker.java
Ardagan: That was bad way to formulate on my side:
CounterUpdates are generated within Java DF Worker. So if we receive different type of CounterUpdate here, that means that someone modified DF worker.

Ardagan:
Run Java PreCommit

Ardagan:
run java precommit

aaltay:
Thank you all for fixing this!

PR #9331, opened 2019-08-13
[BEAM-7966] Write portable Flink application jar

ibzib:
This is the first half of the implementation outlined [here](https://docs.google.com/document/d/1kj_9JWxGWOmSGeZ5hbLVDXSTv-zBrx4kQRqOq85RYD4/edit#). This PR will enable any job with the `--output_executable_path` flag set, when sent to the Flink job server, to be output as an jar. Adding the main method to the jar produced here (or any jar with the same format) will be a separate PR.

R: @angoenka @robertwb @tweise 

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkJobInvoker.java
+++ b/runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkJobInvoker.java
@@ -20,13 +20,14 @@ package org.apache.beam.runners.flink;
 import static org.apache.beam.runners.core.construction.PipelineResources.detectClassPathResourcesToStage;
 
 import java.io.IOException;
-import java.util.List;
 import java.util.UUID;
 import javax.annotation.Nullable;
 import org.apache.beam.model.pipeline.v1.RunnerApi;
 import org.apache.beam.runners.core.construction.PipelineOptionsTranslation;
 import org.apache.beam.runners.fnexecution.jobsubmission.JobInvocation;
 import org.apache.beam.runners.fnexecution.jobsubmission.JobInvoker;
+import org.apache.beam.runners.fnexecution.jobsubmission.PortablePipelineJarCreator;
+import org.apache.beam.runners.fnexecution.jobsubmission.PortablePipelineRunner;
 import org.apache.beam.runners.fnexecution.provisioning.JobInfo;
 import org.apache.beam.sdk.options.PortablePipelineOptions;
 import org.apache.beam.vendor.grpc.v1p21p0.com.google.protobuf.Struct;
@@ -63,7 +64,6 @@ public class FlinkJobInvoker extends JobInvoker {
 
     String invocationId =
         String.format("%s_%s", flinkOptions.getJobName(), UUID.randomUUID().toString());
-    LOG.info("Invoking job {}", invocationId);
 
     if (FlinkPipelineOptions.AUTO.equals(flinkOptions.getFlinkMaster())) {
       flinkOptions.setFlinkMaster(serverConfig.getFlinkMasterUrl());
@@ -74,16 +74,23 @@ public class FlinkJobInvoker extends JobInvoker {
       portableOptions.setSdkWorkerParallelism(serverConfig.getSdkWorkerParallelism());
     }
 
+    PortablePipelineRunner pipelineRunner;
+    if (portableOptions.getOutputExecutablePath() == null
+        || portableOptions.getOutputExecutablePath().isEmpty()) {
+      pipelineRunner =
+          new FlinkPipelineRunner(
+              flinkOptions,
+              serverConfig.getFlinkConfDir(),
+              detectClassPathResourcesToStage(FlinkJobInvoker.class.getClassLoader()));
+    } else {
+      pipelineRunner = new PortablePipelineJarCreator(FlinkPipelineRunner.class);
+    }
+
     flinkOptions.setRunner(null);
 
+    LOG.info("Invoking job {} with pipeline runner {}", invocationId, pipelineRunner);
     return createJobInvocation(
-        invocationId,
-        retrievalToken,
-        executorService,
-        pipeline,
-        flinkOptions,
-        serverConfig.getFlinkConfDir(),
-        detectClassPathResourcesToStage(FlinkJobInvoker.class.getClassLoader()));
+        invocationId, retrievalToken, executorService, pipeline, flinkOptions, pipelineRunner);
   }
 
   static JobInvocation createJobInvocation(
@@ -92,16 +99,13 @@ public class FlinkJobInvoker extends JobInvoker {
       ListeningExecutorService executorService,
       RunnerApi.Pipeline pipeline,
       FlinkPipelineOptions flinkOptions,
-      @Nullable String confDir,
-      List<String> filesToStage) {
+      PortablePipelineRunner pipelineRunner) {
     JobInfo jobInfo =
         JobInfo.create(
             invocationId,
             flinkOptions.getJobName(),
             retrievalToken,
             PipelineOptionsTranslation.toProto(flinkOptions));
-    FlinkPipelineRunner pipelineRunner =
-        new FlinkPipelineRunner(flinkOptions, confDir, filesToStage);
     return new JobInvocation(jobInfo, executorService, pipeline, pipelineRunner);
   }
 }
--- a/runners/flink/src/test/java/org/apache/beam/runners/flink/FlinkSavepointTest.java
+++ b/runners/flink/src/test/java/org/apache/beam/runners/flink/FlinkSavepointTest.java
@@ -206,6 +206,7 @@ public class FlinkSavepointTest implements Serializable {
 
     ListeningExecutorService executorService =
         MoreExecutors.listeningDecorator(Executors.newFixedThreadPool(1));
+    FlinkPipelineOptions pipelineOptions = pipeline.getOptions().as(FlinkPipelineOptions.class);
     try {
       JobInvocation jobInvocation =
           FlinkJobInvoker.createJobInvocation(
@@ -213,9 +214,8 @@ public class FlinkSavepointTest implements Serializable {
               "none",
               executorService,
               pipelineProto,
-              pipeline.getOptions().as(FlinkPipelineOptions.class),
-              null,
-              Collections.emptyList());
+              pipelineOptions,
+              new FlinkPipelineRunner(pipelineOptions, null, Collections.emptyList()));
 
       jobInvocation.start();
 
--- a/runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/PortablePipelineJarCreator.java
+++ b/runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/PortablePipelineJarCreator.java
@@ -1,0 +1,293 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.beam.runners.fnexecution.jobsubmission;
+
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.channels.Channels;
+import java.nio.channels.WritableByteChannel;
+import java.nio.charset.StandardCharsets;
+import java.util.Enumeration;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Set;
+import java.util.UUID;
+import java.util.jar.Attributes;
+import java.util.jar.Attributes.Name;
+import java.util.jar.JarEntry;
+import java.util.jar.JarFile;
+import java.util.jar.JarOutputStream;
+import java.util.jar.Manifest;
+import org.apache.beam.model.jobmanagement.v1.ArtifactApi;
+import org.apache.beam.model.jobmanagement.v1.ArtifactApi.ArtifactChunk;
+import org.apache.beam.model.jobmanagement.v1.ArtifactApi.ArtifactMetadata;
+import org.apache.beam.model.jobmanagement.v1.ArtifactApi.GetArtifactRequest;
+import org.apache.beam.model.jobmanagement.v1.ArtifactApi.GetManifestRequest;
+import org.apache.beam.model.jobmanagement.v1.ArtifactApi.GetManifestResponse;
+import org.apache.beam.model.jobmanagement.v1.ArtifactApi.ProxyManifest;
+import org.apache.beam.model.jobmanagement.v1.ArtifactApi.ProxyManifest.Location;
+import org.apache.beam.model.jobmanagement.v1.ArtifactRetrievalServiceGrpc;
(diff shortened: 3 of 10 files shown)


ibzib:
Run Java PreCommit

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/PortablePipelineJarCreator.java
```diff
@@ -0,0 +1,289 @@
...
+   *
+   * @return A {@link ProxyManifest} pointing to the artifacts' location in the output jar.
+   */
+  @VisibleForTesting
+  static ProxyManifest copyStagedArtifacts(
+      String retrievalToken, JarOutputStream outputStream, ArtifactRetriever retrievalServiceStub)
+      throws IOException {
+    GetManifestRequest manifestRequest =
+        GetManifestRequest.newBuilder().setRetrievalToken(retrievalToken).build();
+    ArtifactApi.Manifest manifest = retrievalServiceStub.getManifest(manifestRequest).getManifest();
```
angoenka: This require double copying of the artifacts. 1. To staging location 2. In to the jar.
This should be ok for now but will it be easy use artifact staging service to write directly to the jar.
Also, should we delete the temporarily staged artifacts?

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/PortablePipelineJarCreator.java
```diff
@@ -0,0 +1,289 @@
...
+
+  private static void writePipelineOptions(Struct optionsProto, JarOutputStream outputStream)
+      throws IOException {
+    JarEntry jarEntry = new JarEntry(PortablePipelineJarUtils.PIPELINE_OPTIONS_FILE_NAME);
+    outputStream.putNextEntry(jarEntry);
+    optionsProto.writeTo(outputStream);
+  }
+
+  @VisibleForTesting
+  interface ArtifactRetriever {
```
angoenka: Can we directly use ArtifactRetrievalServiceBlockingStub and mock it where needed?

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/PortablePipelineJarCreator.java
```diff
@@ -0,0 +1,289 @@
...
+  static ProxyManifest copyStagedArtifacts(
+      String retrievalToken, JarOutputStream outputStream, ArtifactRetriever retrievalServiceStub)
+      throws IOException {
+    GetManifestRequest manifestRequest =
+        GetManifestRequest.newBuilder().setRetrievalToken(retrievalToken).build();
+    ArtifactApi.Manifest manifest = retrievalServiceStub.getManifest(manifestRequest).getManifest();
+    // Create a new proxy manifest to locate artifacts at jar runtime.
+    ProxyManifest.Builder proxyManifestBuilder = ProxyManifest.newBuilder().setManifest(manifest);
+    for (ArtifactMetadata artifact : manifest.getArtifactList()) {
+      String outputPath = PortablePipelineJarUtils.ARTIFACT_FOLDER_NAME + "/" + artifact.getName();
```
angoenka: The actual name of the artifact might have special characters so we should not put it in the path.
I think such issues can be avoided if we directly use JarArticactSageingService.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/PortablePipelineJarUtils.java
```diff
@@ -0,0 +1,28 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.beam.runners.fnexecution.jobsubmission;
+
+/** Contains common code for writing and reading portable pipeline jars. */
+public class PortablePipelineJarUtils {
```
angoenka: Lets make it abstract as it only hold constants.

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkJobInvoker.java
```diff
@@ -61,9 +62,20 @@ protected JobInvocation invokeWithExecutor(
     FlinkPipelineOptions flinkOptions =
         PipelineOptionsTranslation.fromProto(options).as(FlinkPipelineOptions.class);
 
+    PortablePipelineRunner pipelineRunner;
+    if (flinkOptions.getOutputJar() == null) {
```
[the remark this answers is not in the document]
ibzib: done

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkPipelineOptions.java
```diff
@@ -247,4 +247,9 @@
   Boolean isAutoBalanceWriteFilesShardingEnabled();
 
   void setAutoBalanceWriteFilesShardingEnabled(Boolean autoBalanceWriteFilesShardingEnabled);
+
+  @Description("The output path for the Jar file created.")
+  String getOutputJar();
```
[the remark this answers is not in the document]
ibzib: I went for `outputExecutablePath`, which is hopefully a little more descriptive.

runners/flink/src/test/java/org/apache/beam/runners/flink/FlinkSavepointTest.java
```diff
@@ -214,8 +213,7 @@ private JobID executePortable(Pipeline pipeline) throws Exception {
               executorService,
               pipelineProto,
               pipeline.getOptions().as(FlinkPipelineOptions.class),
-              null,
-              Collections.emptyList());
+              null);
```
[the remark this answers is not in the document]
ibzib: These other files were just changed because I removed an argument. The behavior should not be changed. Please take another look.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/PortablePipelineJarCreator.java
```diff
@@ -0,0 +1,289 @@
...
+  }
+
+  /** Copy resources from {@code classLoader} to {@code outputStream}. */
+  private static void writeClassPathResources(ClassLoader classLoader, JarOutputStream outputStream)
+      throws IOException {
+    List<String> classPathResources =
+        PipelineResources.detectClassPathResourcesToStage(classLoader);
+    Preconditions.checkArgument(
+        classPathResources.size() == 1, "Expected exactly one jar on " + classLoader.toString());
+    copyResourcesFromJar(new JarFile(classPathResources.get(0)), outputStream);
```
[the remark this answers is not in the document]
ibzib: `Preconditions.checkArgument` above throws `IllegalArgumentException` if there is ≠ 1 jar.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/PortablePipelineJarCreator.java
```diff
@@ -0,0 +1,289 @@
...
+    Enumeration<JarEntry> inputJarEntries = inputJar.entries();
+    // The zip spec allows multiple files with the same name; the Java zip libraries do not.
+    // Keep track of the files we've already written to filter out duplicates.
+    // Also, ignore the old manifest; we want to write our own.
+    Set<String> previousEntryNames = new HashSet<>(ImmutableList.of(JarFile.MANIFEST_NAME));
+    while (inputJarEntries.hasMoreElements()) {
+      JarEntry inputJarEntry = inputJarEntries.nextElement();
+      InputStream inputStream = inputJar.getInputStream(inputJarEntry);
+      String entryName = inputJarEntry.getName();
+      if (previousEntryNames.contains(entryName)) {
```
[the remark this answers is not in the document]
ibzib: Sure, good idea.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/PortablePipelineJarCreator.java
ibzib: ArtifactRetrievalServiceBlockingStub is a final class, and the version of Mockito we are using does not seem to support mocking final classes.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/PortablePipelineJarUtils.java
ibzib: done

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/PortablePipelineJarCreator.java
```diff
@@ -0,0 +1,289 @@
...
+    }
+  }
+
+  @VisibleForTesting
+  static Manifest createManifest(Class mainClass) {
+    Manifest manifest = new Manifest();
+    manifest.getMainAttributes().put(Attributes.Name.MANIFEST_VERSION, "1.0");
+    boolean classHasMainMethod = false;
+    try {
+      mainClass.getMethod("main", String[].class);
```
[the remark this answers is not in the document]
ibzib: done

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/PortablePipelineJarCreator.java
```diff
@@ -0,0 +1,289 @@
...
+      LOG.warn("No Main-Class will be set in jar because {} lacks a main method.", mainClass);
+    }
+    if (classHasMainMethod) {
+      manifest.getMainAttributes().put(Name.MAIN_CLASS, mainClass.getName());
+    }
+    return manifest;
+  }
+
+  /** Copy resources from {@code classLoader} to {@code outputStream}. */
+  private static void writeClassPathResources(ClassLoader classLoader, JarOutputStream outputStream)
```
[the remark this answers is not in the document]
ibzib: done

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/PortablePipelineJarCreator.java
ibzib: The JarArtifactStagingService approach is possible, but it would require a few non-trivial changes:

1. Abstract out the parts of [BeamFileSystemArtifactStagingService](https://github.com/apache/beam/blob/c2f0d282337f3ae0196a7717712396a5a41fdde1/runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/artifact/BeamFileSystemArtifactStagingService.java) that write the artifacts.
2. Create a new writer that writes to a JarOutputStream instead of a file system.
3. Find some way to change the output jar name. Right now, output paths are generated automatically, and the artifact API doesn't pass any options to make this configurable.

The way things are now, the temporarily staged artifacts are deleted when `--clean-artifacts-per-job` is enabled.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/PortablePipelineJarCreator.java
ibzib: In practice, arbitrary artifact names would already break the artifact system we have. For instance, the Python SDK [expects](https://github.com/apache/beam/blob/7ea4547a6f2530dffc06cb4dab3cdde58c2fd0dc/sdks/python/apache_beam/runners/worker/sdk_worker_main.py#L211-L212) a file named exactly `pickled_main_session`. This is provided by the Go bootstrap code, which writes [artifacts to files based on their name](https://github.com/apache/beam/blob/master/sdks/go/pkg/beam/artifact/materialize.go#L113).

ibzib:
Run Java PreCommit

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/PortablePipelineJarCreator.java
angoenka: The artifact retrieval service will do the name translation for us. 
So in practice, artifacts are stored with arbitrary name (art001) but at the time of retrieval on worker machine, they are converted back to their original name.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/PortablePipelineJarCreator.java
angoenka: Ya, it will add additional complexity. Lets leave it as it is and enable the `--clean-artifacts-per-job` flag only when creating the jar.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/PortablePipelineJarCreator.java
ibzib: I guess I should have been more specific about what I meant by arbitrary artifact names.

While places like https://issues.apache.org/jira/browse/BEAM-4109 seem to indicate we want to support *artifact names which are not valid file names*, places like [here](https://github.com/apache/beam/blob/master/sdks/go/pkg/beam/artifact/materialize.go#L113) indicate that artifact names actually do have to be valid file names. Also, I don't see the advantage of having artifact names be distinct from file names, and it just makes things more confusing.

In light of that, I want to keep the artifacts saved under their actual names wherever possible instead of gibberish UUIDs, unless there's a compelling reason not to.

ibzib:
Run Java PreCommit

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/PortablePipelineJarCreator.java
ibzib: I changed my mind and decided to use UUIDs after all.

tweise:
It would be good to capture the pipeline option modification at runtime in a follow-up JIRA.

Do you have an example how this jar producing runner would be execute from python side? I'm wondering how to integrate this into an automated build. An example command line that also shows how the runner/job-server jar dependency is specified would be helpful.

ibzib:
@tweise I created https://issues.apache.org/jira/browse/BEAM-8115 for overwriting pipeline options.
For an example usage, check out the integration test I wrote for the subsequent draft PR. https://github.com/apache/beam/blob/725116bfdcc7ff364456be016cb93dc7644ffb02/runners/flink/job-server/pipeline_jar_test.sh

tweise:
Thanks, I think it would be reasonable to merge this PR and then continue with any follow-up on https://github.com/apache/beam/pull/9408

Maybe squash and get tests to pass?

ibzib:
Squashed commits together. I think tests were failing from when master was broken.

ibzib:
Run Java PreCommit

tweise:
Something else: With Flink you can bundle multiple entry points into the same jar file and specify which one to use with optional flags. It may be desirable to allow inclusion of multiple pipelines for this tool also, although that would require a different workflow. Absent this option, it becomes quite convoluted for users that need the flexibility to choose which pipeline to launch at submission time.

ibzib:
> With Flink you can bundle multiple entry points into the same jar file

I think the most feasible way to do this would be to point multiple pipelines at the same output jar:

```py
pipeline1.run().wait_until_finish()
pipeline2.run().wait_until_finish()
```

Then the jar creation code would rebuild the jar, including the new pipeline, options, and artifacts. If we went that route, we would need to somehow differentiate adding to a jar vs overwriting it. Maybe by adding an `--input_executable_path` flag. I also don't know how we could ensure the pipelines were created by the same job server (if they weren't, there could be ugly version mismatches). However, I can't think of a way around this short of larger additions to the SDKs.

tweise:
Run Portable_Python PreCommit

tweise:
@ibzib An incremental mode sounds like a good solution. That could also be automatic if the output file exists. Perhaps it is better we take this to the mailing list and then create a JIRA for it?

PR #9336, opened 2019-08-14
[BEAM-7987] Drop empty Windmill workitem in WindowingWindmillReader

boyuanzz:
R: @lukecwik

### DIFF
--- a/runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/WindowingWindmillReader.java
+++ b/runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/WindowingWindmillReader.java
@@ -122,7 +122,7 @@ class WindowingWindmillReader<K, T> extends NativeReader<WindowedValue<KeyedWork
             && Iterables.isEmpty(keyedWorkItem.elementsIterable()));
     final WindowedValue<KeyedWorkItem<K, T>> value = new ValueInEmptyWindows<>(keyedWorkItem);
 
-    // Return a non-op iterator when current workitem is an empty workitem.
+    // Return a noop iterator when current workitem is an empty workitem.
     if (isEmptyWorkItem) {
       return new NativeReaderIterator<WindowedValue<KeyedWorkItem<K, T>>>() {
         @Override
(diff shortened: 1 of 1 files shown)


runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/WindowingWindmillReader.java
```diff
@@ -116,13 +117,19 @@
     final WorkItem workItem = context.getWork();
     KeyedWorkItem<K, T> keyedWorkItem =
         new WindmillKeyedWorkItem<>(key, workItem, windowCoder, windowsCoder, valueCoder);
+    final boolean isEmptyWorkItem =
```
lukecwik: It might be easier to understand the code if we return a new `NativeReaderIterator` that returns false for start/advance and throws NoSuchElementException instead of modifying the existing one.
```
if (isEmptyWorkItem) {
  return new NativeReaderIterator() { ... }
}
...
```

lukecwik:
Run Dataflow PostCommit

boyuanzz:
Run Dataflow PostCommit

boyuanzz:
Run Dataflow PostCommit

runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/WindowingWindmillReader.java
```diff
@@ -116,31 +117,54 @@
...
 
-    return new NativeReaderIterator<WindowedValue<KeyedWorkItem<K, T>>>() {
-      private WindowedValue<KeyedWorkItem<K, T>> current;
-
-      @Override
-      public boolean start() throws IOException {
-        current = value;
-        return true;
-      }
+    // Return a non-op iterator when current workitem is an empty workitem.
```
lukecwik:
```suggestion
    // Return a noop iterator when current workitem is an empty workitem.
```

lukecwik:
Run Java PreCommit

lukecwik:
Run Java PostCommit

boyuanzz:
Run Java PostCommit

runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/WindowingWindmillReader.java
boyuanzz: Done. Thanks : )

PR #9362, opened 2019-08-16
[BEAM-7994] Fixing unsafe pointer usage for Go 1.13

wcn3:
The Go team identified a problem in the Beam SDK that appears due to runtime changes in Go 1.13, which is upcoming. This is the backwards compatible fix the team recommended.

### DIFF
--- a/sdks/go/pkg/beam/core/util/reflectx/functions.go
+++ b/sdks/go/pkg/beam/core/util/reflectx/functions.go
@@ -40,6 +40,8 @@ func FunctionName(fn interface{}) string {
 // points to a valid function implementation.
 func LoadFunction(ptr uintptr, t reflect.Type) interface{} {
 	v := reflect.New(t).Elem()
-	*(*uintptr)(unsafe.Pointer(v.Addr().Pointer())) = (uintptr)(unsafe.Pointer(&ptr))
+	p := new(uintptr)
+	*p = ptr
+	*(*unsafe.Pointer)(unsafe.Pointer(v.Addr().Pointer())) = unsafe.Pointer(p)
 	return v.Interface()
 }
--- a/sdks/go/pkg/beam/core/util/reflectx/functions_test.go
+++ b/sdks/go/pkg/beam/core/util/reflectx/functions_test.go
@@ -1,0 +1,43 @@
+// Licensed to the Apache Software Foundation (ASF) under one or more
+// contributor license agreements.  See the NOTICE file distributed with
+// this work for additional information regarding copyright ownership.
+// The ASF licenses this file to You under the Apache License, Version 2.0
+// (the "License"); you may not use this file except in compliance with
+// the License.  You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package reflectx
+
+import (
+	"reflect"
+	"testing"
+)
+
+func testFunction() int64 {
+	return 42
+}
+
+func TestLoadFunction(t *testing.T) {
+	val := reflect.ValueOf(testFunction)
+	fi := uintptr(val.Pointer())
+	typ := val.Type()
+
+	callable := LoadFunction(fi, typ)
+
+	cv := reflect.ValueOf(callable)
+	out := cv.Call(nil)
+	if len(out) != 1 {
+		t.Errorf("got %d return values, wanted 1.", len(out))
+	}
+
+	if out[0].Int() != testFunction() {
+		t.Errorf("got %d, wanted %d", out[0].Int(), testFunction())
+	}
+}


wcn3:
R: @lostluck @lukecwik

sdks/go/pkg/beam/core/util/reflectx/functions_test.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"reflect"
+	"testing"
+)
+
+func testFunction() int {
+	return 42
+}
+
+func TestXxx(t *testing.T) {
```
dsnet: `TestLoadFunction`?

dsnet:
More background can be found here: golang/go#33662

sdks/go/pkg/beam/core/util/reflectx/functions.go
```diff
@@ -40,6 +40,8 @@ func FunctionName(fn interface{}) string {
 // points to a valid function implementation.
 func LoadFunction(ptr uintptr, t reflect.Type) interface{} {
 	v := reflect.New(t).Elem()
-	*(*uintptr)(unsafe.Pointer(v.Addr().Pointer())) = (uintptr)(unsafe.Pointer(&ptr))
+	p := new(uintptr)
+	*p = ptr
+	*(*unsafe.Pointer)(unsafe.Pointer(v.Addr().Pointer())) = unsafe.Pointer(p)
```
youngoli: I had to read up a bit on how reflect and unsafe work to understand what was happening here, but I still don't understand exactly why the previous version was invalid and this one valid. (I'm using https://golang.org/pkg/unsafe/#Pointer as a guideline).

Is it because in the old version we simply put in the address to `ptr`, and in the new one we actually create a new copy?

Or is it because in the old version we're reassigning to `v` as a `uintptr` instead of as an `unsafe.Pointer`?

Those are the only real differences I could figure out.

sdks/go/pkg/beam/core/util/reflectx/functions_test.go
youngoli: Yeah, the name here seems wrong.

sdks/go/pkg/beam/core/util/reflectx/functions_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+	cv := reflect.ValueOf(callable)
+	out := cv.Call(nil)
+	if len(out) != 1 {
+		t.Errorf("got %d return values, wanted 1.", len(out))
+	}
+	// TODO: check type?
+	if out[0].Int() != 42 {
+		t.Errorf("got %d, wanted 42", out[0].Int())
+	}
```
youngoli: Nit: Instead of comparing to 42 directly, could you do something like:

```
expected := testFunction()
if out[0].Int() != expected {
	t.Errorf("got %d, wanted %d", out[0].Int(), expected)
}
```

sdks/go/pkg/beam/core/util/reflectx/functions_test.go
youngoli: +1, name seems off.

wcn3: I thought the last commit fixed this, but clearly not. Will fix.

sdks/go/pkg/beam/core/util/reflectx/functions.go
wcn3: The problem was that the original code relies on the behavior that when &ptr is evaluated, the resulting anonymous value will be on the heap. The new compiler is able to not put it on the heap, and does so, which causes this code to crash in Go 1.13.

This change makes the expected behavior explicit by allocating p on the heap and then assigning it's pointed-to value to ptr, thus p == &ptr and the remaining code works as before.

The unit test exercises this function. I verified that the old code crashed under Go 1.13 (worked in Go 1.12) and the new code works in both versions.

sdks/go/pkg/beam/core/util/reflectx/functions_test.go
wcn3: Good point, will change.

sdks/go/pkg/beam/core/util/reflectx/functions.go
youngoli: Thanks for the nice summary!

lukecwik:
Similar to how we are testing different versions of Python 3.5, 3.6, 3.7 with the Beam SDK, is there a way to test different Go versions against the Beam SDK to catch these kinds of things?

wcn3:
This was a problem revealed in a beta test in an upcoming version of the Go
SDK. I am not sure Beam wants to set up test coverage for beta SDKs. In
general, Go has good backwards compatibility guarantees so testing version
compatibility of older SDKs hasn’t been needed.

It is entirely possible to set up the sort of testing you describe, but it
would be of limited use, particularly with regards to this issue. In this
case, I believe the Go community process worked as intended: as a precursor
to the Go team releasing betas, they check it against the Google code
corpus. Since Beam is used by Google, Beam gets the benefit of this testing
coverage. This PR came about precisely because of those tests.

sdks/go/pkg/beam/core/util/reflectx/functions_test.go
lostluck: It's a bit more verbose but I've liked the following style since it avoids comparing then printing the wrong things accidentally.

if got,want := out[0].Int(), testFunction(); got != want {
  t.Errorf("got %d, wanted %d", got, want)
}

lostluck:
I have no additional comments. The discussion here was excellent, and I'm glad we'll be able to transition to go1.13 relatively easily. (Beam's Go Module status not withstanding...)

sdks/go/pkg/beam/core/util/reflectx/functions_test.go
wcn3: I like this style, since the comma operator helps keep it compact.

Joe made an offline comment that every usage of the unsafe package should be accompanied by a unit test. I agree with that, and I'm looking at the remaining unsafe usages in the SDK right now. I'll make this tweak in that followup PR.

dsnet: Yes, please! :)

The lack of a unit test on this function meant that it took several days to track down the issue when there was mysterious memory corruption happening in the entire beam job. A single unit test would have immediately identified this as the culprit.

lostluck:
IIRC there should only be one other use of Unsafe and that's an optmization
in the ioutilx package so a []byte we know should be allocated on the stack
isn't put on the heap. (Noescape essentially).
The only other workaround would be to plumb a concrete type all the way to
that spot instead of an io.Reader.

PR #9366, opened 2019-08-16
[BEAM-8143] Build simple LineSource directly on top of SDF

boyuanzz:
R: @robertwb

### DIFF
--- a/sdks/python/apache_beam/io/sdf_line_source.py
+++ b/sdks/python/apache_beam/io/sdf_line_source.py
@@ -1,0 +1,88 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+from __future__ import absolute_import
+
+import os
+
+from apache_beam import Create
+from apache_beam import pvalue
+from apache_beam.transforms import core
+from apache_beam.transforms import ptransform
+from apache_beam.io.restriction_trackers import OffsetRestrictionTracker
+from apache_beam.io.restriction_trackers import OffsetRange
+
+class SdfReadLineRestrictionProvider(core.RestrictionProvider):
+  """ A `core.RestrictionProvider` implementation which provides
+  `OffsetRestrictionTracker`."""
+  def _get_file_size(self, file_name):
+    with open(file_name, 'rb') as f:
+      f.seek(0, os.SEEK_END)
+      return f.tell()
+
+  def initial_restriction(self, element):
+    return OffsetRange(0, self._get_file_size(element['file_pattern']))
+
+  def create_tracker(self, restriction):
+    return OffsetRestrictionTracker(restriction)
+
+  def split(self, element, restriction):
+    return restriction.split(element['desired_chunk_size'])
+
+  def restriction_size(self, element, restriction):
+    return restriction.size()
+
+
+class SdfReadLine(core.DoFn):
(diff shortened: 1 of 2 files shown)


sdks/python/apache_beam/io/sdf_line_source.py
```diff
@@ -0,0 +1,89 @@
...
+  """ An SplittableDoFn which reads lines from file and as output."""
+  def process(
+      self,
+      element,
+      restriction_tracker=core.DoFn.RestrictionParam(
+          SdfReadLineRestrictionProvider())):
+    with open(element['file_pattern'], 'rb') as f:
+      start = restriction_tracker.start_position()
+      f.seek(start)
+      if start > 0:
```
robertwb: This probably deserves some comments. (I assume you're trying to see if we're starting at a line? What about multi-char newlines (e.g. \r\f)?

sdks/python/apache_beam/io/sdf_line_source.py
```diff
@@ -0,0 +1,89 @@
...
+        f.seek(-1, os.SEEK_CUR)
+        start -= 1
+        start += len(f.readline())
```
robertwb: start = f.tell()?

sdks/python/apache_beam/io/sdf_line_source.py
```diff
@@ -0,0 +1,89 @@
...
+      current = start
+      line = f.readline()
+      while restriction_tracker.try_claim(current):
+        if not line:
+          return
+        yield line.rstrip(b'\n')
+        current += len(line)
```
robertwb: Won't this get off by one due to stripping the newline before adding the value to current? Possibly current = f.tell() would be more consistent.

sdks/python/apache_beam/io/sdf_line_source.py
```diff
@@ -0,0 +1,89 @@
...
+          SdfReadLineRestrictionProvider())):
+    with open(element['file_pattern'], 'rb') as f:
+      start = restriction_tracker.start_position()
+      f.seek(start)
+      if start > 0:
+        f.seek(-1, os.SEEK_CUR)
+        start -= 1
+        start += len(f.readline())
+      current = start
+      line = f.readline()
```
robertwb: I would put this line inside the loop, rather than right before it and at the end of the loop.

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
```diff
@@ -1569,6 +1570,22 @@ def test_split_half(self):
     raise unittest.SkipTest("This test is for a single worker only.")
 
 
+class FnApiRunnerSdfSourceTest(unittest.TestCase):
```
robertwb: Should this be in apache_beam/io/sdf_line_source_test.py?

sdks/python/apache_beam/io/sdf_line_source.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+class SdfReadLine(core.DoFn):
+  """ An SplittableDoFn which reads lines from file and as output."""
+  def process(
+      self,
+      element,
+      restriction_tracker=core.DoFn.RestrictionParam(
+          SdfReadLineRestrictionProvider())):
+    with open(element['file_pattern'], 'rb') as f:
```
robertwb: Should we be using Beam's filesystem here?

sdks/python/apache_beam/io/sdf_line_source.py
```diff
@@ -0,0 +1,89 @@
...
+    return OffsetRestrictionTracker(restriction)
+
+  def split(self, element, restriction):
+    return restriction.split(element['desired_chunk_size'])
+
+  def restriction_size(self, element, restriction):
+    return restriction.size()
+
+
+class SdfReadLine(core.DoFn):
```
robertwb: ReadLines (plural) would be more natural.

sdks/python/apache_beam/io/sdf_line_source.py
```diff
@@ -0,0 +1,89 @@
...
+from apache_beam.io.restriction_trackers import OffsetRestrictionTracker
+from apache_beam.transforms import core
+from apache_beam.transforms import ptransform
+
+
+class SdfReadLineRestrictionProvider(core.RestrictionProvider):
+  """ A `core.RestrictionProvider` implementation which provides
+  `OffsetRestrictionTracker`."""
+  def _get_file_size(self, file_name):
+    with open(file_name, 'rb') as f:
```
robertwb: There are generally more efficient ways to get filesize than opening it up.

PR #9368, opened 2019-08-17
[BEAM-7999] Fix BigQuery timestamp handling for Schema Aware PCollection

alexvanboxel:
This fixes the handling of timestamps when reading for BigQuery. As it
was assumed BigQuery always outputs a Double, reading a table directly
outputs a textual version in the JSON. This is now handled. Also the
from Beam Row also outputs the BigQuery textual version, iso the ISO
standard representation, this makes the implementation symmetric.

_Tested on DirectRunner and DataflowRunner_


alexvanboxel:
Abandoning... as rebase seems to be incorrect

PR #8620, opened 2019-05-20
[BEAM-6673] Add schema support to BigQuery reads

charithe:
Adds support for Beam schemas in BigQuery reads.

Direct reads are not supported yet. Support for different date/time types and RECORD types are pending as well.

The largest change is the refactoring of `BigQueryTableSource` and `BigQueryQuerySource` classes. As the schema inference has to happen at pipeline construction time, the reusable code for accessing the table/query results have been extracted out to "source definition" classes named `BigQueryTableSourceDef` and `BigQueryQuerySourceDef`. 

------------------------

Thank you for your contribution! Follow this checklist to help us incorporate your contribution quickly and easily:

 - [ ] [**Choose reviewer(s)**](https://beam.apache.org/contribute/#make-your-change) and mention them in a comment (`R: @username`).
 - [ ] Format the pull request title like `[BEAM-XXX] Fixes bug in ApproximateQuantiles`, where you replace `BEAM-XXX` with the appropriate JIRA issue, if applicable. This will automatically link the pull request to the issue.
 - [ ] If this contribution is large, please file an Apache [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

Lang | SDK | Apex | Dataflow | Flink | Gearpump | Samza | Spark
--- | --- | --- | --- | --- | --- | --- | ---
Go | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go/lastCompletedBuild/) | --- | --- | --- | --- | --- | ---
[...]

PR #9369, opened 2019-08-17
[BEAM-7999] Fix BigQuery timestamp handling for Schema Aware PCollection

alexvanboxel:
This fixes the handling of timestamps when reading for BigQuery. As it
was assumed BigQuery always outputs a Double, reading a table directly
outputs a textual version in the JSON. This is now handled. Also the
from Beam Row also outputs the BigQuery textual version, iso the ISO
standard representation, this makes the implementation symmetric.

_Tested on DirectRunner and DataflowRunner_


alexvanboxel:
Inviting @charithe and @reuvenlax for this fix as they were the creators/reviewers of the original implementation (see https://github.com/apache/beam/pull/8620) . As soon as I tried to use the BigQuery schema aware read I ran into problems when running against the **DirectRunner**. As the DirectRunner force the serialization steps to kick in the problem described in the ticket turned up. This implementation fixes this.

reuvenlax:
this looks good, but we should run SQL BigQuery tests as well. @apilloud  what is the incantation to run all the SQL tests?

alexvanboxel:
ping @apilloud could be cool, if it gets verified so it's part of the next release.

PR #9370, opened 2019-08-17
[BEAM-7882] Invoke Spark API incompatible methods by reflection

iemejia:
It adds also Spark 3 specific version of the methods

R: @RyanSkraba


iemejia:
Run Spark ValidatesRunner

iemejia:

iemejia:
Thanks @RyanSkraba  Yes it seems like the easiest maintainable path. I hesitated between two modules but the changes are relatively small and they impact just a tiny part of the codebase to justify the effort of having separate modules +implementations. Of course we can re evaluate this in the future, the path towards two modules or to let support only for Spark 3 in the future is super straightforward because of this class.

PR #9372, opened 2019-08-19
[Beam-6858] Validate that side-input parameters match the type of the PCollectionView

reuvenlax:
Validate that the type of a processElement side-input parameter matches that of the PCollectionView.

Note: Suppliers are used because often the TypeDescriptor is not Serializable, and so cannot be captured in the PCollectionView.

### DIFF
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/ParDo.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/ParDo.java
@@ -53,6 +53,7 @@ import org.apache.beam.sdk.transforms.reflect.DoFnSignature.FieldAccessDeclarati
 import org.apache.beam.sdk.transforms.reflect.DoFnSignature.MethodWithExtraParameters;
 import org.apache.beam.sdk.transforms.reflect.DoFnSignature.OnTimerMethod;
 import org.apache.beam.sdk.transforms.reflect.DoFnSignature.Parameter.SchemaElementParameter;
+import org.apache.beam.sdk.transforms.reflect.DoFnSignature.Parameter.SideInputParameter;
 import org.apache.beam.sdk.transforms.reflect.DoFnSignatures;
 import org.apache.beam.sdk.transforms.windowing.BoundedWindow;
 import org.apache.beam.sdk.transforms.windowing.WindowFn;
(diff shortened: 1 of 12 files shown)


runners/direct-java/src/main/java/org/apache/beam/runners/direct/ParDoEvaluatorFactory.java
```diff
@@ -107,7 +108,8 @@ public void cleanup() throws Exception {
       List<PCollectionView<?>> sideInputs,
       TupleTag<OutputT> mainOutputTag,
       List<TupleTag<?>> additionalOutputTags,
-      DoFnSchemaInformation doFnSchemaInformation)
+      DoFnSchemaInformation doFnSchemaInformation,
+      Map<String, String> sideInputMapping)
```
salmanVD: @reuvenlax This should be String, String? or PCollectionView

reuvenlax:
Yes, we should change to Map<String, PCollectionView<?>>

reuvenlax:
R: @TheNeuralBit

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/ParDo.java
```diff
@@ -437,6 +438,29 @@ private static void validateStateApplicableForInput(DoFn<?, ?> fn, PCollection<?
...
+    for (SideInputParameter sideInput : processElementMethod.getSideInputParameters()) {
+      PCollectionView<?> view = sideInputs.get(sideInput.sideInputId());
+      checkArgument(
+          view != null,
+          "the ProcessElement method expects a side input identified with the tag %s, but no such side input was"
+              + " supplied. Use withSideInput(String, PCollectionView) to supply this side input.",
+          sideInput.sideInputId());
+      TypeDescriptor<?> viewType = view.getViewFn().getTypeDescriptor();
+
+      // Currently check that the types exactly match, even if the types are convertible.
```
TheNeuralBit: Are there plans to handle convertible types? Should there be a JIRA thats linked here?

sdks/java/core/src/test/java/org/apache/beam/sdk/transforms/ParDoTest.java
```diff
@@ -737,6 +737,76 @@ public void testParDoWithSideInputs() {
...
+          };
+
+      thrown.expect(IllegalArgumentException.class);
+      PCollection<List<Integer>> output =
+          pipeline
+              .apply("Create main input", Create.of(2))
+              .apply(ParDo.of(fn).withSideInput(sideInputTag1, sideInput1));
+      pipeline.run();
+    }
+
```
TheNeuralBit: Would you mind adding tests for the remaining view types - Map, Multimap, Iterable?

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/ParDo.java
reuvenlax: We have the same issue with Element parameters. I'll check to see if we have a JIRA for this (I think a user actually filed one some time back).

sdks/java/core/src/test/java/org/apache/beam/sdk/transforms/ParDoTest.java
reuvenlax: tests added

reuvenlax:
all comments addressed

PR #9373, opened 2019-08-19
[BEAM-7264] Remove set/getClientConfiguration in JdbcIO

iemejia:
R: @aromanenko-dev

### DIFF
--- a/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcIO.java
+++ b/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcIO.java
@@ -531,11 +531,6 @@ public class JdbcIO {
   /** Implementation of {@link #read}. */
   @AutoValue
   public abstract static class Read<T> extends PTransform<PBegin, PCollection<T>> {
-    /** @deprecated It is not needed anymore. It will be removed in a future version of Beam. */
-    @Deprecated
-    @Nullable
-    abstract DataSourceConfiguration getDataSourceConfiguration();
-
     @Nullable
     abstract SerializableFunction<Void, DataSource> getDataSourceProviderFn();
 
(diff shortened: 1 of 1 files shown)


iemejia:
`withDataSourceConfiguration` is the previous (and still) more used API, it allows the users to pass a configuration object that it is converted into the request internally by the more generic `withDataSourceProviderFn`. I am a bit hesitant to deprecate it (or remove it) due to its expanded use and also because it is more user friendly.
However you may have a point, we should consider as an option to maybe deprecate it and suggest to the users to use the other more explicit transform.

iemejia:
Thanks for the review @aromanenko-dev !

aromanenko-dev:
@iemejia I don't propose to deprecate and remove `withDataSourceConfiguration` since, yes, it should be heavily used by users. My point was a bit different, let me explain in more details. 

When one calls `withDataSourceConfiguration` then `withDataSourceProviderFn` will be called implicitly and new instance of data source provider will be created there. On the hand, if `withDataSourceProviderFn` is called explicitly in the same transform, then another data source provider will be created and the order of creating them is undetermined. It looks contradictory  and not obvious for users. So, to avoid such potential ambiguity, we need to not allow to call these two methods in the same transform. WDYT?

iemejia:
Ok I see what you mean now (thanks for the clarification), If I understood correctly, what you mean is to make the methods exclusive. If so, yes, it could be good for user friendliness. Only thing that it is not clear to me is `the order of creating them is undetermined` I think it is deterministic to the latest method call the user does of the two (internal attribute will be overwritten).

I am not 100% sure if the improvement is worth, currently users can also call `withDataSourceConfiguration` with some configuration and then `withDataSourceConfiguration` with a different configuration and we do not validate this. I am weary of the extra complexity for not much in return and the fact of differing of how other IOs with similar APIs work.

aromanenko-dev:
Well, maybe "undetermined" was not a right word for this - I meant that the creating of final "working" instance of data source provider depends on order of calling build methods and it can be different depending on user's code. So, for the same pipeline but with different order of these methods calling, the results could be different too. 

I agree that we don't check calling of the same methods with different arguments (which is quite logical in general case) but I'd argue that we can avoid potential ambiguity in such specific case (since `withDataSourceConfiguration()` calls `withDataSourceProviderFn()` implicitly) by adding the check I proposed above. For example, we have a similar check in `KinesisIO`: 
```
checkArgument(
    getPartitionKey() == null || (getPartitioner() == null),
    "only one of either withPartitionKey() or withPartitioner() is possible");
```
I don't think it will complicate `JdbcIO` API too much. We can have a different Jira for that.

iemejia:
Filled https://issues.apache.org/jira/browse/BEAM-8033 to track this. Thanks @aromanenko-dev !

PR #9374, opened 2019-08-19
[BEAM-5428] Implement Runner support for cache tokens

mxm:
This adds support for sending cache tokens for state requests from the
SDK. Cache tokens are managed by the StateRequestHandler implemented by the
Runner.

User state cache tokens are valid for the lifetime of a Runner<=>SDK Harness
connection. If the application fails for any reason, (e.g. error in user code,
failed checkpointing, etc.) the cache token must be renewed.

--- | --- | --- | --- | ---

### DIFF
--- a/runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
+++ b/runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
@@ -31,6 +31,8 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Optional;
+import java.util.UUID;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
@@ -71,6 +73,7 @@ import org.apache.beam.runners.fnexecution.state.StateRequestHandler;
 import org.apache.beam.runners.fnexecution.state.StateRequestHandlers;
 import org.apache.beam.sdk.coders.Coder;
 import org.apache.beam.sdk.coders.VoidCoder;
+import org.apache.beam.sdk.fn.IdGenerator;
 import org.apache.beam.sdk.fn.data.FnDataReceiver;
 import org.apache.beam.sdk.options.PipelineOptions;
 import org.apache.beam.sdk.state.BagState;
@@ -86,6 +89,7 @@ import org.apache.beam.sdk.values.PCollectionView;
 import org.apache.beam.sdk.values.TupleTag;
 import org.apache.beam.sdk.values.WindowingStrategy;
 import org.apache.beam.vendor.grpc.v1p21p0.com.google.protobuf.ByteString;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Charsets;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Preconditions;
 import org.apache.beam.vendor.sdk.v2.sdk.extensions.protobuf.ByteStringCoder;
 import org.apache.flink.api.common.state.ListStateDescriptor;
@@ -238,7 +242,10 @@ public class ExecutableStageDoFnOperator<InputT, OutputT> extends DoFnOperator<I
           StateRequestHandlers.forBagUserStateHandlerFactory(
               stageBundleFactory.getProcessBundleDescriptor(),
               new BagUserStateFactory(
-                  keyedStateInternals, getKeyedStateBackend(), stateBackendLock));
+                  () -> UUID.randomUUID().toString(),
+                  keyedStateInternals,
+                  getKeyedStateBackend(),
+                  stateBackendLock));
     } else {
       userStateRequestHandler = StateRequestHandler.unsupported();
     }
@@ -250,30 +257,37 @@ public class ExecutableStageDoFnOperator<InputT, OutputT> extends DoFnOperator<I
     return StateRequestHandlers.delegateBasedUponType(handlerMap);
   }
 
-  private static class BagUserStateFactory<K extends ByteString, V, W extends BoundedWindow>
+  static class BagUserStateFactory<K extends ByteString, V, W extends BoundedWindow>
       implements StateRequestHandlers.BagUserStateHandlerFactory<K, V, W> {
 
     private final StateInternals stateInternals;
     private final KeyedStateBackend<ByteBuffer> keyedStateBackend;
     private final Lock stateBackendLock;
+    /** Holds the valid cache token for user state for this operator. */
+    private final ByteString cacheToken;
 
-    private BagUserStateFactory(
+    BagUserStateFactory(
+        IdGenerator cacheTokenGenerator,
         StateInternals stateInternals,
         KeyedStateBackend<ByteBuffer> keyedStateBackend,
         Lock stateBackendLock) {
       this.stateInternals = stateInternals;
       this.keyedStateBackend = keyedStateBackend;
       this.stateBackendLock = stateBackendLock;
+      this.cacheToken = ByteString.copyFrom(cacheTokenGenerator.getId().getBytes(Charsets.UTF_8));
     }
 
     @Override
     public StateRequestHandlers.BagUserStateHandler<K, V, W> forUserState(
+        // Transform id not used because multiple operators with state will not
+        // be fused together. See GreedyPCollectionFusers
         String pTransformId,
         String userStateId,
         Coder<K> keyCoder,
         Coder<V> valueCoder,
         Coder<W> windowCoder) {
       return new StateRequestHandlers.BagUserStateHandler<K, V, W>() {
+
         @Override
         public Iterable<V> get(K key, W window) {
           try {
@@ -290,6 +304,7 @@ public class ExecutableStageDoFnOperator<InputT, OutputT> extends DoFnOperator<I
             }
             BagState<V> bagState =
                 stateInternals.state(namespace, StateTags.bag(userStateId, valueCoder));
+
             return bagState.read();
           } finally {
             stateBackendLock.unlock();
@@ -340,6 +355,12 @@ public class ExecutableStageDoFnOperator<InputT, OutputT> extends DoFnOperator<I
           } finally {
             stateBackendLock.unlock();
           }
+        }
+
+        @Override
+        public Optional<ByteString> getCacheToken() {
+          // Cache tokens remains valid for the life time of the operator
+          return Optional.of(cacheToken);
         }
 
         private void prepareStateBackend(K key) {
--- a/runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/state/InMemoryBagUserStateFactory.java
+++ b/runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/state/InMemoryBagUserStateFactory.java
@@ -20,6 +20,8 @@ package org.apache.beam.runners.fnexecution.state;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Optional;
+import java.util.UUID;
 import org.apache.beam.runners.core.InMemoryStateInternals;
 import org.apache.beam.runners.core.StateInternals;
 import org.apache.beam.runners.core.StateNamespace;
@@ -29,6 +31,8 @@ import org.apache.beam.runners.core.StateTags;
 import org.apache.beam.sdk.coders.Coder;
 import org.apache.beam.sdk.state.BagState;
 import org.apache.beam.sdk.transforms.windowing.BoundedWindow;
+import org.apache.beam.vendor.grpc.v1p21p0.com.google.protobuf.ByteString;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Charsets;
 
 /**
  * Holds user state in memory. Only one key is active at a time due to the GroupReduceFunction being
@@ -70,6 +74,7 @@ public class InMemoryBagUserStateFactory<K, V, W extends BoundedWindow>
 
     private final StateTag<BagState<V>> stateTag;
     private final Coder<W> windowCoder;
+    private final ByteString cacheToken;
 
     /* Lazily initialized state internals upon first access */
     private volatile StateInternals stateInternals;
@@ -77,6 +82,7 @@ public class InMemoryBagUserStateFactory<K, V, W extends BoundedWindow>
     InMemorySingleKeyBagState(String userStateId, Coder<V> valueCoder, Coder<W> windowCoder) {
       this.windowCoder = windowCoder;
       this.stateTag = StateTags.bag(userStateId, valueCoder);
+      this.cacheToken = ByteString.copyFrom(UUID.randomUUID().toString().getBytes(Charsets.UTF_8));
     }
 
     @Override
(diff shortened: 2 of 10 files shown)


mxm:
Requesting a review from @tweise but also feel free to review @rakeshcusat.

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
```diff
@@ -248,33 +253,45 @@ private StateRequestHandler getStateRequestHandler(ExecutableStage executableSta
     return StateRequestHandlers.delegateBasedUponType(handlerMap);
   }
 
-  private static class BagUserStateFactory<K extends ByteString, V, W extends BoundedWindow>
+  static class BagUserStateFactory<K extends ByteString, V, W extends BoundedWindow>
       implements StateRequestHandlers.BagUserStateHandlerFactory<K, V, W> {
 
+    /** Upper limit of the number of valid cache tokens to hand out to the SDK. */
+    private static final int MAX_CACHE_SIZE = 100;
```
rakeshcusat: I think it is good for the initial implementation. Do we want to make it configuration based later on?

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
```diff
@@ -248,33 +253,45 @@ private StateRequestHandler getStateRequestHandler(ExecutableStage executableSta
...
+
     private final StateInternals stateInternals;
     private final KeyedStateBackend<ByteBuffer> keyedStateBackend;
     private final Lock stateBackendLock;
+    /** Cache is scoped by state id. */
```
rakeshcusat: is it possible that there are two DoFns in the pipeline and they accidentally using the same state id? In this case there could be collision of stateid. I think we also need other parameters in the cache key.

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
```diff
@@ -248,33 +253,45 @@ private StateRequestHandler getStateRequestHandler(ExecutableStage executableSta
...
+    BagUserStateFactory(
         StateInternals stateInternals,
         KeyedStateBackend<ByteBuffer> keyedStateBackend,
         Lock stateBackendLock) {
 
       this.stateInternals = stateInternals;
       this.keyedStateBackend = keyedStateBackend;
       this.stateBackendLock = stateBackendLock;
+      this.cacheTokens = CacheBuilder.newBuilder().maximumSize(MAX_CACHE_SIZE).build();
+      this.cacheTokenGenerator = IdGenerators.incrementingLongs();
```
rakeshcusat: I am assuming this will be used later part of the implementation.

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
```diff
@@ -336,11 +356,29 @@ public void clear(K key, W window) {
...
+        }
+
+        @Override
+        public void clearCacheTokens() {
+          cacheTokens.invalidateAll();
+        }
+
+        private ByteString generateAndRegisterCacheKey() {
+          ByteString cacheToken = ByteString.copyFrom(cacheTokenGenerator.getId(), Charsets.UTF_8);
+          cacheTokens.put(userStateId, cacheToken);
```
rakeshcusat: Correct me if I am wrong there. The bag states are also separated by window, would it make sense to include window in the cache key?

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/state/InMemoryBagUserStateFactory.java
```diff
@@ -80,29 +82,33 @@ public void resetForNewKey() {
...
-    public void append(K key, W window, Iterator<V> values) {
+    @Nullable
+    public ByteString append(K key, W window, Iterator<V> values) {
       initStateInternals(key);
       StateNamespace namespace = StateNamespaces.window(windowCoder, window);
       BagState<V> bagState = stateInternals.state(namespace, stateTag);
       while (values.hasNext()) {
         bagState.add(values.next());
       }
+      return null;
```
rakeshcusat: just for consistency I would have similar cache token implementation for this class as well.

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
```diff
@@ -314,13 +332,15 @@ public void append(K key, W window, Iterator<V> values) {
             while (values.hasNext()) {
               bagState.add(values.next());
             }
+
+            return generateAndRegisterCacheKey();
```
rakeshcusat: In some scenarios runner wants sdk worker to not cache the value. It indicates it by not passing any cache token. How are we addressing that scenario?

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
mxm: That is not possible. GreedyPipelineFuser will not fuse two ParDos if they both access user state.

>I think we also need other parameters in the cache key.

Which parameters do we need?

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
mxm: I cannot think of any scenarios for user state where we wouldn't want to cache. It is important though that the cache tokens can be cleared such that they are not sent to the SDK anymore, e.g. when restoring from a checkpoint.

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
mxm: The same cache token can be handed out for all state cells across all windows. On the SDK side we need to scope the cache per window because state needs to be retrieved at least once per window.

On this note, I will also remove the cache token per user state because we just need one cache token for the key range (i.e. operator instance) for all user state. The reason is that user state does not get invalidated for other reasons than an updated key range. For side input, each side input needs a separate cache token to be able to invalidate them individually in case they get updated.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/state/InMemoryBagUserStateFactory.java
mxm: +1

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
mxm: Actually, this will become obsolete if we start to maintain only a single cache token for all user state and one for each side input.

mxm:
PR updated with the changes in #9440.

mxm:
Master is currently broken: https://builds.apache.org/job/beam_PreCommit_Java_Commit/7505/

mxm:
Run Java PreCommit

mxm:

mxm:
Issues with the master are resolved.

mxm:
Run Java PreCommit

mxm:
Run Java PreCommit

mxm:
Run Java PreCommit

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
```diff
@@ -248,31 +250,37 @@ private StateRequestHandler getStateRequestHandler(ExecutableStage executableSta
...
-    private BagUserStateFactory(
+    BagUserStateFactory(
         StateInternals stateInternals,
         KeyedStateBackend<ByteBuffer> keyedStateBackend,
         Lock stateBackendLock) {
 
       this.stateInternals = stateInternals;
       this.keyedStateBackend = keyedStateBackend;
       this.stateBackendLock = stateBackendLock;
+      this.cacheToken = ByteString.copyFrom(UUID.randomUUID().toString(), Charsets.UTF_8);
```
lukecwik: nit: Please take in an `IdGenerator` as a parameter since it will allow one to explicitly test knowing what the cache token value(s) will be. Also consider using `IdGenerators#incrementingLongs`

mxm:
>How do you imagine cache invalidation would happen (I'm asking because I wonder whether we should be exposing invalidateCacheToken APIs as part of the shared runner fn-execution code)?

In my mind, the Runner simply redeploys the operator in case of a failure and that discards all its ephemeral state (e.g. cache tokens). Only state persisted in the Runner's state backend will be restored after a failure. I know this model works for Flink and Spark. If such a model works for all Runners, then we could indeed remove the `invalidateCacheToken()` method from the user state handler.

So I'll remove the invalidation logic from `BagUserStateHandler`. Are you also suggesting, we should remove it from `StateRequestHandler` as well? Then nothing would be exposed in fn-execution. All the invalidation logic would be inside the state handler implementations.

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
mxm: I had the assumption that the cache token would be globally unique. That avoids the cleanup scenario when dealing with a long-running SDK Harness. You are right though that testing/debugging is much easier with increasing integers. Have made the cache token generation configurable via `IdGenerator`.

mxm:
Run Java PreCommit

mxm:
Run Java PreCommit

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/state/StateRequestHandlers.java
```diff
@@ -278,6 +291,11 @@ public static StateRequestHandler forSideInputHandlerFactory(
       }
     }
 
+    @Override
+    public Iterable<BeamFnApi.ProcessBundleRequest.CacheToken> getCacheTokens() {
```
lukecwik: I don't think we need this since this is the default implementation?

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/state/StateRequestHandlers.java
```diff
@@ -137,6 +143,12 @@ static SideInputHandlerFactory unsupported() {
 
     /** Clears the bag user state for the given key and window. */
     void clear(K key, W window);
+
+    /** Returns the currently valid cache token. */
```
lukecwik:
```suggestion
    /** Returns the currently valid cache token. There is no valid cache token if null is returned. */
```

lukecwik: nit: We might want to go with an Iterable here or Optional instead of using null to mark missing cache tokens.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/state/StateRequestHandlers.java
```diff
@@ -414,6 +432,21 @@ public static StateRequestHandler forBagUserStateHandlerFactory(
       }
     }
 
+    @Override
+    public Iterable<BeamFnApi.ProcessBundleRequest.CacheToken> getCacheTokens() {
+      return handlerCache.values().stream()
```
lukecwik: nit: [streams are siginficantly slower than for loops even with JDK 12](https://medium.com/@milan.mimica/slow-like-a-stream-fast-like-a-loop-524f70391182)

runners/java-fn-execution/src/test/java/org/apache/beam/runners/fnexecution/control/SdkHarnessClientTest.java
```diff
@@ -574,6 +579,41 @@ public void handleCleanupWithStateWhenAwaitingOnClosingOutputReceivers() throws
...
+            BeamFnApi.ProcessBundleRequest.CacheToken.newBuilder().getDefaultInstanceForType());
+    when(stateRequestHandler.getCacheTokens()).thenReturn(cacheTokens);
+
+    processor1.newBundle(
+        ImmutableMap.of(SDK_GRPC_WRITE_TRANSFORM, mock(RemoteOutputReceiver.class)),
+        stateRequestHandler,
+        BundleProgressHandler.ignored());
+
+    // Verify that the cache token list is retrieved from the state handler
+    Mockito.verify(stateRequestHandler).getCacheTokens();
```
lukecwik: We should be checking that the mock FnApiControlClient got a ProcessBundle InstructionRequest with the cache token specified and not verifying that the state request handler was invoked.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/state/StateRequestHandlers.java
mxm: That's correct. Thanks.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/state/StateRequestHandlers.java
mxm: Agree, for now `Optional` looks like the best choice.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/state/StateRequestHandlers.java
mxm: Not sure how much this contributes here to the overall computing, but the performance of the blog post indeed looks pretty bad. I've replaced this with for loops.

runners/java-fn-execution/src/test/java/org/apache/beam/runners/fnexecution/control/SdkHarnessClientTest.java
mxm: Fair point. Updated.

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
mxm: Note that this is possible now because the cache token is passed via an `Optional`.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/state/StateRequestHandlers.java
lukecwik: It grinds my gears when people use streams for things because they want a fluent builder pattern for processing when a for loop is easier to read/understand for the majority of people.

mxm: It depends what you are used to. I find the fluent pattern easier to read, but I agree that the procedural definition is probably easier for most people.

PR #9376, opened 2019-08-19
[BEAM-8014] Using OffsetRange as restriction for OffsetRestrictionTracker

boyuanzz:
R: @robertwb

### DIFF
--- a/sdks/python/apache_beam/io/restriction_trackers.py
+++ b/sdks/python/apache_beam/io/restriction_trackers.py
@@ -66,6 +66,9 @@ class OffsetRange(object):
       yield OffsetRange(current_split_start, current_split_stop)
       current_split_start = current_split_stop
 
+  def split_at(self, split_pos):
+    return OffsetRange(self.start, split_pos), OffsetRange(split_pos, self.stop)
+
   def new_tracker(self):
     return OffsetRangeTracker(self.start, self.stop)
 
@@ -162,8 +165,7 @@ class OffsetRestrictionTracker(RestrictionTracker):
             cur + int(max(1, (self._range.stop - cur) * fraction_of_remainder)))
         if split_point < self._range.stop:
           prev_stop, self._range.stop = self._range.stop, split_point
-          return (OffsetRange(self._range.start, split_point),
-                  OffsetRange(split_point, prev_stop))
+          return self._range.split_at(split_point)
 
   # TODO(SDF): Replace all calls with try_claim(0).
   def checkpoint(self):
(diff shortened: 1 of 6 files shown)


boyuanzz:
Kindly pinged.

sdks/python/apache_beam/io/restriction_trackers.py
```diff
@@ -159,7 +162,8 @@ def try_split(self, fraction_of_remainder):
             cur + int(max(1, (self._range.stop - cur) * fraction_of_remainder)))
         if split_point < self._range.stop:
           prev_stop, self._range.stop = self._range.stop, split_point
-          return (self._range.start, split_point), (split_point, prev_stop)
+          return (OffsetRange(self._range.start, split_point),
```
robertwb: I wonder if a `OffsetRange.split(split_pos)` method would make sense here?

sdks/python/apache_beam/io/restriction_trackers.py
```diff
@@ -171,7 +175,7 @@ def checkpoint(self):
       else:
         end_position = self._current_position + 1
 
-      residual_range = (end_position, self._range.stop)
+      residual_range = OffsetRange(end_position, self._range.stop)
```
robertwb: In which case this could be

`self._range, residual_range = self._range.split(end_position)`

sdks/python/apache_beam/testing/synthetic_pipeline.py
```diff
@@ -520,12 +520,12 @@ def process(
       element,
       restriction_tracker=beam.DoFn.RestrictionParam(
           SyntheticSDFSourceRestrictionProvider())):
-    for k in range(*restriction_tracker.current_restriction()):
-      if not restriction_tracker.try_claim(k):
-        return
-      r = np.random.RandomState(k)
+    cur = restriction_tracker.start_position()
+    while restriction_tracker.try_claim(cur):
```
robertwb: I think a for loop would still be more natural here.

sdks/python/apache_beam/io/restriction_trackers.py
boyuanzz: Yes, I agree. There is a `OffsetRange.split(bundle_size)` to perform initial split already. If we want to provide splitting into 2 pieces based on given pos, we need to have a different naming, like 'split_at'?

sdks/python/apache_beam/testing/synthetic_pipeline.py
boyuanzz: Yep, I can change back to for loop.

PR #9398, opened 2019-08-22
[BEAM-7980] Exactly once artifact retrieval for Python SDK worker pool

tweise:
This is the follow-up for exactly once artifact retrieval. 

------------------------

--- | --- | --- | --- | ---


sdks/python/container/boot.go
```diff
@@ -105,18 +107,57 @@ func main() {
...
+		dir := filepath.Join(*semiPersistDir, "staged")
+
+		files, err := artifact.Materialize(ctx, *artifactEndpoint, info.GetRetrievalToken(), dir)
+		if err != nil {
+			log.Fatalf("Failed to retrieve staged files: %v", err)
+		}
+
+		// TODO(herohde): the packages to install should be specified explicitly. It
+		// would also be possible to install the SDK in the Dockerfile.
+		if setupErr := installSetupPackages(files, dir); setupErr != nil {
```
tweise: In the future we could look into inverting this logic by extracting the exactly-once logic into a separate function that can be shared between the SDK containers.

aaltay: Could we actually separate the artifact/install logic L144-L155 into a function from the locking related logic.

sdks/python/container/boot.go
```diff
@@ -105,18 +107,57 @@ func main() {
...
-	}
+		// skip if install already complete
+		_, err = os.Stat(installCompleteFile)
+		if err == nil {
+			return
+		}
+
+		// lock to guard from concurrent artifact retrieval and installation,
+		// when called by child processes in a worker pool
+		lock, err := lockfile.New(filepath.Join(os.TempDir(), "beam.install.lck"))
```
aaltay: Should we separate `filepath.Join(os.TempDir(), "beam.install.lck")` to a variable similar to `installCompleteFile`? Not sure what is the go convention if it is being used only in one place.

sdks/python/container/boot.go
```diff
@@ -105,18 +107,57 @@ func main() {
...
+		if setupErr := installSetupPackages(files, dir); setupErr != nil {
+			log.Fatalf("Failed to install required packages: %v", setupErr)
+		}
+
+		// mark install complete
+		os.OpenFile(installCompleteFile, os.O_RDONLY|os.O_CREATE, 0666)
+
+    }()
 
 	// (3) Invoke python
```
aaltay: Related to env variables below. We need to make sure that no worker receives different endpoint. (This is a non-issue today.)

sdks/python/container/boot.go
```diff
@@ -33,6 +34,7 @@ import (
 	"github.com/apache/beam/sdks/go/pkg/beam/util/execx"
 	"github.com/apache/beam/sdks/go/pkg/beam/util/grpcx"
 	"github.com/golang/protobuf/proto"
+	"github.com/nightlyone/lockfile"
```
aaltay: For my learning, how does one pick/vet right go packages?

tweise:
@aaltay thanks for taking a look, is there any other Go expert that can be tagged here?

sdks/python/container/boot.go
lostluck: About the same way as any other package: looking at the code. There's no good third party vetting system for Go at the moment.

As for this one, it's a small package, (~200 lines), and only imports from the standard library, making it relatively safe, and easy to inspect.
https://github.com/nightlyone/lockfile/blob/master/lockfile.go

sdks/python/container/boot.go
lostluck: If it's only used in one place, this is fine as is. It would not be wrong to move the path definition elsewhere, but it would be indirect.

sdks/python/container/boot.go
```diff
@@ -105,18 +107,57 @@ func main() {
...
 	// (2) Retrieve and install the staged packages.
 
-	dir := filepath.Join(*semiPersistDir, "staged")
+	func() {
 
-	files, err := artifact.Materialize(ctx, *artifactEndpoint, info.GetRetrievalToken(), dir)
-	if err != nil {
-		log.Fatalf("Failed to retrieve staged files: %v", err)
-	}
+        installCompleteFile := filepath.Join(os.TempDir(), "beam.install.complete")
```
lostluck: The indentation here is wonky (it's using spaces instead of tabs). Please run go fmt on this file.
https://blog.golang.org/go-fmt-your-code

sdks/python/container/boot.go
```diff
@@ -105,18 +107,57 @@ func main() {
...
+		// lock to guard from concurrent artifact retrieval and installation,
+		// when called by child processes in a worker pool
+		lock, err := lockfile.New(filepath.Join(os.TempDir(), "beam.install.lck"))
+		if err != nil {
+			log.Fatalf("Cannot init artifact retrieval lock: %v", err)
+		}
+
+		for err = lock.TryLock(); err != nil; err = lock.TryLock() {
+			switch err {
+			case lockfile.ErrBusy, lockfile.ErrNotExist:
```
lostluck: Nit: This is fine as is, but here's an optional suggestion.
In this instance the lockfile package author has a type TemporaryError to indicate the error can be retried,which we can check for instead of listing all temporary errors.

if _, ok := err.(lockfile.TemporaryError); ok {
  //sleep etc
} else {
  log.Fatalf(...
}

Or if you're keen on a switch, how about a type switch?
switch err.(type) {
case lockfile.TemporaryError:
  //sleep etc
default:
  log.Fatalf(...
}

sdks/python/container/boot.go
lostluck: Is that to mean that the loggingEndpoint and controlEndpoints are the same?

sdks/python/container/boot.go
```diff
@@ -105,18 +107,57 @@ func main() {
...
 
-	// TODO(herohde): the packages to install should be specified explicitly. It
-	// would also be possible to install the SDK in the Dockerfile.
-	if setupErr := installSetupPackages(files, dir); setupErr != nil {
-		log.Fatalf("Failed to install required packages: %v", setupErr)
-	}
+		// skip if install already complete
+		_, err = os.Stat(installCompleteFile)
+		if err == nil {
+			return
```
mxm: I think this should live within the locked section.

sdks/python/container/boot.go
```diff
@@ -105,18 +107,57 @@ func main() {
...
+			default:
+				log.Fatalf("Worker %v could not obtain artifact retrieval lock: %v", *id, err)
+			}
+		}
+		defer lock.Unlock()
+
+		// skip if install already complete
+		_, err = os.Stat(installCompleteFile)
+		if err == nil {
+			return
```
mxm: Why is this not within the locked section?

mxm: Do you want to avoid blocking the workers?

aaltay: This is within the lock section. defer ... above executes after the completion of this function.

I think we should block all workers as intended here, because workers will need the environment to be setup first.

sdks/python/container/boot.go
tweise: There is no need to enter the locked section when the install was already completed (hence the comment above).

lostluck:
FYI I'm gone until Sept 3rd, please don't block on unnecessarily on further feed back from me.

sdks/python/container/boot.go
mxm: Is this check really necessary? In case of a race condition when multiple workers complete this section before the install file has been created, only the later check will catch that. For simplicity, I'd remove this check entirely, unless the locking is very expensive.

sdks/python/container/boot.go
mxm: Makes sense. I'm new to Go. However, if you look at the above similar check (line 115), there is a possible race condition which can only be caught here where the lock is in place.

sdks/python/container/boot.go
tweise: Done! Can this be accomplished through the gradle plugin also (w/o extra go install)? Would be nice to have this integrated like Spotless for Java.

sdks/python/container/boot.go
tweise: They don't have to be same on Flink, but the staged artifacts have to be identical.

tweise:
@lostluck @aaltay @mxm PTAL since I made a few more changes than those requested to improve the synchronization logic.

mxm:
I think this can be squashed and merged. @lostluck's comments seem to be resolved.

aaltay:
New changes LGTM.

lostluck:
LGTM after the fact. Good use of the first class function for the lock wrapping.

PR #9417, opened 2019-08-23
[BEAM-8021] Add Automatic-Module-Name headers to beam's artifacts.

lgajowy:
The motivation is best described here:
https://github.com/GoogleCloudPlatform/cloud-opensource-java/blob/master/library-best-practices/JLBP-20.md

Quote:
"For compatibility with the Java Platform Module System (JPMS) in Java 9
and later, every JAR you publish should have a module name, even if
the library does not itself use modules. More precisely, the JAR manifest
in META-INF/MANIFEST.MF should have an Automatic-Module-Name field"

R: @iemejia could you take a look? In case you think someone else should take a look too (it's a big pr, sorry) feel free to tag them

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy
+++ b/buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy
@@ -144,6 +144,9 @@ class BeamModulePlugin implements Plugin<Project> {
 
     /** Override the default "beam-" + `dash separated path` archivesBaseName. */
     String archivesBaseName = null;
+
+    /** Controls whether this project is published to Maven. */
+    boolean publish = true
   }
 
   // A class defining the set of configurable properties for createJavaExamplesArchetypeValidationTask
@@ -1566,6 +1569,7 @@ class BeamModulePlugin implements Plugin<Project> {
       project.ext.applyJavaNature(
               exportJavadoc: false,
               enableSpotbugs: false,
+              publish: configuration.publish,
               archivesBaseName: configuration.archivesBaseName,
               shadowJarValidationExcludes: it.shadowJarValidationExcludes,
               shadowClosure: GrpcVendoring.shadowClosure() << {
--- a/buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy
+++ b/buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy
@@ -130,6 +130,14 @@ class BeamModulePlugin implements Plugin<Project> {
 
     /** Controls whether javadoc is exported for this project. */
     boolean exportJavadoc = true
+
+    /**
+     * Automatic-Module-Name Header value to be set in MANFIEST.MF file.
+     * This is a required parameter unless publishing to Maven is disabled for this project.
+     *
+     * @see: https://github.com/GoogleCloudPlatform/cloud-opensource-java/blob/master/library-best-practices/JLBP-20.md
+     */
+    String automaticModuleName = null
   }
 
   /** A class defining the set of configurable properties accepted by applyPortabilityNature. */
@@ -147,6 +155,14 @@ class BeamModulePlugin implements Plugin<Project> {
 
     /** Controls whether this project is published to Maven. */
     boolean publish = true
+
+    /**
+     * Automatic-Module-Name Header value to be set in MANFIEST.MF file.
+     * This is a required parameter unless publishing to Maven is disabled for this project.
+     *
+     * @see: https://github.com/GoogleCloudPlatform/cloud-opensource-java/blob/master/library-best-practices/JLBP-20.md
+     */
+    String automaticModuleName
   }
 
   // A class defining the set of configurable properties for createJavaExamplesArchetypeValidationTask
@@ -920,6 +936,8 @@ class BeamModulePlugin implements Plugin<Project> {
       }
 
       project.jar {
+        setAutomaticModuleNameHeader(configuration, project)
+
         zip64 true
         into("META-INF/") {
           from "${project.rootProject.projectDir}/LICENSE"
@@ -1571,6 +1589,7 @@ class BeamModulePlugin implements Plugin<Project> {
               enableSpotbugs: false,
               publish: configuration.publish,
               archivesBaseName: configuration.archivesBaseName,
+              automaticModuleName: configuration.automaticModuleName,
               shadowJarValidationExcludes: it.shadowJarValidationExcludes,
               shadowClosure: GrpcVendoring.shadowClosure() << {
                 // We perform all the code relocations but don't include
@@ -2046,6 +2065,16 @@ class BeamModulePlugin implements Plugin<Project> {
         ->
         addPortableWordCountTask(false)
         addPortableWordCountTask(true)
+      }
+    }
+  }
+
+  private void setAutomaticModuleNameHeader(JavaNatureConfiguration configuration, Project project) {
+    if (configuration.publish && !configuration.automaticModuleName) {
+      throw new GradleException("Expected automaticModuleName to be set for the module that is published to maven repository.")
+    } else if (configuration.automaticModuleName) {
+      project.jar.manifest {
+        attributes 'Automatic-Module-Name': configuration.automaticModuleName
       }
     }
   }
--- a/model/fn-execution/build.gradle
+++ b/model/fn-execution/build.gradle
@@ -17,7 +17,10 @@
  */
 
 plugins { id 'org.apache.beam.module' }
-applyPortabilityNature(shadowJarValidationExcludes: ["org/apache/beam/model/fnexecution/v1/**"])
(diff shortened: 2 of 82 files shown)


runners/flink/1.5/build.gradle
```diff
@@ -28,6 +28,7 @@ project.ext {
   main_resources_dirs = ["$basePath/src/main/resources"]
   test_resources_dirs = ["$basePath/src/test/resources"]
   archives_base_name = 'beam-runners-flink_2.11'
+  automatic_module_name = 'beam.runners.flink.2.11'
```
lgajowy: I am not sure if this is the right module name for flink in different versions but couldn't come up with anything else. Please take extra look here. :)

iemejia:
I am adding to the review both @lukecwik and @kennknowles since both are probably the most knowledgeable people about the Beam build system and may also spot some unnamed artifact or Java 11 consequence easily. I think we need to look at this with extreme detail.

model/fn-execution/build.gradle
```diff
@@ -17,7 +17,10 @@
  */
 
 plugins { id 'org.apache.beam.module' }
-applyPortabilityNature(shadowJarValidationExcludes: ["org/apache/beam/model/fnexecution/v1/**"])
+applyPortabilityNature(
+    automaticModuleName: 'beam.model.fn.execution',
```
iemejia: wondering if for this and the other modules it would be better to have the fully qualified name: `org.apache.beam.model.fn.execution`, `org.apache.beam....`. It could be long but it seems to be the convention, What do others think?
I assume that some project can choose the beam project and have some clash with us, however in principle nobody will choose the fully qualified Apache path.

runners/flink/1.5/build.gradle
iemejia: I don't know if the _2.11 path will be removed in the future I suppose that naming it with the exact version is probably better. Maybe @mxm may have a say. 

I am wondering also if the `.` in the version of the name can produce unintented consequences (package wise) in Java 11 but I don't think so.

runners/google-cloud-dataflow-java/examples/build.gradle
```diff
@@ -19,7 +19,7 @@
 import groovy.json.JsonOutput
 
 plugins { id 'org.apache.beam.module' }
-applyJavaNature(publish: false, exportJavadoc: false)
+applyJavaNature(publish: false, exportJavadoc: false, setAutomaticModuleName: false)
```
iemejia: I suppose this does not have automtatic module name because people are not suppose to build from it? We should probably document this as part of `setAutomaticModuleName`in `BeamModulePlugin.groovy`. In what particular reasons we do not set automatic module names. I suppose this will be the case also for some test pure modules like the Elasticsearch ones.

sdks/java/io/hadoop-common/build.gradle
```diff
@@ -17,7 +17,7 @@
  */
 
 plugins { id 'org.apache.beam.module' }
-applyJavaNature()
+applyJavaNature(automaticModuleName: 'beam.sdks.java.io.hadoop.common')
```
iemejia: This is a valid name, but I somehow expected it to coincide with the package definition in the code, currently `org.apache.beam.sdk.io.hadoop` I think is ok as it is, but maybe someone else can argue pro being consistent.

runners/flink/1.5/build.gradle
mxm: I think `beam.runners.flink` would suffice here. If you want to include the version, you can use the new artifact naming convention, e.g. `beam.runners.flink.1.5`

buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy
```diff
@@ -38,6 +37,8 @@ import org.gradle.api.tasks.javadoc.Javadoc
 import org.gradle.api.tasks.testing.Test
 import org.gradle.testing.jacoco.tasks.JacocoReport
 
+import java.util.concurrent.atomic.AtomicInteger
```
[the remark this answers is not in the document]
lgajowy: I think I have something incorrectly set up in my formatting rules. According to our convention in Beam, there should be alphabetical order. I will change it back.

model/fn-execution/build.gradle
lgajowy: You are right - I should change it as you suggest. There are strong arguments to do so: https://blog.joda.org/2017/04/java-se-9-jpms-module-naming.html

runners/google-cloud-dataflow-java/examples/build.gradle
lgajowy: I'm still not 100% sure if we should leave modules without these headers, to be honest. My motivation was to skip it in aggregated docs (javadoc module), examples and testing modules because we know that people should not depend on it (do we?). I also find it as a form of discouragement to depend on them so but a very weak one.

sdks/java/io/hadoop-common/build.gradle
lgajowy: The reason I named it "common" is because it holds some commons so I didn't want to block the hadoop name for the future.

runners/flink/1.5/build.gradle
lgajowy: I will keep the version as Max suggests.

mxm: I would go without the version. No need to encode it.

lgajowy: I think I misunderstood you before. 

I did some thinking - If we decide to name the modules without version `beam.runners.flink`, we will prevent users from importing 2 different versions of Flink a once*. Preventing this sounds sane now but will it always be the case?

According to my knowledge and current reasoning, the answer is "yes". Could you confirm?

*JPMS will simply refuse to start a modulepath where two modules have the same name, even if they contain different packages.

mxm: Exactly. That's why I would prefer `beam.runners.flink`.

lgajowy:
I posted new fixes. Thanks to @iemejia 's comments I've read some more about the topic (https://blog.joda.org/2017/04/java-se-9-jpms-module-naming.html) and it turned out most of the names didn't adhere this rule: 

> Module names are strongly recommended to be the same as the name of the super-package.

Therefore, I changed the name wherever it was needed.

@iemejia could you take a look again?

lgajowy:
I also changed flink module name as @mxm suggested

lgajowy:
@lukecwik thank you. I agree that we should not publish the artifacts in the first place. What I didn't realize is that this is a matter of setting one `publish` switch in nature's configuration. I changed the PR according to your suggestions.

lgajowy:
@lukecwik devlist thread: https://lists.apache.org/thread.html/379d5b6a556530b157b3bafdb21d135152cba6f65d87a9bc7d2f8246@%3Cdev.beam.apache.org%3E

runners/google-cloud-dataflow-java/worker/legacy-worker/build.gradle
```diff
@@ -81,6 +81,7 @@ def excluded_dependencies = [
 ]
 
 applyJavaNature(
+        automaticModuleName: 'org.apache.beam.runners.dataflow.legacy.worker',
```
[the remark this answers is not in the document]
lgajowy: Good catch, thanks

lgajowy:
Thanks! I posted the fixes for zetasketch, removed the legacy-worker's unneeded module name and added the documentation. Now I'll think about the task you suggested (`validateShadedJarDoesntLeakNonProjectClasses`). 

Let's also wait for the devlist to respond a few days.

lukecwik:
Run Java PreCommit

pabloem:
Run Java PreCommit

lgajowy:
@lukecwik I posted another round of fixes, more specifically: 
 - nexmark, test-utils and expansion-service are still published with a proper atomaticModuleName. This comes from the devlist discussion.
 - I changed the prefixes of module names such as `org.apache.beam.sdk...` names to `org.apache.beam.sdks.java...`. I think we should still include the sdk name in case we have different sdks in JVM languages (kotlin?, scala?). As far as I understand, they will also require the automaticModuleName so it's good to prevent possible future collisions. 

Could you take a look again?

lgajowy:
@lukecwik you're right - I was too hasty. Reverted the name change to the previous state.

sdks/java/io/common/build.gradle
```diff
@@ -17,7 +17,7 @@
  */
 
 plugins { id 'org.apache.beam.module' }
-applyJavaNature(exportJavadoc: false)
+applyJavaNature(exportJavadoc: false, automaticModuleName: 'org.apache.beam.sdk.io.common')
```
lgajowy: @lukecwik I think that this module should also not be published - I think I missed this. It contains only /test utilities and is used only by test code. 

Let me know if you agree and if I should notify the devlist again or simply stop publishing.

lgajowy:
Thanks!

PR #9418, opened 2019-08-23
[BEAM-5428] Implement cross-bundle user state caching in the Python SDK

mxm:
This caches user state in the Python SDK using a LRU cache with a bounded number
of items. Cache entries are guarded by cache tokens which are supplied by the
Runner in every ProcessBundleRequest. Cache entries are stored throughout
bundles and are evicted either by the least-recently-used strategy or when a new
cache entry for the same state key is stored.

The caching layer sits in front of the state retrieval layer which means that
requests are going to the cache first, if a valid cache token exists and the
cache contains an entry for a state key. Upon first reading state for a given
state key, the caching layer materializes the state and stores it in the
cache.

With the recent changes to the cache token handling in the Proto, append or
clear requests are executed non-blocking because the cache tokens provided for a
bundle remain valid for the life time of a bundle.

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/runners/portability/fn_api_runner.py
+++ b/sdks/python/apache_beam/runners/portability/fn_api_runner.py
@@ -70,6 +70,8 @@ from apache_beam.runners.worker import bundle_processor
 from apache_beam.runners.worker import data_plane
 from apache_beam.runners.worker import sdk_worker
 from apache_beam.runners.worker.channel_factory import GRPCChannelFactory
+from apache_beam.runners.worker.sdk_worker import _Future
+from apache_beam.runners.worker.statecache import StateCache
 from apache_beam.transforms import trigger
 from apache_beam.transforms.window import GlobalWindows
 from apache_beam.utils import profiler
@@ -81,6 +83,11 @@ ENCODED_IMPULSE_VALUE = beam.coders.WindowedValueCoder(
     beam.coders.BytesCoder(),
     beam.coders.coders.GlobalWindowCoder()).get_impl().encode_nested(
         beam.transforms.window.GlobalWindows.windowed_value(b''))
+
+# State caching is enabled in the fn_api_runner for testing, except for one
+# test which runs without state caching (FnApiRunnerTestWithDisabledCaching).
+# The cache is disabled in production for other runners.
+STATE_CACHE_SIZE = 100
 
 
 class ControlConnection(object):
@@ -487,11 +494,11 @@ class FnApiRunner(runner.PipelineRunner):
                 side_input_id=tag,
                 window=window,
                 key=key))
-        worker_handler.state.blocking_append(state_key, elements_data)
+        worker_handler.state.append_raw(state_key, elements_data)
 
   def _run_bundle_multiple_times_for_testing(
       self, worker_handler_list, process_bundle_descriptor, data_input,
-      data_output, get_input_coder_callable):
+      data_output, get_input_coder_callable, cache_token_generator):
 
     # all workers share state, so use any worker_handler.
     worker_handler = worker_handler_list[0]
@@ -501,7 +508,9 @@ class FnApiRunner(runner.PipelineRunner):
         ParallelBundleManager(
             worker_handler_list, lambda pcoll_id: [],
             get_input_coder_callable, process_bundle_descriptor,
-            self._progress_frequency, k, num_workers=self._num_workers
+            self._progress_frequency, k,
+            num_workers=self._num_workers,
+            cache_token_generator=cache_token_generator
         ).process_bundle(data_input, data_output)
       finally:
         worker_handler.state.restore()
@@ -640,7 +649,7 @@ class FnApiRunner(runner.PipelineRunner):
       out = create_OutputStream()
       for element in values:
         element_coder_impl.encode_to_stream(element, out, True)
-      worker_handler.state.blocking_append(
+      worker_handler.state.append_raw(
           beam_fn_api_pb2.StateKey(
               runner=beam_fn_api_pb2.StateKey.Runner(key=token)),
           out.get())
@@ -724,16 +733,18 @@ class FnApiRunner(runner.PipelineRunner):
           ).coder_id
       ]].get_impl()
 
-    self._run_bundle_multiple_times_for_testing(worker_handler_list,
-                                                process_bundle_descriptor,
-                                                data_input,
-                                                data_output,
-                                                get_input_coder_impl)
+    # Change cache token across bundle repeats
+    cache_token_generator = FnApiRunner.get_cache_token_generator(static=False)
+
+    self._run_bundle_multiple_times_for_testing(
+        worker_handler_list, process_bundle_descriptor, data_input, data_output,
+        get_input_coder_impl, cache_token_generator=cache_token_generator)
 
     bundle_manager = ParallelBundleManager(
         worker_handler_list, get_buffer, get_input_coder_impl,
         process_bundle_descriptor, self._progress_frequency,
-        num_workers=self._num_workers)
+        num_workers=self._num_workers,
+        cache_token_generator=cache_token_generator)
 
     result, splits = bundle_manager.process_bundle(data_input, data_output)
 
@@ -916,7 +927,7 @@ class FnApiRunner(runner.PipelineRunner):
     def process_instruction_id(self, unused_instruction_id):
       yield
 
-    def blocking_get(self, state_key, continuation_token=None):
+    def get_raw(self, state_key, continuation_token=None):
       with self._lock:
         full_state = self._state[self._to_key(state_key)]
         if self._use_continuation_tokens:
@@ -937,13 +948,22 @@ class FnApiRunner(runner.PipelineRunner):
           assert not continuation_token
           return b''.join(full_state), None
 
-    def blocking_append(self, state_key, data):
+    def append_raw(self, state_key, data):
       with self._lock:
         self._state[self._to_key(state_key)].append(data)
+      return _Future.done()
 
-    def blocking_clear(self, state_key):
+    def clear(self, state_key):
       with self._lock:
-        del self._state[self._to_key(state_key)]
+        try:
+          del self._state[self._to_key(state_key)]
+        except KeyError:
+          # This may happen with the caching layer across bundles. Caching may
+          # skip this storage layer for a blocking_get(key) request. Without
+          # the caching, the state for a key would be initialized via the
+          # defaultdict that _state uses.
+          pass
+      return _Future.done()
 
     @staticmethod
     def _to_key(state_key):
@@ -959,19 +979,19 @@ class FnApiRunner(runner.PipelineRunner):
       for request in request_stream:
         request_type = request.WhichOneof('request')
         if request_type == 'get':
-          data, continuation_token = self._state.blocking_get(
+          data, continuation_token = self._state.get_raw(
               request.state_key, request.get.continuation_token)
           yield beam_fn_api_pb2.StateResponse(
               id=request.id,
               get=beam_fn_api_pb2.StateGetResponse(
                   data=data, continuation_token=continuation_token))
         elif request_type == 'append':
-          self._state.blocking_append(request.state_key, request.append.data)
+          self._state.append_raw(request.state_key, request.append.data)
           yield beam_fn_api_pb2.StateResponse(
               id=request.id,
               append=beam_fn_api_pb2.StateAppendResponse())
         elif request_type == 'clear':
-          self._state.blocking_clear(request.state_key)
+          self._state.clear(request.state_key)
           yield beam_fn_api_pb2.StateResponse(
               id=request.id,
               clear=beam_fn_api_pb2.StateClearResponse())
@@ -991,6 +1011,46 @@ class FnApiRunner(runner.PipelineRunner):
     def close(self):
       """Does nothing."""
       pass
+
+  @staticmethod
+  def get_cache_token_generator(static=True):
+    """A generator for cache tokens.
+       :arg static If True, generator always returns the same cache token
+                   If False, generator returns a new cache token each time
+       :return A generator which returns a cache token on next(generator)
+    """
+    def generate_token(identifier):
+      return beam_fn_api_pb2.ProcessBundleRequest.CacheToken(
+          user_state=beam_fn_api_pb2
+          .ProcessBundleRequest.CacheToken.UserState(),
+          token="cache_token_{}".format(identifier).encode("utf-8"))
+
+    class StaticGenerator(object):
+      def __init__(self):
+        self._token = generate_token(1)
+
+      def __iter__(self):
+        # pylint: disable=non-iterator-returned
+        return self
+
+      def __next__(self):
+        return self._token
+
+    class DynamicGenerator(object):
+      def __init__(self):
+        self._counter = 0
+        self._lock = threading.Lock()
+
+      def __iter__(self):
+        # pylint: disable=non-iterator-returned
+        return self
+
+      def __next__(self):
+        with self._lock:
+          self._counter += 1
+          return generate_token(self._counter)
+
+    return StaticGenerator() if static else DynamicGenerator()
 
 
 class WorkerHandler(object):
@@ -1069,10 +1129,11 @@ class EmbeddedWorkerHandler(WorkerHandler):
         self, data_plane.InMemoryDataChannel(), state, provision_info)
     self.control_conn = self
     self.data_conn = self.data_plane_handler
-
     self.worker = sdk_worker.SdkWorker(
         sdk_worker.BundleProcessorCache(
-            FnApiRunner.SingletonStateHandlerFactory(self.state),
+            FnApiRunner.SingletonStateHandlerFactory(
+                sdk_worker.CachingMaterializingStateHandler(
+                    StateCache(STATE_CACHE_SIZE), state)),
             data_plane.InMemoryDataChannelFactory(
                 self.data_plane_handler.inverse()),
             {}))
@@ -1303,15 +1364,21 @@ class ExternalWorkerHandler(GrpcWorkerHandler):
 
 @WorkerHandler.register_environment(python_urns.EMBEDDED_PYTHON_GRPC, bytes)
 class EmbeddedGrpcWorkerHandler(GrpcWorkerHandler):
-  def __init__(self, num_workers_payload, state, provision_info, grpc_server):
+  def __init__(self, payload, state, provision_info, grpc_server):
     super(EmbeddedGrpcWorkerHandler, self).__init__(state, provision_info,
                                                     grpc_server)
-    self._num_threads = int(num_workers_payload) if num_workers_payload else 1
+    if payload:
+      num_workers, state_cache_size = payload.decode('ascii').split(',')
+      self._num_threads = int(num_workers)
+      self._state_cache_size = int(state_cache_size)
+    else:
+      self._num_threads = 1
+      self._state_cache_size = STATE_CACHE_SIZE
 
   def start_worker(self):
     self.worker = sdk_worker.SdkHarness(
         self.control_address, worker_count=self._num_threads,
-        worker_id=self.worker_id)
+        state_cache_size=self._state_cache_size, worker_id=self.worker_id)
     self.worker_thread = threading.Thread(
         name='run_worker', target=self.worker.run)
     self.worker_thread.daemon = True
@@ -1512,7 +1579,8 @@ class BundleManager(object):
 
   def __init__(
       self, worker_handler_list, get_buffer, get_input_coder_impl,
-      bundle_descriptor, progress_frequency=None, skip_registration=False):
+      bundle_descriptor, progress_frequency=None, skip_registration=False,
+      cache_token_generator=FnApiRunner.get_cache_token_generator()):
     """Set up a bundle manager.
 
     Args:
@@ -1530,6 +1598,7 @@ class BundleManager(object):
     self._registered = skip_registration
     self._progress_frequency = progress_frequency
     self._worker_handler = None
+    self._cache_token_generator = cache_token_generator
 
   def _send_input_to_worker(self,
                             process_bundle_id,
@@ -1653,7 +1722,8 @@ class BundleManager(object):
     process_bundle_req = beam_fn_api_pb2.InstructionRequest(
         instruction_id=process_bundle_id,
         process_bundle=beam_fn_api_pb2.ProcessBundleRequest(
-            process_bundle_descriptor_id=self._bundle_descriptor.id))
+            process_bundle_descriptor_id=self._bundle_descriptor.id,
+            cache_tokens=[next(self._cache_token_generator)]))
     result_future = self._worker_handler.control_conn.push(process_bundle_req)
 
     split_results = []
@@ -1697,10 +1767,11 @@ class ParallelBundleManager(BundleManager):
   def __init__(
       self, worker_handler_list, get_buffer, get_input_coder_impl,
       bundle_descriptor, progress_frequency=None, skip_registration=False,
-      **kwargs):
+      cache_token_generator=None, **kwargs):
     super(ParallelBundleManager, self).__init__(
         worker_handler_list, get_buffer, get_input_coder_impl,
-        bundle_descriptor, progress_frequency, skip_registration)
+        bundle_descriptor, progress_frequency, skip_registration,
+        cache_token_generator=cache_token_generator)
     self._num_workers = kwargs.pop('num_workers', 1)
 
   def process_bundle(self, inputs, expected_outputs):
@@ -1715,7 +1786,8 @@ class ParallelBundleManager(BundleManager):
       for result, split_result in executor.map(lambda part: BundleManager(
           self._worker_handler_list, self._get_buffer,
           self._get_input_coder_impl, self._bundle_descriptor,
-          self._progress_frequency, self._registered).process_bundle(
+          self._progress_frequency, self._registered,
+          cache_token_generator=self._cache_token_generator).process_bundle(
               part, expected_outputs), part_inputs):
 
         split_result_list += split_result
--- a/sdks/python/apache_beam/runners/portability/portable_runner_test.py
+++ b/sdks/python/apache_beam/runners/portability/portable_runner_test.py
@@ -179,6 +179,8 @@ class PortableRunnerTest(fn_api_runner_test.FnApiRunnerTest):
     # Override the default environment type for testing.
     options.view_as(PortableOptions).environment_type = (
         python_urns.EMBEDDED_PYTHON)
+    # Enable caching (disabled by default)
+    options.view_as(DebugOptions).add_experiment('state_cache_size=100')
     return options
 
   def create_pipeline(self):
@@ -193,6 +195,7 @@ class PortableRunnerOptimized(PortableRunnerTest):
   def create_options(self):
     options = super(PortableRunnerOptimized, self).create_options()
     options.view_as(DebugOptions).add_experiment('pre_optimize=all')
+    options.view_as(DebugOptions).add_experiment('state_cache_size=100')
     return options
 
 
@@ -201,7 +204,8 @@ class PortableRunnerTestWithExternalEnv(PortableRunnerTest):
   @classmethod
   def setUpClass(cls):
     cls._worker_address, cls._worker_server = (
-        worker_pool_main.BeamFnExternalWorkerPoolServicer.start())
+        worker_pool_main.BeamFnExternalWorkerPoolServicer.start(
+            state_cache_size=100))
 
   @classmethod
   def tearDownClass(cls):
@@ -224,6 +228,8 @@ class PortableRunnerTestWithSubprocesses(PortableRunnerTest):
     options.view_as(PortableOptions).environment_config = (
         b'%s -m apache_beam.runners.worker.sdk_worker_main' %
         sys.executable.encode('ascii')).decode('utf-8')
+    # Enable caching (disabled by default)
+    options.view_as(DebugOptions).add_experiment('state_cache_size=100')
     return options
 
   @classmethod
--- a/sdks/python/apache_beam/runners/worker/bundle_processor.py
+++ b/sdks/python/apache_beam/runners/worker/bundle_processor.py
@@ -199,26 +199,19 @@ class DataInputOperation(RunnerIOOperation):
 
 
 class _StateBackedIterable(object):
-  def __init__(self, state_handler, state_key, coder_or_impl):
+  def __init__(self, state_handler, state_key, coder_or_impl,
+               is_cached=False):
     self._state_handler = state_handler
     self._state_key = state_key
     if isinstance(coder_or_impl, coders.Coder):
       self._coder_impl = coder_or_impl.get_impl()
     else:
       self._coder_impl = coder_or_impl
+    self._is_cached = is_cached
 
   def __iter__(self):
-    # This is the continuation token this might be useful
-    data, continuation_token = self._state_handler.blocking_get(self._state_key)
-    while True:
-      input_stream = coder_impl.create_InputStream(data)
-      while input_stream.size() > 0:
-        yield self._coder_impl.decode_from_stream(input_stream, True)
-      if not continuation_token:
-        break
-      else:
-        data, continuation_token = self._state_handler.blocking_get(
-            self._state_key, continuation_token)
+    return self._state_handler.blocking_get(
+        self._state_key, self._coder_impl, is_cached=self._is_cached)
 
   def __reduce__(self):
     return list, (list(self),)
@@ -294,6 +287,7 @@ class StateBackedSideInputMap(object):
 
 
 class CombiningValueRuntimeState(userstate.CombiningValueRuntimeState):
+
   def __init__(self, underlying_bag_state, combinefn):
     self._combinefn = combinefn
     self._underlying_bag_state = underlying_bag_state
@@ -347,8 +341,8 @@ class _ConcatIterable(object):
 coder_impl.FastPrimitivesCoderImpl.register_iterable_like_type(_ConcatIterable)
 
 
-# TODO(BEAM-5428): Implement cross-bundle state caching.
 class SynchronousBagRuntimeState(userstate.BagRuntimeState):
+
   def __init__(self, state_handler, state_key, value_coder):
     self._state_handler = state_handler
     self._state_key = state_key
@@ -359,7 +353,8 @@ class SynchronousBagRuntimeState(userstate.BagRuntimeState):
   def read(self):
     return _ConcatIterable(
         [] if self._cleared else _StateBackedIterable(
-            self._state_handler, self._state_key, self._value_coder),
+            self._state_handler, self._state_key, self._value_coder,
+            is_cached=True),
         self._added_elements)
 
   def add(self, value):
@@ -370,17 +365,20 @@ class SynchronousBagRuntimeState(userstate.BagRuntimeState):
     self._added_elements = []
 
   def _commit(self):
+    to_await = None
     if self._cleared:
-      self._state_handler.blocking_clear(self._state_key)
+      to_await = self._state_handler.clear(self._state_key, is_cached=True)
     if self._added_elements:
-      value_coder_impl = self._value_coder.get_impl()
-      out = coder_impl.create_OutputStream()
-      for element in self._added_elements:
-        value_coder_impl.encode_to_stream(element, out, True)
-      self._state_handler.blocking_append(self._state_key, out.get())
+      to_await = self._state_handler.extend(
+          self._state_key,
+          self._value_coder.get_impl(),
+          self._added_elements,
+          is_cached=True)
+    if to_await:
+      # To commit, we need to wait on the last state request future to complete.
+      to_await.get()
 
 
-# TODO(BEAM-5428): Implement cross-bundle state caching.
 class SynchronousSetRuntimeState(userstate.SetRuntimeState):
 
   def __init__(self, state_handler, state_key, value_coder):
@@ -393,17 +391,17 @@ class SynchronousSetRuntimeState(userstate.SetRuntimeState):
   def _compact_data(self, rewrite=True):
     accumulator = set(_ConcatIterable(
         set() if self._cleared else _StateBackedIterable(
-            self._state_handler, self._state_key, self._value_coder),
+            self._state_handler, self._state_key, self._value_coder,
+            is_cached=True),
         self._added_elements))
 
     if rewrite and accumulator:
-      self._state_handler.blocking_clear(self._state_key)
-
-      value_coder_impl = self._value_coder.get_impl()
-      out = coder_impl.create_OutputStream()
-      for element in accumulator:
-        value_coder_impl.encode_to_stream(element, out, True)
-      self._state_handler.blocking_append(self._state_key, out.get())
+      self._state_handler.clear(self._state_key, is_cached=True)
+      self._state_handler.extend(
+          self._state_key,
+          self._value_coder.get_impl(),
+          accumulator,
+          is_cached=True)
 
       # Since everthing is already committed so we can safely reinitialize
       # added_elements here.
@@ -417,7 +415,7 @@ class SynchronousSetRuntimeState(userstate.SetRuntimeState):
   def add(self, value):
     if self._cleared:
       # This is a good time explicitly clear.
-      self._state_handler.blocking_clear(self._state_key)
+      self._state_handler.clear(self._state_key, is_cached=True)
       self._cleared = False
 
     self._added_elements.add(value)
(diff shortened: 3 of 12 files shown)


sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -464,6 +469,7 @@ def __init__(self, credentials=None):
     self._lock = threading.Lock()
     self._throwing_state_handler = ThrowingStateHandler()
     self._credentials = credentials
+    self.state_cache = StateCache(100)
```
rakeshcusat: do you want have this as a class variable so it can be shared across different instances of this class?

sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -537,15 +545,20 @@ def __init__(self, state_stub):
...
   @contextlib.contextmanager
-  def process_instruction_id(self, bundle_id):
+  def process_instruction_id(self, bundle_id, cache_tokens):
     if getattr(self._context, 'process_instruction_id', None) is not None:
       raise RuntimeError(
           'Already bound to %r' % self._context.process_instruction_id)
     self._context.process_instruction_id = bundle_id
+    if getattr(self._context, 'cache_tokens', None) is not None:
+      raise RuntimeError(
+          'Cache tokens already to %s' % self._context.cache_tokens)
```
rakeshcusat:
```suggestion
          'Cache tokens already set to %s' % self._context.cache_tokens)
```

sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -597,7 +617,7 @@ def blocking_clear(self, state_key):
             state_key=state_key,
             clear=beam_fn_api_pb2.StateClearRequest()))
 
-  def _blocking_request(self, request):
+  def  _blocking_request(self, request):
```
rakeshcusat:
```suggestion
  def _blocking_request(self, request):
```

sdks/python/apache_beam/runners/worker/statecache.py
```diff
@@ -0,0 +1,103 @@
...
+            other.transform_id == self.transform_id and
+            other.state_id == self.state_id and
+            other.window == self.window and
+            other.key == self.key)
+
+  def __hash__(self):
+    return hash((self.transform_id, self.state_id, self.window, self.key))
+
+  def __repr__(self):
+    return ("CacheStateKey["
```
rakeshcusat:
```suggestion
def __repr__(self):
   return  "CacheStateKey(transofrm_id={}, state_id={}, window={}, key={})".format(self.transform_id, self.state_id, self.window, self.key)
```
I would not use the extra parenthesis and replace square bracket with parenthesis.

sdks/python/apache_beam/runners/worker/statecache.py
```diff
@@ -0,0 +1,103 @@
...
+class CacheStateKey(object):
+
+  def __init__(self, transform_id, state_id, window, key):
+    self.transform_id = transform_id
+    self.state_id = state_id
+    self.window = window
+    self.key = key
+
+  def __eq__(self, other):
+    return (isinstance(other, self.__class__) and
```
rakeshcusat: if would also add the id comparison.
```suggesion
if id(self) == id(other):
   return True 
else:
    return (....)

File: sdks/python/apache_beam/runners/worker/sdk_worker.py [hunk 6]
```diff
@@ -613,13 +633,62 @@ def _blocking_request(self, request):
...
+    cache_tokens = self._context.cache_tokens
+    if not cache_tokens:
+      # no cache tokens, can't do a lookup in the cache
+      return super(CachingGrpcStateHandler, self).blocking_get(
+          state_key, coder)
+    cache_state_key = CachingGrpcStateHandler.convert_to_cache_key(state_key)
+    value = self._state_cache.get(cache_state_key, cache_tokens)
+    if not value:
+      # Cache miss, need to retrieve from the Runner
+      value = super(CachingGrpcStateHandler, self).blocking_get(
```
rakeshcusat: We can another TODO here for adding metrics in a separate PR.

sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -613,13 +633,62 @@ def _blocking_request(self, request):
...
+          state_key, coder)
+      # TODO mxm uses always the first cache token for now
+      self._state_cache.put(
+          CachingGrpcStateHandler.convert_to_cache_key(state_key),
+          cache_tokens[0], value)
```
rakeshcusat: I am not sure why we are always taking the first token: `cache_tokens[0]`

rakeshcusat: The current protocol also return the cache token, are we not caching that token here?

sdks/python/apache_beam/runners/worker/statecache.py
```diff
@@ -0,0 +1,103 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""TODO mxm"""
+import collections
+from threading import Lock
+
```
rakeshcusat:
```suggestion
from cachetools import LRUCache
```

sdks/python/apache_beam/runners/worker/sdk_worker.py
rakeshcusat: Implementation of this method suggests that you are passing only all tokens.

sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -613,13 +633,62 @@ def _blocking_request(self, request):
...
+    return value
+
+  def blocking_append(self, state_key, data):
+    cache_tokens = self._context.cache_tokens
+    self._state_cache.put(
```
rakeshcusat: if we are considering the current protocol then put method should return the new token, in that case, we should caching it after updating the runner.

rakeshcusat:
@mxm thanks for taking care of this.

sdks/python/apache_beam/runners/worker/sdk_worker.py
mxm: Good point.

sdks/python/apache_beam/runners/worker/sdk_worker.py
mxm: As per last discussion on the mailing list, there will only be one active cache token for all user state. This token is provided by the Runner in the ProcessBundleRequest. We also decided to ignore the cache token returned by a get/put/clear request because we want to implement async communication and assume validity of the provided token for the duration of a bundle.

That said, if

sdks/python/apache_beam/runners/worker/sdk_worker.py
mxm: I've already implemented the last design discussion around async handling where we assume validity of the Runner-provided cache tokens for the duration of a bundle. The async handling itself is still a TODO.

sdks/python/apache_beam/runners/worker/statecache.py
mxm: Good point. I think this class will be removed because we can do a to `state_key.SerializeToString()` on the Protobuf structure which will be a simpler way to make the state key comparable.

mxm:
Run Portable_Python PreCommit

sdks/python/apache_beam/runners/worker/statecache.py
mxm: I didn't know about that. Why is that? Clearly it can be different for Python 2, but that does not matter for the correctness of the cache.

mxm:
Run Portable_Python PreCommit

mxm:
This is now based on the Proto changes in #9440.

mxm:
@robertwb Could you have a look at the current state of the SDK implementation?

sdks/python/apache_beam/runners/worker/sdk_worker.py
mxm: With the new version, we don't return the token anymore on append/clear.

mxm:

mxm:

mxm:

mxm:
Run Java PreCommit

mxm:

mxm:

sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -620,6 +628,106 @@ def _next_id(self):
...
+          self._underlying.get_raw(state_key, continuation_token)
+      input_stream = coder_impl.create_InputStream(data)
+      while input_stream.size() > 0:
+        yield coder.decode_from_stream(input_stream, True)
+      if not continuation_token:
+        break
+
+  def _should_be_cached(self, request_is_cached):
+    return self._state_cache.is_cache_enabled() and \
+           request_is_cached and \
```
robertwb: Prefer ()'s to backslash line continuations (here and elsewhere).

sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -620,6 +628,106 @@ def _next_id(self):
...
+      # no cache / tokens, can't do a lookup/store in the cache
+      return self._materialize_iter(state_key, coder)
+    # Cache lookup
+    cache_state_key = self._convert_to_cache_key(state_key)
+    cached_value = self._state_cache.get(cache_state_key,
+                                         self._context.cache_token)
+    if cached_value is None:
+      # Cache miss, need to retrieve from the Runner
+      materialized = cached_value = []
+      for val in self._materialize_iter(state_key, coder):
```
robertwb: This is equivalent to 

`materialized = cached_value = list(self._materialize_iter(state_key, coder))`

sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -620,6 +628,106 @@ def _next_id(self):
...
+    if not self._should_be_cached(is_cached):
+      # no cache / tokens, can't do a lookup/store in the cache
+      return self._materialize_iter(state_key, coder)
+    # Cache lookup
+    cache_state_key = self._convert_to_cache_key(state_key)
+    cached_value = self._state_cache.get(cache_state_key,
+                                         self._context.cache_token)
+    if cached_value is None:
+      # Cache miss, need to retrieve from the Runner
+      materialized = cached_value = []
```
robertwb: This makes it impossible to have state larger than fits into memory, whereas before one could page it in. Also, this means the whole thing must be read before the first element gets processed which could be a performance regression.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -308,7 +311,8 @@ def __init__(
       default_environment=None,
       bundle_repeat=0,
       use_state_iterables=False,
-      provision_info=None):
+      provision_info=None,
+      state_cache_size=100):
```
robertwb: Is this number of items? It'd probably be better for this to be sensitive to the size of the items (e.g. estimated by the serialized size we get when making state requests).

sdks/python/apache_beam/runners/worker/sdk_worker.py
robertwb: Possibly one idea would be to only cache the first page (with the reasoning that for something so large one may be OK with repeated fetches).

sdks/python/apache_beam/runners/worker/statecache.py
```diff
@@ -0,0 +1,105 @@
...
+    with self._lock:
+      return self._cache.put(state_key, (cache_token, value))
+
+  def append(self, state_key, cache_token, elements):
+    assert cache_token and self.is_cache_enabled()
+    with self._lock:
+      token, value = self._cache.get(state_key)
+      if token != cache_token:
+        # Discard cached state if tokens do not match
+        value = []
```
robertwb: We can't safely assume the value is empty on a token mismatch though, can we? Seems we'd have to expunge the cache entry altogether in this case.

sdks/python/apache_beam/runners/worker/statecache.py
```diff
@@ -0,0 +1,105 @@
...
+      for element in elements:
+        value.append(element)
```
robertwb: lists have an extend method that could be used here rather than a looping append.

sdks/python/apache_beam/runners/worker/statecache.py
```diff
@@ -0,0 +1,105 @@
...
+      self._cache.put(state_key, (cache_token, value))
+
+  def clear(self, state_key):
+    assert self.is_cache_enabled()
+    with self._lock:
+      self._cache.clear(state_key)
```
robertwb: Clearing an entry from the cache is different than setting the value as cleared (but still being able to look it up). We want the latter, right?

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1660,6 +1671,12 @@ def process_bundle(self, inputs, expected_outputs):
 
     return result, split_results
 
+  @staticmethod
+  def _generate_cache_token():
+    return beam_fn_api_pb2.ProcessBundleRequest.CacheToken(
+        user_state=beam_fn_api_pb2.ProcessBundleRequest.CacheToken.UserState(),
+        token=bytes(bytearray(random.getrandbits(8) for _ in range(16))))
```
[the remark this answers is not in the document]
mxm: Have replaced this with a generator with a counter and a human-readable name.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1622,7 +1632,8 @@ def process_bundle(self, inputs, expected_outputs):
     process_bundle_req = beam_fn_api_pb2.InstructionRequest(
         instruction_id=process_bundle_id,
         process_bundle=beam_fn_api_pb2.ProcessBundleRequest(
-            process_bundle_descriptor_reference=self._bundle_descriptor.id))
+            process_bundle_descriptor_reference=self._bundle_descriptor.id,
+            cache_tokens=[self._generate_cache_token()]))
```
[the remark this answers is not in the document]
mxm: Correct. The cache token should be regenerated per bundle repeat only. Have pushed an update.

sdks/python/apache_beam/runners/worker/sdk_worker.py
mxm: Changed, but kept it for

```python
data, continuation_token = \
          self._underlying.get_raw(state_key, continuation_token)
```

because

```python
data, continuation_token = (
          self._underlying.get_raw(state_key, continuation_token))
```

looks harder to understand.

sdks/python/apache_beam/runners/worker/sdk_worker.py
mxm: Thanks, much more concise.

sdks/python/apache_beam/runners/worker/sdk_worker.py
mxm: For the cache to work effectively, we want to cache the whole value. Note that if the cache is disabled, we still retain the old behavior of only materializing a page at a time.

Yes, it has the drawback of the value to fit into memory. The design currently doesn't do memory estimation but keeps a fixed upper limit of items. If we move to a memory estimation, then also reflecting this here would make sense.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
mxm: This is the maximum number of items to hold in the cache. I agree it would be nice to have this size-based. Though this will also require a more sophisticated memory model to determine the correct memory size. We could go with something reasonable, e.g. 100mb or so.

sdks/python/apache_beam/runners/worker/statecache.py
mxm: Good catch. Have change it to expunge the cache in this case.

sdks/python/apache_beam/runners/worker/statecache.py
mxm: Thanks, that's better.

sdks/python/apache_beam/runners/worker/statecache.py
mxm: Correct, this would invalidate the cache which results in an unnecessary retrieval of the state from the Runner. We should instead just clear the value.

sdks/python/apache_beam/runners/worker/statecache.py
```diff
@@ -0,0 +1,105 @@
...
+
+class StateCache(object):
+  """ Cache for Beam state access, scoped by state key and cache_token.
+
+  For a given state_key, caches a (cache_token, value) tuple and allows to
+    a) read from the cache,
+           if the currently stored cache_token matches the provided
+    a) write to the cache,
+           storing the new value alongside with a cache token
+    c) append to the cache cache,
```
rakeshcusat:
```suggestion
    c) append to the cache,
```

sdks/python/apache_beam/runners/portability/fn_api_runner.py
robertwb: Yeah, 100mb sounds like a reasonable value. It also makes more sense as a good default (who knows what 100 items is) and something possibly a user could specify. 

I don't think the memory model would have to be that sophisticated; the serialized size should be a reasonable approximation. It would require tracking it though. 

I would be OK with a JIRA and a TODO here.

sdks/python/apache_beam/runners/worker/sdk_worker.py
robertwb: The pagination is intended for when the value is too big to send, which probably means it's too big to cache (especially many of them given the current number-of-items bound). I don't think one would want to turn off all caching to avoid caching just these.

sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -620,6 +628,106 @@ def _next_id(self):
...
+    self._underlying = underlying_state
+    self._state_cache = global_state_cache
+    self._context = threading.local()
+
+  @contextlib.contextmanager
+  def process_instruction_id(self, bundle_id, cache_tokens):
+    if getattr(self._context, 'cache_token', None) is not None:
+      raise RuntimeError(
+          'Cache tokens already set to %s' % self._context.cache_token)
+    # TODO Also handle cache tokens for side input, if present
```
tweise: Create JIRA for it?

sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -331,7 +334,8 @@ def shutdown(self):
 
 class SdkWorker(object):
 
-  def __init__(self, bundle_processor_cache, profiler_factory=None):
+  def __init__(self, bundle_processor_cache,
+               profiler_factory=None):
```
tweise: unintended change?

sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -107,7 +110,7 @@ def run(self):
       # SdkHarness manage function registration and share self._fns with all
       # the workers. This is needed because function registration (register)
       # and exceution(process_bundle) are send over different request and we
```
tweise: "execution"

sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -51,7 +53,7 @@ class SdkHarness(object):
 
   def __init__(
       self, control_address, worker_count, credentials=None, worker_id=None,
-      profiler_factory=None):
+      state_cache_size=100, profiler_factory=None):
```
tweise: Does this mean we can only cache 100 keys? That may be a bit little, but probably it is best left to the user. How about making the default 0 and have that turn the cross bundle caching off?

tweise: The cache size setting would need to be passed to the entry point (`apache_beam.runners.worker.sdk_worker_main`) as pipeline option, so that in the end-to-end scenario the user can specify it. https://github.com/apache/beam/blob/5b43602109648830522159221a283754ed959afa/sdks/python/apache_beam/runners/worker/sdk_worker_main.py#L121

sdks/python/apache_beam/runners/worker/sdk_worker.py
mxm: Will do!

sdks/python/apache_beam/runners/worker/sdk_worker.py
mxm: Intended (separation of the kwargs from the positional args).

sdks/python/apache_beam/runners/worker/sdk_worker.py
mxm:
```suggestion
      # and execution (process_bundle) are send over different request and we
```
Also space missing.

sdks/python/apache_beam/runners/worker/sdk_worker.py
mxm: >Does this mean we can only cache 100 keys? 

Yes.

>How about making the default 0 and have that turn the cross bundle caching off?

Was thinking about this as well. Seems fair to default to 0 in the beginning but IMHO we should change the default to cached after experimentation. Would still keep caching enabled for the tests, except for the test suite which explicitly has caching disabled.

>The cache size setting would need to be passed to the entry point as pipeline option

You can read my mind. Just added this in another commit :)

sdks/python/apache_beam/runners/worker/sdk_worker_main.py
```diff
@@ -149,6 +149,7 @@ def main(unused_argv):
         control_address=service_descriptor.url,
         worker_count=_get_worker_count(sdk_pipeline_options),
         worker_id=_worker_id,
+        state_cache_size=_get_state_cache_size(sdk_pipeline_options),
```
tweise: It might be better to pass the pipeline options to SDKHarness and have it extract what it needs. There is currently quite some noise wherever SdkHarness is constructed in general.

sdks/python/apache_beam/runners/worker/worker_pool_main.py
```diff
@@ -47,21 +47,26 @@
...
+            state_cache_size=0, container_executable=None):
     worker_server = grpc.server(futures.ThreadPoolExecutor(max_workers=10))
     worker_address = 'localhost:%s' % worker_server.add_insecure_port(
         '[::]:%s' % port)
-    worker_pool = cls(worker_threads, use_process=use_process,
-                      container_executable=container_executable)
+    worker_pool = cls(worker_threads,
+                      use_process=use_process,
+                      container_executable=container_executable,
+                      state_cache_size=state_cache_size)
```
tweise: This should not occur here but come from the pipeline options from the provisioning endpoint.

mxm: We need this here as well because this class is also used for testing. It will be overridden for the non-testing case.

mxm:
Run Python PreCommit

sdks/python/apache_beam/runners/worker/sdk_worker.py
mxm: https://issues.apache.org/jira/browse/BEAM-8297

sdks/python/apache_beam/runners/portability/fn_api_runner.py
mxm: Jira: https://issues.apache.org/jira/browse/BEAM-8297

sdks/python/apache_beam/runners/worker/sdk_worker.py
mxm: https://issues.apache.org/jira/browse/BEAM-8298

sdks/python/apache_beam/runners/portability/fn_api_runner.py
mxm: Also added in in the code.

sdks/python/apache_beam/runners/worker/sdk_worker.py
mxm: I think caching only the first page makes sense. From looking at the Runner code, it looks like continuation tokens are not used currently. It seems like it is fair to defer this until they are used; also considering that the state caching is disabled by default.

mxm:
Benchmark with a synthetic source with parallelism of 1 on my local machine:

```python
max_items = 100000
num_keys = 10

class GenerateInput(beam.DoFn):

  def process(self, byte_array):
    key_string = byte_array
    key_int = int(key_string)
    yield (key_int % num_keys, "value" + key_string)

class StatefulProcessing(beam.DoFn):
  count_state_spec = userstate.CombiningValueStateSpec(
    'count', beam.coders.IterableCoder(beam.coders.VarIntCoder()), sum)

  def process(self, kv, count=beam.DoFn.StateParam(count_state_spec)):
    k, v = kv
    count.add(len(v))

(p
 | "Generate data" >> FlinkStreamingImpulseSource().set_message_count(max_items).set_interval_ms(0).with_output_types(bytes)
 | "Reshuffle" >> beam.Reshuffle()
 | "Format Data" >> beam.ParDo(GenerateInput()).with_output_types(KV[int, str])
 | "Stateful Processing" >> beam.ParDo(StatefulProcessing())
)

```

```
# elements: 100,000 
# bundle size: 10
# number of unique keys: 10
Took 205.37934804  with experiments=state_cache_size=0
Took 129.933783054 with experiments=state_cache_size=10
```

The version with caching is about 1,59 times faster. These results are consistent across multiple runs. This is of course a bit of an extreme situation in terms of the number of unique keys, but it shows the potential for the caching.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -1412,11 +1470,13 @@ def stop_worker(self):
...
 
 class WorkerHandlerManager(object):
-  def __init__(self, environments, job_provision_info):
+  def __init__(self, environments, job_provision_info, state_cache_size):
     self._environments = environments
     self._job_provision_info = job_provision_info
     self._cached_handlers = collections.defaultdict(list)
-    self._state = FnApiRunner.StateServicer() # rename?
+    self._state = sdk_worker.CachingMaterializingStateHandler(
+        StateCache(state_cache_size),
```
tweise: For my learning, why is the cache required inside the fn_api_runner vs. on the SDK side?

sdks/python/apache_beam/runners/portability/portable_runner_test.py
```diff
@@ -201,7 +204,8 @@ class PortableRunnerTestWithExternalEnv(PortableRunnerTest):
   @classmethod
   def setUpClass(cls):
     cls._worker_address, cls._worker_server = (
-        worker_pool_main.BeamFnExternalWorkerPoolServicer.start())
+        worker_pool_main.BeamFnExternalWorkerPoolServicer.start(
```
tweise: Logically this is duplicates the experiment flag above. I can see why that is currently necessary, but maybe it would be better to provide the pipeline options to the worker pool servicer? (Inside the container this would happen via provisioning endpoint and environment.)

sdks/python/apache_beam/runners/worker/bundle_processor.py
```diff
@@ -199,26 +199,19 @@ def finish(self):
...
-    while True:
-      input_stream = coder_impl.create_InputStream(data)
-      while input_stream.size() > 0:
-        yield self._coder_impl.decode_from_stream(input_stream, True)
-      if not continuation_token:
-        break
-      else:
-        data, continuation_token = self._state_handler.blocking_get(
-            self._state_key, continuation_token)
+    return self._state_handler.blocking_get(
```
tweise: Is there a JIRA for the continuation token that can be referenced here?

sdks/python/apache_beam/runners/worker/bundle_processor.py
```diff
@@ -430,13 +424,13 @@ def clear(self):
 
   def _commit(self):
     if self._cleared:
-      self._state_handler.blocking_clear(self._state_key)
+      self._state_handler.clear(self._state_key, is_cached=True).get()
```
tweise: Is this the call where we don't need to wait for a response?

sdks/python/apache_beam/runners/worker/bundle_processor.py
```diff
@@ -430,13 +424,13 @@ def clear(self):
...
     if self._added_elements:
-      value_coder_impl = self._value_coder.get_impl()
-      out = coder_impl.create_OutputStream()
-      for element in self._added_elements:
-        value_coder_impl.encode_to_stream(element, out, True)
-      self._state_handler.blocking_append(self._state_key, out.get())
+      self._state_handler.append(
```
tweise: An explicit comment regarding the need for blocking call might be useful here.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
mxm: Good catch. This does not seem to be necessary. Let me revisit.

sdks/python/apache_beam/runners/portability/portable_runner_test.py
mxm: Will look into this.

sdks/python/apache_beam/runners/worker/bundle_processor.py
mxm: Could you elaborate? The continuation token is respected as before.

sdks/python/apache_beam/runners/worker/sdk_worker_main.py
```diff
@@ -205,6 +206,28 @@ def _get_worker_count(pipeline_options):
...
+  experiments = pipeline_options.view_as(DebugOptions).experiments
+  experiments = experiments if experiments else []
+
+  for experiment in experiments:
+    # There should only be 1 match so returning from the loop
+    if re.match(r'state_cache_size=', experiment):
+      return int(
+          re.match(r'state_cache_size=(?P<state_cache_size>.*)',
+                   experiment).group('state_cache_size'))
+  return 100
```
tweise: Should be 0 (the default)?

sdks/python/apache_beam/runners/worker/bundle_processor.py
mxm: Yes, and no. We need to wait for the last response here. For example, if there is no following append, we'll have to wait on the clear. So best to safe the last returned future in a variable and call `.get()`on it here.

sdks/python/apache_beam/runners/worker/sdk_worker_main.py
mxm: Good catch, this was actually pending in my branch.

sdks/python/apache_beam/runners/worker/bundle_processor.py
mxm: Will consolidate with the above.

sdks/python/apache_beam/runners/worker/statecache.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def put(self, key, value):
+      self._cache[key] = value
+      while len(self._cache) > self._max_entries:
+        self._cache.popitem(last=False)
+
+    def evict(self, key):
+      self._cache.pop(key, self._default_entry)
+
+    def evict_all(self):
```
tweise: nit: would clear be the better name for this?

sdks/python/apache_beam/runners/portability/fn_api_runner.py
mxm: I added this because the WorkerHandlerManager will insert the state handler into the WorkerHandlerFactory, which generates a cached BundleProcessorCache with that state handler for the EmbeddedWorkerHandler. It is not necessary otherwise and just does unnecessary caching on the FnApiRunner side.

sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -620,6 +632,107 @@ def _next_id(self):
...
+  def get_raw(self, state_key, continuation_token=None):
+    return self._underlying.get_raw(state_key, continuation_token)
+
+  def append_raw(self, state_key, data):
+    return self._underlying.append_raw(state_key, data)
+
+  def restore(self):
+    self._underlying.restore()
+
+  def checkpoint(self):
```
mxm: Note, these and the above until the comment will go away (fix will come tomorrow), as they won't be necessary with the cache not being inserted for the state handler of the fn_api_runner.

sdks/python/apache_beam/runners/portability/fn_api_runner.py
mxm: Fixed now.

sdks/python/apache_beam/runners/worker/statecache.py
mxm: Maybe, but only if we also change `evict(key)`to `clear` as well. IMHO evict suits for a cache.

sdks/python/apache_beam/runners/portability/portable_runner_test.py
mxm: Since this affects multiple parameters of the `start` method (e.g. `worker_threads`), I would handle this in a follow-up. It looks like either way, via individual parameters or via pipeline options here, there will be duplicate code paths, due to this method being used in a test context and for the main entry point in `if __name__ == '__main__'`. To fix this it would require restructuring of the entire file, which I see out of scope for this PR. But good to think about doing that eventually.

mxm:
Rebased to master.

mxm:

sdks/python/apache_beam/runners/worker/bundle_processor.py
tweise: For the continuation token logic that was removed, what is the replacement?

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -82,6 +84,11 @@
     beam.coders.coders.GlobalWindowCoder()).get_impl().encode_nested(
         beam.transforms.window.GlobalWindows.windowed_value(b''))
 
+# State caching is enabled in fn fn_api_runner for testing, except for one
```
tweise:
```suggestion
# State caching is enabled in the fn_api_runner for testing, except for one
```

sdks/python/apache_beam/runners/portability/fn_api_runner.py
```diff
@@ -82,6 +84,11 @@
...
+# test which runs without state caching (FnApiRunnerTestWithDisabledCaching).
+# The Cache is disabled in production for other Runners.
```
tweise:
```suggestion
# The cache is disabled in production for other runners.
```

sdks/python/apache_beam/runners/worker/bundle_processor.py
```diff
@@ -370,17 +365,20 @@ def clear(self):
...
     if self._cleared:
-      self._state_handler.blocking_clear(self._state_key)
+      to_await = self._state_handler.clear(self._state_key, is_cached=True)
     if self._added_elements:
-      value_coder_impl = self._value_coder.get_impl()
-      out = coder_impl.create_OutputStream()
-      for element in self._added_elements:
-        value_coder_impl.encode_to_stream(element, out, True)
-      self._state_handler.blocking_append(self._state_key, out.get())
+      to_await = self._state_handler.append(
```
robertwb: Just a thought: I wonder if append should be renamed extend (analogous to list.append vs list.extend) as we're concatenating multiple elements here now.

sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -620,6 +632,93 @@ def _next_id(self):
...
+    if not self._should_be_cached(is_cached):
+      # no cache / tokens, can't do a lookup/store in the cache
+      return self._materialize_iter(state_key, coder)
+    # Cache lookup
+    cache_state_key = self._convert_to_cache_key(state_key)
+    cached_value = self._state_cache.get(cache_state_key,
+                                         self._context.cache_token)
+    if cached_value is None:
+      # Cache miss, need to retrieve from the Runner
+      materialized = cached_value = list(
```
robertwb: This will break if the value is too large to fit into memory. Of particular concern is the case of a GBK result that doesn't fit into memory and is fetched via the state API, but very large bag states are not that uncommon either.

sdks/python/apache_beam/runners/worker/bundle_processor.py
mxm: There was no continuation token logic removed. Robert raised the point that the state is materialized no matter how many continuation tokens are available (only if Caching is enabled). This can result in too large state being loaded. However, neither is the continuation token used by the Runner currently, nor is this likely to happen for user state, rather for side inputs which are currently not cached.

tweise: The code that is removed keeps fetching the state until there is no token. My question was what the replacement for this is? If you are saying that there was no need for it in first place because there is no support on the runner side, then that's OK. It might be appropriate to capture this and the question raised by Robert in a comment and/or link to the appropriate JIRA.

mxm: >The code that is removed keeps fetching the state until there is no token.

This code has _not_ been removed. It has simply been refactored to support fetching all state at once instead of just one element at a time. The logic regarding the continuation token is unchanged. Please look inside the `materialize_iter` method.

Of course I will address all remaining comments before merging the PR, as soon as I get a chance.

sdks/python/apache_beam/runners/worker/sdk_worker.py
mxm: Without a size optimization, I don't think there is much we can do at the moment. For large state, the continuation token could be used, but it is not utilized by the Runner for now.

Considering that state caching is disabled for now and the previous state materialization strategy (lazily fetch one item at a time) has not been altered, it seems fair to address this in a follow-up and mark this a TODO for now.

sdks/python/apache_beam/runners/worker/bundle_processor.py
mxm: Good suggestion, although I think we have been adding multiple elements before.

mxm: Done.

sdks/python/apache_beam/runners/worker/bundle_processor.py
mxm: Added a comments to further clarify.

mxm:
Latest changes here: https://github.com/apache/beam/compare/40cb35361234d12bd4ff8f8126aa5128bc07fcba..55588e91ed8e3e25bb661a6202c31e99297e0e79

From my perspective, all comments are addressed now.

mxm:
Run Portable_Python PreCommit

mxm:
Py37 somehow broken (also showing up in other builds):

```
18:34:00 FAILURE: Build completed with 2 failures.
18:34:00 
18:34:00 1: Task failed with an exception.
18:34:00 -----------
18:34:00 * What went wrong:
18:34:00 Execution failed for task ':sdks:python:test-suites:tox:py37:testPy37Gcp'.
18:34:00 > Process 'command 'sh'' finished with non-zero exit value 1
18:34:00 
18:34:00 * Try:
18:34:00 Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
18:34:00 ==============================================================================
18:34:00 
18:34:00 2: Task failed with an exception.
18:34:00 -----------
18:34:00 * Where:
18:34:00 Build file '/home/jenkins/jenkins-slave/workspace/beam_PreCommit_Python_Commit/src/sdks/python/test-suites/dataflow/py37/build.gradle' line: 66
18:34:00 
18:34:00 * What went wrong:
18:34:00 Execution failed for task ':sdks:python:test-suites:dataflow:py37:preCommitIT'.
18:34:00 > Process 'command 'sh'' finished with non-zero exit value 1
18:34:00 
```

mxm:
Run Python PreCommit

mxm:
Run Portable_Python PreCommit

mxm:
Python tests broken on master, see https://builds.apache.org/job/beam_PreCommit_Python_Cron/ and https://builds.apache.org/job/beam_PreCommit_Portable_Python_Cron/

mxm:
Run Portable_Python PreCommit

mxm:
Run Python PreCommit

sdks/python/apache_beam/runners/worker/sdk_worker.py
mxm: Also reflected this in the comment here with a JIRA link. Just to restate, the materialization logic with disabled caching (default) is unaltered here.

PR #9451, opened 2019-08-29
[BEAM-8113] Stage files from context classloader

je-ik:
Fixes [BEAM-8113].

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/PipelineResources.java
+++ b/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/PipelineResources.java
@@ -65,16 +65,16 @@ public class PipelineResources {
    * @return A list of absolute paths to the resources the class loader uses.
    */
   @VisibleForTesting
-  static List<String> detectClassPathResourcesToStage(ClassLoader loader) {
+  static List<String> detectClassPathResourcesToStage(final ClassLoader loader) {
     Set<String> files = new HashSet<>();
     ClassLoader stoppingLoader = loader;
     ClassLoader currentLoader = ReflectHelpers.findClassLoader();
+    if (loader != null) {
+      files.addAll(extractResourcesFromClassLoader(loader));
+    }
     while (currentLoader != null && stoppingLoader != currentLoader) {
       files.addAll(extractResourcesFromClassLoader(currentLoader));
       currentLoader = currentLoader.getParent();
-    }
-    if (loader != null) {
-      files.addAll(extractResourcesFromClassLoader(loader));
     }
     if (files.isEmpty()) {
       throw new IllegalArgumentException("Unable to use ClassLoader to detect classpath elements.");
--- a/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/PipelineResources.java
+++ b/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/PipelineResources.java
@@ -33,6 +33,7 @@ import java.util.Set;
 import java.util.stream.Collectors;
 import org.apache.beam.sdk.util.ZipFiles;
 import org.apache.beam.sdk.util.common.ReflectHelpers;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.annotations.VisibleForTesting;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Preconditions;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Strings;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.hash.Funnels;
@@ -53,14 +54,28 @@ public class PipelineResources {
    * @return A list of absolute paths to the resources the class loader uses.
    */
   public static List<String> detectClassPathResourcesToStage(Class<?> cls) {
+    return detectClassPathResourcesToStage(cls.getClassLoader());
+  }
+
+  /**
+   * Detect resources to stage, by using hierarchy between current context classloader and the given
+   * one. Always include the given class loader in the process.
+   *
+   * @param loader the loader to use as stopping condition when traversing context class loaders
+   * @return A list of absolute paths to the resources the class loader uses.
+   */
+  @VisibleForTesting
+  static List<String> detectClassPathResourcesToStage(ClassLoader loader) {
     Set<String> files = new HashSet<>();
-    ClassLoader stoppingLoader = cls.getClassLoader();
+    ClassLoader stoppingLoader = loader;
     ClassLoader currentLoader = ReflectHelpers.findClassLoader();
     while (currentLoader != null && stoppingLoader != currentLoader) {
       files.addAll(extractResourcesFromClassLoader(currentLoader));
       currentLoader = currentLoader.getParent();
     }
-    files.addAll(extractResourcesFromClassLoader(currentLoader));
+    if (loader != null) {
+      files.addAll(extractResourcesFromClassLoader(loader));
+    }
     if (files.isEmpty()) {
       throw new IllegalArgumentException("Unable to use ClassLoader to detect classpath elements.");
     }
--- a/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/PipelineResources.java
+++ b/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/PipelineResources.java
@@ -27,9 +27,12 @@ import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import java.util.stream.Collectors;
 import org.apache.beam.sdk.util.ZipFiles;
+import org.apache.beam.sdk.util.common.ReflectHelpers;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Preconditions;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Strings;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.hash.Funnels;
@@ -40,34 +43,28 @@ import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.hash.Hashing;
 public class PipelineResources {
 
   /**
-   * Attempts to detect all the resources the class loader has access to. This does not recurse to
-   * class loader parents stopping it from pulling in resources from the system class loader.
+   * Detects all URLs that are present in all class loaders in between context class loader of
+   * calling thread and class loader of class passed as parameter. It doesn't follow parents above
+   * this class loader stopping it from pulling in resources from the system class loader.
    *
-   * @param classLoader The URLClassLoader to use to detect resources to stage.
-   * @throws IllegalArgumentException If either the class loader is not a URLClassLoader or one of
-   *     the resources the class loader exposes is not a file resource.
+   * @param cls Class whose class loader stops recursion into parent loaders
+   * @throws IllegalArgumentException no classloader in context hierarchy is URLClassloader or if
+   *     one of the resources any class loader exposes is not a file resource.
    * @return A list of absolute paths to the resources the class loader uses.
    */
-  public static List<String> detectClassPathResourcesToStage(ClassLoader classLoader) {
-    if (!(classLoader instanceof URLClassLoader)) {
-      String message =
-          String.format(
-              "Unable to use ClassLoader to detect classpath elements. "
-                  + "Current ClassLoader is %s, only URLClassLoaders are supported.",
-              classLoader);
-      throw new IllegalArgumentException(message);
+  public static List<String> detectClassPathResourcesToStage(Class<?> cls) {
+    Set<String> files = new HashSet<>();
+    ClassLoader stoppingLoader = cls.getClassLoader();
+    ClassLoader currentLoader = ReflectHelpers.findClassLoader();
+    while (currentLoader != null && stoppingLoader != currentLoader) {
+      files.addAll(extractResourcesFromClassLoader(currentLoader));
+      currentLoader = currentLoader.getParent();
     }
-
-    List<String> files = new ArrayList<>();
-    for (URL url : ((URLClassLoader) classLoader).getURLs()) {
-      try {
-        files.add(new File(url.toURI()).getAbsolutePath());
-      } catch (IllegalArgumentException | URISyntaxException e) {
-        String message = String.format("Unable to convert url (%s) to file.", url);
-        throw new IllegalArgumentException(message, e);
-      }
+    files.addAll(extractResourcesFromClassLoader(currentLoader));
+    if (files.isEmpty()) {
+      throw new IllegalArgumentException("Unable to use ClassLoader to detect classpath elements.");
     }
-    return files;
+    return new ArrayList<>(files);
   }
 
   /**
@@ -92,6 +89,21 @@ public class PipelineResources {
                   : file.getAbsolutePath();
             })
         .collect(Collectors.toList());
+  }
+
+  private static List<String> extractResourcesFromClassLoader(ClassLoader loader) {
+    List<String> files = new ArrayList<>();
+    if (loader instanceof URLClassLoader) {
+      for (URL url : ((URLClassLoader) loader).getURLs()) {
+        try {
+          files.add(new File(url.toURI()).getAbsolutePath());
+        } catch (IllegalArgumentException | URISyntaxException e) {
+          String message = String.format("Unable to convert url (%s) to file.", url);
+          throw new IllegalArgumentException(message, e);
+        }
+      }
+    }
+    return files;
   }
 
   private static String packageDirectoriesToStage(File directoryToStage, String tmpJarLocation) {
--- a/runners/core-construction-java/src/test/java/org/apache/beam/runners/core/construction/PipelineResourcesTest.java
+++ b/runners/core-construction-java/src/test/java/org/apache/beam/runners/core/construction/PipelineResourcesTest.java
@@ -17,9 +17,9 @@
  */
 package org.apache.beam.runners.core.construction;
 
-import static junit.framework.TestCase.assertTrue;
-import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertThrows;
+import static org.junit.Assert.assertTrue;
 
 import java.io.File;
 import java.io.IOException;
@@ -28,7 +28,9 @@ import java.net.URLClassLoader;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
-import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.ImmutableList;
+import java.util.Optional;
+import org.junit.After;
+import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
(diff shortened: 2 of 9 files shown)


je-ik:
R: @mxm

je-ik:
Run Flink ValidatesRunner

je-ik:
Still debugging this locally. The fix seems not to be doing what I was expecting.

je-ik:
Run Portable_Python PreCommit

je-ik:
@mxm I think this can be merged. My intention was to enable the auto staging feature to work on JDK >= 9 (if user supplied URLClassLoader in context classloader). The problem is, unfortunately, that local flink runner does not use the context classloader (not sure why, still looking into that), but this PR seems to me to be valid on its own. If the user creates some context classloader before submitting the pipeline, we can do our best to retrieve as many dependencies as possible.

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkRunner.java
```diff
@@ -76,8 +80,15 @@ public static FlinkRunner fromOptions(PipelineOptions options) {
     }
 
     if (flinkOptions.getFilesToStage() == null) {
-      flinkOptions.setFilesToStage(
-          detectClassPathResourcesToStage(FlinkRunner.class.getClassLoader()));
+      List<String> filesToStage =
+          Stream.concat(
+                  Stream.of(FlinkRunner.class.getClassLoader()),
```
lukecwik: Please use `ReflectHelpers#findClassLoader`

je-ik: :+1:

je-ik:
The discussion in #8775 is interesting, but I think that it would need an extension. What I actually need is a way of propagating classes that are generated *at runtime*, and therefore cannot be part of any system variable (classes are generated inside a REPL shell). The general case is that application starts with some classloader, then creates another and sets it as context classloader. Now we have two issues:
 a) classes that are part of this new classloader have to be externalized (stored to jar), and
 b) this new jar have to be made available to *all possible runners*
The problem a) is not related to Beam and have to be handled by the application, but problem b) is actually more complicated, than it seems, because runners don't have any common way to ship dependencies (maybe this is changed in portability, I'm not sure). The only way, that seems to work is actually not to use the `filesToStage` pipeline option (because it might be ignored, and for instance `FlinkRunner` (local) ignores it), but to propagate this file to applications classloader (essentially by something like `Thread.currentThread().setContextClassLoader(new URLClassLoader(new URL[] { myNewJar }, Thread.currentThread().getContextClassLoader())`. This works well for both local and remote submission of a job (tested on flink and spark), but currently breaks for JDK >= 9.
This PR fixes that partly, flink local runner is unfortunately somehow broken and doesn't use context classloader to deserialize UDFs. There is a workaround (see [this hack](https://github.com/datadrivencz/proxima-platform/blob/363dee45e9c1edff3b6383b3fe13c463a0169f01/beam/tools/src/main/java/cz/o2/proxima/beam/tools/groovy/BeamStreamProvider.java#L264)), but it most probably needs fixing in flink directly. The hack is obviously not working for JDK >= 9 either.

je-ik:
Run Java PreCommit

je-ik:
I have localized the issue in flink: https://github.com/apache/flink/blob/ce557839d762b5f1ec92aa1885fd3d2ae33d0d0b/flink-runtime/src/main/java/org/apache/flink/runtime/execution/librarycache/BlobLibraryCacheManager.java#L280
It I use as parent classloader context classloader, I can build with JDK11 after applying this PR as well.

je-ik:
Hmm, looks like we should. Will do and then ping here again. :+1:

je-ik:
Fair points, thanks. I thought this will be a "small enough" change, so that it doesn't need any tests, but it is actually bigger. Regarding user specified classpath - true. But then I'd say we shouldn't auto stage nothing. Currently it already behaves the sometimes-things-work-sometimes-not-way. I'm fine with dropping the autostaging, but that would be a breaking change. The current state is not working on JDK9 and later, so we cannot keep it either.

That seems to give a conclusion, that this proposed change is "somewhat better than current state", although not perfect. So, if I will add tests (I'm a little tempted to put that into @ValidatesRunner, but I'm afraid, that it would be much more work, than I previously planned to invest into this :-)), will you then be fine with the proposed approach?

mxm:
@lukecwik Like you, I'm inclined to remove auto-staging entirely because it causes too much problems across different classloaders and java versions. As @je-ik pointed out, it would be a breaking change. IMHO the changes here do not fundamentally alter the current behavior, perhaps just make it execute correctly by traversing also the context classloader.

runners/flink/src/main/java/org/apache/beam/runners/flink/ClassLoaderUtils.java
```diff
@@ -0,0 +1,50 @@
...
+import static org.apache.beam.runners.core.construction.PipelineResources.detectClassPathResourcesToStage;
+
+import java.net.URLClassLoader;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.apache.beam.sdk.util.common.ReflectHelpers;
+
+class ClassLoaderUtils {
```
mxm: Why is this a Flink specific util?

runners/flink/src/main/java/org/apache/beam/runners/flink/ClassLoaderUtils.java
```diff
@@ -0,0 +1,50 @@
...
+
+  static List<String> detectFilesToStageFromClass(Class<?> cls) {
```
mxm: Should this go into `PipelineResources` and replace `detectFilesToStageFromClass`?

je-ik: I was thinking about that, but I was not 100% sure about the consequences. I will explore the contexts from which this call is made and put the code there, if possible.

je-ik:
Run Dataflow ValidatesRunner

je-ik:
Run Flink ValidatesRunner

je-ik:
Run Spark ValidatesRunner

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/PipelineResources.java
```diff
@@ -40,34 +43,28 @@
...
-    for (URL url : ((URLClassLoader) classLoader).getURLs()) {
-      try {
-        files.add(new File(url.toURI()).getAbsolutePath());
-      } catch (IllegalArgumentException | URISyntaxException e) {
-        String message = String.format("Unable to convert url (%s) to file.", url);
-        throw new IllegalArgumentException(message, e);
-      }
+    files.addAll(extractResourcesFromClassLoader(currentLoader));
+    if (files.isEmpty()) {
+      throw new IllegalArgumentException("Unable to use ClassLoader to detect classpath elements.");
```
mxm: Why do we throw an error here? It doesn't seem consistent because we skip over non UrlClassloader below which may also contain required files.

runners/core-construction-java/src/test/java/org/apache/beam/runners/core/construction/PipelineResourcesTest.java
```diff
@@ -44,25 +46,68 @@
...
 
   @Test
   public void detectClassPathResourcesWithUnsupportedClassLoader() {
     ClassLoader mockClassLoader = Mockito.mock(ClassLoader.class);
     thrown.expect(IllegalArgumentException.class);
     thrown.expectMessage("Unable to use ClassLoader to detect classpath elements.");
-
-    PipelineResources.detectClassPathResourcesToStage(mockClassLoader);
+    Thread.currentThread().setContextClassLoader(mockClassLoader);
+    PipelineResources.detectClassPathResourcesToStage(getClass());
```
mxm: Not sure why we throw here. According to the current logic, if there was one other UrlClassloader, we would not fail.

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkRunner.java
```diff
@@ -76,8 +76,7 @@ public static FlinkRunner fromOptions(PipelineOptions options) {
     }
 
     if (flinkOptions.getFilesToStage() == null) {
-      flinkOptions.setFilesToStage(
-          detectClassPathResourcesToStage(FlinkRunner.class.getClassLoader()));
+      flinkOptions.setFilesToStage(detectClassPathResourcesToStage(FlinkRunner.class));
```
mxm: This would step at FlinkRunner classloader, no? FlinkRunner is not part of the Flink cluster and needs to be shipped.

runners/core-construction-java/src/test/java/org/apache/beam/runners/core/construction/PipelineResourcesTest.java
```diff
@@ -44,25 +46,68 @@
...
+            new URL[] {file.toURI().toURL()}, Thread.currentThread().getContextClassLoader());
+    ClassLoader classLoader2 = new ClassLoader(classLoader1) {};
+
+    URLClassLoader classLoader3 =
+        new URLClassLoader(new URL[] {file2.toURI().toURL()}, classLoader2);
 
-    assertEquals(
-        ImmutableList.of(file.getAbsolutePath(), file2.getAbsolutePath()),
-        PipelineResources.detectClassPathResourcesToStage(classLoader));
+    Thread.currentThread().setContextClassLoader(classLoader3);
```
mxm: It looks like we do not test non-context classloaders?

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/PipelineResources.java
je-ik: Actually, there is a test in `PipelineResourcesTest`, that explicitly tests this behavior. And I think it is fine, because otherwise we would again introduce breaking change for JDK >= 9, because we would just silently skip all classloaders and users would get weird exceptions at pipeline runtime (although it have worked on JDK8). I think it is better to let users know about that in a way that is more understandable.

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkRunner.java
je-ik: No, the test explicitly stops at classloader of the given class (which is included, see [here](https://github.com/apache/beam/pull/9451/files#diff-c7786e41ca10361c918552b033e7cbffR63)).

je-ik:
> Thanks for the changes @je-ik. It looks like this is centred around context class loaders now. We should also support shipping required Beam classes from other class loaders. The way this is currently laid out, we would miss important classes, e.g. FlinkRunner which the remote Flink cluster does not have.
Yes, agree that we need to ship these files. And they will be shipped, as long as the classloader that loaded `FlinkRunner` is parent of current context classloader. I agree that that might not be the case for 100% of cases (although it seems to be weird, the JVM would be somewhat broken I'd say). Nevertheless, I'll fix that.

je-ik:
Run Dataflow ValidatesRunner

je-ik:
Run Flink ValidatesRunner

je-ik:
Run Spark ValidatesRunner

je-ik:
@mxm added the possibly non-context class loader. See 5140875.

runners/core-construction-java/src/test/java/org/apache/beam/runners/core/construction/PipelineResourcesTest.java
je-ik: Fixed.

runners/core-construction-java/src/test/java/org/apache/beam/runners/core/construction/PipelineResourcesTest.java
je-ik: Same reasoning as above.

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/PipelineResources.java
```diff
@@ -40,34 +44,42 @@
...
-      }
+  /**
+   * Detect resources to stage, by using hierarchy between current context classloader and the given
+   * one. Always include the given class loader in the process.
+   *
+   * @param loader the loader to use as stopping condition when traversing context class loaders
+   * @return A list of absolute paths to the resources the class loader uses.
+   */
+  @VisibleForTesting
+  static List<String> detectClassPathResourcesToStage(ClassLoader loader) {
```
mxm: Would make `loader` final.

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/PipelineResources.java
```diff
@@ -40,34 +44,42 @@
...
+    ClassLoader stoppingLoader = loader;
+    ClassLoader currentLoader = ReflectHelpers.findClassLoader();
+    while (currentLoader != null && stoppingLoader != currentLoader) {
+      files.addAll(extractResourcesFromClassLoader(currentLoader));
+      currentLoader = currentLoader.getParent();
     }
-    return files;
+    if (loader != null) {
+      files.addAll(extractResourcesFromClassLoader(loader));
+    }
```
mxm: This block should be moved to before the loop.

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/PipelineResources.java
mxm: Simply because there are several class loader variables and we want to make sure to include this one.

je-ik: :+1:

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/PipelineResources.java
je-ik: Makes sense. :+1:

je-ik:
Run Flink ValidatesRunner

je-ik:
@lukecwik can I take your last comment as that you are okay with this PR? I'd like to merge it tomorrow, so that I don't have to rebase it again and again. :-)

lukecwik:
I have to take another look through the comments/updates.

je-ik:
@lukecwik sure ... the related code goes apparently through kinda rapid development, as there are more and more uses of the modified method in core. Essentially, that was a sort of why I wanted to keep that in flink runner code only. But if we want to solve the issues for all runners, then it should be placed in core, but that will probably make this PR more complicated to merge.

runners/core-construction-java/src/test/java/org/apache/beam/runners/core/construction/PipelineResourcesTest.java
```diff
@@ -44,24 +49,98 @@
   @Rule public transient TemporaryFolder tmpFolder = new TemporaryFolder();
   @Rule public transient ExpectedException thrown = ExpectedException.none();
 
+  ClassLoader currentContext = null;
```
lukecwik: nit: Consider making this a test rule like [RestoreSystemProperties](https://github.com/apache/beam/blob/master/sdks/java/core/src/test/java/org/apache/beam/sdk/testing/RestoreSystemProperties.java)

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/PipelineResources.java
```diff
@@ -40,34 +44,42 @@
...
+  /**
+   * Detect resources to stage, by using hierarchy between current context classloader and the given
+   * one. Always include the given class loader in the process.
+   *
+   * @param loader the loader to use as stopping condition when traversing context class loaders
+   * @return A list of absolute paths to the resources the class loader uses.
+   */
+  @VisibleForTesting
+  static List<String> detectClassPathResourcesToStage(final ClassLoader loader) {
+    Set<String> files = new HashSet<>();
```
lukecwik: Making this a set even when there is a single classloader means that we will reorder that classloaders classpath which will cause lots of dependency conflict issues.

mxm: You are right, this is bound to cause trouble. We need to preserve the class loader hierarchy.

je-ik: I can change that to `LinkedHashSet`. But that doesn't solve it entirely, because there is still question in what order to append new files from class loaders. Parent-first? Child-first? What if the class loader of class passed as parameter is not in parent hierarchy of context class loader? Don't see how to define this correctly.

je-ik:
I cannot say I don't agree with both of you. I totally do. Let me recap the current state:
 - we extract jar from classloader of some arbitrarily picked class
 - we assume it is URLClassLoader
 - if user creates any user-supplied class loader and passes that as context class loader, we don't extract classes from that

That has several consequences:
 a) it fails on JDK >= 9
 b) it stages some arbitrary subset of possible jar that exist on class path, even if more of them could be extracted

The best way to provide jars to use is actually to specify them by hand, but that
 - is not standardized among runners (it probably should be part of `PipelineOptions`, but currently is not)
 - even if we add some standard way to stage files ti `PipelineOptions`, some runners (typically local runners) will tend to ignore those (because they assume, that all classes are loaded or able to load when the pipeline is run)

One might argue, that it is wrongif local runners ignore these files, but there is currently no way to supply any jars to local flink for instance.

So, to conclude what actually was my intent here:
 - one can always take list of jars and create a context class loader like that
   ```java
    Thread.currentThread().setContextClassLoader(
        new URLClassLoader(new URL[] { /* my jars */}));
    Pipeline p = ...;
    p.run();
   ```
 - if we correctly stage files, that should work for both *all* distributed and *all* local runners (provided they are well behaved, which unfortunately flink is not, but that is [different issue](https://issues.apache.org/jira/browse/FLINK-13925))

I really don't think that serializing class loader hierarchy along with the DoFns is a solution, because even if it would be possible (which seems to me is not), then it would still be fragile and error prone.

I'd be very glad to hear about some other possibility how to do the staging robust enough to
 * work well for all runners (even runners that beam actually doesn't know about, because everybody can in theory create his own runner outside of beam repo)
 * work well on all JDKs
but I currently don't know any other way.

lukecwik:
> I cannot say I don't agree with both of you. I totally do. Let me recap the current state:
> 
> * we extract jar from classloader of some arbitrarily picked class
> * we assume it is URLClassLoader
the most common of classloaders before JDK 9

> * if user creates any user-supplied class loader and passes that as context class loader, we don't extract classes from that
supporting all uses cases was never the goal but supporting common ones could be.

> 
> That has several consequences:
> a) it fails on JDK >= 9
> b) it stages some arbitrary subset of possible jar that exist on class path, even if more of them could be extracted
> 
> The best way to provide jars to use is actually to specify them by hand, but that
> 
> * is not standardized among runners (it probably should be part of `PipelineOptions`, but currently is not)
> * even if we add some standard way to stage files ti `PipelineOptions`, some runners (typically local runners) will tend to ignore those (because they assume, that all classes are loaded or able to load when the pipeline is run)
> 
> One might argue, that it is wrongif local runners ignore these files, but there is currently no way to supply any jars to local flink for instance.
> 
> So, to conclude what actually was my intent here:
> 
> * one can always take list of jars and create a context class loader like that
>   ```java
>    Thread.currentThread().setContextClassLoader(
>        new URLClassLoader(new URL[] { /* my jars */}));
>    Pipeline p = ...;
>    p.run();
>   ```
> * if we correctly stage files, that should work for both _all_ distributed and _all_ local runners (provided they are well behaved, which unfortunately flink is not, but that is [different issue](https://issues.apache.org/jira/browse/FLINK-13925))
> 
> I really don't think that serializing class loader hierarchy along with the DoFns is a solution, because even if it would be possible (which seems to me is not), then it would still be fragile and error prone.
> 
I believe this is possible but not worth the effort since the majority of users have simple scenarios (one flat classpath). Advanced users will be able to modify the Beam Java container and then do whatever they want.

> I'd be very glad to hear about some other possibility how to do the staging robust enough to
> 
> * work well for all runners (even runners that beam actually doesn't know about, because everybody can in theory create his own runner outside of beam repo)
> * work well on all JDKs
>   but I currently don't know any other way.

Out of a change in this space, I was looking for anything that would improve the situation for all users by either handling a wider range of supported scenarios or by expanding how many runners honor a classpath the user can provide. On #8775, I suggested that we use a technique where we attempt to get the classpath from a UrlClassLoader and if that isn't available then fallback to using java.class.path system property. This would then enable the most common scenario where there users application runs in a flat classpath in JDK 8 and below and also JDK 9 and above.

je-ik:
I'm closing this, because it turns out, that even if merged as proposed, it will not solve what I was hoping for. I will propose a different approach.

PR #9452, opened 2019-08-29
[BEAM-7945] Allow runner to configure semi_persist_dir which is used …

sunjincheng121:
Currently "semi_persist_dir" is not configurable. This may become a problem in certain scenarios. For example, the default value of "semi_persist_dir" is "/tmp" (https://github.com/apache/beam/blob/master/sdks/python/container/boot.go#L48) in Python SDK harness. When the environment type is "PROCESS", the disk of "/tmp" may be filled up and unexpected issues will occur in production environment. 

So, This pull request makes the semi_persist_dir configurable through adding a new PipelineOption(RemoteEnvironmentOptions).The Pipeline option will be passed to the `DefaultJobBundleFactory` and then be used in each EnvironmentFactory(docker, process, external and embedded).

For details of the discussion can be found in [1].

[1] https://lists.apache.org/list.html?dev@beam.apache.org:lte=1M:%5BDISCUSS%5D%20Turn%20%60WindowedValue

--- | --- | --- | --- | ---

### DIFF
--- a/runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
+++ b/runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
@@ -92,18 +92,18 @@ public class DefaultJobBundleFactory implements JobBundleFactory {
   private final int environmentExpirationMillis;
 
   public static DefaultJobBundleFactory create(JobInfo jobInfo) {
+    PipelineOptions pipelineOptions =
+        PipelineOptionsTranslation.fromProto(jobInfo.pipelineOptions());
     Map<String, EnvironmentFactory.Provider> environmentFactoryProviderMap =
         ImmutableMap.of(
             BeamUrns.getUrn(StandardEnvironments.Environments.DOCKER),
-            new DockerEnvironmentFactory.Provider(
-                PipelineOptionsTranslation.fromProto(jobInfo.pipelineOptions())),
+            new DockerEnvironmentFactory.Provider(pipelineOptions),
             BeamUrns.getUrn(StandardEnvironments.Environments.PROCESS),
-            new ProcessEnvironmentFactory.Provider(),
+            new ProcessEnvironmentFactory.Provider(pipelineOptions),
             BeamUrns.getUrn(StandardEnvironments.Environments.EXTERNAL),
             new ExternalEnvironmentFactory.Provider(),
             Environments.ENVIRONMENT_EMBEDDED, // Non Public urn for testing.
-            new EmbeddedEnvironmentFactory.Provider(
-                PipelineOptionsTranslation.fromProto(jobInfo.pipelineOptions())));
+            new EmbeddedEnvironmentFactory.Provider(pipelineOptions));
     return new DefaultJobBundleFactory(jobInfo, environmentFactoryProviderMap);
   }
 
--- a/runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/environment/DockerEnvironmentFactory.java
+++ b/runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/environment/DockerEnvironmentFactory.java
@@ -39,6 +39,7 @@ import org.apache.beam.runners.fnexecution.provisioning.StaticGrpcProvisionServi
 import org.apache.beam.sdk.fn.IdGenerator;
 import org.apache.beam.sdk.options.ManualDockerEnvironmentOptions;
 import org.apache.beam.sdk.options.PipelineOptions;
+import org.apache.beam.sdk.options.RemoteEnvironmentOptions;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Preconditions;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.ImmutableList;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.net.HostAndPort;
@@ -62,7 +63,7 @@ public class DockerEnvironmentFactory implements EnvironmentFactory {
       GrpcFnServer<StaticGrpcProvisionService> provisioningServiceServer,
       ControlClientPool.Source clientSource,
       IdGenerator idGenerator,
-      boolean retainDockerContainer) {
+      PipelineOptions pipelineOptions) {
     return new DockerEnvironmentFactory(
         docker,
         controlServiceServer,
@@ -71,7 +72,7 @@ public class DockerEnvironmentFactory implements EnvironmentFactory {
         provisioningServiceServer,
         idGenerator,
         clientSource,
-        retainDockerContainer);
+        pipelineOptions);
   }
 
   private final DockerCommand docker;
@@ -81,7 +82,7 @@ public class DockerEnvironmentFactory implements EnvironmentFactory {
   private final GrpcFnServer<StaticGrpcProvisionService> provisioningServiceServer;
   private final IdGenerator idGenerator;
   private final ControlClientPool.Source clientSource;
-  private final boolean retainDockerContainer;
+  private final PipelineOptions pipelineOptions;
 
   private DockerEnvironmentFactory(
       DockerCommand docker,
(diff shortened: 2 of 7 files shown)


sunjincheng121:
I appreciate if you have time to look up the changes @robertwb @mxm :)

model/fn-execution/src/main/proto/beam_fn_api.proto
```diff
@@ -815,6 +815,7 @@ message StartWorkerRequest {
   org.apache.beam.model.pipeline.v1.ApiServiceDescriptor logging_endpoint = 3;
   org.apache.beam.model.pipeline.v1.ApiServiceDescriptor artifact_endpoint = 4;
   org.apache.beam.model.pipeline.v1.ApiServiceDescriptor provision_endpoint = 5;
+  string semi_persist_dir = 6;
```
mxm: Should this be dynamic or rather configured up front for the worker pool?

sunjincheng121: Great to have your suggestions. :)
Maybe I have not understood your idea. If we configure the dir for the whole pool, we 
may loose the flexibility that different jobs may set different semi_persist_dir for the workers in the same worker pool?

mxm: I'm not sure whether this flexibility is desired. I could imagine that the person who starts the worker pool does not want arbitrary persist directories, but rather rather a fixed one.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
```diff
@@ -92,18 +92,18 @@
   private final int environmentExpirationMillis;
 
   public static DefaultJobBundleFactory create(JobInfo jobInfo) {
+    PipelineOptions pipelineOption =
```
tweise: "pipelineOptions"

model/fn-execution/src/main/proto/beam_fn_api.proto
tweise: This should not be added here. Pipeline options are provided through the provisioning endpoint.

sunjincheng121: Oh, Yes, I see, this is useless changes. `pipeline_options` already defined in the `ProvisionInfo`.

sdks/java/core/src/test/java/org/apache/beam/sdk/options/RemoteEnvironmentOptionsTest.java
```diff
@@ -0,0 +1,37 @@
...
+/** Tests for {@link RemoteEnvironmentOptions}. */
+@RunWith(JUnit4.class)
+public class RemoteEnvironmentOptionsTest {
+
+  @Test
+  public void testSemiDirectory() {
+    RemoteEnvironmentOptions options = PipelineOptionsFactory.as(RemoteEnvironmentOptions.class);
+    String semiDir = "/ab/cd";
+    options.setSemiPersistDir(semiDir);
+    assertEquals(semiDir, options.getSemiPersistDir());
```
mxm: This should test the default.

sdks/java/core/src/main/java/org/apache/beam/sdk/options/RemoteEnvironmentOptions.java
```diff
@@ -0,0 +1,43 @@
...
+import org.apache.beam.sdk.annotations.Experimental;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.ImmutableList;
+
+/** Options that are used to control configuration of the remote environment. */
+@Experimental
+@Hidden
+public interface RemoteEnvironmentOptions extends PipelineOptions {
+
+  @Description("Local semi-persistent directory")
+  @Default.String("/tmp")
```
mxm: Should this be `System.getProperty('java.io.tmpdir')`?

sunjincheng121: Currently, we keep the same as the default value of other default configuration, such as:`boot.go`.
- https://github.com/apache/beam/blob/d21bbaf4c70986c2dbdbe8f6fce35b2b2cb4843d/sdks/go/container/boot.go#L41
- https://github.com/apache/beam/blob/d21bbaf4c70986c2dbdbe8f6fce35b2b2cb4843d/sdks/python/container/boot.go#L51

So, how about we keep using `/tmp` as default value ?

mxm: Let's keep the existing default.

sunjincheng121:
Thanks for the review @mxm !
I have update the PR according your comments. I appreciate if you can have a another look.

sdks/java/core/src/main/java/org/apache/beam/sdk/options/RemoteEnvironmentOptions.java
tweise: I think the default should be null, so that the environment can pick its suitable tmp directory when nothing is specified by the user.

sunjincheng121:
@mxm @tweise  Thanks a lot for the review. Sorry that I missed the comments from @tweise.
I have updated the PR, would be great if you can take another look.

sunjincheng121:
Run Java PreCommit

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/environment/DockerEnvironmentFactory.java
```diff
@@ -132,20 +133,30 @@ public RemoteEnvironment createEnvironment(Environment environment) throws Excep
...
+
+    String semiPersistDir = pipelineOptions.as(RemoteEnvironmentOptions.class).getSemiPersistDir();
+    ImmutableList.Builder<String> argsBuilder =
+        ImmutableList.<String>builder()
+            .add(String.format("--id=%s", workerId))
+            .add(String.format("--logging_endpoint=%s", loggingEndpoint))
+            .add(String.format("--artifact_endpoint=%s", artifactEndpoint))
+            .add(String.format("--provision_endpoint=%s", provisionEndpoint))
+            .add(String.format("--control_endpoint=%s", controlEndpoint));
+    if (semiPersistDir != null) {
```
tweise: Isn't this provided to the environment through the provision endpoint?

tweise: So we essentially pass the same piece of information to the worker twice: As entry point argument and then again within the pipeline options. It needs to be done this way due to the container contract. Would be nice to revisit in the future.

mxm: That's right. This is redundant here.

mxm: Actually, the semi_persist_dir is not inferred from the pipeline options in the bootloader code. Like you said, it has to be this way currently, but it would be nice to not duplicate this information in the future.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/environment/DockerEnvironmentFactory.java
```diff
@@ -132,20 +133,30 @@ public RemoteEnvironment createEnvironment(Environment environment) throws Excep
...
-        ImmutableList.of(
-            String.format("--id=%s", workerId),
-            String.format("--logging_endpoint=%s", loggingEndpoint),
-            String.format("--artifact_endpoint=%s", artifactEndpoint),
-            String.format("--provision_endpoint=%s", provisionEndpoint),
-            String.format("--control_endpoint=%s", controlEndpoint));
+    Boolean retainDockerContainer =
+        pipelineOptions.as(ManualDockerEnvironmentOptions.class).getRetainDockerContainers();
+    if (!retainDockerContainer) {
+      dockerOptsBuilder.add("--rm");
```
ibzib, line 139: Why was this added in this PR? It seems orthogonal to `semi_persist_dir`. I believe this is a regression; perhaps a rebasing error?

mxm: Indeed looks like a rebasing error. A bit tricky one to get right because we changed the way the container removal worked before the rebasing was done. We were using the `--rm` initially but before the rebase we changed it to remove the container explicitly via `docker remove <container_id>`.

mxm: Fix: https://github.com/apache/beam/pull/11303

PR #9462, opened 2019-08-31
Fix typos

KangZhiDong:
fix typos

### DIFF
--- a/runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/ShuffleSink.java
+++ b/runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/ShuffleSink.java
@@ -238,7 +238,7 @@ public class ShuffleSink<T> extends Sink<WindowedValue<T>> {
           // key by concatenating the nested encoding of sortKey and outer encoding of the
           // timestamp. An alternative implementation would be to use OrderedCode but it
           // is unnecessary here because the nested encoding of sortKey achieves the same effect,
-          // due to the nested encoding being a https://en.wikipedia.org/wiki/Prefix_code.
+          // due to the nested encoding being an https://en.wikipedia.org/wiki/Prefix_code.
           // Move forward enough bytes so we can prefix the size on after performing the write
           int initialChunkSize = chunk.size();
           chunk.resetTo(initialChunkSize + Ints.BYTES);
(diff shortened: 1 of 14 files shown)


runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/ShuffleSink.java
```diff
@@ -238,7 +238,7 @@ public long add(WindowedValue<T> windowedElem) throws IOException {
           // key by concatenating the nested encoding of sortKey and outer encoding of the
           // timestamp. An alternative implementation would be to use OrderedCode but it
           // is unnecessary here because the nested encoding of sortKey achieves the same effect,
-          // due to the nested encoding being a https://en.wikipedia.org/wiki/Prefix_code.
+          // due to the nested encoding being an https://en.wikipedia.org/wiki/Prefix_code.
```
RyanSkraba: Not sure about this one! :D

KangZhiDong: an https

RyanSkraba: I had to read it a few times because my brain automatically "pronounces" `an https://en.wikipedia.org/wiki/Prefix_code`  as `an Prefix Code` -- I would have no objection either leaving it or changing it!

KangZhiDong: O(∩_∩)O~

sdks/go/pkg/beam/core/typex/class_test.go
```diff
@@ -105,7 +105,7 @@ func TestClassOf(t *testing.T) {
 		{reflect.TypeOf(uintptr(0)), Invalid},          // uintptr
 		{reflect.TypeOf(func() {}), Invalid},           // function
 		{reflect.TypeOf(make(chan int)), Invalid},      // chan
-		{reflect.TypeOf(struct{ A error }{}), Invalid}, // public interface field
+		{reflect.TypeOf(struct{ An error }{}), Invalid}, // public interface field
```
ibzib: This one is code, not English, so it should probably have been excluded from linting.

KangZhiDong: this ‘A’ is a variable

runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/ShuffleSink.java
robertwb: I would leave it as "a" in this case.

ibzib:
Test failures are all unrelated.

PR #9479, opened 2019-09-05
Revert "[BEAM-7819] Python - parse PubSub message_id into attributes property"

aaltay:
Reverts apache/beam#9232 - Post commits are failing (https://issues.apache.org/jira/browse/BEAM-8153)


udim:
run python 2 postcommit

PR #8472, opened 2019-05-02
[BEAM-7126] Remove duplicate key encoding for state backend

mxm:
Unknown Coders are wrapped inside a LengthPrefixCoder to ship serialized data
without knowledge of the Coder used to serialize the data. When operators use
state or timers, we rely on data to be partitioned by key. We included the
LengthPrefixCoder for the key Coder which led to issues with handling state
requests which do not include the length prefix in the key encoding.

We can safely remove the LengthPrefixCoder from the key because we do not store
the key in a nested context where we might need the length prefix.

This also expands the state cleanup tests to further verify that state is
cleaned up correctly.

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

Lang | SDK | Apex | Dataflow | Flink | Gearpump | Samza | Spark
--- | --- | --- | --- | --- | --- | --- | ---
Go | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go/lastCompletedBuild/) | --- | --- | --- | --- | --- | ---
Java | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Apex/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Dataflow/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Flink/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Flink_Batch/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Flink_Streaming/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Gearpump/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Samza/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Spark/lastCompletedBuild/)
[...]

PR #9484, opened 2019-09-05
[BEAM-8157] Ensure key encoding for state requests is consistent across SDKs

mxm:
The key encoding for state requests across the SDKs, i.e. Python and Java, is
not consistent. Java uses a NESTED encoding, whereas Python uses OUTER. We
should settle on one encoding.

The Flink runner so far uses OUTER encoding, but it seems sensible to switch to
NESTED encoding since inferring the OUTER encoding for a key coder in a KV<K,V>
structure is a potential source of error.

--- | --- | --- | --- | ---

### DIFF
--- a/runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
+++ b/runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
@@ -261,7 +261,6 @@ public class ExecutableStageDoFnOperator<InputT, OutputT> extends DoFnOperator<I
         StateInternals stateInternals,
         KeyedStateBackend<ByteBuffer> keyedStateBackend,
         Lock stateBackendLock) {
-
       this.stateInternals = stateInternals;
       this.keyedStateBackend = keyedStateBackend;
       this.stateBackendLock = stateBackendLock;
@@ -344,11 +343,8 @@ public class ExecutableStageDoFnOperator<InputT, OutputT> extends DoFnOperator<I
         }
 
         private void prepareStateBackend(K key) {
-          // Key for state request is shipped already encoded as ByteString,
-          // this is mostly a wrapping with ByteBuffer. We still follow the
-          // usual key encoding procedure.
-          // final ByteBuffer encodedKey = FlinkKeyUtils.encodeKey(key, keyCoder);
-          final ByteBuffer encodedKey = ByteBuffer.wrap(key.toByteArray());
+          // Key for state request is shipped encoded with NESTED context.
+          ByteBuffer encodedKey = FlinkKeyUtils.fromEncodedKey(key);
           keyedStateBackend.setCurrentKey(encodedKey);
         }
       };
--- a/runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkStreamingPortablePipelineTranslator.java
+++ b/runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkStreamingPortablePipelineTranslator.java
@@ -68,7 +68,6 @@ import org.apache.beam.sdk.coders.ByteArrayCoder;
 import org.apache.beam.sdk.coders.Coder;
 import org.apache.beam.sdk.coders.IterableCoder;
 import org.apache.beam.sdk.coders.KvCoder;
-import org.apache.beam.sdk.coders.LengthPrefixCoder;
 import org.apache.beam.sdk.coders.VoidCoder;
 import org.apache.beam.sdk.io.UnboundedSource;
 import org.apache.beam.sdk.options.PipelineOptions;
@@ -676,11 +675,6 @@ public class FlinkStreamingPortablePipelineTranslator
                 valueCoder.getClass().getSimpleName()));
       }
       keyCoder = ((KvCoder) valueCoder).getKeyCoder();
-      if (keyCoder instanceof LengthPrefixCoder) {
-        // Remove any unnecessary length prefixes which add more payload
-        // but also are not expected for state requests inside the operator.
-        keyCoder = ((LengthPrefixCoder) keyCoder).getValueCoder();
-      }
       keySelector = new KvToByteBufferKeySelector(keyCoder);
       inputDataStream = inputDataStream.keyBy(keySelector);
     }
--- a/runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/FlinkKeyUtils.java
+++ b/runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/FlinkKeyUtils.java
@@ -32,6 +32,7 @@ import org.apache.beam.sdk.coders.ByteArrayCoder;
 import org.apache.beam.sdk.coders.Coder;
 import org.apache.beam.sdk.coders.StructuredCoder;
 import org.apache.beam.sdk.util.CoderUtils;
+import org.apache.beam.vendor.grpc.v1p21p0.com.google.protobuf.ByteString;
 
 /**
  * Utility functions for dealing with key encoding. Beam requires keys to be compared in binary
@@ -44,7 +45,7 @@ public class FlinkKeyUtils {
     checkNotNull(keyCoder, "Provided coder must not be null");
     final byte[] keyBytes;
     try {
-      keyBytes = CoderUtils.encodeToByteArray(keyCoder, key);
+      keyBytes = CoderUtils.encodeToByteArray(keyCoder, key, Coder.Context.NESTED);
     } catch (Exception e) {
       throw new RuntimeException(String.format(Locale.ENGLISH, "Failed to encode key: %s", key), e);
     }
@@ -52,20 +53,24 @@ public class FlinkKeyUtils {
   }
 
   /** Decodes a key from a ByteBuffer containing a byte array. */
-  static <K> K decodeKey(ByteBuffer byteBuffer, Coder<K> keyCoder) {
+  public static <K> K decodeKey(ByteBuffer byteBuffer, Coder<K> keyCoder) {
     checkNotNull(byteBuffer, "Provided ByteBuffer must not be null");
     checkNotNull(keyCoder, "Provided coder must not be null");
     checkState(byteBuffer.hasArray(), "ByteBuffer key must contain an array.");
     @SuppressWarnings("ByteBufferBackingArray")
     final byte[] keyBytes = byteBuffer.array();
     try {
-      return CoderUtils.decodeFromByteArray(keyCoder, keyBytes);
+      return CoderUtils.decodeFromByteArray(keyCoder, keyBytes, Coder.Context.NESTED);
     } catch (Exception e) {
       throw new RuntimeException(
           String.format(
               Locale.ENGLISH, "Failed to decode encoded key: %s", Arrays.toString(keyBytes)),
           e);
     }
+  }
+
+  static ByteBuffer fromEncodedKey(ByteString encodedKey) {
+    return ByteBuffer.wrap(encodedKey.toByteArray());
   }
 
   /** The Coder for the Runner's encoded representation of a key. */
--- a/runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/state/FlinkStateInternals.java
+++ b/runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/state/FlinkStateInternals.java
@@ -28,8 +28,8 @@ import org.apache.beam.runners.core.StateInternals;
 import org.apache.beam.runners.core.StateNamespace;
 import org.apache.beam.runners.core.StateTag;
 import org.apache.beam.runners.flink.translation.types.CoderTypeSerializer;
+import org.apache.beam.runners.flink.translation.wrappers.streaming.FlinkKeyUtils;
 import org.apache.beam.sdk.coders.Coder;
-import org.apache.beam.sdk.coders.CoderException;
 import org.apache.beam.sdk.coders.InstantCoder;
 import org.apache.beam.sdk.coders.VoidCoder;
 import org.apache.beam.sdk.state.BagState;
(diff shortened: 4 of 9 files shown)


mxm:
CC @sunjincheng121 @dmvk

mxm:
It might be worth to add a test for the state handler, to guard against changes in the future.

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
```diff
@@ -343,11 +343,11 @@ public void clear(K key, W window) {
         }
 
         private void prepareStateBackend(K key) {
-          // Key for state request is shipped already encoded as ByteString,
-          // this is mostly a wrapping with ByteBuffer. We still follow the
-          // usual key encoding procedure.
-          // final ByteBuffer encodedKey = FlinkKeyUtils.encodeKey(key, keyCoder);
-          final ByteBuffer encodedKey = ByteBuffer.wrap(key.toByteArray());
+          // Key for state request is shipped already encoded as ByteString, but it is
```
tweise: In which situation is it NESTED? We made the change originally because it wasn't prefixed, so trying to understand what changed.

mxm: It is always `NESTED` although it is not always apparent because not all coders use a length prefix. To correctly match the Flink Runner's key encoding, we need to remove the length prefix, i.e. convert `NESTED` to `OUTER` context.

tweise: Based on https://github.com/apache/beam/pull/8472/files there shouldn't be a length prefix?

mxm: Yes, but that is incorrect. It worked before because we were applying a double nested encoding.

mxm: I should say, the bug only manifests together with timers which access state, because there the key encoding will not match between the two.

mxm: The timer test should be adjusted accordingly to catch that.

mxm:
Test failures unrelated to changes in this PR: https://builds.apache.org/job/beam_PreCommit_Java_Commit/7598/

```
Test Result (2 failures / +2)

    org.apache.beam.sdk.transforms.ParDoLifecycleTest.testTeardownCalledAfterExceptionInProcessElementStateful
    org.apache.beam.sdk.transforms.ParDoLifecycleTest.testTeardownCalledAfterExceptionInStartBundle
```

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
tweise: There is something that I'm missing. State and timers work as of release 2.14 with Flink 1.8. What has caused it to break?

tweise: https://gist.github.com/tweise/e1d95369234a52b12f90794f6032f039

That's working on 2.14 and also on master as of 4460f03cefbde5cf66c797408439321de2f40692

mxm: Here is why this works for Flink <= 1.8 but definitely needs to be fixed moving forward:

Keys for state requests are encoded incorrectly for the Flink Runner's state backend because they may contain a length prefix, which is not in lines with how the key is encoded for data partitioning. This causes state to be written to the wrong key group. However, the encoding scheme is consistent, so reading from state in `ProcessElement` or `OnTimer` methods works correctly. Also when inter-playing with timers. However, there are potential issues with checkpoints/savepoints due to this.

In Flink 1.9, the key encoding mismatch is visible because reading state from a key not part of the partition _can_ return a `null` value. That is, not always, likely due to the state compaction logic. This explains why `PortableStateExecutionTest` works correctly but `PortableTimersExecutionTest` does not. That said, I think we should merge this fix ASAP for the upcoming release.

tweise: OK, thanks for the RCA. I will take this PR for a test drive and report back. (Just FYI no checkpointing issues with the current encoding on Flink 1.8).

tweise:
Unfortunately with this change the test pipeline fails:
```
RuntimeError: java.lang.RuntimeException: Failed to remove nested context from key: XXXXXX
	at org.apache.beam.runners.flink.translation.wrappers.streaming.FlinkKeyUtils.removeNestedContext(FlinkKeyUtils.java:85)
	at org.apache.beam.runners.flink.translation.wrappers.streaming.ExecutableStageDoFnOperator$BagUserStateFactory$1.prepareStateBackend(ExecutableStageDoFnOperator.java:370)
```

mxm:
Thanks for testing. I've pushed an update. It was wrong to assume we could simply strip the length prefix from the ByteBuffer, as it simply holds whatever bytes the original coder produced. We need to perform an encoding round-trip to recover the original key in `OUTER` encoding.

@tweise Could you please test again with the new update?

mxm:

mxm:
Updated the description and squashed the commit. After we give this another test, this should be good to be merged.

tweise:
This actually fails the test that I had provided in the gist.
```
RuntimeError: java.lang.IllegalArgumentException: Forbidden IOException when reading from InputStream
        at org.apache.beam.sdk.util.CoderUtils.decodeFromSafeStream(CoderUtils.java:118)
        at org.apache.beam.sdk.util.CoderUtils.decodeFromByteArray(CoderUtils.java:98)
        at org.apache.beam.runners.flink.translation.wrappers.streaming.FlinkKeyUtils.removeNestedContext(FlinkKeyUtils.java:81)
        at org.apache.beam.runners.flink.translation.wrappers.streaming.ExecutableStageDoFnOperator$BagUserStateFactory$1.prepareStateBackend(ExecutableStageDoFnOperator.java:351)
        at org.apache.beam.runners.flink.translation.wrappers.streaming.ExecutableStageDoFnOperator$BagUserStateFactory$1.get(ExecutableStageDoFnOperator.java:283)
        at org.apache.beam.runners.flink.translation.wrappers.streaming.ExecutableStageDoFnOperator$BagUserStateFactory$1.get(ExecutableStageDoFnOperator.java:278)
        at org.apache.beam.runners.fnexecution.state.StateRequestHandlers$ByteStringStateRequestHandlerToBagUserStateHandlerFactoryAdapter.handleGetRequest(StateRequestHandlers.java:435)
        at org.apache.beam.runners.fnexecution.state.StateRequestHandlers$ByteStringStateRequestHandlerToBagUserStateHandlerFactoryAdapter.handle(StateRequestHandlers.java:400)
        at org.apache.beam.runners.fnexecution.state.StateRequestHandlers$StateKeyTypeDelegatingStateRequestHandler.handle(StateRequestHandlers.java:205)
        at org.apache.beam.runners.fnexecution.state.GrpcStateService$Inbound.onNext(GrpcStateService.java:130)
        at org.apache.beam.runners.fnexecution.state.GrpcStateService$Inbound.onNext(GrpcStateService.java:118)
        at org.apache.beam.vendor.grpc.v1p21p0.io.grpc.stub.ServerCalls$StreamingServerCallHandler$StreamingServerCallListener.onMessage(ServerCalls.java:249)
        at org.apache.beam.vendor.grpc.v1p21p0.io.grpc.ForwardingServerCallListener.onMessage(ForwardingServerCallListener.java:33)
        at org.apache.beam.vendor.grpc.v1p21p0.io.grpc.Contexts$ContextualizedServerCallListener.onMessage(Contexts.java:76)
        at org.apache.beam.vendor.grpc.v1p21p0.io.grpc.internal.ServerCallImpl$ServerStreamListenerImpl.messagesAvailable(ServerCallImpl.java:297)
        at org.apache.beam.vendor.grpc.v1p21p0.io.grpc.internal.ServerImpl$JumpToApplicationThreadServerStreamListener$1MessagesAvailable.runInContext(ServerImpl.java:738)
        at org.apache.beam.vendor.grpc.v1p21p0.io.grpc.internal.ContextRunnable.run(ContextRunnable.java:37)
        at org.apache.beam.vendor.grpc.v1p21p0.io.grpc.internal.SerializingExecutor.run(SerializingExecutor.java:123)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
Caused by: java.io.EOFException: reached end of stream after reading 1 bytes; 107 bytes expected
        at org.apache.beam.vendor.guava.v26_0_jre.com.google.common.io.ByteStreams.readFully(ByteStreams.java:780)
        at org.apache.beam.vendor.guava.v26_0_jre.com.google.common.io.ByteStreams.readFully(ByteStreams.java:762)
        at org.apache.beam.sdk.coders.ByteArrayCoder.decode(ByteArrayCoder.java:108)
        at org.apache.beam.sdk.coders.ByteArrayCoder.decode(ByteArrayCoder.java:41)
        at org.apache.beam.sdk.util.CoderUtils.decodeFromSafeStream(CoderUtils.java:115)
```

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
```diff
@@ -343,11 +345,10 @@ public void clear(K key, W window) {
         }
 
         private void prepareStateBackend(K key) {
-          // Key for state request is shipped already encoded as ByteString,
-          // this is mostly a wrapping with ByteBuffer. We still follow the
-          // usual key encoding procedure.
-          // final ByteBuffer encodedKey = FlinkKeyUtils.encodeKey(key, keyCoder);
-          final ByteBuffer encodedKey = ByteBuffer.wrap(key.toByteArray());
+          // Key for state request is shipped encoded. It may be encoded with NESTED context which
```
tweise: Don't we need to know for sure it was encoded as NESTED to perform such re-encoding? FWIW the test fails because of the attempt to decode a non-nested (not length prefixed key). ` [107, 49]`

mxm:
@tweise Are you sure you tested with the most recent version of this PR? I had updated the PR. Sorry if that was not clear. We need the orginal coder used to removed the nested encoding, since the ByteStringCoder in the state handler does not provide any information of the key encoding used.

tweise:
I'm reasonably sure. But if you tell me that you tested your latest change with the pipeline in the gist then I can repeat it.

lukecwik:
The coder for the main input's key should match the coder being chosen for the state key.

e.g., if the main input coder is KV<LP<K>, LP<V>> then the state key coder should be LP<K> since encode(LP<K>) != encode(K) and even if the runner drops the length prefixing from encode(LP<K>)'s bytes, it still may not be equal to encode(K).

mxm:
Run Portable_Python PreCommit

mxm:
@lukecwik @tweise @robertwb PR is updated. Encoding for state keys was not consistent across the Java and Python SDK. This has been addressed here by letting them both use the `NESTED` context, i.e. adjusting the key encoding in the Python SDK.

mxm:
Any opinions here? I think this warrants some attention to finally fix the inconsistent key encoding across SDKs.

mxm:
>When the key encoding was changed prior to the Python SDK change, this PR was green, but my test failed. Are we missing (Flink runner) integration test coverage for state and timers for Python?

I had assumed that the Python SDK would do nested encoding for the key in StateRequest, just like Java. Turns out, it did not. That was the reason why after removing the nested context for Java (first version of the PR), your tests failed because it would try removing the nested context from the Python encoded key, which was encoded using outer context.

Now, the real problem is that the inconsistent key encoding is shadowed by Flink <=1.8 because it accepts keys for the state backend which do not belong to the partition key range. The Flink 1.9 (#9296) made it visible that the Java SDK was using the nested encoding, contrary to Python and the Runner itself, which were using the outer encoding. The reason the Python PVR tests in the Flink 1.9 PR were _not_ failing was that we are using parallelism 1, which I had advised against in the past (and I now have a better reason why). So we should increase the parallelism in a follow-up.

sunjincheng121:
Hi @mxm, Sorry for late reply. I go through the changes of this PR and found that the solution in this PR is the same as #9464. The functionality of Java side changes is the same as #9464. Regarding to the Python part, it's a good catch. I think it will be better to recover the changes of #9464 and add fixes for the Python part on top of it. I don't think it's a best way to revert the contributions in this case. Even we revert a commit in some reason, we should also recover it if we finally find the original solution is correct. This is a respect for the contributors and the committers.

What do you think?  @mxm @tweise @lukecwik 

Best, Jincheng

mxm:
I very much appreciate your work on this @sunjincheng121. Let it be said, your contributions are recognized. I also have to repeat that your solution in #9464 was incomplete. Not only did it not adhere to our contribution standard because it was originally without a JIRA ("hotfix"), but it also broke the Python support, as our tests showed. It required careful investigation to find out the root of the problem and to not break any other SDK as the result of a quick fix. We could not iterate on your PR because it was already merged, so reverting it was the only option.

The changes, while similar, are not identical. I could have based them on your original PR, but that would have still required to do more changes, since there are differences related to the code and the tests. I know that getting one's commit reverted is not a good feeling. It happens to everyone eventually, and it is not to diminish your contributors in any way. Keep up the great contributions. I know that everyone in the Beam community greatly appreciates your work.

sunjincheng121:
Thanks for your reply in both this PR and #9464 @mxm, I left my thoughts in #9464. (I suggest that we can discuss in one PR).

PR #9485, opened 2019-09-05
Make JRH logs less noisy

boyuanzz:
Downgrade warning log to debug log, to make it less noisy for streaming.
R: @lukecwik 
cc: @y1chi @angoenka


runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/fn/control/BeamFnMapTaskExecutor.java
```diff
@@ -365,20 +365,11 @@ void updateProgress() {
         if (!isTransientProgressError(exn.getMessage())) {
           grpcWriteOperationElementsProcessed.accept(-1); // Not supported.
           progressErrors++;
```
lukecwik: I was hoping that we would still see this error message occasionally to know that it is still happening.

Instead I would lower bound it so that we see the warning if there have been at least 100 errors and it is a power of 2.

boyuanzz:

boyuanzz:

boyuanzz:
PTAL @lukecwik  : )

boyuanzz:
Run Java PreCommit

PR #9487, opened 2019-09-05
[BEAM-8165]Change default docker images name

Hannah-Jiang:
As we decided to add `_sdk` to image name, this PR is changing default image name for Python, Java and Go images.

Publishing docker images as part of release is tackled with following three PRs.
1. Change default image name. (current PR, #9487)
2. Add staging and publishing scripts for docker images to release procedure. (#9506 )
3. Updating release_guide.md (#9510)

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/runners/portability/portable_runner.py
+++ b/sdks/python/apache_beam/runners/portability/portable_runner.py
@@ -84,13 +84,12 @@ class PortableRunner(runner.PipelineRunner):
                       'has Python %d.%d interpreter.' % (
                           sys.version_info[0], sys.version_info[1]))
 
-      # Perhaps also test if this was built?
-      image = ('{user}-docker-apache.bintray.io/beam/python'
-               '{version_suffix}:latest'.format(
-                   user=os.environ['USER'],
-                   version_suffix=version_suffix))
+      image = ('apachebeam/python{version_suffix}_sdk:latest'.format(
+          version_suffix=version_suffix))
       logging.info(
-          'Using latest locally built Python SDK docker image: %s.' % image)
+          'Using latest built Python SDK docker image: %s. If the image is not '
+          'available at local, it will be pulled from docker.io/apachebeam/'
+          'python%s_sdk repository.' % (image, version_suffix))
       return image
 
     else:
--- a/sdks/python/apache_beam/runners/portability/portable_runner.py
+++ b/sdks/python/apache_beam/runners/portability/portable_runner.py
@@ -21,13 +21,13 @@ import functools
 import itertools
 import json
 import logging
-import os
 import sys
 import threading
 import time
 
 import grpc
 
+from apache_beam import version as beam_version
 from apache_beam import metrics
 from apache_beam.options.pipeline_options import DebugOptions
 from apache_beam.options.pipeline_options import PortableOptions
@@ -78,23 +78,19 @@ class PortableRunner(runner.PipelineRunner):
 
   @staticmethod
   def default_docker_image():
-    if 'USER' in os.environ:
-      version_suffix = '.'.join([str(i) for i in sys.version_info[0:2]])
-      logging.warning('Make sure that locally built Python SDK docker image '
-                      'has Python %d.%d interpreter.' % (
-                          sys.version_info[0], sys.version_info[1]))
+    sdk_version = beam_version.__version__
+    version_suffix = '.'.join([str(i) for i in sys.version_info[0:2]])
+    logging.warning('Make sure that locally built Python SDK docker image '
+                    'has Python %d.%d interpreter.' % (
+                        sys.version_info[0], sys.version_info[1]))
 
-      image = ('apachebeam/python{version_suffix}_sdk:latest'.format(
-          version_suffix=version_suffix))
-      logging.info(
-          'Using latest built Python SDK docker image: %s. If the image is not '
-          'available at local, it will be pulled from docker.io/apachebeam/'
-          'python%s_sdk repository.' % (image, version_suffix))
-      return image
-
-    else:
-      logging.warning('Could not find a Python SDK docker image.')
-      return 'unknown'
+    image = ('apachebeam/python{version_suffix}_sdk:{tag}'.format(
+        version_suffix=version_suffix, tag=sdk_version))
+    logging.info(
+        'Using Python SDK docker image: %s. If the image is not '
+        'available at local, we will try to pull from hub.docker.com'
+        % (image))
+    return image
 
   @staticmethod
   def _create_environment(options):
--- a/sdks/python/container/py2/build.gradle
+++ b/sdks/python/container/py2/build.gradle
@@ -52,7 +52,10 @@ task copyLauncherDependencies(type: Copy) {
 }
 
 docker {
-  name containerImageName(name: "python2.7_sdk")
+  name containerImageName(
+          name: "python2.7_sdk", 
+          tag: project.rootProject.hasProperty(["docker-tag"]) ?
+                  project.rootProject["docker-tag"] : project['python_sdk_version'])
   files "../Dockerfile", "./build"
   buildArgs(['py_version': "2.7"])
 }
--- a/sdks/python/container/py2/build.gradle
+++ b/sdks/python/container/py2/build.gradle
@@ -52,7 +52,7 @@ task copyLauncherDependencies(type: Copy) {
 }
 
 docker {
-  name containerImageName(name: "python2.7")
+  name containerImageName(name: "python2.7_sdk")
   files "../Dockerfile", "./build"
   buildArgs(['py_version': "2.7"])
 }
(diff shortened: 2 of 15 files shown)


Hannah-Jiang:

Hannah-Jiang:

Hannah-Jiang:
Run Python PreCommit

Hannah-Jiang:
R: @aaltay 
Can you please review or find someone else to review it? Thanks.

sdks/python/apache_beam/runners/portability/portable_runner.py
```diff
@@ -84,13 +84,12 @@ def default_docker_image():
                       'has Python %d.%d interpreter.' % (
                           sys.version_info[0], sys.version_info[1]))
 
-      # Perhaps also test if this was built?
-      image = ('{user}-docker-apache.bintray.io/beam/python'
-               '{version_suffix}:latest'.format(
-                   user=os.environ['USER'],
-                   version_suffix=version_suffix))
+      image = ('apachebeam/python{version_suffix}_sdk:latest'.format(
```
aaltay: We may need to give people a way to run with their locally built containers for testing purposes, or for using custom containers.

We could reuse worker_harness_container_image flag, introduce a new flag. Or perhaps something else to do this in an easy way.

@tweise might have suggestions.

tweise: I have mixed feelings about this. When working with master, wouldn't I expect the locally built image to be used? What does it mean to refer to hub for something that wasn't released.

Perhaps do that depending on whether current version is dev/snapshot or a release?

tweise: And thanks for the ping @aaltay !

aaltay: To build on Thomas's idea, we could do something similar to what dataflowrunner does today (https://github.com/apache/beam/blob/9043b3e0bc138e63123d3fa55d4b7280eb96a82e/sdks/python/apache_beam/runners/dataflow/internal/apiclient.py#L946)

- If this is a released version on release branch (i.e. no 'dev' in version') default to image in hub. (release manager will have to build this container as part of the release anyway.)
- If it is a dev version --> this is a question. Dataflowrunner's solution is to have a fixed image that is occasionally updated as needed or if provided a flag use that as an override. (Here we might just simplify and require a locally built version. This occasionally built image is convenient but is error-prone and a hassle.)

Hannah-Jiang: Thanks @aaltay and @tweise for comments.

`--environment_config` option can be used to pass customized images when use Portable runner. The default image is used when no docker image is specified.

My understanding is, from users perspective, they don't need to worry about creating docker images at local when they write pipelines, so it's ok to pull it from remote. From developers perspective, developers who are working on sdks are expected to know how portable runner is working, so as how docker is used. If they want to test their developing sdks, they should build an image at local and use it. The image name can be same as the default one, then the local image is used instead of pulling it from remote, or can pass it with  `--environment_config` option if it has different name.

I can add more comments here to explain how to pass customized images.

What scenarios am I missing?

tweise: When the Beam version is not a development version (a release), then by default use the image corresponding to that release. Not `latest`, because that could be anything and may not be compatible with the release version? If the Python Beam version is 2.16, then it should not pick up something that could be 2.17, for example.

If the version of Beam is a development version, continue to use the locally built image as before.

tweise: But if a locally built container would automatically be picked up, then the distinction won't be necessary. Is that already the case with these changes?

Hannah-Jiang: Ah, I see, it’s possible that Beam version is not up to date. 
It will check local first, if the image is not available, them pull it from remote.

Hannah-Jiang:

Hannah-Jiang:
@aaltay , I fixed to use version as tag when pull default image. For quick workaround, I hardcoded version for docker tasks. Created a ticket for later improvement. https://issues.apache.org/jira/browse/BEAM-8192

tweise:
The version is available as property `project.version`

sdks/python/container/py2/build.gradle
```diff
@@ -51,8 +52,13 @@ task copyLauncherDependencies(type: Copy) {
   }
 }
 
+def sdk_version = '2.16.0'
```
aaltay: This is probably not a good idea. The reason is:
- This needs to match the https://github.com/apache/beam/blob/master/sdks/python/apache_beam/version.py - So it needs to be 2.16.0.dev here
- Release process needs to be updated to change these versions along with other versions. Release process automatically changes the version in the release branch to remove the ".dev" part and changes master branch to the next version.

- If we have to do this, at least is it possible to put this in a single gradle file?
- Another idea, we can execute or parse https://github.com/apache/beam/blob/master/sdks/python/apache_beam/version.py and extract version from here. That should be possible with gradle.

@yifanzou could you help with this?
/cc @markflyhigh as the release manager.

tweise: It would also be good to retain the distinction between -SNAPSHOT/.dev version and release version in the container tag. It can lead to confusing results when a locally built container from snapshot source overrides the image of a release.

aaltay:
`project.version` does not match Python version in master branch. It will be `2.16.0-SNAPSHOT` vs `2.16.0.dev`. I am guessing we can still use it and with a simple string raplace change '-SNAPSHOT' to '.dev' for python.

Hannah-Jiang:
> The version is available as property `project.version`

I investigated it, this would be the best way to go with, but this version is always `{version}-SNAPSHOT`, both for released and developing version. How do we know which tag we should use between `{version}`, `{version}.dev` and `{version}-SNAPSHOT`?

tweise:
It is indeed very odd that the version is set to -SNAPSHOT in release tags: https://github.com/apache/beam/blob/v2.15.0/gradle.properties#L26

tweise:
Versions changes are applied partially: https://github.com/apache/beam/commit/c071613ad92012dd1b431a41a325ab22f6bdfaa5

sdks/python/container/py2/build.gradle
yifanzou: https://docs.gradle.org/current/dsl/org.gradle.api.resources.TextResource.html

Hannah-Jiang:
> `project.version` does not match Python version in master branch. It will be `2.16.0-SNAPSHOT` vs `2.16.0.dev`. I am guessing we can still use it and with a simple string raplace change '-SNAPSHOT' to '.dev' for python.

I also investigated this option during weekend and there is a scenarios we cannot support with string replacement. 
For released versions, it would be `{version}-SNAPSHOT` at `gradle.properties`, but `{version}` at `version.py`. So if a user build an image from released version, it would be tagged with `{version}.dev`, however, an image with `{version}` tag will be used as a default image, which would make difficult to investigate for users who are not familiar with how tagging works with Beam.
This is the same issue as I mentioned previously, given `{version}-SNAPSHOT`, we cannot tell if it's developing version or released version. I also tried to change `{version}-SNAPSHOT` to `{version}` for released versions, and `{version}.dev` for developing versions, but 2.16 cutoff date is so close and the impact with the change would be wide, so I decided to introduce a new variable for now and make the change later.

We have the version defined at several different locations and suffixes are not consistent, which should be improved. I changed scope of [the Jira ticket](https://issues.apache.org/jira/browse/BEAM-8192) to include this work.

sdks/python/container/py2/build.gradle
Hannah-Jiang: Thanks for reviewing and comments.
I wanted to make `sdk_version` consistent with the one from `version.py`, forgot to change it back after testing. Thanks for catching it.

After evaluating many options, I think introducing a new variable `sdk_version` to `gradle.properties` is the easiest way for now. This version will be consistent to the one from `version.py`, so we need to update it when we release it. It would be better to read it from `version.py` and assign to `sdk_version`, however, as far as I know, `gradle.properties` is not supposed to read variables from another file. If it's possible, I am happy to try.

I made a commit with this change, can you please review and make comments?

Thanks,
Hannah

aaltay:
OK. I am fine with this change. I will only call 'sdk_version', 'python_sdk_version', since that is only applicable to python. And let's plan to clean it after 2.16. - LGTM.

@tweise - What do you think?

Hannah-Jiang:
> OK. I am fine with this change. I will only call 'sdk_version', 'python_sdk_version', since that is only applicable to python. And let's plan to clean it after 2.16. - LGTM.
Thanks, I addressed your comment.

tweise:
I think it is reasonable to go with this for 2.16 and work on a long term solution in master as follow-up.

Hannah-Jiang:
Run Java PreCommit

Hannah-Jiang:

Hannah-Jiang:

Hannah-Jiang:
Can we merge it?

Hannah-Jiang:
Thank you @aaltay .

ibzib:
@Hannah-Jiang Could you update https://github.com/apache/beam/blob/master/sdks/CONTAINERS.md to reflect these changes?

Hannah-Jiang:
> @Hannah-Jiang Could you update https://github.com/apache/beam/blob/master/sdks/CONTAINERS.md to reflect these changes?

We are already working on it, a tech writer is on it and will be released as part of 2.16.

PR #9499, opened 2019-09-06
[BEAM-8100] Exception handling for AsJsons and ParseJsons

aromanenko-dev:
This PR is based on work done in #7736 and it adds a possibility to handle exceptions for `AsJsons` and `ParseJsons` transforms.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/extensions/jackson/src/main/java/org/apache/beam/sdk/extensions/jackson/AsJsons.java
+++ b/sdks/java/extensions/jackson/src/main/java/org/apache/beam/sdk/extensions/jackson/AsJsons.java
@@ -17,13 +17,27 @@
  */
 package org.apache.beam.sdk.extensions.jackson;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import edu.umd.cs.findbugs.annotations.Nullable;
 import java.io.IOException;
+import java.util.Arrays;
+import java.util.Map;
+import java.util.Optional;
+import org.apache.beam.sdk.annotations.Experimental;
+import org.apache.beam.sdk.transforms.Contextful;
+import org.apache.beam.sdk.transforms.InferableFunction;
 import org.apache.beam.sdk.transforms.MapElements;
 import org.apache.beam.sdk.transforms.PTransform;
+import org.apache.beam.sdk.transforms.ProcessFunction;
+import org.apache.beam.sdk.transforms.Requirements;
 import org.apache.beam.sdk.transforms.SimpleFunction;
+import org.apache.beam.sdk.transforms.WithFailures;
+import org.apache.beam.sdk.values.KV;
 import org.apache.beam.sdk.values.PCollection;
-import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Optional;
+import org.apache.beam.sdk.values.TypeDescriptor;
+import org.apache.beam.sdk.values.TypeDescriptors;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.ImmutableMap;
 
 /**
  * {@link PTransform} for serializing objects to JSON {@link String Strings}. Transforms a {@code
@@ -41,12 +55,12 @@ public class AsJsons<InputT> extends PTransform<PCollection<InputT>, PCollection
    * into a {@link PCollection} of JSON {@link String Strings} representing those objects using a
    * Jackson {@link ObjectMapper}.
    */
-  public static <OutputT> AsJsons<OutputT> of(Class<? extends OutputT> outputClass) {
-    return new AsJsons<>(outputClass);
+  public static <InputT> AsJsons<InputT> of(Class<? extends InputT> inputClass) {
+    return new AsJsons<>(inputClass);
   }
 
-  private AsJsons(Class<? extends InputT> outputClass) {
-    this.inputClass = outputClass;
+  private AsJsons(Class<? extends InputT> inputClass) {
+    this.inputClass = inputClass;
   }
 
   /** Use custom Jackson {@link ObjectMapper} instead of the default one. */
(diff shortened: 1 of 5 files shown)


sdks/java/extensions/jackson/src/main/java/org/apache/beam/sdk/extensions/jackson/AsJsons.java
```diff
@@ -56,6 +63,35 @@ private AsJsons(Class<? extends InputT> outputClass) {
...
+   *     pojos.apply(
+   *         AsJsons.of(MyPojo.class)
+   *             .withFailures(new WithFailures.ExceptionAsMapHandler<MyPojo>() {}));
+   *
+   * PCollection<String> output = result.output(); // valid json elements
+   * PCollection<KV<MyPojo, Map<String, String>>> failures = result.failures();
+   * }</pre>
+   */
+  @Experimental(Experimental.Kind.WITH_EXCEPTIONS)
+  public <FailureT> AsJsonsWithFailures<FailureT> withFailures(
```
aromanenko-dev: I'm not sure that `withFailures` is a good name in this case and I couldn't find out if any agreement about that exists. So, any advices are welcomed.

aromanenko-dev:
R: @jklukas 
R: @reuvenlax 
Could you take a look?

sdks/java/extensions/jackson/src/main/java/org/apache/beam/sdk/extensions/jackson/AsJsons.java
jklukas: My bias would be towards following the API in `MapElements` as closely as possible, which would make this method `exceptionsVia` and would suggest that we should also have `exceptionsInto` and `exceptionsVia(ProcessFunction)` to allow users to define exception processing inline. I would love to see this factored out somehow so it's easier to add exception handling to different classes with a stable API and without repeating all this boilerplate, but I haven't found a suitable alternative.

aromanenko-dev:
Run Java PreCommit

sdks/java/extensions/jackson/src/main/java/org/apache/beam/sdk/extensions/jackson/AsJsons.java
```diff
@@ -56,6 +63,35 @@ private AsJsons(Class<? extends InputT> outputClass) {
     return newTransform;
   }
 
+  /**
+   * Returns a new {@link AsJsonsWithFailures} transform that catches exceptions raised while
+   * writing JSON elements, passing the raised exception instance and the input element being
+   * processed through the given {@code exceptionHandler} and emitting the result to a failure
+   * collection.
+   *
+   * <p>Example usage:
```
jklukas:
```suggestion
   * <p>See {@link WithFailures} documentation for usage patterns of the returned {@link
   * WithFailures.Result}.
   *
   * <p>Example usage:
```

sdks/java/extensions/jackson/src/main/java/org/apache/beam/sdk/extensions/jackson/ParseJsons.java
```diff
@@ -55,6 +61,34 @@ private ParseJsons(Class<? extends OutputT> outputClass) {
     return newTransform;
   }
 
+  /**
+   * Returns a new {@link ParseJsonsWithFailures} transform that catches exceptions raised while
+   * parsing elements, passing the raised exception instance and the input element being processed
+   * through the given {@code exceptionHandler} and emitting the result to a failure collection.
+   *
+   * <p>Example usage:
```
jklukas:
```suggestion
   * <p>See {@link WithFailures} documentation for usage patterns of the returned {@link
   * WithFailures.Result}.
   *
   * <p>Example usage:
```

sdks/java/extensions/jackson/src/main/java/org/apache/beam/sdk/extensions/jackson/AsJsons.java
jklukas: And perhaps these JSON cases are constrained enough that we could provide a default exceptionHandler so that the user doesn't need to specify one. Perhaps it would catch only the exceptions we expect to be raised (`JsonProcessingException` for AsJsons and `IOException` for ParseJsons), reraising an other exceptions, and return a KV that contains the input element along with a string containing the exception message.

sdks/java/extensions/jackson/src/main/java/org/apache/beam/sdk/extensions/jackson/AsJsons.java
```diff
@@ -56,6 +63,35 @@ private AsJsons(Class<? extends InputT> outputClass) {
...
+   *
+   * <pre>{@code
+   * WithFailures.Result<PCollection<String>, KV<MyPojo, Map<String, String>>> result =
+   *     pojos.apply(
+   *         AsJsons.of(MyPojo.class)
+   *             .withFailures(new WithFailures.ExceptionAsMapHandler<MyPojo>() {}));
```
jklukas: Perhaps we should provide specific default exception handlers for AsJsons and ParseJsons that are identical to ExceptionAsMapHandler except that they catch only the exceptions we expect to be raised (`JsonProcessingException` for AsJsons and `IOException` for ParseJsons), reraising any other exceptions.

sdks/java/extensions/jackson/src/main/java/org/apache/beam/sdk/extensions/jackson/AsJsons.java
aromanenko-dev: Agree to use the same API names for all transforms. Perhaps, it would make sense to add an interface for that?

Also, do we need to have separate method `exceptionsInto` only for lambdas?

And finally, yes, I also was thinking to have default exception handler. So, I think it  will be a good idea to add this, thanks.

jklukas: > Agree to use the same API names for all transforms. Perhaps, it would make sense to add an interface for that?

Yes, if it doesn't end cluttering the API too much. We could potentially put this at `WithFailures.ExceptionsVia` and it would have to be templated on `<InputT, FailureT, TransformT>` so we would end up with something like:

    implements WithFailures.ExceptionsVia<String, FailureT, AsJsonsWithFailures<FailureT>>

> Also, do we need to have separate method `exceptionsInto` only for lambdas?

Yes, I think we need a single `exceptionsInto` method. If the handler is passed as a lambda, Beam likely won't be able to infer a coder for the failure output collection without explicitly giving type information via `exceptionsInto`.

> And finally, yes, I also was thinking to have default exception handler. So, I think it will be a good idea to add this, thanks.

👍

aromanenko-dev:
@jklukas I addressed your review comments and added support of lambda and default handlers for `AsJsons` and `ParseJasons` transforms.

jklukas:
I'm having a realization that we could probably avoid most of the complexity here if `ParseJsons` simply returned the `MapElements` transform it builds rather than wrapping it. I don't think that's particularly an option since users already have code that expects an object of type `ParseJsons`, but what if we exposed a method to return the underlying `MapElements` so that users could call the existing error handling methods? It might look like:

```
PCollection<String> p = ...
p.apply(ParseJsons
               .of(MyClass.class)
               .unwrapped()
               .exceptionsVia(...))
```

This is slightly less convenient for the user, as they have to know about and invoke the `unwrapped` method (we can probably come up with a better name), but it avoids us having to re-implement all the logic from `MapElements` and it would establish a pattern by which other simple transforms that wrap a `MapElements` invocation could provide error handling.

What do you think of that approach?

lukecwik:
It is easier to maintain but I do think it would be easy for us to have a
copy of all the methods call the underlying MapElements without needing for
users to know to invoke unwrapped. Some people discover the API via code
completion and unwrapped is an unlikely place users will look.

aromanenko-dev:
I like the idea of simplifying the needed boiler code but I tend to agree with @lukecwik that, from user perspective, it will be lees evident. So, "wrapping" does this job. 

Not sure that this feature (exception handling) is requested by other transforms as some asked for Json ones, but it would be great to have simple and unified way how to add it into transform API and how to use it.

sdks/java/extensions/jackson/src/main/java/org/apache/beam/sdk/extensions/jackson/AsJsons.java
```diff
@@ -56,6 +70,60 @@ private AsJsons(Class<? extends InputT> outputClass) {
...
+   * PCollection<KV<MyPojo, Map<String, String>>> failures = result.failures();
+   * }</pre>
+   */
+  @Experimental(Experimental.Kind.WITH_EXCEPTIONS)
+  public <FailureT> AsJsonsWithFailures<FailureT> exceptionsVia(
+      InferableFunction<WithFailures.ExceptionElement<InputT>, FailureT> exceptionHandler) {
+    return new AsJsonsWithFailures<>(exceptionHandler, exceptionHandler.getOutputTypeDescriptor());
+  }
+
+  @Experimental(Experimental.Kind.WITH_EXCEPTIONS)
```
jklukas: Need Javadoc with example usage here.

sdks/java/extensions/jackson/src/main/java/org/apache/beam/sdk/extensions/jackson/AsJsons.java
```diff
@@ -64,13 +132,75 @@ private AsJsons(Class<? extends InputT> outputClass) {
...
+    @Nullable private final transient TypeDescriptor<FailureT> failureType;
+
+    AsJsonsWithFailures(
+        InferableFunction<WithFailures.ExceptionElement<InputT>, FailureT> exceptionHandler,
+        TypeDescriptor<FailureT> failureType) {
+      this.exceptionHandler = exceptionHandler;
+      this.failureType = failureType;
+    }
+
+    public AsJsonsWithFailures<FailureT> exceptionsVia(
```
jklukas: Needs Javadoc.

sdks/java/extensions/jackson/src/main/java/org/apache/beam/sdk/extensions/jackson/AsJsons.java
```diff
@@ -64,13 +132,75 @@ private AsJsons(Class<? extends InputT> outputClass) {
...
+                  Contextful.fn(
+                      (Contextful.Fn<InputT, String>) (input1, c) -> writeValue(input1),
+                      Requirements.empty()))
+              .exceptionsInto(failureType)
+              .exceptionsVia(exceptionHandler));
+    }
+  }
+
+  /**
+   * A default handler that extracts information from an exception to a {@code Map<String, String>}
```
jklukas: The info from this class doc would be perfect for filling out Javadoc for `exceptionsVia` that instantiates this.

sdks/java/extensions/jackson/src/main/java/org/apache/beam/sdk/extensions/jackson/ParseJsons.java
```diff
@@ -55,6 +67,61 @@ private ParseJsons(Class<? extends OutputT> outputClass) {
...
+   */
+  @Experimental(Experimental.Kind.WITH_EXCEPTIONS)
+  public <FailureT> ParseJsonsWithFailures<FailureT> exceptionsVia(
+      InferableFunction<WithFailures.ExceptionElement<String>, FailureT> exceptionHandler) {
+    return new ParseJsonsWithFailures<>(
+        exceptionHandler, exceptionHandler.getOutputTypeDescriptor());
+  }
+
+  @Experimental(Experimental.Kind.WITH_EXCEPTIONS)
+  public ParseJsonsWithFailures<KV<String, Map<String, String>>> exceptionsVia() {
```
jklukas: Javadoc needed.

sdks/java/extensions/jackson/src/main/java/org/apache/beam/sdk/extensions/jackson/ParseJsons.java
```diff
@@ -73,4 +139,66 @@ public OutputT apply(String input) {
...
+    @Nullable private final transient TypeDescriptor<FailureT> failureType;
+
+    ParseJsonsWithFailures(
+        InferableFunction<WithFailures.ExceptionElement<String>, FailureT> exceptionHandler,
+        TypeDescriptor<FailureT> failureType) {
+      this.exceptionHandler = exceptionHandler;
+      this.failureType = failureType;
+    }
+
+    public ParseJsonsWithFailures<FailureT> exceptionsVia(
```
jklukas: Javadoc needed.

sdks/java/extensions/jackson/src/main/java/org/apache/beam/sdk/extensions/jackson/AsJsons.java
aromanenko-dev: Done

sdks/java/extensions/jackson/src/main/java/org/apache/beam/sdk/extensions/jackson/AsJsons.java
aromanenko-dev: Done

sdks/java/extensions/jackson/src/main/java/org/apache/beam/sdk/extensions/jackson/AsJsons.java
aromanenko-dev: I added a link there to this javadoc to avoid copy-paste

sdks/java/extensions/jackson/src/main/java/org/apache/beam/sdk/extensions/jackson/ParseJsons.java
aromanenko-dev: Done

sdks/java/extensions/jackson/src/main/java/org/apache/beam/sdk/extensions/jackson/ParseJsons.java
aromanenko-dev: Done

aromanenko-dev:
@jklukas I added javadoc and examples there

aromanenko-dev:
I'm merging since it was already approved earlier.

PR #9237, opened 2019-08-03
[BEAM-5878] support DoFns with Keyword-only arguments

lazylynx:
support DoFns with Keyword-only arguments in Python 3 and add test reverted in #8750 with no syntax error in Python 2

tests condition are fixed as follows due to errors:

##### test_side_input_keyword_only_args

```
result2 = pcol | 'compute2' >> beam.FlatMap(
  sort_with_side_inputs,
  beam.pvalue.AsIter(side))
```
to
```
result2 = pcol | 'compute2' >> beam.FlatMap(
  sort_with_side_inputs,
  beam.pvalue.AsList(side))
```

##### test_combine_keyword_only_args

```
assert_that(result2, equal_to([23]), label='assert2')
```
to
```
assert_that(result2, equal_to([49]), label='assert2')
```
------------------------

Thank you for your contribution! Follow this checklist to help us incorporate your contribution quickly and easily:

 - [x] [**Choose reviewer(s)**](https://beam.apache.org/contribute/#make-your-change) and mention them in a comment (`R: @username`).
 - [x] Format the pull request title like `[BEAM-XXX] Fixes bug in ApproximateQuantiles`, where you replace `BEAM-XXX` with the appropriate JIRA issue, if applicable. This will automatically link the pull request to the issue.
 - [x] If this contribution is large, please file an Apache [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

Lang | SDK | Apex | Dataflow | Flink | Gearpump | Samza | Spark
--- | --- | --- | --- | --- | --- | --- | ---
Go | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go/lastCompletedBuild/) | --- | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_VR_Flink/lastCompletedBuild/) | --- | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_VR_Spark/lastCompletedBuild/)
[...]

PR #9517, opened 2019-09-07
Revert "[BEAM-5878] support DoFns with Keyword-only arguments"

tvalentyn:
Reverts apache/beam#9237


tvalentyn:
R: @lazylynx

tvalentyn:
R: chamikaramj

chamikaramj:
LGTM. Thanks.

To clarify, I believe this is being reverted since the monkey patch caused some user pipelines to fail with an infinite recursion.

tvalentyn:
Thanks @chamikaramj . Reasons for revert were also discussed on the original PR.

PR #9544, opened 2019-09-11
[BEAM-8191] Fixes potentially large number of tasks on Spark after Flatten.pCollections()

pbackx:
In the Spark runner (Beam 2.14.0 and 2.15.0), a Flatten.pCollections() PTransform is translated into a Spark union operation. This union will create a potentially large number of partitions in the RDD that can overload the driver.

This PR does a coalesce operation after the union, which will reduce the number of partitions in the RDD at virtually no cost.

See the JIRA ticket for more detail:
https://issues.apache.org/jira/browse/BEAM-8191

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/runners/spark/src/main/java/org/apache/beam/runners/spark/translation/SparkBatchPortablePipelineTranslator.java
+++ b/runners/spark/src/main/java/org/apache/beam/runners/spark/translation/SparkBatchPortablePipelineTranslator.java
@@ -353,6 +353,11 @@ public class SparkBatchPortablePipelineTranslator {
         index++;
       }
       unionRDD = context.getSparkContext().union(rdds);
+
+      Partitioner partitioner = getPartitioner(context);
+      if (partitioner != null) {
+        unionRDD = unionRDD.coalesce(partitioner.numPartitions(), false);
+      }
     }
     context.pushDataset(getOutputId(transformNode), new BoundedDataset<>(unionRDD));
   }
--- a/runners/spark/src/main/java/org/apache/beam/runners/spark/translation/TransformTranslator.java
+++ b/runners/spark/src/main/java/org/apache/beam/runners/spark/translation/TransformTranslator.java
@@ -99,6 +99,11 @@ public final class TransformTranslator {
             index++;
           }
           unionRDD = context.getSparkContext().union(rdds);
+
+          Partitioner partitioner = getPartitioner(context);
+          if (partitioner != null) {
+            unionRDD = unionRDD.coalesce(partitioner.numPartitions(), false);
+          }
         }
         context.putDataset(transform, new BoundedDataset<>(unionRDD));
       }


pbackx:
R: @iemejia @dmvk

iemejia:
Oups sorry for late answer, @RyanSkraba can you PTAL.

runners/spark/src/main/java/org/apache/beam/runners/spark/translation/SparkBatchPortablePipelineTranslator.java
```diff
@@ -353,6 +353,11 @@ public void setName(String name) {
         index++;
       }
       unionRDD = context.getSparkContext().union(rdds);
+
+      Partitioner partitioner = getPartitioner(context);
```
RyanSkraba: In principle, I don't see any problem with coalescing partitions to avoid the overhead of too many partitions -- it's certainly something that a developer would do when hand-coding spark jobs.

In this case, I'm not sure about your logic here -- as far as I can tell, this will *always* coalesce to the `sc.defaultParallelism()` (or not at all if bundleSize is set in the pipeline options).  

As I understand, bundleSize is only relevant for sourceRDD, but the union causing the partition explosion can be anywhere in the DAG.

What do you think about checking the number of partitions in the unionRDD against a threshold, and doing the coalesce if there are "too many" instead?

What do you think about something like: 

```
if (unionRdd.partitions.size > sc.defaultParallelism() * THRESHOLD) {
  coalesce(sc.defaultParallelism()  * 2) // or three according to the spark tuning recommendations.
}
```

Even setting the threshold to a large, but non-negligable fixed value like 5-10K might be enough.

What do you think?

pbackx: Hi @RyanSkraba, thanks for reviewing.
The default parallelism is kind of arbitrary, I understand and I also wanted a better option.

I'm not sure about the threshold being that large. On our 150 machine cluster with 13 usable cores in each machine and 2 tasks per core, we have parallelism of 3900. Multiplying that by 5000 is exactly what is causing us troubles at the moment.

Maybe this threshold could be a Spark pipeline option? Something like "spark.max.parallelism"?

Regarding the note about multiplying by 2 or 3. I already take that into account when setting the default parallelism, so I don't think this is needed. This value could maybe be replaced by the max parallelism config option? 

Are you guys open for adding an extra pipeline option? That could solve this problem. 
* If I like to keep the maximum parallelism at the default parallelism, I can set both values through the pipeline options. 
* The default value could be positive infinity, which would be the exact behavior as it is today.

Sounds like a good idea?

RyanSkraba: This *does* sound like a good idea!  I linked your JIRA to https://issues.apache.org/jira/browse/BEAM-8384 .  Before adding some a new pipeline option, it would be great if there were a better "overall" view of how the SparkRunner is managing parallelism.

This seems like it would be a good area to collaborate.

pbackx: Due to changes in priorities, this has moved a bit down my list of todo's, hence the delay in getting back.

Yes, the way parallelism (and partitioning) works in the SparkRunner could use some improvements. The problem is that it is difficult to make this generic. In some parts of the pipeline you may want to have more parallelism, while in others you don't.

Having a maximum is just a first step. I think ideally, you want some way to supply the Spark partitioner yourself.

I do need to study Spark partitioning more before I feel confident in making a proper suggestion. But in the mean time, if you have ideas, I'm certainly not afraid of just trying things out.

iemejia:
stale bot calm down this fix seems worth the attention still

PR #9454, opened 2019-08-29
[BEAM-8111] Add ValidatesRunner test to AvroSchemaTest

TheNeuralBit:
Adds a ValidatesRunner test to AvroSchemaTest that would have caught [BEAM-8111](https://issues.apache.org/jira/browse/BEAM-8111).

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

Lang | SDK | Apex | Dataflow | Flink | Gearpump | Samza | Spark
--- | --- | --- | --- | --- | --- | --- | ---
Go | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go/lastCompletedBuild/) | --- | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_VR_Flink/lastCompletedBuild/) | --- | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_VR_Spark/lastCompletedBuild/)
Java | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Apex/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Dataflow/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Flink/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Flink_Batch/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Flink_Streaming/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Gearpump/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Samza/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Spark/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Spark_Batch/lastCompletedBuild/)
[...]

PR #9541, opened 2019-09-11
[BEAM-8204][BEAM-8205] Exclude AvroSchemaTest for Apex and Flink runner for now

robinyqiu:
Exclude the test because they are currently breaking post-commit tests.

The test was added in https://github.com/apache/beam/pull/9454

r: @TheNeuralBit @reuvenlax 

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

Lang | SDK | Apex | Dataflow | Flink | Gearpump | Samza | Spark
--- | --- | --- | --- | --- | --- | --- | ---
Go | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go/lastCompletedBuild/) | --- | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_VR_Flink/lastCompletedBuild/) | --- | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_VR_Spark/lastCompletedBuild/)
Java | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Apex/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Dataflow/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Flink/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Flink_Batch/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Flink_Streaming/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Gearpump/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Samza/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Spark/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Spark_Batch/lastCompletedBuild/)
[...]

PR #9550, opened 2019-09-11
Revert "[BEAM-8111] Add ValidatesRunner test to AvroSchemaTest"

markflyhigh:
Reverts apache/beam#9454 due to AvroSchemaTest.testAvroPipelineGroupBy consistently failed in Flink ValidatesRunner test suite ([link](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Flink/)).

[This](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Flink/5178/) is the first failure instance. You can go to console output or Gradle scan for error details.

+R: @reuvenlax @kennknowles @lukecwik

### DIFF
--- a/sdks/java/core/src/test/java/org/apache/beam/sdk/schemas/AvroSchemaTest.java
+++ b/sdks/java/core/src/test/java/org/apache/beam/sdk/schemas/AvroSchemaTest.java
@@ -32,17 +32,9 @@ import org.apache.avro.reflect.AvroName;
 import org.apache.avro.reflect.AvroSchema;
 import org.apache.avro.util.Utf8;
 import org.apache.beam.sdk.schemas.LogicalTypes.FixedBytes;
-import org.apache.beam.sdk.schemas.Schema.Field;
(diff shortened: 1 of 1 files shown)


reuvenlax:
FYI this appears to be Flink ValidatesRunner, not Dataflow?

reuvenlax:
Also it appears that Flink relies on all the Coders being exactly the same,
which is far more restrictive than other runners. Is there another way for
Flink to do this - maybe comparing the TypeDescriptor instead of the Coder?

reuvenlax:
Also I suspect that https://github.com/apache/beam/pull/9446 will fix this.

markflyhigh:
It should be Flink. Thank you Reuven!

Robin already had https://github.com/apache/beam/pull/9541 to skip this test and test is done. I think we can continue with https://github.com/apache/beam/pull/9541 and ignore this rollback.

robertwb:
So should this be closed?

PR #9119, opened 2019-07-22
[BEAM-6783] byte[] breaks in BeamSQL codegen

kanterov:
Fix casting in BeamCalcRel.

------------------------

Thank you for your contribution! Follow this checklist to help us incorporate your contribution quickly and easily:

 - [x] [**Choose reviewer(s)**](https://beam.apache.org/contribute/#make-your-change) and mention them in a comment (`R: @username`).
 - [x] Format the pull request title like `[BEAM-XXX] Fixes bug in ApproximateQuantiles`, where you replace `BEAM-XXX` with the appropriate JIRA issue, if applicable. This will automatically link the pull request to the issue.
 - [x] If this contribution is large, please file an Apache [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

Lang | SDK | Apex | Dataflow | Flink | Gearpump | Samza | Spark
--- | --- | --- | --- | --- | --- | --- | ---
Go | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go/lastCompletedBuild/) | --- | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_VR_Flink/lastCompletedBuild/) | --- | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_VR_Spark/lastCompletedBuild/)
[...]

PR #9586, opened 2019-09-16
[BEAM-8241] SQL code gen is more restrictive than Calcite

11moon11:
Fix for the SQL code gen to be less restrictive.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/ScalarFunctionImpl.java
+++ b/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/ScalarFunctionImpl.java
@@ -138,11 +138,15 @@ public class ScalarFunctionImpl extends UdfImplReflectiveFunctionBase
     }
 
     private static List<Expression> translate(List<Type> types, List<Expression> expressions) {
+      // https://issues.apache.org/jira/browse/BEAM-8241
+      // In user defined functions Calcite allows variants with fewer arguments than Beam defined.
       Preconditions.checkArgument(
           types.size() >= expressions.size(), "types.size() < expressions.size()");
 
       final List<Expression> translated = new ArrayList<>();
       for (int i = 0; i < expressions.size(); i++) {
+        // TODO: [BEAM-8255] Add support for user defined function with var-arg
+        // Ex: types: [String[].class], expression: [param1, param2, ...]
         translated.add(translate(types.get(i), expressions.get(i)));
       }
 
--- a/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/ScalarFunctionImpl.java
+++ b/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/ScalarFunctionImpl.java
@@ -139,7 +139,7 @@ public class ScalarFunctionImpl extends UdfImplReflectiveFunctionBase
 
     private static List<Expression> translate(List<Type> types, List<Expression> expressions) {
       Preconditions.checkArgument(
-          types.size() == expressions.size(), "types.size() != expressions.size()");
+          types.size() >= expressions.size(), "types.size() < expressions.size()");
 
       final List<Expression> translated = new ArrayList<>();
       for (int i = 0; i < expressions.size(); i++) {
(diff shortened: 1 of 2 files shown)


11moon11:
R: @apilloud

apilloud:
R: @kanterov You were the original author of this assert in #9119. It looks like it was just intended to guard the for loop and it still does that. Just wanted to make sure there wasn't something deeper that I am missing.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/ScalarFunctionImpl.java
```diff
@@ -139,7 +139,7 @@ public CallImplementor getImplementor() {
 
     private static List<Expression> translate(List<Type> types, List<Expression> expressions) {
       Preconditions.checkArgument(
-          types.size() == expressions.size(), "types.size() != expressions.size()");
+          types.size() >= expressions.size(), "types.size() < expressions.size()");
```
amaliujia: Could you explain a bit on this change? I am not quite following how this will fix those tests?

amaliujia: Also, can we move this check completely?

11moon11: Concat function with only 5 arguments is registered: "org/apache/beam/sdk/extensions/sql/zetasql/BeamBuiltinMethods.java", but there are other concat functions that handle various number of arguments (1, 2, 3, 4, 5): "org/apache/beam/sdk/extensions/sql/zetasql/StringFunctions.java".

When using concat with less arguments (than it was registered with), Calcite will generate a call to the method with the correct number of arguments, but the precondition check will fail, since it compares the number of arguments passed to the number of arguments registered function accepts (always 5 for concat).

amaliujia: I believe the right resolution is to register a var-arg version of function to Calcite. For example, for concat, the expected input should be:
types: [String[].class]
expression: [param1, param2, ...]
We will still be able to verify the input types and expression.

However my recall is Calcite doesn't support var-arg in user defined function (maybe I am wrong). So if we cannot define the var-arg version, we actually should remove this check as it's not longer useful.

apilloud: This check is still useful, it is protecting the for loop right after it. It is just too restrictive.

amaliujia: The following loop was written based on the assumption that types = expression so each expression has a corresponding type. The check was based on the same assumption.

But now the assumption is proven to be wrong.

For example, I can see a valid input is:
types: [String[].class]
expression: [param1, param2, ...]

so the for loop will need to be modified to make this input work(every param match with String type). The loop happens to be right with `types.size() >= expressions.size()`.

I don't want to make this PR a one that needs long time to fix something. So if we want to check-in `types.size() >= expressions.size()`, at least can @11moon11 add comments to document what is going on, why need this change and leave a TODO in the for loop with a JIRA to support user defined function with var-arg

11moon11: I did add comments, please take a look when you have a chance.

11moon11:

11moon11:
Run Java PreCommit

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/ScalarFunctionImpl.java
amaliujia: LGTM thanks.

11moon11:

apilloud:
The portability test failure is in Docker, so unrelated. I'm going to merge.
LGTM

kanterov:
@apilloud @11moon11 thanks for taking care of it, I agree, the change makes sense, didn't know that Calcite supports that.

PR #9619, opened 2019-09-19
[BEAM-7760] Added pipeline_instrument module

KevinGG:
1. Added the pipeline_instrument module to automatically instrument a
given Interactive Beam pipeline by mutating it with additional cache
based PTransforms if available so that, within an interactive
environment, each pipeline run could have effect against the future
runs to provide an interactive experience when executing Beam
pipelines.
2. The pipeline_instrument module will replace pipeline_analyzer module
when the integration with a re-written of display module interfaces is
done since the interactivity instrument (i.e., parameters passed and
used) has changed. Most of the display logic won't change.
3. An optional pruning logic is marked as TODO so that when executing
an instrumented pipeline, any sub graph doesn't generate new states
should not be re-executed if implemented.
4. Tests included.
5. The philosophy is to keep the pipeline instance defined by user code
intact and mutate directly on a copied pipeline instance; to always
convert the instrumented pipeline to a portable pipeline and pass it
to runner for execution; to maintain the mapping relationship from
original user defined pipeline to instrumented copied pipeline instances
and jobs executed by runners.
6. Additional complexity occurs when there are multiple pipeline
instances defined in user code, multiple runners instantiated, and
multiple jobs running from those pipeline instances by the runner
instances. Currently, the only guarantee is that a pipeline result
bounded to a job must be the return of a run by a runner and originate
from a pipeline instance. Some design change proposals are marked as TODO
around the cache scoping to solve the context problem in interactive
environment: when and what to instrument when the additional complexity
occurs.
------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
+++ b/sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
@@ -1,0 +1,470 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""Module to instrument interactivity to the given pipeline.
+
+For internal use only; no backwards-compatibility guarantees.
+This module accesses current interactive environment and analyzes given pipeline
+to transform original pipeline into a one-shot pipeline with interactivity.
+"""
+from __future__ import absolute_import
+
+import logging
+
+import apache_beam as beam
+from apache_beam.pipeline import PipelineVisitor
+from apache_beam.runners.interactive import cache_manager as cache
+from apache_beam.runners.interactive import interactive_environment as ie
+
+READ_CACHE = "_ReadCache_"
+WRITE_CACHE = "_WriteCache_"
+
+
+class PipelineInstrument(object):
+  """A pipeline instrument for pipeline to be executed by interactive runner.
+
+  This module should never depend on underlying runner that interactive runner
+  delegates. It instruments the original instance of pipeline directly by
+  appending or replacing transforms with help of cache. It provides
+  interfaces to recover states of original pipeline. It's the interactive
+  runner's responsibility to coordinate supported underlying runners to run
+  the pipeline instrumented and recover the original pipeline states if needed.
+  """
+
+  def __init__(self, pipeline, options=None):
+    self._pipeline = pipeline
+    # The cache manager should be initiated outside of this module and outside
+    # of run_pipeline() from interactive runner so that its lifespan could cover
+    # multiple runs in the interactive environment. Owned by
+    # interactive_environment module. Not owned by this module.
+    # TODO(BEAM-7760): change the scope of cache to be owned by runner or
+    # pipeline result instances because a pipeline is not 1:1 correlated to a
+    # running job. Only complete and read-only cache is valid across multiple
+    # jobs. Other cache instances should have their own scopes. Some design
+    # change should support only runner.run(pipeline) pattern rather than
+    # pipeline.run([runner]) and a runner can only run at most one pipeline at a
+    # time. Otherwise, result returned by run() is the only 1:1 anchor.
+    self._cache_manager = ie.current_env().cache_manager()
+
+    # Invoke a round trip through the runner API. This makes sure the Pipeline
+    # proto is stable. The snapshot of pipeline will not be mutated within this
+    # module and can be used to recover original pipeline if needed.
+    self._pipeline_snap = beam.pipeline.Pipeline.from_runner_api(
+        pipeline.to_runner_api(use_fake_coders=True),
+        pipeline.runner,
+        options)
+    # Snapshot of original pipeline information.
+    (self._original_pipeline_proto,
+     self._original_context) = self._pipeline_snap.to_runner_api(
+         return_context=True, use_fake_coders=True)
+
+    # All compute-once-against-original-pipeline fields.
+    self._has_unbounded_source = has_unbounded_source(self._pipeline_snap)
+    # TODO(BEAM-7760): once cache scope changed, this is not needed to manage
+    # relationships across pipelines, runners, and jobs.
+    self._pcolls_to_pcoll_id = pcolls_to_pcoll_id(self._pipeline_snap,
+                                                  self._original_context)
+
+    # A mapping from PCollection id to python id() value in user defined
+    # pipeline instance.
+    (self._pcoll_version_map,
+     self._cacheables) = cacheables(self.pcolls_to_pcoll_id())
+
+    # A dict from cache key to PCollection that is read from cache.
+    # If exists, caller should reuse the PCollection read. If not, caller
+    # should create new transform and track the PCollection read from cache.
+    # (Dict[str, AppliedPTransform]).
+    self._cached_pcoll_read = {}
+
+  def instrumented_pipeline_proto(self):
+    """Always returns a new instance of portable instrumented proto."""
+    return self._pipeline.to_runner_api(use_fake_coders=True)
+
+  def has_unbounded_source(self):
+    """Checks if a given pipeline has any source that is unbounded.
+
+    The function directly checks the source transform definition instead
+    of pvalues in the pipeline. Thus manually setting is_bounded field of
+    a PCollection or switching streaming mode will not affect this
+    function's result. The result is always deterministic when the source
+    code of a pipeline is defined.
+    """
+    return self._has_unbounded_source
+
+  def cacheables(self):
+    """Finds cacheable PCollections from the pipeline.
+
+    The function only treats the result as cacheables since there is no
+    guarantee whether the cache desired PCollection has been cached or
+    not. A PCollection desires caching when it's bound to a user defined
+    variable in source code. Otherwise, the PCollection is not reusale
+    nor introspectable which nullifying the need of cache.
+    """
+    return self._cacheables
+
+  def pcolls_to_pcoll_id(self):
+    """Returns a dict mapping str(PCollection)s to IDs."""
+    return self._pcolls_to_pcoll_id
+
+  def original_pipeline_proto(self):
+    """Returns the portable proto representation of the pipeline before
+    instrumentation."""
+    return self._original_pipeline_proto
+
+  def original_pipeline(self):
+    """Returns a snapshot of the pipeline before instrumentation."""
+    return self._pipeline_snap
+
+  def instrument(self):
+    """Instruments original pipeline with cache.
+
+    For cacheable output PCollection, if cache for the key doesn't exist, do
+    _write_cache(); for cacheable input PCollection, if cache for the key
+    exists, do _read_cache(). No instrument in any other situation.
+
+    Modifies:
+      self._pipeline
+    """
+    self._preprocess()
+    cacheable_inputs = set()
+
+    class InstrumentVisitor(PipelineVisitor):
+      """Visitor utilizes cache to instrument the pipeline."""
+
+      def __init__(self, pin):
+        self._pin = pin
+
+      def enter_composite_transform(self, transform_node):
+        self.visit_transform(transform_node)
+
+      def visit_transform(self, transform_node):
+        cacheable_inputs.update(self._pin._cacheable_inputs(transform_node))
+
+    v = InstrumentVisitor(self)
+    self._pipeline.visit(v)
+    # Create ReadCache transforms.
+    for cacheable_input in cacheable_inputs:
+      self._read_cache(cacheable_input)
+    # Replace/wire inputs w/ cached PCollections from ReadCache transforms.
+    self._replace_with_cached_inputs()
+    # Write cache for all cacheables.
+    for _, cacheable in self.cacheables().items():
+      self._write_cache(cacheable['pcoll'])
+    # TODO(BEAM-7760): prune sub graphs that doesn't need to be executed.
+
+  def _preprocess(self):
+    """Pre-processes the pipeline.
+
+    Since the pipeline instance in the class might not be the same instance
+    defined in the user code, the pre-process will figure out the relationship
+    of cacheable PCollections between these 2 instances by replacing 'pcoll'
+    fields in the cacheable dictionary with ones from the running instance.
+    """
+
+    class PreprocessVisitor(PipelineVisitor):
+
+      def __init__(self, pin):
+        self._pin = pin
+
+      def enter_composite_transform(self, transform_node):
+        self.visit_transform(transform_node)
+
+      def visit_transform(self, transform_node):
+        for in_pcoll in transform_node.inputs:
+          self._process(in_pcoll)
+        for out_pcoll in transform_node.outputs.values():
+          self._process(out_pcoll)
+
+      def _process(self, pcoll):
+        pcoll_id = self._pin.pcolls_to_pcoll_id().get(str(pcoll), '')
+        if pcoll_id in self._pin._pcoll_version_map:
+          cacheable_key = self._pin._cacheable_key(pcoll)
+          if (cacheable_key in self._pin.cacheables() and
+              self._pin.cacheables()[cacheable_key]['pcoll'] != pcoll):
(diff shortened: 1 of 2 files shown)


KevinGG:
retest this please

KevinGG:
R:@aaltay
PTAL
I'll fix the PyLint checks in the PreCommit. There are many warning level (exit code 4) lint reports related to Python2 and Beam pipeline definition in unit test code that fail some Gradle tasks.

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
```diff
@@ -0,0 +1,470 @@
...
+    # multiple runs in the interactive environment. Owned by
+    # interactive_environment module. Not owned by this module.
+    # TODO(BEAM-7760): change the scope of cache to be owned by runner or
+    # pipeline result instances because a pipeline is not 1:1 correlated to a
+    # running job. Only complete and read-only cache is valid across multiple
+    # jobs. Other cache instances should have their own scopes. Some design
+    # change should support only runner.run(pipeline) pattern rather than
+    # pipeline.run([runner]) and a runner can only run at most one pipeline at a
+    # time. Otherwise, result returned by run() is the only 1:1 anchor.
+    self._cache_manager = ie.current_env().cache_manager()
```
aaltay: Curious, why not pass this information through the constructor?

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
```diff
@@ -0,0 +1,470 @@
...
+    # Invoke a round trip through the runner API. This makes sure the Pipeline
+    # proto is stable. The snapshot of pipeline will not be mutated within this
+    # module and can be used to recover original pipeline if needed.
+    self._pipeline_snap = beam.pipeline.Pipeline.from_runner_api(
+        pipeline.to_runner_api(use_fake_coders=True),
+        pipeline.runner,
+        options)
+    # Snapshot of original pipeline information.
+    (self._original_pipeline_proto,
+     self._original_context) = self._pipeline_snap.to_runner_api(
```
aaltay: Would not the original be the result of `pipeline.to_runner_api(use_fake_coders=True)` ?

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
```diff
@@ -0,0 +1,470 @@
...
+    (self._pcoll_version_map,
+     self._cacheables) = cacheables(self.pcolls_to_pcoll_id())
+
+    # A dict from cache key to PCollection that is read from cache.
+    # If exists, caller should reuse the PCollection read. If not, caller
+    # should create new transform and track the PCollection read from cache.
+    # (Dict[str, AppliedPTransform]).
+    self._cached_pcoll_read = {}
+
+  def instrumented_pipeline_proto(self):
```
aaltay: What exactly is an instrumented pipeline? Is it a pipeline, that is made from a snapshot? Would this concept be user visible?

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
```diff
@@ -0,0 +1,470 @@
...
+    """Always returns a new instance of portable instrumented proto."""
+    return self._pipeline.to_runner_api(use_fake_coders=True)
+
+  def has_unbounded_source(self):
```
aaltay: You might convert this to  a `@property`.

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
```diff
@@ -0,0 +1,470 @@
...
+
+    The function directly checks the source transform definition instead
+    of pvalues in the pipeline. Thus manually setting is_bounded field of
+    a PCollection or switching streaming mode will not affect this
+    function's result. The result is always deterministic when the source
+    code of a pipeline is defined.
+    """
+    return self._has_unbounded_source
+
+  def cacheables(self):
```
aaltay: You can probably cut through some of this by changing self._x to be self.x instead.

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
```diff
@@ -0,0 +1,470 @@
...
+    for cacheable_input in cacheable_inputs:
+      self._read_cache(cacheable_input)
+    # Replace/wire inputs w/ cached PCollections from ReadCache transforms.
+    self._replace_with_cached_inputs()
+    # Write cache for all cacheables.
+    for _, cacheable in self.cacheables().items():
+      self._write_cache(cacheable['pcoll'])
+    # TODO(BEAM-7760): prune sub graphs that doesn't need to be executed.
+
+  def _preprocess(self):
```
aaltay: Where is `_preprocess` used other than the `instrument` above?

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
```diff
@@ -0,0 +1,470 @@
...
+    v = PreprocessVisitor(self)
+    self._pipeline.visit(v)
+
+  def _write_cache(self, pcoll):
+    """Caches a cacheable PCollection.
+
+    For the given PCollection, by appending sub transform part that materialize
+    the PCollection through sink into cache implementation. The cache write is
+    not immediate. It happens when the runner runs the transformed pipeline
+    and thus not usable for this run as intended. It's the caller's
```
aaltay: Who is the caller? Is not it the same class?

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
```diff
@@ -0,0 +1,470 @@
...
+    And if a cache is invalidated, the PCollection's new key is guaranteed to
+    not exist in current cache.
+
+    Modifies:
+      self._pipeline
+    """
+    if pcoll.pipeline is not self._pipeline:
+      return
+    key = self.cache_key(pcoll)
+    if not self._cache_manager.exists('full', key):
```
aaltay: What is being checked here? Could you also clarify the log message below?

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
```diff
@@ -0,0 +1,470 @@
...
+      if key not in self._cached_pcoll_read:
+        logging.debug('read cache: %s', key)
+        # Mutates the pipeline with cache read transform attached
+        # to root of the pipeline.
+        pcoll_from_cache = (
+            self._pipeline
+            | '{}{}'.format(READ_CACHE, key) >> cache.ReadCache(
+                self._cache_manager, key))
+        self._cached_pcoll_read[key] = pcoll_from_cache
+    # else: NOOP when cache doesn't exist.
```
aaltay: Should the user know about this?

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
```diff
@@ -0,0 +1,470 @@
...
+      if self._cacheable_key(in_pcoll) in self.cacheables():
+        inputs.add(in_pcoll)
+    return inputs
+
+  def _cacheable_key(self, pcoll):
+    """Gets the key a cacheable PCollection is tracked within the instrument."""
+    return cacheable_key(pcoll, self.pcolls_to_pcoll_id(),
+                         self._pcoll_version_map)
+
+  def cache_key(self, pcoll):
```
aaltay: What is the difference between cacheable_key and cache_key, they both seemingly return a key.

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
```diff
@@ -0,0 +1,470 @@
...
+    """
+    cacheable = self.cacheables().get(self._cacheable_key(pcoll), None)
+    if cacheable:
+      return '_'.join((cacheable['var'],
+                       cacheable['version'],
+                       cacheable['pcoll_id'],
+                       cacheable['producer_version']))
+    return ''
+
+  def _debug_appliedptransform(self, transform_node):
```
aaltay: This seems like a pure debugging addition, do you need it here?

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
```diff
@@ -0,0 +1,470 @@
...
+        self.visit_transform(transform_node)
+
+      def visit_transform(self, transform_node):
+        self._pin._debug_appliedptransform(transform_node)
+
+    v = DebugVisitor(self)
+    self._pipeline.visit(v)
+
+
+def pin(pipeline, options=None):
```
aaltay: What does `pin` mean?

KevinGG:
@aaltay Thanks for the quick response! I'm currently oncall till next Tuesday. I'll get back to the PR once I'm off duty asap. Thank you very much for the detailed review!

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
KevinGG: There is some work from Alexey to decouple the caching from a centralized cache manager instance. So a cache manager might not exist anymore. Doing this to avoid exposing it in the constructor.

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
KevinGG: Inside this module, the pipeline instance is actually modified (with read/write cache transforms). This approach takes a snapshot of the original pipeline before mutating it so that the invoking [Interactive]Runner can always recover the original pipeline.

Note: the pipeline instance received by this module is from the runner. It's highly possible that a round-trip between pipeline and runner_api proto has been done by the runner itself.
For example, the pipeline instance the notebook user has defined (and continue developing) in their notebook will not be the same instance (but a copy) used during run_pipeline(pipeline) by InteractiveRunner.
This module will receive that copied pipeline instance, snapshot it and mutate it.

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
KevinGG: I'll compose and share a 'vocabulary' document soon. We can vote and decide how we call different things for Beam in an Interactive Notebook (Jupyter lab) environment.

An instrumented pipeline is a plain Python `apache_beam.Pipeline` object with cache related transforms so that interactivity features such as PCollection introspection and partial pipeline (or pipeline segment) execution can be supported.
It has the same functionality of an original pipeline being instrumented but with additional implicit effect for interactivity.

As long as caching is implicit, this concept should not be user visible.

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
KevinGG: Acked! I'll convert this and other applicable properties to `@property`. Thanks!

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
KevinGG: Agreed! Will do that for the compute-once fields.

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
KevinGG: They should not know about this.
The partial execution optimization is done at a best-effort basis.
If a cache is not available for some reason (corrupted cache, lost file and etc.), re-computation will be sufficient. There is no need to fail the pipeline execution or notify the user about the implicit cache.

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
KevinGG: A cacheable_key is the key to a cacheable.
A cacheable means that the PCollection described by it could be cached. But whether a cache exists (a pipeline has been executed before, a PCollection has been re-evaluated), the pipeline_instrument doesn't know.

A cache_key is the key to a cache.
When given a cache_key, if caching mechanism cannot find a cache with that key. It means that the cache has not been created. And this cache_key will eventually lead you to some cache as pipeline executions occur in current notebook environment, until the pipeline_instrument stops generating this cache_key but other new keys (caused by changes in the user defined pipeline in notebook). Old cache is invalidated in this way.

Because there are 2 context:
1. In notebook, there is an `apache_beam.Pipeline` instance where user has a handle.
2. In runner and this pipeline_instrument, there is a copied `apache_beam.Pipeline` instance where the runner and pipeline_instrument has a handle and potentially mutates.
Context 1 isolates the scope of caching. When the `pipeline` object is gone in context 1, all the caching will be invalidated. 
Example scenarios to refresh context 1:
1. Restarts the IPython kernel;
2. Assign a different instance to `pipeline` in notebook.

Within the scope of context 2, pipeline_instrument accesses context 1 to figure out the relationship between PCollections in copied `pipeline` (in context 2) and original `pipeline` (in context 1) to decide where the cache should be written to or read from.

cache_key functions anywhere (pipeline execution, PCollection visualization or simply read a cache file) within context 1.
cacheable_key points to cacheable linking context 1 and context 2 within a single `PipelineInstrument` instance. The instance is only valid during `pipeline.run()`.
Whenever the `pipeline` changes in context 1, this `PipelineInstrument` should also be re-evaluated (within InteractiveRunner).

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
KevinGG: It's better suited for some testing package. I'll remove all the debug statement and functions from this module.

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
KevinGG: It's short for `pipeline instrument`, also (verb.) `pin`s the `pipeline` at this moment, with original pipeline information snapshotted and caching (interactivity) transforms instrumented once by mutating the pipeline object from the state when it receives the pipeline.

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
KevinGG: First it checks whether the PCollection belongs to the pipeline being instrumented. Since users can have multiple pipelines in the notebook. When accessing `watch()`ed pipelines, PCollection defined in other pipelines might come into play.
`_write_cache()` and `_read_cache()` rules out this concern by simply checking whether the `pcoll.pipeline` matches `self._pipeline` without worrying about how they are invoked within this `PipelineInstrument`.

If the PCollection belongs to `self._pipeline`, it calculates an expected cache_key to locate the cache (to read or to write). 

I'll remove the debug level log messages since they are just for debugging.

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
KevinGG: It's only used in `instrument`.

A potential change will happen when TestStream preprocess is added.
By then, we shall have 2 pipelines running in-parallel each with its own instrument (first one writes cache, second one polling reads cache). This preprocess will only be done once. The rest instrument logic would be called like this: preprocess(), instrument_first(), run_first_in_background(), instrument_second(). Then the process goes back to normal one pipeline execution route where the pipeline_result of second pipeline execution is returned by InteractiveRunner.

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
KevinGG: It's the same class. It means this method always writes cache for the given PCollection as long as the PCollection belongs to the pipeline and the keyed cache is absent.
This method doesn't concern about whether the PCollection should be cached.

KevinGG:
@aaltay I've added comments and made some change to the original PR in the 2nd commit.
I'll draft a "vocabulary" doc for us to vote and discuss names of different components in an Interactive Beam notebook.
PTAL. Thank you very much!

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
aaltay: Ack. After the vocabulary is agreed upon please update the comments on these files. We can always have improved documentations.

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
aaltay: OK. Please do the necessary clean ups as more PRs are added.

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
aaltay: I understand. But this is very confusing. I do suggest simplification of terms, concepts and improvement of docs. Some of your explanations could be comments here.

I would not block the PR but please make an overall pass to update the readability. Ideally it should be possible for another engineer to come, learn and be able to contribute these code paths as easily as possible.

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
aaltay: Which one it is? Is it the abbreviation or the verb in this context?

aaltay:
@KevinGG thank you for answering my questions. I think this is ready to merge.

However documentation, and concepts are complex. Please make an overall pass and improve this and make it easy for new folks to contribute.

(cc: @davidyan74 @rohdesamuel other folks working in this area.)

PR #9635, opened 2019-09-23
[website] Update Flink runner page to refer to classic instead of legacy runner

tweise:


iemejia:
Oups I accidentally merged without fixing the name of the commit, sorry.

mxm:
Should we give all requested reviewers a chance to review? Why the rush to merge this so quickly?

mxm:
In any case, the updated terminology is much better.

PR #9636, opened 2019-09-23
[BEAM-8299] Upgrade Jackson to version 2.9.10

iemejia:
Jackson 2.9.10 addresses multiple CVE issues from previous Jackson versions, so we need to upgrade it.
R: @pabloem


pabloem:
LGTM.

PR #9642, opened 2019-09-23
[BEAM-8213] Split up monolithic python preCommit tests on jenkins

chadrik:
See the Jira issue for details: https://issues.apache.org/jira/browse/BEAM-8213#

------------------------

--- | --- | --- | --- | ---


chadrik:
R: @lgajowy
R: @kkucharc
R: @echauchot
R: @robertwb 
R: @udim 

Well, I tried my hand at this, but it's not showing the new jobs, so I'm not sure whether there's something I need to update to make the Jenkins config pull from this PR, or if we've got a chicken/egg situation wrt to testing if this new configuration works.

chadrik:
Run PythonLint PreCommit

aaltay:
I believe you need to run a seed job to get the new jobs recognized by Jenkins. 

R: @yifanmai could help with the specifics of seed job.
R: @tvalentyn for reviewing wrt to python 3 jobs.

tvalentyn:
To run a seed job on your PR you can say "Run Seed Jøb" (using ø to avoid triggering the job by this comment). 
After seed job finishes (~10 min), run you can run jenkins jobs defined in this PR. 

Note that seed job launched on the PR will affect other test executions, so consider giving a heads-up on dev. Seed job runs periodically, so at some point we will restore to the job specs using SOT in master.

tvalentyn:
Overall, this LGTM. We need to make sure that trigger phrases work, are visible, and by default all precommits run on python PRs . 

Main concern: this change may increase precommit queue, since each job will require a jenkins slot, of which we have 16 VMs * 2 slots per VM. What required 1 slot will now require 5. @yifanzou  what's your take on this? We may want to increase the amount of slots, and should monitor the precommit queue time after this is merged.

yifanzou:
> We're still doing the same amount of work, so IIUC, assuming we get similar CPU-utilization in this new configuration, these 5 jobs should finish in the time it took the previous single job to finish, plus whatever overhead is required per job to bootstrap the tests.

I think the concern is that we may have higher test concurrency since we splitting the big python job into multiple pieces. The Jenkins job waiting queue cloud be longer. In addition, this may also lead higher stress on gcp resources usage, such as CreateJobPerMinutePerUser, and computing engine quotas.

chadrik:
> I think the concern is that we may have higher test concurrency since we splitting the big python job into multiple pieces. 

I don't really follow this point. We can't have higher concurrency than the 32 slots that are available, right?  Splitting the work previously done by one job across 5 jobs only affects the amount of work done concurrently at times when all the slots in the queue were not previously full, i.e. there were more slots that could have been expanded to, but not enough jobs to fill them.  But even in the case where we might now make use of those free slots (which is a good thing!), the number of slots in use would still be less than or equal to 32.  

I'm new to this project so there are some important details that I may be overlooking, but that's how it looks to me based on my experience with work queues. 

> In addition, this may also lead higher stress on gcp resources usage, such as CreateJobPerMinutePerUser, and computing engine quotas.

After this change it will be slightly more likely that multiple slots are consumed by python jobs than before, since there are now more python job types (if we assume the type of job in each spot is a random sample of all job types).  So in terms of CPU resources, there would be more stress than before if we think that the python jobs are using more CPU resources than the Java and Go jobs, which I find quite unlikely.  However, If we're talking about IO usage, it is plausible that there will be more stress on the system because it will be more likely than before that two or more jobs are accessing the same data for integration tests.

But all-in-all, since we're doing exactly the same work as before, and we have a limit on the number of jobs that can run concurrently, it seems unlikely to me that there will be a significant increase in resources used, time spent waiting for the queue to clear, or stress on the existing resources.

chadrik:
> Note that seed job launched on the PR will affect other test executions, so consider giving a heads-up on dev. 

Can you describe how other jobs will be affected so that I can send out a proper warning, pease?

youngoli:
As far as queue times go, I think the impact of this should be minimal. Like Chad mentioned, although it will take up 5 Jenkins slots, the overall work is equivalent they should take about 1/5 as long (or slightly longer if you include setup time).

I think the main mitigation here is that these are precommits and not postcommits. Since it's not set to run on any kind of regular schedule I'm less worried about sudden spikes in Jenkins usage. And even if the slots do all get full, the shorter time for these precommits means that slots would be available sooner.

So I don't think this is worth blocking on for that reason. But I'd measure how long the tests take to check that splitting the test didn't drastically increase the total amount of work, and I'd keep an eye on overall latency via these two pages:

http://104.154.241.245/d/_TNndF2iz/pre-commit-test-latency?orgId=1
https://builds.apache.org/label/beam/load-statistics?type=min

youngoli:
> Can you describe how other jobs will be affected so that I can send out a proper warning, pease?

If you do seed job then it will update Jenkins and the github setup here in total, not just for this PR. So during that time where your change is live, all PRs will run the split precommits. If those precommits are failing for whatever reason then you'll be blocking those PRs. So ideally you'll want to warn devs first and if it causes problems, the do it again with a clean environment to undo your changes (this also occurs automatically in 8 hour intervals, but that's too long if the precommits are failing).

tvalentyn:
> the overall work is equivalent they should take about 1/5 as long (or slightly longer if you include setup time).
I don't think we will have 1/5 reduction, since currently Python 2.7, 3.5, 3.6, 3.7 precommits are already executing in parallel (by way of gradle parallelism) on the same Jenkins worker taking up one slot, so we will still have a parallel execution but will require 4x more slots. 

Increasing slots per worker may help, but there are some potentially heavy-weight tests, such as portable python precommit tests that bring up Flink, that may cause jenkins VMs to OOM if we run a lot of them in parallel on the same VM. I have heard of a second hand account that parallelizing  portable precommit tests 4x on the same Jenkins worker caused OOMs, but did not verify myself. Perhaps not an issue, but we need a reliable way to monitor Jenkins worker health / utilization to be confident.

chadrik:
> I don't think we will have 1/5 reduction, since currently Python 2.7, 3.5, 3.6, 3.7 precommits are already executing in parallel (by way of gradle parallelism) on the same Jenkins worker taking up one slot, so we will still have a parallel execution but will require 4x more slots.

I believe he was saying that each of the 5 jobs will take on average 1/5th the time it took before.  Each job will continue to run in parallel, so that I think this is basically true, excepting for the fact that there are some differences between the jobs (lint is fast, some python versions run IT, some don't), but I agree in principle.

> Increasing slots per worker may help, but there are some potentially heavy-weight tests, such as portable python precommit tests that bring up Flink, that may cause jenkins VMs to OOM if we run a lot of them in parallel on the same VM.

This PR does not split up the Python_Portable PreCommit tests that start up flink.

chadrik:
Run Seed Job

chadrik:
Run PythonLint PreCommit

tvalentyn:
> I believe he was saying that each of the 5 jobs will take on average 1/5th the time of the current monolithic job. 

The monolythic job already runs 4 out of 5 tasks in parallel - I do not see where 80% speedup will come from. 

> 
> > Increasing slots per worker may help, but there are some potentially heavy-weight tests, such as portable python precommit tests that bring up Flink, that may cause jenkins VMs to OOM if we run a lot of them in parallel on the same VM.
> 
> This PR does not split up the Python_Portable PreCommit tests that start up flink.

My comment refers to a possible mitigation option for the increased demand for Jenkins slots as a result of this PR. Such option is to increase amount of Jenkins slots per Jenkins VM (for example, 5 slots instead of 2 slots). However it can possibly create as scenario where Jenkins VM has to run several high-demanding jobs, and will be overutilized.

chadrik:
Retest this please

chadrik:
@youngoli @tvalentyn   I could use a little help.  I ran the seed job phrase, waited 20 minutes, and tried to retrigger the tests, but no luck.

tvalentyn:
Run Seed Job

chadrik:
> The monolythic job already runs 4 out of 5 tasks in parallel - I do not see where 80% speedup will come from.

I think this is just a communication error.  We're saying that the sum of the running time of all the 5 jobs will be the same or more than the old monolithic job, but each will be on average 1/5th the time of the old monolithic job.  We're not saying that the sum of all 5 jobs will be faster.

tvalentyn:
I now see: Jenkins: "Seed Job Pending — Build triggered for merge commit." - once started, you should be able to monitor status at https://builds.apache.org/job/beam_SeedJob/

chadrik:
> I now see: Jenkins: "Seed Job Pending — Build triggered for merge commit." - once started, you should be able to monitor status at https://builds.apache.org/job/beam_SeedJob/

I don't see the new jobs listed at https://builds.apache.org/job/beam_SeedJob/.  Should I?

tvalentyn:
> I think this is just a communication error. We're saying that the sum of the running time of all the 5 jobs will be the same

I think the walltime run of new jobs will be the same or slightly faster, since lint will run in parallel.

> but each will be on average 1/5th the time of the old monolithic job. 

I don't see why this would be the case - lint job will run much faster, precommits - same time as now minus time to run lint, so slightly faster, but not 1/5th of the time.

tvalentyn:
This is the seed job: https://builds.apache.org/job/beam_SeedJob/4938/. Once this job completes, you should be able to trigger the jobs defined in this PR using respective trigger phrases.

chadrik:
> I don't see why this would be the case - lint job will run much faster, precommits - same time as now minus time to run lint, so slightly faster, but not 1/5th of the time.

I think I see your point now.  Gradle runs tasks in parallel, but after this change each job will have only one or two Gradle tasks, instead of the 7 or 8 it had before.  How many CPUs does each VM have?

We could mitigate this by running nose tests in parallel:  https://nose.readthedocs.io/en/latest/doc_tests/test_multiprocess/multiprocess.html

I've done this with nose and also with pytest using the xdist plugin.

youngoli:
> > The monolythic job already runs 4 out of 5 tasks in parallel - I do not see where 80% speedup will come from.
> 
> I think this is just a communication error. We're saying that the sum of the running time of all the 5 jobs will be the same or more than the old monolithic job, but each individual job will be on average 1/5th the time of the old monolithic job. We're not saying that the sum of all 5 jobs will be faster.

What Valentyn means is that the monolithic job is already running the tests in parallel within one Jenkins slot. I didn't know that, but if that's the case then splitting the tests wouldn't make them finish faster, it would just continue running the tests in parallel, but using 5 Jenkins slots instead of 1.

> Increasing slots per worker may help, but there are some potentially heavy-weight tests, such as portable python precommit tests that bring up Flink, that may cause jenkins VMs to OOM if we run a lot of them in parallel on the same VM. I have heard of a second hand account that parallelizing portable precommit tests 4x on the same Jenkins worker caused OOMs, but did not verify myself. Perhaps not an issue, but we need a reliable way to monitor Jenkins worker health / utilization to be confident.

Is there a way to distribute tests to workers with the lowest resource utilization? Or even better, have resource benchmarks for our various test suites so we can avoid sending resource-intensive tests to workers that don't have enough available resources? I don't really know how Jenkins works so that might be a little advanced, but it would definitely avoid that problem and let us increase the number of slots in the workers without hitting resource limits.

chadrik:
PythonLint completed successfully in 3:40!

chadrik:
Python35, 36, and 37 all finished successfully in about 46 minutes.   Python2 is still going.

So the total is definitely going to be higher.  It was about 75 minutes total before.

chadrik:

chadrik:
Python2 completed successfully in 75 minutes.

chadrik:
All new tests passing.  This PR looks good to go.  

I'll peek in at the [PR for porting tests to pytest](https://github.com/apache/beam/pull/7949) to see about using xdist there to speed these up.

Perhaps in the meantime we can consider more jenkins slots per VM, as discussed.

aaltay:
Should not we address the jenkins slots issue before merging this PR. Based on my understanding so far, we are reducing a 75 min job to 4 40 something jobs. This will increase the queue lengths and will impact the developer experience for everyone. Am I missing something?

youngoli:
No, I think you're right. These tests shouldn't be seperated just yet (with the possible exception of PythonLint. At less than 5 minutes it seems ok to split). I think the best approach is to look into increasing the number of jenkins slots open, along with trying to find some solution that can load balance the resource usage between VMs.

If we can do that, then this change should be a net positive, since it doesn't actually increase resource usage (just jenkins slots used), and splitting up the precommits makes logical sense. I think it would be helpful to find someone who knows a lot about Jenkins to see if that load balancing idea is possible.

chadrik:
> Should not we address the jenkins slots issue before merging this PR. Based on my understanding so far, we are reducing a 75 min job to 4 40 something jobs. This will increase the queue lengths and will impact the developer experience for everyone. Am I missing something?

It's actually worse than that.  The total time for all 5 python jobs was 217 minutes, which is 142 minutes longer than before.  

Heres a break down of the time:

- Python (old): 75
- Metrics: 2
- Go: 3.5
- Java: 56
- JavaBeamZetaSQL: 2.5
- JavaPortability: 14
- Java_Examples_Dataflow: 15
- Portable_Python: 22
- Python_PVR_FLink: 12
- RAT: 0.5
- Spotless: 1
- Website: 5
- Website_Stage_GCS:  4.5

Total: 213 minutes

Adding 142 minutes to this is a 66% increase.  That's pretty bad. 

I'll leave it up to you all to decide the best course of action.  Longer term, I think parallelizing the python tests at the runner level (nose or pytest) is the best solution, but in the short term increasing the number of slots per VM is worth exploring, but not without risks.  Not merging this is the other option.

aaltay:
I support the idea of separating PyLint for immediate benefit. I see this impacts most first time contributors. And separately figuring out how are we going to parallelize the other tests.

tvalentyn:
I looked at Python 2 tests. The reason they are slower is that they run 2 precommit integration tests - batch and streaming wordcount tests, taking up 33 min.
We used to have a task  ":sdks:python:test-suites:dataflow:py37:preCommitIT", and these tests were supposed to run in Python 3.7 in addition to Python 2.7. This was added in https://github.com/apache/beam/pull/8728/. We decided to pick only one Py3 version for these tests to save quota. I looks that we may have stopped running these ITs on Python 3.

tvalentyn:
Nevermind, we still run wordcounts  them on Python 3.7.

tvalentyn:
https://builds.apache.org/job/beam_PreCommit_Python2_Commit/1/testReport/ says Py2 suite completed in 56 min, while it took 75 mi as per https://builds.apache.org/job/beam_PreCommit_Python2_Commit/1, not including waiting time. Not sure where the 20 min difference comes from.

tvalentyn:
But, if we get to run precommit IT suites in parallel, we should significantly lower the execution time. I think it makes sense to move them into separate jenkins job.

chadrik:
> I support the idea of separating PyLint for immediate benefit. I see this impacts most first time contributors. And separately figuring out how are we going to parallelize the other tests.

I might not have time to get back to this until Friday.  What should we do in the meantime?

tvalentyn:
> I think I see your point now. Gradle runs tasks in parallel, but after this change each job will have only one or two Gradle tasks, instead of the 7 or 8 it had before. How many CPUs does each VM have?

We use n1-highmem-16 (16 vCPUs, 104 GB memory)

chadrik:
> We use n1-highmem-16 (16 vCPUs, 104 GB memory)

Ah yeah, that's a lot of CPUs for these python tests.

> But, if we get to run precommit IT suites in parallel, we should significantly lower the execution time. I think it makes sense to move them into separate jenkins job.

Should I try that first and see how it goes, or do the idea of just creating a separate Lint job first?

tvalentyn:
> Is there a way to distribute tests to workers with the lowest resource utilization? Or even better, have resource benchmarks for our various test suites so we can avoid sending resource-intensive tests to workers that don't have enough available resources

We could classify the jobs into high-demand and low-demand, and use separate jenkins worker pools for them. Low-demand can use workers with a lot of slots.

Perhaps we can somehow configure Jenkins to not take more than X jobs of type Y, if that is possible. Folks familiar Jenkins could advise here. Do you have any thoughts @yifanzou ?

tvalentyn:
> > We use n1-highmem-16 (16 vCPUs, 104 GB memory)
> 
> Ah yeah, that's a lot of CPUs for these python tests.
> 
> > But, if we get to run precommit IT suites in parallel, we should significantly lower the execution time. I think it makes sense to move them into separate jenkins job.
> 
> Should I try that first and see how it goes, or do the idea of just creating a separate Lint job first?

+1 for separting linter, we can just do it and since this job completes fast, it will be an immediate benefit with little/no overhead on the submission queue.

Separating Precommit ITs,  and the other precommit tests,  and parallelizing running unit test all seems worthwhile. Increasing a number of slots I think is very possible here, we may need someone from test infrastructure help us to do this without breaking workers and decreasing the submission  queue. CCing a few more folks (@alanmyrvold @yifanzou @markflyhigh )

tvalentyn:
> > Should I try that first and see how it goes, or do the idea of just creating a separate Lint job first?
> 
> +1 for separting linter, we can just do it and since this job completes fast, it will be an immediate benefit with little/no overhead on the submission queue.

Consider doing this in separate PR, to keep the history of jenkins runs on this PR.

tvalentyn:
And, thanks a lot for helping with making precommits faster, @chadrik !

markflyhigh:
I suggest move this discussion to dev@ since we had similar discussions before and many other people also have insights to this problem. 

For me, I don't see big benefit to do this split (1 job to 5 jobs). 
- The painpoints you mentioned about pylint failure doesn't require this change. I agree with the approach to split pylint alone. Similar thing is done in Java (RAT) and we could move pylint into that as well (or put in a separate job).
- For logging, Gradle scan organize logs by task and provide a pretty good UI to surface the error. The link is in Jenkins job page. Did you have a chance to explore that?
- For efficiency, this split will **not shorten** the walltime of the precommit run, on the contrary, adding requirement of extra 4 job slots. Given that those are precommit and the run frequency is very likely high (triggered by each commit push and manual in PR), it's likely to increase the precommit queue time.
- I agree with Daniel's proposal in https://github.com/apache/beam/pull/9642#issuecomment-535241993 if we want to proceed this change.

chadrik:
Closing this until we move to pytest and xdist

PR #9647, opened 2019-09-24
[BEAM-6923] limit gcs buffer size to 1MB for artifact upload

angoenka:

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/runners/flink/flink_runner.gradle
+++ b/runners/flink/flink_runner.gradle
@@ -106,6 +106,7 @@ dependencies {
   compile project(":runners:core-construction-java")
   compile project(":runners:java-fn-execution")
   compile project(":sdks:java:build-tools")
+  compile project(":sdks:java:extensions:google-cloud-platform-core")
   compile library.java.vendored_grpc_1_21_0
   compile library.java.jackson_annotations
   compile library.java.slf4j_api
--- a/runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkJobServerDriver.java
+++ b/runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkJobServerDriver.java
@@ -21,7 +21,9 @@ import javax.annotation.Nullable;
 import org.apache.beam.runners.fnexecution.ServerFactory;
 import org.apache.beam.runners.fnexecution.jobsubmission.JobInvoker;
 import org.apache.beam.runners.fnexecution.jobsubmission.JobServerDriver;
+import org.apache.beam.sdk.extensions.gcp.options.GcsOptions;
 import org.apache.beam.sdk.io.FileSystems;
+import org.apache.beam.sdk.options.PipelineOptions;
 import org.apache.beam.sdk.options.PipelineOptionsFactory;
 import org.kohsuke.args4j.CmdLineException;
 import org.kohsuke.args4j.CmdLineParser;
@@ -59,8 +61,11 @@ public class FlinkJobServerDriver extends JobServerDriver {
 
   public static void main(String[] args) throws Exception {
     // TODO: Expose the fileSystem related options.
+    PipelineOptions options = PipelineOptionsFactory.create();
+    // Limiting gcs upload buffer to reduce memory usage while doing parallel artifact uploads.
+    options.as(GcsOptions.class).setGcsUploadBufferSizeBytes(1024 * 1024);
     // Register standard file systems.
-    FileSystems.setDefaultPipelineOptions(PipelineOptionsFactory.create());
+    FileSystems.setDefaultPipelineOptions(options);
     fromParams(args).run();
   }
 
--- a/runners/spark/build.gradle
+++ b/runners/spark/build.gradle
@@ -59,6 +59,7 @@ dependencies {
   compile project(":runners:core-construction-java")
   compile project(":runners:core-java")
   compile project(":runners:java-fn-execution")
+  compile project(":sdks:java:extensions:google-cloud-platform-core")
   compile library.java.jackson_annotations
   compile library.java.slf4j_api
   compile library.java.joda_time
--- a/runners/spark/src/main/java/org/apache/beam/runners/spark/SparkJobServerDriver.java
+++ b/runners/spark/src/main/java/org/apache/beam/runners/spark/SparkJobServerDriver.java
@@ -20,7 +20,9 @@ package org.apache.beam.runners.spark;
 import org.apache.beam.runners.fnexecution.ServerFactory;
 import org.apache.beam.runners.fnexecution.jobsubmission.JobInvoker;
 import org.apache.beam.runners.fnexecution.jobsubmission.JobServerDriver;
+import org.apache.beam.sdk.extensions.gcp.options.GcsOptions;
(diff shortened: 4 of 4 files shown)


angoenka:
R: @robertwb @pabloem @lgajowy

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/artifact/BeamFileSystemArtifactStagingService.java
```diff
@@ -77,6 +78,13 @@
   public static final String MANIFEST = "MANIFEST";
   public static final String ARTIFACTS = "artifacts";
 
+  private final Semaphore permittedConcurrentWrite;
+
+  public BeamFileSystemArtifactStagingService() {
+    super();
+    permittedConcurrentWrite = new Semaphore(8);
```
pabloem: Maybe make the `8` a constant?

lukecwik:
Note that gRPC does support pushback and signalling to the client via the HTTP/2 protocol it is built on top of, should we be using that?

robertwb:
> Note that gRPC does support pushback and signalling to the client via the HTTP/2 protocol it is built on top of, should we be using that?

If that works, it'd be good to use, but would this apply to multiple concurrent connections, which I think is the issue here?

lukecwik:
Yes since you can keep track of the number of calls and on a per call basis
provide pushback.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/artifact/BeamFileSystemArtifactStagingService.java
angoenka: The name of the variable `permittedConcurrentWrite` seems to be descriptive enough hence i avoided creating a new constant.
Let me know if you think it will be better to create a constant.

angoenka:
I searched a bit but could't find good documentation and usage.
Their is also an ongoing issue open on grps for limiting concurrent connection https://github.com/grpc/grpc-java/issues/1886
Do you have any pointers which we can use?

pabloem:
Fair enough : )

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/artifact/BeamFileSystemArtifactStagingService.java
```diff
@@ -77,6 +78,13 @@
   public static final String MANIFEST = "MANIFEST";
   public static final String ARTIFACTS = "artifacts";
 
+  private final Semaphore permittedConcurrentWrite;
+
+  public BeamFileSystemArtifactStagingService() {
+    super();
```
lukecwik:
```suggestion
```

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/artifact/BeamFileSystemArtifactStagingService.java
```diff
@@ -227,9 +235,11 @@ public void onNext(PutArtifactRequest putArtifactRequest) {
                   encodedFileName(metadata.getMetadata()), StandardResolveOptions.RESOLVE_FILE);
           LOG.debug(
               "Going to stage artifact {} to {}.", metadata.getMetadata().getName(), artifactId);
-          artifactWritableByteChannel = FileSystems.create(artifactId, MimeTypes.BINARY);
           hasher = Hashing.sha256().newHasher();
+          permittedConcurrentWrite.acquire();
```
lukecwik: Since gRPC does a lot on a single thread since it multiplexes multiple RPCs on a single channel, you may be blocking other RPCs.

angoenka: I agree, this might block threads in the pool but they are anyways going to be blocked while the write is happening as we accept connections in parallel.
This take 1 call from a connection which is currently unlimited in Netty.
So I don't think this will have any added impact on the usage of grpc.

angoenka:
> It may be much simpler to set the GCS upload buffer size to 1MiB to solve BEAM-6923 via setting [GcsUploadBufferSize](https://github.com/apache/beam/blob/c2f0d282337f3ae0196a7717712396a5a41fdde1/sdks/java/extensions/google-cloud-platform-core/src/main/java/org/apache/beam/sdk/extensions/gcp/options/GcsOptions.java#L91). I think historically we have used 1MiB in the past while the default is 64MiB

This makes sense. Do you think we should apply this configuration for all the gcs write.
I am a bit concern as it can impact performance of data path of gcs sink. We can enhance the filesystems configuration to allow passing filesystem specific options.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/artifact/BeamFileSystemArtifactStagingService.java
lukecwik: The FileSystem may push the write to another thread unblocking the gRPC thread.

lukecwik:
> > It may be much simpler to set the GCS upload buffer size to 1MiB to solve BEAM-6923 via setting [GcsUploadBufferSize](https://github.com/apache/beam/blob/c2f0d282337f3ae0196a7717712396a5a41fdde1/sdks/java/extensions/google-cloud-platform-core/src/main/java/org/apache/beam/sdk/extensions/gcp/options/GcsOptions.java#L91). I think historically we have used 1MiB in the past while the default is 64MiB
> 
> This makes sense. Do you think we should apply this configuration for all the gcs write.
> I am a bit concern as it can impact performance of data path of gcs sink. We can enhance the filesystems configuration to allow passing filesystem specific options.

In portable pipeline execution we don't expect to run user transforms for reading/writing from GCS and this would only be for Runner interactions with GCS so I believe we could make it 1MiB for the GcsFileSystem by default.

angoenka:
> > > It may be much simpler to set the GCS upload buffer size to 1MiB to solve BEAM-6923 via setting [GcsUploadBufferSize](https://github.com/apache/beam/blob/c2f0d282337f3ae0196a7717712396a5a41fdde1/sdks/java/extensions/google-cloud-platform-core/src/main/java/org/apache/beam/sdk/extensions/gcp/options/GcsOptions.java#L91). I think historically we have used 1MiB in the past while the default is 64MiB
> > 
> > 
> > This makes sense. Do you think we should apply this configuration for all the gcs write.
> > I am a bit concern as it can impact performance of data path of gcs sink. We can enhance the filesystems configuration to allow passing filesystem specific options.
> 
> In portable pipeline execution we don't expect to run user transforms for reading/writing from GCS and this would only be for Runner interactions with GCS so I believe we could make it 1MiB for the GcsFileSystem by default.

Makes sense. 
Made the change to limit the buffer usage only for flink and spark job server

pabloem:
Run Java PreCommit

PR #9673, opened 2019-09-26
[BEAM-7933] Add job server request timeout (default to 60 seconds)

ecanzonieri:
Add a new option --job_server_timeout that default to 60 seconds. The request timeout is user for all job server requests (with exception to the ones that are expected to last long or hang). The request timeout is also used upon channel creation so that the Beam driver will fail if the job server is not available.

Let me know if the option name looks fine or it needs renaming. Finally, there are other grpc requests (non job service) that I'm not covering as part of this pr. Let me know if I should add the timeout to those as well.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/options/pipeline_options.py
+++ b/sdks/python/apache_beam/options/pipeline_options.py
@@ -818,11 +818,17 @@ class PortableOptions(PipelineOptions):
   """
   @classmethod
   def _add_argparse_args(cls, parser):
-    parser.add_argument('--job_endpoint',
-                        default=None,
-                        help=
-                        ('Job service endpoint to use. Should be in the form '
-                         'of address and port, e.g. localhost:3000'))
+    parser.add_argument(
+        '--job_endpoint', default=None,
+        help=('Job service endpoint to use. Should be in the form of address '
+              'and port, e.g. localhost:3000'))
+    parser.add_argument(
+        '--job-server-timeout', default=60, type=int,
+        help=('Job service request timeout in seconds. The timeout '
+              'determines the max time the driver program will wait to '
+              'get a response from the job server. NOTE: the timeout does not '
+              'apply to the actual pipeline run time. The driver program can '
+              'still wait for job completion indefinitely.'))
     parser.add_argument(
         '--environment_type', default=None,
         help=('Set the default environment type for running '
--- a/sdks/python/apache_beam/runners/portability/local_job_service.py
+++ b/sdks/python/apache_beam/runners/portability/local_job_service.py
@@ -89,7 +89,7 @@ class LocalJobServicer(beam_job_api_pb2_grpc.JobServiceServicer):
     if os.path.exists(self._staging_dir) and self._cleanup_staging_dir:
       shutil.rmtree(self._staging_dir, ignore_errors=True)
 
-  def Prepare(self, request, context=None):
+  def Prepare(self, request, context=None, timeout=None):
     # For now, just use the job name as the job id.
     logging.debug('Got Prepare request.')
     preparation_id = '%s-%s' % (request.job_name, uuid.uuid4())
@@ -121,13 +121,13 @@ class LocalJobServicer(beam_job_api_pb2_grpc.JobServiceServicer):
         artifact_staging_endpoint=self._artifact_staging_endpoint,
         staging_session_token=preparation_id)
 
-  def Run(self, request, context=None):
+  def Run(self, request, context=None, timeout=None):
     job_id = request.preparation_id
     logging.info("Runing job '%s'", job_id)
     self._jobs[job_id].start()
     return beam_job_api_pb2.RunJobResponse(job_id=job_id)
 
-  def GetJobs(self, request, context=None):
+  def GetJobs(self, request, context=None, timeout=None):
     return beam_job_api_pb2.GetJobsResponse(
         [job.to_runner_api(context) for job in self._jobs.values()])
 
@@ -135,16 +135,16 @@ class LocalJobServicer(beam_job_api_pb2_grpc.JobServiceServicer):
     return beam_job_api_pb2.GetJobStateResponse(
         state=self._jobs[request.job_id].state)
 
-  def GetPipeline(self, request, context=None):
+  def GetPipeline(self, request, context=None, timeout=None):
     return beam_job_api_pb2.GetJobPipelineResponse(
         pipeline=self._jobs[request.job_id]._pipeline_proto)
 
-  def Cancel(self, request, context=None):
+  def Cancel(self, request, context=None, timeout=None):
     self._jobs[request.job_id].cancel()
     return beam_job_api_pb2.CancelJobRequest(
         state=self._jobs[request.job_id].state)
 
-  def GetStateStream(self, request, context=None):
+  def GetStateStream(self, request, context=None, timeout=None):
     """Yields state transitions since the stream started.
       """
     if request.job_id not in self._jobs:
@@ -154,7 +154,7 @@ class LocalJobServicer(beam_job_api_pb2_grpc.JobServiceServicer):
     for state in job.get_state_stream():
       yield beam_job_api_pb2.GetJobStateResponse(state=state)
 
-  def GetMessageStream(self, request, context=None):
+  def GetMessageStream(self, request, context=None, timeout=None):
     """Yields messages since the stream started.
       """
     if request.job_id not in self._jobs:
@@ -169,7 +169,7 @@ class LocalJobServicer(beam_job_api_pb2_grpc.JobServiceServicer):
         resp = beam_job_api_pb2.JobMessagesResponse(message_response=msg)
       yield resp
 
-  def DescribePipelineOptions(self, request, context=None):
+  def DescribePipelineOptions(self, request, context=None, timeout=None):
     return beam_job_api_pb2.DescribePipelineOptionsResponse()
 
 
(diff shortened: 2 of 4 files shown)


ecanzonieri:
R: @ibzib @aaltay

sdks/python/apache_beam/options/pipeline_options.py
```diff
@@ -808,11 +808,15 @@ class PortableOptions(PipelineOptions):
...
-                        default=None,
-                        help=
-                        ('Job service endpoint to use. Should be in the form '
-                         'of address and port, e.g. localhost:3000'))
+    parser.add_argument(
+        '--job_endpoint', default=None,
+        help=('Job service endpoint to use. Should be in the form of address '
+              'and port, e.g. localhost:3000'))
+    parser.add_argument(
+            '--job_server_timeout', default=60, type=int,
```
ibzib: indentation here looks off

sdks/python/apache_beam/runners/portability/portable_runner.py
```diff
@@ -251,7 +253,11 @@ def send_options_request(max_retries=5):
           # This reports channel is READY but connections may fail
           # Seems to be only an issue on Mac with port forwardings
           return job_service.DescribePipelineOptions(
-              beam_job_api_pb2.DescribePipelineOptionsRequest())
+              beam_job_api_pb2.DescribePipelineOptionsRequest(),
+                  timeout=portable_options.job_server_timeout)
+        except grpc.FutureTimeoutError:
+          # no retry for timeout errors
```
ibzib: Are we sure we don't want retries? We could do exponential backoff, similar to what SubprocessJobServer already does.

sdks/python/apache_beam/runners/portability/portable_runner.py
```diff
@@ -141,23 +141,25 @@ def _create_environment(options):
           payload=(portable_options.environment_config.encode('ascii')
                    if portable_options.environment_config else None))
 
-  def default_job_server(self, options):
+  def default_job_server(self, portable_options):
```
ibzib: why this change?

sdks/python/apache_beam/runners/portability/portable_runner.py
```diff
@@ -141,23 +141,25 @@ def _create_environment(options):
...
     # TODO Provide a way to specify a container Docker URL
     # https://issues.apache.org/jira/browse/BEAM-6328
     if not self._dockerized_job_server:
       self._dockerized_job_server = job_server.StopOnExitJobServer(
           job_server.DockerizedJobServer())
     return self._dockerized_job_server
 
-  def create_job_service(self, options):
-    job_endpoint = options.view_as(PortableOptions).job_endpoint
+  def create_job_service(self, portable_options):
```
ibzib: I'd just leave this as `options`.

sdks/python/apache_beam/runners/portability/portable_runner.py
```diff
@@ -240,7 +242,7 @@ def run_pipeline(self, pipeline, options):
             known_runner_urns=flink_known_urns,
             partial=True)
 
-    job_service = self.create_job_service(options)
+    job_service = self.create_job_service(portable_options)
```
ibzib: Not sure this is necessary?

sdks/python/apache_beam/runners/portability/portable_runner.py
```diff
@@ -251,7 +253,11 @@ def send_options_request(max_retries=5):
           # This reports channel is READY but connections may fail
           # Seems to be only an issue on Mac with port forwardings
           return job_service.DescribePipelineOptions(
-              beam_job_api_pb2.DescribePipelineOptionsRequest())
+              beam_job_api_pb2.DescribePipelineOptionsRequest(),
+                  timeout=portable_options.job_server_timeout)
```
ibzib: I think this line is causing tests to fail. Also, we're already retrying this request. So I would either keep the existing logic or add the new timeout, but not both.

sdks/python/apache_beam/runners/portability/portable_runner.py
ecanzonieri: I changed it because it is actually immediately converted back to portable_options. Basically the caller takes options and converts it to portable_options, then to the create_job_service method it passes options which is again converted to portable_options. It seems to be that we could pass portable_options in the first place and simplify the code a bit. However, I don't have a strong option, I'll revert back to options.

sdks/python/apache_beam/runners/portability/portable_runner.py
ecanzonieri: My guess is that this is invoking the LocalJobServicer that doesn't have the timeout argument. I'll try to repro locally to verify. If that's the case I'd like to add the timeout argument to all LocalJobServicer methods.

sdks/python/apache_beam/runners/portability/portable_runner.py
ecanzonieri: My guess is that we want to retry for other errors, but we should treat timeout as unretriable, so that we never wait for longer than the specified timeout (including retries). We could implement exponential backoff so that it will overall still wait for no longer than the timeout, but I'm not sure it's worth the additional complexity.

sdks/python/apache_beam/runners/portability/portable_runner.py
ecanzonieri: I reverted it.

sdks/python/apache_beam/runners/portability/portable_runner.py
ecanzonieri: I fixed the tests by adding the timeout to the stubs of the LocalJobServicer.

sdks/python/apache_beam/runners/portability/portable_runner.py
ibzib: Okay.

sdks/python/apache_beam/runners/portability/portable_runner.py
ibzib: Adding the timeout to the stubs fixes the tests, but are you sure adding the timeout like that is actually effective? In all the other places you changed, you put timeout in the requests, but here you put it in the service method, which I assumed was the problem. ie `DescribePipelineOptions(DescribePipelineOptionsRequest(), timeout=...)` should be  `DescribePipelineOptions(DescribePipelineOptionsRequest(timeout=...))`

sdks/python/apache_beam/runners/portability/local_job_service.py
```diff
@@ -169,7 +169,7 @@ def GetMessageStream(self, request, context=None):
         resp = beam_job_api_pb2.JobMessagesResponse(message_response=msg)
       yield resp
 
-  def DescribePipelineOptions(self, request, context=None):
+  def DescribePipelineOptions(self, request, context=None, timeout=None):
```
ibzib: It doesn't look like the `timeout` argument is actually used -- see other comment below

sdks/python/apache_beam/runners/portability/portable_runner.py
ecanzonieri: hmmm, I may be missing something but the timeout is never added to the request. My understanding is that timeout itself is not a request parameter but rather a grpc method call parameter. @ibzib can you point me to the code where you see me adding the timeout to the request? Perhaps, I accidentally added it to the wrong place. 
I'm no grpc expert, but I've ran some manual tests on this feature:
1) run python driver without job server running and default timeout -> The driver times out as expected after 60 seconds.
2) run python driver without job server running and change timeout -> The driver times out as expected after the specified timeout.
3) run python driver with job server running but code changed to slow down Prepare() -> The driver times out as expected.
4) run python driver with long running streaming job (Run() method) -> the driver program never times out (note this is the expected behavior since the driver is stuck waiting for completion.

ibzib: Oh, I read everything wrong. :sweat_smile: 

It'd be great to have real tests to verify all these timeouts, but I guess they would not make suitable unit tests, as they would all require at least 1 second to complete.

ibzib:
Please fix lint errors.

ecanzonieri:
Yeah, sorry about the lint errors. I still need to find a good configuration for PyCharm that actually works for Beam.

ecanzonieri:
Apparently `:sdks:python:test-suites:tox:py2:docs` failed. I'm locally to check whether related to my pr.

ibzib:
Docs failure is probably https://issues.apache.org/jira/browse/BEAM-8286

ibzib:
Run Python PreCommit

sdks/python/apache_beam/options/pipeline_options.py
```diff
@@ -818,11 +818,15 @@ class PortableOptions(PipelineOptions):
...
-                         'of address and port, e.g. localhost:3000'))
+    parser.add_argument(
+        '--job_endpoint', default=None,
+        help=('Job service endpoint to use. Should be in the form of address '
+              'and port, e.g. localhost:3000'))
+    parser.add_argument(
+        '--job-server-timeout', default=60, type=int,
+        help=('Job service request timeout in seconds. The timeout '
+              'determines the max time the driver program will wait to '
+              'get a response from the job server.'))
```
ibzib: Might want to make a note here that timeouts do not apply to the actual pipeline itself.

ecanzonieri:
I messed up the rebase. I'll fix it and repush.

ecanzonieri:
Squashed everything to a single commit now.

PR #9703, opened 2019-10-01
[BEAM-6995] Beam basic aggregation rule only when not windowed

11moon11:
Beam basic aggregation rule should not be applied on Calc, Project, and Filter when their parents/they utilize windowed functions.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rule/BeamBasicAggregationRule.java
+++ b/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rule/BeamBasicAggregationRule.java
@@ -17,14 +17,23 @@
  */
 package org.apache.beam.sdk.extensions.sql.impl.rule;
 
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
 import org.apache.beam.sdk.extensions.sql.impl.rel.BeamAggregationRel;
 import org.apache.beam.sdk.extensions.sql.impl.rel.BeamLogicalConvention;
 import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.plan.RelOptRule;
 import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.plan.RelOptRuleCall;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.plan.volcano.RelSubset;
 import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rel.RelNode;
 import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rel.core.Aggregate;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rel.core.Calc;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rel.core.Filter;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rel.core.Project;
 import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rel.core.RelFactories;
-import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rel.core.TableScan;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rex.RexCall;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rex.RexNode;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.sql.SqlKind;
 import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.tools.RelBuilderFactory;
 
 /**
@@ -40,15 +49,22 @@ public class BeamBasicAggregationRule extends RelOptRule {
 
   public BeamBasicAggregationRule(
       Class<? extends Aggregate> aggregateClass, RelBuilderFactory relBuilderFactory) {
-    super(operand(aggregateClass, operand(TableScan.class, any())), relBuilderFactory, null);
+    super(operand(aggregateClass, operand(RelNode.class, any())), relBuilderFactory, null);
   }
 
   @Override
   public void onMatch(RelOptRuleCall call) {
     Aggregate aggregate = call.rel(0);
-    TableScan tableScan = call.rel(1);
+    RelNode relNode = call.rel(1);
 
-    RelNode newTableScan = tableScan.copy(tableScan.getTraitSet(), tableScan.getInputs());
+    if (relNode instanceof Project || relNode instanceof Calc || relNode instanceof Filter) {
+      if (isWindowed(relNode) || hasWindowedParents(relNode)) {
+        // This case is expected to get handled by the 'BeamAggregationRule'
+        return;
+      }
+    }
+
+    RelNode newTableScan = relNode.copy(relNode.getTraitSet(), relNode.getInputs());
 
     call.transformTo(
         new BeamAggregationRel(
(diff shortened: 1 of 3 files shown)


11moon11:
R: @apilloud

apilloud:
Run Direct Runner Nexmark Tests

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rule/BeamBasicAggregationRule.java
```diff
@@ -40,15 +50,21 @@
 
   public BeamBasicAggregationRule(
       Class<? extends Aggregate> aggregateClass, RelBuilderFactory relBuilderFactory) {
-    super(operand(aggregateClass, operand(TableScan.class, any())), relBuilderFactory, null);
+    super(operand(aggregateClass, operand(AbstractRelNode.class, any())), relBuilderFactory, null);
```
apilloud: Looking at examples of this in Calcite, I think `RelNode` is preferable to `AbstractRelNode`.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rule/BeamBasicAggregationRule.java
```diff
@@ -40,15 +50,21 @@
...
   @Override
   public void onMatch(RelOptRuleCall call) {
     Aggregate aggregate = call.rel(0);
-    TableScan tableScan = call.rel(1);
+    AbstractRelNode relNode = call.rel(1);
 
-    RelNode newTableScan = tableScan.copy(tableScan.getTraitSet(), tableScan.getInputs());
+    if (relNode instanceof Project || relNode instanceof Calc || relNode instanceof Filter) {
+      if (isWindowed(relNode) || hasWindowedParents(relNode)) {
+        return;
```
apilloud: Probably worth adding a comment here that this case is expected to be handled by `BeamAggregationRule`.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rule/BeamBasicAggregationRule.java
11moon11: You are correct, updated match condition to use RelNode instead

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rule/BeamBasicAggregationRule.java
11moon11: Added a comment.

apilloud:
Run Direct Runner Nexmark Tests

11moon11:
Run Direct Runner Nexmark Tests

apilloud:
can you do a `git pull origin && git rebase origin/master` on this?

amaliujia:
cc: @amaliujia

sdks/java/extensions/sql/src/test/java/org/apache/beam/sdk/extensions/sql/BeamSqlDslAggregationTest.java
```diff
@@ -701,7 +700,6 @@ public void testSupportsAggregationWithoutProjection() throws Exception {
   }
 
   @Test
-  @Ignore("https://issues.apache.org/jira/browse/BEAM-8317")
   public void testSupportsAggregationWithFilterWithoutProjection() throws Exception {
```
amaliujia: @11moon11 @apilloud 

What I really want to propose is when we add new test cases with SQL queries, run the test for both dialects unless there is a query syntax mismatch.

Using which planner is controlled by https://github.com/apache/beam/blob/master/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/BeamSqlPipelineOptions.java#L28.


I am looking for a way of `@RunWith(Parameterized.class)` so it's easy to run tests for both dialect by an annotation.

11moon11: Found a useful reference link with examples: https://github.com/Pragmatists/JUnitParams/blob/master/src/test/java/junitparams/usage/SamplesOfUsageTest.java

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamAggregationRel.java
```diff
@@ -84,6 +85,31 @@ public BeamAggregationRel(
...
+        traits,
+        child,
+        indicator,
+        groupSet,
+        groupSets,
+        aggCalls,
+        windowFn,
+        windowFieldIndex);
+
+    this.rowType = rowType;
```
amaliujia: Where is this rowType used?

amaliujia: RelNode type is already inferred from input nodes. Usually when you need to use it, you can use getRowType() function to get it than save it as a class member.

11moon11: There is a JIRA issue: https://jira.apache.org/jira/browse/BEAM-7609, when running queries with "SELECT DISTINCT + JOIN", resulting field names are not assigned proper name.
Even though it does not solve this particular issue, preserving the rowType should not hurt (where previously it would just ignore it and set it to null).

11moon11: I see, in that case I will remove this constructor. Do you think adding `deriveRowType();` to the original constructor makes sense or it would be redundant?

amaliujia: I think `deriveRowType()` can be called when it's needed?

11moon11:
Run JavaBeamZetaSQL PreCommit

sdks/java/extensions/sql/src/test/java/org/apache/beam/sdk/extensions/sql/BeamSqlDslAggregationTest.java
apilloud: Cool. I agree this is a good idea, but we should hold off on doing this until #9737 is in. We will need to relocate common tests into another package.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamAggregationRel.java
11moon11: Agreed, reverted BeamAggregationRel back.

apilloud:
LGTM

sdks/java/extensions/sql/src/test/java/org/apache/beam/sdk/extensions/sql/BeamSqlDslAggregationTest.java
amaliujia: @apilloud 

I don't agree though. At least a duplicate test can be created but run for ZetaSQL only and then we can have a migration.

11moon11: I tried adding `pipeline.getOptions().as(BeamSqlPipelineOptions.class).setPlannerName("org.apache.beam.sdk.extensions.sql.zetasql.ZetaSQLPlannerImpl");` to the test to see if it will work using `ZetaSqlPlanner`, but I get `Class not found exception`.
I assume it is because ZetaSQL is not in the build file. After attempting to add a dependency there is the following error: `Circular dependency`, probably because ZetaSQL depends on BeamSQL?

PR #9720, opened 2019-10-02
[BEAM-8335] Add initial modules for interactive streaming support

rohdesamuel:
- Adds the InteractiveService proto and impl for sending remote scripts to a TestStream
- Adds the StreamingCache that knows how to read from an underlying cache and correctly emit elements to the TestStream

------------------------

--- | --- | --- | --- | ---


rohdesamuel:
run PythonLint

rohdesamuel:
Run PythonLint PreCommit

rohdesamuel:
R: @robertwb can you please review?

model/fn-execution/src/main/proto/beam_interactive_api.proto
```diff
@@ -0,0 +1,106 @@
...
+
+import "beam_runner_api.proto";
+import "google/protobuf/timestamp.proto";
+
+service InteractiveService {
+
+  // A TestStream will request for events using this RPC.
+  rpc Events(EventsRequest) returns (stream EventsResponse) {}
+
+  rpc Start (StartRequest) returns (StartResponse) {}
```
robertwb: Why do we need a GRPC service for this? My understanding was that TestStream could be backed by a simple Grpc service that just streamed TestStreamPayload.Event messages.

model/fn-execution/src/main/proto/beam_interactive_api.proto
```diff
@@ -0,0 +1,106 @@
...
+package org.apache.beam.model.fn_execution.v1;
+
+option go_package = "fnexecution_v1";
+option java_package = "org.apache.beam.model.fnexecution.v1";
+option java_outer_classname = "BeamInteractiveApi";
+
+import "beam_runner_api.proto";
+import "google/protobuf/timestamp.proto";
+
+service InteractiveService {
```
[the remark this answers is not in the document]
rohdesamuel: The Stop request acts as the Reset. Added comments to make this more clear.

model/fn-execution/src/main/proto/beam_interactive_api.proto
rohdesamuel: We need a GRPC to simultaneously allow for streaming of events to the TestStream and for user interactivity in a portable way. When the TestStream is implemented in the FnApiRunner, there will be a hard container boundary to both the runner and worker that needs to be crossed with some IPC. GRPC is the best way for that, currently.

model/fn-execution/src/main/proto/beam_interactive_api.proto
```diff
@@ -0,0 +1,106 @@
+/*
```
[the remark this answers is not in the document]
rohdesamuel: Moved to model/interactive. Also, this file can work for Batch and Streaming. Technically, it will only be used for Batch pipelines because the interactive runner will be running streaming pipelines as Batch during re-run.

sdks/python/apache_beam/runners/interactive/caching/streaming_cache.py
```diff
@@ -0,0 +1,123 @@
...
+def to_timestamp(timestamp_secs):
+  """Converts seconds since epoch to an apache_beam.util.Timestamp.
+  """
+  return timestamp.Timestamp.of(timestamp_secs)
+
+def from_timestamp_proto(timestamp_proto):
+  return timestamp.Timestamp(seconds=timestamp_proto.seconds,
+                             micros=timestamp_proto.nanos * 1000)
+
+def to_timestamp_usecs(ts):
```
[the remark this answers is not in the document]
rohdesamuel: I moved the methods to be on the timestamp.Timestamp object.

sdks/python/apache_beam/runners/interactive/caching/streaming_cache.py
```diff
@@ -0,0 +1,123 @@
...
+    """Abstraction that reads from PCollection readers.
+
+    This class is an Abstraction layer over multiple PCollection readers to be
+    used for supplying the Interactive Service with TestStream events.
+
+    This class is also responsible for holding the state of the clock, injecting
+    clock advancement events, and watermark advancement events.
+    """
+    def __init__(self, readers):
+      self._readers = [reader.read() for reader in readers]
```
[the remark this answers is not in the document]
rohdesamuel: No, the elements in the self._readers are generators that produce the events we want. These fit the interface in https://github.com/apache/beam/pull/8884 for the PCollectionCache.

sdks/python/apache_beam/runners/interactive/caching/streaming_cache.py
```diff
@@ -0,0 +1,123 @@
...
+      self._watermark = timestamp.MIN_TIMESTAMP
```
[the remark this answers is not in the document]
rohdesamuel: Good point, I forgot to update this to be per PCollection tag.

sdks/python/apache_beam/runners/interactive/caching/streaming_cache.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def advance_processing_time(self, processing_time, events):
+      """Advances the internal clock state and injects an AdvanceProcessingTime
+         event.
+      """
+      if self._timestamp != processing_time:
+        duration = timestamp.Duration(
+            micros=processing_time.micros - self._timestamp.micros)
+        if self._timestamp == timestamp.MIN_TIMESTAMP:
+          duration = timestamp.Duration(micros=processing_time.micros)
```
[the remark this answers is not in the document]
rohdesamuel: Good catch, I simplified the implementation to assume the clock and self._timestamp starts at t=0.

sdks/python/apache_beam/testing/interactive_stream.py
```diff
@@ -0,0 +1,123 @@
...
+  if state == 'STOPPED':
+    return beam_interactive_api_pb2.StatusResponse.STOPPED
+  if state == 'PAUSED':
+    return beam_interactive_api_pb2.StatusResponse.PAUSED
+  return beam_interactive_api_pb2.StatusResponse.RUNNING
+
+class InteractiveStreamController(InteractiveServiceServicer):
+  def __init__(self, endpoint, streaming_cache):
+    self._endpoint = endpoint
+    self._server = grpc.server(ThreadPoolExecutor(max_workers=10))
```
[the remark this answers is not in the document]
rohdesamuel: Probably not. I dropped it down to 2. One thread to handle the TestStream persistent connection and another to handle any user requests.

sdks/python/apache_beam/testing/interactive_stream.py
```diff
@@ -0,0 +1,123 @@
...
+    resp.state = to_api_state(self._state)
+    return resp
+
+  def _reset_state(self):
+    self._reader = None
+    self._playback_speed = 1.0
+    self._state = 'STOPPED'
+
+  def _next_state(self, state):
+    if not self._state or self._state == 'STOPPED':
```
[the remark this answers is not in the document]
rohdesamuel: deleted "not self._state"

sdks/python/apache_beam/testing/interactive_stream.py
```diff
@@ -0,0 +1,123 @@
...
+import time
+
+from apache_beam.portability.api import beam_interactive_api_pb2
+from apache_beam.portability.api import beam_interactive_api_pb2_grpc
+from apache_beam.portability.api.beam_interactive_api_pb2_grpc import InteractiveServiceServicer
+from concurrent.futures import ThreadPoolExecutor
+
+
+def to_api_state(state):
+  if state == 'STOPPED':
```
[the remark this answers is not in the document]
rohdesamuel: done

model/fn-execution/src/main/proto/beam_interactive_api.proto
robertwb: Sorry for not being more clear. 

Currently, we have a TestStream that reads its elements from memory. We are hoping to augment that so that its elements can be supplied externally, over GRPC, so we have

TestStreamImpl <--- TestStreamPayload.Event ---- Driver

Where TestStreamImpl and Driver sit in separate processes. To do this, we should augment TestStreamPayload with an (optional) address (you can probably reuse Endpoint) to which it should connect over GRPC to get a stream of TestStreamPayload.Events. 

This Service seems to be trying to split Driver across processes, which seems to introduce unnecessary complexity.

model/interactive/src/main/proto/beam_interactive_api.proto
```diff
@@ -0,0 +1,124 @@
...
+message StatusResponse {
+
+  // The current timestamp of the replay stream. Is MIN_TIMESTAMP when state
+  // is STOPPED.
+  google.protobuf.Timestamp stream_time = 1;
+
+  // The minimum watermark across all of the faked replayable unbounded sources.
+  // Is MIN_TIMESTAMP when state is STOPPED.
+  google.protobuf.Timestamp watermark = 2;
+
```
davidyan74: Do we need more info in the StatusResponse like:
1. the latest timestamp of the replayable stream (i.e. the current time if the recording job is still running, or the time when the recording job stopped)
2. the current playback speed if running (either playback speed of StartRequest when stream_time < latest timestamp of the replayable stream, or 1 if stream_time == latest timestamp of the replayable stream and the recording job is still running, or 0 if not running).

I think these two fields will be valuable for the user to find out more about the replay.

rohdesamuel: My intention is that this will be used to interact with the TestStream executing in the runner, but I see no harm with adding extra information here.

model/fn-execution/src/main/proto/beam_interactive_api.proto
rohdesamuel: Thanks for your clarification. We are actually on the same page. To clarify my position, the service will be sitting in the InteractiveRunner and the TestStream will ask for events from it using the EventsRequest. A user can then start replaying the stream using the StartRequest. At a later time, the user can also pause or prematurely stop the replay.

I added an endpoint to the TestStreamPayload.

model/interactive/src/main/proto/beam_interactive_api.proto
davidyan74: Thanks. I think the information is necessary to explain to the user why e.g. can't replay past a certain time, or why the current playback speed != the playback speed they specify in StartRequest.

model/interactive/src/main/proto/beam_interactive_api.proto
```diff
@@ -0,0 +1,130 @@
...
+    // The InteractiveService is paused from replaying events. Goes to RUNNING
+    // with either a StartRequest or a StepRequest. Goes to STOPPED with a
+    // StopRequest.
+    PAUSED = 2;
+  }
+  State state = 5;
+}
+
+message StepRequest { }
```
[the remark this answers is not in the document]
rohdesamuel: Done.

model/interactive/src/main/proto/beam_interactive_api.proto
```diff
@@ -0,0 +1,130 @@
...
+  rpc Stop (StopRequest) returns (StopResponse) {}
+
+  // Pauses the stream of events to the EventsRequest. If there is already an
+  // outstanding EventsRequest streaming events, then the stream will pause
+  // after the EventsResponse is completed.
+  rpc Pause (PauseRequest) returns (PauseResponse) {}
+
+  // Sends a single element to the EventsRequest then closes the stream.
+  rpc Step (StepRequest) returns (StepResponse) {}
+
```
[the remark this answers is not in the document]
rohdesamuel: Good point, I added the option to either start at a specified timestamp or to advance the stream in the StartRequest.

model/interactive/src/main/proto/beam_interactive_api.proto
```diff
@@ -0,0 +1,145 @@
...
+  double playback_speed = 1;
+
+  oneof starting_placement {
+    // (Optional) if present, will start the stream at the specified timestamp.
+    google.protobuf.Timestamp start_at = 2;
+
+    // (Optional) if present, will advance the stream by replaying events as
+    // quickly as possible until the stream timestamp has advanced by the
+    // specified amount.
+    google.protobuf.Duration advance_by = 3;
```
davidyan74: I think we want to make that a separate method because the user might want to advance to an arbitrary timestamp during the course of the replay, and they might want to do so multiple times (e.g. start replay, ..., advance to time A, ... advance to time B, ...).

sdks/python/apache_beam/testing/interactive_stream.py
```diff
@@ -0,0 +1,125 @@
...
+        self._reader = self._streaming_cache.reader()
+    elif self._state == 'RUNNING':
+      if state == 'STOPPED':
+        self._reset_state()
+    self._state = state
+
+  def Events(self, request, context):
+    # The TestStream will wait until the stream starts.
+    while self._state != 'RUNNING' and self._state != 'STEP':
+      time.sleep(0.01)
```
aaltay: Would it make sense to increase the delay here? I am worried that 0.01 might be too frequent for a check.

sdks/python/apache_beam/testing/interactive_stream.py
```diff
@@ -0,0 +1,125 @@
...
+              e.processing_time_event.advance_duration * self._playback_speed
+              ) * 10**-6
+          time.sleep(sleep_duration)
+        yield beam_interactive_api_pb2.EventsResponse(events=[e])
+    else:
+      resp = beam_interactive_api_pb2.EventsResponse()
+      resp.end_of_stream = True
+      self._next_state('STOPPED')
+      yield resp
+      return
```
aaltay: You may not need this return. It is probably fine to have it here for clarity.

sdks/python/apache_beam/utils/timestamp.py
```diff
@@ -29,6 +29,7 @@
 
 import dateutil.parser
 import pytz
+from google.protobuf import timestamp_pb2
```
aaltay: We should not have this import here. Google dependencies are installed optionally and may not be available. (Or is this coming through just protobuf dependency?)

rohdesamuel: This is just through the protobuf dependency

model/interactive/src/main/proto/beam_interactive_api.proto
rohdesamuel: Ack, I changed it into its own method. When a user sends an AdvanceRequest, the stream will run as fast as possible to the specified offset then pause.

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -500,6 +500,7 @@ message TestStreamPayload {
 
     message AdvanceWatermark {
       int64 new_watermark = 1;
+      string tag = 2;
```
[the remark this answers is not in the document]
rohdesamuel: It will behave as expected: if unspecified then it will default to the Main PCollection Input. If specified then it will emit to that tagged PCollection. It should throw an exception if the tag doesn't exist. I added a comment to clarify the expected behavior.

sdks/python/apache_beam/testing/interactive_stream.py
rohdesamuel: Ack, changed to 0.25. I want to balance the feeling of "snappiness" and not wasting the user's CPU here.

sdks/python/apache_beam/testing/interactive_stream.py
rohdesamuel: ack

sdks/python/apache_beam/testing/interactive_stream_test.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def get_open_port():
+  import socket
+  s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+  s.bind(('', 0))
+  s.listen(1)
+  port = s.getsockname()[1]
+  s.close()
+  return port
```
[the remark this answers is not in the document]
rohdesamuel: Awesome thanks for this. I didn't know about this functionality.

model/interactive/src/main/proto/beam_interactive_api.proto
```diff
@@ -81,7 +86,6 @@ message PauseResponse { }
```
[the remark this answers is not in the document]
rohdesamuel: done

model/interactive/src/main/proto/beam_interactive_api.proto
```diff
@@ -94,6 +98,7 @@ message StatusResponse {
   // no recording.
   google.protobuf.Timestamp recording_time = 3;
 
+  // The set playback_speed from either the StartRequest or the AdvanceRequest.
```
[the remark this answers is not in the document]
rohdesamuel: Does it make sense for the  StartRequest to do that? I don't want to have too many methods on the API if possible.

model/interactive/src/main/proto/beam_interactive_api.proto
```diff
@@ -0,0 +1,160 @@
...
+  // starts the stream if it is not RUNNING.
+  rpc Advance (AdvanceRequest) returns (AdvanceResponse) {}
+
+  // Stops and resets the stream to the beginning.
+  rpc Stop (StopRequest) returns (StopResponse) {}
+
+  // Pauses the stream of events to the EventsRequest. If there is already an
+  // outstanding EventsRequest streaming events, then the stream will pause
+  // after the EventsResponse is completed.
+  // To un-pause, send either a Start or Advance request.
```
davidyan74: If the playback is paused, then an Advance request is issued, does the playback resume at the previous speed after the Advance, or should it remain paused? I think the latter makes more sense.

model/interactive/src/main/proto/beam_interactive_api.proto
```diff
@@ -0,0 +1,160 @@
...
+import "google/protobuf/timestamp.proto";
+import "google/protobuf/duration.proto";
+
+
+service InteractiveService {
+  // A TestStream will request for events using this RPC.
+  rpc Events(EventsRequest) returns (stream EventsResponse) {}
+
+  // Starts the stream of events to the EventsRequest.
+  rpc Start (StartRequest) returns (StartResponse) {}
```
davidyan74: [couldn't reply to an outdated comment]
I think it's okay for the Start method to also change the playback speed for a running replay. We just need to document that.

model/interactive/src/main/proto/beam_interactive_api.proto
```diff
@@ -0,0 +1,160 @@
...
+}
+
+message StartRequest {
+  // (Optional) How quickly the stream will be played back, e.g. if
+  // playback_speed == 2, then the stream will replay events twice as fast as
+  // they were recorded. If unspecified, this will default to 1.
+  double playback_speed = 1;
+
+  // (Optional) if present, will start the stream at the specified timestamp.
+  google.protobuf.Timestamp start_at = 2;
```
davidyan74: I'm not sure whether we should allow this feature since that means re-execution with start_at specified means different input data and it kinda defeats the purpose of replaying. Also, if we use Start method for also unpausing and changing playback speed, start_at either should be ignored or should return an error.

model/interactive/src/main/proto/beam_interactive_api.proto
```diff
@@ -0,0 +1,160 @@
...
+
+  // Responds with debugging and other cache-specific metadata.
+  rpc Status (StatusRequest) returns (StatusResponse) {}
+}
+
+message StartRequest {
+  // (Optional) How quickly the stream will be played back, e.g. if
+  // playback_speed == 2, then the stream will replay events twice as fast as
+  // they were recorded. If unspecified, this will default to 1.
+  double playback_speed = 1;
```
davidyan74: Should we just add [default = 1]?

rohdesamuel: Proto3 doesn't allow for explicit default values :(

model/interactive/src/main/proto/beam_interactive_api.proto
rohdesamuel: Good point, removed it.

model/interactive/src/main/proto/beam_interactive_api.proto
rohdesamuel: Done

model/interactive/src/main/proto/beam_interactive_api.proto
rohdesamuel: Ack, commented on the new protocol.

model/interactive/src/main/proto/beam_interactive_api.proto
```diff
@@ -0,0 +1,159 @@
...
+import "google/protobuf/timestamp.proto";
+import "google/protobuf/duration.proto";
+
+
+service InteractiveService {
+  // A TestStream will request for events using this RPC.
+  rpc Events(EventsRequest) returns (stream EventsResponse) {}
+
+  // Starts the stream of events to the EventsRequest. It is also allowed for
+  // setting the playback_speed while RUNNING.
```
davidyan74: Please also add unpause.

model/interactive/src/main/proto/beam_interactive_api.proto
```diff
@@ -0,0 +1,159 @@
...
+  // possible until the stream timestamp has advanced by the specified amount.
+  google.protobuf.Duration advance_by = 1;
+}
+message AdvanceResponse {}
+
+message StopRequest { }
+message StopResponse { }
+
+message PauseRequest { }
+message PauseResponse { }
```
davidyan74: Does it make sense to include the stream timestamp in PauseResponse?

model/interactive/src/main/proto/beam_interactive_api.proto
davidyan74: Ack.

model/interactive/src/main/proto/beam_interactive_api.proto
davidyan74: Thanks. I think we need to make the doc here better since Advance doesn't really "un-pause" since after Advance, the replay is still paused, but at a different timestamp.

model/interactive/src/main/proto/beam_interactive_api.proto
```diff
@@ -0,0 +1,159 @@
...
+    // with either a StartRequest or a StepRequest. Goes to STOPPED with a
+    // StopRequest.
+    PAUSED = 2;
+  }
+  State state = 5;
+}
+
+message StepRequest {
+  // (Optional) The count of events to stream. If unspecified, the default is
+  // 1, i.e. only read a single element from each source and send it through
```
davidyan74: Should we read a single element from any source, instead of one element from each source? It can be impossible to read one single element from each source since you might need to read many elements from one source before an element is available from another source.

rohdesamuel: Ack, changed to any source.

sdks/python/apache_beam/testing/interactive_stream.py
```diff
@@ -0,0 +1,136 @@
...
+
+from apache_beam.portability.api import beam_interactive_api_pb2
+from apache_beam.portability.api import beam_interactive_api_pb2_grpc
+from apache_beam.portability.api.beam_interactive_api_pb2_grpc import InteractiveServiceServicer
+
+STRING_TO_API_STATE = {
+    'STOPPED': beam_interactive_api_pb2.StatusResponse.STOPPED,
+    'PAUSED': beam_interactive_api_pb2.StatusResponse.PAUSED,
+    'RUNNING': beam_interactive_api_pb2.StatusResponse.RUNNING,
+}
```
davidyan74: What is the initial state?

model/interactive/src/main/proto/beam_interactive_api.proto
rohdesamuel: Ack, I changed the Start comment to reflect that it can also unpause.

sdks/python/apache_beam/testing/interactive_stream.py
rohdesamuel: Thanks, added a comment to the proto saying the initial state is STOPPED.

model/interactive/src/main/proto/beam_interactive_api.proto
```diff
@@ -0,0 +1,178 @@
...
+  // The InteractiveService is not replaying. Goes to RUNNING with a
+  // StartRequest.
+  STOPPED = 0;
+
+  // The InteractiveService is replaying events. Goes to PAUSED with a
+  // PauseRequest. Goes to STOPPED with a StopRequest.
+  RUNNING = 1;
+
+  // The InteractiveService is paused from replaying events. Goes to RUNNING
+  // with either a StartRequest or a StepRequest. Goes to STOPPED with a
```
davidyan74: Does it go to STEPPING (instead of RUNNING) with a StepRequest?
Also, does it go to ADVANCING with an AdvanceRequest?

model/interactive/src/main/proto/beam_interactive_api.proto
```diff
@@ -0,0 +1,178 @@
...
+
+
+service InteractiveService {
+  // A TestStream will request for events using this RPC.
+  rpc Events(EventsRequest) returns (stream EventsResponse) {}
+
+  // Starts the stream of events to the EventsRequest. It is also allowed for
+  // setting the playback_speed while RUNNING. If the stream is paused, this
+  // will unpause the stream at the specified playback speed.
+  rpc Start (StartRequest) returns (StartResponse) {}
```
davidyan74: nit: remove space between method name and open parenthesis for all the rpc declarations?

model/interactive/src/main/proto/beam_interactive_api.proto
```diff
@@ -0,0 +1,178 @@
...
+  State state = 5;
+}
+
+message StepRequest {
+  // (Optional) The count of events to stream. If unspecified, the default is
+  // 1, i.e. only read a single element from any source and send it through
+  // the pipeline.
+  int64 count = 1;
+}
+message StepResponse { }
```
davidyan74: Should this also return stream_time and watermark?

model/interactive/src/main/proto/beam_interactive_api.proto
rohdesamuel: Gotcha, I rewrote the state comments to be very explicit about the state machine and what happens at each state.

model/interactive/src/main/proto/beam_interactive_api.proto
rohdesamuel: done.

model/interactive/src/main/proto/beam_interactive_api.proto
rohdesamuel: sgtm

rohdesamuel:
I discussed with Robert the necessity of the GRPC service. In it, we agreed that the easiest way to support an interactive streaming job is through an HTTP service. Unfortunately, using HTTP transcoding with GRPC requires having a copy certain protos in the repo to be used during proto compilation. Thus, I will make a separate HTTP server without using the GRPC service definition and remove all InteractiveService rpcs _except_ for the Events request.

sdks/python/apache_beam/runners/interactive/caching/streaming_cache.py
```diff
@@ -0,0 +1,124 @@
...
+
+      events = []
+      if not records:
+        self.advance_watermark(timestamp.MAX_TIMESTAMP, events)
+
+      records = sorted(records.items(), key=lambda x: x[1].processing_time)
+      for tag, r in records:
+        # We always send the processing time event first so that the TestStream
+        # can sleep so as to emulate the original stream.
+        self.advance_processing_time(
```
robertwb: Return by mutable parameters is a C++ism that makes this code hard to follow. Please return the event to add, if any, instead.

sdks/python/apache_beam/runners/interactive/caching/streaming_cache.py
```diff
@@ -0,0 +1,124 @@
...
+        self._headers[header.tag if header.tag else None] = header
+        self._readers[header.tag if header.tag else None] = r
+
+      self._watermarks = {tag: timestamp.MIN_TIMESTAMP for tag in self._headers}
+
+    def read(self):
+      """Reads records from PCollection readers.
+      """
+      records = {}
+      for tag, r in self._readers.items():
```
robertwb: This logic is incorrect. One may have to return multiple events from one stream before any events of another. This should be tested.

sdks/python/apache_beam/runners/interactive/caching/streaming_cache.py
```diff
@@ -0,0 +1,124 @@
...
+        self._watermarks[tag] = watermark
+        payload = TestStreamPayload.Event(
+            watermark_event=TestStreamPayload.Event.AdvanceWatermark(
+                new_watermark=self._watermarks[tag].micros, tag=tag))
+        events.append(payload)
+
+    def stream_time(self):
+      return self._timestamp
+
+    def watermark(self):
```
robertwb: Why is this needed?

sdks/python/apache_beam/runners/interactive/caching/streaming_cache_test.py
```diff
@@ -0,0 +1,167 @@
...
+  """Converts seconds since epoch to a google.protobuf.Timestamp.
+  """
+  seconds = int(timestamp_secs)
+  nanos = int((timestamp_secs - seconds) * 10**9)
+  return timestamp_pb2.Timestamp(seconds=seconds, nanos=nanos)
+
+
+class InMemoryReader(object):
+  def __init__(self, tag=None):
+    self._records = [InteractiveStreamHeader(tag=tag).SerializeToString()]
```
robertwb: It'd be cleaner if the stream returned proto objects, rather than the stream returning strings which need to be parsed in a certain way.

sdks/python/apache_beam/testing/interactive_stream.py
```diff
@@ -0,0 +1,72 @@
...
+      self.endpoint = endpoint
+      self._server.add_insecure_port(self.endpoint)
+    else:
+      port = self._server.add_insecure_port('[::]:0')
+      self.endpoint = '[::]:{}'.format(port)
+
+    beam_interactive_api_pb2_grpc.add_InteractiveServiceServicer_to_server(
+        self, self._server)
+    self._streaming_cache = streaming_cache
+    self._playback_speed = 1 / 1000000.0
```
robertwb: I thought we were getting rid of this. If not, at least set it to rather than an arbitrary small value.

sdks/python/apache_beam/testing/interactive_stream.py
```diff
@@ -0,0 +1,72 @@
...
+
+  def start(self):
+    self._server.start()
+    self._reader = self._streaming_cache.reader()
```
robertwb: Isn't it a slight race to start the server before initializing the reader? Better would be to simply initialize it in Events() rather than store it as a local (which would break concurrently serving requests).

sdks/python/apache_beam/testing/interactive_stream.py
```diff
@@ -0,0 +1,72 @@
...
+
+  def stop(self):
+    self._server.stop(0)
+    self._server.wait_for_termination()
+
+  def Events(self, request, context):
+    events = self._reader.read()
+    if events:
+      for e in events:
+        # Here we assume that the first event is the processing_time_event so
```
robertwb: If we want to assume something, we should assert it to be more future proof.

sdks/python/apache_beam/testing/interactive_stream.py
```diff
@@ -0,0 +1,72 @@
...
+        # that we can sleep and then emit the element. Thereby, trying to
+        # emulate the original stream.
+        if e.HasField('processing_time_event'):
+          sleep_duration = (
+              e.processing_time_event.advance_duration * self._playback_speed
+              ) * 10**-6
```
robertwb: Whence this multiplier?

sdks/python/apache_beam/testing/interactive_stream.py
```diff
@@ -0,0 +1,72 @@
...
+          if sleep_duration > 0.001:
```
robertwb: `if sleep_duration`

sdks/python/apache_beam/testing/interactive_stream.py
```diff
@@ -0,0 +1,72 @@
...
+
+  def Events(self, request, context):
+    events = self._reader.read()
+    if events:
+      for e in events:
+        # Here we assume that the first event is the processing_time_event so
+        # that we can sleep and then emit the element. Thereby, trying to
+        # emulate the original stream.
+        if e.HasField('processing_time_event'):
+          sleep_duration = (
```
robertwb: Technically, the right thing to do is record the time since last event, and subtract off any elapsed time from the time we're supposed to sleep. If we're keeping this code here, at least add a TODO.

model/interactive/OWNERS
```diff
@@ -0,0 +1,7 @@
+# See the OWNERS docs at https://s.apache.org/beam-owners
+
+reviewers:
```
aaltay: You might want to limit this list to people who are likely to maintain this path.

model/interactive/src/main/proto/beam_interactive_api.proto
```diff
@@ -0,0 +1,69 @@
...
+// about the stream and how to properly process it.
+message InteractiveStreamHeader {
+  // The PCollection tag this stream is associated with.
+  string tag = 1;
+}
+
+// A record is a recorded element that sound source produced. Its function is
+// to give enough information to the InteractiveService to create a faithful
+// recreation of the original source of data.
+message InteractiveStreamRecord {
```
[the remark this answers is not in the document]
rohdesamuel: I don't fully understand what you are saying, but I'm guessing you want a single proto with a repeated field for events and watermark/clock advances? In that case, it wouldn't work. This is going to be the file format for storing the events of an unbounded source. The expectation is that the file may not fit into memory, so it will be streamed in one-by-one. This necessitates having individual records instead of a single proto for the file.

model/interactive/src/main/proto/beam_interactive_api.proto
```diff
@@ -0,0 +1,69 @@
...
+option java_package = "org.apache.beam.model.interactive.v1";
+option java_outer_classname = "BeamInteractiveApi";
+
+import "beam_runner_api.proto";
+import "google/protobuf/timestamp.proto";
+
+
+service InteractiveService {
+  // A TestStream will request for events using this RPC.
+  rpc Events(EventsRequest) returns (stream EventsResponse) {}
```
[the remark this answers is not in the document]
rohdesamuel: In order to allow for quiescence, I changed this from a stream to a single response per request.

model/interactive/src/main/proto/beam_interactive_api.proto
```diff
@@ -0,0 +1,69 @@
...
+
+option go_package = "interactive_v1";
+option java_package = "org.apache.beam.model.interactive.v1";
+option java_outer_classname = "BeamInteractiveApi";
+
+import "beam_runner_api.proto";
+import "google/protobuf/timestamp.proto";
+
+
+service InteractiveService {
```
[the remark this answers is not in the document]
rohdesamuel: I would rather not, this is pretty specific to the InteractiveRunner.

sdks/python/apache_beam/runners/interactive/caching/streaming_cache.py
rohdesamuel: ack, changed to return the event.

sdks/python/apache_beam/runners/interactive/caching/streaming_cache.py
rohdesamuel: Good catch, I changed it to be able to read from multiple underlying caches. PTAL.

sdks/python/apache_beam/runners/interactive/caching/streaming_cache.py
rohdesamuel: For introspection by the user.

sdks/python/apache_beam/runners/interactive/caching/streaming_cache_test.py
rohdesamuel: The code for the underlying PCollection Cache isn't merged yet, and I haven't decided if I want to create a new subclass or use it directly. This code assumes that the PCollection cache will just writes strings to a file. I can change it later once the PCollection Cache code is merged.

sdks/python/apache_beam/testing/interactive_stream.py
rohdesamuel: Removed

sdks/python/apache_beam/testing/interactive_stream.py
rohdesamuel: Yeah, that's a good point. I changed it from an implicit session (which only supported a single job) to explicit session which can support an arbitrary number of jobs.

sdks/python/apache_beam/testing/interactive_stream.py
rohdesamuel: To convert from us to secs

sdks/python/apache_beam/testing/interactive_stream.py
rohdesamuel: Simplified the code quite a bit and moved the logic into the StreamingCache. Removed this assumption.

sdks/python/apache_beam/testing/interactive_stream.py
rohdesamuel: Removed

sdks/python/apache_beam/testing/interactive_stream.py
rohdesamuel: Ack, PTAL in the StreamingCache. I moved the logic to add in AdvanceProcessingTime events into there.

model/interactive/OWNERS
rohdesamuel: Done

sdks/python/apache_beam/runners/interactive/caching/streaming_cache.py
```diff
@@ -0,0 +1,124 @@
...
+    def __init__(self, readers):
+      self._timestamp = Timestamp.of(0)
+      self._readers = {}
+      self._headers = {}
+      readers = [r.read() for r in readers]
+
+      # The header allows for metadata about an entire stream, so that the data
+      # isn't copied per record.
+      for r in readers:
+        header = InteractiveStreamHeader()
```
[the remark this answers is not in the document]
rohdesamuel: In a different comment, I wrote that I don't know if I want to subclass the PCollection Cache yet. So I assume here that it will be a simple file based cache of strings that I will read row-by-row without any assumption of underlying format. I may change that assumption in the future when the PCollection Cache code is merged.

rohdesamuel:
Run Python PreCommit

rohdesamuel:
Run Java PreCommit

rohdesamuel:
Run Python PreCommit

rohdesamuel:

sdks/python/apache_beam/runners/interactive/caching/streaming_cache.py
robertwb: I didn't think the user had a handle to this, or was interested in the min of all watermarks.

sdks/python/apache_beam/runners/interactive/caching/streaming_cache.py
```diff
@@ -0,0 +1,179 @@
...
+      e = TestStreamPayload.Event(
+          watermark_event=TestStreamPayload.Event.AdvanceWatermark(
+              new_watermark=self._watermarks[tag].micros, tag=tag))
+      return e
+
+    def stream_time(self):
+      return self._monotonic_clock
+
+    def watermarks(self):
+      return self._watermarks
```
robertwb: Similarly, this seems unused.

sdks/python/apache_beam/runners/interactive/caching/streaming_cache_test.py
robertwb: The point stands whether or not we'll merge with PCollectionCache later.

sdks/python/apache_beam/runners/interactive/interactive_stream.py
```diff
@@ -0,0 +1,84 @@
...
+    Token behavior: the first request should have a token of "None". Each
+    subsequent request should use the previously received token from the
+    response. The stream ends when the returned token is the empty string.
+    """
+    assert request.session_id in self._sessions, (\
+        'Session "{}" was not found. Did you forget to call Connect ' +
+        'first?').format(request.session_id)
+
+    reader = self._sessions[request.session_id]
+    token = (int(request.token) if request.token else 0) + 1
```
robertwb: There is no checking done here whether the token is actually sequential, nor any support for multiple callers of this API.

model/interactive/src/main/proto/beam_interactive_api.proto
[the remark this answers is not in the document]
robertwb: Whether this is streaming or not has nothing to do with quiescence. By spreading the call across several RPC calls the service has become stateful, more complicated, and idempotent. Much better for the RPC to be "give me the events of this test stream" than have a (one-time-use) multi-RPC protocol to fetch them.

sdks/python/apache_beam/runners/interactive/caching/streaming_cache.py
```diff
@@ -0,0 +1,179 @@
...
+        events = events + self._read_next()
+        events = sorted(events,
+                        key=lambda x: Timestamp.from_proto(
+                            x[1].processing_time),
+                        reverse=True)
+        if not events:
+          break
+
+        # Retrieves the minimum timestamp from the read events.
+        min_timestamp = (
```
robertwb: Why is this a callable?

sdks/python/apache_beam/runners/interactive/caching/streaming_cache.py
robertwb: I'm still having  hard time convincing myself that the logic between _read_next() and read() is correct. 

What you're really trying to do is perform a merge sort of the reader on timestamp. For readability and maintainability, it would probably be best expressed that way. Whether these are Events that already have the tag embedded (I'm assuming the backing file format will be compressed, so little to no overhead) or the input is a dict of {tag: iterator} and the tags applied when merging depends on which is easier (e.g. when writing).

model/interactive/src/main/proto/beam_interactive_api.proto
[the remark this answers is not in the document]
rohdesamuel: Sure, I'll move it to the beam_runner_api.proto.

model/interactive/src/main/proto/beam_interactive_api.proto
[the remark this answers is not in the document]
rohdesamuel: It's not really isomorphic. The InteractiveStreamRecord has the added processing_time field. This is the absolute wall time seen at when that element/watermark event was recorded. The TestStreamPaylaod.AdvanceProcessingTime event only records deltas between two timestamps. It's also good practice to separate the proto formats for storage and in-memory use.

sdks/python/apache_beam/runners/interactive/caching/streaming_cache.py
rohdesamuel: Ack, I broke it up to smaller methods, ptal.

sdks/python/apache_beam/runners/interactive/caching/streaming_cache.py
rohdesamuel: It's callable because it retrieves the the minimum timestamp of the events. Made it clearer by adding events as an argument and into its own method.

sdks/python/apache_beam/runners/interactive/caching/streaming_cache.py
rohdesamuel: Removed

sdks/python/apache_beam/runners/interactive/caching/streaming_cache.py
rohdesamuel: Removed

sdks/python/apache_beam/runners/interactive/caching/streaming_cache_test.py
rohdesamuel: I'm hoping that we will have the PCollectionCache! Regardless, In the meantime I changed it to return proto objects instead of the string.

sdks/python/apache_beam/runners/interactive/interactive_stream.py
rohdesamuel: Added a check and a test for multiple sessions.

rohdesamuel:
> Whether this is streaming or not has nothing to do with quiescence. By spreading the call across several RPC calls the service has become stateful, more complicated, and idempotent. Much better for the RPC to be "give me the events of this test stream" than have a (one-time-use) multi-RPC protocol to fetch them.

The current implementation in the DirectRunner is that the TestStreamEvaluator reads a single event from the TestStream at a time. It is stateful in nature as it has to keep its current index into the TestStream events between calls. Remember that the DirectRunner creates a TransformEvaluator per application of a bundle. In order to support that semantic of requesting and sending a single element, the service will also reflect that. 

> This PR does three separate things that all fall under the heading of "add initial modules." To facilitate getting this in, I suggest braking it up in to multiple PRs.
> 
> (1) Augment TestStream to support multiple output, with tests.
> (2) Add a service backing for TestStream.
> (3) Add the StreamingCache stream merging classes.

With all due respect, this should have been brought up at the start of the review. With that in mind, I am inclined to finish this PR as-is, without breaking it up. I will keep your recommendation in mind in future PRs.

model/interactive/src/main/proto/beam_interactive_api.proto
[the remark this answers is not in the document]
robertwb: Given a start time, they are isomorphic. I'm not sold on the value having a redundant format for use on disk vs. in memory, but if we go this route the file-based ones are really test specific even if the service isn't.

sdks/python/apache_beam/runners/interactive/caching/streaming_cache.py
```diff
@@ -0,0 +1,166 @@
...
+    This class is an Abstraction layer over multiple PCollection readers to be
+    used for supplying a TestStream service with events.
+
+    This class is also responsible for holding the state of the clock, injecting
+    clock advancement events, and watermark advancement events.
+    """
+    def __init__(self, readers):
+      # This timestamp is used as the monotonic clock to order events in the
+      # replay.
+      self._monotonic_clock = timestamp.Timestamp.of(0)
```
robertwb: MIN_TIMESTAMP?

sdks/python/apache_beam/runners/interactive/caching/streaming_cache.py
```diff
@@ -0,0 +1,166 @@
...
+      self._readers = {r.header().tag : r.read() for r in readers}
+
+      # The watermarks per tag. Useful for introspection in the stream.
+      self._watermarks = {tag: timestamp.MIN_TIMESTAMP for tag in self._headers}
+
+      # The most recently read timestamp per tag.
+      self._stream_times = {tag: timestamp.MIN_TIMESTAMP
+                            for tag in self._headers}
+
+    def _test_stream_events_before_target(self, target_timestamp):
```
robertwb: This doesn't seem to read all events before the target timestamp, but at most one per stream, right?

sdks/python/apache_beam/runners/interactive/caching/streaming_cache.py
rohdesamuel: We do timestamp differences between timestamps in the _advance_processing_time method, which can overflow if this is MIN_TIMESTAMP. This isn't a Watermark, so it's okay to set this to 0 as we can't have an element with a wall time at 0.

sdks/python/apache_beam/runners/interactive/caching/streaming_cache.py
rohdesamuel: Yep, I added a more in depth comment to the function signature.

rohdesamuel:
> > > Whether this is streaming or not has nothing to do with quiescence. By spreading the call across several RPC calls the service has become stateful, more complicated, and idempotent. Much better for the RPC to be "give me the events of this test stream" than have a (one-time-use) multi-RPC protocol to fetch them.
> > 
> > 
> > The current implementation in the DirectRunner is that the TestStreamEvaluator reads a single event from the TestStream at a time. It is stateful in nature as it has to keep its current index into the TestStream events between calls. Remember that the DirectRunner creates a TransformEvaluator per application of a bundle. In order to support that semantic of requesting and sending a single element, the service will also reflect that.
> 
> We should not let an implementation, especially one that is hopefully going away soon, dictate what our APIs should be. Instead, we should come up with clean APIs that represent the abstraction well, and then figure out how they can be implemented. Only if it's truly impossible (or undesirable) to support a clean API should we mangle it to reflect the implementation.
Alright, I rewrote the API to return a stream of TestStream events back. This also allowed to get rid of the ConnectRequest.

> > > This PR does three separate things that all fall under the heading of "add initial modules." To facilitate getting this in, I suggest braking it up in to multiple PRs.
> > > (1) Augment TestStream to support multiple output, with tests.
> > > (2) Add a service backing for TestStream.
> > > (3) Add the StreamingCache stream merging classes.
> > 
> > 
> > With all due respect, this should have been brought up at the start of the review. With that in mind, I am inclined to finish this PR as-is, without breaking it up. I will keep your recommendation in mind in future PRs.
> 
> I don't see how splitting it up now is harder than it would have been at the start (though, yes, had we done it sooner perhaps some pieces could have been in already). But this is a huge change with many orthogonal features that should be separated out, if just for ease of rollback in case something is broken, and for clean history when trying to understand the code.

model/interactive/src/main/proto/beam_interactive_api.proto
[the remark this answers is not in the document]
rohdesamuel: Ok, I rewrote the API to return a stream of TestStreamPayload events. This also allowed the removal of the Connect request.

rohdesamuel:
Run Portable_Python PreCommit

rohdesamuel:
Run Java PreCommit

sdks/python/apache_beam/runners/interactive/caching/streaming_cache.py
robertwb: Thanks. The key to making things readable is not just having smaller methods, but having each piece a self-contained unit of abstraction. For example, making `target_timestamp` a parameter of _test_stream_events_before_target rather than setting a member on self and inspecting it in that function is a nice step forward. On the other hand, _min_timestamp_of now has an (implicit) assumption that the events are passed in (reversed) timestamp order which makes it harder to understand in isolation. 

I think, as far as I am able to follow it, the code as written is now correct.

sdks/python/apache_beam/runners/interactive/caching/streaming_cache.py
```diff
@@ -0,0 +1,171 @@
...
+          continue
+        try:
+          record = next(r)
+          records.append((tag, record))
+          self._stream_times[tag] = Timestamp.from_proto(record.processing_time)
+        except StopIteration:
+          pass
+      return records
+
+    def _merge_sort(self, previous_events, new_events):
```
robertwb: By merge sort, I meant basically https://guava.dev/releases/20.0/api/docs/com/google/common/collect/Iterables.html#mergeSorted-java.lang.Iterable-java.util.Comparator-

In short, one maintains a heap of readers sorted by the timestamp of their (peeked at) next element. Then

```
while not reader_heap.empty():
  first_reader = reader_heap.pop()
  try:
    yield first_reader.read()
  except StopIteration:
    pass
  else:
    reader_heap.push(first_reader)
```

This should be equivalent, so you can keep things as is if you want.

sdks/python/apache_beam/testing/test_stream_service.py
```diff
@@ -0,0 +1,55 @@
...
+from concurrent.futures import ThreadPoolExecutor
+
+import grpc
+
+from apache_beam.portability.api import beam_runner_api_pb2_grpc
+from apache_beam.portability.api.beam_runner_api_pb2_grpc import TestStreamServiceServicer
+
+
+class TestStreamServiceController(TestStreamServiceServicer):
+  def __init__(self, streaming_cache, endpoint=None):
```
robertwb: TestStreamServiceController should not depend on StreamingCache. Instead, it should just take an `Iterable[TestStreamPayload.Event]` in its constructor. This would greatly simplify the testing below as well.

sdks/python/apache_beam/runners/interactive/caching/streaming_cache.py
rohdesamuel: Ack, changed comment.

sdks/python/apache_beam/runners/interactive/caching/streaming_cache.py
rohdesamuel: Ack, I think that would be a good refactor for a follow-up PR.

sdks/python/apache_beam/testing/test_stream_service.py
rohdesamuel: If it's an iterable, then the service can only loop through it once. I changed it to a bound method that returns an iterable. How does that sound? I also got rid of the StreamingCache dependency.

sdks/python/apache_beam/runners/interactive/caching/streaming_cache.py
rohdesamuel: Cool, ty

rohdesamuel:
I broke up the PRs into: 

- https://github.com/apache/beam/pull/10120 : TestStreamService
- https://github.com/apache/beam/pull/10119 : StreamingCache
- https://github.com/apache/beam/pull/10117 : Protos

PTAL at your earliest convenience.

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -489,33 +490,66 @@ message TestStreamPayload {
...
+      // (Required) The duration to advance by.
       int64 advance_duration = 1;
     }
 
+    // Adds elements to the stream to be emitted.
     message AddElements {
+      // (Required) The elements to add to the TestStream.
       repeated TimestampedElement elements = 1;
+
+      // (Optional) The output PCollection tag to add these elements to. If
```
lukecwik: Ditto on required.

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -489,33 +490,66 @@ message TestStreamPayload {
...
       AdvanceProcessingTime processing_time_event = 2;
       AddElements element_event = 3;
     }
 
+    // Advances the watermark to the specified timestamp.
     message AdvanceWatermark {
+      // (Required) The watermark to advance to.
       int64 new_watermark = 1;
+
+      // (Optional) The output watermark tag for a PCollection. If unspecified
```
lukecwik: I believe this should be required instead of having SDKs specially handle this logic.

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -489,33 +490,66 @@ message TestStreamPayload {
...
+    // (Required) The element encoded. Currently the TestStream only supports
+    // encoding primitives.
     bytes encoded_element = 1;
+
+    // (Required) The event timestamp of this element.
     int64 timestamp = 2;
   }
 }
+
+service TestStreamService {
```
lukecwik: Why is this needed?

The current ProcessBundleDescriptor and ProcessBundleRequest can already split the execution so that the Runner and the SDK perform their respective responsibilities:
```
Runner that is processing and emitting TestStream events -> Data Channel -> SDK processing events
```

model/interactive/src/main/proto/beam_interactive_api.proto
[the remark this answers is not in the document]
lukecwik: I would have suggested to make a new transform definition called ExternalTestStream (or some similar name) which is independent of this TestStream.

In general I believe it would be best if transforms had a narrow scope as the cost of choosing a new URN and to have a specific payload is meant to be inexpensive. The payload can be arbitrarily chosen and does not necessarily need to be in proto.

Now for a runner to support TestStream it now has to say what parts of the TestStream specification it supports which is a hassle to deal with. Also this leads to others following the same pattern since adding a new field to an existing transform may be backwards compatible from a decoding perspective but isn't backwards compatible if it changes the intent of the transform since a runner may not know of that field.

PR #9741, opened 2019-10-07
[BEAM-7926] Visualize PCollection

KevinGG:
1. Added pcoll_visualization module to visualize materialized
PCollection data.
2. A visualization mainly consists of 3 parts: facets-dive,
facets-overview and data-table.
3. The visualization is always non-blocking w/ 2 modes: one-shot and
dynamic plotting.
4. One-shot visualization renders the visualization as HTML statically.
5. Dynamical plotting polls materialized data, initializes a
visualization and then updates the visualization continuously.
6. It's the pipeline state, i.e., whether the PCollection
materialization is being updated that decides if the visualization is
dynamic plotting or one-shot. It's not the boundedness of PCollection
that decides the visualization mode. However, when engineering for
Interactive Beam, it's alright to always use the dynamic plotting
mode as long as the InteractiveRunner(s) manages jobs and tracks the
pipeline(job) results correctly. Failing to terminate a visualization
job might cause resource leak.
7. The materialized PCollection data goes through the following process
flow: cache -> element list -> normalization -> dataframe ->
HTMLs/scripts.
8. A visualization is always self-contained. Meaning it controls when
the visualization starts and ends by itself. Once visualized, the
visualization is anchored within the notebook cell's output area. Code
in other cells can affect the visualization in the cell but will not
change the existing visualization's position in the notebook. Multiple
visualize(pcoll) calls will render multiple independent distinct
visualization instances/jobs. Each visualization runs in its own thread
outside of the notebook main thread.
9. Unittest added focusing on the above visualization logic other than
the visualization UI (those should belong to Facets, pandas and jQuery,
we'll trust them work as intended). Integration tests can be added
later.
10. Each visualization job regularly queries interactive environment for
running pipeline(job) results implicitly to determine the default
visualization end condition. It's based on the limitation that within
current interactive environment, at a time, at most one job is in
non-terminated state for an end user defined pipeline instance (in
whatever state). The end user could also use the handle returned by
visualize() to stop() the visualization job manually.
11. Added INTERACTIVE_BEAM extras_require and tests_require in setup.
They are only needed for data visualization in notebook. `jsons` is for
easy serialization of any object and `timeloop` is for self-contained
periodic async tasks. They both use MIT licenses. `facets-overview` is
for overview part of facets. It uses apache 2.0 license. `ipython` is the
kernel used by jupyter, included for non-jupyter environment though it
always comes with jupyter. It uses BSD license. The lower version limit
set is the highest version listed by flink.
------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
+++ b/sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
@@ -103,13 +103,21 @@ def visualize(pcoll, dynamic_plotting_interval=None):
     # Otherwise, dynamic plotting ends when pipeline is not running anymore.
 
   If dynamic_plotting is not enabled (by default), None is returned.
+
+  The function is experimental. For internal use only; no
+  backwards-compatibility guarantees.
   """
   if not _pcoll_visualization_ready:
     return None
   pv = PCollectionVisualization(pcoll)
-  pv.display_facets()
+  if ie.current_env().is_in_notebook:
+    pv.display_facets()
+  else:
+    pv.display_plain_text()
+    # We don't want to do dynamic plotting if there is no notebook frontend.
+    return None
 
-  if dynamic_plotting_interval:
+  if ie.current_env().is_in_notebook and dynamic_plotting_interval:
     # Disables the verbose logging from timeloop.
     logging.getLogger('timeloop').disabled = True
     tl = Timeloop()
@@ -168,6 +176,19 @@ class PCollectionVisualization(object):
     self._overview_display_id = 'facets_overview_{}_{}'.format(self._cache_key,
                                                                id(self))
     self._df_display_id = 'df_{}_{}'.format(self._cache_key, id(self))
+
+  def display_plain_text(self):
+    """Displays a random sample of the normalized PCollection data.
+
+    This function is used when the ipython kernel is not connected to a
+    notebook frontend such as when running ipython in terminal or in unit tests.
+    """
+    # Double check if the dependency is ready in case someone mistakenly use
+    # the function.
+    if _pcoll_visualization_ready:
+      data = self._to_dataframe()
+      data_sample = data.sample(n=25 if len(data) > 25 else len(data))
+      display(data_sample)
 
   def display_facets(self, updating_pv=None):
     """Displays the visualization through IPython.
--- a/sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
+++ b/sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
@@ -32,38 +32,23 @@ from apache_beam import pvalue
 from apache_beam.runners.interactive import interactive_environment as ie
 from apache_beam.runners.interactive import pipeline_instrument as instr
 
-# jsons doesn't support < Python 3.5. Work around with json for legacy tests.
-# TODO(BEAM-8288): clean up once Py2 is deprecated from Beam.
 try:
-  import jsons
-  _pv_jsons_load = jsons.load
-  _pv_jsons_dump = jsons.dump
-except ImportError:
-  import json
-  _pv_jsons_load = json.load
-  _pv_jsons_dump = json.dump
+  import jsons  # pylint: disable=import-error
+  from IPython import get_ipython  # pylint: disable=import-error
+  from IPython.core.display import HTML  # pylint: disable=import-error
+  from IPython.core.display import Javascript  # pylint: disable=import-error
+  from IPython.core.display import display  # pylint: disable=import-error
+  from IPython.core.display import display_javascript  # pylint: disable=import-error
+  from IPython.core.display import update_display  # pylint: disable=import-error
+  from facets_overview.generic_feature_statistics_generator import GenericFeatureStatisticsGenerator  # pylint: disable=import-error
+  from timeloop import Timeloop  # pylint: disable=import-error
 
-try:
-  from facets_overview.generic_feature_statistics_generator import GenericFeatureStatisticsGenerator
-  _facets_gfsg_ready = True
+  if get_ipython():
+    _pcoll_visualization_ready = True
+  else:
+    _pcoll_visualization_ready = False
 except ImportError:
-  _facets_gfsg_ready = False
-
-try:
-  from IPython.core.display import HTML
-  from IPython.core.display import Javascript
-  from IPython.core.display import display
-  from IPython.core.display import display_javascript
-  from IPython.core.display import update_display
-  _ipython_ready = True
-except ImportError:
-  _ipython_ready = False
-
-try:
-  from timeloop import Timeloop
-  _tl_ready = True
-except ImportError:
-  _tl_ready = False
+  _pcoll_visualization_ready = False
 
 # 1-d types that need additional normalization to be compatible with DataFrame.
 _one_dimension_types = (int, float, str, bool, list, tuple)
@@ -119,12 +104,12 @@ def visualize(pcoll, dynamic_plotting_interval=None):
 
   If dynamic_plotting is not enabled (by default), None is returned.
   """
-  if not _ipython_ready:
+  if not _pcoll_visualization_ready:
     return None
   pv = PCollectionVisualization(pcoll)
   pv.display_facets()
 
-  if dynamic_plotting_interval and _tl_ready:
+  if dynamic_plotting_interval:
     # Disables the verbose logging from timeloop.
     logging.getLogger('timeloop').disabled = True
     tl = Timeloop()
@@ -160,6 +145,11 @@ class PCollectionVisualization(object):
   """
 
   def __init__(self, pcoll):
+    assert _pcoll_visualization_ready, (
+        'Dependencies for PCollection visualization are not available. Please '
+        'use `pip install apache-beam[interactive]` to install necessary '
+        'dependencies and make sure that you are executing code in an '
+        'interactive environment such as a Jupyter notebook.')
     assert isinstance(pcoll, pvalue.PCollection), (
         'pcoll should be apache_beam.pvalue.PCollection')
     self._pcoll = pcoll
@@ -219,13 +209,10 @@ class PCollectionVisualization(object):
       display(HTML(html))
 
   def _display_overview(self, data, update=None):
-    if _facets_gfsg_ready:
-      gfsg = GenericFeatureStatisticsGenerator()
-      proto = gfsg.ProtoFromDataFrames(
-          [{'name': 'data', 'table': data}])
-      protostr = base64.b64encode(proto.SerializeToString()).decode('utf-8')
-    else:  # GenericFeatureStatisticsGenerator is not supported.
-      protostr = ''
+    gfsg = GenericFeatureStatisticsGenerator()
+    proto = gfsg.ProtoFromDataFrames(
+        [{'name': 'data', 'table': data}])
+    protostr = base64.b64encode(proto.SerializeToString()).decode('utf-8')
     if update:
       script = _OVERVIEW_SCRIPT_TEMPLATE.format(
           display_id=update,
--- a/sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
+++ b/sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
@@ -259,7 +259,7 @@ class PCollectionVisualization(object):
         # Makes such data structured.
         normalized_list.append({normalized_column: el})
       else:
-        normalized_list.append(_pv_jsons_load(_pv_jsons_dump(el)))
+        normalized_list.append(jsons.load(jsons.dump(el)))
     # Creates a dataframe that str() 1-d iterable elements after
     # normalization so that facets_overview can treat such data as categorical.
     return json_normalize(normalized_list).applymap(
--- a/sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
+++ b/sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
@@ -117,7 +117,7 @@ def visualize(pcoll, dynamic_plotting_interval=None):
     # We don't want to do dynamic plotting if there is no notebook frontend.
     return None
 
-  if ie.current_env().is_in_notebook and dynamic_plotting_interval:
+  if dynamic_plotting_interval:
     # Disables the verbose logging from timeloop.
     logging.getLogger('timeloop').disabled = True
     tl = Timeloop()
@@ -183,7 +183,7 @@ class PCollectionVisualization(object):
     This function is used when the ipython kernel is not connected to a
     notebook frontend such as when running ipython in terminal or in unit tests.
     """
-    # Double check if the dependency is ready in case someone mistakenly use
+    # Double check if the dependency is ready in case someone mistakenly uses
     # the function.
     if _pcoll_visualization_ready:
       data = self._to_dataframe()
--- a/sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
+++ b/sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
@@ -1,0 +1,282 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""Module visualizes PCollection data.
+
+For internal use only; no backwards-compatibility guarantees.
+Only works with Python 3.5+.
+"""
+from __future__ import absolute_import
+
+import base64
+import logging
+from datetime import timedelta
+
+from pandas.io.json import json_normalize
+
+from apache_beam import pvalue
+from apache_beam.runners.interactive import interactive_environment as ie
+from apache_beam.runners.interactive import pipeline_instrument as instr
+
+# jsons doesn't support < Python 3.5. Work around with json for legacy tests.
+# TODO(BEAM-8288): clean up once Py2 is deprecated from Beam.
+try:
+  import jsons
+  _pv_jsons_load = jsons.load
+  _pv_jsons_dump = jsons.dump
+except ImportError:
+  import json
+  _pv_jsons_load = json.load
+  _pv_jsons_dump = json.dump
+
+try:
+  from facets_overview.generic_feature_statistics_generator import GenericFeatureStatisticsGenerator
+  _facets_gfsg_ready = True
+except ImportError:
+  _facets_gfsg_ready = False
+
+try:
+  from IPython.core.display import HTML
+  from IPython.core.display import Javascript
+  from IPython.core.display import display
+  from IPython.core.display import display_javascript
+  from IPython.core.display import update_display
+  _ipython_ready = True
+except ImportError:
+  _ipython_ready = False
+
+try:
+  from timeloop import Timeloop
+  _tl_ready = True
+except ImportError:
+  _tl_ready = False
+
+# 1-d types that need additional normalization to be compatible with DataFrame.
+_one_dimension_types = (int, float, str, bool, list, tuple)
+
+_DIVE_SCRIPT_TEMPLATE = """
+            document.querySelector("#{display_id}").data = {jsonstr};"""
+_DIVE_HTML_TEMPLATE = """
+            <script src="https://cdnjs.cloudflare.com/ajax/libs/webcomponentsjs/1.3.3/webcomponents-lite.js"></script>
+            <link rel="import" href="https://raw.githubusercontent.com/PAIR-code/facets/1.0.0/facets-dist/facets-jupyter.html">
+            <facets-dive sprite-image-width="{sprite_size}" sprite-image-height="{sprite_size}" id="{display_id}" height="600"></facets-dive>
+            <script>
+              document.querySelector("#{display_id}").data = {jsonstr};
+            </script>"""
+_OVERVIEW_SCRIPT_TEMPLATE = """
+              document.querySelector("#{display_id}").protoInput = "{protostr}";
+              """
+_OVERVIEW_HTML_TEMPLATE = """
+            <script src="https://cdnjs.cloudflare.com/ajax/libs/webcomponentsjs/1.3.3/webcomponents-lite.js"></script>
+            <link rel="import" href="https://raw.githubusercontent.com/PAIR-code/facets/1.0.0/facets-dist/facets-jupyter.html">
+            <facets-overview id="{display_id}"></facets-overview>
+            <script>
+              document.querySelector("#{display_id}").protoInput = "{protostr}";
+            </script>"""
+_DATAFRAME_PAGINATION_TEMPLATE = """
+            <script src="https://ajax.googleapis.com/ajax/libs/jquery/2.2.2/jquery.min.js"></script> 
+            <script src="https://cdn.datatables.net/1.10.16/js/jquery.dataTables.js"></script> 
+            <link rel="stylesheet" href="https://cdn.datatables.net/1.10.16/css/jquery.dataTables.css">
+            {dataframe_html}
+            <script>
+              $("#{table_id}").DataTable();
+            </script>"""
+
+
+def visualize(pcoll, dynamic_plotting_interval=None):
+  """Visualizes the data of a given PCollection. Optionally enables dynamic
+  plotting with interval in seconds if the PCollection is being produced by a
+  running pipeline or the pipeline is streaming indefinitely. The function
+  always returns immediately and is asynchronous when dynamic plotting is on.
+
+  If dynamic plotting enabled, the visualization is updated continuously until
+  the pipeline producing the PCollection is in an end state. The visualization
+  would be anchored to the notebook cell output area. The function
+  asynchronously returns a handle to the visualization job immediately. The user
+  could manually do::
+
+    # In one notebook cell, enable dynamic plotting every 1 second:
+    handle = visualize(pcoll, dynamic_plotting_interval=1)
+    # Visualization anchored to the cell's output area.
+    # In a different cell:
+    handle.stop()
+    # Will stop the dynamic plotting of the above visualization manually.
+    # Otherwise, dynamic plotting ends when pipeline is not running anymore.
+
+  If dynamic_plotting is not enabled (by default), None is returned.
+  """
+  if not _ipython_ready:
+    return None
+  pv = PCollectionVisualization(pcoll)
+  pv.display_facets()
+
+  if dynamic_plotting_interval and _tl_ready:
+    # Disables the verbose logging from timeloop.
+    logging.getLogger('timeloop').disabled = True
+    tl = Timeloop()
+
+    def dynamic_plotting(pcoll, pv, tl):
+      @tl.job(interval=timedelta(seconds=dynamic_plotting_interval))
+      def continuous_update_display():  # pylint: disable=unused-variable
+        # Always creates a new PCollVisualization instance when the
+        # PCollection materialization is being updated and dynamic
+        # plotting is in-process.
+        updated_pv = PCollectionVisualization(pcoll)
+        updated_pv.display_facets(updating_pv=pv)
+        if ie.current_env().is_terminated(pcoll.pipeline):
+          try:
+            tl.stop()
+          except RuntimeError:
+            # The job can only be stopped once. Ignore excessive stops.
+            pass
+
+      tl.start()
+      return tl
+
+    return dynamic_plotting(pcoll, pv, tl)
+  return None
+
+
+class PCollectionVisualization(object):
+  """A visualization of a PCollection.
+
+  The class relies on creating a PipelineInstrument w/o actual instrument to
+  access current interactive environment for materialized PCollection data at
+  the moment of self instantiation through cache.
+  """
+
+  def __init__(self, pcoll):
+    assert isinstance(pcoll, pvalue.PCollection), (
+        'pcoll should be apache_beam.pvalue.PCollection')
+    self._pcoll = pcoll
+    # This allows us to access cache key and other meta data about the pipeline
+    # whether it's the pipeline defined in user code or a copy of that pipeline.
+    # Thus, this module doesn't need any other user input but the PCollection
+    # variable to be visualized. It then automatically figures out the pipeline
+    # definition, materialized data and the pipeline result for the execution
+    # even if the user never assigned or waited the result explicitly.
+    # With only the constructor of PipelineInstrument, any interactivity related
+    # pre-process or instrument is not triggered for performance concerns.
+    self._pin = instr.PipelineInstrument(pcoll.pipeline)
+    self._cache_key = self._pin.cache_key(self._pcoll)
+    self._dive_display_id = 'facets_dive_{}_{}'.format(self._cache_key,
+                                                       id(self))
+    self._overview_display_id = 'facets_overview_{}_{}'.format(self._cache_key,
+                                                               id(self))
+    self._df_display_id = 'df_{}_{}'.format(self._cache_key, id(self))
+
+  def display_facets(self, updating_pv=None):
+    """Displays the visualization through IPython.
+
+    Args:
+      updating_pv: A PCollectionVisualization object. When provided, the
+        display_id of each visualization part will inherit from the initial
+        display of updating_pv and only update that visualization web element
+        instead of creating new ones.
+
+    The visualization has 3 parts: facets-dive, facets-overview and paginated
+    data table. Each part is assigned an auto-generated unique display id
+    (the uniqueness is guaranteed throughout the lifespan of the PCollection
+    variable).
+    """
+    # Ensures that dive, overview and table render the same data because the
+    # materialized PCollection data might being updated continuously.
+    data = self._to_dataframe()
+    if updating_pv:
+      self._display_dive(data, updating_pv._dive_display_id)
+      self._display_overview(data, updating_pv._overview_display_id)
+      self._display_dataframe(data, updating_pv._df_display_id)
+    else:
+      self._display_dive(data)
+      self._display_overview(data)
+      self._display_dataframe(data)
+
+  def _display_dive(self, data, update=None):
+    sprite_size = 32 if len(data.index) > 50000 else 64
+    jsonstr = data.to_json(orient='records')
+    if update:
+      script = _DIVE_SCRIPT_TEMPLATE.format(display_id=update,
+                                            jsonstr=jsonstr)
+      display_javascript(Javascript(script))
+    else:
+      html = _DIVE_HTML_TEMPLATE.format(display_id=self._dive_display_id,
+                                        jsonstr=jsonstr,
+                                        sprite_size=sprite_size)
+      display(HTML(html))
+
+  def _display_overview(self, data, update=None):
+    if _facets_gfsg_ready:
+      gfsg = GenericFeatureStatisticsGenerator()
+      proto = gfsg.ProtoFromDataFrames(
+          [{'name': 'data', 'table': data}])
+      protostr = base64.b64encode(proto.SerializeToString()).decode('utf-8')
+    else:  # GenericFeatureStatisticsGenerator is not supported.
+      protostr = ''
+    if update:
+      script = _OVERVIEW_SCRIPT_TEMPLATE.format(
+          display_id=update,
+          protostr=protostr)
+      display_javascript(Javascript(script))
+    else:
+      html = _OVERVIEW_HTML_TEMPLATE.format(
+          display_id=self._overview_display_id,
+          protostr=protostr)
+      display(HTML(html))
+
+  def _display_dataframe(self, data, update=None):
+    if update:
+      table_id = 'table_{}'.format(update)
+      html = _DATAFRAME_PAGINATION_TEMPLATE.format(
+          dataframe_html=data.to_html(notebook=True,
+                                      table_id=table_id),
+          table_id=table_id)
+      update_display(HTML(html), display_id=update)
+    else:
+      table_id = 'table_{}'.format(self._df_display_id)
+      html = _DATAFRAME_PAGINATION_TEMPLATE.format(
+          dataframe_html=data.to_html(notebook=True,
+                                      table_id=table_id),
+          table_id=table_id)
+      display(HTML(html), display_id=self._df_display_id)
+
+  def _to_element_list(self):
+    pcoll_list = []
+    if ie.current_env().cache_manager().exists('full', self._cache_key):
+      pcoll_list, _ = ie.current_env().cache_manager().read('full',
+                                                            self._cache_key)
+    return pcoll_list
+
+  def _to_dataframe(self):
+    normalized_list = []
+    # Column name for _one_dimension_types if presents.
+    normalized_column = str(self._pcoll)
+    # Normalization needs to be done for each element because they might be of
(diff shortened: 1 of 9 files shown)


KevinGG:
R:rohdesamuel
PTAL. Will fix the precommit checks once they are done add aaltay@ later.

KevinGG:
R: @aaltay
R: @rohdesamuel

PTAL, thanks!

aaltay:
I will wait for @rohdesamuel to make the first review pass. Could you also check the failing tests please?

KevinGG:
Hi @pabloem , thanks for volunteering to review this PR!
I've fixed most of the test related issues.
There is a Python fatal stack overflow error that I couldn't really make sense of. It seems that the error exists before my change but past Jenkins tests skipped it due to some execution ordering reason? I've asked the question to the dev mail list: https://mail.google.com/mail/u/0?ui=2&ik=64050edf4e&view=lg&permmsgid=msg-f%3A1647143158414277283&ser=1

aaltay:
Any updates @rohdesamuel @KevinGG ?

KevinGG:
retest this please.

KevinGG:
retest this please.

KevinGG:
retest this please.

pabloem:
FYI you don't need to retest the whole thing. You just need to rerun the specific tests that are failing (e.g. Run Portable_Python PreCommit / Run Python PreCommit / Run PythonLint PreCommit / ....)

KevinGG:
> FYI you don't need to retest the whole thing. You just need to rerun the specific tests that are failing (e.g. Run Portable_Python PreCommit / Run Python PreCommit / Run PythonLint PreCommit / ....)

Got it. Thanks!
Now the tests have passed, we can start reviewing it.

sdks/python/setup.py
```diff
@@ -107,13 +107,17 @@ def get_version():
     'crcmod>=1.7,<2.0',
     # Dill doesn't guarantee comatibility between releases within minor version.
     'dill>=0.3.0,<0.3.1',
+    'facets-overview>=1.0.0,<2',
```
aaltay: This change is significantly expanding the required Beam dependencies. These dependencies are by default not used by existing users and also would not be needed in execution environments.

I suggest making another extras_require similar to gcp and group all interactive dependency requirements under that group.

sdks/python/apache_beam/runners/interactive/interactive_environment.py
```diff
@@ -105,3 +113,33 @@ def set_cache_manager(self, cache_manager):
   def cache_manager(self):
     """Gets the cache manager held by current Interactive Environment."""
     return self._cache_manager
+
+  def set_pipeline_result(self, pipeline, result):
```
rohdesamuel: Why do we need this method? It looks like it's only being used in tests.

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization_test.py
```diff
@@ -0,0 +1,133 @@
...
+  def test_raise_error_for_non_pcoll_input(self):
+    class Foo(object):
+      pass
+
+    with self.assertRaises(ValueError) as ctx:
+      pv.PCollVisualization(Foo())
+      self.assertTrue('pcoll should be apache_beam.pvalue.PCollection' in
+                      ctx.exception)
+
+  @unittest.skipIf(sys.version_info < (3, 5, 3),
```
rohdesamuel: Is there a way to do this check only once? Maybe in the module's "if __name__ == "main""?

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
```diff
@@ -0,0 +1,258 @@
...
+  always returns immediately and is asynchronous when dynamical plotting is on.
+
+  If dynamical plotting enabled, the visualization is updated continuously until
+  the pipeline producing the PCollection is in an end state. The visualization
+  would be anchored to the notebook cell output area. The function
+  asynchronously returns a handle to the visualization job immediately. The user
+  could manually do::
+
+    # In one notebook cell, enable dynamical plotting every 1 second:
+    handle = visualize(pcoll, dynamical_plotting_interval=1)
```
rohdesamuel: Please change "dynamical" to "dynamic" here and elsewhere. Dynamic is the proper antonym of static, which is what I assume you are going for.

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
```diff
@@ -0,0 +1,258 @@
...
+  pv = PCollVisualization(pcoll)
+  pv.display_facets()
+
+  if dynamical_plotting_interval:
+    # Disables the verbose logging from timeloop.
+    logging.getLogger('timeloop').disabled = True
+    tl = Timeloop()
+
+    def dynamical_plotting(pcoll, pv, tl):
+      @tl.job(interval=timedelta(seconds=dynamical_plotting_interval))
```
rohdesamuel: Very cool!

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
```diff
@@ -0,0 +1,258 @@
...
+
+  The class relies on creating a PipelineInstrument w/o actual instrument to
+  access current interactive environment for materialized PCollection data at
+  the moment of self instantiation through cache. It utilizes Facets, pandas
+  DataFrame and jQuery DataTable to visualize the PCollection data.
+  """
+
+  def __init__(self, pcoll):
+    if not isinstance(pcoll, pvalue.PCollection):
+      raise ValueError('pcoll should be apache_beam.pvalue.PCollection')
```
rohdesamuel: An assert would be cleaner here.

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
```diff
@@ -0,0 +1,258 @@
...
+    for el in self._to_element_list():
+      if self._is_one_dimension_type(el):
+        # Makes such data structured.
+        normalized_list.append({normalized_column: el})
+      else:
+        normalized_list.append(_pv_jsons_load(_pv_jsons_dump(el)))
+    # Creates a dataframe that str() 1-d iterable elements after
+    # normalization so that facets_overview can treat such data as categorical.
+    return json_normalize(normalized_list).applymap(
+        lambda x: str(x) if type(x) in (list, tuple) else x)
```
rohdesamuel: I think this is a super useful method for InteractiveRunner users in general. Would it be possible to move this to the InteractiveRunner class so that anyone can take advantage of converting a PCollection to a Dataframe?

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
```diff
@@ -0,0 +1,258 @@
...
+            <script src="https://ajax.googleapis.com/ajax/libs/jquery/2.2.2/jquery.min.js"></script> 
+            <script src="https://cdn.datatables.net/1.10.16/js/jquery.dataTables.js"></script> 
+            <link rel="stylesheet" href="https://cdn.datatables.net/1.10.16/css/jquery.dataTables.css">
+            {dataframe_html}
+            <script>
+              $("#{table_id}").DataTable();
+            </script>"""
+
+
+def visualize(pcoll, dynamical_plotting_interval=None):
```
rohdesamuel: How does one stop visualizing a PCollection?

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization_test.py
```diff
@@ -0,0 +1,133 @@
...
+  @patch('apache_beam.runners.interactive.display.pcoll_visualization'
+         '.PCollVisualization._to_element_list', _mock_to_element_list)
+  @patch('apache_beam.runners.interactive.display.pcoll_visualization'
+         '.PCollVisualization.display_facets')
+  def test_dynamical_plotting_update_same_display(self,
+                                                  mocked_display_facets):
+    # Starts async dynamical plotting.
+    h = pv.visualize(self._pcoll, dynamical_plotting_interval=0.001)
+    # Blocking so the above async task can execute a few iterations.
+    time.sleep(0.1)
```
rohdesamuel: Please don't use time.sleep() in unit tests. Not only does it slow down the testing, but it also generally leads to flaky tests, and is a bad code smell.

sdks/python/setup.py
KevinGG: Thanks for the suggestions!
Yes, I feel the same way but didn't know what to do about it. Will make the change.

sdks/python/apache_beam/runners/interactive/interactive_environment.py
KevinGG: It's not used yet. It should be invoked by the InteractiveRunner once we have all the building blocks ready.
What we are planning to do is to only allow one job for one user pipeline running at the same time within a runner instance.
So if the user re-executes p.run() for an async running job.  The runner should cancel the running job using current tracked pipeline_result and starts a new one, then track the new  pipeline_result.

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization_test.py
KevinGG: Yes, you can do it by putting skipIf at the class level.
It's just one of the tests is depending on (3,6,3) (where assert_called is introduced in unittest). So not all tests are depending on the same version. Decorating at function level gives more flexibility.

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
KevinGG: Thanks, I'll go for it and change it from everywhere.

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
KevinGG: Yes, I'll change it to assert!

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
KevinGG: We've discussed whether we are planning to expose such API to users. Temporarily, the schema support is not done and future Dataframe API work is planned for next Q1. We'll consolidate all the related work and expose such API when everything is clear.
Here the implementation is a naive python solution. So probably don't expose it for general usage yet.

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
KevinGG: It automatically checks the running job's pipeline result and stop itself when the job is in an terminated state.
The end user could also assign the returned handle from visualize() to a variable, say `handle = visualize(pcoll)` and call `handle.stop()` to explicitly manually stop the visualization from anywhere in the notebook.

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization_test.py
KevinGG: Using while loop querying for an ending condition to block then. Also added a timeout to ensure the while loop to never exceed 0.1 second. Even if the logic runs into issue, it will just fail and will not block the test forever.

sdks/python/setup.py
```diff
@@ -219,11 +227,15 @@ def run(self):
...
-    tests_require=REQUIRED_TEST_PACKAGES,
+    tests_require=[
+        REQUIRED_TEST_PACKAGES,
+        INTERACTIVE_BEAM,
+    ],
     extras_require={
         'docs': ['Sphinx>=1.5.2,<2.0'],
         'test': REQUIRED_TEST_PACKAGES,
         'gcp': GCP_REQUIREMENTS,
+        'ib': INTERACTIVE_BEAM,
```
aaltay: how about `interactive` instead of `ib`. This leads to a more readable `pip install apache_beam[interactive]` instead of `pip install apache_beam[ib]`

sdks/python/setup.py
```diff
@@ -219,11 +227,15 @@ def run(self):
     install_requires=REQUIRED_PACKAGES,
     python_requires=python_requires,
     test_suite='nose.collector',
-    tests_require=REQUIRED_TEST_PACKAGES,
+    tests_require=[
+        REQUIRED_TEST_PACKAGES,
+        INTERACTIVE_BEAM,
```
aaltay: Why tests would require interactive beam packages?

KevinGG: Because there are unit tests around interactive beam packages.
When those packages are under REQUIRED_PACKAGES, they are always installed and available during tests.
Once we move them into an extras_require, they are not installed nor available during tests anymore.
For example, *-gcp tests will install [gcp,test] to pick up the extras_package GCP_REQUIREMENTS.
Since we don't have/need *-interactive test suites, we can make the group to always be installed as required test packages for tests.

sdks/python/setup.py
KevinGG: Agreed!

KevinGG:
isort and pylint are crazy and contradicting about where you should put `import apache_beam as beam`. And the order suggested changes over time.

KevinGG:
Run Python PreCommit

rohdesamuel:
lgtm

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization_test.py
```diff
@@ -0,0 +1,152 @@
...
+    # The following iterations use the same updating_pv to display_facets and so
+    # on.
+    _, second_kwargs = mocked_display_facets.call_args_list[1]
+    updating_pv = second_kwargs['updating_pv']
+    for call in mocked_display_facets.call_args_list[2:]:
+      _, kwargs = call
+      self.assertIs(kwargs['updating_pv'], updating_pv)
+    h.stop()
+
+  @unittest.skipIf(sys.version_info < (3, 6),
```
davidyan74: This is inconsistent with the rest of the tests. Perhaps make the rest of the tests 3.6+?

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
```diff
@@ -0,0 +1,258 @@
...
+from apache_beam.runners.interactive import pipeline_instrument as instr
+from facets_overview.generic_feature_statistics_generator import GenericFeatureStatisticsGenerator
+from IPython.core.display import HTML
+from IPython.core.display import Javascript
+from IPython.core.display import display
+from IPython.core.display import display_javascript
+from IPython.core.display import update_display
+from timeloop import Timeloop
+
+# jsons doesn't support < Python 3.5. Work around with json for legacy tests.
```
davidyan74: Why do we need this workaround if we only support Python 3.5+?

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization_test.py
KevinGG: The version is for testing code, not for code being tested.
The code being tested supports 3.5.3+
This specific test has assertion feature that was introduced in 3.6.

So the first 3 tests will be executed in both Py36 and Py37 test suites.
The last test will only be executed in Py37 suite.

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
KevinGG: All the Gradle tasks still supports py2.
So you still have docs, lint and tasks running in py2 virtual env. We'll have to pass those pre-commit checks.

davidyan74: Can we just print a warning and exit, or exclude this from running in py2? We probably don't want this to create an impression that this will work with py2.

KevinGG: This will still work Py2 though. And a sys.exit() would probably break some gradle tasks waiting for 0 exit code.
I've put a TODO item here. It's a general ticket for Interactive Beam to mark for cleanup once py2 is deprecated completely from Beam.

davidyan74: Can we at least print a warning that says "interactive is not supported in py2", like we do for python 3 with beam <= 2.15?

KevinGG: I've added such warning message in the interactive_environment module to check for Python version and print out warnings.

KevinGG:
R: @aaltay 
The tests have passed and Sam has completed his review. Do you have any other comments for this PR?

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
```diff
@@ -0,0 +1,279 @@
...
+# TODO(BEAM-8288): clean up once Py2 is deprecated from Beam.
+try:
+  import jsons
+  _pv_jsons_load = jsons.load
+  _pv_jsons_dump = jsons.dump
+except ImportError:
+  import json
+  _pv_jsons_load = json.load
```
aaltay: If `json` works, why is `jsons` needed?

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
```diff
@@ -0,0 +1,279 @@
...
+  _pv_jsons_dump = json.dump
+
+try:
+  from facets_overview.generic_feature_statistics_generator import GenericFeatureStatisticsGenerator
+  _facets_gfsg_ready = True
+except ImportError:
+  _facets_gfsg_ready = False
```
aaltay: Is there a reason for _facets_gfsg_ready, _ipython_ready, _tl_ready to be separately ready? Does this module work with some of them installed and some not?

I think this import section could be reduced to 

try:
... import all
except ImportError
... warning, missing dependencies and ask user to install [interactive]

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
```diff
@@ -0,0 +1,279 @@
...
+except ImportError:
+  _tl_ready = False
+
+# 1-d types that need additional normalization to be compatible with DataFrame.
+_one_dimension_types = (int, float, str, bool, list, tuple)
+
+_DIVE_SCRIPT_TEMPLATE = """
+            document.querySelector("#{display_id}").data = {jsonstr};"""
+_DIVE_HTML_TEMPLATE = """
+            <script src="https://cdnjs.cloudflare.com/ajax/libs/webcomponentsjs/1.3.3/webcomponents-lite.js"></script>
```
aaltay: I am not sure how to verify that this template is correct. How can we test this?

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
```diff
@@ -0,0 +1,279 @@
...
+            pass
+
+      tl.start()
+      return tl
+
+    return dynamic_plotting(pcoll, pv, tl)
+  return None
+
+
+class PCollVisualization(object):
```
aaltay: Should this be named `PCollectionVisualization` because the name of what it visualizes is `PCollection` and not the short form.

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
```diff
@@ -0,0 +1,279 @@
...
+  """A visualization of a PCollection.
+
+  The class relies on creating a PipelineInstrument w/o actual instrument to
+  access current interactive environment for materialized PCollection data at
+  the moment of self instantiation through cache. It utilizes Facets, pandas
```
aaltay: I think you can drop the last sentence. What it uses to visualize does not need to be part of the documentation.

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
```diff
@@ -0,0 +1,279 @@
...
+                                                               id(self))
+    self._df_display_id = 'df_{}_{}'.format(self._cache_key, id(self))
+
+  def display_facets(self, updating_pv=None):
+    """Displays the visualization through IPython.
+
+    The visualization has 3 parts: facets-dive, facets-overview and paginated
+    data table. Each part is assigned an auto-generated unique display id
+    (the uniqueness is guaranteed throughout the lifespan of the PCollection
+    variable). If a PCollVisualization instance is provided as updating_pv,
```
aaltay: You might document updating_pv and extract out that part of the single paragraph here.

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization_test.py
```diff
@@ -0,0 +1,152 @@
...
+
+class PCollVisualizationTest(unittest.TestCase):
+
+  def setUp(self):
+    self._p = beam.Pipeline()
+    # pylint: disable=range-builtin-not-iterating
+    self._pcoll = self._p | 'Create' >> beam.Create(range(1000))
+
+  @unittest.skipIf(sys.version_info < (3, 5, 3),
+                   'PCollVisualization is not supported on Python 2.')
```
aaltay: This is more like not supported < 3.5.3.

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization_test.py
aaltay: It would be useful to add your response as code comments here.

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization_test.py
```diff
@@ -0,0 +1,152 @@
...
+  def test_auto_stop_dynamic_plotting_when_job_is_terminated(
+      self,
+      mocked_timeloop,
+      mocked_display_facets):
+    fake_pipeline_result = runner.PipelineResult(runner.PipelineState.RUNNING)
+    ie.current_env().set_pipeline_result(self._p, fake_pipeline_result)
+    # Starts non-stopping async dynamic plotting until the job is terminated.
+    pv.visualize(self._pcoll, dynamic_plotting_interval=0.001)
+    # Blocking so the above async task can execute at least 3 iterations.
+    timeout = time.time() + 1
```
aaltay: This might be replaced with time.sleep(1) to simplify it.

From the while loop and brake, we cannot guarantee the 3 time execution anyway.

sdks/python/apache_beam/runners/interactive/interactive_environment.py
```diff
@@ -61,8 +66,20 @@ def __init__(self, cache_manager=None):
...
+    # Each key is a pipeline instance defined by the end user. The
+    # InteractiveRunner is responsible for populating this dictionary
+    # implicitly.
+    self._pipeline_results = {}
     # Always watch __main__ module.
     self.watch('__main__')
+    # Do a warning level logging if current python version is below 2 or 3.5.3.
+    if sys.version_info < (3,):
+      logging.warning('Interactive Beam does not support Python 2.')
+    elif sys.version_info < (3, 5, 3):
```
aaltay: Should we simply say interactivity required 3.5.3 ?

sdks/python/apache_beam/runners/interactive/interactive_environment_test.py
```diff
@@ -88,6 +91,72 @@ def test_watch_class_instance(self):
...
+    ie.current_env().set_pipeline_result(pipeline, pipeline_result)
+    self.assertIs(ie.current_env().pipeline_result(pipeline), pipeline_result)
+
+  def test_determine_terminal_state(self):
+    for state in (runner.PipelineState.DONE,
+                  runner.PipelineState.FAILED,
+                  runner.PipelineState.CANCELLED,
+                  runner.PipelineState.UPDATED,
+                  runner.PipelineState.DRAINED):
+      ie.current_env().set_pipeline_result(self._p, runner.PipelineResult(
```
aaltay: PipelineState has an is_terminal method defined on the class that can be used here.

sdks/python/setup.py
```diff
@@ -166,6 +166,14 @@ def get_version():
     'google-cloud-bigtable>=0.31.1,<1.1.0',
 ]
 
+INTERACTIVE_BEAM = [
+    'facets-overview>=1.0.0,<2',
```
aaltay: Could you sort these dependencies?

sdks/python/setup.py
aaltay: We can add a tox suite to test interactivity similar to gcp suites and not install these dependencies for all tests.

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
KevinGG: `json` wouldn't support all the data structure that might come from the user's data.
But `jsons` and `json` have similar usage and interfaces. So it's a try-best effort to serialize the user's data. We can also simply NOOP if jsons is not available.
Here is the reference _Why not use the standard json library_: https://jsons.readthedocs.io/en/latest/faq.html#why-not-use-the-standard-json-library

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
KevinGG: They can work separately.
If not _ipython_ready, the user is not even in an interactive environment. There is no kernel, so there is definitely no frontend connected to the kernel. Thus no notebook, nowhere to visualize, so basically everything is NOOP.
If not _facets_gfsg_ready, the user wouldn't see the facets-dive widget, but they can see facets-overview and data-table widgets.
If not _tl_ready, the user wouldn't see dynamic plotting but one-shot plotting.

Warning about the user to install [interactive] is a great suggestion!

I can 
1. make it import all and NOOP if any of the dependency is not ready
2. make it import separately and try-best to do visualization
Either way, I'll add the warning for users to `pip install apache-beam[interactive]`. Which option do you like?

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
KevinGG: The `facets` templates originate from examples in notebooks shared among the OSS community.
The `jquery` template simply uses a jquery DataTable widget.
For unittest, the logic of each template is too simple that there is nothing really to be tested.
We'll have some integration test later to execute notebook and verify output area. Such integration test framework is being developed by some other team. We'll merge our efforts there.
For manual test, I've manually tested them from notebook. I can attach png and gif if needed.

Off the topic:
In future PRs, we might have `%%javascript` magic involved in the python code. Then we can have some library such as `js2py` for javascript unittests in Python.

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization_test.py
KevinGG: Sam was against using time.sleep() in test code because they block the test for exactly that amount of time. I don't know if using a while loop really helps with that. Worst case, it also blocks for ~1s and uses a little bit more CPU. Best case, it unblocks faster than time.sleep(1) and doesn't have potential side effects than time.sleep(1) based on the sleep implementation.

It's a tiny bit of trade-off, which one do you recommend? I'm fine with time.sleep(1) though.

sdks/python/apache_beam/runners/interactive/interactive_environment.py
KevinGG: We can. I'll make it simpler. Let's just consider every different component a big feature. The feature requires Py 3.5.3. And we NOOP if dependency is not ready.

sdks/python/apache_beam/runners/interactive/interactive_environment_test.py
KevinGG: Great, thanks! I totally missed it. I'll change it to use that class method.

sdks/python/setup.py
KevinGG: Yes, I agree! This could avoid running those tests when python version is under 3.5.3.
I guess then the only task would require [interactive] for older versioned python will be "docs" where it scans the source code. I'll just add the try-import statements everywhere to work around it.

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
KevinGG: Will change to the full name.

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization_test.py
KevinGG: Added

sdks/python/setup.py
KevinGG: They are already sorted. Do you want me to also sort GCP_REQUIREMENTS?

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
KevinGG: Done.

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization_test.py
KevinGG: Changing the error message.

KevinGG:
I've added two tox env and corresponding Py36 and Py37 test suite. Removed [interactive] from test required packages. Added it to py2:docs env.
I'll add [interactive] for other specific envs defined in tox.ini if any of the gradle task needs it.

I've also added check and warnings in `interactive_environment.InteractiveEnvironment` to check prerequisites for interactive beam (didn't put them in the global scope because global scoped warning level logging will fail gradle tasks in pre-commit):
1. If Python is above 3.5.3;
2. If [interactive] dependencies are fully available;
3. If current runtime is within an interactive environment, i.e., within an ipython kernel.

For the `time.sleep()` and `pcoll_visualization` NOOP strategy when dependencies are not fully installed, please let me know your preference, @aaltay. Personally I'm fine with either approaches, and I'll make change if needed.

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
aaltay: I would suggest first option. Require installation of full [interactive] extra package.

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
aaltay: Integration tests sounds good. If not with this PR, I will suggest doing sooner.

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization_test.py
aaltay: loop with time.time() > timeout is practically a time.sleep. I agree with Sam that it would be better to not having timing dependencies in tests but this test already has it. Changing from sleep to this version does not address the underlying problem.

I would prefer to simplify:

```
    # Blocking so the above async task can execute at least another 3
    # iterations.
    timeout = time.time() + 1
    while len(mocked_display_facets.call_args_list) < 6:
      if time.time() > timeout:
        break
```
to 
`time.sleep(1)`

sdks/python/setup.py
aaltay: Nevermind. This looks good.

sdks/python/tox.ini
```diff
@@ -189,7 +189,7 @@ commands =
   time {toxinidir}/scripts/run_pylint.sh
 
 [testenv:docs]
-extras = test,gcp,docs
+extras = test,gcp,docs,interactive
```
aaltay: Does docs need it?

sdks/python/tox.ini
```diff
@@ -208,3 +208,13 @@ commands =
   coverage report --skip-covered
   # Generate report in xml format
   coverage xml
+
+[testenv:py36-interactive]
+extras = test,interactive
+commands =
+  python setup.py nosetests --where 'apache_beam/runners/interactive' {posargs}
```
aaltay: Is it possible that anything else will be broken in interactive environment?

sdks/python/tox.ini
KevinGG: Yes, `docs` needs it because it's generating documentation for all source code. Kind of like a static scan. That's why `test` and `gcp` packages have also been included here.

sdks/python/tox.ini
KevinGG: TL;DR: For simplicity, I'll just change to run all tests.

It's possible that after setting up a Py36 (or any other) environment with interactive dependencies, some dependencies would conflict or be incompatible. Like the rollback we had, in some test setup (could be totally different from the tox env we have in this repo), some execution path used a newer versioned IPython through transitive dependency causing package usage to fail if there is no try-except for relative import path.
To ensure at least all test environments work without conflict in dependencies, we can use the same `commands` used by `testenv:py36-gcp`. It's just now we rerun many tests simply because the result of `pip install ...` might be different.
But the test suites are executed in parallel, this just adds more workload to Jenkins without increasing sequenced task elapse, so it should be fine.

KevinGG:
> Thank you.
> 
> Could you add your changes in new commits, instead of force pushing. Reviewing incremental changes that way is easier.

Yes, I'll do that from now on. Thanks!

aaltay:
Looks cleaner. Could you ping once all open comments are resolved.

(I will be OOO most of next week but @pabloem can finish the review.)

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
KevinGG: Yes, we'll do the integration tests as soon as the framework is ready, and OSS those tests when the framework supports OSS.

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization_test.py
KevinGG: Simplified it.

KevinGG:
> Looks cleaner. Could you ping once all open comments are resolved.
> 
> (I will be OOO most of next week but @pabloem can finish the review.)

Sure, thanks! I think all the open comments are marked as resolved. I'll let Pablo go over it once more next week. Have a nice weekend~

sdks/python/apache_beam/runners/runner.py
```diff
@@ -84,6 +85,10 @@ def create_runner(runner_name):
         raise ImportError(
             'Google Cloud Dataflow runner not available, '
             'please install apache_beam[gcp]')
+      elif 'interactive' in runner_name.lower():
+        raise ImportError(
```
aaltay: Have you tested this to be working as expected?

sdks/python/test-suites/tox/py37/build.gradle
```diff
@@ -40,6 +40,10 @@ test.dependsOn testPy37Gcp
 
 toxTask "testPy37Cython", "py37-cython"
 test.dependsOn testPy37Cython
+
+toxTask "testPy37Interactive", "py37-interactive"
```
aaltay: Maybe just limit tests to py 3.7 to limit the added initial test time.

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization_test.py
```diff
@@ -0,0 +1,156 @@
...
+import sys
+import time
+import unittest
+
+import apache_beam as beam
+from apache_beam.runners import runner
+from apache_beam.runners.interactive import interactive_environment as ie
+from apache_beam.runners.interactive.display import pcoll_visualization as pv
+
+# Work around nose tests using Python2 without unittest.mock module.
```
aaltay: Could you add a py2 todo here, for this to be removed later. You might be able to use BEAM-1251 or maybe @tvalentyn can suggest a better JIRA.

tvalentyn: I think this part is OK. We will later have to search the codebase for all comments including Python2, Py2, Python 3, etc for this cleanup. If you want a JIRA, feel free to use BEAM-7372.

sdks/python/apache_beam/runners/runner.py
KevinGG: Yes, tested locally by moving the `_ALL_KNOWN_RUNNERS` for `InteractiveRunner` to a non-existing path and when creating the runner, it raises the ImportError.

sdks/python/test-suites/tox/py37/build.gradle
KevinGG: Got it, I've removed py 3.6 configuration and tasks.

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization_test.py
KevinGG: Thanks @aaltay and @tvalentyn, I also have a [JIRA](https://issues.apache.org/jira/browse/BEAM-8288) to clean up for interactive packages. I'll use that one for this too.

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
```diff
@@ -0,0 +1,269 @@
...
+            <script src="https://ajax.googleapis.com/ajax/libs/jquery/2.2.2/jquery.min.js"></script> 
+            <script src="https://cdn.datatables.net/1.10.16/js/jquery.dataTables.js"></script> 
+            <link rel="stylesheet" href="https://cdn.datatables.net/1.10.16/css/jquery.dataTables.css">
+            {dataframe_html}
+            <script>
+              $("#{table_id}").DataTable();
+            </script>"""
+
+
+def visualize(pcoll, dynamic_plotting_interval=None):
```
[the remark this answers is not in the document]
KevinGG: Yes, I'll mark it as experimental! Actually the whole module is marked as `For internal use only; no backwards-compatibility guarantees.`

For renaming, we only need to touch the `visualize()` function that has not been implemented in module `interactive_beam`. Since a notebook user should only use APIs provided in that module.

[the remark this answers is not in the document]
pabloem: SGTM.

sdks/python/apache_beam/runners/interactive/display/pcoll_visualization.py
```diff
@@ -0,0 +1,269 @@
...
+    jsonstr = data.to_json(orient='records')
+    if update:
+      script = _DIVE_SCRIPT_TEMPLATE.format(display_id=update,
+                                            jsonstr=jsonstr)
+      display_javascript(Javascript(script))
+    else:
+      html = _DIVE_HTML_TEMPLATE.format(display_id=self._dive_display_id,
+                                        jsonstr=jsonstr,
+                                        sprite_size=sprite_size)
+      display(HTML(html))
```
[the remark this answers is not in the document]
KevinGG: For ipython terminal, the users would see something similar to below:
```
<IPython.core.display.HTML object>
<IPython.core.display.HTML object>
<IPython.core.display.HTML object>
```
I think the `show`(`visualization`) here heavily depends on HTML/JS/CSS to make sense (because it tries to load the whole PCollection cache into memory for rendering), so if the kernel is connected to a frontend that supports these, they should all support the visualization.

We can add a default pretty print that works similar to existing `print(result.get_list(pcoll))` and it's nothing close to what current `visualization()` does. However, the kernel doesn't know whether it is connected to a frontend / what kinds of frontends, so it's either always pretty-print or not doing it at all.
I would propose limit the work here for the scenarios when the kernel is actually connected to a frontend. And data visualization in a plain ipython terminal is not what we want to achieve here since one cannot interact with the rendering in a terminal easily (with mouse and etc.).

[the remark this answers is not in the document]
pabloem: I see. I don't want to block on this, so LGTM. But robert had also mentioned that he wanted this to be useful outside of a Jupyter notebook. Also the Ipython terminal. We can discuss this separately.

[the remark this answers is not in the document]
KevinGG: I mean I can add something to print a dataframe-printing-like output in terminal that samples some of the data and use some "hacky" way (details can be found in this stackoverflow [discussion](https://stackoverflow.com/questions/15411967/how-can-i-check-if-code-is-executed-in-the-ipython-notebook)) to determine if the kernel is connected to an IPython notebook.

I'll send an additional commit in this PR to implement that.

But we are not going to provide other GUI support in terminal for now.

KevinGG:
Run Python PreCommit

KevinGG:
Run Python PreCommit

KevinGG:
Run Python PreCommit

pabloem:
Run Python PreCommit

pabloem:
Squashed and merged after Ahmet LGTMd, and myself as well.

PR #9745, opened 2019-10-08
[BEAM-8352] Add "withMaxCapacityPerShard()" to KinesisIO.Read

aromanenko-dev:
Added new configuration option `withMaxCapacityPerShard()` to `KinesisIO.Read` to provide a possibility to control the size of `ShardReadersPool.recordsQueue` in case if default value (10K records per shard) can cause NPE for topics with big amount of shards or large messages that can't fit into memory.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/KinesisIO.java
+++ b/sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/KinesisIO.java
@@ -239,6 +239,7 @@ public final class KinesisIO {
         .setMaxNumRecords(Long.MAX_VALUE)
         .setUpToDateThreshold(Duration.ZERO)
         .setWatermarkPolicyFactory(WatermarkPolicyFactory.withArrivalTimePolicy())
+        .setMaxCapacityPerShard(ShardReadersPool.DEFAULT_CAPACITY_PER_SHARD)
         .build();
   }
 
@@ -272,6 +273,8 @@ public final class KinesisIO {
 
     abstract WatermarkPolicyFactory getWatermarkPolicyFactory();
 
+    abstract Integer getMaxCapacityPerShard();
+
     abstract Builder toBuilder();
 
     @AutoValue.Builder
@@ -292,6 +295,8 @@ public final class KinesisIO {
       abstract Builder setRequestRecordsLimit(Integer limit);
 
       abstract Builder setWatermarkPolicyFactory(WatermarkPolicyFactory watermarkPolicyFactory);
+
+      abstract Builder setMaxCapacityPerShard(Integer maxCapacity);
 
       abstract Read build();
     }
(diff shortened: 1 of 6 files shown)


lukecwik:
Can't we instead use some other limit such as sum of queued message sizes to limit memory pressure?

aromanenko-dev:
@lukecwik I think we **can** do this by adding a consumed messages size counter and checking this before putting new messages into queue. If total size is exceeded then we can stop and block fetching and adding new messages until the queue will be dequeued and there will be a free space for new records. 
In the same time, it will be a more complicated implementation comparing to what we have now and I don't see too much profit from it. Do I miss something?

lukecwik:
Your right that you can provided an immediate solution right now and this is an acceptable change that will solve a real problem.

What I wanted to highlight about "tuning" flags in general is that:
1) Users need to know to set it to some value but what value can we suggest to them?
2) It doesn't work dynamically over time so that users need to choose a value that is very conservative.

aromanenko-dev:
@lukecwik Yes, I agree that we need to reduce the number of tuning knobs to as less as possible and try to configure transforms automatically and dynamically if it's feasible.  But here is a question how we can estimate a maximum size in bytes in this case...

lukecwik:
You can use the size of the KinesisRecords data field or you can assume each is the maximum size of 1MiB (https://docs.aws.amazon.com/kinesis/latest/APIReference/API_Record.html).

aromanenko-dev:
@lukecwik Regarding your questions:

1. Default value was (and stays) 10K records per shard and it worked well for topics with not large amount of shards per worker (we've got this issue only after 2 years of KinesisIO utilisation by different users). So, in general, we can conclude that user doesn't need to configure it manually and this option has to be used only in corner cases.

2. Yes, this is a field for improvement. Though, it's not very clear under which conditions user would need to be able change this option in the runtime dynamically.
> you can assume each is the maximum size of 1MiB
I mean the maximum size of available memory that can be allocated for this queue. Actually, it goes to more complicated question how effectively implement backpressure to have effectively working pair of consumer/producer in terms of consumed memory and total performance.

lukecwik:
Yeah, the process needs to have a memory budget and every "largeish" thing should be accounted for. I have been thinking about how to best do this for the Java portability container since it would allow for removing several knobs that already exist.

aromanenko-dev:
I agree that limitation based on payload size would be better, though it's still unclear how we can manage it dynamically.
1. Should be initial value based on process heap size as default value or configured by user? 
2. How much of free space we should leave unused to prevent NPE because of new arrived records (we don't know their size in advance)? 
3. Do we need to adapt the maximum queue size in runtime? 
4. If yes, what are the causes for that?

Btw, do you already have a solution for Java portability container problem? Either maybe we have something similar implemented for other IO?

kennknowles:
This is associated with a bug marked release-blocking. Is that accurate? Is it important enough that we should go with an early approach and make improvements later? Maybe mark experimental/deprecate immediately if you know it will not last... or if it is not urgent to make 2.17.0 perhaps adjust the Jira.

aromanenko-dev:
@kennknowles This one is definitively not a release blocker since it's not a regression and this bug can occur only under some load conditions. We haven't come to final solution for the moment (I raised several questions in above discussion) and final fix may affect user API. So I'd not merge this (even as experimental) until we decide how to proceed further.  For now, I just removed `Fix Version/s` in associated Jira. Sorry for noice.

aromanenko-dev:
@lukecwik If we are agree on current solution (limit internal message queue by number of records) then I'm fine to merge it.

PR #9758, opened 2019-10-10
[BEAM-8374] Enable returning missing PublishResult fields in SnsIO.Write

jfarr:
This PR adds new coders for returning the entire PublishResult object from SnsIO, or the entire PublishResult without HTTP headers. This is necessary in order to check the HTTP status for errors since the current implementation only encodes the messageId field. These new coders can be selected by calling the new `withFullPublishResult()` and `withFullPublishResultWithoutHeaders()` methods on SnsIO.Write. The `withCoder()` method also allows setting a custom coder. This PR also adds coders for common AWS SDK objects `ResponseMetadata` and `SdkHttpMetadata`.
R: @lukecwik 

------------------------

--- | --- | --- | --- | ---


jfarr:
> Note that this is a breaking change to any Dataflow streaming pipelines using the PublishResult.

@lukecwik Thanks for the feedback. I did not realize that, I'm very new to Beam. What do you suggest would be the best way to deal with that?

sdks/java/io/amazon-web-services/src/test/java/org/apache/beam/sdk/io/aws/sns/AmazonSNSMockSuccess.java
```diff
@@ -29,10 +30,13 @@
   @Override
   public PublishResult publish(PublishRequest publishRequest) {
     PublishResult result = Mockito.mock(PublishResult.class);
-    SdkHttpMetadata metadata = Mockito.mock(SdkHttpMetadata.class);
-    Mockito.when(metadata.getHttpHeaders()).thenReturn(new HashMap<>());
-    Mockito.when(metadata.getHttpStatusCode()).thenReturn(200);
-    Mockito.when(result.getSdkHttpMetadata()).thenReturn(metadata);
+    SdkHttpMetadata httpMetadata = Mockito.mock(SdkHttpMetadata.class);
```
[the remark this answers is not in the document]
jfarr: Sure, I can make that change.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/sns/PublishResultCoder.java
```diff
@@ -42,12 +50,27 @@ static PublishResultCoder of() {
   public void encode(PublishResult value, OutputStream outStream)
       throws CoderException, IOException {
     StringUtf8Coder.of().encode(value.getMessageId(), outStream);
+    StringUtf8Coder.of().encode(value.getSdkResponseMetadata().getRequestId(), outStream);
```
iemejia: Since `SdkResponseMetadata` and `SdkHttpMetadata` seem to be used by multiple parts of AWS SDK I was wondering if we could create specific Coders (and tests) for both objects and just reuse them here. I know this could add some extra chores to this PR but we may benefit from them eventually in other IOs..

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/sns/PublishResultCoder.java
```diff
@@ -42,12 +50,27 @@ static PublishResultCoder of() {
...
+    MapCoder.of(StringUtf8Coder.of(), StringUtf8Coder.of())
+        .encode(value.getSdkHttpMetadata().getHttpHeaders(), outStream);
+    VarInt.encode(value.getSdkHttpMetadata().getHttpStatusCode(), outStream);
```
iemejia: nit: I will prefer the HTTP status code to be first just for consistency on the way HTTP works.

iemejia:
One extra issue that I have not really thought about is the size difference that we will have by adding the full headers/sdk metadata to all answers, mmm... now I am hesitating about maybe having a `V2` like approach for the cases where we care about the full metadata (akin to FileIO).
What is the use case that is pushing into this approach?

jfarr:
> Last week we merged SnsIO based on Amazon SDK for Java v2 I was wondering if these fixes could apply there too. We may consider this optional for this PR but maybe worth for consistency to check if we can have a similar approach there too. CC @cmachgodaddy

Sure, I work with Cam and we discussed it today. I can apply the changes to v2 in this PR, no problem.

cmachgodaddy:
I wonder if we should add unit test for this coder? I know we did not have it, b/c we just had one property to encode and decode. But, now we have quite a few.

jfarr:
@iemejia @lukecwik How does this look? If you're happy with it I'll port it over to `amazon-web-services2`.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/sns/PublishResultBuilder.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+package org.apache.beam.sdk.io.aws.sns;
+
+import com.amazonaws.ResponseMetadata;
+import com.amazonaws.http.HttpResponse;
+import com.amazonaws.http.SdkHttpMetadata;
+import com.amazonaws.services.sns.model.PublishResult;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.ImmutableMap;
+
+public class PublishResultBuilder {
```
cmachgodaddy: Shouldn't build this class using auto service (@AutoValue), so it can be consistent with Beam model? @iemejia and @lukecwik what do you think?

lukecwik: +1 on using AutoValue and its builder.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/sns/PublishResultCoder.java
lukecwik: Agree with @iemejia

jfarr:
@lukecwik @iemejia I reverted the original coder and added a new "full" (all fields) and "minimal" (no HTTP headers) coder, and also factored out new coders for the common SDK objects. If this looks more like it then I'll flesh out the unit tests, please let me know.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/sns/SnsIO.java
```diff
@@ -256,9 +269,36 @@ public Write withResultOutputTag(TupleTag<PublishResult> results) {
...
+          .setCoderProvider(
+              CoderProviders.forCoder(
+                  TypeDescriptor.of(PublishResult.class), MinimalPublishResultCoder.of()))
+          .build();
+    }
+
     @Override
     public PCollectionTuple expand(PCollection<PublishRequest> input) {
       checkArgument(getTopicName() != null, "withTopicName() is required");
+      if (getCoderProvider() != null) {
```
lukecwik: I don't think we want to use a coder provider here but just a coder directly and call `setCoder` on the resulting `PCollection` from `input.apply`

Coder providers are meant to allow users a way to bind object -> coder resolution logic for an entire pipeline (not just a single transform).

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/coders/MinimalSdkHttpMetadataCoder.java
```diff
@@ -0,0 +1,66 @@
...
+import java.io.OutputStream;
+import java.io.Serializable;
+import java.util.List;
+import org.apache.beam.sdk.coders.Coder;
+import org.apache.beam.sdk.coders.CoderException;
+import org.apache.beam.sdk.coders.VarIntCoder;
+import org.apache.beam.sdk.util.VarInt;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.ImmutableList;
+
+public class MinimalSdkHttpMetadataCoder extends Coder<SdkHttpMetadata> implements Serializable {
```
lukecwik: Style note, you can create one class called `AwsCoders` which has three static methods `minimalSdkHttpMetadata()` and `fullSdkHttpMetadata()` and `responseMetadata()` that return the coders. This way you can make all the coder implementations private inner static classes and they all become implementation details without needing to expose them to users. Users can still get an instance if they need one but it gives us flexibility from a maintenance point of view how the coders are organized.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/sns/SnsIO.java
```diff
@@ -256,9 +269,36 @@ public Write withResultOutputTag(TupleTag<PublishResult> results) {
...
+      return builder()
+          .setCoderProvider(
+              CoderProviders.forCoder(
+                  TypeDescriptor.of(PublishResult.class), FullPublishResultCoder.of()))
+          .build();
+    }
+
+    /**
+     * Encode the minimal {@code PublishResult} object, including HTTP status code but excluding
+     * HTTP response headers.
```
lukecwik: You might want to mention that this is the default here.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/sns/FullPublishResultCoder.java
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.beam.sdk.io.aws.sns;
+
+import com.amazonaws.services.sns.model.PublishResult;
```
lukecwik: Might as well make the publish result coders internal implementation details of the SnsIO class and/or an SnsCoders class similar to the AwsCoders class I described above.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/sns/SnsIO.java
jfarr: Actually I left the default behavior the same for backward compatibility, i.e. encode only the messageId using (what is now called) DefaultPublishResultCoder. This method selects the MinimalPublishResultCoder which encodes messageId, requestId, and statusCode but not the HTTP headers.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/sns/SnsIO.java
jfarr: I understand now, thanks.

jfarr:
Hi @lukecwik thank you for the feedback. I believe all of the requested changes have been addressed now.

jfarr:
Hi @lukecwik @iemejia this PR was marked stale. I believe I have addressed all of your feedback. Is there anything else you would like me to change?

lukecwik:
Sorry for the delay, taking a look now.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/coders/AwsCoders.java
```diff
@@ -0,0 +1,169 @@
...
+import org.apache.beam.sdk.coders.Coder;
+import org.apache.beam.sdk.coders.CoderException;
+import org.apache.beam.sdk.coders.MapCoder;
+import org.apache.beam.sdk.coders.StringUtf8Coder;
+import org.apache.beam.sdk.coders.VarIntCoder;
+import org.apache.beam.sdk.util.VarInt;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.ImmutableList;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.ImmutableMap;
+
+public final class AwsCoders {
```
lukecwik: Please add class comment.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/coders/AwsCoders.java
```diff
@@ -0,0 +1,169 @@
...
+    private ResponseMetadataCoder() {}
+
+    public static ResponseMetadataCoder of() {
+      return INSTANCE;
+    }
+
+    @Override
+    public void encode(ResponseMetadata value, OutputStream outStream)
+        throws CoderException, IOException {
+      MapCoder.of(StringUtf8Coder.of(), StringUtf8Coder.of())
```
lukecwik: The encoding of this changed where before we assumed this would always exist but now we handle the null case but we don't actually return the same object that was provided when decoding. Would you rather handle the null case by wrapping the response metadata coder returned in `responseMetadata` with the [NullableCoder](https://github.com/apache/beam/blob/master/sdks/java/core/src/main/java/org/apache/beam/sdk/coders/NullableCoder.java)?

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/coders/AwsCoders.java
```diff
@@ -0,0 +1,169 @@
...
+    private FullSdkHttpMetadataCoder() {}
+
+    public static FullSdkHttpMetadataCoder of() {
+      return INSTANCE;
+    }
+
+    @Override
+    public void encode(SdkHttpMetadata value, OutputStream outStream)
+        throws CoderException, IOException {
+      VarInt.encode(value != null ? value.getHttpStatusCode() : 0, outStream);
```
lukecwik: Same suggestion for wrapping NullableCoder within `fullSdkHttpMetadata` and not handling the null case here.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/coders/AwsCoders.java
```diff
@@ -0,0 +1,169 @@
...
+    private MinimalSdkHttpMetadataCoder() {}
+
+    public static MinimalSdkHttpMetadataCoder of() {
+      return INSTANCE;
+    }
+
+    @Override
+    public void encode(SdkHttpMetadata value, OutputStream outStream)
+        throws CoderException, IOException {
+      VarInt.encode(value != null ? value.getHttpStatusCode() : 0, outStream);
```
lukecwik: Same suggestion for wrapping NullableCoder within `minimalSdkHttpMetadata` and not handling the null case here.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/sns/PublishResultCoders.java
```diff
@@ -0,0 +1,155 @@
...
+import java.io.OutputStream;
+import java.io.Serializable;
+import java.util.List;
+import org.apache.beam.sdk.coders.Coder;
+import org.apache.beam.sdk.coders.CoderException;
+import org.apache.beam.sdk.coders.StringUtf8Coder;
+import org.apache.beam.sdk.io.aws.coders.AwsCoders;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.ImmutableList;
+
+public final class PublishResultCoders {
```
lukecwik: class comment

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/sns/PublishResultCoders.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+  }
+
+  private abstract static class AbstractPublishResultCoder extends Coder<PublishResult>
+      implements Serializable {
+
+    private static final Coder<ResponseMetadata> RESPONSE_METADATA_CODER =
+        AwsCoders.responseMetadata();
+
+    protected abstract Coder<SdkHttpMetadata> getHttpMetadataCoder();
```
lukecwik: nit: stylewise I think its simpler to understand if the abstract class has a constructor that takes the sdkhttpmetadata coder and you pass in the coder instance in the concrete subclass implementations instead of asking authors to override `getHttpMetadataCoder`

jfarr:
Hi @lukecwik and @iemejia, I believe I have incorporated all of your review feedback. I also made a few changes to the interface that I hope will improve clarity:

- Instead of the confusingly named `fullSdkHttpMetadata()` and `minimalSdkHttpMetadata()` methods, the `AwsCoders` class now has a method `sdkHttpMetadata()` which creates a coder that by default does not include the response headers, and if you want to include the headers there is also a method `sdkHttpMetadata(boolean includeHeaders)`. I'm on the fence about getting rid of the boolean parameter and just including a method like `sdkHttpMetadataIncludingHeaders()` so let me know if you have a preference.

- I changed the names of the SnsIO methods to `withFullPublishResultNoHeaders` and `withFullPublishResultIncludingHeaders`. The default is still to encode only the messageId for backward compatibility. I also added a method to specify your own coder for future extensibility.

I hope this PR is ready to merge now, please let me know if it needs any additional changes.

lukecwik:
Thanks for returning back to this. Taking a look now.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/coders/AwsCoders.java
```diff
@@ -0,0 +1,156 @@
...
+
+    @Override
+    public List<? extends Coder<?>> getCoderArguments() {
+      return ImmutableList.of();
+    }
+
+    @Override
+    public void verifyDeterministic() throws NonDeterministicException {
+      STATUS_CODE_CODER.verifyDeterministic();
+      HEADERS_ENCODER.verifyDeterministic();
```
lukecwik:
```suggestion
      if (includeHeaders) {
        HEADERS_ENCODER.verifyDeterministic();
      }
```

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/coders/AwsCoders.java
```diff
@@ -0,0 +1,156 @@
...
+      return ImmutableList.of();
+    }
+
+    @Override
+    public void verifyDeterministic() throws NonDeterministicException {
+      METADATA_ENCODER.verifyDeterministic();
+    }
+  }
+
+  private static class SdkHttpMetadataCoder extends Coder<SdkHttpMetadata> implements Serializable {
```
lukecwik: Please extend CustomCoder and remove the getCoderArguments method below.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/sns/SnsIO.java
```diff
@@ -257,12 +269,37 @@ public Write withResultOutputTag(TupleTag<PublishResult> results) {
...
+     */
+    public Write withFullPublishResultNoHeaders() {
+      return builder().setCoder(PublishResultCoders.fullPublishResultNoHeaders()).build();
+    }
+
+    /**
+     * Encode the full {@code PublishResult} object, including sdkResponseMetadata and
+     * sdkHttpMetadata with the HTTP response headers.
+     */
+    public Write withFullPublishResultIncludingHeaders() {
```
lukecwik:
```suggestion
    public Write withFullPublishResult() {
```

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/sns/SnsIO.java
```diff
@@ -257,12 +269,37 @@ public Write withResultOutputTag(TupleTag<PublishResult> results) {
       return builder().setResultOutputTag(results).build();
     }
 
+    /**
+     * Encode the full {@code PublishResult} object, including sdkResponseMetadata and
+     * sdkHttpMetadata but excluding the HTTP response headers.
+     */
+    public Write withFullPublishResultNoHeaders() {
```
lukecwik:
```suggestion
    public Write withFullPublishResultMinusHeaders() {
```

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/sns/SnsIO.java
```diff
@@ -257,12 +269,37 @@ public Write withResultOutputTag(TupleTag<PublishResult> results) {
...
+    /**
+     * Encode the full {@code PublishResult} object, including sdkResponseMetadata and
+     * sdkHttpMetadata with the HTTP response headers.
+     */
+    public Write withFullPublishResultIncludingHeaders() {
+      return builder().setCoder(PublishResultCoders.fullPublishResultIncludingHeaders()).build();
+    }
+
+    /** Encode the {@code PublishResult} with the given coder. */
+    public Write withPublishResultCoder(Coder<PublishResult> coder) {
```
lukecwik:
```suggestion
    public Write withCoder(Coder<PublishResult> coder) {
```

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/sns/SnsIO.java
```diff
@@ -257,12 +269,37 @@ public Write withResultOutputTag(TupleTag<PublishResult> results) {
...
+    public Write withFullPublishResultNoHeaders() {
+      return builder().setCoder(PublishResultCoders.fullPublishResultNoHeaders()).build();
+    }
+
+    /**
+     * Encode the full {@code PublishResult} object, including sdkResponseMetadata and
+     * sdkHttpMetadata with the HTTP response headers.
+     */
+    public Write withFullPublishResultIncludingHeaders() {
+      return builder().setCoder(PublishResultCoders.fullPublishResultIncludingHeaders()).build();
```
lukecwik:
```suggestion
      return setCoder(PublishResultCoders.fullPublishResultIncludingHeaders());
```

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/sns/SnsIO.java
```diff
@@ -257,12 +269,37 @@ public Write withResultOutputTag(TupleTag<PublishResult> results) {
       return builder().setResultOutputTag(results).build();
     }
 
+    /**
+     * Encode the full {@code PublishResult} object, including sdkResponseMetadata and
+     * sdkHttpMetadata but excluding the HTTP response headers.
+     */
+    public Write withFullPublishResultNoHeaders() {
+      return builder().setCoder(PublishResultCoders.fullPublishResultNoHeaders()).build();
```
lukecwik:
```suggestion
      return setCoder(PublishResultCoders.fullPublishResultNoHeaders());
```

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/sns/SnsIO.java
```diff
@@ -257,12 +269,37 @@ public Write withResultOutputTag(TupleTag<PublishResult> results) {
...
       checkArgument(getTopicName() != null, "withTopicName() is required");
-      return input.apply(
-          ParDo.of(new SnsWriterFn(this))
-              .withOutputTags(getResultOutputTag(), TupleTagList.empty()));
+      PCollectionTuple result =
+          input.apply(
+              ParDo.of(new SnsWriterFn(this))
+                  .withOutputTags(getResultOutputTag(), TupleTagList.empty()));
+      Optional.ofNullable(getCoder())
+          .ifPresent(coder -> result.get(getResultOutputTag()).setCoder(coder));
```
lukecwik: nit: It is much easier for someone to understand the following logic:
```suggestion
      if (getCoder() != null) {
        result.get(getResultOutputTag()).setCoder(coder);
      }
```

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/sns/PublishResultCoders.java
```diff
@@ -0,0 +1,129 @@
...
+  private static class PublishResultCoder extends Coder<PublishResult> implements Serializable {
+
+    private final Coder<ResponseMetadata> responseMetadataEncoder;
+    private final Coder<SdkHttpMetadata> sdkHttpMetadataCoder;
+
+    protected PublishResultCoder(
+        Coder<ResponseMetadata> responseMetadataEncoder,
+        Coder<SdkHttpMetadata> sdkHttpMetadataCoder) {
+      this.responseMetadataEncoder =
+          Optional.ofNullable(responseMetadataEncoder).map(NullableCoder::of).orElse(null);
```
lukecwik: Why add this complexity here, just pass in a NullableCoder.of(RESPONSE_METADATA_CODER) and perform assignment here.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/coders/AwsCoders.java
```diff
@@ -0,0 +1,156 @@
...
+   *
+   * @param includeHeaders <code>true</code> to serialize the SDK response headers, otherwise <code>
+   *     false</code>
+   * @return the SdkHttpMetadata coder
+   */
+  public static Coder<SdkHttpMetadata> sdkHttpMetadata(boolean includeHeaders) {
+    return new SdkHttpMetadataCoder(includeHeaders);
+  }
+
+  private static class ResponseMetadataCoder extends Coder<ResponseMetadata>
```
lukecwik: Please extend AtomicCoder and get rid of getCoderArguments

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/sns/PublishResultCoders.java
```diff
@@ -0,0 +1,129 @@
...
+  public static Coder<PublishResult> fullPublishResultIncludingHeaders() {
+    return new PublishResultCoder(RESPONSE_METADATA_CODER, AwsCoders.sdkHttpMetadata(true));
+  }
+
+  private static class PublishResultCoder extends Coder<PublishResult> implements Serializable {
+
+    private final Coder<ResponseMetadata> responseMetadataEncoder;
+    private final Coder<SdkHttpMetadata> sdkHttpMetadataCoder;
+
+    protected PublishResultCoder(
```
lukecwik:
```suggestion
    private PublishResultCoder(
```

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/coders/AwsCoders.java
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  /**
+   * Returns a new coder for SdkHttpMetadata that by default does not serialize the response
+   * headers.
+   *
+   * @return the SdkHttpMetadata coder
+   */
+  public static Coder<SdkHttpMetadata> sdkHttpMetadata() {
+    return new SdkHttpMetadataCoder(false);
```
lukecwik:
```suggestion
  public static Coder<SdkHttpMetadata> sdkHttpMetadataWithoutHeaders() {
    return new SdkHttpMetadataCoder(true);
```

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/coders/AwsCoders.java
```diff
@@ -0,0 +1,156 @@
...
+
+  /**
+   * Returns a new coder for SdkHttpMetadata that will serialize the response headers if indicated.
+   *
+   * @param includeHeaders <code>true</code> to serialize the SDK response headers, otherwise <code>
+   *     false</code>
+   * @return the SdkHttpMetadata coder
+   */
+  public static Coder<SdkHttpMetadata> sdkHttpMetadata(boolean includeHeaders) {
+    return new SdkHttpMetadataCoder(includeHeaders);
```
lukecwik:
```suggestion
  public static Coder<SdkHttpMetadata> sdkHttpMetadata() {
    return new SdkHttpMetadataCoder(true);
```

lukecwik:
Run Java PreCommit

jfarr:
Thanks for the feedback @lukecwik. I've addressed your comments and also added one more unit test in SnsIOTest. Please let me know what you think now.

lukecwik:
retest this please

sdks/java/io/amazon-web-services/src/test/java/org/apache/beam/sdk/io/aws/sns/SnsIOTest.java
```diff
@@ -131,6 +139,52 @@ public void testRetries() throws Throwable {
...
+            .apply(MapElements.into(TypeDescriptors.strings()).via(result -> result.getMessageId()))
+            .apply(Count.globally());
+    PAssert.that(publishedResultsSize).containsInAnyOrder(ImmutableList.of(1L));
+    p.run().waitUntilFinish();
+    assertThat(mockCoder.captured).isNotNull();
+  }
+
+  // Hand-code mock because Mockito mocks cause NotSerializableException even with
+  // withSettings().serializable().
+  private static class MockCoder extends AtomicCoder<PublishResult> implements Serializable {
```
lukecwik: All Coders implement Serializable already.
```suggestion
  private static class MockCoder extends AtomicCoder<PublishResult> {
```

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/sns/PublishResultCoders.java
```diff
@@ -0,0 +1,128 @@
...
+   * but does not include the HTTP response headers.
+   *
+   * @return the PublishResult coder
+   */
+  public static Coder<PublishResult> fullPublishResultWithoutHeaders() {
+    return new PublishResultCoder(
+        RESPONSE_METADATA_CODER, NullableCoder.of(AwsCoders.sdkHttpMetadataWithoutHeaders()));
+  }
+
+  static class PublishResultCoder extends Coder<PublishResult> implements Serializable {
```
lukecwik:
```suggestion
  static class PublishResultCoder extends CustomCoder<PublishResult> {
```

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/coders/AwsCoders.java
```diff
@@ -0,0 +1,141 @@
...
+   * Returns a new coder for SdkHttpMetadata that does not serialize the response headers.
+   *
+   * @return the SdkHttpMetadata coder
+   */
+  public static Coder<SdkHttpMetadata> sdkHttpMetadataWithoutHeaders() {
+    return new SdkHttpMetadataCoder(false);
+  }
+
+  private static class ResponseMetadataCoder extends AtomicCoder<ResponseMetadata>
+      implements Serializable {
```
lukecwik:
```suggestion
  private static class ResponseMetadataCoder extends AtomicCoder<ResponseMetadata> {
```

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/coders/AwsCoders.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    @Override
+    public ResponseMetadata decode(InputStream inStream) throws CoderException, IOException {
+      return new ResponseMetadata(METADATA_ENCODER.decode(inStream));
+    }
+  }
+
+  private static class SdkHttpMetadataCoder extends CustomCoder<SdkHttpMetadata>
+      implements Serializable {
```
lukecwik:
```suggestion
  private static class SdkHttpMetadataCoder extends CustomCoder<SdkHttpMetadata> {
```

jfarr:
@lukecwik thanks for the suggestions!

jfarr:
One last question @lukecwik, would you prefer to apply these changes in amazon-web-services2 in this PR? Or should we take that to a separate PR? That might be a good opportunity to discuss making some improvements to how errors are handled in this IO. In my opinion it's somewhat suboptimal that `IOException`s and the like are handled in the IO but HTTP error responses have to be handled in the caller, for example.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/coders/AwsCoders.java
```diff
@@ -0,0 +1,138 @@
...
+    return ResponseMetadataCoder.of();
+  }
+
+  /**
+   * Returns a new coder for SdkHttpMetadata.
+   *
+   * @return the SdkHttpMetadata coder
+   */
+  public static Coder<SdkHttpMetadata> sdkHttpMetadata() {
+    return new SdkHttpMetadataCoder(true);
```
cmachgodaddy: Just curious, why this coder is not following the pattern `coder.of(xxx)` ?

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/coders/AwsCoders.java
```diff
@@ -0,0 +1,138 @@
...
+        Optional.ofNullable(HEADERS_ENCODER.decode(inStream))
+            .ifPresent(
+                headers ->
+                    headers.keySet().forEach(k -> httpResponse.addHeader(k, headers.get(k))));
+      }
+      return SdkHttpMetadata.from(httpResponse);
+    }
+
+    @Override
+    public void verifyDeterministic() throws NonDeterministicException {
```
cmachgodaddy: And one more curiosity, why we have to implement `verifyDeterministic` for this coder?

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/sns/SnsCoderProviderRegistrar.java
```diff
@@ -32,6 +32,7 @@
   @Override
   public List<CoderProvider> getCoderProviders() {
     return ImmutableList.of(
-        CoderProviders.forCoder(TypeDescriptor.of(PublishResult.class), PublishResultCoder.of()));
+        CoderProviders.forCoder(
+            TypeDescriptor.of(PublishResult.class), PublishResultCoders.defaultPublishResult()));
```
cmachgodaddy: Same here, any reasons this is not `XXXcoder.of(...)` pattern?

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/coders/AwsCoders.java
jfarr: @cmachgodaddy That's because this class is designed as a factory for coders, with the coder implementation being an internal detail, as @lukecwik suggested in an earlier comment thread.

> Style note, you can create one class called `AwsCoders` which has three static methods `minimalSdkHttpMetadata()` and `fullSdkHttpMetadata()` and `responseMetadata()` that return the coders. This way you can make all the coder implementations private inner static classes and they all become implementation details without needing to expose them to users. Users can still get an instance if they need one but it gives us flexibility from a maintenance point of view how the coders are organized.

I could create a static factory method `of()` on SdkHttpMetadata but it wouldn't serve any purpose, it would only be called by sdkHttpMetadata() which is already a static factory method. SdkHttpMetadata is not meant to be instantiated directly by the caller.

lukecwik: of() prevents moving the class while a top level static method allows for stuff to still be moved around.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/sns/SnsCoderProviderRegistrar.java
jfarr: It's sort of the same reason as above, PublishResultCoders is a factory for creating several different coder implementations so the factory methods have different names to indicate which one. I could rename this method to `of()` but since we are introducing new behavior I wanted to make it very clear that this is the default, or existing, behavior. Also I'm not sure if it makes sense to have a method called `of()` on the factory not on the coder itself.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/coders/AwsCoders.java
lukecwik: By default coders are considered non-deterministic which prevent them from being used as the key coder for many things (GBK, map/multimap side input keys, map state keys, ...).

Implementing it allows us to be deterministic if the coders we use are deterministic. So if the internally used coders become deterministic then we get that for free.

lukecwik:
> One last question @lukecwik, would you prefer to apply these changes in amazon-web-services2 in this PR? Or should we take that to a separate PR? That might be a good opportunity to discuss making some improvements to how errors are handled in this IO. In my opinion it's somewhat suboptimal that `IOException`s and the like are handled in the IO but HTTP error responses have to be handled in the caller, for example.

I would prefer a separate PR so that people can start using this sooner then later.

lukecwik:
@iemejia Any final comments?

Otherwise I intend to merge when the tests are green.

sdks/java/io/amazon-web-services/src/main/java/org/apache/beam/sdk/io/aws/coders/AwsCoders.java
jfarr: `CustomCoder`'s implementation always throws NonDeterministicException so this seemed more correct.
https://github.com/apache/beam/blob/33ec0bb052e6711efaa833d6c044680e4de30902/sdks/java/core/src/main/java/org/apache/beam/sdk/coders/CustomCoder.java#L47-L53

lukecwik:
Run Java PreCommit

lukecwik:
Run Java PreCommit

PR #9765, opened 2019-10-11
[BEAM-8382] Add rate limit policy to KinesisIO.Read

jfarr:
This PR adds an optional rate limit policy to KinesisIO.Read that can be configured with the following methods:

`withFixedDelayRateLimitPolicy()`
Specifies a fixed delay rate limit policy with the default delay of 1 second.

`withFixedDelayRateLimitPolicy(Duration delay)`
Specifies a fixed delay rate limit policy with the given delay.

`withDynamicDelayRateLimitPolicy(Supplier<Duration> delay)`
Specifies a dynamic delay rate limit policy with the given function being called at each polling interval to get the next delay value.

`withCustomRateLimitPolicy(RateLimitPolicyFactory rateLimitPolicyFactory)`
Specifies the RateLimitPolicyFactory for a custom rate limiter.

This provides a strategy interface for dealing with the Kinesis hard limit of 5 read requests per second per shard. The FixedDelayRateLimiter will apply a fixed delay between read calls. The default is a "no limiter" policy which preserves the behavior of the current implementation.
 R: @aromanenko-dev

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/KinesisIO.java
+++ b/sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/KinesisIO.java
@@ -40,6 +40,7 @@ import java.util.Properties;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.concurrent.LinkedBlockingDeque;
+import java.util.function.Supplier;
 import javax.annotation.Nullable;
 import org.apache.beam.sdk.annotations.Experimental;
 import org.apache.beam.sdk.io.Read.Unbounded;
@@ -94,7 +95,6 @@ import org.slf4j.LoggerFactory;
  *
  * <pre>{@code
  * public class MyCustomKinesisClientProvider implements AWSClientsProvider {
- *   {@literal @}Override
  *   public AmazonKinesis getKinesisClient() {
  *     // set up your client here
  *   }
@@ -148,12 +148,10 @@ import org.slf4j.LoggerFactory;
  *       this.customWatermarkPolicy = new WatermarkPolicyFactory.CustomWatermarkPolicy(WatermarkParameters.create());
  *     }
  *
- *     @Override
  *     public Instant getWatermark() {
  *       return customWatermarkPolicy.getWatermark();
  *     }
  *
- *     @Override
  *     public void update(KinesisRecord record) {
  *       customWatermarkPolicy.update(record);
  *     }
@@ -161,7 +159,6 @@ import org.slf4j.LoggerFactory;
  *
  * // custom factory
  * class MyCustomPolicyFactory implements WatermarkPolicyFactory {
- *     @Override
  *     public WatermarkPolicy createWatermarkPolicy() {
  *       return new MyCustomPolicy();
  *     }
@@ -171,6 +168,69 @@ import org.slf4j.LoggerFactory;
  *    .withStreamName("streamName")
  *    .withInitialPositionInStream(InitialPositionInStream.LATEST)
  *    .withCustomWatermarkPolicy(new MyCustomPolicyFactory())
+ * }</pre>
+ *
+ * <p>By default Kinesis IO will poll the Kinesis getRecords() API as fast as possible which may
+ * lead to excessive read throttling. To limit the rate of getRecords() calls you can set a rate
+ * limit policy. For example, the default fixed delay policy will limit the rate to one API call per
+ * second per shard:
+ *
+ * <pre>{@code
+ * p.apply(KinesisIO.read()
+ *    .withStreamName("streamName")
+ *    .withInitialPositionInStream(InitialPositionInStream.LATEST)
+ *    .withFixedDelayRateLimitPolicy())
+ * }</pre>
+ *
+ * <p>You can also use a fixed delay policy with a specified delay interval, for example:
+ *
+ * <pre>{@code
+ * p.apply(KinesisIO.read()
+ *    .withStreamName("streamName")
+ *    .withInitialPositionInStream(InitialPositionInStream.LATEST)
+ *    .withFixedDelayRateLimitPolicy(Duration.millis(500))
+ * }</pre>
+ *
+ * <p>If you need to change the polling interval of a Kinesis pipeline at runtime, for example to
+ * compensate for adding and removing additional consumers to the stream, then you can supply the
+ * delay interval as a function so that you can obtain the current delay interval from some external
+ * source:
+ *
+ * <pre>{@code
+ * p.apply(KinesisIO.read()
+ *    .withStreamName("streamName")
+ *    .withInitialPositionInStream(InitialPositionInStream.LATEST)
+ *    .withDynamicDelayRateLimitPolicy(() -> Duration.millis(<some delay interval>))
+ * }</pre>
+ *
+ * <p>Finally, you can create a custom rate limit policy that responds to successful read calls
+ * and/or read throttling exceptions with your own rate-limiting logic:
+ *
+ * <pre>{@code
+ * // custom policy
+ * public class MyCustomPolicy implements RateLimitPolicy {
+ *
+ *   public void onSuccess(List<KinesisRecord> records) throws InterruptedException {
+ *     // handle successful getRecords() call
+ *   }
+ *
+ *   public void onThrottle(KinesisClientThrottledException e) throws InterruptedException {
+ *     // handle Kinesis read throttling exception
+ *   }
+ * }
+ *
+ * // custom factory
+ * class MyCustomPolicyFactory implements RateLimitPolicyFactory {
+ *
+ *   public RateLimitPolicy getRateLimitPolicy() {
+ *     return new MyCustomPolicy();
+ *   }
+ * }
+ *
+ * p.apply(KinesisIO.read()
+ *    .withStreamName("streamName")
+ *    .withInitialPositionInStream(InitialPositionInStream.LATEST)
+ *    .withCustomRateLimitPolicy(new MyCustomPolicyFactory())
  * }</pre>
  *
  * <h3>Writing to Kinesis</h3>
@@ -430,15 +490,13 @@ public final class KinesisIO {
       return toBuilder().setWatermarkPolicyFactory(watermarkPolicyFactory).build();
     }
 
-    /**
-     * Specifies the rate limit policy as FixedDelayRateLimiter with the default delay of 1 second.
-     */
+    /** Specifies a fixed delay rate limit policy with the default delay of 1 second. */
     public Read withFixedDelayRateLimitPolicy() {
       return toBuilder().setRateLimitPolicyFactory(RateLimitPolicyFactory.withFixedDelay()).build();
     }
 
     /**
-     * Specifies the rate limit policy as FixedDelayRateLimiter with the given delay.
+     * Specifies a fixed delay rate limit policy with the given delay.
      *
      * @param delay Denotes the fixed delay duration.
      */
@@ -447,6 +505,18 @@ public final class KinesisIO {
       return toBuilder()
           .setRateLimitPolicyFactory(RateLimitPolicyFactory.withFixedDelay(delay))
           .build();
+    }
+
+    /**
+     * Specifies a dynamic delay rate limit policy with the given function being called at each
+     * polling interval to get the next delay value. This can be used to change the polling interval
+     * of a running pipeline based on some external configuration source, for example.
+     *
+     * @param delay The function to invoke to get the next delay duration.
+     */
+    public Read withDynamicDelayRateLimitPolicy(Supplier<Duration> delay) {
+      checkArgument(delay != null, "delay cannot be null");
+      return toBuilder().setRateLimitPolicyFactory(RateLimitPolicyFactory.withDelay(delay)).build();
     }
 
     /**
--- a/sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/KinesisIO.java
+++ b/sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/KinesisIO.java
@@ -239,6 +239,7 @@ public final class KinesisIO {
         .setMaxNumRecords(Long.MAX_VALUE)
         .setUpToDateThreshold(Duration.ZERO)
         .setWatermarkPolicyFactory(WatermarkPolicyFactory.withArrivalTimePolicy())
+        .setRateLimitPolicyFactory(RateLimitPolicyFactory.withoutLimiter())
         .setMaxCapacityPerShard(ShardReadersPool.DEFAULT_CAPACITY_PER_SHARD)
         .build();
   }
@@ -273,6 +274,8 @@ public final class KinesisIO {
 
     abstract WatermarkPolicyFactory getWatermarkPolicyFactory();
 
+    abstract RateLimitPolicyFactory getRateLimitPolicyFactory();
+
     abstract Integer getMaxCapacityPerShard();
 
     abstract Builder toBuilder();
@@ -295,6 +298,8 @@ public final class KinesisIO {
       abstract Builder setRequestRecordsLimit(Integer limit);
 
       abstract Builder setWatermarkPolicyFactory(WatermarkPolicyFactory watermarkPolicyFactory);
+
+      abstract Builder setRateLimitPolicyFactory(RateLimitPolicyFactory rateLimitPolicyFactory);
 
       abstract Builder setMaxCapacityPerShard(Integer maxCapacity);
 
@@ -425,6 +430,35 @@ public final class KinesisIO {
       return toBuilder().setWatermarkPolicyFactory(watermarkPolicyFactory).build();
     }
 
+    /**
+     * Specifies the rate limit policy as FixedDelayRateLimiter with the default delay of 1 second.
+     */
+    public Read withFixedDelayRateLimitPolicy() {
+      return toBuilder().setRateLimitPolicyFactory(RateLimitPolicyFactory.withFixedDelay()).build();
+    }
+
+    /**
+     * Specifies the rate limit policy as FixedDelayRateLimiter with the given delay.
+     *
+     * @param delay Denotes the fixed delay duration.
+     */
+    public Read withFixedDelayRateLimitPolicy(Duration delay) {
+      checkArgument(delay != null, "delay cannot be null");
+      return toBuilder()
+          .setRateLimitPolicyFactory(RateLimitPolicyFactory.withFixedDelay(delay))
+          .build();
+    }
+
+    /**
+     * Specifies the {@code RateLimitPolicyFactory} for a custom rate limiter.
+     *
+     * @param rateLimitPolicyFactory Custom rate limit policy factory.
+     */
+    public Read withCustomRateLimitPolicy(RateLimitPolicyFactory rateLimitPolicyFactory) {
+      checkArgument(rateLimitPolicyFactory != null, "rateLimitPolicyFactory cannot be null");
+      return toBuilder().setRateLimitPolicyFactory(rateLimitPolicyFactory).build();
+    }
+
     /** Specifies the maximum number of messages per one shard. */
     public Read withMaxCapacityPerShard(Integer maxCapacity) {
       checkArgument(maxCapacity > 0, "maxCapacity must be positive, but was: %s", maxCapacity);
@@ -441,6 +475,7 @@ public final class KinesisIO {
                   getInitialPosition(),
                   getUpToDateThreshold(),
                   getWatermarkPolicyFactory(),
+                  getRateLimitPolicyFactory(),
                   getRequestRecordsLimit(),
                   getMaxCapacityPerShard()));
 
--- a/sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/RateLimitPolicyFactory.java
+++ b/sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/RateLimitPolicyFactory.java
@@ -19,6 +19,7 @@ package org.apache.beam.sdk.io.kinesis;
 
 import java.io.Serializable;
 import java.util.List;
+import java.util.function.Supplier;
 import org.joda.time.Duration;
 
 /**
@@ -34,30 +35,34 @@ public interface RateLimitPolicyFactory extends Serializable {
   }
 
   static RateLimitPolicyFactory withFixedDelay() {
-    return FixedDelayRateLimiter::new;
+    return DelayIntervalRateLimiter::new;
   }
 
   static RateLimitPolicyFactory withFixedDelay(Duration delay) {
-    return () -> new FixedDelayRateLimiter(delay);
+    return () -> new DelayIntervalRateLimiter(() -> delay);
   }
 
-  class FixedDelayRateLimiter implements RateLimitPolicy {
+  static RateLimitPolicyFactory withDelay(Supplier<Duration> delay) {
+    return () -> new DelayIntervalRateLimiter(delay);
+  }
 
-    private static final Duration DEFAULT_DELAY = Duration.standardSeconds(1);
+  class DelayIntervalRateLimiter implements RateLimitPolicy {
 
-    private final Duration delay;
+    private static final Supplier<Duration> DEFAULT_DELAY = () -> Duration.standardSeconds(1);
 
-    public FixedDelayRateLimiter() {
+    private final Supplier<Duration> delay;
+
+    public DelayIntervalRateLimiter() {
       this(DEFAULT_DELAY);
     }
 
-    public FixedDelayRateLimiter(Duration delay) {
+    public DelayIntervalRateLimiter(Supplier<Duration> delay) {
       this.delay = delay;
     }
 
     @Override
     public void onSuccess(List<KinesisRecord> records) throws InterruptedException {
-      Thread.sleep(delay.getMillis());
+      Thread.sleep(delay.get().getMillis());
     }
   }
 }
--- a/sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/RateLimitPolicyFactory.java
+++ b/sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/RateLimitPolicyFactory.java
@@ -1,0 +1,63 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.beam.sdk.io.kinesis;
+
+import java.io.Serializable;
+import java.util.List;
+import org.joda.time.Duration;
+
+/**
+ * Implement this interface to create a {@code RateLimitPolicy}. Used to create a rate limiter for
+ * each shard.
+ */
+public interface RateLimitPolicyFactory extends Serializable {
+
+  RateLimitPolicy getRateLimitPolicy();
+
+  static RateLimitPolicyFactory withoutLimiter() {
+    return () -> new RateLimitPolicy() {};
+  }
+
+  static RateLimitPolicyFactory withFixedDelay() {
+    return FixedDelayRateLimiter::new;
+  }
+
+  static RateLimitPolicyFactory withFixedDelay(Duration delay) {
+    return () -> new FixedDelayRateLimiter(delay);
+  }
+
+  class FixedDelayRateLimiter implements RateLimitPolicy {
+
+    private static final Duration DEFAULT_DELAY = Duration.standardSeconds(1);
+
+    private final Duration delay;
+
+    public FixedDelayRateLimiter() {
+      this(DEFAULT_DELAY);
+    }
+
+    public FixedDelayRateLimiter(Duration delay) {
+      this.delay = delay;
+    }
+
+    @Override
+    public void onSuccess(List<KinesisRecord> records) throws InterruptedException {
+      Thread.sleep(delay.getMillis());
+    }
+  }
+}
--- a/sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/ShardReadersPool.java
+++ b/sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/ShardReadersPool.java
@@ -75,6 +75,7 @@ class ShardReadersPool {
 
   private final SimplifiedKinesisClient kinesis;
   private final WatermarkPolicyFactory watermarkPolicyFactory;
+  private final RateLimitPolicyFactory rateLimitPolicyFactory;
   private final KinesisReaderCheckpoint initialCheckpoint;
   private final int queueCapacityPerShard;
   private final AtomicBoolean poolOpened = new AtomicBoolean(true);
@@ -83,10 +84,12 @@ class ShardReadersPool {
       SimplifiedKinesisClient kinesis,
       KinesisReaderCheckpoint initialCheckpoint,
       WatermarkPolicyFactory watermarkPolicyFactory,
+      RateLimitPolicyFactory rateLimitPolicyFactory,
       int queueCapacityPerShard) {
     this.kinesis = kinesis;
     this.initialCheckpoint = initialCheckpoint;
     this.watermarkPolicyFactory = watermarkPolicyFactory;
+    this.rateLimitPolicyFactory = rateLimitPolicyFactory;
     this.queueCapacityPerShard = queueCapacityPerShard;
     this.executorService = Executors.newCachedThreadPool();
     this.numberOfRecordsInAQueueByShard = new ConcurrentHashMap<>();
@@ -115,11 +118,12 @@ class ShardReadersPool {
   void startReadingShards(Iterable<ShardRecordsIterator> shardRecordsIterators) {
     for (final ShardRecordsIterator recordsIterator : shardRecordsIterators) {
       numberOfRecordsInAQueueByShard.put(recordsIterator.getShardId(), new AtomicInteger());
-      executorService.submit(() -> readLoop(recordsIterator));
+      executorService.submit(
+          () -> readLoop(recordsIterator, rateLimitPolicyFactory.getRateLimitPolicy()));
     }
   }
 
-  private void readLoop(ShardRecordsIterator shardRecordsIterator) {
+  private void readLoop(ShardRecordsIterator shardRecordsIterator, RateLimitPolicy rateLimiter) {
     while (poolOpened.get()) {
       try {
         List<KinesisRecord> kinesisRecords;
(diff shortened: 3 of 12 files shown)


iemejia:
Oups my github page hadn't updated and I had not seen Alexey was already on this. I -1 myself from the review then. Thanks for the contribution @jfarr

jfarr:
> Thank you for contribution. Beam tends to minimize number of tuning knobs. So, I'm wondering if it's possible to detect such throttling behaviour and increase timeout automatically in the runtime?

Hi @aromanenko-dev, thanks for the feedback. That makes sense and I think that it would be possible but I'm concerned that it would overly complicate the implementation and the runtime behavior. Also, an algorithm that could perform well would likely only add more knobs and if those aren't exposed then they just aren't tunable. Personally I would prefer the simplicity and predictability of a simple rate limit. If preserving the default behavior unchanged isn't a concern what do you think about just setting a reasonable default (1 second is the AWS recommendation) but leaving that knob in for the rare case that users need to tune it? I doubt that the vast majority would ever need to touch it.

aromanenko-dev:
@jfarr I guess it would not complicate things too much. My idea is the following - when we fetch new records in `ShardReadersPool.readLoop()` with `shardRecordsIterator.readNextBatch()` then it will finally end up in calling `SimplifiedKinesisClient.getRecords()` where we execute `AmazonKinesis.getRecords()` and wrap exceptions, if any. So, it looks like that we can re-throw `KMSThrottlingException` and catch it in `ShardReadersPool.readLoop()`. In case of this exception we can write warning in logs about that, wait with progressive timeout and recall `shardRecordsIterator.readNextBatch()` again. In this case no additional knobs well be needed.
Wdyt?

jfarr:
@aromanenko-dev If we're the only consumer on the stream we can call getRecords every 200ms without getting throttled (in theory). If we introduce a 1 sec delay on the first KMSThrottlingException then we have an unnecessary 800ms of latency. Let's say we reduce the initial delay to 100ms. Assuming a successful call takes about 10ms, we'll hit an exception after 50ms and for the next 950ms thereafter. On the 4th retry we'll succeed but we'll be at 800ms. Reduce it to 10ms and on the 7th retry we'll succeed but we'll be at 640ms. In any case we've introduced another knob (initial backoff delay time).

So that's assuming delay time starts at zero and only increases when we get a KMSThrottlingException. Since that's 100% guaranteed to happen we could try with a non-zero initial delay instead. Maybe if we start at 200ms we won't get throttled at all and we won't overshoot. Another knob.

OK so now we've overshot. Maybe we can ease back on the delay. We can speculatively try a shorter delay time and see if we still get throttled. How often to try? How much to ease back? More knobs.

You can fiddle with these knobs and come up with something that works well when you're the only consumer, but as soon as you have 2 or more consumers you have to throw that out the window. If you're pulling back more or less records that may take more or less time and throw off your timing. You can't really hardcode anything because what works well in one scenario may not work well in another.

So those are my initial thoughts. I'm totally open to the idea that I'm just overthinking it. I think the ultimate test would be to try it out and see what works. But if you want to minimize latency (and we do) I think the 1 second rolling window introduces a feedback delay that makes this a little more complicated than first glance. What do you think?

aromanenko-dev:
@jfarr I agree that there is no perfect solution for that. In the same time, I don't see too much difference between hardcoded value and configured one taking into account that Kinesis is unbounded source and user pipeline has to run for a long time without interruption. So, what's happened if we set initial polling interval too small or too big? Would we need to stop a pipeline, set new value and restart it again?
 
Unfortunately, there is no guarantee that this value will be efficient and work fine all the way since, as you said above, other client can start consume from the same shard in parallel. So, in this case it would better to provide adaptive solution. Even if we should expose some knobs to user, so what's about using `FluentBackoff` [1] instead? It can be configured with retry configuration object, like we do in `SolrIO` and some other IOs with `RetryConfiguration`. 

Other option could be to allow to provide UDF object which will manage polling delay dynamically.

Wdyt?

[1] https://github.com/apache/beam/blob/master/sdks/java/core/src/main/java/org/apache/beam/sdk/util/FluentBackoff.java

jfarr:
@aromanenko-dev To me FluentBackoff seems like the wrong way to deal with rate limiting. I would rather account for the rate limit up front than try to react to it as if it were an error. I think the end result is going to be once every second you'll get a few successful getRecords calls off then spend the rest of that second being throttled. Possibly longer because as the AWS documentation states, "Too many of these exceptions can result in exponential back-offs and thereby cause significant unexpected latencies in processing." They also state that they recommend a 1 second polling interval because, "processing larger batches of data tends to be more efficient at reducing network and other downstream latencies in your application," so polling as fast as possible then getting throttled will not be as efficient according to AWS.

You have a good point about changing the polling interval while the pipeline is running. A UDF sounds like it could be a good solution for that. We do not have that use case though. We know ahead of time how many consumers our stream will have and what their polling intervals will be. That would not change unless our architecture changes.

If it's reasonable to allow configuring a backoff strategy why would it not be reasonable to allow configuring a polling interval? I thought your goal was to not introduce any new knobs. If that's the case honestly I would rather have a hard-coded 1 second polling interval.

aromanenko-dev:
@jfarr Regarding backoff strategy  - I think it's a good way in case if your backend doesn't provide any statistics in advance at which rate you have to perform next queries. We know only maximum limits and the result of last `kinesis.getRecords()` execution. So, starting with recommended default value and use proper backoff strategy in case of failures it could be optimal solution when you need to limit rate of requests.  Otherwise, there are big chances that backend resources won't be utilised efficiently. `FluentBackoff` is used in many Beam IOs for similar reasons.

I'd like to ask you questions about your use case. Let's suppose we can have ability set static polling interval with `withPollingInterval(Duration)` (as implemented in current PR). What would be default value in your case? 1 sec? What would you do in case if it won't be enough for already running pipeline? Would your need to change this value and restart pipeline manually?

Btw, default delay of 1 sec looks quite pessimistic since according to AWS doc `each shard can support up to five read transactions per second.` So, in case of one consumer per shard (I can guess this is a quite common case) it should be 200 ms as default value. 

PS: The goal of minimizing the number of tuning knobs is to reduce the number of possible config combinations (which grows exponentially) in case if it's possible to configure this automatically [1]. If it's not possible then the better solution will be to provide flexible API, especially for unbounded sources (as Kinesis, Kafka, etc), that are being used in long-running pipelines.  

[1] https://beam.apache.org/contribute/ptransform-style-guide/#configuration

jfarr:
@aromanenko-dev thanks for the link, that's really helpful. For our use case we'll probably poll once per second. That's what the SDK docs recommend, Kinesis Firehose polls on a fixed 1 second interval, Lambda triggers poll on a fixed 1 second interval, the KCL has a configurable polling interval but it defaults to 1 second, etc., so that seems pretty standard across the board. We currently have 2 Firehoses on this stream polling once per second so in theory we could go as low as 350ms or so but as you say with a static configuration we would have to stop the running pipeline if we needed to change this later. This would allow us to add one maybe two consumers before we needed to deal with that. I know that my colleague @cmachgodaddy and possibly others are working on a v2 KinesisIO that uses subscribeToShard instead of polling getRecords so hopefully we can switch to using that instead before then.

To give you some more context, we are trying to read records from Kinesis, extract a few fields, then send the record to SNS with those fields as message attributes. So far we can't even get `read from Kinesis -> write to SNS` to work at our scale which isn't particularly large, only thousands of records per second. We are seeing latencies grow to 15-20 mins or more under peak load. To be honest at this point I can't say for sure whether it's KinesisIO or SnsIO or some combination, but I can say that adding the polling interval solved our ReadProvisionedThroughputExceeded problem but hasn't completely solved the latency problem. What I would like to do next week is build a test pipeline that just reads from Kinesis, calculates the latency at the moment we receive the record, then pushes that metric to CloudWatch. Then I'd like to test the current implementation vs adding FluentBackoff vs adding a polling interval (my PR) vs adding both. That will allow us to isolate KinesisIO and get some hard data that we can compare. I'll let you know what I find ASAP. How does that sound?

cmachgodaddy:
@aromanenko-dev , Alex correct me if I am wrong. I am trying to understand the existing KinesisIO to migrate it, and I found in each split it will generate a bunch of threads, and each thread will pull records from a shard (so 1:1 , 1 thread one shard). But, the point is if we have 10 splits (or parallelisms), then there will be 10 x # threads reading a shard. In other words, there will be 10 clients reading one shard concurrently. That's why we are experiencing this throttling. I am wondering why couldn't it be designed in a way that each split read one shard (1:1) ?

aromanenko-dev:
@jfarr Sounds great for me! Thank you for your contribution on this! Actually, I think it's very interesting and important topic for KinesisIO and we have a field for improvement there.

aromanenko-dev:
@cmachgodaddy Hmm, what do you mean by "split"? One split is one shard, and as you properly said before, only one thread from thread pool will consume messages from this shard. Did I miss something?

cmachgodaddy:
@aromanenko-dev , a bit technical, in KinesisSource, we have spit method, this method generate a number of sources or readers (or KinesisReader) ? This split method is now based on either the desiredNumSplits (which is configured by user as a parallelism parameter) or partitions (number of shards) to generate num of readers. However, in each reader (KinesisReader), we again generate another number of threads and each read one shard. So this looks to me a parallelism inside a parallelism (or nested parallelism). Here is in the KinesisReader, which call ShareReadersPool::start(), and which uses ExecutorService to execute 'readLoop' in parallel (in diff thread) https://github.com/apache/beam/blob/master/sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/ShardReadersPool.java#L124. Let me know if I am missing something?

aromanenko-dev:
@cmachgodaddy My understanding is the following. `KinesisSource.split()` method is called on driver, before running a pipeline. So, it will create a list of `KinesisSource`s for every split/partition (based on `initialCheckpointGenerator`) that will be distributed on workers and run there. Then every `KinesisSource` will create `KinesisReader` with own `CheckpointGenerator` (on worker) and instantiate `ShardReadersPool` which will create thread for every shard iterator (since every reader can read from more than one shard). So, I don't see nested parallelism here. Though, it would be great to test it in case if we have a bug there since this logic is quite complicated.

cmachgodaddy:
@alexvanboxel , I think we all understand it right, from what you described above we just say it in different ways ;-) . In simple words, each split, run on each worker, will create a number of threads base off num of shards. That's means if we have 10 splits, we will have 10 guys/reader read one shard (of course, it will read all shards)? Hope you agree with me this point? Now, what Amazon recommend is we should have one client read one shard. Here is what they say: 

> Typically, when you use the KCL, you should ensure that the number of instances does not exceed the number of shards (except for failure standby purposes). Each shard is processed by exactly one KCL worker and has exactly one corresponding record processor, so you never need multiple instances to process one shard. However, one worker can process any number of shards, so it's fine if the number of shards exceeds the number of instances.
And, here is why they do that, https://docs.aws.amazon.com/streams/latest/dev/kinesis-low-latency.html

But, don't argue that, here they use KCL, so pls consider one KCL is one KinesisClient that our split use to connect to Kinesis and read a shard. And don't misunderstand the `one worker can read a number of shard`. Here in their context, or the example in their doc, their worker is an application that run on EC2 instance, and when that scale this application, it will loadbalance the KCL, e.g. two EC2 will have two KCL, each read two shards (assume a stream has 4 shards). 

The point is we don't want to have number readers greater number of shards in one application (or one pipeline in our context)? Imagine, if we have 10 pipelines deployed in our runners, and which has parallelism of 10, then we will have 10 x 10 readers reading one shard ? And this 10 x 10 readers are not loadbalanced since we are not using KCL? (even with KCL, this number still exceed num of shard?).  I am sure we will get "throughput exception" in the log :-)

alexvanboxel:
> @alexvanboxel , I think we all understand it right, from what you described above we just say it in different ways ;-) . In simple words, each split, run on each worker, will create a number of threads 

I think you're talking about another Alex, probably @aromanenko-dev :-)

jfarr:
@cmachgodaddy I’ll have to confirm but I’m pretty sure each ShardReadersPool doesn’t read every shard just the ones assigned to that worker. That’s pretty much what I’ve observed.

cmachgodaddy:
@jfarr , uhmmm , what happens when you set parallelism of 1? Do you see one worker just pull one shard (which assigned to it) ? and the rest of the shards just seating idle? does it mean we just read 1/3 of data stream, assuming you have 3 shards?

jfarr:
> @jfarr , uhmmm , what happens when you set parallelism of 1? Do you see one worker just pull one shard (which assigned to it) ? and the rest of the shards just seating idle? does it mean we just read 1/3 of data stream, assuming you have 3 shards?

@cmachgodaddy I believe if you have one worker and three shards then that worker will be assigned all three shards. Also if you have one shard and three workers then you will have one worker reading the shard and two workers sitting idle. That I can confirm, I have seen it with my own eyes.

aromanenko-dev:
@cmachgodaddy I think @jfarr is right - in simple case all shards are discovered by `DynamicCheckpointGenerator` and put into the list of `KinesisSource`s during `split()` executing on driver. Every `KinesisSource` may contain more than 1 shard (see `KinesisReaderCheckpoint`) in case if number of workers is less the total number of shards but it's not possible that 2 different sources had reference to the same shard. Then, all these split sources will be distributed over workers by your resource manager. So, every Source will be responsible for one or more different shards and it will spawn one thread per shard in `ShardReadersPool`.

cmachgodaddy:
Thanks @jfarr and @aromanenko-dev for sharing the info! That make sense!

jfarr:
@aromanenko-dev I'm still testing and gathering data but one of my colleagues had an idea I wanted to run by you. What if we add a polling rate policy object and have fixed and fluent versions with fluent as the default?

aromanenko-dev:
@jfarr Sounds good to me (it's similar to what I proposed by adding UDF object before).

jfarr:
@aromanenko-dev Great, I will start working on that too then. Thanks!

jfarr:
Hi @aromanenko-dev, I replaced the polling interval with a rate limit policy configured with the following methods:

`withBackoffRateLimitPolicy(FluentBackoff fluentBackoff)`
Specifies the rate limit policy as BackoffRateLimiter.

`withFixedDelayRateLimitPolicy()`
Specifies the rate limit policy as FixedDelayRateLimiter with the default delay of 1 second.

`withFixedDelayRateLimitPolicy(Duration delay)`
Specifies the rate limit policy as FixedDelayRateLimiter with the given delay.

`withCustomRateLimitPolicy(RateLimitPolicyFactory rateLimitPolicyFactory)`
Specifies the RateLimitPolicyFactory for a custom rate limiter.

The default is a no limiter policy that preserves the behavior of the current implementation.

wdyt?

sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/RateLimitPolicyFactory.java
```diff
@@ -0,0 +1,105 @@
...
+
+  class BackoffRateLimiter implements RateLimitPolicy {
+
+    private final BackOff backoff;
+
+    public BackoffRateLimiter(FluentBackoff fluentBackoff) {
+      this.backoff =
+          fluentBackoff
+              // never stop retrying
+              .withMaxRetries(Integer.MAX_VALUE)
```
aromanenko-dev: What is a reason to override user's parameters and never stop backoff process? I guess that initial idea was to allow user to do that by providing custom `FluentBackoff` object.

sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/KinesisIO.java
```diff
@@ -420,6 +426,47 @@ public Read withCustomWatermarkPolicy(WatermarkPolicyFactory watermarkPolicyFact
       return toBuilder().setWatermarkPolicyFactory(watermarkPolicyFactory).build();
     }
 
+    /**
+     * Specifies the rate limit policy as BackoffRateLimiter.
+     *
+     * @param fluentBackoff The {@code FluentBackoff} used to create the backoff policy.
+     */
+    public Read withBackoffRateLimitPolicy(FluentBackoff fluentBackoff) {
```
aromanenko-dev: Please, add an example into KinesisIO javadoc how to use it.

sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/KinesisIO.java
```diff
@@ -420,6 +426,47 @@ public Read withCustomWatermarkPolicy(WatermarkPolicyFactory watermarkPolicyFact
...
+    public Read withFixedDelayRateLimitPolicy() {
+      return toBuilder().setRateLimitPolicyFactory(RateLimitPolicyFactory.withFixedDelay()).build();
+    }
+
+    /**
+     * Specifies the rate limit policy as FixedDelayRateLimiter with the given delay.
+     *
+     * @param delay Denotes the fixed delay duration.
+     */
+    public Read withFixedDelayRateLimitPolicy(Duration delay) {
```
aromanenko-dev: Please, add an example into KinesisIO javadoc how to use it.

sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/KinesisIO.java
```diff
@@ -420,6 +426,47 @@ public Read withCustomWatermarkPolicy(WatermarkPolicyFactory watermarkPolicyFact
...
+          .setRateLimitPolicyFactory(RateLimitPolicyFactory.withFixedDelay(delay))
+          .build();
+    }
+
+    /**
+     * Specifies the {@code RateLimitPolicyFactory} for a custom rate limiter.
+     *
+     * @param rateLimitPolicyFactory Custom rate limit policy factory.
+     */
+    public Read withCustomRateLimitPolicy(RateLimitPolicyFactory rateLimitPolicyFactory) {
```
aromanenko-dev: Please, add an example into KinesisIO javadoc how to use it.

sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/RateLimitPolicyFactory.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.List;
+import org.apache.beam.sdk.util.BackOff;
+import org.apache.beam.sdk.util.FluentBackoff;
+import org.joda.time.Duration;
+
+/**
+ * Implement this interface to create a {@code RateLimitPolicy}. Used to create a rate limiter for
+ * each shard.
+ */
+public interface RateLimitPolicyFactory extends Serializable {
```
aromanenko-dev: Please, add tests for different types of factory.

sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/KinesisIO.java
lukecwik: org.apache.beam.sdk.util.FluentBackoff is not meant to be on the public API surface and is only meant for internal use. All classes within org.apache.beam.sdk.util are like that.

If we would want to promote it to be public, we should find the appropriate place for it and move it out of util.

aromanenko-dev: What's about wrapping it into KinesisIO specific backoff wrapper? 
PS: it's not evident that this is for internal use only. Do we need to annotate it?

sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/RateLimitPolicyFactory.java
```diff
@@ -0,0 +1,105 @@
...
+    public FixedDelayRateLimiter() {
+      this(DEFAULT_DELAY);
+    }
+
+    public FixedDelayRateLimiter(Duration delay) {
+      this.delay = delay;
+    }
+
+    @Override
+    public void onSuccess(List<KinesisRecord> records) throws InterruptedException {
```
cmachgodaddy: records param is not being used?

sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/RateLimitPolicyFactory.java
```diff
@@ -0,0 +1,105 @@
...
+        if (backOffMillis != BackOff.STOP) {
+          Thread.sleep(backOffMillis);
+        }
+      } catch (IOException e) {
+        // do nothing
+      }
+    }
+
+    @Override
+    public void onSuccess(List<KinesisRecord> records) {
```
cmachgodaddy: This onSuccess is called from ShardReadersPool and pass in list of records, but I don't see any implementations of the interface use this param?

sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/ShardReadersPool.java
```diff
@@ -149,10 +157,20 @@ private void readLoop(ShardRecordsIterator shardRecordsIterator) {
           recordsQueue.put(kinesisRecord);
           numberOfRecordsInAQueueByShard.get(kinesisRecord.getShardId()).incrementAndGet();
         }
+        rateLimiter.onSuccess(kinesisRecords);
+      } catch (KinesisClientThrottledException e) {
+        try {
+          rateLimiter.onThrottle(e);
+        } catch (InterruptedException ex) {
+          LOG.warn("Thread was interrupted, finishing the read loop", ex);
+          Thread.currentThread().interrupt();
```
cmachgodaddy: uhmmm, we catch an interrupted exception, and then throw out another one. Any reasons why we want to interrupt this thread?

sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/ShardReadersPool.java
```diff
@@ -149,10 +157,20 @@ private void readLoop(ShardRecordsIterator shardRecordsIterator) {
...
+          break;
+        }
       } catch (TransientKinesisException e) {
         LOG.warn("Transient exception occurred.", e);
       } catch (InterruptedException e) {
         LOG.warn("Thread was interrupted, finishing the read loop", e);
+        Thread.currentThread().interrupt();
```
cmachgodaddy: same here

sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/KinesisIO.java
aromanenko-dev: @lukecwik In this case, I suggest to change a signature of this method to `Read withBackoffRateLimitPolicy(int maxRetries, Duration maxCumulativeBackoff)` and insatiate `FluentBackoff` internally with provided params. 
PS: Updated `sdk/util package-info.java` doc,  #10009

aromanenko-dev:
@cmachgodaddy 
1)Good point, but we use `AmazonKinesis` as a client for Kinesis. Can we leverage `RetryPolicy` in this case?
2) I believe the last one should win but it would make sense to add more checks to avoid an ambiguity.

cmachgodaddy:
> @cmachgodaddy
> 
> 1. Good point, but we use `AmazonKinesis` as a client for Kinesis. Can we leverage `RetryPolicy` in this case?
> 2. I believe the last one should win but it would make sense to add more checks to avoid an ambiguity.

--> #1 Yes, every AWS client (Dynamo, Sns, Sqs, Kinesis, ...) can take in ClientConfiguration as a argument. And we can set any configurations with this object.
--> #2 Of course, we can add a ton of checkers, but users will be confuse of which one of withRateLimitXXXs to use. I would rather have just one withRateLimitXXX and let user pass in an Enum.

sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/ShardReadersPool.java
```diff
@@ -51,15 +51,13 @@
 
   /**
    * Executor service for running the threads that read records from shards handled by this pool.
-   * Each thread runs the {@link ShardReadersPool#readLoop(ShardRecordsIterator)} method and handles
-   * exactly one shard.
+   * Each thread runs the {@link ShardReadersPool#readLoop} method and handles exactly one shard.
```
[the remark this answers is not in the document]
jfarr: Sure, I'll change that.

sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/KinesisIO.java
jfarr: I was considering exposing a wrapper object like SnsIO. I didn't realize FluentBackoff was internal. I think that makes a lot of sense here too then.

What does it mean to give up retrying with an unbounded source though? What should we do in that case? Throw an unchecked exception and let the pipeline crash? That's why I explicitly overrode max retries and max cumulative backoff because I didn't think that made sense but I can change that if you want.

sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/KinesisIO.java
jfarr: > 1. Have we try to use Kinesis's RetryPolicy and BackoffStrategy? This look like we reengineer this feature, which already supported by Kinesis's API. Here is the example: https://www.programcreek.com/java-api-examples/?api=com.amazonaws.retry.RetryPolicy. If we already tried it, then what would be the reasons not to use the Kinesis's one ?

I'm not sure if RetryPolicy and BackoffStrategy apply to LimitExceededException / ProvisionedThroughputExceededException but I can look into that. If so I think it makes sense to just configure this in the Kinesis client instead of having a BackoffRateLimitPolicy. What do you think @aromanenko-dev and @lukecwik ?

jfarr:
> 1. Have we try to use Kinesis's RetryPolicy and BackoffStrategy? This look like we reengineer this feature, which already supported by Kinesis's API. Here is the example: https://www.programcreek.com/java-api-examples/?api=com.amazonaws.retry.RetryPolicy. If we already tried it, then what would be the reasons not to use the Kinesis's one ?

I'm not sure if RetryPolicy and BackoffStrategy apply to LimitExceededException / ProvisionedThroughputExceededException but I can look into that. If so I think it makes sense to just configure this in the Kinesis client instead of having a BackoffRateLimitPolicy. What do you think @aromanenko-dev and @lukecwik ?

jfarr:
> 2\. From user perspective, what if users set all 3 or 4 withRateLimitXXX ? which one will override what? or we apply all of them? Would it be helpful if we provide an enum?

If you call one of these methods again it will override the previous call.

sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/RateLimitPolicyFactory.java
jfarr: It's hard to anticipate what state a custom rate limit policy will want to keep but since we have the list of records and it might want them I just passed them in here. Same with passing the KinesisClientThrottledException in onThrottled(). Currently, no, there are no implementations using these.

sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/ShardReadersPool.java
jfarr: Thread.interrupt() does not interrupt the current thread, it sets the thread's interrupt status. The short answer is that catching InterruptedException clears the interrupt status so in order to handle it properly you should always either rethrow the exception or reset the thread's interrupt status so any code higher up the call stack can still detect the interruption. There is a good article by Brian Goetz that goes into the details: https://www.ibm.com/developerworks/java/library/j-jtp05236/index.html.

aromanenko-dev:
@cmachgodaddy 
>  #2 Of course, we can add a ton of checkers, but users will be confuse of which one of withRateLimitXXXs to use. I would rather have just one withRateLimitXXX and let user pass in an Enum.
We already have the similar API for watermarks factory in KinesisIO, so I think it will make sense to keep the same style here. And I don't think users will be very confused if we have clear Java doc for each method and examples of usage in the class Java doc. New checks are actually not needed, they could just prevent users from potential mistakes.

@jfarr I'm ok with that if it won't fail a pipeline in case of throttling and work in the similar way as FluentBackoff.

aromanenko-dev:
Btw, I merged another KinesisIO PR, so this one needs to be rebased.

sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/KinesisIO.java
aromanenko-dev: @jfarr I'd leave this responsibility to user and that's why I proposed to expose these parameters.

aromanenko-dev:
I'd like to come back to idea of using `RetryPolicy` and `BackoffStrategy` classes from AWS SDK proposed by @cmachgodaddy above. We recently had a similar issue with failing a pipeline because of `TransientKinesisException ` caused by reading 2 pipelines from the same shard. It was solved by creating a new `AWSClientsProvider`, based on current `BasicKinesisProvider`, used by default, and just overriding a method `getKinesisClient()` where we configured our custom `RetryPolicy` and `BackoffStrategy`. 

The simplified example could look like this:
```
  @Override
  public AmazonKinesis getKinesisClient() {
    AmazonKinesisClientBuilder clientBuilder = AmazonKinesisClientBuilder.standard()
        .withCredentials(getCredentialsProvider());
    clientBuilder.withRegion(region);
    
    RetryPolicy.BackoffStrategy backoffStrategy = 
        new PredefinedBackoffStrategies.ExponentialBackoffStrategy(baseDelay, maxBackoffTime);
    RetryPolicy retryPolicy = new RetryPolicy(DEFAULT_RETRY_CONDITION, backoffStrategy,
        DEFAULT_MAX_ERROR_RETRY, true);

    ClientConfiguration clientConfiguration = new ClientConfiguration();
    clientBuilder.withClientConfiguration(clientConfiguration.withRetryPolicy(retryPolicy));

    return clientBuilder.build();
  }
```
So, instead of adding new API, which will do similar things that already implemented in AWS SDK, I think it would be enough to just update KinesisIO javadoc and add an example showing how to configure and leverage custom `AWSClientsProvider` in such case.
WDYT?

PS: @jfarr I'm sorry to ask you initially to add this functionality in KinesisIO using Beam SDK, I was not aware about such configuration option in `AmazonKinesisClient`.

jfarr:
Sure @aromanenko-dev, let me try that out when I get some time and I can document it in the javadoc as you suggest and also remove the BackoffRateLimiter. This should also apply to the other AWS IOs like SNS and DynamoDB, would it make sense to remove the FluentBackoff-based implementation there too in v2 for consistency?

cmachgodaddy:
All of our AWS APIs should have this. But out API just need to call `getKinesisClient()`, which means we delegate it to users, and it's up to them to how build the client.

aromanenko-dev:
Yes, I think we should be consistent in terms of user API. So, since most (all?) of our AWS-based IOs provide a way to build client object, which can be configured to use custom retry/backoff policy, we may, at least, want to document it properly. 

In the same time, some of the AWS IOs (like DynamoDBIO, SnsIO) already provide a method `withRetryConfiguration(RetryConfiguration)` to define retry configuration. Under the hood, it actually uses Beam's `BackOff` and own implementation of retry cycle. So, I think we need to change it to use internal AWS client implementation (as we discussed before). 

In the end, it would be great to have unified API for all AWS-based IOs, at least for SDK V2.

cmachgodaddy:
They are all in my TODOs @aromanenko-dev . When I finish the other PRs, I'll do it :-)

jfarr:
Hi @aromanenko-dev, I finally got around to testing this AWS SDK RetryPolicy-based approach but with very little success. First of all, the SDK client by default uses the following retry policy for all throttling exceptions:

- backoff strategy = PredefinedBackoffStrategies.EqualJitterBackoffStrategy
- base delay = 500ms
- max backoff = 20s

Where EqualJitterBackoffStrategy is an exponential backoff with jitter. This strategy will ensure that the jitter will never be more than 1/2 the max delay for that retry attempt. So the initial delay will always be between 250-500ms.

I added some logging and I can see the retry policy behaving as expected. For example:

```
2020-02-09T21:48:00.413+00:00 com.godaddy..KinesisClientsProvider shouldRetry exception `com.amazonaws.services.kinesis.model.ProvisionedThroughputExceededException: Rate exceeded for shard shardId-000000000000 in stream beam-test ...`: true
2020-02-09T21:48:00.413+00:00 com.godaddy..KinesisClientsProvider delayBeforeNextRetry: 333 retriesAttempted: 0
2020-02-09T21:48:00.413+00:00 com.amazonaws.http.AmazonHttpClient Retriable error detected, will retry in 333ms, attempt number: 0
```

Indeed the initial backoff always seems to be within 250 to 500ms. The exponential backoff is pretty irrelevant, though, because the initial delay also seems sufficient that the first retry attempt always succeeds. I didn't see any instances of a second retry attempt in my testing.

However, the cumulative effect of all of these initial instances of ProvisionedThroughputExceededException is that approximately 30% of all getRecords() calls fail. You can see that in the attached charts where GetRecords.Success is 0.7 and ReadProvisionedThroughputExceeded is 0.3. I don't think it matters whether the retry policy is implemented in Beam or in the AWS SDK, I don't think it can be tuned to perform much better than this.

Screen Shot 2020-02-09 at 1 10 18 PM
Screen Shot 2020-02-09 at 1 10 46 PM


Here is the code I used just for reference:
```
    RetryCondition retryCondition = new RetryCondition() {
      RetryCondition defaultCondition = PredefinedRetryPolicies.DEFAULT_RETRY_CONDITION;
      @Override
      public boolean shouldRetry(AmazonWebServiceRequest originalRequest, AmazonClientException exception, int retriesAttempted) {
        boolean retry = defaultCondition.shouldRetry(originalRequest, exception, retriesAttempted);
        LOG.info("shouldRetry exception `{}`: {}", exception, retry);
        return retry;
      }
    };
    BackoffStrategy backoffStrategy = new BackoffStrategy() {
      BackoffStrategy defaultStrategy = PredefinedRetryPolicies.DEFAULT_BACKOFF_STRATEGY;
      @Override
      public long delayBeforeNextRetry(AmazonWebServiceRequest originalRequest, AmazonClientException exception, int retriesAttempted) {
        long delay = defaultStrategy.delayBeforeNextRetry(originalRequest, exception, retriesAttempted);
        LOG.info("delayBeforeNextRetry: {} retriesAttempted: {}", delay, retriesAttempted);
        return delay;
      }
    };
    RetryPolicy retryPolicy = new RetryPolicy(retryCondition, backoffStrategy,
        PredefinedRetryPolicies.DEFAULT_MAX_ERROR_RETRY, false);

    return AmazonKinesisClientBuilder.standard()
        .withClientConfiguration(PredefinedClientConfigurations.defaultConfig()
            .withRetryPolicy(retryPolicy))
        .build();
```

aromanenko-dev:
@jfarr Thank you very much for detailed testing and explanation. In the end, what do you think would be the best solution in this case?

jfarr:
Hi, @aromanenko-dev I have already removed the BackoffRateLimiter from this PR, and I would like to finish the unit tests and documentation for FixedDelayRateLimiter and narrow the scope of the PR back to that since it's a variant of what we're currently using in production. Is there any chance we can get that in? It would be good to know before I put in the remaining work. I think the custom rate limit policy is a nice to have so I can leave that in if you want in case anyone wants to implement their own adaptive rate limit policy in the future.

jfarr:
And actually, we could support the use case of changing the polling interval at runtime by having a version of FixedDelayRateLimiter that takes a function for the delay.

iemejia:
@jfarr just for info Alexey is OOO this week, so don't be surprised if the answer takes a bit of time.

jfarr:
No problem, @iemejia, thanks for the heads up.

aromanenko-dev:
@jfarr Thank you very much for update and keep working on this! I'll take a look on this today and get back to you after.

aromanenko-dev:
@jfarr It looks fine but, as you mentioned above, it misses some tests and documentation to be merged.

aromanenko-dev:
@jfarr Btw, did you do the similar tests using these changes with fixed delay as you did for `PredefinedBackoffStrategies.EqualJitterBackoffStrategy` above? Sorry, if I missed that.

jfarr:
Hi @aromanenko-dev, sure we have tested it. Here's what read throttles look like with a 1s fixed delay:
Screen Shot 2020-03-01 at 2 46 05 PM

jfarr:
@aromanenko-dev if I wanted to make FixedDelayRateLimiter take a function as an argument, can I just declare it as `Supplier<Integer>` or should I declare it some other way due to Beam's serializability requirements?

jfarr:
Hi @aromanenko-dev as long as the `Supplier<Duration>` param is OK then I think this PR is ready. What do you think?

aromanenko-dev:
retest me please

aromanenko-dev:
retest this please

aromanenko-dev:
Run Java PreCommit

jfarr:
@aromanenko-dev Sure, happy to help!

aromanenko-dev:
Hmm, there is some flaky test in KinesisIO caused by `org.apache.beam.sdk.io.kinesis.ShardReadersPoolTest.shouldCallRateLimitPolicy` introduced in this PR. Could you take a look please on [this issue](https://issues.apache.org/jira/browse/BEAM-8382)?

jfarr:
Sure, I will take a look.

jfarr:
https://github.com/apache/beam/pull/11090

PR #9769, opened 2019-10-11
[BEAM-8383] Add metrics to the Python state cache

mxm:
For more insight into how effectively the cache works, metrics have been added
to the Python SDK. All the state operations are counted, as well as metrics like
the current size of the cache, cache hits/misses, and the capacity of the cache.

--- | --- | --- | --- | ---

### DIFF
--- a/runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/GaugeData.java
+++ b/runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/GaugeData.java
@@ -20,6 +20,7 @@ package org.apache.beam.runners.core.metrics;
 import com.google.auto.value.AutoValue;
 import java.io.Serializable;
 import org.apache.beam.sdk.metrics.GaugeResult;
+import org.apache.beam.sdk.transforms.windowing.GlobalWindow;
 import org.joda.time.Instant;
 
 /**
@@ -57,7 +58,7 @@ public abstract class GaugeData implements Serializable {
   public static class EmptyGaugeData extends GaugeData {
 
     private static final EmptyGaugeData INSTANCE = new EmptyGaugeData();
-    private static final Instant EPOCH = new Instant(0);
+    private static final Instant EPOCH = new Instant(GlobalWindow.TIMESTAMP_MIN_VALUE);
 
     private EmptyGaugeData() {}
 
--- a/runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MetricsContainerImpl.java
+++ b/runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MetricsContainerImpl.java
@@ -30,6 +30,7 @@ import org.apache.beam.model.pipeline.v1.MetricsApi.CounterData;
 import org.apache.beam.model.pipeline.v1.MetricsApi.ExtremaData;
 import org.apache.beam.model.pipeline.v1.MetricsApi.IntDistributionData;
 import org.apache.beam.model.pipeline.v1.MetricsApi.MonitoringInfo;
+import org.apache.beam.runners.core.construction.BeamUrns;
 import org.apache.beam.runners.core.metrics.MetricUpdates.MetricUpdate;
 import org.apache.beam.sdk.annotations.Experimental;
 import org.apache.beam.sdk.annotations.Experimental.Kind;
@@ -60,6 +61,9 @@ import org.slf4j.LoggerFactory;
 public class MetricsContainerImpl implements Serializable, MetricsContainer {
 
   private static final Logger LOG = LoggerFactory.getLogger(MetricsContainerImpl.class);
+
+  private static final String GAUGE_URN =
+      BeamUrns.getUrn(MetricsApi.MonitoringInfoTypeUrns.Enum.LATEST_INT64_TYPE);
 
   @Nullable private final String stepName;
 
@@ -306,8 +310,13 @@ public class MetricsContainerImpl implements Serializable, MetricsContainer {
           if (metric.hasCounterData()) {
             CounterData counterData = metric.getCounterData();
             if (counterData.getValueCase() == CounterData.ValueCase.INT64_VALUE) {
-              Counter counter = getCounter(metricName);
-              counter.inc(counterData.getInt64Value());
+              if (GAUGE_URN.equals(monitoringInfo.getType())) {
+                GaugeCell gauge = getGauge(metricName);
+                gauge.set(counterData.getInt64Value());
+              } else {
+                Counter counter = getCounter(metricName);
+                counter.inc(counterData.getInt64Value());
+              }
             } else {
               LOG.warn("Unsupported CounterData type: {}", counterData);
             }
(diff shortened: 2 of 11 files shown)


mxm:

mxm:
Run Java PreCommit

mxm:
I've ran some cluster tests with this. The metrics provide pretty good insight into how well the state caching performance. The hit/miss gauges as well as the current state size/capacity allows users to tune the cache to a decent size. The metrics are available per-bundle and as totals.

This could need a review now :)

mxm:
Just to summarize, these are the metrics provided:

```
# Per-bundle gauges
'get'
'put'
'extend'
'miss'
'hit'
'clear'
'evict'
'size'
'capacity'
# Counters
'get_total'
'put_total'
'extend_total'
'miss_total'
'hit_total'
'clear_total'
'evict_total'
```

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/MetricsContainerImpl.java
```diff
@@ -306,8 +307,15 @@ public void update(Iterable<MonitoringInfo> monitoringInfos) {
           if (metric.hasCounterData()) {
             CounterData counterData = metric.getCounterData();
             if (counterData.getValueCase() == CounterData.ValueCase.INT64_VALUE) {
-              Counter counter = getCounter(metricName);
-              counter.inc(counterData.getInt64Value());
+              String gaugeURN =
+                  BeamUrns.getUrn(MetricsApi.MonitoringInfoTypeUrns.Enum.LATEST_INT64_TYPE);
```
tweise: `gaugeURN` is constant and can be retrieved once.

sdks/python/apache_beam/runners/portability/flink_runner_test.py
```diff
@@ -219,42 +223,72 @@ def test_flattened_side_input(self):
...
+      experiments = options.view_as(DebugOptions).experiments or []
+      experiments.append('state_cache_size=123')
+      options.view_as(DebugOptions).experiments = experiments
       with Pipeline(self.get_runner(), options) as p:
         # pylint: disable=expression-not-assigned
         (p
-         | beam.Create(list(range(n)))
-         | beam.ParDo(DoFn()))
-
+         | "source" >> FlinkStreamingImpulseSource()
```
tweise: Is it necessary to use the (deprecated) source here? Wouldn't `Create` also produce multiple bundles when  `max_bundle_size` is set?

sdks/python/apache_beam/runners/portability/flink_runner_test.py
```diff
@@ -219,42 +223,72 @@ def test_flattened_side_input(self):
...
+                        # only getting initialized after the first bundle.
+                        # Caching depends on the cache token which is lazily
+                        # initialized by the Runner's StateRequestHandlers).
+                        'stateful.beam.metric:statecache:get_total: 100',
+                        'stateful.beam.metric:statecache:miss_total: 10',
+                        'stateful.beam.metric:statecache:hit_total: 90',
+                        'stateful.beam.metric:statecache:put_total: 10',
+                        'stateful.beam.metric:statecache:extend_total: 100',
+                        'stateful.beam.metric:statecache:evict_total: 0',
+                       }
```
tweise: nice!

sdks/python/apache_beam/runners/worker/statecache.py
```diff
@@ -43,19 +131,23 @@ class StateCache(object):
   def __init__(self, max_entries):
     logging.info('Creating state cache with size %s', max_entries)
     self._cache = self.LRUCache(max_entries, (None, None))
-    self._lock = Lock()
+    self._lock = threading.RLock()
+    self._metrics = Metrics()
 
+  @Metrics.counter_hit_miss("get", "hit", "miss")
```
tweise: looks like black magic :)

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/GaugeData.java
```diff
@@ -57,7 +58,7 @@ public GaugeResult extractResult() {
   public static class EmptyGaugeData extends GaugeData {
 
     private static final EmptyGaugeData INSTANCE = new EmptyGaugeData();
-    private static final Instant EPOCH = new Instant(0);
+    private static final Instant EPOCH = new Instant(GlobalWindow.TIMESTAMP_MIN_VALUE);
```
tweise: Why this change?

sdks/python/apache_beam/runners/portability/flink_runner_test.py
```diff
@@ -219,42 +223,72 @@ def test_flattened_side_input(self):
...
-         | beam.ParDo(DoFn()))
-
+         | "source" >> FlinkStreamingImpulseSource()
+         .set_interval_ms(0).set_message_count(110)
+         | "mapper" >> beam.Map(lambda x: (int(x.decode('utf-8')) % 10, 'val'))
+         | "stateful" >> beam.ParDo(DoFn()))
+
+      lines_expected = {'counter: 110',
+                        # Gauges for the last finished bundle
+                        'stateful.beam.metric:statecache:capacity: 123',
```
tweise: Why are the metrics scoped by transform, aren't they by worker? Are the same metrics reported for each transform that executes in a worker?

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/GaugeData.java
mxm: I think this is a bug because the empty gauge will only be overridden with a timestamp > 0, but any time stamp > min timestamp should override it.

sdks/python/apache_beam/runners/portability/flink_runner_test.py
mxm: The metrics are reported per transform because we can have multiple stateful transforms inside an SdkHarness. The only metrics which are per SDK Harness are `size` and `capacity` because the state cache is shared across all the transforms. Metrics currently cannot be reported for the worker itself; that's why these two are duplicated across all transforms. Since the user never knows which Sdk Harness runs the transform, I think it makes sense to always include those two.

mxm:
>Since this throws a few things into the per-record path, have you compared performance (CPU, throughput) before/after?

I've compared the CPU usage and didn't see any difference. Of course there might be a slight overhead but I think it is very small compared to all the other instructions which are involved with state. That said, I didn't observe any CPU increase.

sdks/python/apache_beam/runners/worker/statecache.py
mxm: Decorators are magical :)

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/GaugeData.java
tweise: Better to rename the constant in this case?

mxm: We can rename it in a follow-up if you don't mind.

mxm:
Run Python PreCommit

mxm:
Run Python PreCommit

mxm:
Run Python PreCommit

mxm:
Builds for Python time out, but tests were consistently passing before on Jenkins and locally. There appears to be an issue with the test infrastructure.

PR #9772, opened 2019-10-11
[BEAM-1440] Create a BigQuery source that implements iobase.BoundedSource for Python

kamilwu:
This is an implementation of iobase.BoundedSource for BigQuery.
It works the same way as in Java SDK, which means that it exports BigQuery table to JSON and returns TextSource objects in the split() call.

Please also do note that the source does not currently work with DirectRunner (tested with BundleBasedDirectRunner and SwitchingDirectRunner)

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/io/gcp/bigquery.py
+++ b/sdks/python/apache_beam/io/gcp/bigquery.py
@@ -232,9 +232,11 @@ returned as base64-encoded bytes.
 from __future__ import absolute_import
 
 import collections
+import decimal
 import itertools
 import json
 import logging
+import os
 import time
 import uuid
 from builtins import object
@@ -248,8 +250,13 @@ from apache_beam import coders
 from apache_beam import pvalue
 from apache_beam.internal.gcp.json_value import from_json_value
 from apache_beam.internal.gcp.json_value import to_json_value
+from apache_beam.io.filesystems import CompressionTypes
+from apache_beam.io.filesystems import FileSystems
 from apache_beam.io.gcp import bigquery_tools
 from apache_beam.io.gcp.internal.clients import bigquery
+from apache_beam.io.iobase import BoundedSource
+from apache_beam.io.iobase import SourceBundle
+from apache_beam.io.textio import _TextSource as TextSource
 from apache_beam.options import value_provider as vp
 from apache_beam.options.pipeline_options import DebugOptions
 from apache_beam.options.pipeline_options import GoogleCloudOptions
@@ -262,6 +269,7 @@ from apache_beam.transforms.display import DisplayDataItem
 from apache_beam.transforms.window import GlobalWindows
 from apache_beam.utils import retry
 from apache_beam.utils.annotations import deprecated
+from apache_beam.utils.annotations import experimental
 
 __all__ = [
     'TableRowJsonCoder',
@@ -269,6 +277,7 @@ __all__ = [
     'BigQuerySource',
     'BigQuerySink',
     'WriteToBigQuery',
+    'ReadFromBigQuery',
     'SCHEMA_AUTODETECT',
     ]
 
@@ -497,6 +506,181 @@ class BigQuerySource(dataflow_io.NativeSource):
         use_legacy_sql=self.use_legacy_sql,
         flatten_results=self.flatten_results,
         kms_key=self.kms_key)
+
+
+SchemaFields = collections.namedtuple('SchemaFields', 'fields mode name type')
+
+
+def _to_bool(value):
+  return value == 'true'
+
+
+def _to_decimal(value):
+  return decimal.Decimal(value)
+
+
+class _BigQueryRowCoder(coders.Coder):
+  """A coder for a table row (represented as a dict) from a JSON string which
+  applies additional conversions.
+  """
+
+  def __init__(self, table_schema):
+    # bigquery.TableSchema type is unpickable so we must translate it to a
+    # pickable type
+    self.fields = [SchemaFields(x.fields, x.mode, x.name, x.type)
+                   for x in table_schema.fields]
+    self._converters = {
+        'INTEGER': int,
+        'INT64': int,
+        'FLOAT': float,
+        'BOOLEAN': _to_bool,
+        'NUMERIC': _to_decimal,
+    }
+
+  def decode(self, value):
+    value = json.loads(value)
+    for field in self.fields:
+      if field.name not in value:
+        # The field exists in the schema, but it doesn't exist in this row.
+        # It probably means its value was null, as the extract to JSON job
+        # doesn't preserve null fields
+        value[field.name] = None
+        continue
+
+      try:
+        converter = self._converters[field.type]
+        value[field.name] = converter(value[field.name])
+      except KeyError:
+        # No need to do any conversion
+        pass
+    return value
+
+  def is_deterministic(self):
+    return True
+
+  def to_type_hint(self):
+    return dict
+
+
+class _BigQuerySource(BoundedSource):
+  def __init__(self, table=None, dataset=None, project=None, query=None,
+               validate=False, coder=None, use_standard_sql=False,
+               flatten_results=True, kms_key=None, gcs_bucket_name=None):
+    if gcs_bucket_name is None:
+      raise ValueError('The name of the GCS bucket must be specified')
+    self.gcs_bucket_name = gcs_bucket_name
+
+    if table is not None and query is not None:
+      raise ValueError('Both a BigQuery table and a query were specified.'
+                       ' Please specify only one of these.')
+    elif table is None and query is None:
+      raise ValueError('A BigQuery table or a query must be specified')
+    elif table is not None:
+      self.table_reference = bigquery_tools.parse_table_reference(
+          table, dataset, project)
+      self.query = None
+      self.use_legacy_sql = True
+    else:
+      self.query = query
+      # TODO(BEAM-1082): Change the internal flag to be standard_sql
+      self.use_legacy_sql = not use_standard_sql
+      self.table_reference = None
+
+    self.project = project
+    self.validate = validate
+    self.flatten_results = flatten_results
+    self.coder = coder or _BigQueryRowCoder
+    self.kms_key = kms_key
+    self.split_result = None
+
+  def estimate_size(self):
+    bq = bigquery_tools.BigQueryWrapper()
+    if self.table_reference is not None:
+      table = bq.get_table(self.table_reference.projectId,
+                           self.table_reference.datasetId,
+                           self.table_reference.tableId)
+      return int(table.numBytes)
+    else:
+      self._setup_temporary_dataset(bq)
+      job = bq._start_query_job(self.project, self.query,
+                                self.use_legacy_sql, self.flatten_results,
+                                job_id=uuid.uuid4().hex, dry_run=True,
+                                kms_key=self.kms_key)
+      size = int(job.statistics.totalBytesProcessed)
+
+      bq.clean_up_temporary_dataset(self.project)
+
+      return size
+
+  def split(self, desired_bundle_size, start_position=None, stop_position=None):
+    if self.split_result is None:
+      bq = bigquery_tools.BigQueryWrapper()
+
+      if self.query is not None:
+        self._setup_temporary_dataset(bq)
+        self.table_reference = self._execute_query(bq)
+
+      schema, metadata_list = self._export_files(bq)
+      self.split_result = [TextSource(metadata.path, 0,
+                                      CompressionTypes.UNCOMPRESSED, True,
+                                      self.coder(schema))
+                           for metadata in metadata_list]
+
+      if self.query is not None:
+        bq.clean_up_temporary_dataset(self.project)
+
+    for source in self.split_result:
+      yield SourceBundle(0, source, None, None)
+
+  def get_range_tracker(self, start_position, stop_position):
+    raise NotImplementedError('BigQuery source must be split before being read')
+
+  def read(self, range_tracker):
+    raise NotImplementedError('BigQuery source must be split before being read')
+
+  def _setup_temporary_dataset(self, bq):
+    location = bq.get_query_location(self.project, self.query,
+                                     self.use_legacy_sql)
+    bq.create_temporary_dataset(self.project, location)
+
+  def _execute_query(self, bq):
+    job = bq._start_query_job(self.project, self.query,
+                              self.use_legacy_sql, self.flatten_results,
+                              job_id=uuid.uuid4().hex, kms_key=self.kms_key)
+    job_ref = job.jobReference
+    bq.wait_for_bq_job(job_ref)
+    return bq._get_temp_table(self.project)
+
+  def _export_files(self, bq):
+    """Runs a BigQuery export job.
+
+    Returns:
+      bigquery.TableSchema instance, a list of FileMetadata instances
+    """
+    job_id = uuid.uuid4().hex
+    destination = self._get_destination_uri(self.gcs_bucket_name, job_id)
+    job_ref = bq.perform_extract_job([destination], job_id,
+                                     self.table_reference,
+                                     bigquery_tools.ExportFileFormat.JSON,
+                                     include_header=False)
+    bq.wait_for_bq_job(job_ref)
+    metadata_list = FileSystems.match([destination])[0].metadata_list
+
+    table = bq.get_table(self.table_reference.projectId,
+                         self.table_reference.datasetId,
+                         self.table_reference.tableId)
+
+    return table.schema, metadata_list
+
+  @staticmethod
+  def _get_destination_uri(gcs_bucket_name, subdir_name):
+    """Returns the fully qualified Google Cloud Storage URI where the
+    extracted table should be written.
+    """
+    # We need subdirectories to prevent from overwriting files when
+    # multiple export jobs run in parallel.
+    return os.path.join(gcs_bucket_name, subdir_name,
+                        'bigquery-table-dump-*.json')
 
 
 @deprecated(since='2.11.0', current="WriteToBigQuery")
@@ -1274,3 +1458,60 @@ bigquery_v2_messages.TableSchema):
                                    tableSpec)
       res['table'] = DisplayDataItem(tableSpec, label='Table')
     return res
+
+
+@experimental()
+class ReadFromBigQuery(PTransform):
+  """Read data from BigQuery.
+
+    This PTransform uses a BigQuery export job to take a snapshot of the table
+    on GCS, and then reads from each produced JSON file.
+
+    Do note that currently this source does not work with DirectRunner.
+
+  Args:
+    table (str, callable, ValueProvider): The ID of the table, or a callable
+      that returns it. The ID must contain only letters ``a-z``, ``A-Z``,
+      numbers ``0-9``, or underscores ``_``. If dataset argument is
+      :data:`None` then the table argument must contain the entire table
+      reference specified as: ``'DATASET.TABLE'``
+      or ``'PROJECT:DATASET.TABLE'``. If it's a callable, it must receive one
+      argument representing an element to be written to BigQuery, and return
+      a TableReference, or a string table name as specified above.
+    dataset (str): The ID of the dataset containing this table or
+      :data:`None` if the table reference is specified entirely by the table
+      argument.
+    project (str): The ID of the project containing this table.
+    query (str): A query to be used instead of arguments table, dataset, and
+      project.
+    validate (bool): If :data:`True`, various checks will be done when source
+      gets initialized (e.g., is table present?). This should be
+      :data:`True` for most scenarios in order to catch errors as early as
+      possible (pipeline construction instead of pipeline execution). It
+      should be :data:`False` if the table is created during pipeline
+      execution by a previous step.
+    coder (~apache_beam.coders.coders.Coder): The coder for the table
+      rows. If :data:`None`, then the default coder is
+      :class:`~apache_beam.io.gcp.bigquery._BigQueryRowCoder`,
+      which will interpret every line in a file as a JSON serialized
+      dictionary. This argument needs a value only in special cases when
+      returning table rows as dictionaries is not desirable.
+    use_standard_sql (bool): Specifies whether to use BigQuery's standard SQL
+      dialect for this query. The default value is :data:`False`.
+      If set to :data:`True`, the query will use BigQuery's updated SQL
+      dialect with improved standards compliance.
+      This parameter is ignored for table inputs.
+    flatten_results (bool): Flattens all nested and repeated fields in the
+      query results. The default value is :data:`True`.
+    kms_key (str): Experimental. Optional Cloud KMS key name for use when
+      creating new temporary tables.
+    gcs_bucket_name (str): The name of the Google Cloud Storage bucket where
+      the extracted table should be written.
+   """
+  def __init__(self, *args, **kwargs):
+    self._args = args
+    self._kwargs = kwargs
+
+  def expand(self, pcoll):
+    return (pcoll
+            | beam.io.Read(_BigQuerySource(*self._args, **self._kwargs)))
--- a/sdks/python/apache_beam/io/gcp/bigquery.py
+++ b/sdks/python/apache_beam/io/gcp/bigquery.py
@@ -236,7 +236,6 @@ import decimal
 import itertools
 import json
 import logging
-import os
 import time
 import uuid
 from builtins import object
@@ -261,6 +260,8 @@ from apache_beam.options import value_provider as vp
 from apache_beam.options.pipeline_options import DebugOptions
 from apache_beam.options.pipeline_options import GoogleCloudOptions
 from apache_beam.options.pipeline_options import StandardOptions
+from apache_beam.options.value_provider import StaticValueProvider
+from apache_beam.options.value_provider import ValueProvider
 from apache_beam.runners.dataflow.native_io import iobase as dataflow_io
 from apache_beam.transforms import DoFn
 from apache_beam.transforms import ParDo
@@ -543,7 +544,7 @@ class _JsonToDictCoder(coders.Coder):
   def _convert_to_tuple(cls, table_field_schemas):
     """Recursively converts the list of TableFieldSchema instances to the
     list of tuples to prevent errors when pickling and unpickling
-    TableFieldSchema` instances.
+    TableFieldSchema instances.
     """
     if not table_field_schemas:
       return []
@@ -585,13 +586,9 @@ class _JsonToDictCoder(coders.Coder):
 
 
 class _BigQuerySource(BoundedSource):
-  def __init__(self, table=None, dataset=None, project=None, query=None,
-               validate=False, coder=None, use_standard_sql=False,
-               flatten_results=True, kms_key=None, gcs_bucket_name=None):
-    if gcs_bucket_name is None:
-      raise ValueError('The name of the GCS bucket must be specified')
-    self.gcs_bucket_name = gcs_bucket_name
-
+  def __init__(self, gcs_location=None, table=None, dataset=None,
+               project=None, query=None, validate=False, coder=None,
+               use_standard_sql=False, flatten_results=True, kms_key=None):
     if table is not None and query is not None:
       raise ValueError('Both a BigQuery table and a query were specified.'
                        ' Please specify only one of these.')
@@ -608,6 +605,7 @@ class _BigQuerySource(BoundedSource):
       self.use_legacy_sql = not use_standard_sql
       self.table_reference = None
 
+    self.gcs_location = gcs_location
     self.project = project
     self.validate = validate
     self.flatten_results = flatten_results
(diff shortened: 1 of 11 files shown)


kamilwu:
Run Python 2 PostCommit

kamilwu:
Run Python 2 PostCommit

kamilwu:
Run Java PreCommit

kamilwu:
Run Python 37 PostCommit

kamilwu:
Run Python 2 PostCommit

kamilwu:
Run PythonLint PreCommit

kamilwu:
Run Python 3 PostCommit

kamilwu:
Run Python 3.7 PostCommit

kamilwu:
Run Python 3.7 PostCommit

kamilwu:
Run Python 2 PostCommit

kamilwu:
Run Java PreCommit

kamilwu:
Run Python 2 PostCommit

kamilwu:
Run Python 3.7 PostCommit

kamilwu:
Run JavaBeamZetaSQL PreCommit

kamilwu:
Run Python 3.7 PostCommit

kamilwu:
Run Python 2 PostCommit

kamilwu:
Run Java PreCommit

kamilwu:
Run Python PreCommit

kamilwu:

kamilwu:
R: @pabloem @chamikaramj  Please take a look. Any thoughts are welcome

pabloem:
taking a look

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -496,6 +505,233 @@ def reader(self, test_bigquery_client=None):
...
+      self.query = query
+      # TODO(BEAM-1082): Change the internal flag to be standard_sql
+      self.use_legacy_sql = not use_standard_sql
+      self.table_reference = None
+
+    self.project = project
+    self.validate = validate
+    self.flatten_results = flatten_results
+    self.coder = coder or _BigQueryRowCoder
+    self.kms_key = kms_key
```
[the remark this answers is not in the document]
pabloem: Ah nvm I see now. This kms key should be used when creating a temporary dataset with the query, right? It seems likeit's not used for that atm.

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -496,6 +505,233 @@ def reader(self, test_bigquery_client=None):
...
+      query results. The default value is :data:`True`.
+    kms_key (str): Experimental. Optional Cloud KMS key name for use when
+      creating new tables.
+    gcs_bucket_name (str): The name of the Google Cloud Storage bucket where
+      the extracted table should be written.
+   """
+  def __init__(self, table=None, dataset=None, project=None, query=None,
+               validate=False, coder=None, use_standard_sql=False,
+               flatten_results=True, kms_key=None, gcs_bucket_name=None):
+    if gcs_bucket_name is None:
```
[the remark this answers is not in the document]
pabloem: Also, we should probably have something that does all the following:
- Support valueproviders
- Try to use temp_location as the gcs location

If none of these are a gcs location, or they're all none, then we can error out at execution time.

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -1265,3 +1501,14 @@ def display_data(self):
                                    tableSpec)
       res['table'] = DisplayDataItem(tableSpec, label='Table')
     return res
+
+
+@experimental()
+class ReadFromBigQuery(PTransform):
+  def __init__(self, *args, **kwargs):
```
pabloem: Since `ReadFromBigQuery` is the user-facing transform, it should have all the Pydoc.
That being said, I'll defer @chamikaramj whether we want to expose this already.

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -496,6 +505,233 @@ def reader(self, test_bigquery_client=None):
...
+  return decimal.Decimal(value)
+
+
+def _to_bytes(value):
+  """Converts value from str to bytes on Python 3.x. Does nothing on
+  Python 2.7."""
+  return value.encode('utf-8')
+
+
+class _BigQueryRowCoder(coders.Coder):
```
pabloem: I feel silly - but didn't we have a coder for this? I see that `TableRowJsonCoder` does not do the full tablerow to dict conversion... does it make sense to extend it? Or not really?
[1] https://github.com/apache/beam/blob/12d07745835e1b9c1e824b83beeeadf63ab4b234/sdks/python/apache_beam/io/gcp/bigquery.py#L312-L349

sdks/python/apache_beam/io/gcp/bigquery_tools.py
```diff
@@ -370,7 +383,37 @@ def _start_query_job(self, project_id, query, use_legacy_sql, flatten_results,
...
+        reached.
+    """
+    retry = 0
+    while True:
+      retry += 1
+      job = self.get_job(job_reference.projectId, job_reference.jobId,
+                         job_reference.location)
+      logging.info('Job status: %s', job.status.state)
+      if job.status.state == 'DONE' and job.status.errorResult:
+        raise RuntimeError("BigQuery job %s failed. Error Result: %s",
```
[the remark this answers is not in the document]
kamilwu: Oh, you're right. In that some maybe I'll use the `format` method

sdks/python/apache_beam/io/gcp/bigquery_tools.py
```diff
@@ -695,10 +769,12 @@ def get_or_create_table(
 
   def run_query(self, project_id, query, use_legacy_sql, flatten_results,
                 dry_run=False):
-    job_id, location = self._start_query_job(project_id, query,
-                                             use_legacy_sql, flatten_results,
-                                             job_id=uuid.uuid4().hex,
-                                             dry_run=dry_run)
+    job = self._start_query_job(project_id, query, use_legacy_sql,
```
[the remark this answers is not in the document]
kamilwu: Yes. It returns the whole job object because I needed its `statistics` property. See usage: bigquery.py:655

sdks/python/apache_beam/io/gcp/bigquery_read_it_test.py
```diff
@@ -86,23 +117,61 @@ def create_table(self, tablename):
...
         {'number': 3, 'str': u'你好'},
         {'number': 4, 'str': u'привет'}
     ]
-    self.bigquery_client.insert_rows(
-        self.project, self.dataset_id, tablename, table_data)
+    cls.bigquery_client.insert_rows(
+        cls.project, cls.dataset_id, table_name, table_data)
 
-  def create_table_new_types(self, table_name):
+  def get_expected_data(self):
```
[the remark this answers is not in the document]
kamilwu: +1

sdks/python/apache_beam/io/gcp/bigquery_read_it_test.py
```diff
@@ -86,23 +117,61 @@ def create_table(self, tablename):
...
+          query=self.query, use_standard_sql=True)))
+      assert_that(result, equal_to(self.get_expected_data()))
+
+  @skip(['DirectRunner', 'TestDirectRunner'])
+  @attr('IT')
+  def test_iobase_source(self):
+    with beam.Pipeline(argv=self.args) as p:
+      result = (p | 'read' >> beam.io.ReadFromBigQuery(
+          query=self.query, use_standard_sql=True, project=self.project,
+          gcs_bucket_name='gs://temp-storage-for-end-to-end-tests'))
```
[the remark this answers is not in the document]
kamilwu: > Why are we skipping this for DirectRunner? This should work there, right?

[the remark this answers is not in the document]
kamilwu: > Why are we skipping this for DirectRunner? This should work there, right?

sdks/python/test-suites/portable/py37/build.gradle
```diff
@@ -30,3 +33,25 @@ task preCommitPy37() {
...
+
+  doLast {
+    def tests = [
+        "apache_beam.io.gcp.bigquery_read_it_test",
+    ]
+    def testOpts = ["--tests=${tests.join(',')}"]
+    def cmdArgs = mapToArgString([
+        "test_opts": testOpts,
+        "suite": "postCommitIT-flink-py37",
+        "pipeline_opts": "--runner=FlinkRunner --project=apache-beam-testing --environment_type=LOOPBACK",
```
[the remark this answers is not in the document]
kamilwu: Yes, locally. I would phrase trigger a PostCommit check, but it is getting aborted almost always: https://builds.apache.org/view/A-D/view/Beam/view/All/job/beam_PostCommit_Python37/

sdks/python/apache_beam/io/gcp/bigquery_read_it_test.py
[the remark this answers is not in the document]
kamilwu: > Why are we skipping this for DirectRunner? This should work there, right?
Unfortunately, no. My solution doesn't work with DirectRunner.

The direct cause is `get_range_tracker` and `read` methods aren't implemented in my source (they're raising NotImplementedError exception). This is purposeful — the runner is expected to call `split` instead. See Java implementation which works the same way: [link](https://github.com/apache/beam/blob/c2f0d282337f3ae0196a7717712396a5a41fdde1/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQuerySourceBase.java)

It seems that DataflowRunner and Flink are able to catch these exceptions somehow, while DirectRunner is not.

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -496,6 +505,233 @@ def reader(self, test_bigquery_client=None):
...
+    for source in self.split_result:
+      yield SourceBundle(0, source, None, None)
+
+  def get_range_tracker(self, start_position, stop_position):
+    raise NotImplementedError('BigQuery source must be split before being read')
+
+  def read(self, range_tracker):
+    raise NotImplementedError('BigQuery source must be split before being read')
+
+  def _setup(self, bq):
```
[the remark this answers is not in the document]
kamilwu: No, just a dataset setup. `setup_temporary_dataset` sounds good, +1

sdks/python/apache_beam/io/gcp/bigquery.py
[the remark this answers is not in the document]
kamilwu: Thanks for spotting this. I will write a fix so that the kms key will be used when creating a temporary dataset. There are no other places where kms key should be used, right?

sdks/python/apache_beam/io/gcp/bigquery.py
kamilwu: +1

sdks/python/apache_beam/io/gcp/bigquery.py
kamilwu: Looks interesting. If I combine this coder with this function[1], maybe that would be the solution. I'll investigate this further — first I need to solve the problem that `bigquery.TableSchema` is unpickable.

[1] https://github.com/apache/beam/blob/03f780c7329e0eca692baef44874056b7d263303/sdks/python/apache_beam/io/gcp/bigquery_tools.py#L892-L920

sdks/python/apache_beam/io/gcp/bigquery_read_it_test.py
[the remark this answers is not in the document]
pabloem: I see. Can you file a bug for this?

pabloem:
thanks!

sdks/python/apache_beam/io/gcp/bigquery_read_it_test.py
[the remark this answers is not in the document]
kamilwu: Done: https://issues.apache.org/jira/browse/BEAM-8528

sdks/python/apache_beam/io/gcp/bigquery.py
kamilwu: I looked at this and, sadly, it doesn't work in practice. For each field `TableRowJsonCoder` returns `JsonValue` instance which doesn't contain the name of the given field. This makes finding an appropriate field type and matching a conversion function quite hard.

That's why I decided to stay with my own coder.

kamilwu:
@pabloem Could you take a look once again? In addition to what you had suggested, I've added the functionality of removing JSON file after being read.

kamilwu:
Run Python 3.7 PostCommit

kamilwu:
Run Python 2 PostCommit

kamilwu:
Run Java PreCommit

kamilwu:
Run Python 2 PostCommit

kamilwu:
Run Java PreCommit

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -496,6 +506,189 @@ def reader(self, test_bigquery_client=None):
...
+        'FLOAT': float,
+        'BOOLEAN': _to_bool,
+        'NUMERIC': _to_decimal,
+        'BYTES': _to_bytes,
+    }
+
+  @classmethod
+  def _convert_to_tuple(cls, table_field_schemas):
+    """Recursively converts the list of TableFieldSchema instances to the
+    list of tuples to prevent errors when pickling and unpickling
```
kamilwu: This error was quite interesting. It seems that it's impossible to serialize and deserialize nested `bigquery.TableFieldSchema` instances:
```

from apache_beam.internal import pickler
from apache_beam.io.gcp.internal.clients import bigquery

obj = bigquery.TableFieldSchema(fields=[bigquery.TableFieldSchema()])
pickler.loads(pickler.dumps(obj))
```

This snippet triggers the following exception:
`AttributeError: 'FieldList' object has no attribute '_FieldList__field'`.

My workaround was to rewrite all TableFieldSchema instances to an equivalent tuple, which can be serialized and deserialized without problems.

pabloem:
I had been traveling. I'll take look now.

kamilwu:
Resolved merge conflicts. 
@pabloem Could you take a look once again?

kamilwu:
Run Python PreCommit

kamilwu:
Run Python PreCommit

pabloem:
Looking once more.

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -499,6 +509,189 @@ def reader(self, test_bigquery_client=None):
...
+    return value
+
+  def is_deterministic(self):
+    return True
+
+  def to_type_hint(self):
+    return dict
+
+
+class _BigQuerySource(BoundedSource):
```
pabloem: Please rename it _CustomBigQuerySource - as we already have a BigQuerySource. To avoid confusion.

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -1268,3 +1461,140 @@ def display_data(self):
                                    tableSpec)
       res['table'] = DisplayDataItem(tableSpec, label='Table')
     return res
+
+
+@experimental()
+class PassThroughThenCleanup(PTransform):
```
pabloem: Does this class need to be public? If it should be public, it should be in a different file. If not, let's make it private and only use it in this file.

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -1268,3 +1461,140 @@ def display_data(self):
...
+    _ = (input.pipeline
+         | beam.Create([None])
+         | beam.ParDo(self.cleanup_dofn, beam.pvalue.AsSingleton(
+             cleanup_signal)))
+
+    return main_output
+
+
+@experimental()
+class ReadFromBigQuery(PTransform):
```
pabloem: This is looking great. I've discussed with Cham, and let's rename this as `_ReadFromBigQuery` (with underscore) to prevent users from picking it up before we have tested it. We have some tests that we'll run on it, and once we're confident of performance/functionality, we can remove the underscore, and rename to `ReadFromBigQuery`. WDYT?

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -499,6 +509,189 @@ def reader(self, test_bigquery_client=None):
...
+        continue
+
+      if field.type == 'RECORD':
+        value[field.name] = self._decode_with_schema(value[field.name],
+                                                     field.fields)
+      else:
+        try:
+          converter = self._converters[field.type]
+          value[field.name] = converter(value[field.name])
+        except KeyError:
```
pabloem: Does this mean that for other data types, we pass them as they are? e.g. for datetime data, or other like that?

ananvay:
@robertwb 
/cc: @ananvay

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -499,6 +509,189 @@ def reader(self, test_bigquery_client=None):
...
+    return bq._get_temp_table(self.project)
+
+  def _export_files(self, bq):
+    """Runs a BigQuery export job.
+
+    Returns:
+      bigquery.TableSchema instance, a list of FileMetadata instances
+    """
+    job_id = uuid.uuid4().hex
+    job_ref = bq.perform_extract_job([self.gcs_location], job_id,
```
pabloem: after checking with @chamikaramj - it seems we should export using avro to gain performance.

pabloem: We should consider the possibility of failures:
If this call fails and is retried - but an export job still runs twice, we may duplicate the files for export. Is that a correct assessment? 
If duplication of files on retries is a possibility, then we should generate a random subdirectory within `gcs_location` to only read from that subdirectory when the call succeeds. Does that make sense?

chamikaramj: Yeah, I believe all existing (Java and Python) BQ sources use Avro as the export format. We switched from JSON to Avro sometime back due to performance gains. Sink still uses JSON.

kamilwu: Maybe I'll create a jira for this as soon as this PR is merged. Is that alright?

sdks/python/apache_beam/io/gcp/bigquery.py
kamilwu: I'm totally fine with it. 
Once this PR is merged, I'm going to make changes to the Chicago Taxi Example, so that the example would use this transform. That would be a great opportunity to check stability and measure performance of the transform

sdks/python/apache_beam/io/gcp/bigquery.py
kamilwu: I don't think it needs to be public. I'll make it private then.

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -700,30 +675,19 @@ def _export_files(self, bq):
...
     job_id = uuid.uuid4().hex
-    destination = self._get_destination_uri(self.gcs_bucket_name, job_id)
-    job_ref = bq.perform_extract_job([destination], job_id,
+    job_ref = bq.perform_extract_job([self.gcs_location], job_id,
                                      self.table_reference,
                                      bigquery_tools.ExportFileFormat.JSON,
                                      include_header=False)
     bq.wait_for_bq_job(job_ref)
-    metadata_list = FileSystems.match([destination])[0].metadata_list
+    metadata_list = FileSystems.match([self.gcs_location])[0].metadata_list
```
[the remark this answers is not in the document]
kamilwu: Yes. The thing I used is called `Single wildcard URI`[1].  In this case, an extract job creates one or many files and all of them are created in the same directory.

[1] https://cloud.google.com/bigquery/docs/exporting-data#exporting_data_into_one_or_more_files

sdks/python/apache_beam/io/gcp/bigquery.py
kamilwu: +1

sdks/python/apache_beam/io/gcp/bigquery.py
kamilwu: Yes. I wanted to implement the same behavior as we already have when using default coder for the native BigQuerySource. There is a [test](https://github.com/apache/beam/blob/03f780c7329e0eca692baef44874056b7d263303/sdks/python/apache_beam/io/gcp/bigquery_read_it_test.py#L232) that checks BigQuery data types conversions across both sources.

sdks/python/apache_beam/io/gcp/bigquery.py
kamilwu: I found in the doc[1] something that may answer your question:
```
Calling jobs.insert on a given job ID is idempotent; in other words, you can retry as many times as you like on the same job ID, and at most one of those operations will succeed.
```
When retry is called, job ID is the same, so I think we are save.

[1] https://cloud.google.com/bigquery/docs/exporting-data#exporting_data_stored_in

kamilwu:
Run Python 2 PostCommit

kamilwu:
Run Python PreCommit

kamilwu:
Run Python PreCommit

sdks/python/apache_beam/io/gcp/bigquery.py
pabloem: Yes, that will be great to compare the Native Dataflow source with the new Beam source.

sdks/python/apache_beam/io/gcp/bigquery.py
pabloem: Except the job id is not the same, right? because `uuid.uuid4().hex` returns different IDs every time?

kamilwu: `_export_files()` is called only once, so the job id is the same within a pipeline. And in the case when two or more pipelines are ran at the same time, we generate a random subdirectory within `gcs_location` right here:

https://github.com/apache/beam/blob/894431000cf72a122422257365e0f3b775c0c44f/sdks/python/apache_beam/io/gcp/bigquery.py#L1571-L1572

kamilwu:
Run Portable_Python PreCommit

kamilwu:
Run Python 2 PostCommit

kamilwu:
I've tidied up commit history a bit. Also, I've renamed `ReadFromBigQuery` to `_ReadFromBigQuery` - I forgot to do this in an earlier commit

kamilwu:
Run Python 3.5 PostCommit

sdks/python/apache_beam/io/gcp/bigquery.py
pabloem: gotcha. This makes sense. Thanks Kamil!

pabloem:
as a final note, please create issues to switch to avro, and to enable tests in other runners (e.g. direct runner).

pabloem:
Run Python 3.5 PostCommit

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -499,6 +509,189 @@ def reader(self, test_bigquery_client=None):
...
+                           self.table_reference.datasetId,
+                           self.table_reference.tableId)
+      return int(table.numBytes)
+    else:
+      self._setup_temporary_dataset(bq)
+      job = bq._start_query_job(self.project, self.query,
+                                self.use_legacy_sql, self.flatten_results,
+                                job_id=uuid.uuid4().hex, dry_run=True,
+                                kms_key=self.kms_key)
+      size = int(job.statistics.totalBytesProcessed)
```
robertwb: Is this input or output bytes?

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -499,6 +509,189 @@ def reader(self, test_bigquery_client=None):
...
+      bq.clean_up_temporary_dataset(self.project)
+
+      return size
+
+  def split(self, desired_bundle_size, start_position=None, stop_position=None):
+    if self.split_result is None:
+      bq = bigquery_tools.BigQueryWrapper()
+
+      if self.query is not None:
+        self._setup_temporary_dataset(bq)
```
robertwb: Is this redundant with what's done in `estimate_size()`? Is it expensive?

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -499,6 +509,189 @@ def reader(self, test_bigquery_client=None):
...
+        self.table_reference = self._execute_query(bq)
+
+      schema, metadata_list = self._export_files(bq)
+      self.split_result = [TextSource(metadata.path, 0,
```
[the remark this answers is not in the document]
kamilwu: I prefer to leave it as it is now. The PR is now two months old and I already had to put an effort to keep it up-to-date (the PR is also quite big and this doesn't help). If you are OK with that, for now, I'll create a jira for switching to avro. WDYT?

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -499,6 +509,189 @@ def reader(self, test_bigquery_client=None):
...
+  def estimate_size(self):
+    bq = bigquery_tools.BigQueryWrapper()
+    if self.table_reference is not None:
+      table = bq.get_table(self.table_reference.projectId,
+                           self.table_reference.datasetId,
+                           self.table_reference.tableId)
+      return int(table.numBytes)
+    else:
+      self._setup_temporary_dataset(bq)
+      job = bq._start_query_job(self.project, self.query,
```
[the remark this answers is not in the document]
kamilwu: This one does not block due to dry_run being `True`. When dry_run is true, the job is not actually ran. Instead, we get some processing statistics (one of them is the number of bytes read by the query)

> Is estimate_size() called during pipeline construction? Is it guaranteed to be called only (exactly) once?

I think a runner is free to call estimate_size() as many times as it needs. 
I see one major improvement in this particular case - temporary datasets don't need to be created when estimating size. As the result, estimate_size() becomes even more simple. Then, I think we are fine with estimate_size() being called multiple times.

sdks/python/apache_beam/io/gcp/bigquery.py
kamilwu: Please see one of my comment above, where I described that creating temporary dataset in `estimate_size()` is unnecessary. I think that solves the problem

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -499,6 +509,189 @@ def reader(self, test_bigquery_client=None):
...
+    elif table is None and query is None:
+      raise ValueError('A BigQuery table or a query must be specified')
+    elif table is not None:
+      self.table_reference = bigquery_tools.parse_table_reference(
+          table, dataset, project)
+      self.query = None
+      self.use_legacy_sql = True
+    else:
+      self.query = query
+      # TODO(BEAM-1082): Change the internal flag to be standard_sql
```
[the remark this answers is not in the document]
kamilwu: It was copied-pasted from the native BigQuery source (`BigQuerySource` class), because their interfaces are mostly the same. I'll check if this can be improved.

sdks/python/apache_beam/io/gcp/bigquery.py
kamilwu: Input. And to be more precise - that's the number of bytes which must be read from BigQuery.

kamilwu:
Thanks @robertwb for your comments!

> Why does this not work on the direct runners. Is it an issue of needing to be split first?
Yes. I've already created a jira for this: https://issues.apache.org/jira/browse/BEAM-8528

> would it make sense to implement this as an SDF instead?

My first attempt was a regular (non splittable) DoFn that triggers export job followed by `MatchAll` and `ReadMatches` transforms. This worked, but I had troubles with implementing the rest: waiting for query job, waiting for export job and removing json files after reading. Using Source API turned out to be simpler.

kamilwu:
Jira issue to switch to avro: https://issues.apache.org/jira/browse/BEAM-8910

kamilwu:
Run Python 2 PostCommit

kamilwu:
Run Python PreCommit

kamilwu:
Run Python PreCommit

kamilwu:
Run Python PreCommit

pabloem:
Run Python PreCommit

kamilwu:
Run Python 2 PostCommit

kamilwu:
Run Python 2 PostCommit

kamilwu:
Run Python PreCommit

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -1274,3 +1463,139 @@ def display_data(self):
...
+                      len(match_result))
+        paths = [x.path for x in match_result]
+        FileSystems.delete(paths)
+
+    temp_location = pcoll.pipeline.options.view_as(
+        GoogleCloudOptions).temp_location
+    gcs_location = self._get_destination_uri(temp_location)
+
+    return (pcoll
+            | beam.io.Read(_CustomBigQuerySource(gcs_location=gcs_location,
```
chamikaramj: Quick question. Have we considered workiem failures and retries ? We have to be careful to not to loose data in such cases. For example, what will happen if one of the workitems is failed and retried after deleting some files in _PassThroughThenCleanup step ?

pabloem: The pasthroughthencleanup step produces and consumes side inputs, which means that it happens as part of a separate stage. The stage boundary 'commits' the data from files before deleting them.

sdks/python/apache_beam/io/gcp/bigquery.py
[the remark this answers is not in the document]
robertwb: As long as it actually happens right away. Create the JIRA assigned to yourself and drop a TODO into the code.

sdks/python/apache_beam/io/gcp/bigquery.py
robertwb: It seems the reported estimated size should be the output bytes, but if that's not easy to get hopefully this is at least a reasonable approximation.

sdks/python/apache_beam/io/gcp/bigquery.py
chamikaramj: Makes sense. Thanks.

kamilwu: It is just like @pabloem wrote. When _PassThroughThenCleanup step executes a cleanup function, we can be sure that the input has already been processed.

tvalentyn:
Postcommits tests are failing with this change: https://issues.apache.org/jira/browse/BEAM-8988.

kamilwu:
Thanks to everyone involved.

@tvalentyn PR with a fix is ready: https://github.com/apache/beam/pull/10412

TheNeuralBit:
It looks like there's another error on the Py3.5 PostCommit due to str vs bytes in json.loads: https://issues.apache.org/jira/browse/BEAM-8995

kamilwu:
@TheNeuralBit  Thanks. I've added a commit with a fix to the PR above

PR #9777, opened 2019-10-11
[BEAM-8341]: basic bundling support for portable runner

lhaiesp:
basic bundling support for portable runner

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/runners/samza/src/main/java/org/apache/beam/runners/samza/runtime/DoFnOp.java
+++ b/runners/samza/src/main/java/org/apache/beam/runners/samza/runtime/DoFnOp.java
@@ -17,6 +17,8 @@
  */
 package org.apache.beam.runners.samza.runtime;
 
+import static org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Preconditions.checkState;
+
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
@@ -24,6 +26,8 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.ServiceLoader;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
 import org.apache.beam.model.pipeline.v1.RunnerApi;
 import org.apache.beam.runners.core.DoFnRunner;
 import org.apache.beam.runners.core.DoFnRunners;
@@ -32,14 +36,15 @@ import org.apache.beam.runners.core.SideInputHandler;
 import org.apache.beam.runners.core.SimplePushbackSideInputDoFnRunner;
 import org.apache.beam.runners.core.StateNamespace;
 import org.apache.beam.runners.core.StateNamespaces;
+import org.apache.beam.runners.core.StateTags;
 import org.apache.beam.runners.core.TimerInternals;
-import org.apache.beam.runners.core.construction.SerializablePipelineOptions;
 import org.apache.beam.runners.core.construction.graph.ExecutableStage;
-import org.apache.beam.runners.core.serialization.Base64Serializer;
 import org.apache.beam.runners.fnexecution.control.StageBundleFactory;
 import org.apache.beam.runners.samza.SamzaExecutionContext;
 import org.apache.beam.runners.samza.SamzaPipelineOptions;
 import org.apache.beam.sdk.coders.Coder;
+import org.apache.beam.sdk.state.BagState;
+import org.apache.beam.sdk.state.TimeDomain;
 import org.apache.beam.sdk.transforms.DoFn;
 import org.apache.beam.sdk.transforms.DoFnSchemaInformation;
 import org.apache.beam.sdk.transforms.join.RawUnionValue;
@@ -57,6 +62,7 @@ import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.Iterator
 import org.apache.samza.config.Config;
 import org.apache.samza.context.Context;
 import org.apache.samza.operators.Scheduler;
+import org.joda.time.Duration;
 import org.joda.time.Instant;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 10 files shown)


lhaiesp:
@xinyuiscool

runners/samza/src/main/java/org/apache/beam/runners/samza/runtime/DoFnOp.java
```diff
@@ -237,9 +292,39 @@ private String getTimerStateId(DoFnSignature signature) {
     return builder.toString();
   }
 
+  private void attemptStartBundle() {
+    if (isBundleStarted.compareAndSet(false, true)) {
+      currentBundleElementCount.set(0L);
+      bundleStartTime.set(System.currentTimeMillis());
+      pushbackFnRunner.startBundle();
```
xinyuiscool: Shall we invoke scheduleNextBundleCheck() after startBundle()? Right now schedule() is invoked in multiple places. Seems we only need to schedule the check here all the time.

runners/samza/src/main/java/org/apache/beam/runners/samza/runtime/DoFnOp.java
```diff
@@ -237,9 +292,39 @@ private String getTimerStateId(DoFnSignature signature) {
     return builder.toString();
   }
 
+  private void attemptStartBundle() {
+    if (isBundleStarted.compareAndSet(false, true)) {
```
xinyuiscool: thread safety is still bit awkward here as if another thread invokes attemptStartBundle() in a different thread after the the line of finishBundle, the count and time can still be set in a non-consistent manner. I guess we don't allow multiple threads in the same task right now. But it can cause problems. Could you add a check in the translator to make sure we don't have task.max.concurrency bigger than 1 if bundling is enabled?

runners/samza/src/main/java/org/apache/beam/runners/samza/runtime/DoFnOp.java
```diff
@@ -280,9 +365,26 @@ public void processWatermark(Instant watermark, OpEmitter<OutT> emitter) {
...
+        // for batch mode, the max watermark should force the bundle to close
+        finishBundle(emitter);
+      }
+    }
+  }
+
   @Override
   public void processSideInput(
       String id, WindowedValue<? extends Iterable<?>> elements, OpEmitter<OutT> emitter) {
+    checkState(
```
xinyuiscool: Move this check to the translator so it can throw before running.

runners/samza/src/main/java/org/apache/beam/runners/samza/runtime/DoFnOp.java
```diff
@@ -308,6 +410,9 @@ public void processSideInput(
 
   @Override
   public void processSideInputWatermark(Instant watermark, OpEmitter<OutT> emitter) {
+    checkState(
```
xinyuiscool: Move this check to the translator so it can throw before running.

runners/samza/src/main/java/org/apache/beam/runners/samza/runtime/DoFnOp.java
```diff
@@ -317,7 +422,18 @@ public void processSideInputWatermark(Instant watermark, OpEmitter<OutT> emitter
...
-  public void processTimer(KeyedTimerData<Void> keyedTimerData) {
+  public void processTimer(KeyedTimerData<Void> keyedTimerData, OpEmitter<OutT> emitter) {
+    // this is internal timer in processing time to check whether a bundle should be closed
+    if (bundleCheckTimerId.equals(keyedTimerData.getTimerData().getTimerId())) {
+      attemptFinishBundle(emitter);
+      scheduleNextBundleCheck();
+      return;
+    }
+
+    checkState(
```
xinyuiscool: Move this check to the translator so it can throw before running.

We might need this in the long run. Please create a Beam ticket for this and add a comment to link the ticket here.

runners/samza/src/main/java/org/apache/beam/runners/samza/runtime/SamzaDoFnRunners.java
```diff
@@ -213,31 +218,65 @@ public void processElement(WindowedValue<InT> elem) {
...
-        }
+                BundleProgressHandler.ignored());
 
-        // RemoteBundle close blocks until all results are received
-        KV<String, FnOutT> result;
-        while ((result = outputQueue.poll()) != null) {
-          outputManager.output(
-              idToTupleTagMap.get(result.getKey()), (WindowedValue) result.getValue());
-        }
+        inputReceiver = Iterables.getOnlyElement(remoteBundle.getInputReceivers().values());
```
xinyuiscool: I guess this is to say we only support main output right now? Maybe worth a comment for this.

runners/samza/src/main/java/org/apache/beam/runners/samza/runtime/DoFnOp.java
lhaiesp: There are two models. 1) the current model is to regularly check in the background to see whether max bundling time is hit; this is only roughly accurate but more efficient 2) schedule a timer for EVERY bundle start. This is more accurate but we may end up scheduling too many timer when the QPS is too high. Let's discuss this

runners/samza/src/main/java/org/apache/beam/runners/samza/runtime/DoFnOp.java
lhaiesp: Make sense. Will do

runners/samza/src/main/java/org/apache/beam/runners/samza/runtime/SamzaDoFnRunners.java
lhaiesp: Igor already has a RB for side output. I think this is now addressed

lhaiesp:
Looking at the end to end test but it may be hard to just do one for bundling. Instead, I've been looking into validatesPortableRunner and testPipelineJar example from flink runner. The latter one seems more achievable at this point, while validatesPortableRunner seems more like the real long term solution.

runners/samza/src/main/java/org/apache/beam/runners/samza/translation/ParDoBoundMultiTranslator.java
```diff
@@ -76,11 +79,36 @@
     doFnInvokerRegistrar = invokerReg.hasNext() ? Iterators.getOnlyElement(invokerReg) : null;
   }
 
+  /*
+   * Perform some common pipeline option validation. Bundling related logic for now
+   */
+  private static void validatePipelineOption(
```
xinyuiscool: we actually have a SamzaPipelineOPtionsValidator. Can we move the logic there?

PR #9779, opened 2019-10-11
[BEAM-8388] Updating avro dependency from 1.8.2 to 1.9.1

realjordanna:

------------------------

--- | --- | --- | --- | ---


sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/utils/AvroUtils.java
```diff
@@ -75,8 +75,8 @@
   static {
     // This works around a bug in the Avro library (AVRO-1891) around SpecificRecord's handling
     // of DateTime types.
-    SpecificData.get().addLogicalTypeConversion(new TimeConversions.TimestampConversion());
-    GenericData.get().addLogicalTypeConversion(new TimeConversions.TimestampConversion());
+    SpecificData.get().addLogicalTypeConversion(new JodaTimeConversions.TimestampConversion());
+    GenericData.get().addLogicalTypeConversion(new JodaTimeConversions.TimestampConversion());
```
TheNeuralBit: This is necessary because `TimeConversions` converts to `java.time` now?

Maybe we need to consider adding support for `java.time` in `AvroSchemaProvider` now

TheNeuralBit:
At first I thought the test failures were flakes or checkstyle or something but it looks like it's a real failure in AvroSchemaTest. From [gradle scan](https://scans.gradle.com/s/kinsk2kpzcupa/console-log?task=:sdks:java:core:compileTestJava):

```
../sdks/java/core/src/test/java/org/apache/beam/sdk/schemas/AvroSchemaTest.java:274: error: incompatible types: org.joda.time.LocalDate cannot be converted to java.time.LocalDate
          DATE,
```

This seems to be failing to create an instance of the schema defined in [test.avsc](https://github.com/apache/beam/blob/master/sdks/java/core/src/test/avro/org/apache/beam/sdk/schemas/test.avsc#L14) because the date logical type maps to `java.time` now. Is there some way we can change this back to joda time?

realjordanna:
> At first I thought the test failures were flakes or checkstyle or something but it looks like it's a real failure in AvroSchemaTest. From [gradle scan](https://scans.gradle.com/s/kinsk2kpzcupa/console-log?task=:sdks:java:core:compileTestJava):
> 
> ```
> ../sdks/java/core/src/test/java/org/apache/beam/sdk/schemas/AvroSchemaTest.java:274: error: incompatible types: org.joda.time.LocalDate cannot be converted to java.time.LocalDate
>           DATE,
> ```
> 
> This seems to be failing to create an instance of the schema defined in [test.avsc](https://github.com/apache/beam/blob/master/sdks/java/core/src/test/avro/org/apache/beam/sdk/schemas/test.avsc#L14) because the date logical type maps to `java.time` now. Is there some way we can change this back to joda time?

Since joda time is deprecated in favor of java8 I doubt the avro folks would want to roll it back. I'm going to try some manual conversions between the two to get this passing again.

TheNeuralBit:
I didn't mean get Avro to roll it back, I was wondering if there's anything we can do to make the generated `TestAvro` class use joda time instead of java8 time, e.g. change test.avsc somehow, change some configuration, etc..

realjordanna:
@dkulp who did the joda/time migration on the avro package -> is it safe to only change the test? Will this expose users of GenericRecordBuilder to runtime exceptions if they still use joda time?

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/utils/AvroUtils.java
realjordanna: It was renamed to JodatimeConversions in https://github.com/apache/avro/pull/248 for "clarity".

It says it defaults to joda time, but that doesn't explain why it's forcing me to update the tests since I haven't changed this flag value.

TheNeuralBit:
It looks like the default was [changed](https://github.com/apache/avro/blob/release-1.9.1/lang/java/compiler/src/main/java/org/apache/avro/compiler/specific/SpecificCompiler.java#L121) before 1.9.1, and joda has since been removed: https://github.com/apache/avro/pull/631

We use a gradle plugin to compile test.avsc into TestAvro, configured [here](https://github.com/apache/beam/blob/master/buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy#L1535). From the README it looks like we can configure it to use JODA: https://github.com/davidmc24/gradle-avro-plugin#datetimelogicaltype

Hopefully that can fix these tests, but we should probably consider dropping joda as well.

TheNeuralBit:
I wrote a a [patch](https://github.com/TheNeuralBit/beam/commit/b0e2cf443d51192019b46fa33eb938b1ed40c2c0) to configure the plugin to use joda (branched before your changed to AvroSchemaTests), but it's not having any effect, TestAvro still uses java time. I tried setting it to a bad value just to see if it's being read and I do get an error:

```
❯ ./gradlew :sdks:java:core:test
Configuration on demand is an incubating feature.
> Task :sdks:java:core:generateTestAvroJava FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':sdks:java:core:generateTestAvroJava'.
> Invalid dateTimeLogicalType 'FOO'.  Value values are: [JODA, JSR310]
...
```

So it seems to be reading the value, but not using it. Not sure if this is a bug in the plugin or maybe my limited gradle knowledge is failing me.

iemejia:
[BEAM-7328](https://issues.apache.org/jira/browse/BEAM-7328) (the previous issue to migrate to Avro 1.9.x) ended up being 'stalled' for two reasons still worth to consider in this PR:

1. Potential compatibility issues for Beam Java users given that Avro is transitively 'leaked' as part of Core SDK public API. This is the same issue for any 'transitive' dependency we offer on Beam (e.g. Jackson), but there may be some minor issues for users because of Avro API changes, remember Avro minor versions are not backwards compatible code-wise (but they are format-wise). Anyway migrating is good for multiple other reasons like security fixes, but we probably should warn or discuss on the consequences of this on the ML.

2. Issues on target runners where Avro 1.9.x has not been downstreamed. More concretely Spark runner core module depends still on Avro 1.8.x (AFAIK Spark runner does not use Avro at all, but maybe Spark core does) so we should probably as part of this PR guarantee that all the Spark related tests pass too).

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/utils/AvroUtils.java
RyanSkraba: Hello!  Just for information, Avro 1.9.x supports `joda-time`  and `java.time`, with the strong recommendation that users start preferring the `java.time` classes.  Avro 1.10.x removes built-in support for joda. 

In any case, good news: AVRO-1891 has been fixed in Avro 1.9.x, so these lines can probably be removed!

RyanSkraba:
@iemejia 

> 2\. Spark runner core module depends still on Avro 1.8.x

Except where it still depends on 1.7.x  underneath it all... (See [Spark 2.3.3](https://mvnrepository.com/artifact/org.apache.spark/spark-core_2.11/2.3.3))

I think 1.7 and 1.8 are API compatible-ish enough, but it hasn't been exhaustively tested.  I ran into a problem with ParquetIO using a 1.8.x Avro API that required shading for older versions of spark (documented here: https://beam.apache.org/documentation/io/built-in/parquet/)

It might be worth checking this sooner rather than later, in case bumping Avro implies dropping some versions for Spark support!

TheNeuralBit:
Run Spark ValidatesRunner

kennknowles:
Seems like including Avro in the core SDK might be a mistake / legacy problem. We could potentially create new artifacts for particular versions so users can choose compatible pieces of Beam.

realjordanna:
It's certainly proving challenging to update. I don't disagree with you.

zeidoo:
Avro needs to be update if BEAM wants to support the new Confluent Schema Registry with Protobuf and JSON schemas (https://issues.apache.org/jira/browse/BEAM-9330), as the 5.5.x confluent libs use AVRO 1.9.

From reading this thread, I'm not exactly sure there's a clear plan around this. If there is, could someone tell me, I could implement it.

realjordanna:
Go for it - I got pretty stuck doing in, it was a non trivial update

zeidoo:
Hi everyone,

I think I managed to move (**not upgrade**) all the Avro logic from sdk.core to sdk.extensions.avro (right besides protobuf). I also updated all usages and most tests seem to pass, as seen below. The tests that depended on Avro have also been moved to sdk.extensions.avro. I had to also update some accessibility modifiers from protected/private to public, that might be something undesirable architecturally speaking. Really this is more of a first pass (ie: getting an idea of what needs to be done).

`./gradlew -p sdks/java spotlessApply check --exclude-task :sdks:java:extensions:ml:test --exclude-task :sdks:java:io:google-cloud-platform:test --exclude-task :sdks:java:io:jdbc:test --rerun-tasks`

`[... lots of output...]
BUILD SUCCESSFUL in 49m 45s
640 actionable tasks: 640 executed`

Here are the reasons I excluded some tasks:
- :sdks:java:extensions:ml:test --> no Google Computing Engine setup on my machine
- :sdks:java:io:google-cloud-platform:test --> no Google cloud account set up on my machine
- :sdks:java:io:jdbc:test  --> one test failing two dates are one day apart, no sure why, maybe due to my timezone?

Given I'm brand new to BEAM contributions and that Java isn't something I'm very good at, I'd kindly request that someone who has good experience in both to kindly spend some time with me doing a screen share so we can superficially review my changes, make sure that the tests really do pass and help me do the paperwork. (ie: PR, issue tracking, etc).

Thanks!

TheNeuralBit:
That's great thank you @zeidoo! What is the extent of our use of Avro in core Beam? Is it just AvroCoder and AvroSchema? That does seem to be the case based on the changes in this PR.

I think AvroCoder is used pretty commonly so if we move it out of core we should add something to CHANGES.md to make sure users know they'll need to add a new dependency. We may want to call the new extension avro-1.8 or similar so we can add a 1.9 option down the road.

As far as getting the change reviewed - I think the best bet would actually be to go ahead and put up a PR, you can mark it as a WIP if you don't feel its ready. From there the CI infrastructure will be able to run the tests that need a GCP account. I suspect you're right about the JDBC test failing due to a local timezone issue, it's probably worth filing a JIRA for that so we can fix the test.

zeidoo:
Sure thing! Here we go: https://github.com/apache/beam/pull/12748

PR #9783, opened 2019-10-14
[BEAM-8391] Update type id write methods in `AwsModule`

cowtowncoder:
So: (polymorphic) type id write methods were refactored in Jackson 2.9, and change here is to use new methods: these will work better with some data formats (xml, avro, for example). Deprecated methods would work fine for json in 2.x, but seems better to update.

R: @iemejia


cowtowncoder:
retest this please

cowtowncoder:
retest this please

cowtowncoder:
retest this please

iemejia:
Merged, thanks again. I just saw you are a jackson maintainer catching up downstream, makes sense now :). Never mind about the update suggestion if you don't have the time.

cowtowncoder:
@iemejia You are welcome! I could try 2.10 upgrade, time permitting. There shouldn't be much to do but would definitely want full regression test to be run just in case.

PR #9800, opened 2019-10-15
[BEAM-8403] Guard request id generation to prevent concurrent worker access

mxm:
There is a race condition in GrpcStateRequestHandler which surfaced after the
recent changes to process append/clear state request asynchronously. The race
condition can occur if multiple Runner workers process a transform with state
requests with the same SDK Harness. The SDK Harness has multiple workers threads
which can access the id generation code in parallel.

For example, this setup occurs with Flink when a TaskManager has multiple task
slots and two or more of those slots process the same stateful stage against an
SDK Harness.

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/runners/worker/sdk_worker.py
+++ b/sdks/python/apache_beam/runners/worker/sdk_worker.py
@@ -570,6 +570,7 @@ class GrpcStateHandler(object):
   _DONE = object()
 
   def __init__(self, state_stub):
+    self._lock = threading.Lock()
     self._state_stub = state_stub
     self._requests = queue.Queue()
     self._responses_by_id = {}
@@ -604,6 +605,7 @@ class GrpcStateHandler(object):
     def pull_responses():
       try:
         for response in responses:
+          # Popping an item from a dictionary is atomic in cPython
           future = self._responses_by_id.pop(response.id)
           future.set(response)
           if self._done:
@@ -643,7 +645,9 @@ class GrpcStateHandler(object):
   def _request(self, request):
     request.id = self._next_id()
     request.instruction_id = self._context.process_instruction_id
+    # Adding a new item to a dictionary is atomic in cPython
     self._responses_by_id[request.id] = future = _Future()
+    # Request queue is thread-safe
     self._requests.put(request)
     return future
 
@@ -662,8 +666,14 @@ class GrpcStateHandler(object):
       return response
 
   def _next_id(self):
-    self._last_id += 1
-    return str(self._last_id)
+    with self._lock:
+      # Use a lock here because this GrpcStateHandler is shared across all
+      # requests which have the same process bundle descriptor. State requests
+      # can concurrently access this section if a Runner uses threads / workers
+      # (aka "parallelism") to send data to this SdkHarness and its workers.
+      self._last_id += 1
+      request_id = self._last_id
+    return str(request_id)
 
 
 class CachingMaterializingStateHandler(object):


mxm:
Run Portable_Python PreCommit

lukecwik:
Run Python PreCommit

aaltay:
Change LGTM.

I would suggest against using the atomiclong dependency, it is stale and marked by the owner as do not use this.

Precommit failed for an unrelated issue. Related to BEAM-8194 and pypi being flaky. That would be addressed separately.

tweise:
@aaltay ignoring the specific dependency, what is the generally recommended approach for such situations? Explicit locking or are there other options?

tweise:
@mxm no occurrence of the error during testing with this change, we should be good to go

mxm:
Thanks for the reviews!

I doubt that using the atomic long (native) library will make any measurable difference. The reason is that in Python there can only ever be one thread active at a time (due to the Global Interpreter Lock (GIL)), so the lock basically comes for free. Only in the very rare case of a thread having to wait for the lock (which even in tests it was hard to reproduce), there will be a super minimal delay for switching to another thread to release the lock.

Generally speaking, we should be more concerned about the parallelism model of the Python harness. Effectively, we are scarifying performance by letting multiple Runner workers process data with a GIL-threaded Python harness. I'm not sure if you have done performance measurements but the throughput of a single SDK Harness with N workers vs N instances of an SDK Harness should be huge. The former being effectively single-threaded while the latter having N "real" threads.

mxm:
Run Python PreCommit

mxm:
Unrelated test failures for both testPy37Gcp and testPython37:

```
ERROR: test_concurrent_requests (apache_beam.runners.portability.artifact_service_test.ZipFileArtifactServiceTest)
  File "/home/jenkins/jenkins-slave/workspace/beam_PreCommit_Python_Phrase/src/sdks/python/test-suites/tox/py37/build/srcs/sdks/python/apache_beam/runners/portability/artifact_
service_test.py", line 215, in test_concurrent_requests
    _ = list(pool.map(check, range(100)))
  File "/usr/lib/python3.7/concurrent/futures/_base.py", line 586, in result_iterator
    yield fs.pop().result()
  File "/usr/lib/python3.7/concurrent/futures/_base.py", line 425, in result
    return self.__get_result()
  File "/usr/lib/python3.7/concurrent/futures/_base.py", line 384, in __get_result
    raise self._exception
  File "/usr/lib/python3.7/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/jenkins/jenkins-slave/workspace/beam_PreCommit_Python_Phrase/src/sdks/python/test-suites/tox/py37/build/srcs/sdks/python/apache_beam/runners/portability/artifact_
service_test.py", line 208, in check
    self._service, tokens[session(index)], name(index)))
  File "/home/jenkins/jenkins-slave/workspace/beam_PreCommit_Python_Phrase/src/sdks/python/test-suites/tox/py37/build/srcs/sdks/python/apache_beam/runners/portability/artifact_
service_test.py", line 73, in retrieve_artifact
    name=name)))
  File "/home/jenkins/jenkins-slave/workspace/beam_PreCommit_Python_Phrase/src/sdks/python/test-suites/tox/py37/build/srcs/sdks/python/apache_beam/runners/portability/artifact_
service_test.py", line 70, in <genexpr>
    return b''.join(chunk.data for chunk in retrieval_service.GetArtifact(
  File "/home/jenkins/jenkins-slave/workspace/beam_PreCommit_Python_Phrase/src/sdks/python/test-suites/tox/py37/build/srcs/sdks/python/apache_beam/runners/portability/artifact_
service.py", line 133, in GetArtifact
    chunk = fin.read(self._chunk_size)
  File "/usr/lib/python3.7/zipfile.py", line 899, in read
    data = self._read1(n)
  File "/usr/lib/python3.7/zipfile.py", line 989, in _read1
    self._update_crc(data)
  File "/usr/lib/python3.7/zipfile.py", line 917, in _update_crc
    raise BadZipFile("Bad CRC-32 for file %r" % self.name)
zipfile.BadZipFile: Bad CRC-32 for file '/3b2b55eb92de23535010b7ac80d553ec2d4bae872ac5606bc3042ce9313dff87/763ed77b36bcd01e3c2a2f8e0d615cfe3b1f3ad2d1f93b6948a11b9dd32e6d53'
```

tweise:
You are right that due to the GIL there is no real parallelism. That's why there is the option to use multiple SDK worker processes (configurable via pipeline option `sdk_worker_parallelism`).

As part of the verification for this change I found that using 2 worker processes instead of one leads to throughput increase of ~50%. It won't be 100% because Python is still going to switch threads while waiting for IO etc.

It is still desirable to have a proper abstraction for something as basic as an atomic counter. It results in code that is more readable and easier to understand, and helps avoiding issues such as this in first place. Perhaps that's something the Python SDK experts can consider?

PR #9757, opened 2019-10-10
[BEAM-8484] RowJsonDeserializer/ToJson cleanup

TheNeuralBit:
- Moves both `RowJsonDeserializer` and `RowJsonSerializer` into `RowJson` as suggested in https://github.com/apache/beam/pull/9732#pullrequestreview-298369721
- Passes along exception as suggested in https://github.com/apache/beam/pull/9744#discussion_r332642510
- Makes `ToJson#of` public and `new ToJson()` private

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

Lang | SDK | Apex | Dataflow | Flink | Gearpump | Samza | Spark
--- | --- | --- | --- | --- | --- | --- | ---
Go | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go/lastCompletedBuild/) | --- | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_VR_Flink/lastCompletedBuild/) | --- | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_VR_Spark/lastCompletedBuild/)
Java | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Apex/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Dataflow/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Flink/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Flink_Batch/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Flink_Streaming/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Gearpump/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Samza/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Spark/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Spark_Batch/lastCompletedBuild/)
[...]

PR #9806, opened 2019-10-16
[BEAM-8427] Create a table and a table provider for MongoDB

11moon11:
- Create a Table with read support (for now, write will be added in a separate PR)
- Add some tests

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/MongoDbTable.java
+++ b/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/MongoDbTable.java
@@ -1,0 +1,143 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.beam.sdk.extensions.sql.meta.provider.mongodb;
+
+import static org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Preconditions.checkArgument;
+
+import java.io.Serializable;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.apache.beam.sdk.annotations.Experimental;
+import org.apache.beam.sdk.extensions.sql.impl.BeamTableStatistics;
+import org.apache.beam.sdk.extensions.sql.meta.SchemaBaseBeamTable;
+import org.apache.beam.sdk.extensions.sql.meta.Table;
+import org.apache.beam.sdk.io.mongodb.MongoDbIO;
+import org.apache.beam.sdk.options.PipelineOptions;
+import org.apache.beam.sdk.schemas.Schema;
+import org.apache.beam.sdk.transforms.DoFn;
+import org.apache.beam.sdk.transforms.JsonToRow;
+import org.apache.beam.sdk.transforms.PTransform;
+import org.apache.beam.sdk.transforms.ParDo;
+import org.apache.beam.sdk.values.PBegin;
+import org.apache.beam.sdk.values.PCollection;
+import org.apache.beam.sdk.values.PCollection.IsBounded;
+import org.apache.beam.sdk.values.POutput;
+import org.apache.beam.sdk.values.Row;
+import org.apache.beam.vendor.calcite.v1_20_0.com.google.common.annotations.VisibleForTesting;
+import org.bson.Document;
+import org.bson.json.JsonMode;
+import org.bson.json.JsonWriterSettings;
+
+@Experimental
+public class MongoDbTable extends SchemaBaseBeamTable implements Serializable {
+  // Should match: mongodb://username:password@localhost:27017/database/collection
+  @VisibleForTesting
+  final Pattern locationPattern =
+      Pattern.compile(
+          "(?<credsHostPort>mongodb://(?<usernamePassword>.*(?<password>:.*)?@)?.+:\\d+)/(?<database>.+)/(?<collection>.+)");
+
+  @VisibleForTesting final String dbCollection;
+  @VisibleForTesting final String dbName;
+  @VisibleForTesting final String dbUri;
+
+  MongoDbTable(Table table) {
+    super(table.getSchema());
+
+    String location = table.getLocation();
+    Matcher matcher = locationPattern.matcher(location);
+    checkArgument(
+        matcher.matches(),
+        "MongoDb location must be in the following format: 'mongodb://(username:password@)?localhost:27017/database/collection'");
+    this.dbUri = matcher.group("credsHostPort"); // "mongodb://localhost:27017"
+    this.dbName = matcher.group("database");
+    this.dbCollection = matcher.group("collection");
+  }
+
+  @Override
+  public PCollection<Row> buildIOReader(PBegin begin) {
+    // Read MongoDb Documents
+    PCollection<Document> readDocuments =
+        MongoDbIO.read()
+            .withUri(dbUri)
+            .withDatabase(dbName)
+            .withCollection(dbCollection)
+            .expand(begin);
+
+    return readDocuments.apply(DocumentToRow.withSchema(getSchema()));
+  }
+
+  @Override
+  public POutput buildIOWriter(PCollection<Row> input) {
+    throw new UnsupportedOperationException("Writing to a MongoDB is not supported");
+  }
+
+  @Override
+  public IsBounded isBounded() {
+    return IsBounded.BOUNDED;
+  }
+
+  @Override
+  public BeamTableStatistics getTableStatistics(PipelineOptions options) {
+    long count =
+        MongoDbIO.read()
+            .withUri(dbUri)
+            .withDatabase(dbName)
+            .withCollection(dbCollection)
+            .getDocumentCount();
+
+    if (count < 0) {
+      return BeamTableStatistics.BOUNDED_UNKNOWN;
+    }
+
+    return BeamTableStatistics.createBoundedTableStatistics((double) count);
+  }
+
+  public static class DocumentToRow extends PTransform<PCollection<Document>, PCollection<Row>> {
+    private final Schema schema;
+
+    private DocumentToRow(Schema schema) {
+      this.schema = schema;
+    }
+
+    public static DocumentToRow withSchema(Schema schema) {
+      return new DocumentToRow(schema);
+    }
+
+    @Override
+    public PCollection<Row> expand(PCollection<Document> input) {
+      // TODO(BEAM-8498): figure out a way convert Document directly to Row.
+      return input
+          .apply("Convert Document to JSON", ParDo.of(new DocumentToJsonStringConverter()))
+          .apply("Transform JSON to Row", JsonToRow.withSchema(schema))
+          .setRowSchema(schema);
+    }
+
+    // TODO: add support for complex fields (May require modifying how Calcite parses nested
+    // fields).
+    @VisibleForTesting
+    static class DocumentToJsonStringConverter extends DoFn<Document, String> {
+      @DoFn.ProcessElement
+      public void processElement(ProcessContext context) {
+        context.output(
+            context
+                .element()
+                .toJson(JsonWriterSettings.builder().outputMode(JsonMode.RELAXED).build()));
+      }
+    }
+  }
+}
--- a/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/MongoDbTable.java
+++ b/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/MongoDbTable.java
@@ -62,7 +62,7 @@ public class MongoDbTable extends SchemaBaseBeamTable implements Serializable {
     Matcher matcher = locationPattern.matcher(location);
     checkArgument(
         matcher.matches(),
-        "MongoDb location must be in the following format: 'mongodb://(username:password@)?localhost:27017/database/collection'");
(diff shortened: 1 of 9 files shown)


11moon11:
R: @apilloud 
cc: @amaliujia

apilloud:
R: @TheNeuralBit 
Brian, can you take a look at this? You are probably more familiar with JSON stuff than I am.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/MongoDbTable.java
```diff
@@ -0,0 +1,118 @@
...
+  public BeamTableStatistics getTableStatistics(PipelineOptions options) {
+    return BeamTableStatistics.BOUNDED_UNKNOWN;
+  }
+
+  @VisibleForTesting
+  SingleOutput<Document, String> createParserParDo() {
+    return ParDo.of(new DocumentToJsonStringConverter());
+  }
+
+  // TODO: add support for complex fields.
```
TheNeuralBit: Can you add more detail about this? Is the limitation in `DocumentToJsonStringConverter`, or `JsonToRow`, or both? What happens when a complex field is encountered? A jira tracking this would be good as well.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/MongoDbTable.java
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  @Override
+  public IsBounded isBounded() {
+    return IsBounded.BOUNDED;
+  }
+
+  @Override
+  public BeamTableStatistics getTableStatistics(PipelineOptions options) {
+    return BeamTableStatistics.BOUNDED_UNKNOWN;
```
TheNeuralBit: Does MongoDB have any interface we could use to implement this? If so a TODO/Jira pointing in the right direction would be nice. If not, maybe add a comment stating why.

sdks/java/extensions/sql/src/test/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/MongoDbReadWriteIT.java
```diff
@@ -0,0 +1,195 @@
...
+            (float) 1.0,
+            1.0,
+            true,
+            "varchar",
+            Arrays.asList("123", "456"));
+
+    writePipeline
+        .apply(Create.of(testRow))
+        .setRowSchema(SOURCE_SCHEMA)
+        .apply("Transform Rows to JSON", new ToJson<Row>())
```
TheNeuralBit: If you merge/rebase past https://github.com/apache/beam/pull/9757 now this can (and should) be `ToJson.of()`

sdks/java/extensions/sql/src/test/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/MongoDbTableTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+  @Test
+  public void testDocumentToRowConverter() {
+    PCollection<String> output =
+        pipeline
+            .apply("Create document from JSON", Create.<Document>of(Document.parse(JSON_ROW)))
+            .apply("Decode document back into JSON", SQL_TABLE.createParserParDo());
+
+    // Make sure JSON was decoded correctly
+    PAssert.that(output).containsInAnyOrder(JSON_ROW);
```
TheNeuralBit: I think this is brittle, since the string comparison can fail if the keys are written out in a different order, and I don't think there's anything guaranteeing the order. Maybe instead verify that `output` and `JSON_ROW` can be parsed into equivalent rows?

This is a problem I keep running into as I'm writing tests with JSON, it would be nice to have some general purpose way to assert that two strings are equivalent JSON. Maybe we just need to write our own `Matcher` that parses to jackson nodes and compares. (You don't need to do that here unless you're so inclined, I'm just thinking out loud).

sdks/java/extensions/sql/src/test/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/MongoDbReadWriteIT.java
```diff
@@ -0,0 +1,195 @@
...
+
+  @AfterClass
+  public static void tearDown() throws Exception {
+    dropDatabase();
+  }
+
+  private static void dropDatabase() throws Exception {
+    new MongoClient(options.getMongoDBHostName())
+        .getDatabase(options.getMongoDBDatabaseName())
+        .drop();
```
TheNeuralBit: Shouldn't there be equivalent code to create this database? Or does that happen implicitly when you write to it? Maybe in the comment above state that a database called with the configured named will be created and deleted

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/package-info.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/** Table schema for MongoDb. */
+package org.apache.beam.sdk.extensions.sql.meta.provider.mongodb;
```
TheNeuralBit: I think these package-info files are supposed to have `@DefaultAnnotation(NonNull.class)` like [this](https://github.com/apache/beam/blob/928c4df93491f1d15ccfe36fe7887a27e4bdfbdc/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/pubsub/package-info.java#L20). (Although it looks like not all of them [do right now](https://github.com/apache/beam/blob/928c4df93491f1d15ccfe36fe7887a27e4bdfbdc/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/bigquery/package-info.java#L20)).

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/MongoDbTable.java
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  @VisibleForTesting
+  SingleOutput<Document, String> createParserParDo() {
+    return ParDo.of(new DocumentToJsonStringConverter());
+  }
+
+  // TODO: add support for complex fields.
+  @VisibleForTesting
+  static class DocumentToJsonStringConverter extends DoFn<Document, String> {
```
TheNeuralBit: It might cleaner to just have a separate `DocumentToRow extends PTransform<PCollection<Document>, PCollection<Row>>` that  expands to 'Document to JSON' + 'JsonToRow', then you can test that separately, rather than making this and `createParserParDo` visible for testing.

Bonus: then people using MongoDB but not SQL can use it :)

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/MongoDbTable.java
```diff
@@ -0,0 +1,118 @@
...
+  @VisibleForTesting final String dbUri;
+
+  MongoDbTable(Table table) {
+    super(table.getSchema());
+
+    String location = table.getLocation();
+    Matcher matcher = locationPattern.matcher(location);
+    checkArgument(
+        matcher.matches(),
+        "MongoDb location must be in the following format: 'mongodb://(username:password@)?localhost:27017/database/collection'");
```
TheNeuralBit: I think maybe this `?` is a copy-pasta error. That's not actually part of the format is it?

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/MongoDbTable.java
```diff
@@ -0,0 +1,118 @@
...
+    // Read MongoDb Documents
+    PCollection<Document> readDocuments =
+        MongoDbIO.read()
+            .withUri(dbUri)
+            .withDatabase(dbName)
+            .withCollection(dbCollection)
+            .expand(begin);
+
+    return readDocuments
+        // TODO: figure out a way convert Document directly to Row.
```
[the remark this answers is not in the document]
11moon11: Created a jira and updated a TODO comment.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/MongoDbTable.java
11moon11: After a quick investigation it appears that the main blocker to get this working with nested fields is that Calcite does not fully support nested fields. Nested fields get flattened, thus resulting in a different schema, making it very counter-intuitive for the users to handle.
As far as I can tell `DocumentToJsonStringConverter` and `JsonToRow` do support nested fields.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/MongoDbTable.java
11moon11: Right now MongoDbIO can only get an estimate Byte size via `MongoDbIO.BoundedMongoDbSource#getEstimatedSizeBytes`. I added a similar method to extract a `count` of Documents.

sdks/java/extensions/sql/src/test/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/MongoDbReadWriteIT.java
11moon11: Rebased and updated to use `ToJson.of()`. 
This part will change once `buildIOWriter` for `MongoDbTable` is implemented.

sdks/java/extensions/sql/src/test/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/MongoDbTableTest.java
11moon11: Makes sense, updated the test to check that the same `Row`s are produced. It would be a good thing to add at some point!

sdks/java/extensions/sql/src/test/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/MongoDbReadWriteIT.java
11moon11: I was using MongoDBIOIT as a reference:
https://github.com/apache/beam/blob/c2f0d282337f3ae0196a7717712396a5a41fdde1/sdks/java/io/mongodb/src/test/java/org/apache/beam/sdk/io/mongodb/MongoDBIOIT.java#L126-L135
Updated a comment to remove any uncertainty about the behavior.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/package-info.java
11moon11: Updated the package-info file, thanks for pointing this out!

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/MongoDbTable.java
11moon11: Great suggestion! Refactored.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/MongoDbTable.java
11moon11: I believe that `?` should be there, since `username:password@` are optional according to the JavaDoc here:
https://github.com/apache/beam/blob/c2f0d282337f3ae0196a7717712396a5a41fdde1/sdks/java/io/mongodb/src/main/java/org/apache/beam/sdk/io/mongodb/MongoDbIO.java#L208-L225
It also states that `For some authentication mechanisms, only the username is specified and the password is not, in which case the ":" after the username is left off as well`, updated the pattern to take that into account as well (and added tests to make sure it works as intended).

I decided to make a port number mandatory for now.

I feel like it would be a good idea to break that regular expression down into multiple logical layers (ex: `layer1` -> extract `uri`, `database`, and `collection`; `layer2` -> extract `username:password@`, `port`). The main benefit is improved readability, especially if we plan to add support for all features mentioned in the JavaDoc:
https://github.com/apache/beam/blob/c2f0d282337f3ae0196a7717712396a5a41fdde1/sdks/java/io/mongodb/src/main/java/org/apache/beam/sdk/io/mongodb/MongoDbIO.java#L202

11moon11:
Run Java PreCommit

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/MongoDbTable.java
TheNeuralBit: I guess I was expecting something like the MonogoDbIO string you linked, where the parens (or square brackets) indicate something is optional on their own. To me this looks like the the `?` could be part of the actual location string. This is just a bikeshed, I'm sure it will get the point across to any reasonable user either way, so I'm fine with leaving it.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/MongoDbTable.java
TheNeuralBit: Awesome, thanks!

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/MongoDbTable.java
11moon11: I see what you mean, replaced `?`with `[` `]` to be consistent with MongoDbIO format.

apilloud:
LGTM

PR #9811, opened 2019-10-16
[BEAM-8402] Create a class hierarchy to represent environments

violalyu:
As a first step towards making it possible to assign different environments to sections of a pipeline, we first need to expose environment classes to the pipeline API. Unlike PTransforms, PCollections, Coders, and Windowings, environments exists solely in the portability framework as protobuf objects. By creating a hierarchy of "native" classes that represent the various environment types – external, docker, process, etc – users will be able to instantiate these and assign them to parts of the pipeline. The assignment portion will be covered in a follow-up issue/PR.

Jira issue: https://issues.apache.org/jira/browse/BEAM-8402?jql=text%20~%20%22environment%20class%22

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/transforms/environments.py
+++ b/sdks/python/apache_beam/transforms/environments.py
@@ -1,0 +1,396 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""Environments concepts.
+
+For internal use only. No backwards compatibility guarantees."""
+
+from __future__ import absolute_import
+
+import json
+
+from google.protobuf import message
+
+from apache_beam.portability import common_urns
+from apache_beam.portability import python_urns
+from apache_beam.portability.api import beam_runner_api_pb2
+from apache_beam.portability.api import endpoints_pb2
+from apache_beam.utils import proto_utils
+
+__all__ = ['Environment',
+           'DockerEnvironment', 'ProcessEnvironment', 'ExternalEnvironment',
+           'EmbeddedPythonEnvironment', 'EmbeddedPythonGrpcEnvironment',
+           'SubprocessSDKEnvironment', 'RunnerAPIEnvironmentHolder']
+
+
+class Environment(object):
+  """Abstract base class for environments.
+
+  Represents a type and configuration of environment.
+  Each type of Environment should have a unique urn.
+
+  For internal use only. No backwards compatibility guarantees.
+  """
+
+  _known_urns = {}
+  _urn_to_env_cls = {}
+
+  def to_runner_api_parameter(self, context):
+    raise NotImplementedError
+
+  @classmethod
+  def register_urn(cls, urn, parameter_type, constructor=None):
+
+    def register(constructor):
+      if isinstance(constructor, type):
+        constructor.from_runner_api_parameter = register(
+            constructor.from_runner_api_parameter)
+        # register environment urn to environment class
+        cls._urn_to_env_cls[urn] = constructor
+        return constructor
+
+      else:
+        cls._known_urns[urn] = parameter_type, constructor
+        return staticmethod(constructor)
+
+    if constructor:
+      # Used as a statement.
+      register(constructor)
+    else:
+      # Used as a decorator.
+      return register
+
+  @classmethod
+  def get_env_cls_from_urn(cls, urn):
+    return cls._urn_to_env_cls[urn]
+
+  def to_runner_api(self, context):
+    urn, typed_param = self.to_runner_api_parameter(context)
+    return beam_runner_api_pb2.Environment(
+        urn=urn,
+        payload=typed_param.SerializeToString()
+        if isinstance(typed_param, message.Message)
+        else typed_param if (isinstance(typed_param, bytes) or
+                             typed_param is None)
+        else typed_param.encode('utf-8')
+    )
+
+  @classmethod
+  def from_runner_api(cls, proto, context):
+    if proto is None or not proto.urn:
+      return None
+    parameter_type, constructor = cls._known_urns[proto.urn]
+
+    try:
+      return constructor(
+          proto_utils.parse_Bytes(proto.payload, parameter_type),
+          context)
+    except Exception:
+      if context.allow_proto_holders:
+        return RunnerAPIEnvironmentHolder(proto)
+      raise
+
+  @classmethod
+  def from_options(cls, options):
+    """Creates an Environment object from PipelineOptions.
+
+    Args:
+      options: The PipelineOptions object.
+    """
+    raise NotImplementedError
+
+
+@Environment.register_urn(common_urns.environments.DOCKER.urn,
+                          beam_runner_api_pb2.DockerPayload)
+class DockerEnvironment(Environment):
+
+  def __init__(self, container_image=None):
+    from apache_beam.runners.portability.portable_runner import PortableRunner
+
+    if container_image:
+      self.container_image = container_image
+    else:
+      self.container_image = PortableRunner.default_docker_image()
+
+  def __eq__(self, other):
+    return self.__class__ == other.__class__ \
+           and self.container_image == other.container_image
+
+  def __ne__(self, other):
+    # TODO(BEAM-5949): Needed for Python 2 compatibility.
+    return not self == other
+
+  def __hash__(self):
+    return hash((self.__class__, self.container_image))
+
+  def __repr__(self):
+    return 'DockerEnvironment(container_image=%s)' % self.container_image
+
+  def to_runner_api_parameter(self, context):
+    return (common_urns.environments.DOCKER.urn,
+            beam_runner_api_pb2.DockerPayload(
+                container_image=self.container_image))
+
+  @staticmethod
+  def from_runner_api_parameter(payload, context):
+    return DockerEnvironment(container_image=payload.container_image)
+
+  @classmethod
+  def from_options(cls, options):
+    return cls(container_image=options.environment_config)
+
+
+@Environment.register_urn(common_urns.environments.PROCESS.urn,
+                          beam_runner_api_pb2.ProcessPayload)
+class ProcessEnvironment(Environment):
+
+  def __init__(self, command, os='', arch='', env=None):
+    self.command = command
+    self.os = os
+    self.arch = arch
+    self.env = env or {}
+
+  def __eq__(self, other):
+    return self.__class__ == other.__class__ \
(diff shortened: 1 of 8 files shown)


violalyu:
R: @robertwb 
R: @mxm 
R: @chadrik

sdks/python/apache_beam/transforms/environments.py
```diff
@@ -0,0 +1,388 @@
...
+                          beam_runner_api_pb2.DockerPayload)
+class DockerEnvironment(Environment):
+
+  def __init__(self, container_image=None):
+    from apache_beam.runners.portability.portable_runner import PortableRunner
+
+    if container_image:
+      self.container_image = container_image
+    else:
+      self.container_image = PortableRunner.default_docker_image()
```
chadrik: perhaps this should be allowed to be None, and choosing a default image if it is None can be handled inside the `PortableRunner`?

robertwb:
FWIW, I think we should re-think how what exactly environments mean. In particular, things like "in process" vs. "docker" are very runner-dependent, and I think should be handled as runner-substitutions of well-defined environments like java-2.16 which will make things a lot cleaner in the face of multi-language pipelines.

chadrik:
> FWIW, I think we should re-think how what exactly environments mean. In particular, things like "in process" vs. "docker" are very runner-dependent, and I think should be handled as runner-substitutions of well-defined environments like java-2.16 which will make things a lot cleaner in the face of multi-language pipelines.

Can you explain this a bit more? 

In your opinion, do you think that this re-thinking of environments should precede this PR?  The primary problem I see with that is that Viola and I don't have well-formed opinions on that topic, and making it a prerequisite for this would hold it up and the subsequent features we'd like to add.  My hope is that we can agree on 2 things:
1) that assigning an environment to parts of a pipeline is something that should be possible
2) that the general pattern of the `to_runner_api` / `from_runner_api` protocol for marshaling types between pipeline and portable-runner representations is viable for environments, and would remain viable through any future refinement of the standard environments (which I would assume would happen at the protobuf level and then be propagated out to dependent/adjacent APIs?)

mxm:
I think Robert's idea is to specify a set of required runtime dependencies, in the form of "Java with Beam 2.16.0 with libraries XY"; and then let Beam create an environment that fits these requirements. The question is whether we should really "bake in" the existing environments into the model itself. It does not have to be a contradiction because we could support "legacy" environments  (like the existing) and eventually replace them by the "smart" dynamic environments.

chadrik:
> I think Robert's idea is to specify a set of required runtime dependencies, in the form of "Java with Beam 2.16.0 with libraries XY"; and then let Beam create an environment that fits these requirements. 

That's how I was reading it as well.  While that's potentially more user friendly, it also seems quite challenging.  If I request "Java with Beam 2.16.0", and the runner happens to be using Java, it seems unsafe to assume that an embedded process is a valid replacement for docker, not in the least because I may have my own custom docker container for Beam-Java.   That said, it's still a very interesting idea.  Is there a ticket for this idea yet?

Since that idea seems like it could take awhile to be designed and executed, I'm hoping that we can move forward with this PR in the meantime.   I think that the progress that we make on this and the subsequent work on the semantics of assigning environments to transforms (possibly including [BEAM-7850](https://issues.apache.org/jira/browse/BEAM-7850)), would be useful in any case.  Any hey, we can always mark it as experimental :)

robertwb:
I bring this up because I think it's important to at least have some idea of the long-term story before we start introducing the explicit notion of environments to the public API. (I think there's still the open question on what the intended public API and usecase is, which would be informative in helping guide this story.)

One of the problems is this: if I have a pipeline with an environment "embedded" for a DoFn it now is tied to a particular runner (currently, either a Java runner or a Python runner). This is contrary to the idea that a pipeline proto should be runner-agnostic, and in practice comes up when trying to use an expansion service that asks for the embedded-java environment on the Python direct runner, or the Python embedded environment when running on Flink. And neither work on Dataflow. 

As for the safety of asking for "Java with Beam 2.16.0" and swapping I think that such an ask should be interpreted as the vanilla container--rather than building a custom container with the same name as the Java default one should be building a custom container with its own name. 

I agree that fleshing out the idea completely could take some time, but if we have the basic shape we can leave some details TBD while still having a forward-looking API .

violalyu:
> I bring this up because I think it's important to at least have some idea of the long-term story before we start introducing the explicit notion of environments to the public API. (I think there's still the open question on what the intended public API and usecase is, which would be informative in helping guide this story.)

Hi @robertwb, here's an example of how it might be used:
```
pipe = beam.Pipeline(options=options)

init = pipe | beam.Create(some_iterable)

with pipe.environment(DockerEnvironment(container_image='dummy_registry')):
	section_1 = (
		init
		| beam.ParDo(UserDefinedDoFn())
		| # arbitrary number of operations here
		| 'furtherProcess' >> beam.Map(lambda x: x)
	)

with pipe.environment(DockerEnvironment(container_image='another_registry')):
	section_2 = (
		section_1
		| beam.ParDo(AnotherUDF())
	)

with pipe.environment(ExternalEnvironment(url='localhost:50000')):
    section_3 = (
    	section_1
    	| beam.ParDo(ExternalUDF())
    )

...
```
where `pipe.environment` acts like a context manager which will assign the given environment to all transforms inside it.

robertwb:
Yes, this is mostly about the public API. This does create a module of environments, with several pre-defined, which gets us a step closer to just using these in the public API, which is not I think a scalable/portable way forward. We should at least add a caveat to that module that these are intended for internal use only.

mxm:
Fine. @violalyu  Could we rebase this and document that these are only intended for internal use?

violalyu:
@mxm updated!

sdks/python/apache_beam/transforms/environments.py
```diff
@@ -0,0 +1,394 @@
...
+           'EmbeddedPythonEnvironment', 'EmbeddedPythonGrpcEnvironment',
+           'SubprocessSDKEnvironment', 'RunnerAPIEnvironmentHolder']
+
+
+class Environment(object):
+  """Abstract base class for environments.
+
+  Represents a type and configuration of environment.
+  Each type of Environment should have a unique urn.
+  """
```
chadrik: let's add the internal warning here too.

violalyu: Updated!

violalyu:
Run Python PreCommit

mxm:

mxm:

mxm:
Run Python 3.5 Flink ValidatesRunner

mxm:
Run Python 3.5 PostCommit

mxm:
Run Python 2 PostCommit

mxm:
The metrics failures in https://builds.apache.org/job/beam_PostCommit_Py_VR_Dataflow_PR/156/ look unrelated.

sdks/python/apache_beam/transforms/environments.py
```diff
@@ -0,0 +1,396 @@
...
+    """
+    raise NotImplementedError
+
+
+@Environment.register_urn(common_urns.environments.DOCKER.urn,
+                          beam_runner_api_pb2.DockerPayload)
+class DockerEnvironment(Environment):
+
+  def __init__(self, container_image=None):
+    from apache_beam.runners.portability.portable_runner import PortableRunner
```
chamikaramj: Seems like this creates a new dependency between DataflowRunner and PortableRunner. Dataflow runner uses DockerEnvironment but there was no need for DataflowRunner to depend on the PortableRunner module before. Can we do this change without introducing the new dependency ? For example, can we refactor out default_docker_image() method to this module ?

mxm: +1 for moving `default_docker_image()` to this class.

mxm: @violalyu Do you think you could make this change?

violalyu: Thanks @mxm @chamikaramj ! Yeah I can make this change, should I open a new PR for this?

chamikaramj: Yes, please. Thanks.

violalyu: Opened the fix PR here: https://github.com/apache/beam/pull/10112

aaltay:
Beam 2.18 broke some users. Reverting back to 2.17 resolves the issues. I believe this PR is the root cause. I am trying to get more information to file a JIRA.

A stack trace I got looks like this:

```
  File "/tfx-src/tfx/orchestration/kubeflow/container_entrypoint.py", line 339, in <module>
    main()
  File "/tfx-src/tfx/orchestration/kubeflow/container_entrypoint.py", line 332, in main
    execution_info = launcher.launch()
  File "/tfx-src/tfx/orchestration/launcher/base_component_launcher.py", line 205, in launch
    execution_decision.exec_properties)
  File "/tfx-src/tfx/orchestration/launcher/in_process_component_launcher.py", line 67, in _run_executor
    executor.Do(input_dict, output_dict, exec_properties)
  File "/tfx-src/tfx/components/example_gen/base_example_gen_executor.py", line 230, in Do
    artifact_utils.get_split_uri(output_dict['examples'], split_name)))
  File "/opt/venv/lib/python3.6/site-packages/apache_beam/pipeline.py", line 436, in __exit__
    self.run().wait_until_finish()
  File "/opt/venv/lib/python3.6/site-packages/apache_beam/pipeline.py", line 416, in run
    self._options).run(False)
  File "/opt/venv/lib/python3.6/site-packages/apache_beam/pipeline.py", line 429, in run
    return self.runner.run_pipeline(self, self._options)
  File "/opt/venv/lib/python3.6/site-packages/apache_beam/runners/portability/fn_api_runner.py", line 389, in run_pipeline
    default_environment=self._default_environment))
  File "/opt/venv/lib/python3.6/site-packages/apache_beam/pipeline.py", line 637, in to_runner_api
    default_environment=default_environment)
  File "/opt/venv/lib/python3.6/site-packages/apache_beam/runners/pipeline_context.py", line 134, in __init__
    default_environment, label='default_environment')
  File "/opt/venv/lib/python3.6/site-packages/apache_beam/runners/pipeline_context.py", line 66, in get_id
    if obj not in self._obj_to_id:
TypeError: unhashable type: 'Environment'
```

I am suspecting this is related to the changes in the Environment class.

@violalyu @chadrik or other reviewers, do you know what this might be about?

chadrik:
The `default_environment` argument of `FnApiRunner` changed type from `beam_runner_api_pb2.Environment` to `environment.Environment` subclass.  This tfx code is still passing `beam_runner_api_pb2.Environment`:

https://github.com/tensorflow/tfx/blob/20666f4206b6741d55ed76d9194245600d433da9/tfx/components/base/base_executor.py#L116

Some possible solutions:

1. document the change
2. raise an informative `TypeError` earlier in the process
3. silently convert from `beam_runner_api_pb2.Environment` to `environment.Environment` subclass

I'd opt for solution 2.  Maybe 1 as well, though I'm not sure where this kind of changelog/announcement is supposed to go.

aaltay:
Thank you @chadrik. Option 2 sounds good and I can work with TFX team to update their code.

@udim This I believe came as a breaking change in 2.18 and we might want to add a note to the release notes about this. (Even add a note on 2.19 since that will also have the breaking change. @boyuanzz)

boyuanzz:
> Thank you @chadrik. Option 2 sounds good and I can work with TFX team to update their code.
> 
> @udim This I believe came as a breaking change in 2.18 and we might want to add a note to the release notes about this. (Even add a note on 2.19 since that will also have the breaking change. @boyuanzz)

Ack. Thanks for handling this!

mxm:
We believed this would not directly affect users. Turns out, it did. Next time, we should at least consider adding an entry to the release notes. Thank you for taking care of this.

PR #9820, opened 2019-10-17
[BEAM-8347]: Consistently advance UnboundedRabbitMqReader watermark

drobert:
Originally reported in [BEAM-8347](https://issues.apache.org/jira/browse/BEAM-8347?focusedCommentId=16953004&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16953004) and discussed on [StackOverflow](https://stackoverflow.com/a/58421999/796064) the UnboundedRabbitMqReader has surprising watermarking behavior.

Notably:
- Watermarks only advance if new messages come through. It's easily the case that no new messages come through for some long periods of time, preventing downstream windows from closing.
- The implementation of `getWatermark` runs counter to suggested documentation in the `UnboundedSource` documentation.

From the `getWatermark` javadoc:

> watermark] can be approximate. If records are read that violate this guarantee, they will be considered late, which will affect how they will be processed. ...
> 
> However, this value should be as late as possible. Downstream windows may not be able to close until this watermark passes their end.
> 
> For example, a source may know that the records it reads will be in timestamp order. In this case, the watermark can be the timestamp of the last record read. For a source that does not have natural timestamps, timestamps can be set to the time of reading, in which case the watermark is the current clock time.

From the above, a Rabbit/queue-based client should be able to safely expect records read are in timestamp order. This implies the watermark should be the timestamp of the last record read (or 'now' if the records have no timestamp).

The current implementation uses the following logic for setting the watermark on a checkpoint:
```java
        currentTimestamp = new Instant(delivery.getProperties().getTimestamp());
        if (currentTimestamp.isBefore(checkpointMark.oldestTimestamp)) {
          checkpointMark.oldestTimestamp = currentTimestamp;
        }
```
That is, the timestamp remains stuck at the *oldest* timestamp. The watermark effectively never advances until a call to `finalizeCheckpoint` is triggered, which often does not happen with low volumes of messages.

This PR makes the following significant changes to the implementation:
1. When no new messages are available, the watermark advances to wall clock time
2. When a new message is processed, watermark advances to delivery time (if present), else current time

Monotonically advancing the watermark should be safe, even with 'split'/multi-consumers running against a single queue. Because it's a queue, processing for any given shard should be in consistent FIFO order. The client implementation also preserves in-order semantics of fetched messages. 

This implementation is very similar to the implementation used in [AmqpIO](https://github.com/apache/beam/blob/master/sdks/java/io/amqp/src/main/java/org/apache/beam/sdk/io/amqp/AmqpIO.java#L308) and analogous to the KafkaIO [ProcessingTimePolicy](https://github.com/apache/beam/blob/master/sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/TimestampPolicyFactory.java#L112)

NOTE: I am having a hard time finding prior art on how to test an UnboundedSource.

R: @jbonofre 

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/io/rabbitmq/src/main/java/org/apache/beam/sdk/io/rabbitmq/RabbitMqIO.java
+++ b/sdks/java/io/rabbitmq/src/main/java/org/apache/beam/sdk/io/rabbitmq/RabbitMqIO.java
@@ -543,6 +543,8 @@ public class RabbitMqIO {
         // we consume message without autoAck (we want to do the ack ourselves)
         GetResponse delivery = channel.basicGet(queueName, false);
         if (delivery == null) {
+          currentRecordId = null;
+          current = null;
           checkpointMark.advanceWatermark(Instant.now());
           return false;
         }
--- a/sdks/java/io/rabbitmq/src/main/java/org/apache/beam/sdk/io/rabbitmq/RabbitMqIO.java
+++ b/sdks/java/io/rabbitmq/src/main/java/org/apache/beam/sdk/io/rabbitmq/RabbitMqIO.java
@@ -543,8 +543,6 @@ public class RabbitMqIO {
         // we consume message without autoAck (we want to do the ack ourselves)
         GetResponse delivery = channel.basicGet(queueName, false);
         if (delivery == null) {
-          currentRecordId = null;
-          current = null;
           checkpointMark.advanceWatermark(Instant.now());
           return false;
         }
--- a/sdks/java/io/rabbitmq/src/main/java/org/apache/beam/sdk/io/rabbitmq/RabbitMqIO.java
+++ b/sdks/java/io/rabbitmq/src/main/java/org/apache/beam/sdk/io/rabbitmq/RabbitMqIO.java
@@ -415,7 +415,7 @@ public class RabbitMqIO {
   private static class RabbitMQCheckpointMark
       implements UnboundedSource.CheckpointMark, Serializable {
     transient Channel channel;
-    Instant oldestTimestamp = Instant.now();
+    Instant latestTimestamp = Instant.now();
     final List<Long> sessionIds = new ArrayList<>();
 
     @Override
@@ -424,13 +424,15 @@ public class RabbitMqIO {
         channel.basicAck(sessionId, false);
       }
       channel.txCommit();
-      oldestTimestamp = Instant.now();
+      latestTimestamp = Instant.now();
       sessionIds.clear();
     }
   }
 
   private static class UnboundedRabbitMqReader
       extends UnboundedSource.UnboundedReader<RabbitMqMessage> {
+    private static final IDLE_WATERMARK_DELTA = Duration.standardSeconds(2);
+
     private final RabbitMQSource source;
 
     private RabbitMqMessage current;
@@ -449,7 +451,7 @@ public class RabbitMqIO {
 
     @Override
     public Instant getWatermark() {
-      return checkpointMark.oldestTimestamp;
+      return checkpointMark.latestTimestamp;
     }
 
     @Override
@@ -530,6 +532,10 @@ public class RabbitMqIO {
         // we consume message without autoAck (we want to do the ack ourselves)
         GetResponse delivery = channel.basicGet(queueName, false);
         if (delivery == null) {
+          Instant idleWatermark = Instant.now().minus(IDLE_WATERMARK_DELTA);
+          if (idleWatermark.isAfter(checkpointMark.latestTimestamp)) {
+            checkpointMark.latestTimestamp = idleWatermark;
+          }
           return false;
         }
         if (source.spec.useCorrelationId()) {
@@ -546,8 +552,8 @@ public class RabbitMqIO {
 
         current = new RabbitMqMessage(source.spec.routingKey(), delivery);
         currentTimestamp = new Instant(delivery.getProps().getTimestamp());
-        if (currentTimestamp.isBefore(checkpointMark.oldestTimestamp)) {
-          checkpointMark.oldestTimestamp = currentTimestamp;
+        if (currentTimestamp.isAfter(checkpointMark.latestTimestamp)) {
+          checkpointMark.latestTimestamp = currentTimestamp;
         }
       } catch (IOException e) {
         throw e;
--- a/sdks/java/io/rabbitmq/src/main/java/org/apache/beam/sdk/io/rabbitmq/RabbitMqIO.java
+++ b/sdks/java/io/rabbitmq/src/main/java/org/apache/beam/sdk/io/rabbitmq/RabbitMqIO.java
@@ -32,6 +32,7 @@ import java.nio.charset.StandardCharsets;
 import java.security.KeyManagementException;
 import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
+import java.util.Date;
 import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.concurrent.TimeoutException;
@@ -418,6 +419,18 @@ public class RabbitMqIO {
     Instant latestTimestamp = Instant.now();
     final List<Long> sessionIds = new ArrayList<>();
 
+    /**
+     * Advances the watermark to the provided time, provided said time is after the current
+     * watermark. If the provided time is before the latest, this function no-ops.
+     *
+     * @param time The time to advance the watermark to
+     */
+    public void advanceWatermark(Instant time) {
+      if (time.isAfter(latestTimestamp)) {
+        latestTimestamp = time;
+      }
+    }
+
     @Override
     public void finalizeCheckpoint() throws IOException {
       for (Long sessionId : sessionIds) {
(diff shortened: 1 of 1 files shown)


sdks/java/io/rabbitmq/src/main/java/org/apache/beam/sdk/io/rabbitmq/RabbitMqIO.java
```diff
@@ -376,16 +389,11 @@ public void finalizeCheckpoint() throws IOException {
       this.source = source;
       this.current = null;
       this.checkpointMark = checkpointMark != null ? checkpointMark : new RabbitMQCheckpointMark();
-      try {
-        connectionHandler = new ConnectionHandler(source.spec.uri());
```
drobert: Fixing a separate bug here. Originally, the constructor created the `connectionHandler` instance, which was closed inside `close()`. However `start()` created one as well, but masked the member variable with a local instance, so the channel/connection actually used for processing was not the one managed by the lifecycle of the class.

This is a semantic difference though, as now validation of the connection handler happens during `start()` rather than during construction. If there's a strong preference, I can restore this and instead modify `start()` to use the existing connection handler

sdks/java/io/rabbitmq/src/main/java/org/apache/beam/sdk/io/rabbitmq/RabbitMqIO.java
```diff
@@ -479,10 +488,10 @@ public boolean advance() throws IOException {
         checkpointMark.sessionIds.add(deliveryTag);
 
         current = new RabbitMqMessage(source.spec.routingKey(), delivery);
-        currentTimestamp = new Instant(delivery.getProperties().getTimestamp());
-        if (currentTimestamp.isBefore(checkpointMark.oldestTimestamp)) {
-          checkpointMark.oldestTimestamp = currentTimestamp;
-        }
+        Date deliveryTimestamp = delivery.getProperties().getTimestamp();
```
drobert: The meat of the change. Note I'm having trouble finding any documentation on exactly how `getProperties().getTimestamp()` is populated. I'm falling back to `Instant.now()` as a safeguard (which should actually be safe).

drobert:
I'd love to get some eyes and feedback on this before the 2.17.0 release is cut.
R: @jbonofre (you're listed in OWNERS)
R: @iemejia (you approved my last PR)

Thanks.

kennknowles:
Even though 2.17.0 has been cut, there is still time to cherrypick. Sorry for the large review delay.

sdks/java/io/rabbitmq/src/main/java/org/apache/beam/sdk/io/rabbitmq/RabbitMqIO.java
kennknowles: My impression is that your change here is an improvement.

sdks/java/io/rabbitmq/src/main/java/org/apache/beam/sdk/io/rabbitmq/RabbitMqIO.java
```diff
@@ -376,16 +389,11 @@ public void finalizeCheckpoint() throws IOException {
...
-        connectionHandler = new ConnectionHandler(source.spec.uri());
-      } catch (Exception e) {
-        throw new IOException(e);
-      }
     }
 
     @Override
     public Instant getWatermark() {
-      return checkpointMark.oldestTimestamp;
+      return checkpointMark.latestTimestamp;
```
kennknowles: Ideally, the watermark would be informed by the RabbitMQ broker's backlog. The goal is for the watermark to be a good prediction of the fact "no future records will have timestamps before the watermark's timestamp".

My impression is that you have made the watermark advance more reliably, in good ways, but will perhaps result in a large amount of data being marked late. What do you think?

drobert: I'm not sure I follow. 

>  The goal is for the watermark to be a good prediction of the fact "no future records will have timestamps before the watermark's timestamp".

In this implementation, the timestamp is effectively processing time. The timestamp is the delivery timestamp (which is how this was implemented even before this PR), which I believe is when the consumer picked up the message, not when it was enqueued. The queue is FIFO (we should probably add explicit documentation that priority queues are unsupported).

As noted in the PR description, the AmqpIO and certain Kafka timestamp watermarking implementations directly use 'now', so there's prior art here.

> when there's not a lot of messages (but in that case, we can imagine to update the watermark before the finalize checkpoint

This use case is what prompted this PR originally. 

> when we use an existing exchange (instead of creating one)

I don't see how this is relevant. Messages are pulled from a single queue, regardless of how the exchange was declared. 

> As it depends of the use case

drobert:
> Instead of changing the overall watermark advance behavior, it would make sense to provide for user to define the way to update the watermark.

I have a few concerns here, though this idea did cross my mind.

1. I'm having a hard time envisioning another reasonable watermarking strategy. The only real context is 'delivery time', which as noted will be FIFO, so I'm not sure what situation would cause the strategy in this PR to be inappropriate
2. Pluggable strategies feel like they ought to be a separate PR. It's a larger problem, and I'd set out to solve the more immediate, which is that the current strategy fails with low message counts.

jbonofre:
@drobert yeah, I agree. And actually, I also had hard times to think about an efficient watermark strategy in these IOs (jms, amqp, rabbitmq, mqtt, ...).  I used the same approach in these IOs: the watermark is the oldest timestamp before finalizing the checkpoint. For instance, in JmsIO checkpoint, when the checkpoint is finalized I do:

```
        message.acknowledge();
        Instant currentMessageTimestamp = new Instant(message.getJMSTimestamp());
        snapshot.updateOldestPendingTimestampIf(currentMessageTimestamp, Instant::isAfter);
```

So, the watermark advances with the oldest pending message timestamp.

The finalize checkpoint happens when the bundle is completed, so, I think it's consistent.

Right ?

drobert:
@jkff I looked at the PubSub implementation as well and opted against it *for now* for a few reasons:
1. It has an open ticket for the watermark not advancing with low volumes (https://issues.apache.org/jira/browse/BEAM-7322) which is the issue I set out to solve here for rabbit
2. It's fairly complicated relative to what I have here for rabbit. Given the difficulties in testing watermarking for an unbounded source, I didn't relish making this a blocker.
3. Rabbit has different semantics than pubsub. I actually think the processing time watermark is correct and simple for rabbit.

I'm very open to exploring better watermarking approaches in the future, but I think the current one is broken and I'd like to get a fix in first and improvements to it in second, if that makes sense.

jkff:
@drobert I totally agree that it makes sense to do a stop-the-bleeding fix first. Watermark logic in unbounded sources that don't have good native watermarks has been an issue for a very long time; if one were to attack this problem in a generic way, it would be best to start with a design and discuss it on dev@ - certainly outside the scope of this PR.

drobert:
Quick note: I may be seeing a problem with this implementation locally. This isn't mergeable due to a need to rebase anyway, but there may be an issue updating the watermark when no messages come in (DirectRunner reports "Processing stuck in step (read from rabbit) for at least 05m00s without outputting or completing in state process"). Looking.

drobert:
Update: after running some tests, the migration to the pull-based API in https://github.com/apache/beam/pull/9900 proves to be unsafe for this use case. I misread or misunderstood the semantics but it seems `channel.basicGet` blocks indefinitely, as opposed to returning immediately if there are no messages. This means if no new messages come in for a long period of time, the Runner will detect that the reader is 'stuck' since it's been on a single blocking call for so many minutes. Basically, we need to revert to a push-based-but-blocking API for this.

*However* that is a separate issue from the watermarks advancing. I don't think that broken merge should block this PR once rebased. I'll have a separate PR to fix the other up shortly.

sdks/java/io/rabbitmq/src/main/java/org/apache/beam/sdk/io/rabbitmq/RabbitMqIO.java
```diff
@@ -530,6 +543,10 @@ public boolean advance() throws IOException {
         // we consume message without autoAck (we want to do the ack ourselves)
         GetResponse delivery = channel.basicGet(queueName, false);
         if (delivery == null) {
+          current = null;
```
drobert: Notable change since the last round of reviews: The documentation for `Source`'s `getCurrent`, `getCurrentRecordId` and `getCurrentTimestamp` include the following:

>      * @throws NoSuchElementException if the reader is at the beginning of the input and {@link
>     *     #start} or {@link #advance} wasn't called, or if the last {@link #start} or {@link
>     *     #advance} returned {@code false}.

This states that if `advance()` returns `false`, subsequent calls to these should throw `NoSuchElementException`. As such, I'm explicitly setting these to `null`.

drobert:
@jkff This has been rebased and should be ready to merge.

I did make one notable change since the last review which is I now explicit set `current`, `currentTimestamp` and `currentRecordId` to `null` per this javadoc in `Source` for each corresponding getter:

```java
@throws NoSuchElementException 
   ... if the last {@link #start} or {@link#advance} returned @code false}.
```

jbonofre:
@jkff sure, I do a final pass and I merge ;) Thanks !

PR #9210, opened 2019-07-31
[BEAM-7832] Add ZetaSQL as a dialect in BeamSQL

amaliujia:
Support ZetaSQL(https://github.com/google/zetasql) as a dialect in BeamSQL.

------------------------

Thank you for your contribution! Follow this checklist to help us incorporate your contribution quickly and easily:

 - [x] [**Choose reviewer(s)**](https://beam.apache.org/contribute/#make-your-change) and mention them in a comment (`R: @username`).
 - [ ] Format the pull request title like `[BEAM-XXX] Fixes bug in ApproximateQuantiles`, where you replace `BEAM-XXX` with the appropriate JIRA issue, if applicable. This will automatically link the pull request to the issue.
 - [ ] If this contribution is large, please file an Apache [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

Lang | SDK | Apex | Dataflow | Flink | Gearpump | Samza | Spark
--- | --- | --- | --- | --- | --- | --- | ---
Go | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go/lastCompletedBuild/) | --- | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_VR_Flink/lastCompletedBuild/) | --- | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_VR_Spark/lastCompletedBuild/)
[...]

PR #9828, opened 2019-10-18
Test ZetaSQLPlanner in DataCatalogBigQueryIT

amaliujia:

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/extensions/sql/datacatalog/build.gradle
+++ b/sdks/java/extensions/sql/datacatalog/build.gradle
@@ -26,7 +26,8 @@ dependencies {
   compile library.java.grpc_google_cloud_datacatalog_v1beta1
   compile library.java.proto_google_cloud_datacatalog_v1beta1
   provided project(":sdks:java:extensions:sql")
-
+  provided project(":sdks:java:extensions:sql:zetasql")
+  
   // For Data Catalog GRPC client
   provided library.java.grpc_all
   provided library.java.google_auth_library_oauth2_http
--- a/sdks/java/extensions/sql/datacatalog/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/datacatalog/DataCatalogTableProvider.java
+++ b/sdks/java/extensions/sql/datacatalog/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/datacatalog/DataCatalogTableProvider.java
@@ -98,9 +98,8 @@ public class DataCatalogTableProvider extends FullNameTableProvider {
   }
 
   @Override
-  public @Nullable Table getTable(String tableNamePart) {
-    throw new UnsupportedOperationException(
-        "Loading a table by partial name '" + tableNamePart + "' is unsupported");
+  public @Nullable Table getTable(String tableName) {
+    return loadTable(tableName);
   }
 
   @Override
--- a/sdks/java/extensions/sql/datacatalog/src/test/java/org/apache/beam/sdk/extensions/sql/meta/provider/datacatalog/DataCatalogBigQueryIT.java
+++ b/sdks/java/extensions/sql/datacatalog/src/test/java/org/apache/beam/sdk/extensions/sql/meta/provider/datacatalog/DataCatalogBigQueryIT.java
@@ -25,6 +25,7 @@ import com.google.api.services.bigquery.model.TableReference;
 import com.google.api.services.bigquery.model.TableRow;
 import com.google.api.services.bigquery.model.TableSchema;
 import org.apache.beam.sdk.extensions.sql.SqlTransform;
+import org.apache.beam.sdk.extensions.sql.impl.BeamSqlPipelineOptions;
 import org.apache.beam.sdk.io.gcp.bigquery.BigQueryIO;
 import org.apache.beam.sdk.io.gcp.bigquery.TableRowJsonCoder;
 import org.apache.beam.sdk.io.gcp.bigquery.TestBigQuery;
@@ -64,6 +65,11 @@ public class DataCatalogBigQueryIT {
         String.format(
             "bigquery.`table`.`%s`.`%s`.`%s`",
             bqTable.getProjectId(), bqTable.getDatasetId(), bqTable.getTableId());
+
+    readPipeline
+        .getOptions()
+        .as(BeamSqlPipelineOptions.class)
+        .setPlannerName("org.apache.beam.sdk.extensions.sql.zetasql.ZetaSQLQueryPlanner");
 
     PCollection<Row> result =
         readPipeline.apply(
--- a/sdks/java/extensions/sql/zetasql/src/main/java/org/apache/beam/sdk/extensions/sql/zetasql/ZetaSQLQueryPlanner.java
+++ b/sdks/java/extensions/sql/zetasql/src/main/java/org/apache/beam/sdk/extensions/sql/zetasql/ZetaSQLQueryPlanner.java
@@ -129,11 +129,16 @@ public class ZetaSQLQueryPlanner implements QueryPlanner {
     final SqlOperatorTable opTab0 =
         connection.config().fun(SqlOperatorTable.class, SqlStdOperatorTable.instance());
 
+    Object[] contexts =
+        org.apache.beam.vendor.calcite.v1_20_0.com.google.common.collect.ImmutableList.of(
+                connection.config(), TableResolutionContext.joinCompoundIds("datacatalog"))
+            .toArray();
+
     return Frameworks.newConfigBuilder()
         .parserConfig(parserConfig.build())
         .defaultSchema(defaultSchema)
         .traitDefs(traitDefs)
-        .context(Contexts.of(connection.config()))
+        .context(Contexts.of(contexts))
         .ruleSets(ruleSets)
         .costFactory(null)
         .typeSystem(connection.getTypeFactory().getTypeSystem())


amaliujia:
Run SQL PostCommit

amaliujia:
Run SQL PostCommit

amaliujia:
@kennknowles

sdks/java/extensions/sql/datacatalog/src/test/java/org/apache/beam/sdk/extensions/sql/meta/provider/datacatalog/DataCatalogBigQueryIT.java
```diff
@@ -65,6 +66,11 @@ public void testReadWrite() throws Exception {
             "bigquery.`table`.`%s`.`%s`.`%s`",
             bqTable.getProjectId(), bqTable.getDatasetId(), bqTable.getTableId());
 
+    readPipeline
```
kennknowles: Let us keep tests for both dialects. It is fine to copy/paste duplicate or break into a separate method with two `@Test` calls to it.

sdks/java/extensions/sql/datacatalog/build.gradle
```diff
@@ -26,7 +26,8 @@ dependencies {
   compile library.java.grpc_google_cloud_datacatalog_v1beta1
   compile library.java.proto_google_cloud_datacatalog_v1beta1
   provided project(":sdks:java:extensions:sql")
-
+  provided project(":sdks:java:extensions:sql:zetasql")
```
kennknowles: This should be `testRuntime` since it is not required by the main code.

sdks/java/extensions/sql/datacatalog/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/datacatalog/DataCatalogTableProvider.java
```diff
@@ -98,9 +98,8 @@ public void dropTable(String tableName) {
   }
 
   @Override
-  public @Nullable Table getTable(String tableNamePart) {
-    throw new UnsupportedOperationException(
```
kennknowles: Where does this exception get swallowed and turned into a null? Did you figure that out?

sdks/java/extensions/sql/zetasql/src/main/java/org/apache/beam/sdk/extensions/sql/zetasql/ZetaSQLQueryPlanner.java
```diff
@@ -129,11 +129,16 @@ private FrameworkConfig defaultConfig(JdbcConnection connection, RuleSet[] ruleS
     final SqlOperatorTable opTab0 =
         connection.config().fun(SqlOperatorTable.class, SqlStdOperatorTable.instance());
 
+    Object[] contexts =
+        org.apache.beam.vendor.calcite.v1_20_0.com.google.common.collect.ImmutableList.of(
+                connection.config(), TableResolutionContext.joinCompoundIds("datacatalog"))
```
kennknowles: This should not be necessary. The pipeline registers "datacatalog" as the default table provider.

sdks/java/extensions/sql/datacatalog/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/datacatalog/DataCatalogTableProvider.java
```diff
@@ -98,9 +98,8 @@ public void dropTable(String tableName) {
...
-        "Loading a table by partial name '" + tableNamePart + "' is unsupported");
+  public @Nullable Table getTable(String tableName) {
+    return loadTable(tableName);
```
kennknowles: This should always fail to load, since it will only be called with the last component of a table name. If this works, then I think the problem is that the ZetaSQLPlanner needs to insert the components of the SQL name, splitting them, instead of one string.

sdks/java/extensions/sql/datacatalog/src/test/java/org/apache/beam/sdk/extensions/sql/meta/provider/datacatalog/DataCatalogBigQueryIT.java
amaliujia: I would suggest a to merge this PR firs than keeping both tests. It will be a following work to have tests for both dialects.

sdks/java/extensions/sql/datacatalog/build.gradle
amaliujia: This can be a future improvement.

sdks/java/extensions/sql/datacatalog/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/datacatalog/DataCatalogTableProvider.java
amaliujia: Yes. See next comment.

sdks/java/extensions/sql/datacatalog/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/datacatalog/DataCatalogTableProvider.java
kennknowles: I dug through the code, and I think what is required is to replace "JOIN_INTO_COMPOUND_ID" with identity function here: https://github.com/apache/beam/blob/master/sdks/java/extensions/sql/zetasql/src/main/java/org/apache/beam/sdk/extensions/sql/zetasql/TableResolutionContext.java#L44

sdks/java/extensions/sql/zetasql/src/main/java/org/apache/beam/sdk/extensions/sql/zetasql/ZetaSQLQueryPlanner.java
amaliujia: This line is not to register default table provider. Instead, it's used to load table from "datacatalog". Basically this is a quick fix so I won't expect we have a sold implementation at this moment.

sdks/java/extensions/sql/datacatalog/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/datacatalog/DataCatalogTableProvider.java
amaliujia: I think your understanding is wrong. Apparently it is the whole string passed in to called load table.

amaliujia: My last comment was to address the first comment. 

Regarding to the second comment. @kennknowles, if you know how to fix this problem, can you send a PR?

apilloud: I think this function was correct before and is now wrong.

apilloud:
The semantics of `getTable()` are well defined in both [Calcite](https://github.com/apache/calcite/blob/3ba81e2fda4a66fcc6acdd24a8688c8292bbfcff/core/src/main/java/org/apache/calcite/schema/Schema.java) and [Beam](https://github.com/apache/beam/blob/77f047bd2b166b2c318425b5752b005337633c83/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/FullNameTableProvider.java) as only taking the table name and not a fully qualified path. This changes that behavior to work around a bug. We should fix the root cause, not change the semantics of the API.

amaliujia:
We should have a well documented design on our table naming resolution along with API design, which is a to-do work.

kennknowles:
I think these points are not as mutually-exclusive as it may appear. Currently the protocol as far as I have grokked it is to call `getSubProvider` until it returns `null`, and then call `getTable` with the bit for which it returned `null`.

I don't like this protocol. It seems fragile and it also does not ensure that the entire list of qualifications on the names has been consumed. But in my conversations with Anton, he mentioned that this is what Calcite expects so we are not able to make a better protocol. Though Calcite is unaware of `TableProvider` so it must be via some other interface.

Notably, if you try to resolve `a.b.c` (which parses as `["a", "b", "c"]`) and `getSubProvider("b")` returns `null` then what should you do? It is a bug in the table provider. As I understand it, the ZetaSQLPlanner does a fairly reasonable thing and concatenates the remainder and calls `getTable("b.c")`. This allows a `TableProvider` to own a whole prefix of the fully qualified name. Instead, it might uncover the bug if the ZetaSQLPlanner threw an error: `getSubProvider not found for namespace "b" for TableProvider <current provider>`.

Why is the `TableProvider` returning `null`? Is it because the `DataCatalogTableProvider` needs to have all the names up front and it is not receiving all of them?

apilloud:
Under the calcite API semantics if `getSubProvider` returns `null` and not all the tokens have been consumed then the subschema (and therefore table) doesn't exist. Under the new API established by this PR, it would seem `getSubProvider` is not longer needed and `.` is now a reserved character in table names. It turns out the DataCatalogTableProvider (and other `FullNameTableProvider` classes) have a `getTableByFullName` call that handles  your concern with the calcite APIs, is that something you could use instead?

To answer the last question, `FullNameTableProvider` requires a call to `registerKnownTableNames` to look up tables using the old Calcite APIs. This is a workaround for the inability to lazily load subschemas.

In this PR is a significant change to the semantics of an API. The title doesn't reflect this, there is no JIRA, no discussion, and there are a number of critical issues that are being deferred to followup PRs. There were no failing tests in Beam that would justify the abnormal approach taken here and this bug actually existed before the ZetaSQL dialect went into Beam, so it would seem if we need an emergency fix the correct approach would be to roll back #9210.

The next Beam release isn't scheduled to be cut until Wednesday. Hopefully we can get the right fix in before then?

kennknowles:
I do not agree that this PR changes API semantics. It changes `DataCatalogTableProvider` and `ZetaSQLQueryPlanner`. Neither of these own any API semantics.

Your argument convinces me of something, though: this PR introduces incorrect behavior to `DataCatalogTableProvider` to compensate for incorrect clients. Let's definitely fix that ASAP.

My point is that "if `getSubProvider` returns `null` and not all tokens are consumed" is a fragile contract prerequisite. My understanding is that clients may be somewhat oblivious to whether or not all tokens have been consumed / they may not have access to the full list of tokens. The comments on `FullNameTableProvider` imply a protocol where a client is expected to always ask for a subprovider and only ask for a table if no subprovider is returned. This may be compensation for now-obsolete limitations in Data Catalog. In the world you posit where one can know if there are any further tokens, this exchange is unnecessary. The client knows whether to call `getSubProvider` and when to call `getTable`. I like that world; it is much simpler. Can we move to it immediately?

Minor nit: `TableProvider` is not a Calcite concept. That is our proxy for Calcite's `Schema`, am I wrong? And definitely `.` is not a reserved character. But rather, a client of `TableProvider` (specifically `DataCatalogTableProvider`) that concatenates the tail of a fully qualified name is enabled. As you point out, it should not be.

sdks/java/extensions/sql/datacatalog/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/datacatalog/DataCatalogTableProvider.java
kennknowles: Agreed.

amaliujia:
This is PR is actually a partial rollback of #9353(that PR was long time ago so cannot auto-rollback). 

I am not know what design decision was made in the past. If you guys still think we should use that design, the right fix is to fix #9353 for sure. My fix can be considered as a temporal fix.

sdks/java/extensions/sql/datacatalog/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/datacatalog/DataCatalogTableProvider.java
kennknowles: This line should not be changed.

sdks/java/extensions/sql/zetasql/src/main/java/org/apache/beam/sdk/extensions/sql/zetasql/ZetaSQLQueryPlanner.java
kennknowles: And this line.

PR #9843, opened 2019-10-21
[BEAM-4775] Converting MonitoringInfos to MetricResults in PortableRunner

kamilwu:
This PR enables users to retrieve metrics from portable runners via query() call.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/runners/portability/local_job_service.py
+++ b/sdks/python/apache_beam/runners/portability/local_job_service.py
@@ -31,6 +31,7 @@ from concurrent import futures
 import grpc
 from google.protobuf import text_format
 
+from apache_beam.metrics import monitoring_infos
 from apache_beam.portability.api import beam_artifact_api_pb2
 from apache_beam.portability.api import beam_artifact_api_pb2_grpc
 from apache_beam.portability.api import beam_fn_api_pb2_grpc
@@ -107,6 +108,26 @@ class LocalJobServicer(abstract_job_service.AbstractJobServiceServicer):
     if os.path.exists(self._staging_dir) and self._cleanup_staging_dir:
       shutil.rmtree(self._staging_dir, ignore_errors=True)
 
+  def GetJobMetrics(self, request, context=None):
+    if request.job_id not in self._jobs:
+      raise LookupError("Job {} does not exist".format(request.job_id))
+
+    result = self._jobs[request.job_id].result
+    monitoring_info_list = []
+    for mi in result._monitoring_infos_by_stage.values():
+      monitoring_info_list.extend(mi)
+
+    # Filter out system metrics
+    user_monitoring_info_list = [
+        x for x in monitoring_info_list
+        if monitoring_infos._is_user_monitoring_info(x) or
+        monitoring_infos._is_user_distribution_monitoring_info(x)
+    ]
+
+    return beam_job_api_pb2.GetJobMetricsResponse(
+        metrics=beam_job_api_pb2.MetricResults(
+            committed=user_monitoring_info_list))
+
 
 class SubprocessSdkWorker(object):
   """Manages a SDK worker implemented as a subprocess communicating over grpc.
@@ -176,6 +197,7 @@ class BeamJob(abstract_job_service.AbstractBeamJob):
     self._log_queues = []
     self.state = beam_job_api_pb2.JobState.STARTING
     self.daemon = True
+    self.result = None
 
   @property
   def state(self):
@@ -204,11 +226,12 @@ class BeamJob(abstract_job_service.AbstractBeamJob):
   def _run_job(self):
     with JobLogHandler(self._log_queues):
       try:
-        fn_api_runner.FnApiRunner(
+        result = fn_api_runner.FnApiRunner(
             provision_info=self._provision_info).run_via_runner_api(
                 self._pipeline_proto)
         logging.info('Successfully completed job.')
         self.state = beam_job_api_pb2.JobState.DONE
+        self.result = result
       except:  # pylint: disable=bare-except
         logging.exception('Error running pipeline.')
         logging.exception(traceback)
--- a/sdks/python/apache_beam/runners/portability/portable_metrics.py
+++ b/sdks/python/apache_beam/runners/portability/portable_metrics.py
@@ -1,0 +1,69 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from __future__ import absolute_import
+
+import logging
+
+from apache_beam.metrics import monitoring_infos
+from apache_beam.metrics.execution import MetricKey
+from apache_beam.metrics.metric import MetricName
+
+
+def from_monitoring_infos(monitoring_info_list, user_metrics_only=False):
+  """Groups MonitoringInfo objects into counters, distributions and gauges.
+
+  Args:
+    monitoring_info_list: An iterable of MonitoringInfo objects.
+    user_metrics_only: If true, includes user metrics only.
+  Returns:
+    A tuple containing three dictionaries: counters, distributions and gauges,
+    respectively. Each dictionary contains (MetricKey, metric result) pairs.
+  """
+  counters = {}
+  distributions = {}
+  gauges = {}
+
+  for mi in monitoring_info_list:
+    if (user_metrics_only and
+        not monitoring_infos.is_user_monitoring_info(mi)):
+      continue
+
+    try:
+      key = _create_metric_key(mi)
+    except ValueError as e:
+      logging.debug(str(e))
+      continue
+    metric_result = (monitoring_infos.extract_metric_result_map_value(mi))
+
+    if monitoring_infos.is_counter(mi):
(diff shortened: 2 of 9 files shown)


kamilwu:
Run Python PreCommit

kamilwu:
R: @robertwb @ajamato Could you take a look? 
cc: @lgajowy

pabloem:
We'll have to rely on r:@Ardagan

sdks/python/apache_beam/runners/portability/local_job_service.py
```diff
@@ -107,6 +108,26 @@ def stop(self, timeout=1):
...
+      monitoring_info_list.extend(mi)
+
+    # Filter out system metrics
+    user_monitoring_info_list = [
+        x for x in monitoring_info_list
+        if monitoring_infos._is_user_monitoring_info(x) or
+        monitoring_infos._is_user_distribution_monitoring_info(x)
+    ]
+
+    return beam_job_api_pb2.GetJobMetricsResponse(
```
ajamato: FYI, here is some background. I was not aware that these existed in proto form now. Creating a proto was suggested but not pursued at the time.
https://s.apache.org/get-metrics-api

When I last worked here MetricResult did not have a proto format. The plan was to use MonitoringInfos for a language agnostic format, and then MetricResult would be a language specific format (python, java, go, etc.) Each runner should provide a way to return MonitoringInfos, and each language would have  a library to convert MonitoringInfos to MetricResult protos

Its hard for me to review, I don't think I am up to date on whatever the current plan/usage of these protos are.

sdks/python/apache_beam/runners/portability/portable_metrics.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class ParseMonitoringInfoMixin(object):
+  @staticmethod
+  def from_monitoring_infos(monitoring_info_list, user_metrics_only=False):
+    """Groups MonitoringInfo objects into counters, distributions and gauges.
+
+    Args:
+      monitoring_info_list: An iterable of MonitoringInfo objects.
+      user_metrics_only: If true, includes user metrics only.
```
ajamato: Please add a Returns section to pydoc comment

sdks/python/apache_beam/runners/portability/portable_metrics.py
```diff
@@ -0,0 +1,74 @@
...
+      elif monitoring_infos.is_distribution(mi):
+        distributions[key] = metric_result
+      elif monitoring_infos.is_gauge(mi):
+        gauges[key] = metric_result
+
+    return counters, distributions, gauges
+
+  @staticmethod
+  def _create_metric_key(monitoring_info):
+    step_name = ParseMonitoringInfoMixin._get_step_name(monitoring_info)
```
ajamato: _get_step_name seems like it should live on monitoring_infos.py

sdks/python/apache_beam/runners/portability/portable_metrics.py
```diff
@@ -0,0 +1,74 @@
...
+    namespace, name = monitoring_infos.parse_namespace_and_name(monitoring_info)
+    return MetricKey(step_name, MetricName(namespace, name))
+
+  @staticmethod
+  def _get_step_name(monitoring_info):
+    keys_to_check = [monitoring_infos.PTRANSFORM_LABEL,
```
ajamato: This doesn't seem correct. Why would you consider the step name to be under labels other than PTRANSFORM_LABEL

see montioring_info specs defined here
https://github.com/apache/beam/blob/d4afbabf38a3ab557625c9c091ed5f06ca6731ce/model/pipeline/src/main/proto/metrics.proto

PTRANSFORM_LABEL is the only one used for this purpose

sdks/python/apache_beam/runners/portability/portable_runner.py
```diff
@@ -361,16 +363,31 @@ def add_runner_options(parser):
                           state_stream, cleanup_callbacks)
 
 
-class PortableMetrics(metrics.metric.MetricResults):
+class PortableMetrics(metric.MetricResults,
+                      portable_metrics.ParseMonitoringInfoMixin):
```
ajamato: Please remove the use of double inheritance. You can use the methods from portable_metrics.ParseMonitoringInfoMixin by  making that file define the methods in the module instead of a class, since it only defined static methods

sdks/python/apache_beam/runners/portability/portable_metrics.py
```diff
@@ -0,0 +1,74 @@
...
+#
+
+from __future__ import absolute_import
+
+from apache_beam.metrics import monitoring_infos
+from apache_beam.metrics.execution import MetricKey
+from apache_beam.metrics.metric import MetricName
+
+
+class ParseMonitoringInfoMixin(object):
```
ajamato: Please remove this class and instead define the methods on a module, without a class. All the methods here are static, so there is no need for a class/object instance,

sdks/python/apache_beam/testing/load_tests/load_test.py
```diff
@@ -52,32 +52,24 @@ def setUp(self):
...
-    if not self.are_metrics_collected():
-      logging.info('Metrics will not be collected')
-      self.metrics_monitor = None
-    else:
-      self.metrics_monitor = MetricsReader(
-          project_name=self.project_id,
-          bq_table=self.metrics_namespace,
-          bq_dataset=self.metrics_dataset,
-      )
+    self.metrics_monitor = MetricsReader(
```
ajamato: I am not familiar with MetricsReader

sdks/python/apache_beam/runners/portability/portable_metrics.py
kamilwu: +1

sdks/python/apache_beam/runners/portability/portable_metrics.py
ajamato: No, the intention is that URN + Labels defines the metric instance.

Think of URN as the class of metric, and URN+Labels defines the object instance, as an analogy.

I don't quite remember exactly what MetricKey contained, but generally our collection objects for metrics need to account for URN+Labels to correctly identify the metric instance.

MetricResult was originally designed just for user metrics, which did not have labels. Just a name and namespace. The labels concept was introduced to MonitoringInfos later. Then the name and namespace were reworked to be labels.
See the user metric MonitoringInfoSpec, defining what should be populated on an MI
https://github.com/apache/beam/blob/d4afbabf38a3ab557625c9c091ed5f06ca6731ce/model/pipeline/src/main/proto/metrics.proto#L67

ajamato:
Happy to keep reviewing but please have Ardagan review as well.

sdks/python/apache_beam/runners/portability/portable_metrics.py
Ardagan: In Dataflow runner we have to translate SDK PTransform and PCollection names [back to user defined names](https://github.com/apache/beam/blob/19799cdb1ba0c9b92544a470274d1c5c3fa6ff70/runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/fn/control/ElementCountMonitoringInfoToCounterUpdateTransformer.java#L100). I'm still checking the code to see if that is required here as well.

sdks/python/apache_beam/testing/load_tests/load_test.py
kamilwu: We use MetricsReader to push metrics obtained from PipelineResult to BigQuery in load tests. This diff is just a fix to prevent system metrics from being published, as we don't need them in BigQuery

sdks/python/apache_beam/runners/portability/portable_metrics.py
kamilwu: I see your point now.
I think it might be too difficult (if possible at all) to implement such a mapping right here. Since this is a portable runner, such a logic would have to be aware of all possible runner's implementation details.

I'm going to stay with checking PTRANSFORM_LABEL only for now.

kamilwu:
Thanks for your review so far, I've put some fixes

kamilwu:
Run Python PreCommit

sdks/python/apache_beam/runners/portability/portable_metrics.py
```diff
@@ -22,53 +22,44 @@
...
-
-    for key in keys_to_check:
-      value = monitoring_info.labels.get(key)
-      if value:
-        return value
-    return ''
+def _create_metric_key(monitoring_info):
+  step_name = monitoring_infos.get_step_name(monitoring_info)
+  if not step_name:
+    raise ValueError("Step name is empty")
```
Ardagan: Dump monitoring info to error message. Otherwise this error is indescriptive. Ideally change text to what you expect here, for example: "Monitoring info should contain XXX field. Monitoring info: {}" or in this case it can be: "Failed to deduce step_name from MonitoringInfo: {}"

sdks/python/apache_beam/runners/portability/portable_metrics.py
```diff
@@ -22,53 +22,44 @@
...
+    if (user_metrics_only and
+        not monitoring_infos.is_user_monitoring_info(mi)):
+      continue
 
-      key = ParseMonitoringInfoMixin._create_metric_key(mi)
-      metric_result = (monitoring_infos.extract_metric_result_map_value(mi))
+    try:
+      key = _create_metric_key(mi)
+    except ValueError:
+      continue
```
Ardagan: Add debug log for the error. It is possible that we skip monitoring infos that have incorrect setup and fail to deduce step_name here.

sdks/python/apache_beam/runners/portability/portable_metrics.py
ajamato: If you want to aggregate away the other labels, you could use this approach if you sum everything with the same ptransform_label

sdks/python/apache_beam/runners/portability/portable_metrics.py
kamilwu: "Failed to deduce step_name..." looks good.

kamilwu:
Thanks @ajamato!
@Ardagan I've added dumping monitoring infos. I've also squashed all commits.

Ardagan:
run python precommit

Ardagan:
There are no more comments, I'll merge PR.

PR #9775, opened 2019-10-11
[BEAM-8372] Job server submitting UberJars directly to Flink Runner.

robertwb:
One can now use the FlinkRunner with the --flink_master_url=flink-master-url
and an UberJar will automatically be created with the artifacts and pipeline
definition, uploaded to the master, and started.

------------------------

Thank you for your contribution! Follow this checklist to help us incorporate your contribution quickly and easily:

 - [ ] [**Choose reviewer(s)**](https://beam.apache.org/contribute/#make-your-change) and mention them in a comment (`R: @username`).
 - [ ] Format the pull request title like `[BEAM-XXX] Fixes bug in ApproximateQuantiles`, where you replace `BEAM-XXX` with the appropriate JIRA issue, if applicable. This will automatically link the pull request to the issue.
 - [ ] If this contribution is large, please file an Apache [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

Lang | SDK | Apex | Dataflow | Flink | Gearpump | Samza | Spark
--- | --- | --- | --- | --- | --- | --- | ---
Go | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go/lastCompletedBuild/) | --- | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_VR_Flink/lastCompletedBuild/) | --- | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_VR_Spark/lastCompletedBuild/)
[...]

PR #9844, opened 2019-10-21
[BEAM-8372] Support flink_master parameter in job server

mxm:
The original parameter was named flink_master (defined in FlinkPipelineOptions)
but since there have been inconsistencies. So it's best to support both now.

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/runners/portability/flink_runner.py
+++ b/sdks/python/apache_beam/runners/portability/flink_runner.py
@@ -44,7 +44,12 @@ class FlinkRunnerOptions(pipeline_options.PipelineOptions):
   @classmethod
   def _add_argparse_args(cls, parser):
     parser.add_argument('--flink_master',
-                        default='[local]')
+                        default='[auto]',
+                        help='Flink master address (host:port) to submit the'
+                             ' job against. Use "[local]" to start a local'
+                             ' cluster for the execution. Use "[auto]" if you'
+                             ' plan to either execute locally or submit through'
+                             ' Flink\'s CLI.')
     parser.add_argument('--flink_version',
                         default=PUBLISHED_FLINK_VERSIONS[-1],
                         choices=PUBLISHED_FLINK_VERSIONS,
--- a/sdks/python/apache_beam/runners/portability/flink_runner.py
+++ b/sdks/python/apache_beam/runners/portability/flink_runner.py
@@ -43,7 +43,8 @@ class FlinkRunner(portable_runner.PortableRunner):
 class FlinkRunnerOptions(pipeline_options.PipelineOptions):
   @classmethod
   def _add_argparse_args(cls, parser):
-    parser.add_argument('--flink_master', default='[local]')
+    parser.add_argument('--flink_master',
+                        default='[local]')
     parser.add_argument('--flink_version',
                         default=PUBLISHED_FLINK_VERSIONS[-1],
                         choices=PUBLISHED_FLINK_VERSIONS,
--- a/runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkJobInvoker.java
+++ b/runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkJobInvoker.java
@@ -66,7 +66,7 @@ public class FlinkJobInvoker extends JobInvoker {
         String.format("%s_%s", flinkOptions.getJobName(), UUID.randomUUID().toString());
 
     if (FlinkPipelineOptions.AUTO.equals(flinkOptions.getFlinkMaster())) {
-      flinkOptions.setFlinkMaster(serverConfig.getFlinkMasterUrl());
+      flinkOptions.setFlinkMaster(serverConfig.getFlinkMaster());
     }
 
     PortablePipelineOptions portableOptions = flinkOptions.as(PortablePipelineOptions.class);
--- a/runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkJobServerDriver.java
+++ b/runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkJobServerDriver.java
@@ -38,11 +38,14 @@ public class FlinkJobServerDriver extends JobServerDriver {
 
   /** Flink runner-specific Configuration for the jobServer. */
   public static class FlinkServerConfiguration extends ServerConfiguration {
-    @Option(name = "--flink-master-url", usage = "Flink master url to submit job.")
-    private String flinkMasterUrl = "[auto]";
+    @Option(
+        name = "--flink-master",
+        aliases = {"--flink-master-url"},
+        usage = "Flink master address (host:port) to submit job.")
+    private String flinkMaster = "[auto]";
 
-    String getFlinkMasterUrl() {
-      return this.flinkMasterUrl;
+    String getFlinkMaster() {
+      return this.flinkMaster;
     }
 
     @Option(
--- a/runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkJobServerDriver.java
+++ b/runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkJobServerDriver.java
@@ -41,8 +41,11 @@ public class FlinkJobServerDriver extends JobServerDriver {
     @Option(
         name = "--flink-master",
         aliases = {"--flink-master-url"},
-        usage = "Flink master address (host:port) to submit job.")
-    private String flinkMaster = "[auto]";
+        usage =
+            "Flink master address (host:port) to submit the job against. Use Use \"[local]\" to start a local "
+                + "cluster for the execution. Use \"[auto]\" if you plan to either execute locally or submit through "
+                + "Flink\'s CLI.")
+    private String flinkMaster = FlinkPipelineOptions.AUTO;
 
     String getFlinkMaster() {
       return this.flinkMaster;
--- a/runners/flink/src/test/java/org/apache/beam/runners/flink/FlinkJobServerDriverTest.java
+++ b/runners/flink/src/test/java/org/apache/beam/runners/flink/FlinkJobServerDriverTest.java
@@ -42,7 +42,7 @@ public class FlinkJobServerDriverTest {
     assertThat(config.getPort(), is(8099));
     assertThat(config.getArtifactPort(), is(8098));
     assertThat(config.getExpansionPort(), is(8097));
-    assertThat(config.getFlinkMasterUrl(), is("[auto]"));
+    assertThat(config.getFlinkMaster(), is("[auto]"));
     assertThat(config.isCleanArtifactsPerJob(), is(true));
     FlinkJobServerDriver flinkJobServerDriver = FlinkJobServerDriver.fromConfig(config);
     assertThat(flinkJobServerDriver, is(not(nullValue())));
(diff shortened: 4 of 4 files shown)


sdks/python/apache_beam/runners/portability/flink_runner.py
```diff
@@ -43,7 +43,8 @@ def default_job_server(self, options):
 class FlinkRunnerOptions(pipeline_options.PipelineOptions):
   @classmethod
   def _add_argparse_args(cls, parser):
-    parser.add_argument('--flink_master', default='[local]')
+    parser.add_argument('--flink_master', '--flink_master_url',
+                        default='[local]')
```
ibzib: The default for Java is `[auto]`. Should we change the Python SDK to match?

mxm: Good point! We should.

robertwb:
There's also the question of whether this should include the `http[s]://` portion. (Especially the 's' part is potentially ambiguous.)

mxm:
The address was never meant to be anything different from `host:port`. Somehow the perception of an URL slipped in through time. I tried to clarify that in the parameter description.

mxm:
To clarify, if you want to have an encrypted connection, you need to configure this in the Flink config file. Also with Beam this configuration will be picked up.

mxm:
I was inclined to remove the `flink_master_url` parameters entirely, but I was afraid it might break existing user scripts.

robertwb:
`flink_master_url` is very new, I think we can consolidate on one option for now. 

So, if I understand correctly, if `flink_master` is just host:port, there is no way to (simultaneously) submit a job to two different clusters (one with ssl, one without) without editing the config file in the middle? This seems quite unfortunate. Also, the code here does not read the config file. 

What if, instead, we made the `http[s]://` part optional, where it would be stripped in Java and assumed to be just `http://` in Python if not specified? (Or it could try https and then fall back to http on failure.)

robertwb:
I think it would be good to get this resolved in the current release, if possible.

mxm:
>`flink_master_url` is very new, I think we can consolidate on one option for now. 

The instance in `FlinkJobServerDriver` has been there for over a year. It might be sufficient to convert it to an alias only there.

>I think it would be good to get this resolved in the current release, if possible.

+1

>So, if I understand correctly, if `flink_master` is just host:port, there is no way to (simultaneously) submit a job to two different clusters (one with ssl, one without) without editing the config file in the middle? This seems quite unfortunate. Also, the code here does not read the config file.

The `FlinkJobServerDriver` reads the config file before it submits the Flink job to the cluster. So the Python code does not have to worry about it. Yes, there is no way to submit against two clusters without changing the configuration file. You can start a second job server though, which uses a different configuration file.

>What if, instead, we made the `http[s]://` part optional, where it would be stripped in Java and assumed to be just `http://` in Python if not specified? (Or it could try https and then fall back to http on failure.)

That would be a new feature. Flink does not accept any url schemas like `http://`. It just expects host:port. The config file determines whether the Rest service running at the port uses SSL or not.

sdks/python/apache_beam/runners/portability/flink_runner.py
```diff
@@ -43,7 +43,13 @@ def default_job_server(self, options):
 class FlinkRunnerOptions(pipeline_options.PipelineOptions):
   @classmethod
   def _add_argparse_args(cls, parser):
-    parser.add_argument('--flink_master', default='[local]')
+    parser.add_argument('--flink_master',
+                        default='[auto]',
```
robertwb: `[local]` is also referenced at https://github.com/apache/beam/blob/release-2.17.0/sdks/python/apache_beam/runners/portability/flink_runner.py#L36

sdks/python/apache_beam/runners/portability/flink_runner.py
```diff
@@ -43,7 +43,13 @@ def default_job_server(self, options):
...
+                        help='Flink master address (host:port) to submit the'
+                             ' job against. Use "[local]" to start a local'
+                             ' cluster for the execution. Use "[auto]" if you'
+                             ' plan to either execute locally or submit through'
```
robertwb: How is this choice made if `[auto]` is passed? Should it really be the default?

robertwb:
I guess it's `--flink_master` that's new (in Python's FlinkRunner). But if we don't have the http than we need to change the https://github.com/apache/beam/pull/9775 to add it.

mxm:
>I guess it's `--flink_master` that's new (in Python's FlinkRunner).

This option since the inception of the Flink Runner was `flinkMaster` (Java) or `flink_master` (Python): https://github.com/apache/beam/blob/8972e5e10050b67a4a9a0027c09d2d2eb00b45ed/runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkPipelineOptions.java#L66

https://github.com/apache/beam/blob/50528d024b83a5083afa24bd7c3e3694a6347c15/sdks/python/apache_beam/options/pipeline_options.py#L669

In the light of URLs not even being supported, it also makes sense to keep that name instead of renaming it to `flink_master_url`.

> But if we don't have the http than we need to change the https://github.com/apache/beam/pull/9775 to add it. 

Could there be a misunderstanding? Rest does not mean that we need to prepend `http://`. The format is `host:port` where host is a hostname or a IPv4/IPv6 address, and port is the port number, i.e. of Flink's Rest service (typically running at port 8081). Please correct me, but I don't see that we need to change #9775.

sdks/python/apache_beam/runners/portability/flink_runner.py
mxm: I'm coming from the Java SDK, where we use `[auto]` as the default. I think it makes sense because `[local]` has unexpected consequences when you use it with the Flink CLI (now that we have executable jars that is possible). It will force a local execution instead of submitting the job to the cluster. We could even think about removing the local mode, since it is covered by the auto mode as well.

sdks/python/apache_beam/runners/portability/flink_runner.py
mxm: This looks like it is for testing? That looks ok.

robertwb:
If `flink_master` is the option, and has been for a long time, then we should just use that. 

As for adapting #9775, REST does require specification of a protocol. Currently if we just pass `--runner=FlinkRunner --flink_master=localhost:8081` we get

```
File "/Users/robertwb/Work/beam/incubator-beam/sdks/python/apache_beam/runners/portability/portable_runner.py", line 315, in run_pipeline
    timeout=portable_options.job_server_timeout)
  File "/Users/robertwb/Work/beam/incubator-beam/sdks/python/apache_beam/runners/portability/abstract_job_service.py", line 53, in Prepare
    request.pipeline_options)
  File "/Users/robertwb/Work/beam/incubator-beam/sdks/python/apache_beam/runners/portability/flink_uber_jar_job_server.py", line 77, in create_beam_job
    self.executable_jar(),
  File "/Users/robertwb/Work/beam/incubator-beam/sdks/python/apache_beam/runners/portability/flink_uber_jar_job_server.py", line 66, in executable_jar
    'runners:flink:%s:job-server:shadowJar' % self.flink_version()))
  File "/Users/robertwb/Work/beam/incubator-beam/sdks/python/apache_beam/runners/portability/flink_uber_jar_job_server.py", line 70, in flink_version
    '%s/v1/config' % self._master_url).json()['flink-version']
  File "/Users/robertwb/Work/beam/venv3/lib/python3.6/site-packages/requests-2.22.0-py3.6.egg/requests/api.py", line 75, in get
    return request('get', url, params=params, **kwargs)
  File "/Users/robertwb/Work/beam/venv3/lib/python3.6/site-packages/requests-2.22.0-py3.6.egg/requests/api.py", line 60, in request
    return session.request(method=method, url=url, **kwargs)
  File "/Users/robertwb/Work/beam/venv3/lib/python3.6/site-packages/requests-2.22.0-py3.6.egg/requests/sessions.py", line 533, in request
    resp = self.send(prep, **send_kwargs)
  File "/Users/robertwb/Work/beam/venv3/lib/python3.6/site-packages/requests-2.22.0-py3.6.egg/requests/sessions.py", line 640, in send
    adapter = self.get_adapter(url=request.url)
  File "/Users/robertwb/Work/beam/venv3/lib/python3.6/site-packages/requests-2.22.0-py3.6.egg/requests/sessions.py", line 731, in get_adapter
    raise InvalidSchema("No connection adapters were found for '%s'" % url)
requests.exceptions.InvalidSchema: No connection adapters were found for 'localhost:8081/v1/config'
```

because we're using requests to attempt to communicate with the flink master. (Note that in this case there is no `FlinkJobServerDriver`, etc.) We could automatically append `http` or `https` if we could figure out which one, but it seems strange design to have to read a config file when the address is passed explicitly. 

Taking a step back, here's what I think the ideal experience should be. If one passes just `--runner=FlinkRunner` then it should start up a local runner and run against that, automatically enabling `LOOPBACK` mode if one hasn't been explicitly chosen for ease of use. If one passes `--runner=FlinkRunner --flink_master=address` it would submit the job to the cluster, without having to even download jars and invoke java. It's unclear how `[auto]` would get this behavior (or indeed what the behavior of `[auto]` even is).

tweise:
`[auto]` is the correct default value, it will result in local execution except when a job is run by Flink CLI or REST API. The use case here is a Flink REST client. In that context there is no Flink config, the REST API is fully described here: https://ci.apache.org/projects/flink/flink-docs-stable/monitoring/rest_api.html

I don't see why `flink_master` has to serve as endpoint URL necessarily. Why can there not be a separate property for just for the REST client?

It would be good to separate the more general discussion from this PR and surface it on the mailing list. That new job submission mode and related changes need to become more visible to others interested in Flink.

robertwb:
I'm not sure what you mean by "a job run by Flink CLI." What does that mean
for a Python job? (I was trying to read the docs and code for [auto] but
wasn't able to figure it out.)

As for re-using flink master for the REST client, this is to have the same
behavior as when using the automatically started job server (e.g. for py <
3.6).

+1 to bringing this to the list. It's be good to just enumerate all the
modes we have and how they should work.

mxm:
Let's do the following: 

1. Unify the two parameters `flink_master` and `flink_master_url` as fixed here for the job server and previously by #9803 .
2. Make the master address a url. Add `http://` to the `flink_master` in Python if no scheme is specified. Similarly, remove any `http://` in Java, since the Java rest client does not expect a scheme.
3. Deprecate the `[auto]` and `[local]` property, it should be sufficient to replace them with either an address string or an empty string. The empty string would either mean local execution or, in the context of the Flink CLI tool, loading the master address from the config.

(2) and (3) should be follow-ups. I'll post a summary to the list.

> If flink_master is the option, and has been for a long time, then we should just use that.

+1

> As for adapting #9775, REST does require specification of a protocol. Currently if we just pass --runner=FlinkRunner --flink_master=localhost:8081 we get

Oh I see. I wasn't aware that we need to pass `http://` due to requests. In that regard the `_url` parameter makes sense. Still better to settle for a single parameter.

> The use case here is a Flink REST client. In that context there is no Flink config, the REST API is fully described here: https://ci.apache.org/projects/flink/flink-docs-stable/monitoring/rest_api.html

This is not true. The Flink Rest client loads the config in order to support uses cases like SSL encryption (i.e. loading the trust store, initializing the ssl connection, then sending http requests over it). However, the code in Python's `FlinkRunner` attempts to directly communicate with the Flink master via the `requests` library.

>I don't see why flink_master has to serve as endpoint URL necessarily. Why can there not be a separate property for just for the REST client?

Why should there be multiple options for the same thing? `flink_master` is already the Rest endpoint address. No need to have another `flink_master_url` option. We can add URL support and probably it is fair to add `http://` in case no URL has been supplied.

>I'm not sure what you mean by "a job run by Flink CLI." What does that mean
for a Python job? (I was trying to read the docs and code for [auto] but
wasn't able to figure it out.)

I mean running a jar directly with the Flink command-line tool, i.e. `bin/flink`. This is the defacto standard way to run Flink jobs. It does not apply here but we want to make sure that generated Jars do not break when submitted through the CLI. The `[auto]` mode basically says, either (1) figure out the cluster address from the context (`bin/flink` reads it from a config file), or (2) execute locally if there is no such context or the jar is run directly.

PR #9854, opened 2019-10-22
[BEAM-8457] Label Dataflow jobs from Notebook

KevinGG:
1. Changed the pipeline.run() API to allow a runner and an option
parameter so that a pipeline initially bundled w/ an interactive runner
can be directly run by other runners from notebook.
2. Implicitly added the necessary source information through user labels
when the user does p.run(runner=DataflowRunner(), options=options) or
DataflowRunner().run_pipeline(p, options).

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/pipeline.py
+++ b/sdks/python/apache_beam/pipeline.py
@@ -95,7 +95,7 @@ class Pipeline(object):
   (e.g. ``input | "label" >> my_tranform``).
   """
 
-  def __init__(self, runner=None, options=None, argv=None, interactive=None):
+  def __init__(self, runner=None, options=None, argv=None):
     """Initialize a pipeline object.
 
     Args:
@@ -111,8 +111,6 @@ class Pipeline(object):
         to be used for building a
         :class:`~apache_beam.options.pipeline_options.PipelineOptions` object.
         This will only be used if argument **options** is :data:`None`.
-      interactive (bool): a boolean value indicating whether the pipeline is
-        created in an interactive environment such as interactive notebooks.
 
     Raises:
       ~exceptions.ValueError: if either the runner or options argument is not
@@ -174,13 +172,10 @@ class Pipeline(object):
     # If a transform is applied and the full label is already in the set
     # then the transform will have to be cloned with a new label.
     self.applied_labels = set()
-    from apache_beam.runners.interactive import interactive_runner
-    if interactive:
-      self.interactive = interactive
-    elif isinstance(self.runner, interactive_runner.InteractiveRunner):
-      self.interactive = True
-    else:
-      self.interactive = False
+    # A boolean value indicating whether the pipeline is created in an
+    # interactive environment such as interactive notebooks. Initialized as
+    # None. The value is set ad hoc when `pipeline.run()` is invoked.
+    self.interactive = None
 
   @property
   @deprecated(since='First stable release',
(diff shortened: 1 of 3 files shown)


KevinGG:
R: @pabloem 
PTAL, thanks!

sdks/python/apache_beam/pipeline.py
```diff
@@ -396,28 +396,40 @@ def replace_all(self, replacements):
...
+
+    If another runner instance and options are provided, that runner will
+    execute the pipeline with the given options. If either of them is not set,
+    the default runner will run the pipeline with the original options
+    assigned to the pipeline. The usage is similar to directly invoking
+    `runner.run_pipeline(pipeline, options)`.
+    """
+    runner_in_use = self.runner
+    options_in_use = self._options
+    if runner and options:
```
pabloem: What if either runner or options are not provided? Should that throw an error? Currently, if only one is provided, it'll be ignored - and that would be quite surprising for users.

sdks/python/apache_beam/runners/dataflow/dataflow_runner.py
```diff
@@ -360,6 +360,15 @@ def visit_transform(self, transform_node):
 
   def run_pipeline(self, pipeline, options):
     """Remotely executes entire pipeline or parts reachable from node."""
+    # Label goog-notebook if pipeline is initiated from interactive runner.
+    from apache_beam.runners.interactive import interactive_runner
+    if isinstance(pipeline.runner, interactive_runner.InteractiveRunner):
```
pabloem: This seems fine - but what if we go with the `runner_in_use` codepath? Would users do: `p.run(runner=InteractiveRunner(DataflowRunner()), options=...)`? Or would users create a pipeline with InteractiveRunner and then do `p.run(runner=DataflowRunner()...`? Is it poissible for users to do `p = beam.Pipeline()`, and then do `InteractiveRunner().run_pipeline(p)`/`InteractiveRunner(DataflowRunner()).run_pipeline(p)`?

IIUC users would have to pass the interactive runner in `p = beam.Pipeline()` to activate the interactive mode, right? InteractiveRunner is not automatically selected?

sdks/python/apache_beam/pipeline.py
KevinGG: You're right! This will surprise the user. I've changed it to throw error if either is not provided instead of ignoring the input by default.

sdks/python/apache_beam/runners/dataflow/dataflow_runner.py
KevinGG: I've missed the path where a new Pipeline is created and `run()` is invoked again.
Yes, all of these would be possible.
I've added an `interactive` parameter at the constructor level for `Pipeline` using default value `None`. `run()` and `from_runner_api()` will pass the `None` or `bool` value down no matter how the user chains the runners. I'm not very confident with the naming but the change should be backward compatible for Beam.

Currently, I'm running into a problem when testing. Once I set `labels`, Dataflow job will fail immediately and throw `Error processing pipeline.` error. There will be no job graph, no worker started, no logs. Looks like when there is user label in the job request, Dataflow cannot convert the work item into internal representation.

I'll do some investigation and figure out why.

sdks/python/apache_beam/pipeline.py
```diff
@@ -396,28 +405,46 @@ def replace_all(self, replacements):
...
+                       'is given.')
     # When possible, invoke a round trip through the runner API.
     if test_runner_api and self._verify_runner_api_compatible():
       return Pipeline.from_runner_api(
           self.to_runner_api(use_fake_coders=True),
-          self.runner,
-          self._options).run(False)
+          runner_in_use,
+          options_in_use,
+          interactive=self.interactive).run(False)
```
pabloem: Did you find that this was necessary? I don't think we should change the signature of the `from_runner_api` call. The pipeline protobuf should contain all the necessary information... Though I'd defer to @robertwb on this.

KevinGG: No, it's not necessary. On a second thought, I can just move the logic to determine `interactive` ad hoc in `run()` and put the `interactive` field as a parameter in the `run()` method. Then I don't even need to change the Pipeline constructor.

Also, I've added this to the interactive_runner for below use case:
`InteractiveRunner(underlying_runner=DataflowRunner()).run_pipeline(Pipeline(DataflowRunner()))`

pabloem:
Thanks Ning

charlesccychen:
@KevinGG @pabloem  I believe this PR may be causing issues in environments without IPython because it unconditionally imports the interactive runner (and thereby IPython).

aaltay:
Let's roll this back. I do not think we should be importing interactive in pipeline.py. And also my understanding is that runner will keep track of the interactivity not the pipeline.

robertwb:
+1. We should not be importing the interactive runner (it's causing problems with tests as well), and interactivity should not be a property of the pipeline, but of the runner (and I'd prefer a design that avoid passing an interactive bit around everywhere).

PR #9872, opened 2019-10-24
[BEAM-8471] Flink native job submission for portable pipelines

tweise:
Add a new Flink entry point (main method) that invokes the external SDK client entry point to generate the pipeline and submits the resulting Flink job like any other Flink native driver program would, via the optimizer plan environment ("[auto]").

Note that in this PR, the SDK client is assumed to be on the same host, which is the case when Flink and Python dependencies are in the same container image, for example. While this is something that can be solved at build time, the question from Robert made me realize that the dependency is almost identical to that between the runner and SDK on the execution side, which is abstracted via the environment concept. So we could, in the future, consider introducing a "client environment" or simply expand the existing environment. This would allow the SDK bits for both pipeline construction and execution to live in a side car container, separate from Flink. 

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkJobServerDriver.java
+++ b/runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkJobServerDriver.java
@@ -19,7 +19,6 @@ package org.apache.beam.runners.flink;
 
 import javax.annotation.Nullable;
 import org.apache.beam.runners.fnexecution.ServerFactory;
-import org.apache.beam.runners.fnexecution.jobsubmission.JobInvoker;
 import org.apache.beam.runners.fnexecution.jobsubmission.JobServerDriver;
 import org.apache.beam.sdk.extensions.gcp.options.GcsOptions;
 import org.apache.beam.sdk.io.FileSystems;
@@ -66,7 +65,7 @@ public class FlinkJobServerDriver extends JobServerDriver {
     options.as(GcsOptions.class).setGcsUploadBufferSizeBytes(1024 * 1024);
     // Register standard file systems.
     FileSystems.setDefaultPipelineOptions(options);
-    fromParams(args).run();
+    fromConfig(parseArgs(args)).run();
   }
 
   private static void printUsage(CmdLineParser parser) {
@@ -76,7 +75,7 @@ public class FlinkJobServerDriver extends JobServerDriver {
     System.err.println();
   }
 
-  public static FlinkJobServerDriver fromParams(String[] args) {
+  public static FlinkServerConfiguration parseArgs(String[] args) {
     FlinkServerConfiguration configuration = new FlinkServerConfiguration();
     CmdLineParser parser = new CmdLineParser(configuration);
     try {
@@ -86,33 +85,40 @@ public class FlinkJobServerDriver extends JobServerDriver {
       printUsage(parser);
       throw new IllegalArgumentException("Unable to parse command line arguments.", e);
     }
-
-    return fromConfig(configuration);
+    return configuration;
   }
 
   public static FlinkJobServerDriver fromConfig(FlinkServerConfiguration configuration) {
     return create(
         configuration,
         createJobServerFactory(configuration),
-        createArtifactServerFactory(configuration));
+        createArtifactServerFactory(configuration),
+        () -> FlinkJobInvoker.create(configuration));
   }
 
-  public static FlinkJobServerDriver create(
+  public static FlinkJobServerDriver fromConfig(
+      FlinkServerConfiguration configuration, JobInvokerFactory jobInvokerFactory) {
+    return create(
+        configuration,
+        createJobServerFactory(configuration),
+        createArtifactServerFactory(configuration),
+        jobInvokerFactory);
+  }
+
+  private static FlinkJobServerDriver create(
       FlinkServerConfiguration configuration,
       ServerFactory jobServerFactory,
-      ServerFactory artifactServerFactory) {
-    return new FlinkJobServerDriver(configuration, jobServerFactory, artifactServerFactory);
+      ServerFactory artifactServerFactory,
+      JobInvokerFactory jobInvokerFactory) {
+    return new FlinkJobServerDriver(
+        configuration, jobServerFactory, artifactServerFactory, jobInvokerFactory);
   }
 
   private FlinkJobServerDriver(
       FlinkServerConfiguration configuration,
       ServerFactory jobServerFactory,
-      ServerFactory artifactServerFactory) {
-    super(configuration, jobServerFactory, artifactServerFactory);
-  }
-
-  @Override
-  protected JobInvoker createJobInvoker() {
-    return FlinkJobInvoker.create((FlinkServerConfiguration) configuration);
+      ServerFactory artifactServerFactory,
+      JobInvokerFactory jobInvokerFactory) {
+    super(configuration, jobServerFactory, artifactServerFactory, jobInvokerFactory);
   }
 }
--- a/runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkPortableClientEntryPoint.java
+++ b/runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkPortableClientEntryPoint.java
@@ -1,0 +1,257 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.beam.runners.flink;
+
+import java.io.File;
+import java.nio.charset.Charset;
+import java.nio.file.Files;
+import java.time.Duration;
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import org.apache.beam.model.pipeline.v1.RunnerApi;
+import org.apache.beam.runners.fnexecution.environment.ProcessManager;
+import org.apache.beam.runners.fnexecution.jobsubmission.JobInvocation;
+import org.apache.beam.runners.fnexecution.jobsubmission.JobInvoker;
+import org.apache.beam.runners.fnexecution.jobsubmission.PortablePipelineResult;
+import org.apache.beam.runners.fnexecution.jobsubmission.PortablePipelineRunner;
+import org.apache.beam.runners.fnexecution.provisioning.JobInfo;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Preconditions;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.ImmutableList;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.util.concurrent.ListeningExecutorService;
+import org.apache.flink.api.common.time.Deadline;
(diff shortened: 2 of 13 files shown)


tweise:
R: @chadrik

chadrik:
Thanks for the heads up.  I don't have any input on this, but I do like staying in the loop on this subject!

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkPortableClientRunner.java
```diff
@@ -0,0 +1,250 @@
...
+      } catch (Exception processKillException) {
+        e.addSuppressed(processKillException);
+      }
+      throw e;
+    }
+  }
+
+  private void stopJobService() throws InterruptedException {
+    if (driver != null) {
+      driver.stop();
```
mxm: This should named `jobServer` to separate it from the pipeline submission driver.

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkPortableClientRunner.java
```diff
@@ -0,0 +1,250 @@
...
+    }
+    if (driverThread != null) {
+      driverThread.interrupt();
+      driverThread.join();
+    }
+  }
+
+  private class DetachedJobInvokerFactory implements FlinkJobServerDriver.JobInvokerFactory {
+
+    private CountDownLatch latch = new CountDownLatch(1);
```
mxm:
```suggestion
    private final CountDownLatch latch = new CountDownLatch(1);
```

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkPortableClientRunner.java
```diff
@@ -0,0 +1,250 @@
...
+              executorService,
+              pipeline,
+              flinkOptions,
+              handoverPipelineRunner);
+        }
+      };
+    }
+
+    private void executeDetachedJob() throws Exception {
+      Duration timeout = Duration.ofSeconds(30);
```
mxm: Could we extract these timeouts to static variables? Also in the future, they may have to be configurable.

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkPortableClientRunner.java
```diff
@@ -0,0 +1,250 @@
...
+    if (driverThread != null) {
+      driverThread.interrupt();
+      driverThread.join();
+    }
+  }
+
+  private class DetachedJobInvokerFactory implements FlinkJobServerDriver.JobInvokerFactory {
+
+    private CountDownLatch latch = new CountDownLatch(1);
+    private PortablePipelineRunner actualPipelineRunner;
```
mxm:
```suggestion
    private volatile PortablePipelineRunner actualPipelineRunner;
```

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkPortableClientRunner.java
```diff
@@ -0,0 +1,250 @@
...
+    private RunnerApi.Pipeline pipeline;
```
mxm:
```suggestion
    private volatile RunnerApi.Pipeline pipeline;
```

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkPortableClientRunner.java
mxm: I believe, without the volatile keyword, it is possible that after awaiting the latch, we still see null here under certain conditions.

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkPortableClientRunner.java
mxm: I believe, without the volatile keyword, it is possible that after awaiting the latch, we still see null here under certain conditions.

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkPortableClientRunner.java
```diff
@@ -0,0 +1,250 @@
...
+  private static final String DRIVER_CMD_FLAGS = "--job_endpoint=%s";
+
+  private final String driverCmd;
+  private FlinkJobServerDriver driver;
+  private Thread driverThread;
+  private DetachedJobInvokerFactory jobInvokerFactory;
+  private int jobPort = 0; // any free port
+
+  public FlinkPortableClientRunner(String driverCmd) {
+    this.driverCmd = driverCmd;
```
mxm:
```suggestion
    Preconditions.checkState(!driverCmd.contains(DRIVER_CMD_FLAGS), "Driver command must not contain "+ DRIVER_CMD_FLAGS);
    this.driverCmd = driverCmd;
```

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkPortableClientRunner.java
```diff
@@ -0,0 +1,250 @@
...
+ * "detached" status to the driver program.
+ *
+ * <p>Upon arrival of the job invocation, the entry point executes the runner, which prepares
+ * ("executes") the Flink job through the plan environment.
+ *
+ * <p>Finally Flink launches the job.
+ */
+public class FlinkPortableClientRunner {
+  private static final Logger LOG = LoggerFactory.getLogger(FlinkPortableClientRunner.class);
+  private static final String DRIVER_CMD_FLAGS = "--job_endpoint=%s";
```
mxm:
```suggestion
  private static final String JOB_ENDPOINT_FLAG = "--job_endpoint";
```

mxm: This allows us to check for the args in the user command string. The `%s` would have to be appended in the format string.

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkPortableClientRunner.java
```diff
@@ -0,0 +1,250 @@
...
+
+  public FlinkPortableClientRunner(String driverCmd) {
+    this.driverCmd = driverCmd;
+  }
+
+  /** Main method to be called within the Flink OptimizerPlanEnvironment. */
+  public static void main(String[] args) throws Exception {
+    Preconditions.checkArgument(
+        ExecutionEnvironment.getExecutionEnvironment() instanceof OptimizerPlanEnvironment,
+        "Can only execute in OptimizerPlanEnvironment");
```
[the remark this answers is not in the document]
tweise: I'm going to remove this because it will now also work with other environments.

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkPortableClientRunner.java
tweise: That should not happen because we don't read the value before the latch? I changed it anyways since there isn't any cost associated in this case.

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkJobServerDriver.java
```diff
@@ -76,7 +75,7 @@ private static void printUsage(CmdLineParser parser) {
     System.err.println();
   }
 
-  public static FlinkJobServerDriver fromParams(String[] args) {
+  public static FlinkServerConfiguration fromParams(String[] args) {
```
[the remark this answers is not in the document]
tweise: Renamed, made it a separate method since the configuration is already exposed and used independently.

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkPortableClientRunner.java
mxm: I think there is no guarantee about reading up-to-date references between threads unless you use `volatile`.

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkPortableClientEntryPoint.java
```diff
@@ -0,0 +1,254 @@
...
+  }
+
+  private void runDriverProgram() throws Exception {
+    ProcessManager processManager = ProcessManager.create();
+    String executable = "bash";
+    List<String> args =
+        ImmutableList.of(
+            "-c",
+            String.format(
+                "exec %s " + JOB_ENDPOINT_FLAG + "=%s", driverCmd, jobServer.getJobServerUrl()));
```
mxm:
```suggestion
                "exec %s %s=%s", driverCmd, JOB_ENDPOINT_FLAG, jobServer.getJobServerUrl()));
```
;)

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkPortableClientEntryPoint.java
```diff
@@ -0,0 +1,249 @@
...
+ *
+ * <p>The job service defers execution of the pipeline to the plan environment and returns the
+ * "detached" status to the driver program.
+ *
+ * <p>Upon arrival of the job invocation, the entry point executes the runner, which prepares
+ * ("executes") the Flink job through the plan environment.
+ *
+ * <p>Finally Flink launches the job.
+ */
+public class FlinkPortableClientEntryPoint {
```
tweise: @mxm any preference regarding class name? I first had it as `FlinkPortableClientRunner` but maybe "runner" is misleading. On the other hand,  "entry point" is also redundant.

mxm: How about just `FlinkPortableRunner`? The functionality is closest to the classic `FlinkRunner`, as it allows for any portable pipeline to be submitted.

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkPortableClientEntryPoint.java
```diff
@@ -0,0 +1,249 @@
...
+  private int jobPort = 0; // pick any free port
+
+  public FlinkPortableClientEntryPoint(String driverCmd) {
+    Preconditions.checkState(
+        !driverCmd.contains(JOB_ENDPOINT_FLAG),
+        "Driver command must not contain " + JOB_ENDPOINT_FLAG);
+    this.driverCmd = driverCmd;
+  }
+
+  /** Main method to be called within the Flink OptimizerPlanEnvironment. */
```
mxm: This comment is obsolete, no?

runners/java-fn-execution/src/test/java/org/apache/beam/runners/fnexecution/environment/ProcessManagerTest.java
```diff
@@ -99,4 +106,53 @@ public void testEnvironmentVariables() throws IOException, InterruptedException
...
+    processManager.stopProcess("1");
+    byte[] output = Files.readAllBytes(outputFile.toPath());
+    assertNotNull(output);
+    String outputStr = new String(output, Charset.defaultCharset());
+    assertThat(outputStr, containsString("testing123"));
+  }
+
+  @Ignore("not working yet")
+  @Test
+  public void testInheritIO() throws IOException, InterruptedException {
```
tweise: @mxm I see the redirect working as the expected output is printed in the console. But short of starting a separate java process I don't see a way to capture it in the test.

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkPortableClientEntryPoint.java
tweise: I had it as XYZRunner originally, but figured it will cause confusion. There are `FlinkPipelineRunner` and `PortablePipelineRunner` already (these are used within the job server).

mxm: We could rename those 🤓

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkPortableClientEntryPoint.java
tweise: Yes, updated now.

runners/java-fn-execution/src/test/java/org/apache/beam/runners/fnexecution/environment/ProcessManagerTest.java
tweise: I left the test but skipped the assertion, maybe we can still find a clever way to do this.

PR #9880, opened 2019-10-25
[BEAM-8503] Improve TestBigQuery and TestPubsub

TheNeuralBit:
This PR adds some features to `TestBigQuery` and `TestPubsub`, and modifies some integration tests to use these new features.

### TestBigQuery
- Add ability to insert data in the test table prior to running a test.
- Can now assert on table data `now` as well as `eventually`.

### TestPubsub
- `TestPubsub` now initializes a randomized subscription when its created, so that users can pull all messages received during test execution.
- Adds `TestPubsub#{pull, waitForNMessages, assertThatTopicEventuallyReceives}` for accessing received messages.

### TestBigQueryReadWriteIT
- Original `testSQLRead*` tests are renamed to `testSQLReadWrite*` since they test separate Read and Write SQL pipelines.
- Add `testSQLRead` and `testSQLWrite` tests to just test read path and write path, respectively.

### PubsubJsonIT
- Modifies `testUsesDlq` to access dlq data with `TestPubsub#pull` instead of signaling (although this test is still ignored as I haven't verified that it no longer flakes).
- Adds `testWritesJsonRowsToPubsub`.

### DataCatalogBigQueryIT
- Removes `writeToBqPipeline` and instead uses `TestBigQuery#insertRows` at the start of the test.

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/pubsub/PubsubMessage.java
+++ b/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/pubsub/PubsubMessage.java
@@ -21,6 +21,7 @@ import static org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Prec
 
 import java.util.Map;
 import javax.annotation.Nullable;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.MoreObjects;
 
 /**
  * Class representing a Pub/Sub message. Each message contains a single message payload, a map of
@@ -65,5 +66,14 @@ public class PubsubMessage {
   @Nullable
   public String getMessageId() {
     return messageId;
+  }
+
+  @Override
+  public String toString() {
+    return MoreObjects.toStringHelper(this)
+        .add("message", message)
+        .add("attributes", attributes)
+        .add("messageId", messageId)
+        .toString();
   }
 }
--- a/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/pubsub/TestPubsub.java
+++ b/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/pubsub/TestPubsub.java
@@ -30,6 +30,7 @@ import org.apache.beam.sdk.io.gcp.pubsub.PubsubClient.SubscriptionPath;
 import org.apache.beam.sdk.io.gcp.pubsub.PubsubClient.TopicPath;
 import org.apache.beam.sdk.testing.TestPipeline;
 import org.apache.beam.sdk.testing.TestPipelineOptions;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.ImmutableList;
 import org.joda.time.DateTime;
 import org.joda.time.Duration;
 import org.joda.time.Instant;
@@ -41,9 +42,10 @@ import org.junit.runner.Description;
 import org.junit.runners.model.Statement;
 
 /**
- * Test rule which creates a new topic with randomized name and exposes the APIs to work with it.
+ * Test rule which creates a new topic and subscription with randomized names and exposes the APIs
+ * to work with them.
  *
- * <p>Deletes topic on shutdown.
+ * <p>Deletes topic and subscription on shutdown.
  */
 public class TestPubsub implements TestRule {
   private static final DateTimeFormatter DATETIME_FORMAT =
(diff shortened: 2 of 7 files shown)


TheNeuralBit:
Run SQL PostCommit

TheNeuralBit:
Run SQL PostCommit

TheNeuralBit:
R: @kennknowles

TheNeuralBit:
Run Java PreCommit

TheNeuralBit:
Run Java PreCommit

TheNeuralBit:
Run SQL PostCommit

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/pubsub/TestPubsub.java
```diff
@@ -172,6 +193,65 @@ public void publish(List<PubsubMessage> messages) throws IOException {
...
   }
 
+  /** Pull up to 100 messages from {@link #subscriptionPath()}. */
+  public List<PubsubMessage> pull() throws IOException {
+    return pull(100);
+  }
+
+  /** Pull up to {@code maxBatchSize} messages from {@link #subscriptionPath()}. */
+  public List<PubsubMessage> pull(int maxBatchSize) throws IOException {
+    return pubsub.pull(0, subscriptionPath, 100, true).stream()
```
kennknowles: For the `pull` methods, two things that make sense to me are to have no default and/or to match the defaults of a public Pubsub library. Since `PubsubClient` here is just something we built for Beam's purposes, I wouldn't try to match its API. TBH it seems like the native Pubsub API described at https://cloud.google.com/pubsub/docs/pull might be a better fit to avoid writing your own pulling loop. I assume `PubsubClient` adds conveniences so this may not be a straightforward decision. I'm OK either way, but something to consider.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/pubsub/PubsubMessage.java
```diff
@@ -66,4 +67,17 @@ public String getAttribute(String attribute) {
   public String getMessageId() {
     return messageId;
   }
+
+  @Override
+  public String toString() {
```
kennknowles: `MoreObjects.toStringHelper()` provides a nice builder for a very predictable format in cases like this.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/pubsub/TestPubsub.java
TheNeuralBit: I don't see anything in the public pubsub API that saves me from writing my own pulling loop. The closest thing to what I've done here is the [synchronous pull example](https://cloud.google.com/pubsub/docs/pull#synchronous_pull), but that just does a single pull, no loop to wait for N messages. It also acknowledges the messages, which I wasn't doing.. I've added that now.

All I really want is a simple way to wait for the subscription to receive N messages, and then make an assertion about their content, so I can fluently test pipelines that write to pubsub. The `pull` methods could be a private implementation detail, would that be better? Actually the one that defaults to maxBatchSize = 100 could be removed entirely in that case.

I could also look at switching this from `PubsubClient` to the public library, it doesn't look like it would be too hard.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/pubsub/PubsubMessage.java
TheNeuralBit: Done.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/pubsub/TestPubsub.java
kennknowles: My impression from (very briefly) peeking at their docs was that you could set up an asynchronous pull. This is very low stakes, while the value of your change is high. So I would not block a merge on this. Let's just move forwards.

TheNeuralBit: Ah I see. I dismissed the asynchronous option since it would require some sort of synchronization, but that may be simpler. I can look at doing that as a follow-up.

TheNeuralBit:
Thanks that's a great tip! Here I thought people were just adding that "fixup!" for fun, I didn't realize it worked with `git rebase --autosquash`

kennknowles:
run sql postcommit

PR #9885, opened 2019-10-25
[BEAM-8457] Label Dataflow jobs from Notebook

KevinGG:
1. Added a standalone interactive utils module that provides functionality to detect if current execution is in an interactive notebook environment.
2. Implicitly append a user label (goog-dataflow-notebook=$beam_version) to Dataflow jobs started from interactive notebook environment when the user does DataflowRunner().run_pipeline(p, options).
3. User '--labels' doesn't support character '.' or '"'. When applying version related label, replace '.' with '_' and avoid enclosing any label with '"'.
4. Fixed try-except condition in static imports of display_manager module. Future PRs will remove the usage of display_manager from interactive runner.
------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/pipeline.py
+++ b/sdks/python/apache_beam/pipeline.py
@@ -169,6 +169,10 @@ class Pipeline(object):
     # If a transform is applied and the full label is already in the set
     # then the transform will have to be cloned with a new label.
     self.applied_labels = set()
+    # A boolean value indicating whether the pipeline is created in an
+    # interactive environment such as interactive notebooks. Initialized as
+    # None. The value is set ad hoc when `pipeline.run()` is invoked.
+    self.interactive = None
 
   @property
   @deprecated(since='First stable release',
@@ -405,28 +409,57 @@ class Pipeline(object):
     for override in replacements:
       self._check_replacement(override)
 
-  def run(self, test_runner_api=True):
-    """Runs the pipeline. Returns whatever our runner returns after running."""
+  def run(self, test_runner_api=True, runner=None, options=None,
+          interactive=None):
+    """Runs the pipeline. Returns whatever our runner returns after running.
 
+    If another runner instance and options are provided, that runner will
+    execute the pipeline with the given options. If either of them is not set,
+    a ValueError is raised. The usage is similar to directly invoking
+    `runner.run_pipeline(pipeline, options)`.
+    Additionally, an interactive field can be set to override the pipeline's
+    self.interactive field to mark current pipeline as being initiated from an
+    interactive environment.
+    """
+    if interactive:
+      self.interactive = interactive
+    elif (type(self.runner).__module__
+          == 'apache_beam.runners.interactive.interactive_runner' and
+          type(self.runner).__name__ == 'InteractiveRunner'):
+      self.interactive = True
+    else:
+      self.interactive = False
+    runner_in_use = self.runner
+    options_in_use = self._options
+    if runner and options:
+      runner_in_use = runner
+      options_in_use = options
+    elif not runner and options:
+      raise ValueError('Parameter runner is not given when parameter options '
+                       'is given.')
+    elif not options and runner:
+      raise ValueError('Parameter options is not given when parameter runner '
+                       'is given.')
     # When possible, invoke a round trip through the runner API.
     if test_runner_api and self._verify_runner_api_compatible():
       return Pipeline.from_runner_api(
           self.to_runner_api(use_fake_coders=True),
-          self.runner,
-          self._options).run(False)
+          runner_in_use,
+          options_in_use).run(test_runner_api=False,
+                              interactive=self.interactive)
 
-    if self._options.view_as(TypeOptions).runtime_type_check:
+    if options_in_use.view_as(TypeOptions).runtime_type_check:
       from apache_beam.typehints import typecheck
       self.visit(typecheck.TypeCheckVisitor())
 
-    if self._options.view_as(SetupOptions).save_main_session:
+    if options_in_use.view_as(SetupOptions).save_main_session:
       # If this option is chosen, verify we can pickle the main session early.
       tmpdir = tempfile.mkdtemp()
       try:
         pickler.dump_session(os.path.join(tmpdir, 'main_session.pickle'))
       finally:
         shutil.rmtree(tmpdir)
-    return self.runner.run_pipeline(self, self._options)
+    return runner_in_use.run_pipeline(self, options_in_use)
 
   def __enter__(self):
     return self
--- a/sdks/python/apache_beam/pipeline.py
+++ b/sdks/python/apache_beam/pipeline.py
@@ -169,10 +169,6 @@ class Pipeline(object):
     # If a transform is applied and the full label is already in the set
     # then the transform will have to be cloned with a new label.
     self.applied_labels = set()
-    # A boolean value indicating whether the pipeline is created in an
-    # interactive environment such as interactive notebooks. Initialized as
-    # None. The value is set ad hoc when `pipeline.run()` is invoked.
-    self.interactive = None
 
   @property
   @deprecated(since='First stable release',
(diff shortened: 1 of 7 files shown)


sdks/python/apache_beam/pipeline.py
```diff
@@ -396,28 +400,57 @@ def replace_all(self, replacements):
...
+    `runner.run_pipeline(pipeline, options)`.
+    Additionally, an interactive field can be set to override the pipeline's
+    self.interactive field to mark current pipeline as being initiated from an
+    interactive environment.
+    """
+    if interactive:
+      self.interactive = interactive
+    elif (type(self.runner).__module__
+          == 'apache_beam.runners.interactive.interactive_runner' and
+          type(self.runner).__name__ == 'InteractiveRunner'):
```
KevinGG: This is the difference from previous [PR](https://www.google.com/url?q=https://github.com/apache/beam/pull/9854). 
All runners are using  "new-style" classes in Python, the `type(obj).__module__/__name__` should always work. Please let me know if there would be backward incompatible cases. 
Thanks!

KevinGG:
Run Python PreCommit

KevinGG:
Run Python PreCommit

KevinGG:
Run Python PreCommit

KevinGG:
Run Python PreCommit

KevinGG:
R: @aaltay 
R: @pabloem 
Sorry for the previous rollback. I've removed the dependency of interactive_runner from pipeline, so there shouldn't be import issues now. Thanks!

aaltay:
R: @pabloem could you make a first review pass?

pabloem:
This is pretty much the same PR as before, except it checks the runner via string matching. It LGTM.

sdks/python/apache_beam/pipeline.py
```diff
@@ -172,6 +172,10 @@ def __init__(self, runner=None, options=None, argv=None):
     # If a transform is applied and the full label is already in the set
     # then the transform will have to be cloned with a new label.
     self.applied_labels = set()
+    # A boolean value indicating whether the pipeline is created in an
+    # interactive environment such as interactive notebooks. Initialized as
+    # None. The value is set ad hoc when `pipeline.run()` is invoked.
+    self.interactive = None
```
aaltay: I believe one more suggestion from the previous PR was to not keepting track of interactivity as part of the Pipeline but making it a property of the runner.

sdks/python/apache_beam/pipeline.py
```diff
@@ -396,28 +400,57 @@ def replace_all(self, replacements):
     for override in replacements:
       self._check_replacement(override)
 
-  def run(self, test_runner_api=True):
-    """Runs the pipeline. Returns whatever our runner returns after running."""
-
+  def run(self, test_runner_api=True, runner=None, options=None,
```
aaltay: Why are we adding runner and options parameters here?

sdks/python/apache_beam/runners/dataflow/dataflow_runner.py
```diff
@@ -360,6 +360,16 @@ def visit_transform(self, transform_node):
 
   def run_pipeline(self, pipeline, options):
     """Remotely executes entire pipeline or parts reachable from node."""
+    # Label goog-dataflow-notebook if pipeline is initiated from interactive
+    # runner.
+    if pipeline.interactive:
```
aaltay: The change could be limited to:
- here detect, whether we are in interactive environment or not. (For example, check whether certain imports are loaded?) Could also be a utility method somewhere, to check whether this in an interactive method.
- If yes, add the labels.

sdks/python/apache_beam/pipeline.py
KevinGG: Thanks! I'll go with your suggestion below to deduce if the job is started from a notebook environment instead of determining if the job is started from a pipeline that was originally bundled with an Interactive Runner.

sdks/python/apache_beam/pipeline.py
KevinGG: IIRC, we want to allow the user to switch to `DataflowRunner` using the `p.run()` pattern instead of limiting the user to `Runner().run_pipeline(p, options)`.

Do you think we should put it into a separate PR, or simply not supporting it at all?

sdks/python/apache_beam/runners/dataflow/dataflow_runner.py
KevinGG: I see your point! Yes, I have the [capability](https://github.com/apache/beam/blob/master/sdks/python/apache_beam/runners/interactive/interactive_environment.py#L131) to check if current interpreted code is in a notebook or not. This branch will need a rebase against master to take those changes.

To clartify the process:
When a DataflowRunner tries to run a job from a given pipeline,
1. Check if the module `interactive_environment` is imported by checking the `sys.modules` dictionary;
2. Check if `current_env().is_in_notebook`;
3. If yes, label the job.

aaltay: Could we move https://github.com/apache/beam/blob/master/sdks/python/apache_beam/runners/interactive/interactive_environment.py#L100:L102 to a common utility function, and each runner if they want could call this without worrying about require additional imports?

sdks/python/apache_beam/pipeline.py
aaltay: > Do you think we should put it into a separate PR,

Yes. At least let's have a seperate discussion for API changes like this.

>  or simply not supporting it at all?

Maybe not. This could be just a override for the interactive runners run() (e.g. run_with(NewRunner, NewOptions). At least, let's discuss with all stakeholders.

sdks/python/apache_beam/pipeline.py
KevinGG: I'll go with the check environment route and make it a standalone utility module in the interactive package.

sdks/python/apache_beam/pipeline.py
KevinGG: Putting this discussion on next Monday's agenda and will remove changes to the API.

sdks/python/apache_beam/runners/dataflow/dataflow_runner.py
KevinGG: Discussed with David and Sam. Since we also want to track jobs started from notebook even if the user never uses `InteractiveRunner`, checking the environment might just be the only way to do it.
By putting the logic into a try-except block as it is, we could avoid introducing `ipython` dependency into `DataflowRunner`. If the `[interactive]` dependency is never installed and current execution_path has never imported `ipython`, the code would just never be executed.

I'll move the logic into a standalone utility module and import it in DataflowRunner to do the check.

KevinGG:
Run Portable_Python PreCommit

KevinGG:
Run Portable_Python PreCommit

KevinGG:
Run Portable_Python PreCommit

sdks/python/apache_beam/runners/interactive/interactive_environment.py
```diff
@@ -93,17 +94,7 @@ def __init__(self, cache_manager=None):
...
-    # Check if the runtime is within an interactive environment, i.e., ipython.
-    try:
-      from IPython import get_ipython  # pylint: disable=import-error
-      if get_ipython():
-        self._is_in_ipython = True
-        if 'IPKernelApp' in get_ipython().config:
-          self._is_in_notebook = True
-    except ImportError:
-      pass
+    self._is_in_ipython, self._is_in_notebook = is_interactive()
```
robertwb: Conventionally, `is_xxx` functions return a boolean. Returning a pair will be especially surprising if one writes statements like `if is_interactive()` and the return value is `(False, False)` (which as a non-zero-length tuple evaluates to `True`.

KevinGG: Roger, will make it into 2 separate APIs.

PR #9918, opened 2019-10-29
[BEAM-8507] Fix 'flink_master' parameter for the upcoming release

mxm:
## Add '[auto]' as a supported execution mode for Python FlinkRunner

The `[auto]` mode would work if the Flink job server had a config with the cluster
address. Change default back to '[local]' until it is clear how exactly '[auto]'
would be used.

## Support http protocol scheme for Flink master address

Python uses the requests library to communicate with the Flink master via the
REST API and requires a protocol scheme.

Despite also using the REST API, the Java client does not expect the master
address to contain a protocol scheme. In order to support Python specifying one
in Python, we can remove it in Java.

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/runners/portability/flink_runner.py
+++ b/sdks/python/apache_beam/runners/portability/flink_runner.py
@@ -30,6 +30,7 @@ from apache_beam.runners.portability import job_server
 from apache_beam.runners.portability import portable_runner
 
 PUBLISHED_FLINK_VERSIONS = ['1.7', '1.8', '1.9']
+MAGIC_HOST_NAMES = ['[local]', '[auto]']
 
 
 class FlinkRunner(portable_runner.PortableRunner):
@@ -37,19 +38,20 @@ class FlinkRunner(portable_runner.PortableRunner):
     flink_master = self.add_http_scheme(
         options.view_as(FlinkRunnerOptions).flink_master)
     options.view_as(FlinkRunnerOptions).flink_master = flink_master
-    if flink_master in ['[local]', '[auto]'] or sys.version_info < (3, 6):
-      portable_options = options.view_as(pipeline_options.PortableOptions)
-      if flink_master == '[local]' and not portable_options.environment_type:
-        portable_options.environment_type = 'LOOPBACK'
+    if flink_master in MAGIC_HOST_NAMES or sys.version_info < (3, 6):
       return job_server.StopOnExitJobServer(FlinkJarJobServer(options))
     else:
+      # This has to be changed [auto], otherwise we will attempt to submit a
+      # the pipeline remotely on the Flink JobMaster which will _fail_.
+      # DO NOT CHANGE the following line, unless you have tested this.
+      options.view_as(FlinkRunnerOptions).flink_master = '[auto]'
       return flink_uber_jar_job_server.FlinkUberJarJobServer(flink_master)
 
   @staticmethod
   def add_http_scheme(flink_master):
     """Adds a http protocol scheme if none provided."""
     flink_master = flink_master.strip()
-    if not flink_master in ['[local]', '[auto]'] and \
+    if not flink_master in MAGIC_HOST_NAMES and \
           not re.search('^http[s]?://', flink_master):
       logging.info('Adding HTTP protocol scheme to flink_master parameter: '
                    'http://%s', flink_master)
--- a/sdks/python/apache_beam/runners/portability/flink_runner.py
+++ b/sdks/python/apache_beam/runners/portability/flink_runner.py
@@ -20,6 +20,8 @@
 from __future__ import absolute_import
 from __future__ import print_function
 
+import logging
+import re
 import sys
 
 from apache_beam.options import pipeline_options
@@ -32,7 +34,9 @@ PUBLISHED_FLINK_VERSIONS = ['1.7', '1.8', '1.9']
 
 class FlinkRunner(portable_runner.PortableRunner):
   def default_job_server(self, options):
-    flink_master = options.view_as(FlinkRunnerOptions).flink_master
+    flink_master = self.add_http_scheme(
+        options.view_as(FlinkRunnerOptions).flink_master)
+    options.view_as(FlinkRunnerOptions).flink_master = flink_master
     if flink_master in ['[local]', '[auto]'] or sys.version_info < (3, 6):
       portable_options = options.view_as(pipeline_options.PortableOptions)
       if flink_master == '[local]' and not portable_options.environment_type:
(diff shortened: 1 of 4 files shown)


sdks/python/apache_beam/runners/portability/flink_runner.py
```diff
@@ -32,26 +33,39 @@
...
   @classmethod
   def _add_argparse_args(cls, parser):
     parser.add_argument('--flink_master',
-                        default='[auto]',
-                        help='Flink master address (host:port) to submit the'
-                             ' job against. Use "[local]" to start a local'
-                             ' cluster for the execution. Use "[auto]" if you'
-                             ' plan to either execute locally or submit through'
-                             ' Flink\'s CLI.')
+                        default='[local]',
```
tweise: Per discussion the default should remain '[auto]'?

mxm: The default was `[local]`. We had changed it to auto, but we don't know yet what that means in the context of `flink_runner.py`. So I'd suggest to leave it at `[local]` for now.

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkJobInvoker.java
```diff
@@ -66,7 +66,9 @@ protected JobInvocation invokeWithExecutor(
         String.format("%s_%s", flinkOptions.getJobName(), UUID.randomUUID().toString());
 
     if (FlinkPipelineOptions.AUTO.equals(flinkOptions.getFlinkMaster())) {
-      flinkOptions.setFlinkMaster(serverConfig.getFlinkMaster());
+      // Although Flink uses Rest, it expects the address not to contain a http scheme
+      String flinkMaster = serverConfig.getFlinkMaster().trim().replaceFirst("^http://", "");
```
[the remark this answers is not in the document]
mxm: Not sure whether we should allow `https`, since it is not supported anyway. I'm ok with adding support, if we figure out a way for Python to support it.

sdks/python/apache_beam/runners/portability/flink_runner.py
mxm: Yes, this is the purpose of this PR and fixed here. Changing the default to `[auto]` broke it in the first place. Now it will work for `[auto]` or `[local]`. Still the more sensible default is probably `[local]` for Python at the moment. There is no real use for `[auto]` unless the job server is brought up with a config.

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkJobInvoker.java
```diff
@@ -66,7 +67,9 @@ protected JobInvocation invokeWithExecutor(
         String.format("%s_%s", flinkOptions.getJobName(), UUID.randomUUID().toString());
 
     if (FlinkPipelineOptions.AUTO.equals(flinkOptions.getFlinkMaster())) {
-      flinkOptions.setFlinkMaster(serverConfig.getFlinkMaster());
+      // Although Flink uses Rest, it expects the address not to contain a http scheme
+      String flinkMaster = stripHttpSchema(serverConfig.getFlinkMaster());
```
ibzib: I'm not sure why we should strip the schema in this particular place?

sdks/python/apache_beam/runners/portability/flink_runner.py
mxm: On this note, we should add tests for `flink_runner.py`. Currently there are none: https://jira.apache.org/jira/browse/BEAM-8512

tweise: The previous default of '[local]' that you restore here already had the above mentioned issue. This runner is not only used for the direct job submission but also for outputting generated jar files. We should verify that all the different submit mechanisms work.

ibzib: +1 I think these changes make sense. Although I would definitely like to see some tests (which I guess is [BEAM-8512](https://issues.apache.org/jira/browse/BEAM-8512))

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkJobInvoker.java
mxm: What is a better place?

ibzib: My understanding is that since you're stripping the schema in `FlinkExecutionEnvironments`, it shouldn't be necessary to also do it here.

sdks/python/apache_beam/runners/portability/flink_runner.py
mxm: >+1 I think these changes make sense. Although I would definitely like to see some tests (which I guess is BEAM-8512)

Tests for the Python side are out of scope for this PR. There are currently none and creating them will require some effort. We will have to follow-up with this.

ibzib: Sounds good.

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkJobInvoker.java
mxm: True, that is redundant.

sdks/python/apache_beam/runners/portability/flink_runner.py
tweise: @mxm yes, I believe that was the issue. I think the proper fix is to keep default at "auto" and change the behavior of the newly introduced Python REST API client to handle accordingly. Part of making changes here has to be verification. If there is no automation, then manual.

sdks/python/apache_beam/runners/portability/flink_runner.py
```diff
@@ -32,26 +34,39 @@
...
-    flink_master = options.view_as(FlinkRunnerOptions).flink_master
-    if flink_master == '[local]' or sys.version_info < (3, 6):
+    flink_master = self.add_http_scheme(
+        options.view_as(FlinkRunnerOptions).flink_master)
+    options.view_as(FlinkRunnerOptions).flink_master = flink_master
+    if flink_master in ['[local]', '[auto]'] or sys.version_info < (3, 6):
       portable_options = options.view_as(pipeline_options.PortableOptions)
       if flink_master == '[local]' and not portable_options.environment_type:
-        portable_options.environment_type == 'LOOPBACK'
+        portable_options.environment_type = 'LOOPBACK'
```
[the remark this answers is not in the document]
mxm: (I've verified that LOOPBACK execution works).

sdks/python/apache_beam/runners/portability/flink_runner.py
mxm: @tweise I've fully verified that the current version of this PR works under all deployment scenarios, i.e. Python 2 / Python < 3.6 / Python >= 3.6, with `flink_master` set to `[local]`, `[auto]`, or an actual address.

mxm:
Run Python PreCommit

sdks/python/apache_beam/runners/portability/flink_runner.py
tweise: I retested the jar file upload with the current changes and it now works.

I'm OK with all changes in the PR except the default value modification.

A default value of '[local]' is inconsistent with that in the Java Flink runner and such inconsistencies have bitten us already elsewhere.

There is no reason for the change since '[auto]' already does the right thing, per verification, and accurately reflects the behavior: Run locally except when in Flink context.

robertwb:
Thanks!

robertwb:
This change seems to have undone https://issues.apache.org/jira/browse/BEAM-8396 .

mxm:
>This change seems to have undone https://issues.apache.org/jira/browse/BEAM-8396.

I doubt it since only the following line was removed:

https://github.com/apache/beam/blob/49ef57be306096e54ef5527beb764d6b7f115aae/sdks/python/apache_beam/runners/portability/flink_runner.py#L39

This line is ineffective and even if there was only one `=`, it would still not work properly (I tested that). We need to set the loopback environment earlier for it to work (also tested that).

mxm:
It never seemed to have worked, but of course we can can/could still fix this.

PR #9932, opened 2019-10-30
[BEAM-8524] Use impulse instead of fake pubsub starting signal in Dataflow

boyuanzz:
Based on dataflow windmill changes, stop injecting pubsub into Impulse transform in fnapi streaming mode.

R: @lukecwik


boyuanzz:
Run Python PreCommit

lukecwik:
Run Python PreCommit

PR #9949, opened 2019-10-31
[BEAM-7948] Add time-based cache threshold support in the Java data s…

sunjincheng121:
Currently only size-based cache threshold is supported in data service. It should also support the time-based cache threshold. This is very important, especially for streaming jobs which are sensitive to the delay. So, This PR add the time-based cache threshold support in the Java data service.

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/fn-execution/src/main/java/org/apache/beam/sdk/fn/data/BeamFnDataBufferingOutboundObserver.java
+++ b/sdks/java/fn-execution/src/main/java/org/apache/beam/sdk/fn/data/BeamFnDataBufferingOutboundObserver.java
@@ -17,14 +17,14 @@
  */
 package org.apache.beam.sdk.fn.data;
 
-import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
 import org.apache.beam.model.fnexecution.v1.BeamFnApi;
 import org.apache.beam.sdk.coders.Coder;
-import org.apache.beam.vendor.grpc.v1p21p0.com.google.protobuf.ByteString;
+import org.apache.beam.sdk.options.ExperimentalOptions;
+import org.apache.beam.sdk.options.PipelineOptions;
 import org.apache.beam.vendor.grpc.v1p21p0.io.grpc.stub.StreamObserver;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.annotations.VisibleForTesting;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
  * A buffering outbound {@link FnDataReceiver} for the Beam Fn Data API.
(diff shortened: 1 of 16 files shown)


sdks/java/fn-execution/src/main/java/org/apache/beam/sdk/fn/data/BeamFnDataBufferingOutboundObserver.java
```diff
@@ -43,48 +57,83 @@
...
+  private final int sizeLimit;
   private final Coder<T> coder;
   private final LogicalEndpoint outputLocation;
   private final StreamObserver<BeamFnApi.Elements> outboundObserver;
   private final ByteString.Output bufferedElements;
+  @VisibleForTesting final ScheduledFuture<?> flushFuture;
+  private final AtomicReference<IOException> flushException = new AtomicReference<>(null);
+  private final Object lock = new Object();
 
   private BeamFnDataBufferingOutboundObserver(
```
lukecwik: I think we will want to have the original  implementation of this class so that we don't need to take on the synchronization overhead for flush.

sdks/java/fn-execution/src/test/java/org/apache/beam/sdk/fn/data/BeamFnDataBufferingOutboundObserverTest.java
```diff
@@ -146,6 +159,85 @@ public void testConfiguredBufferLimit() throws Exception {
...
+            options,
+            OUTPUT_LOCATION,
+            CODER,
+            TestStreams.withOnNext(addToValuesConsumer(values))
+                .withOnCompleted(setBooleanToTrue(onCompletedWasCalled))
+                .build());
+
+    // Test that it emits when time passed the time limit
+    consumer.accept(valueInGlobalWindow(new byte[1]));
+    Thread.sleep(10); // allow the flush thread to flush the buffer
```
lukecwik: This test will be flaky since its based upon very small amounts of time. You want to make withOnNext free a lock/semaphore/countdownlatch and in your test you wait on it with a large time limit instead of using thread.sleep

sdks/java/fn-execution/src/test/java/org/apache/beam/sdk/fn/data/BeamFnDataBufferingOutboundObserverTest.java
```diff
@@ -146,6 +159,85 @@ public void testConfiguredBufferLimit() throws Exception {
...
+                    (Consumer<Elements>)
+                        t -> {
+                          throw new RuntimeException("");
+                        })
+                .withOnCompleted(setBooleanToTrue(onCompletedWasCalled))
+                .build());
+
+    // Test that it emits when time passed the time limit
+    consumer.accept(valueInGlobalWindow(new byte[1]));
+    Thread.sleep(100); // allow the flush thread to flush the buffer
```
lukecwik: ditto here to use a semaphore/lock/countdownlatch instead of thread.sleep

sunjincheng121:
@lukecwik  Thanks a lot for your review. I have updated the PR and addressed your comments.

> I was wondering how this manifests?
> 
> Do you have a way to cancel downstream computations based upon the output of this bundle if this bundle fails?

If a bundle fails, the exception will be prograted to the main thread at the next time when an input element is processed(accept/close is called). So I think the failure handle logic for time-based data buffer strategy is the same as the size-based data buffer strategy. Time-based data buffer strategy has not and also should not introduce extra complexity for failure handling.

Looking forward to your feedback.

sunjincheng121:
Hi @lukecwik, Could you please have another look, any comment is welcome, Thanks! :)

sdks/java/fn-execution/src/main/java/org/apache/beam/sdk/fn/data/BeamFnDataTimeBasedBufferingOutboundObserver.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+  }
+
+  @Override
+  public void accept(T t) throws IOException {
+    checkFlushThreadException();
+    super.accept(t);
+  }
+
+  private void periodicFlush() {
```
lukecwik: I don't think you need flushException or this method at all, you can invoke `this::flush` from the scheduled executor directly and check the status of the flushFuture via the `isDone()` method. If its done then you know that an exception was thrown.

sdks/java/fn-execution/src/main/java/org/apache/beam/sdk/fn/data/BeamFnDataTimeBasedBufferingOutboundObserver.java
```diff
@@ -0,0 +1,109 @@
...
+      } else {
+        flushException.set(new IOException(t));
+      }
+      throw new RuntimeException(t);
+    }
+  }
+
+  /** Check if the flush thread failed with an exception. */
+  private void checkFlushThreadException() throws IOException {
+    IOException e = flushException.get();
```
lukecwik:
```
if (flushFuture.isDone()) {
  try {
    flushFuture.get();
    throw new IOException("Periodic flushing thread finished unexpectedly.");
  } catch (ExecutionException e) {
    if (e.getCause() instanceof IOException) {
      throw (IOException) e.getCause();
    }
    throw new RuntimeException(e);
  }
}
```

sdks/java/fn-execution/src/test/java/org/apache/beam/sdk/fn/data/BeamFnDataTimeBasedBufferingOutboundObserverTest.java
```diff
@@ -0,0 +1,139 @@
...
+    try {
+      // Test that the exception caught in the flush thread is propagate to
+      // the main thread when processing the next element
+      consumer.accept(valueInGlobalWindow(new byte[1]));
+      fail();
+    } catch (IOException e) {
+      // expected
+    }
+
+    consumer =
```
lukecwik: It seems like your duplicating the above portion of the test here unless I'm not seeing what the small difference is.

sdks/java/fn-execution/src/main/java/org/apache/beam/sdk/fn/data/BeamFnDataTimeBasedBufferingOutboundObserver.java
```diff
@@ -0,0 +1,109 @@
...
+
+  @Override
+  public void close() throws Exception {
+    checkFlushThreadException();
+    synchronized (lock) {
+      flushFuture.cancel(true);
+      try {
+        flushFuture.get();
+      } catch (CancellationException | ExecutionException | InterruptedException exn) {
+        // expected
```
lukecwik: CancellationException is the only exception that is expected. ExecutionException should be unwrapped and we should throw the cause (if set) otherwise we should throw the execution exception directly. InterruptedException should not be caught here at all.

sdks/java/fn-execution/src/main/java/org/apache/beam/sdk/fn/data/BeamFnDataTimeBasedBufferingOutboundObserver.java
```diff
@@ -0,0 +1,109 @@
...
+                    .build())
+            .scheduleAtFixedRate(this::periodicFlush, timeLimit, timeLimit, TimeUnit.MILLISECONDS);
+    this.flushException = new AtomicReference<>(null);
+  }
+
+  @Override
+  public void close() throws Exception {
+    checkFlushThreadException();
+    synchronized (lock) {
+      flushFuture.cancel(true);
```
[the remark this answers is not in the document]
sunjincheng121: This is done by holding the flush lock and so it's guaranteed that the flush thread is not flush data. I have renamed the "lock" to "flushLock" to make it more clear.

sdks/java/fn-execution/src/main/java/org/apache/beam/sdk/fn/data/BeamFnDataTimeBasedBufferingOutboundObserver.java
sunjincheng121: It's not possible to use the flush method directly as it throws an IOException. It's a good idea to use `flushFuture.isDone()` to check the status of the flush thread. Will be Updated. :)

sdks/java/fn-execution/src/test/java/org/apache/beam/sdk/fn/data/BeamFnDataTimeBasedBufferingOutboundObserverTest.java
sunjincheng121: The first test case verifies that the exception is propagated when processing the next element. The second test case verifies that the exception is propagated when closing.

lukecwik:
Run Java PreCommit

sdks/java/fn-execution/src/main/java/org/apache/beam/sdk/fn/data/BeamFnDataTimeBasedBufferingOutboundObserver.java
lukecwik: nit:
You can use the callable version of schedule to be able to use methods that throw exceptions:
https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ScheduledExecutorService.html#schedule-java.util.concurrent.Callable-long-java.util.concurrent.TimeUnit-

sdks/java/fn-execution/src/main/java/org/apache/beam/sdk/fn/data/BeamFnDataTimeBasedBufferingOutboundObserver.java
```diff
@@ -90,20 +90,27 @@ private void periodicFlush() {
...
+        throw new IOException("Periodic flushing thread finished unexpectedly.");
+      } catch (ExecutionException ee) {
+        // the cause of ExecutionException is always RuntimeException
+        RuntimeException re = (RuntimeException) ee.getCause();
+        if (re.getCause() instanceof IOException) {
+          throw (IOException) re.getCause();
+        }
+        throw re;
+      } catch (InterruptedException | CancellationException e) {
+        // Should never happen
```
lukecwik: Prefer to wrap the InerruptedException and CancellationException in an IOException.

Also for the interrupted exception you should invoke Thread.interrupt() before rethrowing.

sdks/java/fn-execution/src/main/java/org/apache/beam/sdk/fn/data/BeamFnDataTimeBasedBufferingOutboundObserver.java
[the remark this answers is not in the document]
lukecwik: You wouldn't need to have the flushLock object at all if you marked the flush method as synchronized and decided cancel the flushFuture without interrupting it.

sdks/java/fn-execution/src/main/java/org/apache/beam/sdk/fn/data/BeamFnDataTimeBasedBufferingOutboundObserver.java
```diff
@@ -48,25 +46,27 @@
...
     super(sizeLimit, outputLocation, coder, outboundObserver);
-    this.lock = new Object();
+    this.flushLock = new Object();
     this.flushFuture =
         Executors.newSingleThreadScheduledExecutor(
                 new ThreadFactoryBuilder()
                     .setDaemon(true)
                     .setNameFormat("DataBufferOutboundFlusher-thread")
                     .build())
             .scheduleAtFixedRate(this::periodicFlush, timeLimit, timeLimit, TimeUnit.MILLISECONDS);
```
lukecwik: You should be able to use the callable version of schedule:
https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ScheduledExecutorService.html#schedule-java.util.concurrent.Callable-long-java.util.concurrent.TimeUnit-
to invoke the flush method. You might need to provide a cast to Callable<Void> to tell Java to use that version of schedule.

sunjincheng121: I found that the callable version of schedule is not a periodic action. So, we have to create a new method which warps the `flush`, What do you think?

sunjincheng121:
I have update the PR accordingly except one comment I am not pretty sure, and left comment to you. :) Appreciate if you can have another look :) @lukecwik

sdks/java/fn-execution/src/main/java/org/apache/beam/sdk/fn/data/BeamFnDataSizeBasedBufferingOutboundObserver.java
```diff
@@ -85,7 +85,7 @@ public void close() throws Exception {
   }
 
   @Override
-  public void flush() throws IOException {
+  public synchronized void flush() throws IOException {
```
lukecwik: We don't want to make this synchronized since this class is not thread safe and should not take this perf hit. Your previous usage of defining an override that is synchronized for the class that called super.flush() is all you need.

sdks/java/fn-execution/src/main/java/org/apache/beam/sdk/fn/data/BeamFnDataTimeBasedBufferingOutboundObserver.java
lukecwik: Your right, I missed the fact that there was no periodic schedule that took a callable.

lukecwik:
Just remove the synchronized from BeamFnDataSizeBasedBufferingOutboundObserver.java and add
```
@Override
public synchronized void flush() throws IOException {
  super.flush();
}
```
to the time based outbound observer and we can merge.

sunjincheng121:
Thanks for the review @lukecwik 👍

PR #9955, opened 2019-10-31
[BEAM-2572] Python SDK S3 Filesystem

MattMorgis:
Co-authored-by: Matthew Morgis <matthew.morgis@gmail.com>
Co-authored-by: Tamera Lanham <t.lanham@elsevier.com>

This adds an AWS S3 file system implementation to the python SDK. 

------------------------

--- | --- | --- | --- | ---


MattMorgis:
Hi,

We are running into trouble getting the unit tests to pass in the CI environment, and I think we can use help from a core team member. 

We added a new set of extra dependencies when using this new S3 filesystem - we followed the same pattern that GCP did: https://github.com/apache/beam/pull/9955/files#diff-e9d0ab71f74dc10309a29b697ee99330R239

This allows the user to install with `pip install beam[gcp]` or `pip install beam[aws]` in our case. 

Our unit tests are completely mocked out and do not require any of the AWS extra packages, however, we set it up behind a flag so you can bypass the mock and talk to a real S3 bucket over the wire. Because of this, the extra dependencies *do* need to installed when running these new unit tests. 

Again, following the lead of how GCP implemented this, they also skip the unit tests if their extra dependencies are not installed: https://github.com/apache/beam/blob/master/sdks/python/apache_beam/io/gcp/gcsio_test.py#L240

Our question: _**How do we configured CI to install the AWS deps to run the tests?**_ 

I have poked around a bit and found one setting in `tox.ini` that appears to install both the test and gcp deps (https://github.com/apache/beam/blob/master/sdks/python/tox.ini#L200). Addtionally, at the root level of the project, (https://github.com/apache/beam/blob/master/buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy#L1799) I found a `installGcpTest` Gradle task that seems to also install both. This task only seems to be referenced inside of the `test-suites/dataflow` but not `direct` or `portable`. 

Any guidance here would be greatly appreciated!

MattMorgis:
R: @pabloem @robertwb @aaltay @charlesccychen

aaltay:
@MattMorgis thank you for the contribution. The general path of adding [aws] as an extra package sounds reasonable.

@chamikaramj could help with reviews or find a person to review.
@yifanzou could help with CI related questions.

chamikaramj:
R: @pabloem will you be able to review ?

Also cc: @lukecwik

yifanzou:
@markflyhigh would you help on the python test environment setup?

pabloem:
I can review. Looking today.

sdks/python/apache_beam/io/aws/clients/s3/messages.py
```diff
@@ -0,0 +1,166 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from __future__ import absolute_import
+
+
+class GetRequest():
```
pabloem: It seems like most of these messages could be implemented as namedtuple? You'd gain hashing and other utils, but also not a blocker.

sdks/python/apache_beam/io/aws/clients/s3/boto3_client.py
```diff
@@ -0,0 +1,235 @@
...
+
+    item = messages.Item(boto_response['ETag'],
+                         request.object,
+                         boto_response['LastModified'],
+                         boto_response['ContentLength'])
+
+    return item
+
+  def get_range(self, request, start, end):
+    r"""Retrieves an object.
```
pabloem:
```suggestion
    r"""Retrieves an object's contents.
```

sdks/python/apache_beam/io/aws/s3io.py
```diff
@@ -0,0 +1,604 @@
...
+           mode='r',
+           read_buffer_size=16*1024*1024,
+           mime_type='application/octet-stream'):
+    """Open an S3 file path for reading or writing.
+
+    Args:
+      filename (str): S3 file path in the form ``s3://<bucket>/<object>``.
+      mode (str): ``'r'`` for reading or ``'w'`` for writing.
+      read_buffer_size (int): Buffer size to use during read operations.
+      mime_type (str): Mime type to set for write operations.
```
pabloem: `mime_type` is ignored in this function. Maybe AWS always works with byte data? Should we verify that users sre requesting bytes?

sdks/python/apache_beam/utils/retry.py
```diff
@@ -46,8 +46,15 @@
 # TODO(sourabhbajaj): Remove the GCP specific error code to a submodule
```
pabloem: Would you file a JIRA issue and replace it in this line to remove the S3/GCS-specific errors from this file? We don't need to move them now, but it'd be nice to track them for later.

sdks/python/apache_beam/io/aws/s3io.py
tamera-lanham: This is addressed with the most recent changes, so the `mime_type` that the user sets will be reflected in the `ContentType` of the object in S3

pabloem:
There were some errors when creating the fake client:

```
14:34:21 ======================================================================
14:34:21 ERROR: test_delete_tree (apache_beam.io.aws.s3io_test.TestS3IO)
14:34:21 ----------------------------------------------------------------------
14:34:21 Traceback (most recent call last):
14:34:21   File "/home/jenkins/jenkins-slave/workspace/beam_PreCommit_Python_Commit/src/sdks/python/test-suites/tox/py2/build/srcs/sdks/python/apache_beam/io/aws/s3io_test.py", line 488, in test_delete_tree
14:34:21     self.assertTrue(self.aws.exists(path))
14:34:21   File "/home/jenkins/jenkins-slave/workspace/beam_PreCommit_Python_Commit/src/sdks/python/test-suites/tox/py2/build/srcs/sdks/python/apache_beam/io/aws/s3io.py", line 443, in exists
14:34:21     self.client.get_object_metadata(request)
14:34:21   File "/home/jenkins/jenkins-slave/workspace/beam_PreCommit_Python_Commit/src/sdks/python/test-suites/tox/py2/build/srcs/sdks/python/apache_beam/io/aws/clients/s3/fake_client.py", line 88, in get_object_metadata
14:34:21     return file_.get_metadata()
14:34:21   File "/home/jenkins/jenkins-slave/workspace/beam_PreCommit_Python_Commit/src/sdks/python/test-suites/tox/py2/build/srcs/sdks/python/apache_beam/io/aws/clients/s3/fake_client.py", line 49, in get_metadata
14:34:21     len(self.contents))
14:34:21 TypeError: __init__() takes exactly 6 arguments (5 given)
```

pabloem:
@MattMorgis @tamera-lanham There seems to be an issue with the fake client : ) would love to move this forward.

MattMorgis:
@pabloem I'll look into that and the `tox.ini` today or tomorrow

pabloem:
Hi! : ) it would be nice to get this in soon. Would you like us to jump on a call / any help pushing it through? : )

sdks/python/apache_beam/io/aws/clients/s3/fake_client.py
```diff
@@ -0,0 +1,235 @@
...
+  def get_metadata(self):
+    last_modified_datetime = None
+    if self.last_modified:
+      last_modified_datetime = datetime.datetime.utcfromtimestamp(
+          self.last_modified)
+
+    return messages.Item(self.etag,
+                         self.key,
+                         last_modified_datetime,
+                         len(self.contents))
```
pabloem: It seems that the problem is a missing argument here (`mime_type`). : ) @MattMorgis @tamera-lanham

tamera-lanham: Hey! Sorry it's been a bit since I've been in this codebase. A couple notes:
- In the constructor for `messages.Item`, `mime_type` is optional and defaults to `None` (skds/python/apache_beam/io/aws/clients/s3/messages.py:121), so it's surprising that this error is appearing. When I run the tests locally the behavior is what I'd expect - the `mime_type` gets set to `None` and the test passes, so I don't know why it isn't doing that in CI. I can't replicate the failure in my environment (which is Dockerized, if you want to try building it yourself!)
- The reason for using the default of `None` is because it was kind of onerous to build mime type support into the fake client. The only time mime type behavior is tested is in s3io_test.py, in `test_file_mime_type`, and that passes against the real client and is skipped for the fake. I like that metadata from the fake client comes out with a mime type of `None`, to indicate to whoever consumes the fake client that mime types aren't supported. Would you be ok with a default of `None` here instead of `application/octet-stream`? Whatever the default is, it shouldn't show up in any test anyway (except the one that we skip for the mock anyway, in which case we're skipping it)

pabloem: That's totally fine by me : ) - you're right, it seems that the default value for mime type in  `Item` was not set to None. I am okay with letting it be None. Thanks for taking a look.

pabloem:
Run Python PreCommit

tamera-lanham:
I'm checking out the test failures and spot-fixing now, starting with the python 2.7 issues. Seems like most of those can be fixed fairly easily. There are a couple of oddball failures as well in some python 3 environments which might take longer to figure out. I'll commit again when I think I've got some fixes in.

Also, do you know if there's a way to download a whole test report instead of navigating Jenkins to see the results?

pabloem:
Build scan: https://scans.gradle.com/s/tc45x4bplbo6m - only docs are complaining now : )

sdks/python/apache_beam/io/aws/clients/s3/boto3_client.py
```diff
@@ -0,0 +1,239 @@
...
+
+from apache_beam.io.aws.clients.s3 import messages
+
+try:
+  # pylint: disable=wrong-import-order, wrong-import-position
+  # pylint: disable=ungrouped-imports
+  import boto3
+
+except ImportError:
+  raise ImportError('Missing `boto3` requirement')
```
pabloem: Looking at: https://scans.gradle.com/s/tc45x4bplbo6m/console-log?task=:sdks:python:test-suites:tox:pycommon:docs
Seems like it minds this line. Maybe this is the issue?

```suggestion
  raise ImportError('Missing boto3 requirement')
```

sdks/python/apache_beam/io/aws/clients/s3/boto3_client.py
```diff
@@ -0,0 +1,239 @@
...
+
+from apache_beam.io.aws.clients.s3 import messages
+
+try:
+  # pylint: disable=wrong-import-order, wrong-import-position
+  # pylint: disable=ungrouped-imports
+  import boto3
+
+except ImportError:
+  raise ImportError('Missing boto3 requirement')
```
pabloem: Sorry. I think another approach for this line could be:
```suggestion
  boto3 = None
```

tamera-lanham: This file is the real, non-mocked boto3 client, so it only makes sense to use it with boto3 installed. If we catch the `ImportError` and silence it this way the user will just get a more cryptic error later (something like 'NoneType' object has no attribute 'client') which could be harder to debug. If it's our usage of a built-in error that's causing the problem, could we just change the type of error we're raising?

tamera-lanham: We can also just ignore the original import error, that is just delete lines 27 and 28 entirely.

pabloem: You could add this, and in the client `__init__` method (where boto3 is actually called), add a line with something like:

```
assert boto3 is not None, 'Missing boto3 requirement'
```

This would prevent this import-time error, and let the tests pass, and give a reasonable error message during client intialization rather than at import. Thoughts?

tamera-lanham: Works for me!

pabloem:
Looks like errors unrelated to the change. Let me clean up the GCP project that we use for testing

pabloem:

pabloem:
Run Python PreCommit

pabloem:
lovely!

pabloem:
Thanks so much @tamera-lanham @MattMorgis - y'all went the extra mile to write a good feature with testable code. Lots of people have wanted this feature added, so I'm very grateful to you two : )

aaltay:
Thank you all very much!

PR #9959, opened 2019-10-31
[BEAM-8523] JobAPI: Give access to timestamped state change history

chadrik:
When connecting to GetStateStream, the job service will first stream the history of missed state events, thereby catching up the client.

------------------------

--- | --- | --- | --- | ---


model/job-management/src/main/proto/beam_job_api.proto
```diff
@@ -153,6 +154,7 @@ message GetJobStateRequest {
 
 message GetJobStateResponse {
```
lukecwik: I think we will want a different version of GetJobStateRequest and GetJobStateResponse for the streaming and non-streaming versions of the state APIs.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/JobStateEvent.java
```diff
@@ -0,0 +1,55 @@
...
+package org.apache.beam.runners.fnexecution.jobsubmission;
+
+import com.google.auto.value.AutoValue;
+import java.time.Instant;
+import org.apache.beam.model.jobmanagement.v1.JobApi;
+import org.apache.beam.vendor.grpc.v1p21p0.com.google.protobuf.Timestamp;
+
+/** A state transition event. */
+@AutoValue
+public abstract class JobStateEvent {
```
lukecwik: We could rename the proto GetJobStateResponse to JobStateEvent and instead use that proto instead of creating this class.

model/job-management/src/main/proto/beam_job_api.proto
chadrik: Why?  What would be different about them?

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/JobStateEvent.java
chadrik: Do we want a `GetJobStateResponse` to have a `JobStateEvent` ?  i.e.

```proto
message GetJobStateResponse {
  JobStateEvent state_event = 1; // (required)
}
```

model/job-management/src/main/proto/beam_job_api.proto
chadrik: ok, saw your comment from Jira:

> Typically when you would connect you would want to give some lower bound to filter messages. For example GetJobStateRequest could become:
> 
> ```proto
> message GetJobStateRequest {
>   string job_id = 1; // (required)
>   // (Optional) If specified, only state transitions after
>   // the provided timestamp will be provided.
>   google.protobuf.Timestamp min_message_timestamp;
> }
> ```

I agree with this conceptually, but was hoping to defer the addition of this until I get to the message stream.  The state stream is typically going to be just a few events, so practically speaking, the additional of `min_message_timestamp` here will not be very impactful.  The size of the backfill for the message stream on the other hand, with the changes that I'd like to add to it, could be significant, and there could be a number of ways that a user may want to filter (`min_message_timestamp`, `max_num_messages`, `importance`, etc).  I'd like the APIs to be similar between the two, and I think the GetMessageStream will inform what we want to do for GetStateStream. 

tl;dr I'd like to defer the addition of this until later, if possible.

chadrik:
Note that this PR depends on https://github.com/apache/beam/pull/9965 which depends on https://github.com/apache/beam/pull/9969

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/JobStateEvent.java
chadrik: > use that proto instead of creating this class.

I've noticed that there is a pattern in the Java and Python SDKs to avoid using protobuf objects directly as part of the Beam SDK, and instead favor "native" types with methods for converting to/from protobuf representations.  I was trying to maintain that separation by providing this `JobStateEvent` class, but I'm happy to use a protobuf object directly.  Let me know what you think.

chadrik: > We could rename the proto GetJobStateResponse to JobStateEvent and instead use that proto instead of creating this class.

Done.

chadrik:
This is ready for another look.

chadrik:
Run Python PreCommit

chadrik:
I reworked this a bit to make it easier for other job services to implement state history, and I added some tests for the local job service.

chadrik:
R: @mxm 
R: @lukecwik

chadrik:
Run Python PreCommit

chadrik:
Run Python PreCommit

chadrik:
Run Python PreCommit

chadrik:
Run Python PreCommit

chadrik:
Run Python PreCommit

sdks/python/apache_beam/runners/portability/abstract_job_service.py
```diff
@@ -115,13 +127,43 @@ def __init__(self, job_id, job_name, pipeline, options):
...
+  def set_state(self, new_state):
+    """Set the latest state as an int enum and update the state history.
+
+    :param new_state: int
+      latest state enum
+    :return: float or None
+      the new timestamp if the state has not changed, else None
+    """
+    if new_state != self._state_history[-1][0]:
+      timestamp = time.time()
```
lukecwik: Consider using Beam's timestamp object instead of the native time object in Python since it already supports to/from proto: https://github.com/apache/beam/blob/b446304f75078ca9c97437e685409c31ceab7503/sdks/python/apache_beam/utils/timestamp.py#L145

This will prevent the proliferation of multiple time like objects.

chadrik: yeah, good suggestion.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/JobInvocation.java
```diff
@@ -70,8 +72,9 @@ public JobInvocation(
     this.stateObservers = new ArrayList<>();
     this.messageObservers = new ArrayList<>();
     this.invocationFuture = null;
-    this.jobState = JobState.Enum.STOPPED;
+    this.stateHistory = new ArrayList<>();
```
mxm: Does this grow boundlessly?

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/JobInvocation.java
```diff
@@ -200,8 +208,10 @@ public void onFailure(Throwable throwable) {}
   }
 
   /** Listen for job state changes with a {@link Consumer}. */
-  public synchronized void addStateListener(Consumer<JobState.Enum> stateStreamObserver) {
-    stateStreamObserver.accept(getState());
+  public synchronized void addStateListener(Consumer<JobStateEvent> stateStreamObserver) {
+    for (JobStateEvent event : stateHistory) {
+      stateStreamObserver.accept(event);
```
mxm: Couldn't we remove the event once we have informed the observers?

model/job-management/src/main/proto/beam_job_api.proto
```diff
@@ -151,8 +152,9 @@ message GetJobStateRequest {
 
 }
 
-message GetJobStateResponse {
+message JobStateEvent {
   JobState.Enum state = 1; // (required)
+  google.protobuf.Timestamp timestamp = 2; // (required)
```
mxm: Why is the timestamp not set a the receiving side? Do we need to include this in the Proto?

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/JobInvocation.java
chadrik: yes, but based on the feedback from @lukecwik in the last job-state PR, there are no runners that can switch back and forth between states, like STOPPED -> RUNNING -> STOPPED -> RUNNING, so the history should always be quite limited.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/jobsubmission/JobInvocation.java
chadrik: There may be many clients connecting to this stream, and thus many potential observers interested in receiving the full list of state history.

model/job-management/src/main/proto/beam_job_api.proto
chadrik: One of the main users stories behind this PR is that there could be multiple clients, and some may join the stream late, so those late joining clients won't know the timestamps.

chadrik:
Run Python PreCommit

chadrik:
Run Portable_Python PreCommit

sdks/python/apache_beam/runners/portability/abstract_job_service.py
```diff
@@ -16,16 +16,36 @@
...
 
 _LOGGER = logging.getLogger(__name__)
 
 
+def make_state_event(state, timestamp):
+  if isinstance(timestamp, Timestamp):
+    proto_timestamp = timestamp.to_proto()
+  elif isinstance(timestamp, timestamp_pb2.Timestamp):
+    proto_timestamp = timestamp
+  elif isinstance(timestamp, float):
```
lukecwik: nit: I thought you would have removed this float -> proto conversion and users would always use Beam's Timestamp.

chadrik: Fixed.   At the time I thought it could be useful for an end-user implementing their own job service, but it's not too much to ask call `Timestamp.now()`.

chadrik:
All tests passing!

lostluck:
Forward fixed in https://github.com/apache/beam/pull/10280

PR #9942, opened 2019-10-30
[BEAM-3288] Do not drop data after trigger finishes

kennknowles:
Quoting https://lists.apache.org/thread.html/d400848b3ead1a5cf4093328a6ea73082c60823039250d104fd0d588@%3Cdev.beam.apache.org%3E

Problem: a trigger can "finish" which causes a window to "close" and drop
all remaining data arriving for that window.

This has been discussed many times and I thought fixed, but it seems to not
be fixed. It does not seem to have its own Jira or thread that I can find.
But here are some pointers:

 - data loss bug:
https://lists.apache.org/thread.html/ce413231d0b7d52019668765186ef27a7ffb69b151fdb34f4bf80b0f@%3Cdev.beam.apache.org%3E
 - user hitting the bug:
https://lists.apache.org/thread.html/28879bc80cd5c7ef1a3e38cb1d2c063165d40c13c02894bbccd66aca@%3Cuser.beam.apache.org%3E
 - user confusion:
https://lists.apache.org/thread.html/2707aa449c8c6de1c6e3e8229db396323122304c14931c44d0081449@%3Cuser.beam.apache.org%3E
 - thread from 2016 on the topic:
https://lists.apache.org/thread.html/5f44b62fdaf34094ccff8da2a626b7cd344d29a8a0fff6eac8e148ea@%3Cdev.beam.apache.org%3E

In theory, trigger finishing was intended for users who can get their
answers from a smaller amount of data and then drop the rest. In practice,
triggers aren't really expressive enough for this. Stateful DoFn is the
solution for these cases.

I've opened https://github.com/apache/beam/pull/9942 which makes the
following changes:

 - when a trigger says it is finished, it never fires again but data is
still kept
 - at GC time the final output will be emitted

As with all bugfixes, this is backwards-incompatible (if your pipeline
relies on buggy behavior, it will stop working). So this is a major change
that I wanted to discuss on dev@.

------------------------

Thank you for your contribution! Follow this checklist to help us incorporate your contribution quickly and easily:

 - [ ] [**Choose reviewer(s)**](https://beam.apache.org/contribute/#make-your-change) and mention them in a comment (`R: @username`).
 - [x] Format the pull request title like `[BEAM-XXX] Fixes bug in ApproximateQuantiles`, where you replace `BEAM-XXX` with the appropriate JIRA issue, if applicable. This will automatically link the pull request to the issue.
[...]

PR #9960, opened 2019-10-31
[BEAM-3288] Guard against unsafe triggers at construction time

kennknowles:
Some triggers can cause data loss. This PR prevents them from being used.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/GroupByKey.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/GroupByKey.java
@@ -22,9 +22,12 @@ import org.apache.beam.sdk.coders.Coder.NonDeterministicException;
 import org.apache.beam.sdk.coders.IterableCoder;
 import org.apache.beam.sdk.coders.KvCoder;
 import org.apache.beam.sdk.transforms.display.DisplayData;
+import org.apache.beam.sdk.transforms.windowing.AfterWatermark.AfterWatermarkEarlyAndLate;
+import org.apache.beam.sdk.transforms.windowing.AfterWatermark.FromEndOfWindow;
 import org.apache.beam.sdk.transforms.windowing.DefaultTrigger;
 import org.apache.beam.sdk.transforms.windowing.GlobalWindows;
 import org.apache.beam.sdk.transforms.windowing.InvalidWindows;
+import org.apache.beam.sdk.transforms.windowing.Never.NeverTrigger;
 import org.apache.beam.sdk.transforms.windowing.TimestampCombiner;
 import org.apache.beam.sdk.transforms.windowing.Window;
 import org.apache.beam.sdk.transforms.windowing.WindowFn;
@@ -151,9 +154,8 @@ public class GroupByKey<K, V>
         && windowingStrategy.getTrigger() instanceof DefaultTrigger
         && input.isBounded() != IsBounded.BOUNDED) {
       throw new IllegalStateException(
-          "GroupByKey cannot be applied to non-bounded PCollection in "
-              + "the GlobalWindow without a trigger. Use a Window.into or Window.triggering transform "
-              + "prior to GroupByKey.");
+          "GroupByKey cannot be applied to non-bounded PCollection in the GlobalWindow without a"
+              + " trigger. Use a Window.into or Window.triggering transform prior to GroupByKey.");
     }
 
     // Validate the window merge function.
@@ -162,6 +164,45 @@ public class GroupByKey<K, V>
       throw new IllegalStateException(
           "GroupByKey must have a valid Window merge function.  " + "Invalid because: " + cause);
     }
+
+    // Validate that the trigger does not finish before garbage collection time
+    if (!triggerIsSafe(windowingStrategy)) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Unsafe trigger may lose data, see"
+                  + " https://s.apache.org/finishing-triggers-drop-data: %s",
+              windowingStrategy.getTrigger()));
+    }
+  }
+
+  // Note that Never trigger finishes *at* GC time so it is OK, and
+  // AfterWatermark.fromEndOfWindow() finishes at end-of-window time so it is
+  // OK if there is no allowed lateness.
+  private static boolean triggerIsSafe(WindowingStrategy<?, ?> windowingStrategy) {
+    if (!windowingStrategy.getTrigger().mayFinish()) {
+      return true;
+    }
+
+    if (windowingStrategy.getTrigger() instanceof NeverTrigger) {
+      return true;
+    }
+
+    if (windowingStrategy.getTrigger() instanceof FromEndOfWindow
+        && windowingStrategy.getAllowedLateness().getMillis() == 0) {
+      return true;
+    }
+
+    if (windowingStrategy.getTrigger() instanceof AfterWatermarkEarlyAndLate
+        && windowingStrategy.getAllowedLateness().getMillis() == 0) {
+      return true;
+    }
+
+    if (windowingStrategy.getTrigger() instanceof AfterWatermarkEarlyAndLate
+        && ((AfterWatermarkEarlyAndLate) windowingStrategy.getTrigger()).getLateTrigger() != null) {
+      return true;
+    }
+
+    return false;
   }
 
   public WindowingStrategy<?, ?> updateWindowingStrategy(WindowingStrategy<?, ?> inputStrategy) {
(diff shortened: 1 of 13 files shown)


sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/GroupByKey.java
```diff
@@ -162,6 +164,44 @@ public static void applicableTo(PCollection<?> input) {
       throw new IllegalStateException(
           "GroupByKey must have a valid Window merge function.  " + "Invalid because: " + cause);
     }
+
+    // Validate that the trigger does not finish before garbage collection time
+    if (!triggerIsSafe(windowingStrategy)) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Unsafe trigger may finish and drop remaining data: %s",
```
je-ik: This would probably rise many questions on mailing list, can we add here a link to a document describing what just happened?

kennknowles: Interesting idea. Yes, I can make a world-readable doc that is a short summary of the issue and why this is disabled and how to contact us for those rare cases where someone wants it back. We could give them a pipeline option. But I would wait until someone asks for it, to avoid implementing something dangerous.

kennknowles:
Doc is https://s.apache.org/finishing-triggers-drop-data. I will edit the error message to link to it.

kennknowles:
Trying to repro the Jenkins failure locally so I can iterate quickly. So far I cannot reproduce it. I suspect the Gradle configuration may have some inadvertent dependency on context.

kennknowles:
run java precommit

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/GroupByKey.java
```diff
@@ -162,6 +164,45 @@ public static void applicableTo(PCollection<?> input) {
...
+                  + " https://s.apache.org/finishing-triggers-drop-data: %s",
+              windowingStrategy.getTrigger()));
+    }
+  }
+
+  // Note that Never trigger finishes *at* GC time so it is OK, and
+  // AfterWatermark.fromEndOfWindow() finishes at end-of-window time so it is
+  // OK if there is no allowed lateness.
+  private static boolean triggerIsSafe(WindowingStrategy<?, ?> windowingStrategy) {
+    if (!windowingStrategy.getTrigger().mayFinish()) {
```
je-ik: I wonder if `Trigger.mayFinish()` actually should not simply return `false` for all triggers - that is if we can simply drop all the trigger finishing logic.
Let's suppose that a trigger might finish. Then:
 a) if it finishes before window GC time, then it is possibly data loss
 b) if it finishes at window GC time, then it is a no-op
 c) if it finishes after window GC time, then it has no semantic meaning

That seems that we want to avoid a) and b) and c) can be ignored. Do we have any other reason to keep triggers the ability to finish?

sdks/java/testing/nexmark/src/main/java/org/apache/beam/sdk/nexmark/queries/Query10.java
```diff
@@ -314,7 +315,7 @@ public void processElement(ProcessContext c, BoundedWindow window)
             name + ".WindowLogFiles",
             Window.<KV<Void, OutputFile>>into(
                     FixedWindows.of(Duration.standardSeconds(configuration.windowSizeSec)))
-                .triggering(AfterWatermark.pastEndOfWindow())
+                .triggering(DefaultTrigger.of())
```
je-ik:
```suggestion
```

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/GroupByKey.java
```diff
@@ -162,6 +164,45 @@ public static void applicableTo(PCollection<?> input) {
       throw new IllegalStateException(
           "GroupByKey must have a valid Window merge function.  " + "Invalid because: " + cause);
     }
+
+    // Validate that the trigger does not finish before garbage collection time
+    if (!triggerIsSafe(windowingStrategy)) {
```
je-ik: Is this logic GBK specific? Perhaps applies to stateful ParDo and Combine as well?

je-ik: Correction - stateful ParDo doesn't use triggers. But Combine might still be affected by this?

kennknowles: Combine is not a primitive operation. If you look at Combine.java you can see that it expands to include a GroupByKey. When a runner sees it in a pipeline, it will usually replace the operation. But the validation happens on the raw graph.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/GroupByKey.java
kennknowles: I agree exactly with all of your analysis. This pull request will forbid triggers of type a).

The meaning of `Trigger.mayFinish()` is to answer this question "is this a trigger that might finish?"

If the answer to that question is "yes" then it might be a type a) trigger. It depends on allowed lateness. There are a couple of very special triggers where the answer "yes" is OK as long as allowed lateness is zero.

sdks/java/testing/nexmark/src/main/java/org/apache/beam/sdk/nexmark/queries/Query10.java
kennknowles: I will run spotlessApply.

kennknowles:
Curiously, `./gradlew -p sdks/java/core test` does not repro the failure locally. I have noticed some inconsistencies between `./gradlew -p <dir> <task>` and `./gradlew <project>:<task>`. I thought these were the same. So we probably have some issues in our gradle configs.

sdks/java/testing/nexmark/src/main/java/org/apache/beam/sdk/nexmark/queries/Query10.java
kennknowles: (did nothing - the formatting is correct)

kennknowles:
Ah, this is of course part of `./gradlew :runners:direct-java:needsRunnerTests`

kennknowles:
Abandoned node enforcement was the culprit, since I was deliberately not running the pipeline. https://scans.gradle.com/s/ie7pt3dazwpbm/tests

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/GroupByKey.java
je-ik: I'm not sure I follow. If you agree with my analysis, then the only conclusion is that triggers simply *should not finish*. If we want to avoid type a), type b) is no-op (no reason to finish then) and type c) is impossible - why there should still exist "special triggers" that might finish?

kennknowles: Yes, I agree that *top level* triggers should not finish. Subtriggers can finish.

 - Consider `Repeatedly.forever(AfterPane.elementCountAtLeast(7))`: The inner trigger finishes, and the outer repeat starts it over.
 - Consider `AfterEach.inOrder(a, b, c)`: Each of `a`, `b`, `c`, finishes to move to the next one.

So it is fine and necessary for a subtrigger to be able to finish.

The question we have to answer is: what do we do when a user tries to set a finishing trigger to be the top level trigger? The two possibilities discussed are:

 - Reject the pipeline, force the user to choose a safe trigger (this PR)
 - Accept the pipeline, but do not actually drop data. Save the data and emit it in the GC pane as we always do. That is #9942 

How do you know whether or not to reject the pipeline? You need a method like `mayFinish`. Any trigger that may finish before GC time must be rejected.

je-ik: Understood. Cool. :+1:

kennknowles:
I definitely think we could remove OnceTrigger. I deliberately left it out of portability. However it would be backwards incompatible if a user has declared a variable of the type. I do think that after this PR we can still proceed with even more than #9942 to clean up. That PR makes it so that a finishing trigger just stops firing until GC time. But we can simply stop tracking whether or not top level triggers are finished, entirely.

kennknowles:
Fixed commit history and I will merge once Java suite passes.

je-ik:
> I definitely think we could remove OnceTrigger. I deliberately left it out of portability. However it would be backwards incompatible if a user has declared a variable of the type. 
Just a one small note: the class is marked `@Internal` - if that annotation should mean something, then removal of such a class would not be considered backwards incompatible. Another question (and related!) is whether it should be part of public APIs (e.g. `Trigger#orFinally`, but many more).

kennknowles:
Jenkins node "too many open files".

kennknowles:
run java precommit

kennknowles:
Ah, yes. Since it is internal it can be eliminated.

PR #9962, opened 2019-11-01
BEAM-7288: Upgraded bcprov-jdk15on to version 1.64 in beam/buildSrc/b…

IamMujuziMoses:
…uild/pluginUnderTestMetadata and in the build.gradle file

https://issues.apache.org/jira/browse/BEAM-7288

Upgraded bcprov-jdk15on to version 1.64 in beam/buildSrc/build/pluginUnderTestMetadata and in the build.gradle files
------------------------

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -31,6 +31,7 @@ plugins {
   id 'net.researchgate.release' version '2.6.0'
   id 'org.apache.beam.module'
   id "org.sonarqube" version "2.7"
+  id "java"
 }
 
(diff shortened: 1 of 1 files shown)


IamMujuziMoses:
retest this please @iemejia

iemejia:
Looks like this dependency is somehow a transitive dependency that is not explicitly used in the project anymore so this looks like already fixed. I am going to resolve it without adding the changes. Thanks @IamMujuziMoses for taking a look.

PR #9975, opened 2019-11-02
Beam 7290

IamMujuziMoses:
https://issues.apache.org/jira/browse/BEAM-7290

BEAM-7290: Updated hadoop-client to version 3.2.0 in the build.gradle and the src/documentation/sdks/java-dependencies.md file.

Signed-off-by: Mujuzi Moses <mujuzimoses@gmail.com>

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -31,6 +31,7 @@ plugins {
   id 'net.researchgate.release' version '2.6.0'
   id 'org.apache.beam.module'
   id "org.sonarqube" version "2.7"
+  id "java"
 }
 
 /*************************************************************************************************/
@@ -262,13 +263,28 @@ task runBeamDependencyCheck() {
   dependsOn ":sdks:python:dependencyUpdates"
 }
 
+repositories {
+	
+	mavenCentral()
+}
+
+dependencies {
+	
+	// https://mvnrepository.com/artifact/org.bouncycastle/bcprov-jdk15on
+compile group: 'org.bouncycastle', name: 'bcprov-jdk15on', version: '1.64'
+}
+
+jar {
+	
+	from configurations.compile.collect {zipTree it}
+}
 // Configure the release plugin to do only local work; the release manager determines what, if
 // anything, to push. On failure, the release manager can reset the branch without pushing.
 release {
   revertOnFail = false
   tagTemplate = 'v${version}'
   git {
-    requireBranch = 'release-.*|master'
+   requireBranch = 'release-.*|master'
     pushToRemote = ''
   }
 }
--- a/build.gradle
+++ b/build.gradle
@@ -267,11 +267,17 @@ repositories {
 	
 	mavenCentral()
 }
-
+    // to use:
+    // dependencies {
+    //   compile library.java.slf4j_api
 dependencies {
 	
 	// https://mvnrepository.com/artifact/org.bouncycastle/bcprov-jdk15on
 compile group: 'org.bouncycastle', name: 'bcprov-jdk15on', version: '1.64'
+
+// https://mvnrepository.com/artifact/org.apache.hadoop/hadoop-client
+compile group: 'org.apache.hadoop', name: 'hadoop-client', version: '3.2.0'
+
 }
 
 jar {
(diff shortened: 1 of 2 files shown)


IamMujuziMoses:
retest this please @iemejia

website/src/documentation/sdks/java-dependencies.md
```diff
@@ -91,7 +91,7 @@ the relevant section below.</p>
   <tr><td>io.grpc</td><td>grpc-stub</td><td>1.13.1</td></tr>
   <tr><td>com.google.guava</td><td>guava</td><td>20.0</td></tr>
   <tr><td>com.google.guava</td><td>guava-testlib</td><td>20.0</td></tr>
-  <tr><td>org.apache.hadoop</td><td>hadoop-client</td><td>2.7.3</td></tr>
```
lukecwik: This file lists dependency versions at specific SDK versions and isn't meant to be edited when versions change at head.

build.gradle
```diff
@@ -262,13 +263,34 @@ task runBeamDependencyCheck() {
...
+    // to use:
+    // dependencies {
+    //   compile library.java.slf4j_api
+dependencies {
+	
+	// https://mvnrepository.com/artifact/org.bouncycastle/bcprov-jdk15on
+compile group: 'org.bouncycastle', name: 'bcprov-jdk15on', version: '1.64'
+
+// https://mvnrepository.com/artifact/org.apache.hadoop/hadoop-client
+compile group: 'org.apache.hadoop', name: 'hadoop-client', version: '3.2.0'
```
lukecwik: If you want to upgrade hadoop, you want to edit [BeamModulePlugin.groovy](https://github.com/apache/beam/blob/980e655a32a65f464ace85ff518b5d5b346677ba/buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy#L376) since that is where we define shared dependency versions.

lukecwik:
@iemejia How will upgrading to 3.x hadoop from 2.x impact Beam users, do we expect compilation/feature issues or dependency issues. What about runners who depend on Hadoop, will they be impacted?

iemejia:
> @iemejia How will upgrading to 3.x hadoop from 2.x impact Beam users, do we expect compilation/feature issues or dependency issues. What about runners who depend on Hadoop, will they be impacted?

We expect both, I took a quick look and seems to be 'easier' than I expected for the IO modules at least given that dependencies are mostly provided, probably the part where we have to do more changes is on Spark runner once Spark 3 is released. Filled https://issues.apache.org/jira/browse/BEAM-8569 to discuss more and create sub tasks.

iemejia:
And for the sake of this PR we could probably move to a more recent version of the 2.x releases (this may solve some CVEs) but I am still not sure of unintended consequences with versions leaked in the runtime classpath (e.g. with current Spark runner).
A surprising discovery related to this is that 2.7.x that is the one used by mostly every project + many vendors is already EOL.
https://cwiki.apache.org/confluence/display/HADOOP/EOL+%28End-of-life%29+Release+Branches

iemejia:
This seems to be a more complex process than expected as explained in [BEAM-8569](https://issues.apache.org/jira/browse/BEAM-8569). We cannot just update the version, we need to be catious and add first a 'profile' in gradle to dynamically change the provided dependencies in our tests and later in the future (when the Big data ecosystem catches up) make the Hadoop 3 version the default one.

Do you have any idea of how to do that in gradle @lukecwik, this is an issue I had already mentioned in the past in the context of IOs [BEAM-4087](https://issues.apache.org/jira/browse/BEAM-4087). Note that Flink somehow resolves this, but in a more convoluted way so I am looking if there is a better practice/solution.

lukecwik:
It depends, the simplest approach is to mark the dependency as provided and force the user to provide either hadoop 2 or hadoop 3 deps but you can only do this if the hadoop 2 and 3 versions are binary compatible. If they are binary compatible and you don't want to go down the provided route you can create a cross-product of hadoop.version x project.that.depends.on.hadoop.

If they are not binary compatible then you need to create a project that is specific to that version similar to the AWS and AWS2 library.

PR #9979, opened 2019-11-04
[BEAM-7951] Allow runner to configure customization WindowedValue coder.

sunjincheng121:
Motivation:
The coder of WindowedValue cannot be configured and it’s always FullWindowedValueCoder. We don't need to serialize the timestamp, window and pane properties in Flink and so it will be better to make the coder configurable (i.e. allowing to use ValueOnlyWindowedValueCoder)

Detail changes:
- Add ValueOnlyWindowedValueCoder into MODEL_CODER so that we can use value only coder between runner and harness.
- To make it configurable and also maybe for other configurations in futher(i.e., more general), we add a `configurations` map in Components.
- Add python ValueOnlyWindowedValueCoder.

Some concerns reviewers may have:
- Do we have any other options instead of adding the configurations in Components(beam_runner_api.proto).
  Yes, we have other options. For example, we can add some flags in the methods when call `addSdkWireCoder` and `addRunnerWireCoder`. However, the call stack is very deep(from DefaultJobBundleFactory.forStage() to WireCoders.addSdkWireCoder()). It would be a mess if we add a flag for all these methods.
- The current changes are not used in Beam project.
  Due to the doFn can emit timestamp in beam, it seems we can't tell when we can ignore encode or decode the timestamp field, i.e., currently, we only use FullWindowedValueCoder in Beam. However, in Flink we don't have such problem and can make use of the ValueOnlyWindowedValueCoder.
- Have not add support for other SDKs(go)?
  Maybe we can add support for other SDKs later when we find it is useful?

--- | --- | --- | --- | ---

### DIFF
--- a/model/pipeline/src/main/proto/beam_runner_api.proto
+++ b/model/pipeline/src/main/proto/beam_runner_api.proto
@@ -674,6 +674,16 @@ message StandardCoders {
     // Components: The element coder and the window coder, in that order
     WINDOWED_VALUE = 8 [(beam_urn) = "beam:coder:windowed_value:v1"];
 
+    // A windowed value coder with parameterized timestamp, windows and pane info.
+    // Encodes an element with only the value of the windowed value.
+    // Decodes the value and assign the parameterized timestamp, windows and PaneInfo to the
+    // windowed value
+    // Components: The element coder and the window coder, in that order
+    // The payload of this coder is an encoded windowed value using the
+    // beam:coder:windowed_value:v1 coder parameterized by beam:coder:bytes:v1
+    // elements coder and the window coder that this param_windowed_value uses.
+    PARAM_WINDOWED_VALUE = 14 [(beam_urn) = "beam:coder:param_windowed_value:v1"];
+
     // Encodes an iterable of elements, some of which may be stored elsewhere.
     //
     // The encoding for a state-backed iterable is the same as that for
@@ -1063,6 +1073,21 @@ message SideInput {
   FunctionSpec window_mapping_fn = 3;
 }
 
+// Settings that decide the coder type of wire coder.
+message WireCoderSetting {
+  // (Required) The URN of the wire coder.
+  // Note that only windowed value coder or parameterized windowed value coder are supported.
+  string urn = 1;
+
+  // (Optional) The data specifying any parameters to the URN. If
+  // the URN is beam:coder:windowed_value:v1, this may be omitted. If the URN is
+  // beam:coder:param_windowed_value:v1, the payload is an encoded windowed
+  // value using the beam:coder:windowed_value:v1 coder parameterized by
+  // beam:coder:bytes:v1 elements coder and the window coder that this
+  // param_windowed_value uses.
+  bytes payload = 2;
+}
+
 // An environment for executing UDFs. By default, an SDK container URL, but
 // can also be a process forked by a command, or an externally managed process.
 message Environment {
@@ -1270,6 +1295,9 @@ message ExecutableStagePayload {
   // We use an environment rather than environment id
   // because ExecutableStages use environments directly. This may change in the future.
   Environment environment = 1;
+
+  // set the wire coder of this executable stage
+  WireCoderSetting wire_coder_setting = 9;
 
   // (Required) Input PCollection id. This must be present as a value in the inputs of any
   // PTransform the ExecutableStagePayload is the payload of.
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/util/WindowedValue.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/util/WindowedValue.java
@@ -20,8 +20,12 @@ package org.apache.beam.sdk.util;
 import static org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Preconditions.checkArgument;
 import static org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Preconditions.checkNotNull;
 
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
 import java.io.OutputStream;
 import java.util.Arrays;
 import java.util.Collection;
@@ -30,6 +34,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
+import org.apache.beam.sdk.coders.ByteArrayCoder;
 import org.apache.beam.sdk.coders.Coder;
 import org.apache.beam.sdk.coders.CoderException;
 import org.apache.beam.sdk.coders.CollectionCoder;
@@ -505,6 +510,11 @@ public abstract class WindowedValue<T> {
     return ValueOnlyWindowedValueCoder.of(valueCoder);
   }
 
+  /** Returns the {@code ParamWindowedValueCoder} from the given valueCoder. */
+  public static <T> ParamWindowedValueCoder<T> getParamWindowedValueCoder(Coder<T> valueCoder) {
+    return ParamWindowedValueCoder.of(valueCoder);
+  }
+
   /** Abstract class for {@code WindowedValue} coder. */
   public abstract static class WindowedValueCoder<T> extends StructuredCoder<WindowedValue<T>> {
     final Coder<T> valueCoder;
@@ -637,7 +647,11 @@ public abstract class WindowedValue<T> {
    *
    * <p>A {@code ValueOnlyWindowedValueCoder} only encodes and decodes the value. It drops timestamp
    * and windows for encoding, and uses defaults timestamp, and windows for decoding.
+   *
+   * @deprecated Use ParamWindowedValueCoder instead it is general purpose implementation of the
+   *     same concept but makes timestamp, windows and pane info configurable.
    */
+  @Deprecated
   public static class ValueOnlyWindowedValueCoder<T> extends WindowedValueCoder<T> {
     public static <T> ValueOnlyWindowedValueCoder<T> of(Coder<T> valueCoder) {
       return new ValueOnlyWindowedValueCoder<>(valueCoder);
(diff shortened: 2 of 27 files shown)


sunjincheng121:
Run Python PreCommit

sunjincheng121:
Hi, @lukecwik 

Thanks a lot for your advice. I think a discussion in mailing list is a good idea. I will fire a discussion later.

As for the second option listed by you, the PR is quite similar to it. 
With the PR, the (Python)SDK realize it was given a value only coder from the coder_id in `RemoteGrpcPort`. And it seems we don't need to change `BeamFnDataReadRunner` or `BeamFnDataWriteRunner` due to `ValueOnlyWindowedValueCoder` is also a coder of WindowedValue. 

What do you think?

Best,
Jincheng

sunjincheng121:
The discussion mail thread: https://lists.apache.org/thread.html/a93e3985cd729b3cb5a8bb9fbff59851d075d5894b411e2ffaedf999@%3Cdev.beam.apache.org%3E

mxm:
Please let me know once you have updated the PR according to the discussion result.

sunjincheng121:
Hi @mxm, Thanks a lot for your attention! I have updated the PR. The changes mainly include:

1. Add a ParameterizedWindowedValueCoder that with constant timestamp, windows and paneInfo. The ParameterizedWindowedValueCoder at the harness side init the constants from the payload of the coder, and to achieve this, the `message WindowInfo` is added in `beam_runner_api.proto` to carry the constant information.
   
2. Add WireCoderSetting in ExecutableStagePayload. With the WireCoderSetting, we can configure wire coder to `ParameterizedWindowedValueCoder` to avoid redundant encoding and decoding.

Would be great if you can take a look! 
Best,
Jincheng

sunjincheng121:
Run Portable_Python PreCommit

sunjincheng121:
Run Java PreCommit

sunjincheng121:
Run Python PreCommit

sunjincheng121:
Run Python PreCommit

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -1338,3 +1359,44 @@ message ExecutableStagePayload {
...
+  int64 timestamp = 1;
+  repeated BoundedWindow bounded_windows = 2;
+  PaneInfo pane_info = 3;
+}
+
+// Represents window information assigned to data elements
+message BoundedWindow {
+  oneof window {
+    GlobalWindow global_window = 1;
+    IntervalWindow interval_window = 2;
```
mxm: What about session windows?

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/windowing/BoundedWindow.java
```diff
@@ -35,7 +36,7 @@
  * windows that are logically equal will be treated as equal by {@code equals()} and {@code
  * hashCode()}.
  */
-public abstract class BoundedWindow {
+public abstract class BoundedWindow implements Serializable {
```
mxm: Not sure about this one. Why can't we use `WindowCoder` instead?

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/windowing/PaneInfo.java
```diff
@@ -40,7 +41,7 @@
  *
  * <p>Note: This does not uniquely identify a pane, and should not be used for comparisons.
  */
-public final class PaneInfo {
+public final class PaneInfo implements Serializable {
```
mxm: Not sure about this one. Why can't we use `PaneInfoCoder` instead?

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -1338,3 +1359,44 @@ message ExecutableStagePayload {
...
+// Window information in WindowedValue
+message WindowInfo {
+  // timestamp in milliseconds
+  int64 timestamp = 1;
+  repeated BoundedWindow bounded_windows = 2;
+  PaneInfo pane_info = 3;
+}
+
+// Represents window information assigned to data elements
+message BoundedWindow {
```
lukecwik: This limits the set of window types and the number of windows that can be specified. I think we should go with a model where we encode a "dummy element" which is replaced by the actual element when decoding. For example, the payload would be:
```
message ParameterizedWindowedValuePayload {
  // A specification of a windowed value coder where the element coder is always "beam:coder:bytes:v1".
  string coder_id;
  // Contains an encoded windowed value with an empty byte[] element. This elements windowing
  // information should be copied to all elements that are decoded with this coder.
  bytes value;
}
```

This allows us to use the full range of window encodings and the already defined pane/timestamp encodings.

sunjincheng121:
Hi, @mxm @lukecwik Thanks a lot for your review and suggestions. I think it's a good idea to use a `WindowedValue` coder to encode the constant values, and decode the `WindowedValue` at python side. In this way, we don't need to add protobuf class to define window info. I will try to update the PR according to your suggestions. 

As for the coder name, how about call it `constant_windowed_value`?  I haven't found a better name yet. 

Best, 
Jincheng

sunjincheng121:
Sorry for my belated reply as I have just returned from Flink Forward Asia 2019. (Talk about the PyFlink on Beam Portability Framework )

lukecwik:
How about we stick with ValueOnlyWindowedValueCoder (value_only_windowed_value_coder for the URN), and repurpose the Java one to do this as well.

sunjincheng121:
Hi @lukecwik, My first thoughts is that the `ValueOnly` seems give some hint that `timestamp/windows/pane` info are ignored.
So, how about we rename it to `ParamWindowedValueCoder`(param_windowed_value_coder for the URN)?

sunjincheng121:
Hi @lukecwik @mxm 

I have updated the PR. The changes as follows:
The previous Proto definition of windowing information has been removed.
Instead, I use the `Windowcoder` to encode the constant window information and store it in the payload of the `ParamWindowedValueCoder`.

As for the coder name, I name it `ParamWindowedValueCoder` in the PR. We can further discuss about it if you still have other concerns.

The last thing needs to be mentioned is it seems we have to make `BoundedWindow` and `PaneInfo` implement Serializable, because the coder(`ParamWindowedValueCoder`) needs to be serializable, So, I think the members should also be serializable.

Best,
Jincheng

sunjincheng121:
Run Python PreCommit

sdks/python/apache_beam/coders/coders.py
```diff
@@ -1136,6 +1136,62 @@ def __hash__(self):
...
+    return coder_impl.ParamWindowedValueCoderImpl(
+        self._wrapped_value_coder.get_impl(),
+        self.window_coder.get_impl(),
+        self.payload)
+
+  def is_deterministic(self):
+    return all(c.is_deterministic() for c in [self._wrapped_value_coder,
+                                              self.window_coder])
+
+  def as_cloud_object(self, coders_context=None):
```
lukecwik: This is meant to support Dataflow explicitly and is similar to the proto conversion.

You'll need to add one to java in [CloudObjectTranslators](https://github.com/apache/beam/blob/a06b7a6cbad2386b7bee37d3be0d7f3f5b8bfaff/runners/google-cloud-dataflow-java/src/main/java/org/apache/beam/runners/dataflow/util/CloudObjectTranslators.java#L244) and wire it together.

You'll need to add the encoded payload as part of the cloud object as well and I suggest you add it as a 'payload' field. Please use the same byte string -> json compatible string conversion found elsewhere in these as_cloud_object conversions.

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -1059,6 +1066,17 @@ message SideInput {
...
 
+// Settings that decide the coder type of wire coder.
+message WireCoderSetting {
+  // (Required) The URN of the wire coder.
+  // Note that only windowed value coder or parameterized windowed value coder are supported.
+  string urn = 1;
+
+  // (Optional) The data specifying any parameters to the URN. If
+  // the URN does not require any arguments, this may be omitted.
+  bytes payload = 2;
```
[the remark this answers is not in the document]
sunjincheng121: We can get the element coder id and window coder id like the same way of the current windowed value coder. The element coder and window coder can be used to encode and decode the payload. For example, in WireCoders, the elementCoderId and 
windowCoderId are get from PCollectionNode.

sdks/java/core/src/main/java/org/apache/beam/sdk/util/WindowedValue.java
```diff
@@ -693,4 +697,110 @@ public void registerByteSizeObserver(WindowedValue<T> value, ElementByteSizeObse
...
     }
   }
+
+  /**
+   * A parameterized Coder for {@code WindowedValue}.
+   *
+   * <p>A {@code ParamWindowedValueCoder} only encodes and decodes the value. It drops timestamp and
+   * windows for encoding, and uses defaults(constant) timestamp, and windows for decoding.
+   */
+  public static class ParamWindowedValueCoder<T> extends FullWindowedValueCoder<T> {
```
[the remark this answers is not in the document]
sunjincheng121: I was thinking `ParamWindowedValueCoder` is a special kind of `FullWindowedValueCoder` with the parameterized timestamp, windows and pane info for the WindowedValues, i.e., `ParamWindowedValueCoder` is an optimization of `FullWindowedValueCoder`. We can also benefit from it such as reuse the code of testing FullWindowedValueCoder in CommonCoderTest, etc.

sdks/python/apache_beam/coders/coders.py
sunjincheng121: Thanks a lot for the detailed suggestions. How about implementing it in a following PR? We can throw an exception in this PR for now. Involve too much in a single PR seems not a good idea. What do you think?

sunjincheng121:
@lukecwik @mxm Appreciate your review and nice comments. I have updated the PR according to your suggestions. 

I have replied to some comments above. The other comments without reply are all make mor sense to me. Thank you very much!

Best,
Jincheng

sunjincheng121:
Run Python PreCommit

sunjincheng121:
Run Python PreCommit

sunjincheng121:
It seems that the CI fail not due to this PR.

sdks/java/core/src/main/java/org/apache/beam/sdk/util/WindowedValue.java
```diff
@@ -500,11 +505,22 @@ private void ensureWindowsAreASet() {
...
 
-  /** Returns the {@code ValueOnlyCoder} from the given valueCoder. */
+  /**
+   * Returns the {@code ValueOnlyCoder} from the given valueCoder.
+   *
+   * @deprecated Use {@link #getParamWindowedValueCoder(Coder)} instead. The {@link
+   *     ValueOnlyWindowedValueCoder} is deprecated now.
+   */
+  @Deprecated
   public static <T> ValueOnlyWindowedValueCoder<T> getValueOnlyCoder(Coder<T> valueCoder) {
```
mxm: I'm not sure we need to deprecate this method and replace all of the invocations. This is still a useful method if no parameters are supplied. We can simply redirect to the new ParamWindowedValueCoder here.

sdks/python/apache_beam/coders/coders.py
mxm: Seems fair to fix this at a later point in time, since this is not used in Dataflow at the moment.

sdks/java/core/src/main/java/org/apache/beam/sdk/util/WindowedValue.java
sunjincheng121: Hi, return `ParamWindowedValueCoder` would result some compatibility problems for Dataflow, but I think we can keep the `getValueOnlyCoder` as you suggested. We can deprecate it later once we find need to.:)

sunjincheng121:
Hi thanks for the review @mxm ! I have update the PR accordingly.

sunjincheng121:

mxm:
LGTM. I'll give @lukecwik another day to have another look, since he appears to be back.

sunjincheng121:
Run Python PreCommit

sunjincheng121:
Run Python PreCommit

sunjincheng121:
Run Python PreCommit

sunjincheng121:
Run Python PreCommit

mxm:
Seeing an error in the Python tests, but it looks unrelated:

```
09:41:25 ___________________ FnApiLogRecordHandlerTest.test_exc_info ____________________
09:41:25 [gw0] linux2 -- Python 2.7.12 /home/jenkins/jenkins-slave/workspace/beam_PreCommit_Python_Phrase/src/sdks/python/test-suites/tox/py2/build/srcs/sdks/python/target/.tox-py27-cython-pytest/py27-cython-pytest/bin/python
09:41:25 
09:41:25 self = <apache_beam.runners.worker.log_handler_test.FnApiLogRecordHandlerTest testMethod=test_exc_info>
09:41:25 
09:41:25     def test_exc_info(self):
09:41:25       try:
09:41:25         raise ValueError('some message')
09:41:25       except ValueError:
09:41:25         _LOGGER.error('some error', exc_info=True)
09:41:25     
09:41:25       self.fn_log_handler.close()
09:41:25     
09:41:25 >     log_entry = self.test_logging_service.log_records_received[0].log_entries[0]
09:41:25 E     IndexError: list index out of range
09:41:25 
09:41:25 apache_beam/runners/worker/log_handler_test.py:110: IndexError
```

sunjincheng121:
Run Python PreCommit

sunjincheng121:
Yes, I have retried a few times but without success. I just find it is reported recently by [BEAM-8974](https://issues.apache.org/jira/browse/BEAM-8974) and it seems the PR will be merged soon. 

I have retried again. Hope we can make it this time by chance.

sunjincheng121:
Rebase code and squash the commits. :)

lukecwik:
Yeah, I'm back and wanted to go through this so please don't merge as I was looking for a large enough block of time to cover this change again.

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -674,6 +674,16 @@ message StandardCoders {
     // Components: The element coder and the window coder, in that order
     WINDOWED_VALUE = 8 [(beam_urn) = "beam:coder:windowed_value:v1"];
 
+    // A windowed value coder with parameterized timestamp, windows and pane info.
+    // Encodes an element with only the value of the windowed value.
+    // Decodes the value and assign the parameterized timestamp, windows and PaneInfo to the
```
lukecwik:
```suggestion
    // Decodes the value and assigns the parameterized timestamp, windows and pane info to the
```

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -674,6 +674,16 @@ message StandardCoders {
...
+    // windowed value
```
lukecwik:
```suggestion
    // windowed value.
```

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -674,6 +674,16 @@ message StandardCoders {
...
+    // Components: The element coder and the window coder, in that order
+    // The payload of this coder is an encoded windowed value using the
+    // beam:coder:windowed_value:v1 coder parameterized by beam:coder:bytes:v1
```
lukecwik:
```suggestion
    // beam:coder:windowed_value:v1 coder parameterized by a beam:coder:bytes:v1
```

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -674,6 +674,16 @@ message StandardCoders {
...
+    // elements coder and the window coder that this param_windowed_value uses.
```
lukecwik:
```suggestion
    // element coder and the window coder that this param_windowed_value coder uses.
```

sdks/java/core/src/main/java/org/apache/beam/sdk/util/WindowedValue.java
```diff
@@ -637,7 +647,11 @@ public void registerByteSizeObserver(WindowedValue<T> value, ElementByteSizeObse
    *
    * <p>A {@code ValueOnlyWindowedValueCoder} only encodes and decodes the value. It drops timestamp
    * and windows for encoding, and uses defaults timestamp, and windows for decoding.
+   *
+   * @deprecated Use ParamWindowedValueCoder instead it is general purpose implementation of the
```
lukecwik:
```suggestion
   * @deprecated Use ParamWindowedValueCoder instead, it is a general purpose implementation of the
```

sdks/java/core/src/main/java/org/apache/beam/sdk/util/WindowedValue.java
```diff
@@ -693,4 +707,181 @@ public void registerByteSizeObserver(WindowedValue<T> value, ElementByteSizeObse
       return Collections.singletonList(valueCoder);
     }
   }
+
+  /**
+   * A parameterized Coder for {@code WindowedValue}.
```
lukecwik:
```suggestion
   * A parameterized coder for {@code WindowedValue}.
```

sdks/java/core/src/main/java/org/apache/beam/sdk/util/WindowedValue.java
```diff
@@ -693,4 +707,181 @@ public void registerByteSizeObserver(WindowedValue<T> value, ElementByteSizeObse
...
+        Coder<T> valueCoder,
+        Coder<? extends BoundedWindow> windowCoder,
+        Instant timestamp,
+        Collection<? extends BoundedWindow> windows,
+        PaneInfo pane) {
+      return new ParamWindowedValueCoder<>(valueCoder, windowCoder, timestamp, windows, pane);
+    }
+
+    /**
+     * Returns the {@link ParamWindowedValueCoder} from the given valueCoder, windowCoder and the
```
lukecwik:
```suggestion
     * Returns the {@link ParamWindowedValueCoder} for the given valueCoder and windowCoder using
```

sdks/java/core/src/main/java/org/apache/beam/sdk/util/WindowedValue.java
```diff
@@ -693,4 +707,181 @@ public void registerByteSizeObserver(WindowedValue<T> value, ElementByteSizeObse
...
+     * supplied parameterized {@link BoundedWindow#TIMESTAMP_MIN_VALUE}, {@link #GLOBAL_WINDOWS} and
```
lukecwik:
```suggestion
     * {@link BoundedWindow#TIMESTAMP_MIN_VALUE} as the timestamp, {@link #GLOBAL_WINDOWS} as the window, and
```

sdks/java/core/src/main/java/org/apache/beam/sdk/util/WindowedValue.java
```diff
@@ -693,4 +707,181 @@ public void registerByteSizeObserver(WindowedValue<T> value, ElementByteSizeObse
...
+     * {@link PaneInfo#NO_FIRING} for {@link WindowedValue}.
```
lukecwik:
```suggestion
     * {@link PaneInfo#NO_FIRING} as the pane info for parameters.
```

lukecwik:
I went through all the usages of FullWindowedValueCoder and didn't see any issues with any explicit downcasts being done where ParamWindowedvalueCoder wouldn't work in its place.

sunjincheng121:
Many thanks to the review and suggestions. @lukecwik @mxm 👍 

Will open the follow-up PRs soon.

PR #9991, opened 2019-11-05
[BEAM-8557]Add log for the dropped unknown response

sunjincheng121:
In this PR will remove the null check for  `ResponseStreamObserver` and `InboundObserver`
Because before the the `onNext` call, the `Future` already put into the queue in `handle` method.

I found the test as follows:
```
 @Test
 public void testUnknownResponseIgnored() throws Exception
```
I do not know why we need test this case? I think it would be better if we throw the Exception for an UnknownResponse, otherwise, this may hidden a potential bug. 

Please correct me if there anything I misunderstand @kennknowles

--- | --- | --- | --- | ---

### DIFF
--- a/runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/FnApiControlClient.java
+++ b/runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/FnApiControlClient.java
@@ -148,16 +148,18 @@ public class FnApiControlClient implements Closeable, InstructionRequestHandler 
       LOG.debug("Received InstructionResponse {}", response);
       CompletableFuture<BeamFnApi.InstructionResponse> responseFuture =
           outstandingRequests.remove(response.getInstructionId());
-      if (responseFuture != null) {
-        if (response.getError().isEmpty()) {
-          responseFuture.complete(response);
-        } else {
-          responseFuture.completeExceptionally(
-              new RuntimeException(
-                  String.format(
-                      "Error received from SDK harness for instruction %s: %s",
-                      response.getInstructionId(), response.getError())));
-        }
+      if (responseFuture == null) {
+        LOG.warn("Dropped unknown InstructionResponse {}", response);
+        return;
+      }
+      if (response.getError().isEmpty()) {
+        responseFuture.complete(response);
+      } else {
+        responseFuture.completeExceptionally(
+            new RuntimeException(
+                String.format(
+                    "Error received from SDK harness for instruction %s: %s",
+                    response.getInstructionId(), response.getError())));
       }
     }
 
--- a/sdks/java/harness/src/main/java/org/apache/beam/fn/harness/state/BeamFnStateGrpcClientCache.java
+++ b/sdks/java/harness/src/main/java/org/apache/beam/fn/harness/state/BeamFnStateGrpcClientCache.java
@@ -138,12 +138,14 @@ public class BeamFnStateGrpcClientCache {
       public void onNext(StateResponse value) {
         LOG.debug("Received StateResponse {}", value);
         CompletableFuture<StateResponse> responseFuture = outstandingRequests.remove(value.getId());
-        if (responseFuture != null) {
-          if (value.getError().isEmpty()) {
-            responseFuture.complete(value);
-          } else {
-            responseFuture.completeExceptionally(new IllegalStateException(value.getError()));
-          }
+        if (responseFuture == null) {
+          LOG.warn("Dropped unknown StateResponse {}", value);
+          return;
+        }
+        if (value.getError().isEmpty()) {
+          responseFuture.complete(value);
+        } else {
+          responseFuture.completeExceptionally(new IllegalStateException(value.getError()));
         }
       }
 


runners/java-fn-execution/src/test/java/org/apache/beam/runners/fnexecution/control/FnApiControlClientTest.java
```diff
@@ -100,7 +99,8 @@ public void testRequestError() throws Exception {
   }
 
   @Test
-  public void testUnknownResponseIgnored() throws Exception {
+  public void testUnknownResponse() throws Exception {
+    thrown.expect(NullPointerException.class);
```
kennknowles: My opinion: an NPE always indicates a programming error. Null should not exist anymore in high-level programming. Or in most medium-level or low-level programming. The code should prevent an NPE and give a good error message.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/FnApiControlClient.java
```diff
@@ -148,16 +148,14 @@ public void onNext(BeamFnApi.InstructionResponse response) {
...
-        if (response.getError().isEmpty()) {
-          responseFuture.complete(response);
-        } else {
-          responseFuture.completeExceptionally(
-              new RuntimeException(
-                  String.format(
-                      "Error received from SDK harness for instruction %s: %s",
-                      response.getInstructionId(), response.getError())));
-        }
+      if (response.getError().isEmpty()) {
```
kennknowles: I do not like this change, because of my comment below about NPE.

But also, I do not know the code well enough. Is there a distributed systems question? Can a duplicate message be received that should be recieved?

If the answer is "yes" then the previous code is correct. If the answer is "no" then the previous code should crash or at least log an error.

runners/java-fn-execution/src/test/java/org/apache/beam/runners/fnexecution/control/FnApiControlClientTest.java
sunjincheng121: I agree that meaningful message is better than NPE.

sunjincheng121: I agree that meaningful message is better than NPE.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/FnApiControlClient.java
sunjincheng121: Thanks for the quick reply!

Is there some logic for the resent the message?  I have not find it, so  I think in this case there is no duplicate message. 

Please correct me if there anything I misunderstand :)

sunjincheng121:
Run Portable_Python PreCommit

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/FnApiControlClient.java
kennknowles: I read some code and thought about this. The message comes from the other side of a gRPC connection. We cannot assume what code is connected. It could be any of the SDK harnesses, or could be another SDK harness not part of Beam.

Receiving a response for a request where we are not waiting for an answer is OK. Maybe it means there is a bug in code on the other side of the connection. I don't think the runner side should crash and terminate. It may be worth a log message, but then a problem on the SDK harness could cause flooding on the runner.

sunjincheng121: Totally agree that this may mean a bug in the SDK harness if it happens. I'm just thinking that a log message may be not enough as it's easy to be ignored. An exception makes sure that problem was discovered in time. But I'm fine if you think log is enough as users will investigate the log if there are serious problem for them after all. So, we need only add a WARNING log for this case, right?

sunjincheng121:
R: @kennknowles  I have updated the PR, is that make sense to you :) Welcome any feedback.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/FnApiControlClient.java
```diff
@@ -158,6 +158,8 @@ public void onNext(BeamFnApi.InstructionResponse response) {
                       "Error received from SDK harness for instruction %s: %s",
                       response.getInstructionId(), response.getError())));
         }
+      } else {
+        LOG.warn("Dropped unknown InstructionResponse {}", response);
```
kennknowles: Tiny style suggestion: move the error case to the top. Reduce one level of curly braces.

```java
if (responseFuture == null) {
  LOG.warn(...)
  return;
}
```

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/FnApiControlClient.java
lukecwik: It is an error if we receive a response for something we weren't expecting.

We should close the control/data/state channel. The runner is then responsible for restarting the SDK harness and resuming processing.

sdks/java/harness/src/main/java/org/apache/beam/fn/harness/state/BeamFnStateGrpcClientCache.java
```diff
@@ -138,12 +138,14 @@ private synchronized void closeAndCleanUp(RuntimeException cause) {
       public void onNext(StateResponse value) {
         LOG.debug("Received StateResponse {}", value);
         CompletableFuture<StateResponse> responseFuture = outstandingRequests.remove(value.getId());
-        if (responseFuture != null) {
-          if (value.getError().isEmpty()) {
-            responseFuture.complete(value);
-          } else {
-            responseFuture.completeExceptionally(new IllegalStateException(value.getError()));
-          }
+        if (responseFuture == null) {
```
lukecwik: this should be a terminal error on the SDK side where we should close the control/data/state channel and have the runner restart us

PR #10027, opened 2019-11-07
[BEAM-8577] Initialize FileSystems during Coder deserialization in Re…

dmvk:
https://issues.apache.org/jira/browse/BEAM-8577

### DIFF
--- a/runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkBatchTransformTranslators.java
+++ b/runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkBatchTransformTranslators.java
@@ -34,6 +34,7 @@ import org.apache.beam.runners.core.construction.ParDoTranslation;
 import org.apache.beam.runners.core.construction.ReadTranslation;
 import org.apache.beam.runners.flink.translation.functions.FlinkAssignWindows;
 import org.apache.beam.runners.flink.translation.functions.FlinkDoFnFunction;
+import org.apache.beam.runners.flink.translation.functions.FlinkIdentityFunction;
 import org.apache.beam.runners.flink.translation.functions.FlinkMergingNonShuffleReduceFunction;
 import org.apache.beam.runners.flink.translation.functions.FlinkMultiOutputPruningFunction;
 import org.apache.beam.runners.flink.translation.functions.FlinkPartialReduceFunction;
@@ -84,6 +85,7 @@ import org.apache.flink.api.java.operators.FlatMapOperator;
 import org.apache.flink.api.java.operators.GroupCombineOperator;
 import org.apache.flink.api.java.operators.GroupReduceOperator;
 import org.apache.flink.api.java.operators.Grouping;
+import org.apache.flink.api.java.operators.MapOperator;
 import org.apache.flink.api.java.operators.MapPartitionOperator;
 import org.apache.flink.api.java.operators.SingleInputUdfOperator;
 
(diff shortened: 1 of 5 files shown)


runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkBatchTransformTranslators.java
```diff
@@ -306,11 +308,19 @@ public void translateNode(
...
-      context.setOutputDataSet(context.getOutput(transform), inputDataSet.rebalance());
+      @SuppressWarnings("unchecked")
+      final CoderTypeInformation<WindowedValue<KV<K, InputT>>> outputType =
+          ((CoderTypeInformation) inputDataSet.getType())
+              .withPipelineOptions(context.getPipelineOptions());
+      final DataSet<WindowedValue<KV<K, InputT>>> retypedDataSet =
+          new MapOperator<>(
+              inputDataSet,
+              outputType,
+              FlinkIdentityFunction.of(),
```
mxm: It looks like we could stick the existing patterns to load the options as part of the operator's open method.

dmvk: This wouldn't help, the MapOperator is called on "map side" of the rebalance. we encounter problems on "reduce side". Only reason for this `MapOperator` is, that we are unable to change TypeInformation of the input dataset.

However the main reason is, that DoFns, that follow rebalance are currently not chainable, therefore there is a slight chance of this happening.

I reckon that after resolving https://issues.apache.org/jira/browse/BEAM-8608, this would be no longer needed, because we'd init fs in chained DoFn's open method.

mxm: I see. Would be good then to generally move to this approach. Let's add a comment here why this is necessary.

cchepelov:
I've been hit by the same problem in a 

- scio 0.8.0-beta2   (cc @nevillelyh @jto )
- scala 2.12 (.10) 
- beam 2.16.0
- flink 1.8.3 (or 1.9.1) 

job while trying to use the flink runner on a cluster. Same issue as described in BEAM-8577, the root cause was the reshuffle and uninitialized (BEAM) FileSystems registry across the ResourceIdCoder.

Backporting @dmvk's patch to beam 2.16.0 (including persuading gradle into making a local Ivy publication to get picked up by SBT) successfully led to reliably hit another, unrelated problem. Would love to see this PR go through! Thank you

mxm:
This is fine as-is, should have made clearer that I'm just missing a comment in the translator. I'd also favor a unified style to initialize pipeline options, instead of the two different ways we have now. We can tackle this in a follow-up. Merging.

mxm:
Run Java PreCommit

mxm:
Run Spotless PreCommit

dmvk:
@mxm let me rebase this first

mxm:
I thought rebasing wouldn't be necessary, but checkstyle fails. If you rebase, please add a comment in the Reshuffle translator to explain why we're adding the noop function.

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/functions/FlinkIdentityFunction.java
```diff
@@ -0,0 +1,42 @@
...
+import org.apache.flink.api.common.functions.MapFunction;
+
+/**
+ * A map function that outputs the input element without any change.
+ *
+ * @param <T> Input element type.
+ */
+public class FlinkIdentityFunction<T> implements MapFunction<T, T> {
+
+  private static FlinkIdentityFunction<?> INSTANCE = new FlinkIdentityFunction<>();
```
cchepelov: CheckStyle asks this be named something matching '^[a-z][a-zA-Z0-9]*_?$', i.e. "instance" in lowercase

mxm: The reason is that the `final` keyword is missing, only final values are constants and can be capitalized. Fixing this in the merge commit, ensuring that the tests pass.

mxm:
I used a rebased version of David's commit, that's why this PR doesn't close. Sorry about that. Authorship is preserved through in 2dc923b0aeb627d0db5d7b2a41f56865659fd629 via the parent 785609f22d013411b7973bbf9e2d15c3c8171fb2.

Thank you!

PR #10038, opened 2019-11-08
Simplify Python test process

chadrik:
- Use pyproject.toml to install requirements for building beam
  - Consolidate a bunch of requirements that were spread across gen_protos, gradle, and tox
  - Remove the convoluted process in gen_protos where it tries to install deps, they're covered by pyproject.toml now
- Use tox to build sdist of beam instead of gradle:  one less virtualenv and fewer deps installed for no reason
- Remove test_requires because they're already handled by extras
- Don't call python setup.py nosetests because it rebuilds beam:  we want this to be entirely controlled by our tox (our pep517-build system)
- Stop running tests against code in the source directory!  This undermines the sandboxing that tox provides and requires more work from gradle and tox cleanup scripts to try to resolve.

One thing to note here is that despite the progress made by peps 517 and 518 toward improving the python build toolchains, it's still not a turnkey solution.  You can't just `pip build sdist` yet.  I tried a dozen different things, but it looks like the best solution is this [pep517](https://pypi.org/project/pep517/) module.  References:

- https://github.com/pypa/pip/issues/6041
- https://stackoverflow.com/questions/58753970/how-to-build-a-source-distribution-without-using-setup-py-file

Another problem is that we're always building cython on linux, rather than doing it conditionally based on the test we're running.  Need to figure out if we can control that with environment markers

------------------------

--- | --- | --- | --- | ---


chadrik:
Run Python PreCommit

chadrik:

sdks/python/scripts/run_pytest.sh
```diff
@@ -28,10 +28,10 @@ envname=${1?First argument required: suite base name}
 posargs=$2
 
 # Run with pytest-xdist and without.
-python setup.py pytest --addopts="-o junit_suite_name=${envname} \
+pytest --addopts="-o junit_suite_name=${envname} \
```
udim: You should be able to remove the `--addopts` flag and leave the value part.

chadrik: Ah yes, I was wrangling with that the other day.  It would simplify things if we could just get rid of run_pytest.sh.  Do we still need to run the tests with and without xdist?

udim: Yes, there are a few tests that are still incompatible with xdist, such as those using save_main_session.

chadrik: If the goal is to run both tests regardless of whether there is a failure (i.e. avoid early abort) then you can set the tox setting `ignore_errors=true`: https://tox.readthedocs.io/en/latest/config.html#conf-ignore_errors

> If false, a non-zero exit code from one command will abort execution of commands for that environment. If true, a non-zero exit code from one command will be ignored and further commands will be executed. The overall status will be “commands failed”, i.e. tox will exit non-zero in case any command failed.

I was trying to get run_pytest.sh to work with `tox -e py37-pytest -- path/to/test` and I was struggling to get the arg passing correct, and ultimately I just moved it to tox.ini and it worked.   With `ignore_errors=true` in place I don't think the shell script will be necessary.

udim: I could try helping with arg passing.

If you specify a certain test to run, one of the pytest runs will certainly fail since it either has or doesn't have the `no_xdist` marker. The script special-cases the 5 exit status from pytest to handle this scenario.

I'm probably missing something, because the way you set things up `tox -e py37-pytest -- path/to/test` should have failed (either the `-m not no_xdist` or the `-m no_xdist` command should have failed).

In any case, I'm strongly in favor of simplifying the build process by having as few shell scripts as possible.

There's also this plugin to suppress certain exit statuses:
https://github.com/yashtodi94/pytest-custom_exit_code

chadrik:
Run Python PreCommit

kennknowles:
I know it shouldn't be relevant, but I noticed the Java flakes and will file something about them.

kennknowles:
run java precommit

kennknowles:
Both already had Jiras, so Java precommit is g2g.

sdks/python/setup.py
```diff
@@ -226,12 +307,6 @@ def run(self):
     ]),
     install_requires=REQUIRED_PACKAGES,
     python_requires=python_requires,
-    test_suite='nose.collector',
-    setup_requires=['pytest_runner'],
-    tests_require= [
-        REQUIRED_TEST_PACKAGES,
-        INTERACTIVE_BEAM,
```
udim: I would expect that removing `INTERACTIVE_BEAM` would cause some tests that depend on these package to fail. I'm guessing there aren't any tests for this yet.

chadrik:
Thanks.  Slowly chipping away at this.  I'll bring this to the list when I have the remaining kinks worked out.

sdks/python/tox.ini
```diff
@@ -307,3 +331,10 @@ commands =
   coverage report --skip-covered
   # Generate report in xml format
   coverage xml
+
+[testenv:py37-interactive]
```
udim: Does this add any additional test coverage? I'd rather fold any 'interactive' tests into the py37 target if possible. Or at least not repeat all the py37 unit tests.

sdks/python/setup.py
chadrik: Any tests that require the interactive deps should add the 'interactive' extra.  So far I haven't run into any failures related to this.

sdks/python/tox.ini
chadrik: I think this addition is actually a rebase error on my part.  These tests were not intentionally added by me.  IIRC, they were added to tox on master then removed in a later commit, and I must've messed up on one of my rebases and left it in.

chadrik:
My last two jenkins jobs indicated it took 0s to run, but the logs show that it was aborted after 2 hours.  Any idea what could be going on?

udim:
> My last two jenkins jobs indicated it took 0s to run, but the logs show that it was aborted after 2 hours. Any idea what could be going on?

It looks like it hit the 2h Jenkins timeout. Should be extended to 3h once https://github.com/apache/beam/pull/10234 is merged and the next seed job runs. (I'll start one in #10234)

sdks/python/setup.py
```diff
@@ -226,12 +307,6 @@ def run(self):
     ]),
     install_requires=REQUIRED_PACKAGES,
     python_requires=python_requires,
-    test_suite='nose.collector',
```
markflyhigh: What's alternative collector it will use? Can we ensure the collection result unchanged?

udim: I believe it will still use nose, just not via setup.py. We can compare nosetests.xml files to verify.

chadrik: Correct, `test_requires` and `test_suite` are for the `python setup.py test` and derivative workflows.  That whole stack is now deprecated in favor of extras and pep517.

chadrik:
Run Python PreCommit

chadrik:
Run Python PreCommit

chadrik:
Run Python PreCommit

PR #10095, opened 2019-11-14
[BEAM-8664] [SQL] MongoDb project push down

11moon11:
- `MongoDbTable` should perform project push-down.
- Add tests.

Based on top of #10061.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/MongoDbTable.java
+++ b/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/MongoDbTable.java
@@ -33,7 +33,9 @@ import org.apache.beam.sdk.extensions.sql.meta.Table;
 import org.apache.beam.sdk.io.mongodb.FindQuery;
 import org.apache.beam.sdk.io.mongodb.MongoDbIO;
 import org.apache.beam.sdk.options.PipelineOptions;
+import org.apache.beam.sdk.schemas.FieldAccessDescriptor;
 import org.apache.beam.sdk.schemas.Schema;
+import org.apache.beam.sdk.schemas.utils.SelectHelpers;
 import org.apache.beam.sdk.transforms.DoFn;
 import org.apache.beam.sdk.transforms.JsonToRow;
 import org.apache.beam.sdk.transforms.MapElements;
@@ -92,10 +94,14 @@ public class MongoDbTable extends SchemaBaseBeamTable implements Serializable {
   @Override
   public PCollection<Row> buildIOReader(
       PBegin begin, BeamSqlTableFilter filters, List<String> fieldNames) {
-    MongoDbIO.Read readBuilder = MongoDbIO.read()
-        .withUri(dbUri)
-        .withDatabase(dbName)
-        .withCollection(dbCollection);
+    MongoDbIO.Read readBuilder =
+        MongoDbIO.read().withUri(dbUri).withDatabase(dbName).withCollection(dbCollection);
+
+    final FieldAccessDescriptor resolved =
+        FieldAccessDescriptor.withFieldNames(fieldNames)
+            .withOrderByFieldInsertionOrder()
+            .resolve(getSchema());
+    final Schema newSchema = SelectHelpers.getOutputSchema(getSchema(), resolved);
 
     if (!(filters instanceof DefaultTableFilter)) {
       throw new RuntimeException("Unimplemented at the moment.");
@@ -105,7 +111,7 @@ public class MongoDbTable extends SchemaBaseBeamTable implements Serializable {
       readBuilder.withQueryFn(FindQuery.create().withProjection(fieldNames));
     }
 
-    return readBuilder.expand(begin).apply(DocumentToRow.withSchema(getSchema()));
+    return readBuilder.expand(begin).apply(DocumentToRow.withSchema(newSchema));
   }
 
   @Override
@@ -117,7 +123,7 @@ public class MongoDbTable extends SchemaBaseBeamTable implements Serializable {
 
   @Override
   public ProjectSupport supportsProjects() {
-    return ProjectSupport.WITHOUT_FIELD_REORDERING;
+    return ProjectSupport.WITH_FIELD_REORDERING;
   }
 
   @Override
--- a/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/MongoDbTable.java
+++ b/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/MongoDbTable.java
@@ -104,7 +104,7 @@ public class MongoDbTable extends SchemaBaseBeamTable implements Serializable {
     final Schema newSchema = SelectHelpers.getOutputSchema(getSchema(), resolved);
 
     if (!(filters instanceof DefaultTableFilter)) {
-      throw new RuntimeException("Unimplemented at the moment.");
+      throw new AssertionError("Predicate push-down is unsupported, yet received a predicate.");
     }
 
     if (!fieldNames.isEmpty()) {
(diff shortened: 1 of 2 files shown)


11moon11:
Run sql postcommit

11moon11:
R: @TheNeuralBit 
cc: @apilloud

11moon11:
Run sql postcommit

11moon11:
Run sql postcommit

11moon11:
Run sql postcommit

11moon11:

11moon11:

sdks/java/extensions/sql/src/test/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/MongoDbReadWriteIT.java
```diff
@@ -187,7 +199,68 @@ public void testWriteAndRead() {
     PCollection<Row> output =
         BeamSqlRelUtils.toPCollection(readPipeline, sqlEnv.parseQuery("select * from TEST"));
 
-    assertEquals(output.getSchema(), SOURCE_SCHEMA);
+    assertEquals(SOURCE_SCHEMA, output.getSchema());
```
TheNeuralBit: nit: I generally prefer syntax like `assertThat(output.getSchema(), equalTo(SOURCE_SCHEMA))` since it makes it clear which is the value being tested and which is the expectation.

sdks/java/extensions/sql/src/test/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/MongoDbReadWriteIT.java
```diff
@@ -187,7 +199,68 @@ public void testWriteAndRead() {
...
+            + "ARRAY['123', '456']"
+            + ")";
+
+    BeamRelNode insertRelNode = sqlEnv.parseQuery(insertStatement);
+    BeamSqlRelUtils.toPCollection(writePipeline, insertRelNode);
+    writePipeline.run().waitUntilFinish();
+
+    BeamRelNode node = sqlEnv.parseQuery("select c_varchar, c_boolean, c_integer from TEST");
+    // Calc should be dropped, since MongoDb supports project push-down and field reordering.
+    assertThat(node, instanceOf(BeamPushDownIOSourceRel.class));
```
TheNeuralBit: I'm not crazy about this method of checking that the project push down happened, the fact that we produce a `BeamPushDownIOSourceRel` should be an implementation detail as far as this integration test is concerned.

Ideally we would get access to the list of fields retrieved from MongoDB somehow and make assertions about that instead. Is there any way to get that information? I don't know if this embedded mongo server supports that, but maybe some alternative (fongo?) does?

This isn't that big of a deal, if it's going to be a huge pain to re-work I'm fine leaving it as-is, but I think it's worth looking into.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/MongoDbTable.java
```diff
@@ -84,13 +91,41 @@
...
+        MongoDbIO.read().withUri(dbUri).withDatabase(dbName).withCollection(dbCollection);
+
+    final FieldAccessDescriptor resolved =
+        FieldAccessDescriptor.withFieldNames(fieldNames)
+            .withOrderByFieldInsertionOrder()
+            .resolve(getSchema());
+    final Schema newSchema = SelectHelpers.getOutputSchema(getSchema(), resolved);
+
+    if (!(filters instanceof DefaultTableFilter)) {
+      throw new RuntimeException("Unimplemented at the moment.");
```
TheNeuralBit: I think an `AssertionError` would be more appropriate here, since it should be impossible for this to happen. Also a more descriptive message would be good - maybe "Received a predicate even though predicate push-down is not supported"?

11moon11: I agree that the error message could be clearer here, but a predicate push-down PR for MongoDB is in the works, so this exception won't be there for too long.

TheNeuralBit: Well regardless, I think it should be an `AssertionError` to indicate that this is something that we believe to be impossible. Unless I'm misunderstanding and there is some way this could actually happen?

11moon11: Yes, you are correct, this code-path is impossible. I may have misunderstood your first message: do you mean something like `Preconditions.checkArgument` by `AssertionError`?

TheNeuralBit: Oh no I meant just throwing `AssertionError` instead of `RuntimeException`: `throw new AssertionError("blah blah")`

something like checkArgument may be a nice way to do the same thing, I'm not sure.

11moon11: I see, will update in a next commit, thanks!

sdks/java/extensions/sql/src/test/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/MongoDbReadWriteIT.java
11moon11: I agree, `assertEquals` is not very intuitive in terms of which parameter is the expected value and which is the actual. Will change in a next commit.

sdks/java/extensions/sql/src/test/java/org/apache/beam/sdk/extensions/sql/meta/provider/mongodb/MongoDbReadWriteIT.java
11moon11: I see the benefit of checking what actually is being retrieved.

Fields retrieved from MongoDB are returned as a `PCollection<Document>` from `MongoDBIO.expand()`, which is called by `MongoDbTable.buildIOReader(...)`.
An `IOSourceRel` does not do any data manipulations (just wraps an IO and does type conversion).
I am not exactly sure that checking values returned by `MongoDBIO.expand()` would be much different, unless I am misunderstanding something.

I also [looked](https://stackoverflow.com/questions/41550450/mongodb-list-previous-queries) into retrieving some sort of log (from the embedded database) to check what query actually got executed, but I am not sure that we can easily retrieve that data.

Alternatively, we could validate that `withQueryFn` on `MongoDbIO.Read` does indeed have projected fields set.

What do you think?

TheNeuralBit: > I also looked into retrieving some sort of log (from the embedded database) to check what query actually got executed, but I am not sure that we can easily retrieve that data.

This is the sort of thing I had in mind. Making an assertion about the query fn is a little better, but I still think that should be an implementation detail from the perspective of this test.

We're in sort of a weird position, something like a mock mongodb client would make this really easy, but that won't work for us since we need to be able to reference a hostname and port for the mongo instance. 

Your link about setting the profile level seems promising - it looks like if we do that we could get a list of operations/queries from the [`system.profile` collection](https://docs.mongodb.com/manual/tutorial/manage-the-database-profiler/#view-profiler-data). We could query that and assert there's just one operation with the expected projection. That seems like the best bet to me. It's kind of a pain though, so if you want to punt on it I'd be fine with just filing a jira describing this approach and referencing it here.

11moon11: Makes sense. Ended up validating projections logged in `system.profile` collection. Thanks for your help!

11moon11:
Run sql postcommit

TheNeuralBit:
It looks like the test failed in the post commit becase we didn't get anything out of `system.profile`, any idea what went wrong?

11moon11:
I think it is caused because `@After` drops the table.
Moving these lines to `@Before` should fix it.
https://github.com/apache/beam/pull/10095/commits/5c62169b1903107fe39205be47a3991800921164#diff-3660a76b56e047b917aef5fbf45fea3fR133-R135

11moon11:
Run sql postcommit

PR #10118, opened 2019-11-14
[BEAM-7473] Pack RangeTracker into restriction

boyuanzz:
Use (SourceBundle, RangeTracker) as restriction for SDFBoundedSourceWrapper. 

This PR fixes the issue that the system and SDF.process don't use the same instance of RangeTracker, which causes splitting failed.

R: @robertwb 
cc: @chamikaramj

### DIFF
--- a/sdks/python/apache_beam/io/iobase.py
+++ b/sdks/python/apache_beam/io/iobase.py
@@ -1402,6 +1402,18 @@ class _SDFBoundedSourceWrapper(ptransform.PTransform):
 
   NOTE: This transform can only be used with beam_fn_api enabled.
   """
+
+  class _SDFBoundedSourceRestriction(object):
+    """ A restriction wraps SourceBundle and RangeTracker. """
+    def __init__(self, source_bundle, range_tracker=None):
+      self.source_bundle = source_bundle
+      self.range_tracker = range_tracker
+
+    def __reduce__(self):
+      # The instance of RangeTracker shouldn't be serialized.
+      return (self.__class__, (self.source_bundle, ))
+
+
   class _SDFBoundedSourceRestrictionTracker(RestrictionTracker):
     """An `iobase.RestrictionTracker` implementations for wrapping BoundedSource
     with SDF. The tracked restriction is a (SourceBundle, RangeTracker) pair.
@@ -1411,14 +1423,15 @@ class _SDFBoundedSourceWrapper(ptransform.PTransform):
     Delegated RangeTracker guarantees synchronization safety.
     """
     def __init__(self, restriction):
-      source_bundle, _ = restriction
-      if not isinstance(source_bundle, SourceBundle):
+      if not isinstance(restriction,
+                        _SDFBoundedSourceWrapper._SDFBoundedSourceRestriction):
         raise ValueError('Initializing SDFBoundedSourceRestrictionTracker'
-                         ' requires a SourceBundle')
-      self._delegate_range_tracker = source_bundle.source.get_range_tracker(
-          source_bundle.start_position, source_bundle.stop_position)
-      self._source = source_bundle.source
-      self._weight = source_bundle.weight
+                         ' requires a _SDFBoundedSourceRestriction')
+      self._source = restriction.source_bundle.source
+      self._weight = restriction.source_bundle.weight
+      self._delegate_range_tracker = self._source.get_range_tracker(
+          restriction.source_bundle.start_position,
+          restriction.source_bundle.stop_position)
 
     def current_progress(self):
       return RestrictionProgress(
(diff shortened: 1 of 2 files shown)


boyuanzz:
Run Python PreCommit

sdks/python/apache_beam/io/iobase.py
```diff
@@ -1404,18 +1404,21 @@ class _SDFBoundedSourceWrapper(ptransform.PTransform):
   """
   class _SDFBoundedSourceRestrictionTracker(RestrictionTracker):
     """An `iobase.RestrictionTracker` implementations for wrapping BoundedSource
-    with SDF.
+    with SDF. The tracked restriction is a (SourceBundle, RangeTracker) pair.
+    In order to save bytes sent across the wire, the RangeTracker is set as
+    system tracking RangeTracker only when current_restriction is called.
```
robertwb: This is not only to save bytes across the wire--range trackers are not meant to be serialized. I might say that logically the restriction is a (SourceBundle, RangeTracker) pair, but the second argument is never serialized.

sdks/python/apache_beam/io/iobase.py
```diff
@@ -1487,10 +1497,10 @@ def split(self, element, restriction):
...
       for source_bundle in source_bundles:
-        yield source_bundle
+        yield (source_bundle, None)
 
     def restriction_size(self, element, restriction):
-      return restriction.weight
+      return restriction[0].weight
 
     def restriction_coder(self):
       return coders.DillCoder()
```
robertwb: Do we ever encode the result of current_restriction()? (For example, due to self-checkpointing?)

robertwb: Thinking about this some more, I think it'd be safer and cleaner to create an actual _SDFBoundedSourceRestriction class that holds the SourceBundle and the (lazily-initialized) underlying restriction rather than using a tuple. Its `__reduce__` method could return only the former, and its constructor only take the former. Many of the methods here would simply delegate to it.

boyuanzz: > Do we ever encode the result of current_restriction()? (For example, due to self-checkpointing?)

I would say, no. `current_restriction` is expected be called inside SDF.DoFn(). Split(include self-checkpoint) should go through `try_split`.

But I really like the `_SDFBoundedSourceRestriction`, definitely a better way to go. Thanks for your help!

boyuanzz:
Run Python PreCommit

boyuanzz:
Changed usage to `_SDFBoundedSourceRestriction`. Please take another look @robertwb

sdks/python/apache_beam/io/iobase.py
robertwb: Update comment?

sdks/python/apache_beam/io/iobase.py
```diff
@@ -1402,20 +1402,36 @@ class _SDFBoundedSourceWrapper(ptransform.PTransform):
...
         raise ValueError('Initializing SDFBoundedSourceRestrictionTracker'
-                         'requires a SourceBundle')
-      self._delegate_range_tracker = restriction.source.get_range_tracker(
-          restriction.start_position, restriction.stop_position)
-      self._source = restriction.source
-      self._weight = restriction.weight
+                         ' requires a _SDFBoundedSourceRestriction')
+      self._source = restriction.source_bundle.source
+      self._weight = restriction.source_bundle.weight
+      self._delegate_range_tracker = self._source.get_range_tracker(
```
robertwb: How about just adding a range_tracker() method to _SDFBoundedSourceRestriction? Then _delegate_range_tracker would not have to be tracked here at all.

sdks/python/apache_beam/io/iobase.py
```diff
@@ -1424,11 +1440,12 @@ def current_progress(self):
     def current_restriction(self):
       start_pos = self._delegate_range_tracker.start_position()
       stop_pos = self._delegate_range_tracker.stop_position()
-      return SourceBundle(
-          self._weight,
-          self._source,
-          start_pos,
-          stop_pos)
+      return _SDFBoundedSourceWrapper._SDFBoundedSourceRestriction(
```
robertwb: Rather than re-creating it here, how about just storing (and returning) the initial _SDFBoundedSourceRestriction object (which does lazy initialization in range_tracker())?

sdks/python/apache_beam/io/iobase.py
```diff
@@ -1456,10 +1473,17 @@ def try_split(self, fraction_of_remainder):
         residual_weight = self._weight - primary_weight
         # Update self._weight to primary weight
         self._weight = primary_weight
-        return (SourceBundle(primary_weight, self._source, start_pos,
-                             split_pos),
-                SourceBundle(residual_weight, self._source, split_pos,
-                             stop_pos))
+        return (
```
robertwb: You could add a split method to the _SDFBoundedSourceRestriction that would contain most of the logic of this method.

boyuanzz:
Thanks for your help!

PR #10132, opened 2019-11-15
[BEAM-8016] Pipeline Graph

KevinGG:
1. Removed dependency of pipeline_analyzer, display_manager and
pipeline_graph_renderer from interactive_runner.
2. Replaced the logic of analyzing pipeline with pipeline_instrument
module including how PipelineResult subclass works.
3. Replaced the Pipeline Graph rendering logic with pipeline_graph
module. Implemented the display_graph interface.
4. The cache manager is now held by a static environment object and
shared with all interactive runners. The environment and its cache
manager, however, are intialized lazily by any interactive runner in
current main loop/process/thread.
5. Updated the example ipynotebook.
------------------------

--- | --- | --- | --- | ---


sdks/python/apache_beam/runners/interactive/display/pipeline_graph.py
```diff
@@ -136,14 +159,50 @@ def _generate_graph_dicts(self):
...
             self._edge_to_vertex_pairs[pcoll_id].append(
                 (producer_name, consumer_name))
-            edge_dict[(producer_name, consumer_name)] = {}
+            if self._pin:
+              edge_dict[(producer_name, consumer_name)] = {
+                  'label':
+                    self._pin.cacheable_var_by_pcoll_id(pcoll_id)
+              }
+            else:
+              edge_dict[(producer_name, consumer_name)] = {}
```
KevinGG: The entire logic will be simplified once data-centric user flow is in-place and pcollections are rendered as nodes rather than labels on edges.

The difference introduced:
Before:

After:



In the future:
With data-centric user flow, notebook cell execution metadata and new Beam pipeline graph proposal:

KevinGG:
R: @davidyan74 
R: @rohdesamuel 

Adding Sindy to see if the change conforms to what she has done to decouple renderer logic.
R: @qinyeli 
Hi Sindy, we are basically:

- moving away from the display manager which would always clear the output area of a notebook cell.
- rendering the display_graph directly, but still using available render_option.
- not doing sampling cache or visualization at all on pipeline graph. The logic is replaced by new `show()` (currently called `visualize` but not implemented) API in `interactive_beam` module.
- users still can get list of data from pipeline result, so the code in existing notebook example using matplotlib still works.

Please let me know if you have any concern with this change.

Also adding Pablo as the committer.
R: @pabloem 

Thank you all!

KevinGG:
Run Python PreCommit

sdks/python/apache_beam/runners/interactive/interactive_runner.py
```diff
@@ -117,123 +119,43 @@ def apply(self, transform, pvalueish, options):
...
         options)
 
     if not self._skip_display:
-      display = display_manager.DisplayManager(
-          pipeline_proto=pipeline_proto,
-          pipeline_analyzer=analyzer,
-          cache_manager=self._cache_manager,
-          pipeline_graph_renderer=self._renderer)
-      display.start_periodic_update()
+      pg = pipeline_graph.PipelineGraph(pin.original_pipeline,
```
qinyeli: Let's rename this into something more descriptive.

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
```diff
@@ -314,10 +316,29 @@ def cache_key(self, pcoll):
...
+    that referenced PCollection. If the PCollection is anonymous, return ''.
+    """
+    return self._cacheable_var_by_pcoll_id.get(pcoll_id, '')
+
 
 def pin(pipeline, options=None):
-  """Creates PipelineInstrument for a pipeline and its options with cache."""
+  """Creates PipelineInstrument for a pipeline and its options with cache.
+
+  This is the shorthand for doing 3 steps: 1) compute once for metadata of given
```
qinyeli: nit: of the given

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
```diff
@@ -314,10 +316,29 @@ def cache_key(self, pcoll):
...
+  runner pipeline and everything watched from user pipelines; 2) associate info
+  between runner pipeline and its corresponding user pipeline, eliminate data
```
qinyeli: nit: between the runner pipeline

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
```diff
@@ -337,6 +358,7 @@ def cacheables(pcolls_to_pcoll_id):
   """
   pcoll_version_map = {}
   cacheables = {}
+  cacheable_var_by_pcoll_id = {}
```
qinyeli: Can't comment on already-committed lines, so commenting here.
I didn't get what a "cache desired PCollection" is. cache-desired PCollection looks slightly better, but still doesn't tell what it is. PCollections that need to be cached?

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
```diff
@@ -80,7 +80,10 @@ def __init__(self, pipeline, options=None):
     # A mapping from PCollection id to python id() value in user defined
     # pipeline instance.
     (self._pcoll_version_map,
-     self._cacheables) = cacheables(self.pcolls_to_pcoll_id)
+     self._cacheables,
```
qinyeli: Adding comment here since GitHub does not allow comments on unchanged lines.
It's probably not obvious what it means to "instrument" a pipeline. From my context back when working on Interactive Beam, I would guess it's something similar to the original pipeline_analyzer, (and I have to admit that "analyze" pipeline was not a much better.), but it's still fuzzy.

Maybe adding some comment summarizing what pipeline_instrument does would be good. In a separate PR

qinyeli:
Thanks Ning! I haven't been catching up with the latest changes in Interactive Beam, so consider my comments as from someone with somewhat context but not all. A lot of my comments are "This is not very obvious, can you elaborate?"

So it looks good in general. Some comments in line. Also a couple of questions out of curiosity.
* Are we getting rid of the tooltips displaying the intermediate results? Do they not fit in the new model?
* What do the PCollections look like if the user did not specify the PCollection name as a variable?

KevinGG:
Thanks for the comments!
> * Are we getting rid of the tooltips displaying the intermediate results? Do they not fit in the new model?
We are offering a `show` API to users so that they can visualize a larger set of their data dynamically instead of peeking through a random static sample (which we still offer if the user calls `show` in an ipython terminal not a notebook web frontend).
And with new Beam pipeline graph proposal and some pipeline graph library WIP, the tooltip in the future might show other metadata such as elapse/throughput to provide a consistent user experience that is similar to what users have on Dataflow.
> * What do the PCollections look like if the user did not specify the PCollection name as a variable?
Those PCollections will not be cached. The idea is when building a pipeline, if the user does not assign a PCollection to a variable, they would not be able to build the pipeline further from it and they cannot invoke `show(pcoll)` because they don't have access to `pcoll` in their code.
Before, we have had the `leaf pcollection` concept for PCollections who have never been used as inputs. It doesn't work for PCollections consumed by sinks (with input but no output) even if the user has assigned them to a variable and they look like hanging PCollections. It also doesn't work in a notebook environment where new transforms can be added at different locations and PCollections can be re-evaluated due to cell-re-execution.

sdks/python/apache_beam/runners/interactive/interactive_runner.py
KevinGG: Sure! I'll use the full name `pipeline_graph`.

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
KevinGG: Yes, I agree! It's similar to what pipeline_analyzer was doing but using `PipelineVisitor` and some `globals` instead of pipeline protos. We haven't achieved a consensus about what this should be called. We also have `cache augmented pipeline` and etc.. But none of the names describe what the module does accurately because the implementation details cover so many different execution routes and scenarios.
We'll definitely get back to the documentation here once we make the final naming decision. Thanks!

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
KevinGG: Thanks, done.

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
KevinGG: Thanks, done.

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
KevinGG: Yes, it is actually PCollections that need to be cached. I thought it was a little bit wordy and chose the shorter form. I'll rephrase it to "PCollections that need to be cached" instead for clarity.

sdks/python/apache_beam/runners/interactive/interactive_runner.py
```diff
@@ -117,123 +119,43 @@ def apply(self, transform, pvalueish, options):
...
-    self._runner = runner
-    self._pipeline_info = pipeline_info
-    self._cache_manager = cache_manager
-    self._pcolls_to_pcoll_id = pcolls_to_pcoll_id
-
-  def _cache_label(self, pcoll):
-    pcoll_id = self._pcolls_to_pcoll_id[str(pcoll)]
-    return self._pipeline_info.cache_label(pcoll_id)
+    self._underlying_result = underlying_result
+    self._pin = pin
```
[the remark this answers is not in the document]
KevinGG: Thanks! I'll add the docstring for `__init__`.

sdks/python/apache_beam/runners/interactive/display/pipeline_graph.py
```diff
@@ -92,9 +104,20 @@ def __init__(self,
...
     return self._get_graph().to_string()
 
+  def display_graph(self):
+    rendered_graph = self._renderer.render_pipeline_graph(self)
+    if ie.current_env().is_in_notebook:
+      try:
+        from IPython.core import display
+        display.display(display.HTML(rendered_graph))
+      except ImportError:  # Unlikely to happen when is_in_notebook.
+        pass
```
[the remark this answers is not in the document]
KevinGG: Added a warning level logging here.

KevinGG:
Run Portable_Python PreCommit

sdks/python/apache_beam/runners/interactive/interactive_runner.py
```diff
@@ -65,8 +65,12 @@ def __init__(self,
     """
     self._underlying_runner = (underlying_runner
                                or direct_runner.DirectRunner())
-    self._cache_manager = cache.FileBasedCacheManager(cache_dir, cache_format)
-    self._renderer = pipeline_graph_renderer.get_renderer(render_option)
+    if not ie.current_env().cache_manager():
+      ie.current_env().set_cache_manager(
+          cache.FileBasedCacheManager(cache_dir,
+                                      cache_format))
+    self._cache_manager = ie.current_env().cache_manager()
```
rohdesamuel: Why make the CacheManager a global instead of per InteractiveRunner? The way that it is written here is that every InteractiveRunner will receive the same CacheManager. What if the user wants to create a new InteractiveRunner?

KevinGG: Yes, they will share the same cache manager since the PCollections, PTransforms and Pipelines the user has defined are global.

The notebook users own their source code in notebooks, so they should always be able to take advantage of Interactive Beam features if they are using the same Pipeline and same PCollection.

Cache belongs to the implicit implementation details. So the user should not know about it or depend on it directly throughout notebook usage. The most they can do are `watch` and `show`.

sdks/python/apache_beam/runners/interactive/display/pipeline_graph.py
```diff
@@ -107,6 +135,16 @@ def _top_level_transforms(self):
         top_level_transform_proto = transforms[top_level_transform_id]
         yield top_level_transform_id, top_level_transform_proto
 
+  def _decorate(self, value):
+    """Decorates label-ish values used for rendering in dot language.
+
+    Escapes special characters in the given str value for dot language. '"' is
```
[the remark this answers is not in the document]
KevinGG: Thanks, David! Yes, `'\\"'` works for escaping `"` and I'll add a test for this class.

KevinGG:
> I would say always avoid abbreviations that can be confusing. Perhaps just spell it out? >_pipeline_instrument

Spelled out all variable names of this in the module.

sdks/python/apache_beam/runners/interactive/display/pipeline_graph.py
```diff
@@ -107,6 +135,17 @@ def _top_level_transforms(self):
         top_level_transform_proto = transforms[top_level_transform_id]
         yield top_level_transform_id, top_level_transform_proto
 
+  def _decorate(self, value):
+    """Decorates label-ish values used for rendering in dot language.
+
+    Escapes special characters in the given str value for dot language. Please
+    escape all PTransform unique names when building dot representation.
+    Otherwise, special characters will break the graph rendered.
```
davidyan74: Is this statement still valid? Does 'value' need to be escaped any more?

sdks/python/apache_beam/runners/interactive/display/pipeline_graph.py
```diff
@@ -55,9 +63,17 @@ def __init__(self,
       pipeline: (Pipeline proto) or (Pipeline) pipeline to be rendered.
       default_vertex_attrs: (Dict[str, str]) a dict of default vertex attributes
       default_edge_attrs: (Dict[str, str]) a dict of default edge attributes
+      render_option: (str) this parameter decides how the pipeline graph is
+          rendered. See display.pipeline_graph_renderer for available options.
     """
     self._lock = threading.Lock()
     self._graph = None
+    self._pin = None
```
davidyan74: Can we also spell out this one?

sdks/python/apache_beam/runners/interactive/display/pipeline_graph.py
KevinGG: Yes, I think the `escape` has a different meaning in DOT's doc. When we say escape (or 
 URL encoding), we mean it automatically adds things like `\` before special characters or convert ` ` into `%20`. 

>In computing and telecommunication, an escape character is a character which invokes an alternative interpretation on subsequent characters in a character sequence. An escape character is a particular case of metacharacters. Generally, the judgment of whether something is an escape character or not depends on context.

But in DOT, it kind of means the opposite.

Because characters like `:` or `[]` have its own meanings in DOT, the recommended way is to enclose them with `"` so that they will be recognized as string literals.
That's why originally the comment says `"` is the only character that cannot be escaped because `"` is the character that escapes everything in-between.
Now, we do a replacement to convert `"` into `\\"` so that it becomes `\"` in dot and rendered as `"`.
However, `\` is still not escaped because it's used as an escape character even in-between `"`. This never ends as long as python literals and DOT work like this.

sdks/python/apache_beam/runners/interactive/display/pipeline_graph.py
KevinGG: Sure, I'll make the change here.

sdks/python/apache_beam/runners/interactive/display/pipeline_graph.py
KevinGG: We escape both `\` and `"` and have documented how the process works behind the scene.

KevinGG:
Run Python PreCommit

KevinGG:
R: @pabloem 
Hi Pablo, could you please take a last round of review for this PR? Thanks!

pabloem:
Does this mean that non-cached PCollections will also be displayed in the dot graph with a generated name?

KevinGG:
> Does this mean that non-cached PCollections will also be displayed in the dot graph with a generated name?

If any of the non-cached PCollections belongs to an output of a top level transform, then yes, they will be rendered as a circle with no name (we use `label=""` to override whatever name they might have) in it.
Most of the non-cached PCollections exist in-between non-top-level transforms (such as sub transforms of a composite transform the user defines) as intermediate input-outputs within a composite transform. Since the DOT doesn't support expanding composite transforms, they will not be rendered. If in the future, we support expanding those transforms in graph, they will be rendered as circles without names too.

pabloem:
Thanks Ning!

PR #10151, opened 2019-11-19
[BEAM-7116] Remove use of KV in Schema transforms

reuvenlax:
Beam's KV type has no schema and due to special casing of KvCoder in Beam it is difficult to give it one. Here we modify the Beam schema transforms that return PCollection<KV> to instead return PCollection<Row> where the Row contains key and value fields. This is possible now that we support large iterables in schemas.

### DIFF
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/transforms/CoGroup.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/transforms/CoGroup.java
@@ -20,6 +20,7 @@ package org.apache.beam.sdk.schemas.transforms;
 import com.google.auto.value.AutoValue;
 import java.io.Serializable;
 import java.util.Collections;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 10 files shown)


reuvenlax:
R: @TheNeuralBit

reuvenlax:
run sql postcommit

reuvenlax:
friendly ping

TheNeuralBit:
Run Java PreCommit

suztomo:
@reuvenlax @TheNeuralBit  `AvroSchemaTest.testAvroPipelineGroupBy` started failing after merging this commit. Do you think this PR is related?

https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Spark/6002/

```
org.apache.beam.sdk.Pipeline$PipelineExecutionException: java.lang.IllegalStateException: ValueIterator can't be iterated more than once,otherwise there could be data lost
```

image

TheNeuralBit:
Yeah this PR must be the cause. It's not clear to me why this only seems to be a problem with the Spark runner.. I'll file a jira and exclude the test from Spark for now.

suztomo:
@TheNeuralBit Thank you for confirmation.

TheNeuralBit:
#10358 should fix Spark ValidatesRunner

reuvenlax:
The short answer is that the failure is caused by a limitation of the Spark
runner - the Beam model allows the Iterable passed into a GroupByKey to
reiterated, but Spark is unable to support that.

The longer answer is that this is exposing a performance bug in the Group
transform. We were accidentally walking over the iterator, which we should
not be doing here. I'll send a PR to fix it.

PR #10172, opened 2019-11-20
[BEAM-8747] Guava dependency cleanup

suztomo:
Fixes BEAM-8747.

As per email thread "Completeness of Beam Java Dependency Check Report" ([archive link](https://lists.apache.org/thread.html/c477d120a4c4626cbe675f8b03d84c6fe7938e36c8e2b55c492224cf@%3Cdev.beam.apache.org%3E)),  @kennknowles found there are "declared but unused" and "used but undeclared" Guava dependencies in some of modules. This PR addresses the findings.

I used gradle-lint-plugin to confirm Ken's findings. See detail in the ticket https://issues.apache.org/jira/projects/BEAM/issues/BEAM-8747 .

- GCP connector turned out using guava in fully-qualified class name ([link](https://issues.apache.org/jira/browse/BEAM-8747?focusedCommentId=16978573&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16978573); no action taken.
- sdks/java/extensions/sql/jdbc had "declared but unused" Guava dependency.
- runners/google-cloud-dataflow-java/worker depends on 2 classes without declaring. These classes should not depend on Guava class. So I updated them to use vendored Guava.
- sdks/java/io/kinesis has "used but undeclared" Guava dependency.
- sdks/java/extensions/sql/zetasql has "used but undeclared" Guava dependency.
- sdks/java/io/elasticsearch-tests/elasticsearch-tests-2 has "declared but unused" Guava dependency.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/counters/CounterUpdateAggregators.java
+++ b/runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/counters/CounterUpdateAggregators.java
@@ -18,11 +18,11 @@
 package org.apache.beam.runners.dataflow.worker.counters;
 
 import com.google.api.services.dataflow.model.CounterUpdate;
-import com.google.common.collect.ImmutableMap;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import org.apache.beam.runners.dataflow.worker.MetricsToCounterUpdateConverter.Kind;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.ImmutableMap;
 
 public class CounterUpdateAggregators {
 
--- a/sdks/java/extensions/sql/zetasql/build.gradle
+++ b/sdks/java/extensions/sql/zetasql/build.gradle
@@ -31,6 +31,7 @@ dependencies {
   compile project(":sdks:java:core")
   compile project(":sdks:java:extensions:sql")
   compile library.java.vendored_calcite_1_20_0
+  compile library.java.guava
   compile library.java.grpc_all
   compile library.java.protobuf_java
   compile library.java.protobuf_java_util
--- a/sdks/java/io/kinesis/build.gradle
+++ b/sdks/java/io/kinesis/build.gradle
@@ -34,6 +34,7 @@ dependencies {
   compile library.java.slf4j_api
   compile library.java.joda_time
   compile library.java.jackson_dataformat_cbor
+  compile library.java.guava
   compile library.java.aws_java_sdk_cloudwatch
   compile library.java.aws_java_sdk_core
   compile library.java.aws_java_sdk_kinesis
--- a/runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/fn/control/TimerReceiver.java
+++ b/runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/fn/control/TimerReceiver.java
@@ -17,7 +17,6 @@
  */
 package org.apache.beam.runners.dataflow.worker.fn.control;
 
-import com.google.common.annotations.VisibleForTesting;
 import java.io.IOException;
 import java.util.Collections;
 import java.util.HashMap;
@@ -41,6 +40,7 @@ import org.apache.beam.sdk.transforms.windowing.GlobalWindow;
 import org.apache.beam.sdk.transforms.windowing.PaneInfo;
 import org.apache.beam.sdk.util.WindowedValue;
 import org.apache.beam.sdk.values.KV;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.annotations.VisibleForTesting;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
--- a/sdks/java/extensions/sql/jdbc/build.gradle
+++ b/sdks/java/extensions/sql/jdbc/build.gradle
@@ -36,7 +36,6 @@ dependencies {
   compile "jline:jline:2.14.6"
   compile "sqlline:sqlline:1.4.0"
   compile library.java.slf4j_jdk14
-  compile library.java.guava
   testCompile project(path: ":sdks:java:io:google-cloud-platform", configuration: "testRuntime")
   testCompile library.java.junit
   testCompile library.java.hamcrest_core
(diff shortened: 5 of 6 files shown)


suztomo:
R: @kennknowles

sdks/java/io/kinesis/build.gradle
```diff
@@ -34,6 +34,7 @@ dependencies {
   compile library.java.slf4j_api
   compile library.java.joda_time
   compile library.java.jackson_dataformat_cbor
+  compile library.java.guava
```
suztomo: @aromanenko-dev  As Kenn responded to you in the email, it's good practice to explicitly declare dependency if you're using the class. Without an explicit dependency, it's not clear where the class is coming from (could be 3-hop deep away from your project) and may throw difficult-to-troubleshoot compilation/runtime errors in future when some of the dependencies change.

suztomo:
Run Java PreCommit

sdks/java/io/kinesis/build.gradle
aromanenko-dev: Thank you for fixing this. Don't we need to be aligned with Guava version which is used in `com.amazonaws:amazon-kinesis-producer` in this case? In Beam, we have guava version  "20.0", but KPL depends on version "18.0"

suztomo: No, version 18.0 has not been used.

This is dependency tree of sdks/java/io/kinesis **before** this PR: https://gist.github.com/suztomo/22a5d15e21b4d65507c3be3f38139bb9 . Beam and sdks/java/io/kinesis have been pinning `com.google.guava:guava`'s version to 20.0 through `resolutionStrategy` by [BeamModulePlugin](https://github.com/apache/beam/blob/master/buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy#L1245).

suztomo:
Run Java PreCommit

suztomo:
Run Java PreCommit

sdks/java/io/kinesis/build.gradle
aromanenko-dev: Yes, this is what I was concerned about since Amazon KPL depends on Guava 18.0 ([link](https://search.maven.org/artifact/com.amazonaws/amazon-kinesis-producer/0.13.1/jar)). Though, it's a case for other KPL deps as well and it should be ok if it works in the end as expected.

runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/counters/CounterUpdateAggregators.java
```diff
@@ -18,11 +18,11 @@
 package org.apache.beam.runners.dataflow.worker.counters;
 
 import com.google.api.services.dataflow.model.CounterUpdate;
-import com.google.common.collect.ImmutableMap;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import org.apache.beam.runners.dataflow.worker.MetricsToCounterUpdateConverter.Kind;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.ImmutableMap;
```
kennknowles: Is this checkstyle check disabled for the Dataflow worker?  https://github.com/apache/beam/blob/master/sdks/java/build-tools/src/main/resources/beam/checkstyle.xml#L102

sdks/java/io/kinesis/build.gradle
kennknowles: @aromanenko-dev the point is that this library has already been pinned to 20.0 for a long time, I think because `BeamModulePlugin` is going in and overriding things even when it doesn't have to. So we know it currently is working and @suztomo is just recording the status quo.

By the way it isn't so simple as KPL depending on 18.0.  This IO module also links against the Kinesis client which depends on 26.0-jre and it has been pinned downward to 20.0. So there's not one version of Guava that these AWS libs use.

Here's the excerpt from the linked dep tree, which is already how it has been for some time.

```
compile - Dependencies for source set 'main' (deprecated, use 'implementation' instead).
+--- com.amazonaws:amazon-kinesis-client:1.13.0
|    +--- com.google.guava:guava:26.0-jre -> 20.0
+--- com.amazonaws:amazon-kinesis-producer:0.13.1
|    +--- com.google.guava:guava:18.0 -> 20.0
```

Highest priority is choosing a version of Guava that works for this connector. So it is OK for it to diverge and use its own version. The aliases at `library.java.*` exist just to help Beam modules cooperate. We probably want to keep moving `library.java.guava` forward, for example I think we need an upgrade support Java 11. If that breaks this module, then we will just have to fork the version.

Is there enough test coverage to catch any problems? Is the Kinesis client library open source so you can clone it and see if it compiles with a new version? That could be a way to check full API compatibility that might be missed by testing.

sdks/java/extensions/sql/zetasql/build.gradle
```diff
@@ -31,6 +31,7 @@ dependencies {
   compile project(":sdks:java:core")
   compile project(":sdks:java:extensions:sql")
   compile library.java.vendored_calcite_1_20_0
+  compile library.java.guava
```
kennknowles: Here we may need to diverge eventually. ZetaSQL is a Google "throw over the wall" open source. That means that it generally will track against the latest Guava or even newer internal changes. Just a note - if we hit any problems it may simply have to diverge.

sdks/java/io/kinesis/build.gradle
aromanenko-dev: @kennknowles Thank you for details. Yes, there is a known discrepancy between guava versions used in different Kinesis libs and I believe we test it sufficiently with unit and integration tests. Though, we still have a task to move IT into Apache Jenkins env and run it there (the problem that we run it against real AWS instances but it's another question, not elated to this PR).

Back to initial question, to make it clear - I don't insist, I just asked to clarify =) And I agree that since we know that it's working we should not worry too much about that. Kinesis client library is [open source](https://github.com/awslabs/amazon-kinesis-client), so yes, we can check it if needed.

aromanenko-dev: So, it's LGTM for KinesisIO from my side

runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/counters/CounterUpdateAggregators.java
suztomo: Yes, I see it's disabled.
https://github.com/apache/beam/blob/c2f0d28/runners/google-cloud-dataflow-java/worker/build.gradle#L111-L113

sdks/java/extensions/sql/zetasql/build.gradle
lukecwik: diverging will be bad for the project as its unlikely that this won't cause various IOs/runners to fail

kennknowles: I was just pointing out that if there comes a time when there is no version of Guava that works for both ZetaSQL and Kinesis we might have no choice. Neither one can be vendored. ZetaSQL *might* be possible to vendor by putting a layer between user-supplied parameter values for prepared statements.

PR #10222, opened 2019-11-26
[BEAM-8822] Hadoop client version 2.8.5 from 2.7

suztomo:
Upgrading Hadoop client version 2.7 (EOL) to 2.8.5 (the latest in 2.8.X release).

CC: @iemejia 

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/io/hadoop-file-system/src/main/java/org/apache/beam/sdk/io/hdfs/HadoopFileSystemRegistrar.java
+++ b/sdks/java/io/hadoop-file-system/src/main/java/org/apache/beam/sdk/io/hdfs/HadoopFileSystemRegistrar.java
@@ -45,8 +45,9 @@ public class HadoopFileSystemRegistrar implements FileSystemRegistrar {
 
   private static final List<String> HA_SCHEMES = Arrays.asList("hdfs", "webhdfs");
 
-  // Using hard-coded value for compatibility, because hadoop-dfs-client:2.7's DFSConfigKeys holds
-  // this value while it is HdfsClientConfigKeys that holds the value in version 2.8.
+  // Using hard-coded value to avoid incompatibility between HDFS client
+  // (org.apache.hadoop:hadoop-dfs-client) version 2.7's DFSConfigKeys and version 2.8's
+  // HdfsClientConfigKeys.
   private static final String CONFIG_KEY_DFS_NAMESERVICES = "dfs.nameservices";
 
   @Override
--- a/sdks/java/io/hadoop-file-system/src/main/java/org/apache/beam/sdk/io/hdfs/HadoopFileSystemRegistrar.java
+++ b/sdks/java/io/hadoop-file-system/src/main/java/org/apache/beam/sdk/io/hdfs/HadoopFileSystemRegistrar.java
@@ -45,6 +45,10 @@ public class HadoopFileSystemRegistrar implements FileSystemRegistrar {
 
   private static final List<String> HA_SCHEMES = Arrays.asList("hdfs", "webhdfs");
 
+  // Using hard-coded value for compatibility, because hadoop-dfs-client:2.7's DFSConfigKeys holds
+  // this value while it is HdfsClientConfigKeys that holds the value in version 2.8.
+  private static final String CONFIG_KEY_DFS_NAMESERVICES = "dfs.nameservices";
+
   @Override
   public Iterable<FileSystem> fromOptions(@Nonnull PipelineOptions options) {
     final List<Configuration> configurations =
@@ -75,7 +79,7 @@ public class HadoopFileSystemRegistrar implements FileSystemRegistrar {
       builder.add(new HadoopFileSystem(scheme, configuration));
       registeredSchemes.add(scheme);
     }
-    final String nameServices = configuration.get(HdfsClientConfigKeys.DFS_NAMESERVICES);
+    final String nameServices = configuration.get(CONFIG_KEY_DFS_NAMESERVICES);
     if (nameServices != null && !nameServices.isEmpty()) {
       // we can register schemes that are support by HA cluster
       for (String scheme : HA_SCHEMES) {
--- a/sdks/java/io/hadoop-file-system/src/main/java/org/apache/beam/sdk/io/hdfs/HadoopFileSystemRegistrar.java
+++ b/sdks/java/io/hadoop-file-system/src/main/java/org/apache/beam/sdk/io/hdfs/HadoopFileSystemRegistrar.java
@@ -36,7 +36,6 @@ import org.apache.beam.sdk.options.PipelineOptions;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.ImmutableList;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.Iterables;
 import org.apache.hadoop.conf.Configuration;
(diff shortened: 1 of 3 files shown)


sdks/java/io/hadoop-format/build.gradle
```diff
@@ -73,6 +73,8 @@ dependencies {
     exclude group: "org.apache.spark", module: "spark-sql_2.10"
     exclude group: "org.apache.storm", module: "storm-core"
   }
+  // elasticsearch-hadoop 5.0.0 uses commons-httpclient's URIException
+  testCompile "commons-httpclient:commons-httpclient:3.1"
```
suztomo: `Elasticsearch-hadoop` appear 12 lines above this line.

suztomo:
Run Java PostCommit

suztomo:
Run Java PostCommit

suztomo:
Run Python PreCommit

suztomo:

suztomo:
R: @iemejia

suztomo:
Flink Runner Nexmark Tests

suztomo:
Spark Runner Nexmark Tests

suztomo:

suztomo:

suztomo:
Run Java PortabilityApi PostCommit

suztomo:
Run Flink ValidatesRunner

suztomo:
The checks other than the problematic Python checks went successful.

iemejia:
Changes look good for the goal, however it seems that my comment about Hadoop 2.7.x being already EOL in upstream Hadoop was a bit optimistic. After a quick review most of the major users and systems still support 2.7 and ‘even worse’ 2.6. So I was wondering how can we ensure that new patches on Hadoop modules don’t break backwards compatibility.

One idea could be to add an extra set of tests that run the build for the provided previous versions of the dependencies. But I don’t have a clue how to do this in gradle. Do you by any chance? (Note that a fix for this will solve other compatibility issues we have with other systems and provided dependencies e.g. Spark and Kafka). More details in [BEAM-4087](https://issues.apache.org/jira/browse/BEAM-4087)

For the case of the current patch it is not backwards compatible at the source level (because it introduces a dep on a class that did not exist in previous hadoop versions) and the proposed tests may have helped to discover that (but this can be easily fixed by extracting the constant into a local String or just by keeping the warning on the old class).

suztomo:
The lifecycle of a product is to help to avoid supporting old product, and to guide decision like this.

Still, we should be clear about compatibility. I don't use Hadoop and thus cannot test it. Let me send email to dev@ for anyone can verify the compatibility.

lukecwik:
I believe what @iemejia is saying is that we should do something like [Flink](https://github.com/apache/beam/tree/master/runners/flink) where it runs tests using Flink 1.7, 1.8 and 1.9. In the Hadoop case we would run the tests using version 2.7 and 2.8 (and maybe 2.6) to show that our tests pass.

suztomo:
Got it. Thank you for clarification. I'll check how Beam's Hadoop functionality is tested.

Memo: Still I feel I need to verify whether Hadoop-client 2.8 is compatible with Hadoop cluster 2.7.

- If it's compatible, then we can think about creating integration tests for multiple Hadoop versions (2.7 and 2.8).
- If Hadoop-client 2.8 is not compatible with Haodop cluster 2.7, then some modules need to be split in the same manner as Flink's.

kennknowles:
Another example which is an IO is elasticsearch, where we run tests against multiple versions.

Since there is a source incompatibility, you can split into modules (if necessary) in addition to running tests against multiple versions.

 - `beam-io-hadoop-format-lib` that works with all versions and just provides the main functionality

And thin modules:

 - `beam-io-hadoop-format-2.6`
 - `beam-io-hadoop-format-2.7`
 - `beam-io-hadoop-format-2.8` where you can add references to new 2.8 features

These might be able to be `build.gradle` only.

And you may not actually need them.

kennknowles:
I believe @lgajowy is a good contact about these sorts of tests.

lgajowy:
Yes - ping me and @mwalenia if you have any questions

iemejia:
Thanks Luke and the others for improving my explanation. One aspect however that differs in this case (compared with elasticsearch/flink) is that for both of them we 'include' the dependencies in the module. The hadoop modules (both hadoop-file-system and hadoop-format) are provided by the user (commonly by relying on the ones available in their clusters), so doing a test should be as easy as adding an extra test configuration in `hadoop-file-system` and `hadoop-format` with new parameters for the versions we want to test that we support. This is trivial to do in Maven `mvn test -Dhadoop.version=x.x.x` and I hope we could find a solution as simple as that one with gradle to not require the extra complexity of files and hierarchies that Elasticsearch and Flink have (but that’s a valid solution too).

Sorry @suztomo don’t assume my comments/questions are in the sense that I don’t want to have this upgrade done (it is totally the opposite), or that I want to ‘feature creep’ this PR, but compatibility with what users run in their clusters is important and I am afraid of this being broken in the future as a consequence of this version change.

suztomo:
Thanks, everyone. Give me some time to confirm compatibility.
(Now reading https://beam.apache.org/documentation/io/built-in/hadoop/ on how to use Hadoop from Beam)

suztomo:
> The hadoop modules (both hadoop-file-system and hadoop-format) are provided by the user (commonly by relying on the ones available in their clusters)

Now I understand this problem better. The Hadoop client dependencies are marked as `provided`. With this PR, users who touch "hdfs://" scheme will have to upgrade their Hadoop client version (especially `org.apache.hadoop:hadoop-hdfs-client`) at least; otherwise they'll get ClassNotFoundError or NoClassDefFoundError for the new class `HdfsClientConfigKeys`.

sdks/java/io/hadoop-file-system/src/main/java/org/apache/beam/sdk/io/hdfs/HadoopFileSystemRegistrar.java
```diff
@@ -45,6 +45,10 @@
 
   private static final List<String> HA_SCHEMES = Arrays.asList("hdfs", "webhdfs");
 
+  // Using hard-coded value for compatibility, because hadoop-dfs-client:2.7's DFSConfigKeys holds
+  // this value while it is HdfsClientConfigKeys that holds the value in version 2.8.
+  private static final String CONFIG_KEY_DFS_NAMESERVICES = "dfs.nameservices";
```
suztomo: Updated the source to use a string rather than a field of Hadoop class.

iemejia: Yes you read my mind, I was going to subject that just as the best path forward.

iemejia: I think this fix makes this mergeable, do you plan to tackle the multi version testing as well?

suztomo: @iemejia Thanks. Would you merge this?

I don't add the test in this PR. I created a followup issue: https://issues.apache.org/jira/browse/BEAM-8876

suztomo:
Run Java PostCommit

suztomo:
Thank you.

PR #10240, opened 2019-11-28
A fix for some TLS issues in the MongoDB IO

coheigea:
A fix for some TLS issues in the MongoDB IO


jbonofre:
retest this please

PR #10247, opened 2019-11-29
[BEAM-7274] In preparation for protocol-buffer schemas, add OneOf and Enumeration logical types

reuvenlax:
Add new LogicalTypes to represent enumerations and OneOf types. In order to implement this, we extended the LogicalType argument to be able to have a Schema so that we don't have to encode complicated arguments inside of strings.

R:@alexvanboxel


reuvenlax:
run sql postcommit

reuvenlax:
run java precommit

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/LogicalTypes.java
```diff
@@ -111,4 +134,247 @@ public FieldType getBaseType() {
...
+              .map(f -> Field.nullable(f.getName(), f.getType()))
+              .collect(Collectors.toList());
+      List<String> enumValues =
+          nullableFields.stream().map(Field::getName).collect(Collectors.toList());
+      nullableFields.add(
+          Field.of(enumField, FieldType.logicalType(EnumerationType.create(enumValues))));
+
+      oneOfSchema = Schema.builder().addFields(nullableFields).build();
+      schemaProtoRepresentation = SchemaTranslation.schemaToProto(oneOfSchema).toByteArray();
+      this.oneOfCaseFieldId = oneOfSchema.indexOf(enumField);
```
alexvanboxel: I'm wondering if the oneOfCase needs to be stored in the row as it can be inferred by the rest of the fields:

- the non-null value is the selected oneOf value
- if all fields are null the oneOf is NULL

this is how proto does it on the wire. Depends on what's important. Data size (extra field) or Case performance.

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/Schema.java
```diff
@@ -181,6 +183,14 @@ public Builder addBooleanField(String name) {
       return this;
     }
 
+    public Builder addEnumerationField(String name, EnumerationType enumerationType) {
```
alexvanboxel: Does it make sense to add these methods as first class add'ers to Schema? This feels like breaking consistency... as they are logical types.

sdks/java/core/src/main/java/org/apache/beam/sdk/values/Row.java
```diff
@@ -208,6 +210,26 @@ public Boolean getBoolean(String fieldName) {
     return getMap(getSchema().indexOf(fieldName));
   }
 
+  /** Helper method to get an enum field value. */
+  public EnumerationValue getEnumValue(String fieldValue) {
```
alexvanboxel: Same remark. Are they first class citizens to merit extra methods?

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/Schema.java
reuvenlax: Originally I left them out. But then decided that these are fairly "basic" logical types (e.g. they exist in proto, avro, etc.) so it was worth adding this method as syntactic sugar. I don't feel very strongly about it either way though.

sdks/java/core/src/main/java/org/apache/beam/sdk/values/Row.java
reuvenlax: See above (also keep in mind that we want to make DATETIME a logical type as well). However I don't feel strongly one way or the other about this.

reuvenlax:
run java precommit

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/LogicalTypes.java
```diff
@@ -111,4 +134,247 @@ public FieldType getBaseType() {
...
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hash(stringValue, value);
+    }
+  }
+
+  /** This {@link LogicalType} represent an enumeration over a fixed set of values. */
+  public static class EnumerationType implements LogicalType<EnumerationValue, Integer> {
```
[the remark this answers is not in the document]
reuvenlax: good idea - done

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/LogicalTypes.java
reuvenlax: This makes sense. Removed from the row representation

sdks/java/core/src/main/java/org/apache/beam/sdk/values/Row.java
alexvanboxel: Mmm.. yes. But we should keep it to the basic well known types then. Enum, DateTime they make sense. But the OneOf, is not so common.

reuvenlax: IMO unions are pretty basic. However removed these methods for now, as it's not difficult to simply access as a logical type.

reuvenlax:
Also added support for enums in the java type mapping. This means that POJOs or JavaBeans (or AutoValues, or any similar type) with Java enums will map to the new EnumerationType.

reuvenlax:

reuvenlax:
Note: fixing the Avro issue requires refactoring a bit. This refactor I already have in another PR as protos require it, so I'll go send that out now.

alexvanboxel:
> Note: fixing the Avro issue requires refactoring a bit. This refactor I already have in another PR as protos require it, so I'll go send that out now.

I was indeed wondering how this would Avro would be handled. For me it looks all good.

reuvenlax:
Added Enum support to Avro. Adding support for generic Avro unions proved complicated (mostly because Avro compiler represents them using an Object type), so best left for another PR; also, it appears that most usages of Avro unions are to represent nullable types, which we already handle.

reuvenlax:
@alexvanboxel I added Avro support for enums, and have now fixed CheckStyle errors. Let me know how this looks to you.

alexvanboxel:
OK, it took a while... but I've wanted to verify something. I've been experimenting with how Avro compiles to Java code and how the ordinal value is created. I just wanted to check out the following scenario:

` { "name": "foo", "type": { "name": "FooEnum", "type": "enum", "symbols": ["x","y"] }},`

we have a schema evolution to:

` { "name": "foo", "type": { "name": "FooEnum", "type": "enum", "symbols": ["a", "x", "y"] }},`

we **upgrade** the pipeline, so the current state is stored. As the ordinal value is stored in the intermediate state and the schema is evolved by **pre**-pending a enum value, that means the ordinal values changed (as it's sequential as defined in the source file: from `x=0 , y=1` > `a=0, x=1, y=2` ).

I do notice the same holds through to the Java Enum type. Not to **Proto** as it has an explicit number value assigned to an Enum string.

_I don't know if it's an issue... but I wanted to raise it so you're aware of the side effect_. Is a pipeline even upgradable when the schema changes?

reuvenlax:
Right now we don't support evolving schemas. There is a plan (that needs to be discussed on the dev list for this): but essentially it would require the new pipeline to preserve the old job's ordinal values as much as possible, instead of simply taking them in order.

alexvanboxel:
> Right now we don't support evolving schemas. There is a plan (that needs to be discussed on the dev list for this): but essentially it would require the new pipeline to preserve the old job's ordinal values as much as possible, instead of simply taking them in order.

OK, then it's not an issue for now.

reuvenlax:
Run Java PreCommit

reuvenlax:
Run Java PreCommit

reuvenlax:
Run Java PreCommit

reuvenlax:
Run Java PreCommit

reuvenlax:
Run Java PreCommit

PR #10254, opened 2019-12-02
[Beam-8564] Add LZO compression and decompression support

amoght:
LZO is a lossless data compression algorithm which is focused on compression and decompression speeds.

This will enable Apache Beam sdk to compress/decompress files using LZO/LZOP compression algorithm. 

This includes the following functionalities:
1) Appropriate Input and Output streams to enable working with LZO/LZOP files.
2) Compression using LZO/LZOP compression algorithm on Apache beam.
3) Decompression using LZO/LZOP decompression algorithm on Apache beam.


sdks/java/core/src/main/java/org/apache/beam/sdk/io/Compression.java
```diff
@@ -152,6 +156,38 @@ public WritableByteChannel writeCompressed(WritableByteChannel channel) throws I
     }
   },
 
+  /** LZO compression. */
+  LZO(".lzo", ".lzo") {
```
lukecwik: Should this use the `.lzo_deflate` extension for suggested and detected extensions?

sdks/java/core/src/main/java/org/apache/beam/sdk/io/Compression.java
```diff
@@ -152,6 +156,38 @@ public WritableByteChannel writeCompressed(WritableByteChannel channel) throws I
     }
   },
 
+  /** LZO compression. */
```
lukecwik: It may not be obvious to users that the difference between LZO and LZOP is that one has headers and one is just the LZO algorithm. Please expand on this comment and the one below.

lukecwik:
R: @lukecwik

sdks/java/core/build.gradle
```diff
@@ -90,4 +90,6 @@ dependencies {
   shadowTest library.java.avro_tests
   shadowTest library.java.zstd_jni
   testRuntimeOnly library.java.slf4j_jdk14
+  compile 'io.airlift:aircompressor:0.16'
```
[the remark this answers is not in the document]
amoght: The other implementations that are present are either not pure java(contain jni, .c, .h files) or have licensing issues.
This solves both of these issues since it is pure java implemented and is under Apache Licence.
I am not sure if we have any other pure java implementation which is under Apache Licence.

sdks/java/core/src/main/java/org/apache/beam/sdk/io/Compression.java
amoght: Agreed. I have added comments for both the modules (LZO, LZOP). This should make things clear.

sdks/java/core/src/main/java/org/apache/beam/sdk/io/Compression.java
amoght: Thanks for pointing this out. I have made the appropriate changes on this.

sdks/java/core/build.gradle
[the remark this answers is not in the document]
lukecwik: I asked on your original review thread if there were any alternative suggestions. LZO implementations that contain C code would be ok if they ran on the three most popular platforms (Linux, Mac, Windows).

sdks/java/core/src/main/java/org/apache/beam/sdk/util/LzoCompressorInputStream.java
```diff
@@ -0,0 +1,112 @@
...
+    count(ret == -1 ? 0 : 1);
+    return ret;
+  }
+
+  @Override
+  public int read(final byte[] buf, final int off, final int len) throws IOException {
+    if (len == 0) {
+      return 0;
+    }
+    final int ret = lzoIS.read(buf, off, len);
```
gsteelman: What is the reason for not delegating the call in the case of 0 length? Does `lzoIS.read()` not handle that case cleanly?

sdks/java/core/build.gradle
```diff
@@ -90,4 +90,6 @@ dependencies {
...
+  compile 'com.facebook.presto.hadoop:hadoop-apache2:3.2.0-1'
```
gsteelman: LZO itself should have no dependency on anything related to Hadoop, Presto, or Facebook.
1. Why do we need to include this?
2. If `aircompressor` really needs this, why does it need it?

sdks/java/core/src/test/java/org/apache/beam/sdk/io/CompressedSourceTest.java
```diff
@@ -161,6 +189,28 @@ public void testGzipSplittable() throws Exception {
     assertFalse(source.isSplittable());
   }
 
+  /** Test splittability of files in LZO mode -- none should be splittable. */
+  @Test
+  public void testLzoSplittable() throws Exception {
```
gsteelman: Please add a similar test for `testLzopSplittable()`.

sdks/java/core/src/test/java/org/apache/beam/sdk/io/CompressedSourceTest.java
```diff
@@ -235,6 +315,30 @@ public void testReadConcatenatedGzip() throws IOException {
     assertEquals(Bytes.asList(expected), actual);
   }
 
+  /**
+   * Using Lzo Codec Test a concatenation of lzo files is correctly decompressed.
+   *
+   * <p>A concatenation of lzo files as one file is a valid lzo file and should decompress to be the
+   * concatenation of those individual files.
```
gsteelman: I think we need an enclosing `</p>`, or we can simply remove the `<p>` tag. Either way, can you please ensure `testReadConcatenatedGzip()` and `testReadMultiStreamBzip2()` follow the same javadocs format?

sdks/java/core/src/test/java/org/apache/beam/sdk/io/CompressedSourceTest.java
```diff
@@ -235,6 +315,30 @@ public void testReadConcatenatedGzip() throws IOException {
...
+   */
+  @Test
+  public void testReadConcatenatedLzo() throws IOException {
```
gsteelman: Please add a `testReadConcatenatedLzop()` as well if it is applicable. Not sure if it is due to headers.

sdks/java/core/src/test/java/org/apache/beam/sdk/io/CompressedSourceTest.java
```diff
@@ -267,6 +371,69 @@ public void testReadMultiStreamBzip2() throws IOException {
     verifyReadContents(output, tmpFile, mode);
   }
 
+  /**
+   * Test a lzo file containing multiple streams is correctly decompressed.
+   *
+   * <p>A lzo file may contain multiple streams and should decompress as the concatenation of those
+   * streams.
```
gsteelman: I think we need an enclosing </p>, or we can simply remove the <p> tag. Whichever, please be consistent with  `testReadConcatenatedLzo()`.

sdks/java/core/src/test/java/org/apache/beam/sdk/io/CompressedSourceTest.java
```diff
@@ -761,6 +1043,132 @@ public void testGzipProgress() throws IOException {
...
+    writeFile(tmpFile, new byte[0], CompressionMode.LZO);
+
+    PipelineOptions options = PipelineOptionsFactory.create();
+    CompressedSource<Byte> source =
+        CompressedSource.from(new ByteSource(filename, 1)).withDecompression(CompressionMode.LZO);
+    try (BoundedReader<Byte> readerOrig = source.createReader(options)) {
+      assertThat(readerOrig, instanceOf(CompressedReader.class));
+      CompressedReader<Byte> reader = (CompressedReader<Byte>) readerOrig;
+      // before starting
+      assertEquals(0.0, reader.getFractionConsumed(), 1e-6);
```
gsteelman: I see `1e-6` is used by many of the test cases as a threshold. Should it be factored out into a constant?

sdks/java/core/src/main/java/org/apache/beam/sdk/util/LzoCompressorInputStream.java
amoght: No, this case is getting handled. This check has been put simply for the reason that if buffer length is 0, the read method doesn't even get executed and is handled here itself. Basically, to avoid unnecessary method call overhead.

sdks/java/core/build.gradle
amoght: This is included because LzoCodec class that has been used to create Input&Output streams is using some classes of the org.apache.hadoop package, which is a part of com.facebook.presto.hadoop.
Since the aircompressor is designed to also support optional hadoop configurations, hadoop is coming into picture(in our case, hadoop config is null).

sdks/java/core/src/test/java/org/apache/beam/sdk/io/CompressedSourceTest.java
amoght: This is happening when we run the spotlessApply task. When the </p> tag is clossed, the spotlessCheck fails. Not sure of the reason behind that.

sdks/java/core/src/test/java/org/apache/beam/sdk/io/CompressedSourceTest.java
amoght: The current behaviour of LZOP codec is that it returns the contents of the first file only, if concatenated files are given because of the presence of headers. This causes the test to fail. That is why we have not added this test.

sdks/java/core/src/test/java/org/apache/beam/sdk/io/CompressedSourceTest.java
amoght: This is happening during spotlessApply.

sdks/java/core/src/test/java/org/apache/beam/sdk/io/CompressedSourceTest.java
amoght: It can be done. But that would require altering all the tests that use this constant value. Will that be fine if we do that?

sdks/java/core/src/test/java/org/apache/beam/sdk/io/CompressedSourceTest.java
amoght: Thanks for pointing this out, this has been added.

amoght:
While studying the code, we found that the airlift/ aircompressor library only requires some classes which are also present in apache hadoop common package(~3.9MB). Therefore, we are now thinking that if we make changes in the airlift/ aircompressor package, replacing the 
com.facebook.presto.hadoop with org.apache.hadoop.common and remove other compression mechanisms(like zstd, gzip etc) while only keeping the required LZO package.
But if we go ahead with this approach, we will have to manually update this library whenever any changes are made to the airlift/aircompressor's LZO package.
@lukecwik @gsteelman please provide your thoughts on this.

gsteelman:
> While studying the code, we found that the airlift/ aircompressor library only requires some classes which are also present in apache hadoop common package(~3.9MB). Therefore, we are now thinking that of making changes in the airlift/ aircompressor package, replacing the
> com.facebook.presto.hadoop with org.apache.hadoop.common and removing other compression mechanisms present in the airlift/aircompressor package(like zstd, gzip etc) while only keeping the required LZO package.
> But if we go ahead with this approach, we will have to manually update this library whenever any changes are made to the airlift/aircompressor's LZO package.
> @lukecwik @gsteelman please provide your thoughts on this.

Is it possible to instead add the dependencies on the `apache.hadoop.common` package directly in these changes, and not add a dependency on airlift/aircompressor this change? I would prefer to stick with strict dependencies when possible, rather than relying on transitive dependencies to bring in the classes we need.

Relying on the transitive dependencies brought in by airlift/aircompressor has its own set of issues, including having to update our libraries whenever changes are made to airlift.

amoght:
@gsteelman we have used the airlift/aircompressor library to only get the compression and decompression mechanism, the implementation of Input/Output stream there introduces the transitive dependency, which can be removed and replaced with apache hadoop common library. This significantly reduces the size as well.
So, here are the 2 possible options:
1) We only use the compression and decompression mechanism from airlift/aircompressor and design the Input/Output Streams for beam accordingly. This will be needed to be updated if there is any change in those classes on airlift/aircompressor's end. But, since we will only be using the compression and decompression mechanism from airlift/aircompressor, the updates will be small and quite rare. Therefore, this won't be that big of an issue.
2) We introduce LZO as an optional package for beam. As this will give users the option to manage their beam size (if it is a constraint) or if LZO is not required.

gsteelman:
@amoght I don't have enough context to make the call on that, as I am very new to Beam. I have reached out to some others at Twitter to also review this change, as they will have more context.

amoght:
> @amoght I don't have enough context to make the call on that, as I am very new to Beam. I have reached out to some others at Twitter to also review this change, as they will have more context.

Thanks Gary :) appreciate your help!

sdks/java/core/src/test/java/org/apache/beam/sdk/io/CompressedSourceTest.java
gsteelman: Perhaps it would be a good idea to add a test with an expected failure then?

sdks/java/core/src/test/java/org/apache/beam/sdk/io/CompressedSourceTest.java
gsteelman: I think we can add the constant for `CompressedSourceTest.java` at least.

amoght: we have added this in the update

sdks/java/core/src/test/java/org/apache/beam/sdk/io/CompressedSourceTest.java
amoght: We have added this in this update.

amoght:
@lukecwik I've updated the PR based on the discussion that we had. Please let me know your thoughts and suggestions.

sdks/java/core/src/main/java/org/apache/beam/sdk/util/LzoCompressorInputStream.java
```diff
@@ -0,0 +1,112 @@
...
+import org.apache.commons.compress.utils.IOUtils;
+import org.apache.commons.compress.utils.InputStreamStatistics;
+
+/**
+ * {@link CompressorInputStream} implementation to create LZO encoded stream. Library relies on <a
+ * href="https://github.com/airlift/aircompressor/">LZO</a>
+ *
+ * @since 1.18
+ */
+public class LzoCompressorInputStream extends CompressorInputStream
```
lukecwik: Instead of creating a wrapper class that delegates to lzoIS, create a static method which returns the LZO and LZOP input and output streams and invoke the appropriate static method from the enum readDecompressed/writeDecompressed

Putting the code into a static method will prevent the LzoCodec/LzopCodec from being loaded till the static method is called.

sdks/java/core/src/test/java/org/apache/beam/sdk/io/CompressedSourceTest.java
lukecwik: The closing </p> tag isn't needed in javadoc even if your editor is inserting it.

sdks/java/core/src/test/java/org/apache/beam/sdk/io/CompressedSourceTest.java
lukecwik: Can we either add support for multistream or throw an exception if the stream isn't finished?

It would be dangerous for users to have part of their data silently dropped in this scenario. We should also add to the comment that concatenated streams aren't supported.

sdks/java/core/src/test/java/org/apache/beam/sdk/io/CompressedSourceTest.java
```diff
@@ -738,7 +1069,133 @@ public void testGzipProgress() throws IOException {
...
+          assertTrue(reader.start());
+        } else {
+          assertTrue(reader.advance());
+        }
+        assertEquals(0, reader.getSplitPointsConsumed());
+        assertEquals(1, reader.getSplitPointsRemaining());
+      }
+      assertFalse(reader.advance());
+
+      // after reading empty source
```
lukecwik:
```suggestion
      // after reading source
```

sdks/java/core/src/test/java/org/apache/beam/sdk/io/CompressedSourceTest.java
```diff
@@ -738,7 +1069,133 @@ public void testGzipProgress() throws IOException {
...
```
lukecwik:
```suggestion
      // after reading source
```

sdks/java/core/src/test/java/org/apache/beam/sdk/io/CompressedSourceTest.java
```diff
@@ -755,7 +1212,7 @@ public void testGzipProgress() throws IOException {
       assertFalse(reader.advance());
 
       // after reading empty source
```
lukecwik:
```suggestion
      // after reading source
```

sdks/java/core/src/main/java/org/apache/beam/sdk/util/LzoCompressorInputStream.java
amoght: Where do you suggest to keep static methods? Currently we have wrapper class in util package.

sdks/java/core/src/test/java/org/apache/beam/sdk/io/CompressedSourceTest.java
amoght: line 325: public void testReadConcatenatedLzo() throws IOException:
Is this for LZOP codec? Since unlike LZOP, LZO Codec supports file concatenation. Also we have testReadMultiStreamLzo and testFalseReadConcatenatedLzop

sdks/java/core/src/main/java/org/apache/beam/sdk/util/LzoCompressorInputStream.java
lukecwik: A class called LzoCompression in util is fine.

sdks/java/core/src/test/java/org/apache/beam/sdk/io/CompressedSourceTest.java
lukecwik: It would be better if the concatenated streams for LZOP worked or if concatenated streams were detected then an exception was thrown to the user.

lukecwik:
Ping, any updates?

nownikhil:
LGTM
- Twitter Core data libraries team

lukecwik:
I was pinging about whether there was any recent work to address my last review's comments?

gsteelman:
Hi @amoght can you address the open comments? Have reached out to a couple more folks internally at Twitter to request some more eyes on this. Hoping to get this wrapped up soon. Thank you for your work so far.

amoght:
@lukecwik we are working on all the suggestions provided by you, will be updating the PR in a few days. Thank you for your patience.

sdks/java/core/src/test/java/org/apache/beam/sdk/io/CompressedSourceTest.java
amoght: done

sdks/java/core/src/test/java/org/apache/beam/sdk/io/CompressedSourceTest.java
amoght: done

sdks/java/core/src/main/java/org/apache/beam/sdk/util/LzoCompressorInputStream.java
amoght: replaced wrapper classes with static methods

sdks/java/core/src/main/java/org/apache/beam/sdk/io/Compression.java
```diff
@@ -152,6 +156,38 @@ public WritableByteChannel writeCompressed(WritableByteChannel channel) throws I
...
+
+    @Override
+    public WritableByteChannel writeCompressed(WritableByteChannel channel) throws IOException {
+      return Channels.newChannel(new LzoCompressorOutputStream(Channels.newOutputStream(channel)));
+    }
+  },
+
+  /**
+   * LZOP compression using LZOP Codec. .lzo extension is specified for the files with magic bytes
+   * and headers.
```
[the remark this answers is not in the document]
amoght: done

sdks/java/core/src/main/java/org/apache/beam/sdk/io/Compression.java
```diff
@@ -152,6 +156,38 @@ public WritableByteChannel writeCompressed(WritableByteChannel channel) throws I
     }
   },
 
+  /**
+   * LZO compression using LZO Codec. .lzo_deflate extension is specified for the files which just
+   * use the LZO algorithm without headers.
```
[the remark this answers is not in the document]
amoght: done

sdks/java/core/build.gradle
```diff
@@ -91,4 +95,6 @@ dependencies {
   shadowTest library.java.avro_tests
   shadowTest library.java.zstd_jni
   testRuntimeOnly library.java.slf4j_jdk14
+  compileOnly 'io.airlift:aircompressor:0.16'
```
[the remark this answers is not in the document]
amoght: done

sdks/java/core/build.gradle
```diff
@@ -91,4 +95,6 @@ dependencies {
   shadowTest library.java.avro_tests
   shadowTest library.java.zstd_jni
   testRuntimeOnly library.java.slf4j_jdk14
+  compileOnly 'io.airlift:aircompressor:0.16' 
+  compileOnly 'com.facebook.presto.hadoop:hadoop-apache2:3.2.0-1'
```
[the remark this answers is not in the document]
amoght: done

sdks/java/core/build.gradle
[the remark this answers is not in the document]
amoght: done

sdks/java/core/build.gradle
```diff
@@ -58,6 +58,10 @@ test {
   }
 }
 
+configurations {
+    testCompile.extendsFrom compileOnly
+}
+
```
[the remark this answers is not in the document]
amoght: done

sdks/java/core/src/test/java/org/apache/beam/sdk/io/CompressedSourceTest.java
amoght: For now we have added a comment warning users that a concatenated lzo file doesn't gets decompressed correctly. Its added above testFalseReadConcatenatedLzop and testFalseReadMultiStreamLzop methods.

amoght:
@lukecwik I've incorporated mostly all the suggested changes in the PR. Please let me know your thoughts on this.

lukecwik:
retest this please

sdks/java/core/src/test/java/org/apache/beam/sdk/io/CompressedSourceTest.java
lukecwik: Best to add that comment over the LZOP enum since nobody reading the documentation is going to find the comment in the tests.

sdks/java/core/src/main/java/org/apache/beam/sdk/io/Compression.java
```diff
@@ -152,6 +153,54 @@ public WritableByteChannel writeCompressed(WritableByteChannel channel) throws I
     }
   },
 
+  /**
+   * LZO compression using LZO Codec. .lzo_deflate extension is specified for the files which just
```
lukecwik:
```suggestion
   * LZO compression using LZO codec. {@code .lzo_deflate} extension is specified for files which
```

sdks/java/core/src/main/java/org/apache/beam/sdk/io/Compression.java
```diff
@@ -152,6 +153,54 @@ public WritableByteChannel writeCompressed(WritableByteChannel channel) throws I
...
+   * use the LZO algorithm without headers.
+   *
+   * <p>The Beam Java SDK does not pull in the required libraries for LZO compression by default, so
+   * it is the user's responsibility to declare an explicit dependency on {@code
+   * airlift/aircompressor} and {@code presto-hadoop-apache2}. Attempts to read or write
```
lukecwik:
```suggestion
   * io.airlift:aircompressor} and {@code com.facebook.presto.hadoop:hadoop-apache2}. Attempts to read or write
```

sdks/java/core/src/main/java/org/apache/beam/sdk/io/Compression.java
```diff
@@ -152,6 +153,54 @@ public WritableByteChannel writeCompressed(WritableByteChannel channel) throws I
...
+   * .lzo_deflate files without {@code airlift/aircompressor} and {@code presto-hadoop-apache2}
```
lukecwik:
```suggestion
   * {@code .lzo_deflate} files without {@code io.airlift:aircompressor} and {@code com.facebook.presto.hadoop:hadoop-apache2}
```

sdks/java/core/src/main/java/org/apache/beam/sdk/io/Compression.java
```diff
@@ -152,6 +153,54 @@ public WritableByteChannel writeCompressed(WritableByteChannel channel) throws I
...
+   * loaded will result in {@code NoClassDefFoundError} at runtime.
```
lukecwik:
```suggestion
   * loaded will result in a {@code NoClassDefFoundError} at runtime.
```

sdks/java/core/src/main/java/org/apache/beam/sdk/io/Compression.java
```diff
@@ -152,6 +153,54 @@ public WritableByteChannel writeCompressed(WritableByteChannel channel) throws I
...
+
+    @Override
+    public WritableByteChannel writeCompressed(WritableByteChannel channel) throws IOException {
+      return Channels.newChannel(
+          LzoCompression.createLzoOutputStream(Channels.newOutputStream(channel)));
+    }
+  },
+
+  /**
+   * LZOP compression using LZOP Codec. .lzo extension is specified for the files with magic bytes
```
lukecwik:
```suggestion
   * LZOP compression using LZOP codec. {@code .lzo} extension is specified for files with magic bytes
```

sdks/java/core/src/main/java/org/apache/beam/sdk/io/Compression.java
```diff
@@ -152,6 +153,54 @@ public WritableByteChannel writeCompressed(WritableByteChannel channel) throws I
...
+   * and headers.
+   *
+   * <p>The Beam Java SDK does not pull in the required libraries for LZOP compression by default,
+   * so it is the user's responsibility to declare an explicit dependency on {@code
+   * airlift/aircompressor} and {@code presto-hadoop-apache2}. Attempts to read or write .lzo files
```
lukecwik:
```suggestion
   * io.airlift:aircompressor} and {@code com.facebook.presto.hadoop:hadoop-apache2}. Attempts to read or write {@code .lzo} files
```

sdks/java/core/src/main/java/org/apache/beam/sdk/io/Compression.java
```diff
@@ -152,6 +153,54 @@ public WritableByteChannel writeCompressed(WritableByteChannel channel) throws I
...
+   * without {@code airlift/aircompressor} and {@code presto-hadoop-apache2} loaded will result in
```
lukecwik:
```suggestion
   * without {@code io.airlift:aircompressor} and {@code com.facebook.presto.hadoop:hadoop-apache2} loaded will result in a
```

sdks/java/core/src/main/java/org/apache/beam/sdk/io/Compression.java
```diff
@@ -152,6 +153,54 @@ public WritableByteChannel writeCompressed(WritableByteChannel channel) throws I
...
+    public WritableByteChannel writeCompressed(WritableByteChannel channel) throws IOException {
+      return Channels.newChannel(
+          LzoCompression.createLzoOutputStream(Channels.newOutputStream(channel)));
+    }
+  },
+
+  /**
+   * LZOP compression using LZOP Codec. .lzo extension is specified for the files with magic bytes
+   * and headers.
+   *
```
lukecwik:
```suggestion
   *
   * <p><b>Warning:</b> The LZOP codec being used does not support concatenated LZOP streams and will
   * silently ignore data after the end of the first LZOP stream.
   *
```

lukecwik:
After committing the comments, you may need to run spotlessApply again.

amoght:
> After committing the comments, you may need to run spotlessApply again.

Done

lukecwik:
retest this please

lukecwik:
retest this please

lukecwik:

lukecwik:

shubham-srivastav:

shubham-srivastav:
`Run Java_Examples_Dataflow PreCommit`

shubham-srivastav:
retest this please

lukecwik:
Run Java PreCommit

lukecwik:

lukecwik:

shubham-srivastav:
@lukecwik  Do we need to add test dependency for facebook-presto and airlift in  /beam/examples/java/build.gradle

lukecwik:
WordCount doesn't depend on using LZO so it shouldn't be a dependency and the pipeline should execute successfully without it. The test may be picking up a legitimate case which users would hit as well.

shubham-srivastav:
@lukecwik We Observed replacing Compression I/O stream with java.io I/O stream in LzoCompression.java can resolve the issue. Should we go ahead and do that?

lukecwik:
That sounds great. Should have caught that earlier.

amoght:
retest this please

amoght:
Run Java PreCommit

shubham-srivastav:
retest this please

shubham-srivastav:
Run Java PreCommit

lukecwik:
retest this please

sdks/java/core/src/main/java/org/apache/beam/sdk/util/LzoCompression.java
```diff
@@ -22,58 +22,52 @@
...
 import java.io.OutputStream;
-import org.apache.hadoop.io.compress.CompressionInputStream;
-import org.apache.hadoop.io.compress.CompressionOutputStream;
 
 public class LzoCompression {
 
   /**
-   * Create a {@link CompressionInputStream} that will read from the given {@link InputStream} using
-   * {@link LzoCodec}.
+   * Create a {@link InputStream} that will read from the given {@link InputStream} using {@link
```
lukecwik:
```suggestion
   * Create an {@link InputStream} that will read from the given {@link InputStream} using {@link
```

sdks/java/core/src/main/java/org/apache/beam/sdk/util/LzoCompression.java
```diff
@@ -22,58 +22,52 @@
...
-  public static CompressionInputStream createLzoInputStream(InputStream inputStream)
-      throws IOException {
+  public static InputStream createLzoInputStream(InputStream inputStream) throws IOException {
     return new LzoCodec().createInputStream(inputStream);
   }
 
   /**
-   * Create a {@link CompressionInputStream} that will read from the given {@link InputStream} using
-   * {@link LzopCodec}.
+   * Create a {@link InputStream} that will read from the given {@link InputStream} using {@link
```
lukecwik:
```suggestion
   * Create an {@link InputStream} that will read from the given {@link InputStream} using {@link
```

sdks/java/core/src/main/java/org/apache/beam/sdk/util/LzoCompression.java
```diff
@@ -22,58 +22,52 @@
...
    */
-  public static CompressionInputStream createLzopInputStream(InputStream inputStream)
-      throws IOException {
+  public static InputStream createLzopInputStream(InputStream inputStream) throws IOException {
     return new LzopCodec().createInputStream(inputStream);
   }
 
   /**
-   * Create a {@link CompressionOutputStream} that will write to the given {@link OutputStream}.
+   * Create a {@link OutputStream} that will write to the given {@link OutputStream}.
```
lukecwik:
```suggestion
   * Create an {@link OutputStream} that will write to the given {@link OutputStream}.
```

sdks/java/core/src/test/java/org/apache/beam/sdk/io/CompressedSourceTest.java
```diff
@@ -83,7 +82,7 @@
 @RunWith(JUnit4.class)
 public class CompressedSourceTest {
 
-  private final double DELTA = 1e-6;
+  private final double delta = 1e-6;
```
lukecwik: nit: you should have declared this static and kept the capital letters instead of making it a member variable of CompressedSourceTest

PR #10276, opened 2019-12-03
[BEAM-7926] Data-centric Interactive Part1

KevinGG:
1. Changed Beam Python SDK’s pipeline transform apply() logic to
implicitly alter the label of top-level transform to be applied with
ipython-prompt metadata when the code execution is in ipython and during
user-defined pipeline construction time.
2. The users can observe what pieces of code are in-effect by looking at
the labels of applied PTransforms. Thus they can avoid broken in-memory
states due to notebook cell-re-execution/cell-deletion/hidden-states.
3. Allow duplicate PTransform labels in different cells when the user
constructs a pipeline under an ipython/notebook environment because the
label altering de-duplicates the labels.
4. Created common testing package and modules: Created a mock ipython
kernel for testing; Made pipeline assertion a common testing module.
5. Updated tests for ipython related scenarios in a mocked ipython
environment.
6. Updated the example notebook.
------------------------

--- | --- | --- | --- | ---


KevinGG:
The main logic lies in `sdks/python/apache_beam/utils/interactive_utils.py` and `sdks/python/apache_beam/pipeline.py`.
Everything else is either tests/test-utils around the changes or minor fixes.

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
```diff
@@ -314,7 +307,6 @@ def cache_key(self, pcoll):
     if cacheable:
       return '_'.join((cacheable['var'],
                        cacheable['version'],
-                       cacheable['pcoll_id'],
```
KevinGG: `pcoll_id` is not stable for cache_key especially with the coming introduction of `pipeline fragment` in `[BEAM-7926-part2]`. A combination of other fields is sufficient to locate a cache.

KevinGG:
R: @davidyan74 
R: @rohdesamuel 
PTAL, thanks!

R: @aaltay 
Adding Ahmet as the committer.

Thank you all!

sdks/python/apache_beam/runners/interactive/interactive_environment.py
```diff
@@ -221,3 +224,20 @@ def is_terminated(self, pipeline):
     if result:
       return runner.PipelineState.is_terminal(result.state)
     return True
+
+  def track_user_pipelines(self):
+    """Tracks all user-defined pipeline instances.
```
davidyan74: Please explain what it means by "tracking" user pipelines.

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
davidyan74: Thanks! Perhaps put that explanation in the comment?

sdks/python/apache_beam/runners/interactive/testing/pipeline_assertion.py
```diff
@@ -0,0 +1,64 @@
...
+  expected_pipeline_proto = expected_pipeline.to_runner_api(
+      use_fake_coders=True)
+  components1 = actual_pipeline_proto.components
+  components2 = expected_pipeline_proto.components
+  test_case.assertEqual(len(components1.transforms),
+                        len(components2.transforms))
+  test_case.assertEqual(len(components1.pcollections),
+                        len(components2.pcollections))
+
+  # GreatEqual instead of Equal because the pipeline_proto_to_execute could
```
davidyan74: This would mean that the assert_pipeline_equal(a, b) does not mean assert_pipeline_equal(b, a).
We should explain this in detail in the docstring of this function. Also, GreatEqual should be GreaterEqual.

sdks/python/apache_beam/utils/interactive_utils.py
```diff
@@ -0,0 +1,98 @@
...
+    # Tracks user defined pipeline instances in watched scopes so that we only
+    # alter labels for any transform to pvalueish belonging to those pipeline
+    # instances, excluding any transform to be applied in other pipeline
+    # instances the Beam SDK creates implicitly.
+    ie.current_env().track_user_pipelines()
+    from IPython import get_ipython
+    prompt = get_ipython().execution_count
+    pipeline = _extract_pipeline_of_pvalueish(pvalueish)
+    if not pipeline:
+      _LOGGER.warning('Non-fatal error. Failed to alter the label of a '
```
davidyan74: nit: warning already implies non-fatal.

sdks/python/apache_beam/runners/interactive/display/pipeline_graph_test.py
```diff
@@ -83,6 +91,45 @@ def test_get_dot(self):
...
+         '"Cell 3: Square";\n'
+         'squares [shape=circle];\n'
+         '"Cell 4: Cube";\n'
+         'cubes [shape=circle];\n'
+         '"Cell 2: Init" -> init_pcoll;\n'
+         'init_pcoll -> "Cell 3: Square";\n'
+         'init_pcoll -> "Cell 4: Cube";\n'
+         '"Cell 3: Square" -> squares;\n'
+         '"Cell 4: Cube" -> cubes;\n'
+         '}\n'),
```
davidyan74: Should we add a test for composite transforms?

KevinGG: `Create` is a composite transform. The dot based pipeline graph only renders top-level transform and our label altering only happens to top-level transforms. So whether a transform is composite or not doesn't affect neither.

sdks/python/apache_beam/runners/interactive/interactive_environment.py
KevinGG: Rephrased to
```
    """Record references to all user-defined pipeline instances watched in
    current environment.

    Current static global singleton interactive environment holds references to
    a set of pipeline instances defined by the user in the watched scope.
    Interactive Beam features could use the references to determine if a given
    pipeline is defined by user or implicitly created by Beam SDK or runners,
    then handle them differently.

    This is invoked every time a PTransform is to be applied if the current
    code execution is under ipython due to the possibility that any user-defined
    pipeline can be re-evaluated through notebook cell re-execution at any time.

    Each time this is invoked, the tracked user pipelines are refreshed to
    remove any pipeline instances that are no longer in watched scope. For
    example, after a notebook cell re-execution re-evaluating a pipeline
    creation, the last pipeline reference created by last evaluation will not be
    in watched scope anymore.
    """
```

sdks/python/apache_beam/runners/interactive/pipeline_instrument.py
KevinGG: Adding
```
    'pcoll_id' of cacheable is not stable for cache_key, thus not included in
    cache key. A combination of 'var', 'version' and 'producer_version' is
    sufficient to identify a cached PCollection.
```

sdks/python/apache_beam/utils/interactive_utils.py
KevinGG: Removing the non-fatal sentence.

sdks/python/apache_beam/runners/interactive/testing/pipeline_assertion.py
KevinGG: `assert_pipeline_equal(a, b)` should be symmetric.
The assertion code was copied from old pipeline_analyzer tests. We don't need the GreaterEqual any more. Changing the assertion statement and removing the comment.

sdks/python/apache_beam/pipeline.py
```diff
@@ -491,6 +492,11 @@ def apply(self, transform, pvalueish=None, label=None):
       finally:
         transform.label = old_label
 
+    # Attempts to alter the label of the transform to be applied only when it's
+    # a top-level transform.
```
rohdesamuel: Please append: "This is so that the cell number will not be prepended to every child transform in a composite."

KevinGG: Thanks, yes, adding it.

KevinGG:
retest this please.

sdks/python/apache_beam/runners/interactive/testing/mock_ipython.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def __enter__(self):
+      """Marks entering of a cell/prompt."""
+      self._execution_count = self._execution_count + 1
+
+    def __exit__(self, exc_type, exc_value, traceback):
+      """Marks exiting of a cell/prompt."""
+      pass
+
+  global _mocked_get_ipython
```
aaltay: Do you need both L19 and L65 definitions for global?

sdks/python/apache_beam/runners/interactive/testing/pipeline_assertion.py
```diff
@@ -0,0 +1,62 @@
...
+def _assert_transform_equal(test_case, actual_pipeline_proto,
+                            actual_transform_id,
+                            expected_pipeline_proto, expected_transform_id):
+  """Asserts the equivalence between transforms from two given pipelines. """
+  transform_proto1 = actual_pipeline_proto.components.transforms[
+      actual_transform_id]
+  transform_proto2 = expected_pipeline_proto.components.transforms[
+      expected_transform_id]
+  test_case.assertEqual(transform_proto1.spec.urn, transform_proto2.spec.urn)
+  # Skipping payload checking because PTransforms of the same functionality
```
aaltay: Would not this find 2 ParDo's with different user functions equivalent? Is that intended?

sdks/python/apache_beam/utils/interactive_utils.py
```diff
@@ -0,0 +1,98 @@
...
+  is_in_notebook = False
+  if is_in_ipython():
+    # The import and usage must be valid under the execution path.
+    from IPython import get_ipython
+    if 'IPKernelApp' in get_ipython().config:
+      is_in_notebook = True
+  return is_in_notebook
+
+
+def alter_label_if_interactive(transform, pvalueish):
```
aaltay: This is more like alter_label_if_ipython. Do we want/need it to be more generic? Like supporting users running in a python shell in an interactive fashion?

sdks/python/apache_beam/utils/interactive_utils.py
```diff
@@ -0,0 +1,98 @@
...
+                      % pvalueish)
+    if (pipeline
+        # We only alter for transforms to be applied to user-defined pipelines
+        # at pipeline construction time.
+        and pipeline in ie.current_env().tracked_user_pipelines):
+      transform.label = 'Cell {}: {}'.format(prompt, transform.label)
+
+
+def _extract_pipeline_of_pvalueish(pvalueish):
+  """Extracts the pipeline that the given pvalueish belongs to."""
```
aaltay: Extracts the pipeline or a more concrete pvalue type?

sdks/python/apache_beam/runners/interactive/testing/mock_ipython.py
KevinGG: The `global` on L65 is not a definition but telling the interpreter this variable is a global variable (that is defined on L19). Otherwise, this function treats `_mocked_get_ipython` as an undefined local variable because these lines are in the local scope of function `mock_get_ipython()` not global scope.

sdks/python/apache_beam/runners/interactive/testing/pipeline_assertion.py
KevinGG: Probably if they have the same label.
This functionality was originally part of the `pipeline_analyzer_test`. And the module is used to verify if the implicit cache transforms are applied as expected. So there would not be user code involved in any test that uses this assertion.
I'll add a module level docstring to clarify this utility's usage:
```
"""Module to verify implicit cache transforms applied by Interactive Beam.

For internal use only; no backwards-compatibility guarantees.
This utility should only be used by Interactive Beam tests. It verifies if the
implicit cache transforms are applied as expected when running a pipeline with
the InteractiveRunner. It shouldn't be used to verify equivalence between
pipelines if the code to be tested depends or mutates user code in pipelines.
"""
```

sdks/python/apache_beam/utils/interactive_utils.py
KevinGG: I think making it more generic doesn't hurt since we are calling the thing `Interactive Beam` rather than `ipython Beam` though nothing would work without ipython.

We do support using `Interactive Beam` in ipython shell. It will have fewer problems when users are not using a notebook because there will be no "hidden-state"/"out-of-order-execution"/"deleted-cell". And we support a sampled data-table print when `show` is invoked. Duplicate transform labels in different prompts are allowed in ipython shell. And so on...

sdks/python/apache_beam/utils/interactive_utils.py
KevinGG: Extracts the pipeline.

When constructing a pipeline, it's unlikely to run into the first few conditional paths.
When running a pipeline, under the hood, Beam SDK and runners do a lot of magic including implicitly building the real pipelines for execution. I see pvalueish being tuples and dicts from time to time.
So the first step is to get a concrete pvalue from the pvalueish. Any pvalue would be fine.
Then fetch the pipeline the pvalue belongs to.

aaltay: Even if it is unlikely, this might return a pvalue. If that is not expected, maybe we can make those cases errors.

sdks/python/apache_beam/utils/interactive_utils.py
aaltay: Should we make the code more generic? Unless this runs in ipython environment it will not do anything even if the environment is interactive. We can either change the code or the name to make them match each other.

KevinGG: Changing it to `alter_label_if_ipython` to be more descriptive.

sdks/python/apache_beam/utils/interactive_utils.py
KevinGG: Discussed offline, mark it resolved.

KevinGG:
I'll do a rebase to upstream HEAD to resolve the merge conflict.

KevinGG:
Run Python PreCommit

KevinGG:
Run Python PreCommit

KevinGG:
R: @aaltay ,
Hi Ahmet, the merge conflicts have been resolved and the PR is ready to be merged.
Thanks a lot!

PR #10278, opened 2019-12-03
[BEAM-7274] Support recursive type transformation in ByteBuddyUtils

reuvenlax:
This was uncovered by the attempt to add protocol buffer support to schemas. This PR refactors the schema utils to convert parameter types (e.g. for List<T> to convert T as well). As part of this PR, Row now represents ARRAY types as a Collection instead of as a List.

R: @alexvanboxel

### DIFF
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/utils/ReflectUtils.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/utils/ReflectUtils.java
@@ -84,7 +84,9 @@ public class ReflectUtils {
         clazz,
         c -> {
           return Arrays.stream(c.getDeclaredMethods())
-              .filter(m -> !m.isBridge()) // Covariant overloads insert bridge functions, which we must ignore.
+              .filter(
+                  m -> !m.isBridge()) // Covariant overloads insert bridge functions, which we must
+              // ignore.
               .filter(m -> !Modifier.isPrivate(m.getModifiers()))
               .filter(m -> !Modifier.isProtected(m.getModifiers()))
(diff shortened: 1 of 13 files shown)


sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/utils/ReflectUtils.java
```diff
@@ -33,16 +36,19 @@
...
+import org.apache.beam.sdk.values.TypeDescriptor;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.Lists;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.Maps;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.primitives.Primitives;
 
 /** A set of reflection helper methods. */
 public class ReflectUtils {
-  static class ClassWithSchema {
+  /** Represents a class and a schema. */
+  public static class ClassWithSchema {
```
alexvanboxel: I assume it has a reason that they are now public? (was package protected)

alexvanboxel:
In general this LGTM. I see optimizations for the List vs Collection as List will be the most common use case. If I remember correctly the instanceof has minimal impact.

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/utils/ReflectUtils.java
reuvenlax: It's actually public because we need to use this in the protobuf schema provider which is in a different package (and in general, you should be able to put schema providers in other packages).

reuvenlax:
run sql postcommit

PR #10333, opened 2019-12-09
[BEAM-8536] Migrate request_execution_time to request_delay_time in Java

boyuanzz:
+R: @robertwb 
cc: @lukecwik


boyuanzz:
Run Portable_Python PreCommit

boyuanzz:
Run Java PreCommit

boyuanzz:
Run Java PreCommit

boyuanzz:
Kindly pinging : )

model/fn-execution/src/main/proto/beam_fn_api.proto
```diff
@@ -207,18 +207,13 @@ message BundleApplication {
 // Either an absolute timestamp or a relative timestamp can represent a
 // scheduled execution time.
 message DelayedBundleApplication {
-  // Recommended time at which the application should be scheduled to execute
-  // by the runner. Times in the past may be scheduled to execute immediately.
-  // TODO(BEAM-8536): Migrate usage of absolute time to requested_time_delay.
-  google.protobuf.Timestamp requested_execution_time = 1;
```
robertwb: Are there other places that need to be changed (e.g. does this need to be a two-phased commit)?

boyuanzz: I have changed all the usage in this PR. Just talked to Rebo and Daniel, one left here is to generate go proto within this PR. I'll let you know once I finish.

boyuanzz: I have rebuilt the go proto. Waiting for tests pass.

boyuanzz:
Run Java PreCommit

lukecwik:
Run Python PreCommit

model/fn-execution/src/main/proto/beam_fn_api.proto
```diff
@@ -207,18 +207,13 @@ message BundleApplication {
...
-
   // (Required) The application that should be scheduled.
-  BundleApplication application = 2;
+  BundleApplication application = 1;
```
apilloud: Field indexes in protos should never be changed.

boyuanzz: Thanks for your comment! I think we should be fine here given:
1.  The usage of `DelayedBundleApplication` is still under development, we never announced an official supported usage so far IIRC.
2. We release versioned SDK, which should also work with versioned runner. I don't think it will break anything.
Would you mind explaining a little bit more about your concern?

PR #10353, opened 2019-12-11
[release-2.18.0][BEAM-8861] Disallow self-signed certificates by default in ElasticsearchIO

iemejia:
Same that the one I accidentally push forced to cherry pick another issue.
R: @udim 
CC: @coheigea


iemejia:
For info we had to backport this into 2.17.0 too and it was already merged in #10354

PR #10362, opened 2019-12-12
[BEAM-8955] Attach values so we don't iterate over everything to verify types

reuvenlax:
We were accidentally overiterating the result of a Group. This is problematic for the Spark runner and inefficient in general.


reuvenlax:
R: @suztomo

suztomo:
Run Spark ValidatesRunner

PR #10363, opened 2019-12-12
[GoSDK] Improve StateChannel resilience.

lostluck:
Similarly to https://github.com/apache/beam/pull/10238, have the state channels properly terminate & clean themselves up on errors, and allow themselves to be recreated by the manager. Bundles using a bad StateChannel will fail one way or another (their readers will return errors up to them), but subsequent creations will no longer be poisoned. Runners can then retry the bundles at their discretion.

This affects anything reading (and arguably, writing) on the State service, which includes Side Inputs, State Backed Iterables for CoGBKs, and when the Go SDK eventually has State API support.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/go/pkg/beam/core/runtime/harness/statemgr.go
+++ b/sdks/go/pkg/beam/core/runtime/harness/statemgr.go
@@ -347,16 +347,17 @@ func (c *StateChannel) read(ctx context.Context) {
 func (c *StateChannel) write(ctx context.Context) {
 	var err error
 	var id string
-	for req := range c.requests {
+	for {
+		var req *pb.StateRequest
+		select {
+		case req = <-c.requests:
+		case <-c.DoneCh: // Close the goroutine on context cancel.
+			return
+		}
 		err = c.client.Send(req)
 		if err != nil {
 			id = req.Id
 			break // non-nil errors mean the stream is broken and can't be re-used.
-		}
-		select {
-		case <-c.DoneCh: // Close the goroutine on context cancel.
-			return
-		default: // Continue around the loop here.
 		}
 	}
 
--- a/sdks/go/pkg/beam/core/runtime/harness/statemgr.go
+++ b/sdks/go/pkg/beam/core/runtime/harness/statemgr.go
@@ -361,10 +361,7 @@ func (c *StateChannel) write(ctx context.Context) {
 		}
 	}
 
-	if err == nil {
-		log.Warnf(ctx, "StateChannel[%v].write closed", c.id)
-		return
-	} else if err == io.EOF {
+	if err == io.EOF {
 		log.Warnf(ctx, "StateChannel[%v].write EOF on send; fetching real error", c.id)
 		err = nil
 		for err == nil {
--- a/sdks/go/pkg/beam/core/runtime/harness/statemgr.go
+++ b/sdks/go/pkg/beam/core/runtime/harness/statemgr.go
@@ -312,6 +312,7 @@ func makeStateChannel(ctx context.Context, cancelFn context.CancelFunc, id strin
 
 func (c *StateChannel) read(ctx context.Context) {
 	for {
+		// Closing the context will have an error return from this call.
 		msg, err := c.client.Recv()
 		if err != nil {
 			c.terminateStreamOnError(err)
@@ -351,6 +352,11 @@ func (c *StateChannel) write(ctx context.Context) {
 		if err != nil {
 			id = req.Id
 			break // non-nil errors mean the stream is broken and can't be re-used.
+		}
+		select {
+		case <-c.DoneCh: // Close the goroutine on context cancel.
+			return
+		default: // Continue around the loop here.
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


lostluck:
R: @youngoli 
If it looks good to you, please feel free to merge once the tests are passing.

lostluck:
Run Go PostCommit

lostluck:
Run Go Postcommit

lostluck:
Run Go Postcommit

sdks/go/pkg/beam/core/runtime/harness/statemgr_test.go
```diff
@@ -0,0 +1,248 @@
...
+			}
+
+			if test.validateCancelled {
+				select {
+				case <-ctx.Done(): // Assert that the context must have been cancelled on read failures.
+					return
+				case <-time.After(time.Second * 5):
+					t.Fatal("context wasn't cancelled")
+				}
+				if !forceRecreateCalled {
```
youngoli: It looks like this check would never get reached because the select statement above either returns or calls a fatal error. Is the return supposed to be a break?

sdks/go/pkg/beam/core/runtime/harness/statemgr_test.go
```diff
@@ -0,0 +1,248 @@
...
+						Id: req.Id,
+					}
+				}()
+				_, err := c.Send(&pb.StateRequest{})
+				return err
+			},
+			expectedErr:       expectedError,
+			validateCancelled: true,
+		}, {
+			name: "readResponseChannelDeleted",
```
youngoli: This test case gave me some trouble to understand, compared to the others. Is this one meant to check the case in StateChannel.read where there is no consumer for the state response? (Line 332 in statemgr.go)

sdks/go/pkg/beam/core/runtime/harness/statemgr.go
```diff
@@ -307,21 +345,47 @@ func (c *StateChannel) read(ctx context.Context) {
...
 func (c *StateChannel) write(ctx context.Context) {
-	for req := range c.requests {
-		err := c.client.Send(req)
-		if err == nil {
-			continue // ok
+	var err error
+	var id string
+	for {
+		var req *pb.StateRequest
+		select {
```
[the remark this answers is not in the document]
lostluck: Good question.

Channel Select picks "ready" cases arbitrarily. Otherwise it's blocking until the first thing triggers.
Assuming both are ready, that means the client has experienced an error and the context has been cancelled. GRPC clients would send an error in this case. Unfortunately, since this depends on GRPC behavior, it's hard to test and verify.
The only way the context is cancelled is via the terminate method, which means requests and responses are nilled. This happens in a locked state, so that operation is safe. There's arguably a race on nilling c.requests, but lets worry about that in a minute.
 
However, assuming the client returns an error, then the go-routine would close, and then get a subsequent error from the Recv calls (which will always repeat their errors for sure), try to get the  responses back from the nil map (it's ok to "comma ok" look up from a nil map), and then not send anything.

If the context cancellation resolves first, then the goroutine exits as expected.

For the contentious case: If the write loop is waiting on the channel, I don't think there are issues, since it's already got a reference to the request channel.  If it happen before or after the nilling of requests, it's fine. If it happens as a race that's more problematic, but that process was going to have problems anyway. The worker would panic, and then would get restarted, which is fine.



1. waiting on a nil channel is fine, and doesn't cause problems (other than the goroutine deadlocking) https://play.golang.org/p/7NQh8StcNh-
It will block forever, which is what we want. When the context is canceled the done channel is closed and unblocks the code as intended. 
2.

sdks/go/pkg/beam/core/runtime/harness/statemgr.go
```diff
@@ -307,21 +345,47 @@ func (c *StateChannel) read(ctx context.Context) {
...
+			break // non-nil errors mean the stream is broken and can't be re-used.
+		}
+	}
 
-		// Failed to send. Return error.
-		c.mu.Lock()
-		ch, ok := c.responses[req.Id]
-		delete(c.responses, req.Id)
-		c.mu.Unlock()
+	if err == nil {
```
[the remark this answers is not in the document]
lostluck: You're right. I had this when the loop was still ranging over the c.requests channel, and it's no longer necessary.

sdks/go/pkg/beam/core/runtime/harness/statemgr_test.go
lostluck: Correct.

sdks/go/pkg/beam/core/runtime/harness/statemgr_test.go
lostluck: Great spot! Thanks! The tests fortunately still pass. It's a bit redundant with the channel check, but I'd rather ensure that all places cancelFn is called, the forceRecreate is called.

sdks/go/pkg/beam/core/runtime/harness/statemgr.go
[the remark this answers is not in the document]
youngoli: Ok, so if I understand correctly that race condition situation would only be happening when the worker is being terminated, so it's no big deal if it panics. In the other case where c.DoneCh is still called but no race condition happens, the rest of this function still executes fine (not that it makes much of a difference, since the worker is being terminated). Sounds fine to me.

PR #10381, opened 2019-12-13
[BEAM-8962] Add option to disable the metric container accumulator

mxm:
The FlinkMetricContainer wraps the Beam metric container for reporting metrics,
but also stores them as Flink accumulators. With high parallelism jobs with over
a thousand tasks and many built-in Beam metrics for every Beam step, this can
accumulate to over 100MB of serialized data which is stored in the JobManager's
ExecutionGraph. This then fails to even sent over the wire, due to the
akka.framesize limit (10MB by default), and manifests in {{500 Internal Server
Error}}s in the web frontend.

This introduces an option to disable the reporting via accumulators. The
accumulator is mostly useful for batch workloads where you can retrieve the
final accumulator values at the end of the job. It adds a lot of memory and
network overhead.

In the future, we might disable it by default on streaming workloads.

--- | --- | --- | --- | ---

### DIFF
--- a/runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkPipelineOptions.java
+++ b/runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkPipelineOptions.java
@@ -159,6 +159,15 @@ public interface FlinkPipelineOptions
 
   void setEnableMetrics(Boolean enableMetrics);
 
+  @Description(
+      "By default, uses Flink accumulators to store the metrics which allows to query metrics from the PipelineResult. "
+          + "If set to true, metrics will still be reported but can't be queried via PipelineResult. "
+          + "This saves network and memory.")
+  @Default.Boolean(false)
+  Boolean getDisableMetricAccumulator();
+
+  void setDisableMetricAccumulator(Boolean disableMetricAccumulator);
+
   /** Enables or disables externalized checkpoints. */
   @Description(
       "Enables or disables externalized checkpoints. "
--- a/runners/flink/src/main/java/org/apache/beam/runners/flink/metrics/FlinkMetricContainer.java
+++ b/runners/flink/src/main/java/org/apache/beam/runners/flink/metrics/FlinkMetricContainer.java
@@ -46,6 +46,11 @@ import org.slf4j.LoggerFactory;
 /**
  * Helper class for holding a {@link MetricsContainerImpl} and forwarding Beam metrics to Flink
  * accumulators and metrics.
+ *
+ * <p>Using accumulators can be turned off because it is memory and network intensive. The
+ * accumulator results are only meaningful in batch applications or testing streaming applications
+ * which have a defined end. They are not essential during execution because metrics will also be
+ * reported using the configured metrics reporter.
  */
 public class FlinkMetricContainer {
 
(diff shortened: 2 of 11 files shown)


runners/flink/src/main/java/org/apache/beam/runners/flink/metrics/FlinkMetricContainer.java
```diff
@@ -62,22 +67,27 @@
...
     this.flinkDistributionGaugeCache = new HashMap<>();
     this.flinkGaugeCache = new HashMap<>();
 
-    Accumulator<MetricsContainerStepMap, MetricsContainerStepMap> metricsAccumulator =
-        runtimeContext.getAccumulator(ACCUMULATOR_NAME);
-    if (metricsAccumulator == null) {
+    Accumulator<MetricsContainerStepMap, MetricsContainerStepMap> metricsAccumulator;
+    if (accumulatorDisabled) {
+      // Do not register the accumulator with Flink
       metricsAccumulator = new MetricsAccumulator();
```
ibzib: Is this accumulator still doing anything if it's not registered?

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkPipelineOptions.java
```diff
@@ -159,6 +159,15 @@
 
   void setEnableMetrics(Boolean enableMetrics);
 
+  @Description(
+      "By default, uses Flink accumulators to store the metrics which allows to query metrics from the PipelineResult. "
+          + "If set to true, metrics will still be reported but can't be queried via PipelineResult. "
```
ibzib:
```suggestion
          + "If set to true, metrics will still be reported to the configured metrics reporter but can't be queried via PipelineResult. "
```

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkPipelineOptions.java
```diff
@@ -159,6 +159,15 @@
...
+          + "This saves network and memory.")
+  @Default.Boolean(false)
```
ibzib: Nit: I prefer positive flags to avoid the double negative `disabled=false`. Maybe instead something like `useMetricAccumulator` as you did in the tests?

mxm: Unfortunately, we can't set true-defaulting options to false. That's why I inverted the flag's name. See https://issues.apache.org/jira/browse/BEAM-8959

runners/flink/src/main/java/org/apache/beam/runners/flink/metrics/FlinkMetricContainer.java
mxm: Yes, it holds all local metrics.

runners/flink/src/main/java/org/apache/beam/runners/flink/FlinkPipelineOptions.java
tweise: I find it important that flags are named to express their semantics. In this case, the flag's purpose is to turn something off, and there are many similar examples where we would use "skip", "no" etc.

There should not be a case where `disabled=false` needs to be specified by the user. 
But maybe flags should not have default values, since they are fully described by name.

That would also avoid the options to be included with their defaults during job submission, even when the user has not specified them.

mxm:
I think it is worth to think about disabling the metric accumulator by default and only enable it via a `--enable_metric_accumulator`. The reason is that it provides very little value. The accumulator is used to aggregate the final metric values to write them to the configured Beam MetricSink. However, this is only done on job completion which makes this feature useless for streaming applications. Even for batch, you probably want to be able to see metrics during job execution which the accumulator does not provide. I'm inclined to remove the feature entirely.

PR #10384, opened 2019-12-14
[BEAM-8933] Utilities for converting Arrow schemas and reading Arrow batches as Rows

TheNeuralBit:
Adds `ArrowConversion` and `ArrowConversionTest`

--- | --- | --- | --- | ---


iemejia:
I am a bit worried of adding arrow to core, I know Arrow is fantastic but can we envision to package this as an extension? My point is the extra dependencies, in particular netty that has proved to be a total PITA in the past with runners like Spark.

TheNeuralBit:
That's a fair concern. Do you think we should just have an "arrow" extension that includes this utility and anything else arrow-related? 

The first use of these functions will be BigQueryIO (cc: @11moon11) where we would like to have support for both Arrow and Avro in the Storage API. I suppose we would need to have a separate BQ IO with Arrow support in the extension package then?

TheNeuralBit:
> I suppose we would need to have a separate BQ IO with Arrow support in the extension package then?

Oh a better option would be for `:sdks:java:io:google-cloud-platform` to just include a dependency on `:sdks:java:extensions:arrow`, that doesn't seem too onerous

iemejia:
Yes exactly as your last proposal, we already did that for ZetaSQL in the past too (making it an extension) that is a dep of GCP IO. Thanks for understanding my point. As an extra point a module module for Arrow related code may end up benefiting other IOs without the tradeoff of leaking too many things into core.

TheNeuralBit:
Pushed a commit that moved the arrow code into `:sdks:java:extension:arrow`, thanks for the suggestion @iemejia. I want to do a little more clean up tomorrow before sending this out for review but it should be enough for @11moon11 to work from for the gcp io changes.

sdks/java/extensions/arrow/src/main/java/org/apache/beam/sdk/extensions/arrow/ArrowSchema.java
```diff
@@ -0,0 +1,446 @@
...
+import org.apache.beam.sdk.schemas.Schema.Field;
+import org.apache.beam.sdk.schemas.Schema.FieldType;
+import org.apache.beam.sdk.schemas.logicaltypes.FixedBytes;
+import org.apache.beam.sdk.values.Row;
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
+
+/** Utilities to create Beam RowWithGetter instances backed by Arrow record batches. */
+@Experimental(Experimental.Kind.SCHEMAS)
+public class ArrowSchema {
```
11moon11: Just realised that Storage API has a proto with the same name, which makes it a little inconvenient to when both classes are required.
`com.google.cloud.bigquery.storage.v1beta1#ArrowSchema`

11moon11:
It also looks like we should add `package-info.java` and a few other minor changes to make `./gradlew :sdks:java:extensions:arrow:checkstyleMain` happy.

TheNeuralBit:
R: @reuvenlax

TheNeuralBit:
@alexvanboxel do you have time to take a look at this?

cc: @emkornfield in case you want to take a look

TheNeuralBit:
Run Java PreCommit

sdks/java/extensions/arrow/src/main/java/org/apache/beam/sdk/extensions/arrow/ArrowConversion.java
```diff
@@ -0,0 +1,448 @@
...
+                        childrenFields.size() == 1,
+                        "Encountered "
+                            + childrenFields.size()
+                            + " child fields for list type, expected 1");
+                    return FieldType.array(toBeamField(childrenFields.get(0)).getType());
+                  }
+
+                  @Override
+                  public FieldType visit(ArrowType.FixedSizeList type) {
+                    throw new IllegalArgumentException(
```
emkornfield: Is the goal of this to perfectly retain information?  If it isn't then it seems like you could create an FieldType.array() here as well.

emkornfield: might at least be worth a comment as to rationale.

sdks/java/extensions/arrow/src/main/java/org/apache/beam/sdk/extensions/arrow/ArrowConversion.java
```diff
@@ -0,0 +1,448 @@
...
+
+                  @Override
+                  public FieldType visit(ArrowType.Interval type) {
+                    throw new IllegalArgumentException(
+                        "Type \'" + type.toString() + "\' not supported.");
+                  }
+
+                  @Override
+                  public FieldType visit(ArrowType.Duration type) {
+                    throw new IllegalArgumentException(
```
emkornfield: at some point you might consider providing the option of converting this and other underlying logical types to there physical representation and passing along metadata about them.

sdks/java/extensions/arrow/src/main/java/org/apache/beam/sdk/extensions/arrow/ArrowConversion.java
```diff
@@ -0,0 +1,448 @@
...
+                  }
+
+                  @Override
+                  public FieldType visit(ArrowType.Bool type) {
+                    return FieldType.BOOLEAN;
+                  }
+
+                  @Override
+                  public FieldType visit(ArrowType.Decimal type) {
+                    // FieldType.DECIMAL isn't perfect here since arrow decimal has a
```
emkornfield: we the conversion is to Beam though is this really a problem?  or is the goal to only provide perfect round tripping?

sdks/java/extensions/arrow/src/main/java/org/apache/beam/sdk/extensions/arrow/ArrowConversion.java
TheNeuralBit: Honestly I think this was just laziness to begin with. I was only going after the low-hanging fruit to get @11moon11 unblocked.

But to answer your question - I would like to retain information. We should add a FixedSizeList logical type to map this to. I'll add a comment to that effect.

sdks/java/extensions/arrow/src/test/java/org/apache/beam/sdk/extensions/arrow/ArrowConversionTest.java
```diff
@@ -0,0 +1,179 @@
...
+                    new ArrowType.List(),
+                    field("int32s", new ArrowType.Int(32, true))),
+                field("boolean", new ArrowType.Bool()),
+                field("fixed_size_binary", new ArrowType.FixedSizeBinary(3))));
+
+    Schema beamSchema = ArrowConversion.toBeamSchema(schema);
+
+    VectorSchemaRoot expectedSchemaRoot = VectorSchemaRoot.create(schema, allocator);
+    expectedSchemaRoot.setRowCount(16);
+    for (FieldVector vector : expectedSchemaRoot.getFieldVectors()) {
```
emkornfield: I believe VectorSchemaRoot has an allocateNew

sdks/java/extensions/arrow/src/main/java/org/apache/beam/sdk/extensions/arrow/ArrowConversion.java
```diff
@@ -0,0 +1,448 @@
...
+    }
+
+    @Override
+    public boolean hasNext() {
+      return currRowIndex < vectorSchemaRoot.getRowCount();
+    }
+
+    @Override
+    public Row next() {
+      Row result =
```
emkornfield: should there be a precondition check here that to assert there are more rows?

sdks/java/extensions/arrow/src/main/java/org/apache/beam/sdk/extensions/arrow/ArrowConversion.java
```diff
@@ -0,0 +1,448 @@
...
+      }
+
+      @Override
+      public Function<Object, Object> visit(ArrowType.Time type) {
+        throw new IllegalArgumentException("Type \'" + type.toString() + "\' not supported.");
+      }
+
+      @Override
+      public Function<Object, Object> visit(ArrowType.Timestamp type) {
+        DateTimeZone tz = DateTimeZone.forID(type.getTimezone());
```
emkornfield: is it ok for tz to be null or empty string here?  I forget the details on the java implementation but the Arrow spec differentiates between timestamps [with/without timezones](https://github.com/apache/arrow/blob/master/format/Schema.fbs#L178)

sdks/java/extensions/arrow/src/main/java/org/apache/beam/sdk/extensions/arrow/ArrowConversion.java
```diff
@@ -0,0 +1,448 @@
...
+
+  /**
+   * Returns an {@link Iterable<Row>} backed by the Arrow record batch stored in {@code
+   * vectorSchemaRoot}.
+   *
+   * <p>Note this is a lazy interface. The data in the underlying Arrow buffer is not read until a
+   * field of one of the returned {@link Row}s is accessed.
+   */
+  public static Iterable<Row> rowsFromRecordBatch(
+      Schema schema, VectorSchemaRoot vectorSchemaRoot) {
```
emkornfield: you should clarify ownership of VectorSchemaRoot (i.e. will this iterator close after done?)

emkornfield:
@TheNeuralBit @iemejia Regarding Arrow dependencies there was a discussion on potentially moving to a [different allocator recently](https://mail-archives.apache.org/mod_mbox/arrow-dev/202001.mbox/%3CCAK7Z5T_KEnfFmVvg_EUzTH5OFuAu3OFK5NMhVeQ0UTHE%2BQ5smw%40mail.gmail.com%3E) on the Arrow Dev mailing list.  If you have thoughts please chime in.

sdks/java/extensions/arrow/src/main/java/org/apache/beam/sdk/extensions/arrow/ArrowConversion.java
```diff
@@ -0,0 +1,448 @@
...
+    for (org.apache.arrow.vector.types.pojo.Field field : fields) {
+      Field beamField = toBeamField(field);
+      builder.addField(beamField);
+    }
+    return builder.build();
+  }
+
+  /** Get Beam Field from Arrow Field. */
+  private static Field toBeamField(org.apache.arrow.vector.types.pojo.Field field) {
+    FieldType beamFieldType = toFieldType(field.getFieldType(), field.getChildren());
```
reuvenlax: include nullable as well?

sdks/java/extensions/arrow/src/main/java/org/apache/beam/sdk/extensions/arrow/ArrowConversion.java
reuvenlax: +1 to a logical type

sdks/java/extensions/arrow/src/main/java/org/apache/beam/sdk/extensions/arrow/ArrowConversion.java
```diff
@@ -0,0 +1,448 @@
...
+                  @Override
+                  public FieldType visit(ArrowType.FixedSizeList type) {
+                    throw new IllegalArgumentException(
+                        "Type \'" + type.toString() + "\' not supported.");
+                  }
+
+                  @Override
+                  public FieldType visit(ArrowType.Union type) {
+                    throw new IllegalArgumentException(
+                        "Type \'" + type.toString() + "\' not supported.");
```
reuvenlax: We now have OneOf - can you map to that?

sdks/java/extensions/arrow/src/main/java/org/apache/beam/sdk/extensions/arrow/ArrowConversion.java
reuvenlax: We might want to make DECIMAL a logical type at some point, parameterized by the precision.

sdks/java/extensions/arrow/src/main/java/org/apache/beam/sdk/extensions/arrow/ArrowConversion.java
```diff
@@ -0,0 +1,448 @@
...
+                  }
+
+                  @Override
+                  public FieldType visit(ArrowType.Duration type) {
+                    throw new IllegalArgumentException(
+                        "Type \'" + type.toString() + "\' not supported.");
+                  }
+                });
+    return fieldType.withNullable(arrowFieldType.isNullable());
```
reuvenlax: Arrow seems to have isNullable on both the field and the field type. Are they guaranteed to both be set?

sdks/java/extensions/arrow/src/main/java/org/apache/beam/sdk/extensions/arrow/ArrowConversion.java
TheNeuralBit: Yeah I actually lumped that into a jira for making DATETIME logical, [BEAM-7554](https://issues.apache.org/jira/browse/BEAM-7554). I'll add a comment referencing that here

sdks/java/extensions/arrow/src/main/java/org/apache/beam/sdk/extensions/arrow/ArrowConversion.java
TheNeuralBit: I think FieldType here is just a construction for the java library combining Field and Type. In the flatbuffers metadata nullability is only specified on the [Field](https://github.com/apache/arrow/blob/master/format/Schema.fbs#L302)

sdks/java/extensions/arrow/src/main/java/org/apache/beam/sdk/extensions/arrow/ArrowConversion.java
TheNeuralBit: This is handled in `toFieldType`

sdks/java/extensions/arrow/src/main/java/org/apache/beam/sdk/extensions/arrow/ArrowConversion.java
```diff
@@ -0,0 +1,448 @@
...
+import org.joda.time.DateTimeZone;
+
+/**
+ * Utilities to create {@link Iterable}s of Beam {@link Row} instances backed by Arrow record
+ * batches.
+ */
+@Experimental(Experimental.Kind.SCHEMAS)
+public class ArrowConversion {
+  /** Converts Arrow schema to Beam row schema. */
+  public static Schema toBeamSchema(org.apache.arrow.vector.types.pojo.Schema schema) {
```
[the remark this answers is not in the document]
TheNeuralBit: +1 this makes sense, it could be a separate internal class or a separate class entirely (new file).

sdks/java/extensions/arrow/src/main/java/org/apache/beam/sdk/extensions/arrow/ArrowConversion.java
TheNeuralBit: Let's add a comment/jira to do this later, we can verify BigQuery integration without it. (Does BigQuery have a type that would map to FixedSizeList?)

sdks/java/extensions/arrow/src/main/java/org/apache/beam/sdk/extensions/arrow/ArrowConversion.java
```diff
@@ -0,0 +1,448 @@
...
+import org.apache.beam.sdk.schemas.logicaltypes.FixedBytes;
+import org.apache.beam.sdk.values.Row;
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
+
+/**
+ * Utilities to create {@link Iterable}s of Beam {@link Row} instances backed by Arrow record
+ * batches.
+ */
+@Experimental(Experimental.Kind.SCHEMAS)
```
TheNeuralBit, line 48: Consider having a new kind of experimental for Arrow support instead of piggyback on SCHEMAS

sdks/java/extensions/arrow/src/main/java/org/apache/beam/sdk/extensions/arrow/ArrowConversion.java
TheNeuralBit: Ah yes I think that would be a good mapping.
Let's add a comment/jira to do this later though, we can verify BigQuery integration without it.

sdks/java/extensions/arrow/src/main/java/org/apache/beam/sdk/extensions/arrow/ArrowConversion.java
TheNeuralBit: This is still a thorny issue, see this recent [dev@ thread](https://lists.apache.org/thread.html/r9cab743a2ec4278cc6195b624dc920fc42605b0a8966f1dc8058dcd2%40%3Cdev.beam.apache.org%3E). I'm inclined to save this for later as well.

emkornfield:
> I am a bit worried of adding arrow to core, I know Arrow is fantastic but can we envision to package this as an extension? My point is the extra dependencies, in particular netty that has proved to be a total PITA in the past with runners like Spark.

As an update Arrow has now decoupled from the need to netty as a compile time dependency (memory allocation implementations have been separated into separate modules).  I think the two potentially remaining problematic dependencies might be jackson and SLF4j

sdks/java/extensions/arrow/src/main/java/org/apache/beam/sdk/extensions/arrow/ArrowConversion.java
TheNeuralBit: Yes we should clarify this. I think joda DateTime always has a timezone, we may need to reject timezone naive timestamps.

Here's a stable reference to arrow spec regarding timezone-naive timestamps: https://github.com/apache/arrow/blob/6e29200cebeb94f6014c72d25b1bc3a1be9cff1c/format/Schema.fbs#L235-L239

sdks/java/extensions/arrow/src/main/java/org/apache/beam/sdk/extensions/arrow/ArrowConversion.java
TheNeuralBit: yes let's add this

sdks/java/extensions/arrow/src/test/java/org/apache/beam/sdk/extensions/arrow/ArrowConversionTest.java
TheNeuralBit: [You're right](https://arrow.apache.org/docs/java/reference/org/apache/arrow/vector/VectorSchemaRoot.html#allocateNew--), let's use it here.

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/CachingFactory.java
```diff
@@ -31,7 +31,7 @@
  * significant for larger schemas) on each lookup. This wrapper caches the value returned by the
  * inner factory, so the schema comparison only need happen on the first lookup.
  */
-class CachingFactory<CreatedT> implements Factory<CreatedT> {
+public class CachingFactory<CreatedT> implements Factory<CreatedT> {
```
TheNeuralBit, line 34: Let's mark this `@Internal`, it's only public so we can use it from the arrow extension.

sdks/java/extensions/arrow/src/main/java/org/apache/beam/sdk/extensions/arrow/ArrowConversion.java
TheNeuralBit: +1 let's document the ownership model here. Let me know if you need help with this.

TheNeuralBit (commented):
Isidro will be working on reviving this for BEAM-8933. I'm just leaving some comments here to help clarify what work still needs to be done.

sdks/java/extensions/arrow/build.gradle
```diff
@@ -0,0 +1,33 @@
...
+
+description = "Apache Beam :: SDKs :: Java :: Extensions :: Arrow"
+
+def arrow_version = "0.15.1"
+
+dependencies {
+    compile library.java.slf4j_api
+    compile library.java.vendored_guava_26_0_jre
+    compile project(path: ":sdks:java:core", configuration: "shadow")
+    compile "org.apache.arrow:arrow-vector:0.15.1"
```
emkornfield, line 30: lets make sure to upgrade the version to 3.0.0 (4.0.0 should be released pretty soon).

TheNeuralBit: Ah thanks. I actually just noted that in the jira but it's good to have it all in one place :)

PR #10385, opened 2019-12-16
[release-2.18.0][BEAM-8967] Maven artifact beam-sdks-java-core does not have JSR305 specified as "compile"

iemejia:
A tiny error we found when validating the generated pom files from the previous cherry-pick. We should be good to go after this one.

R: @udim 
CC: @suztomo


suztomo:

PR #10408, opened 2019-12-17
[BEAM-8960]: Add an option for user to opt out of using insert id for BigQuery Streaming ingest.

yirutang:
Expose an option so that user can opt out of using insert id while streaming into BigQuery. Insert id only guarantees best effort insert rows deduplication, without it, user will be able to opt into using new streaming backend with higher quotas and reliabilities.


yirutang:
R:reuvenlax

yirutang:
retest this please

PR #10418, opened 2019-12-18
[BEAM-9014] CachingShuffleBatchReader use bytes to limit cache size.

tysonjh:
Limiting cache size by bytes.

The target batch size is 4MiB but may vary depending on many factors. Using a byte limit for the cache is beneficial,
  * large batches will cause OOMs less often since less will be cached
  * more small batches fit into the cache

### DIFF
--- a/runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/ApplianceShuffleEntryReader.java
+++ b/runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/ApplianceShuffleEntryReader.java
@@ -44,9 +44,9 @@ public class ApplianceShuffleEntryReader implements ShuffleEntryReader {
         new ChunkingShuffleBatchReader(executionContext, operationContext, applianceShuffleReader);
 
     if (cache) {
-      // Limit the size of the cache.
-      final int maxBatches = 32;
-      batchReader = new CachingShuffleBatchReader(batchReader, maxBatches);
+      // Limit the size of the cache to ~32 full shuffle batches.
+      final long maxBytes = 128L * 1024 * 1024;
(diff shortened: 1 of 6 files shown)


lukecwik:
> Limiting cache size by bytes stabilizes the memory footprint of the cache since there may be variability in batch sizes.
> 
> NOTE: this is a WIP since I'm currently unable to run the tests and would like to benchmark the performance. The current cache size is set to 32kb, which seems small to me but I am not familiar enough with the general pipeline to be certain.

The current cache size is 32 batches, is each batch really only 1KiB. I was under the impression that batches were larger but I haven't looked at this code in a very long time.

tysonjh:
> > Limiting cache size by bytes stabilizes the memory footprint of the cache since there may be variability in batch sizes.
> > NOTE: this is a WIP since I'm currently unable to run the tests and would like to benchmark the performance. The current cache size is set to 32kb, which seems small to me but I am not familiar enough with the general pipeline to be certain.
> 
> The current cache size is 32 batches, is each batch really only 1KiB? I was under the impression that batches were larger but I haven't looked at this code in a very long time.

The batch size limit is actually 4MiB so I'll change the limit to 128MiB.

tysonjh:
@lukecwik would you review please?

lukecwik:
Thanks for the contribution.

Tyson, could you create a JIRA account as per the [contribution guide](https://beam.apache.org/contribute/#share-your-intent) for sharing your intent. Then I can add you as a contributor to the project which would allow you to assign JIRAs to yourself (specifically BEAM-9014 which I created for this change). Note that all PRs should have an accompanying JIRA associated with them.

tysonjh:
Thanks Luke. My jira account has the same username - tysonjh.

lukecwik:
I have added you as a contributor to the BEAM JIRA project. You can now assign JIRAs to yourself (you could have always been able to create them).

PR #10424, opened 2019-12-19
[BEAM-8996] Auto-generate pipeline options documentation for FlinkRunner

mxm:
The documentation of the pipeline options on the runner page easily becomes
outdated. In order for it to stay up to date, we auto-generate the documentation
from the FlinkPipelineOptions class. This is done for both Java and Python.

These two tasks are run before executing the tests:

- generatePipelineOptionsTableJava
- generatePipelineOptionsTablePython

--- | --- | --- | --- | ---


runners/flink/src/test/java/org/apache/beam/runners/flink/website/PipelineOptionsTableGenerator.java
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  /**
+   * Returns the extracted list of options via reflections on FlinkPipelineOptions. Options are
+   * returned sorted in alphabetical order since Java does not guarantee any consistent order on the
+   * class methods.
+   */
+  private static List<Option> extractOptions(boolean isPython) {
+    List<Option> options = new ArrayList<>();
+    for (Method method : FlinkPipelineOptions.class.getDeclaredMethods()) {
```
lgajowy: What would you say for using filter & map & collect for this? Like this: 

```
private static List<Option> extractOptions(boolean isPython) {
    return Arrays.stream(FlinkPipelineOptions.class.getDeclaredMethods())
        .filter(method -> method.getName().matches("^(get|is).*"))
        .map(method -> {
          String name = extractName(isPython, method);
          String description = extractDescription(method, name);
          String defaultValue = extractDefaultValue(method);
      return new Option(name, description, defaultValue);
    }).collect(Collectors.toList());
  }
```

and the methods: 

```
private static String extractName(boolean isPython, Method method) {
    String name = method.getName().replaceFirst("^(get|is)", "");

    if (isPython) {
      name = CaseFormat.UPPER_CAMEL.to(CaseFormat.LOWER_UNDERSCORE, name);
    } else {
      name = Character.toLowerCase(name.charAt(0)) + name.substring(1);
    }
    return name;
  }

  private static String extractDescription(Method method, String name) {
    Description descriptionAnnotation = method.getAnnotation(Description.class);
    if (descriptionAnnotation == null) {
      throw new RuntimeException(
          "All pipeline options should have a description. Please add one for " + name);
    }
    return descriptionAnnotation.value();
  }
```

runners/flink/src/test/java/org/apache/beam/runners/flink/website/PipelineOptionsTableGenerator.java
```diff
@@ -0,0 +1,179 @@
...
+        }
+
+        Description descriptionAnnotation = method.getAnnotation(Description.class);
+        if (descriptionAnnotation == null) {
+          throw new RuntimeException(
+              "All pipeline options should have a description. Please add one for " + name);
+        }
+        description = descriptionAnnotation.value();
+
+        Optional<String> defaultValueFromAnnotation = getDefaultValueFromAnnotation(method);
```
lgajowy: I'm not sure we should use Optional here if we pass the null further to Options() anyway. Notice that in case when optional is empty the defaultValue is assigned to null anyway

runners/flink/src/test/java/org/apache/beam/runners/flink/website/PipelineOptionsTableGenerator.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+    String arg = args[0].toLowerCase();
+    if (!supportedLanguages.contains(arg)) {
+      throw new RuntimeException("The language is not supported: " + arg);
+    }
+    boolean isPython = arg.equals("python");
+
+    printHeader();
+    List<Option> options = extractOptions(isPython);
+    printOptionsTable(options);
```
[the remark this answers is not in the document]
mxm: I think using a template library or even a template to generate a simple table is overkill here.

runners/flink/src/test/java/org/apache/beam/runners/flink/website/PipelineOptionsTableGenerator.java
mxm: I prefer using iterative methods. Besides, stream is horribly slow. I appreciate the style comments! Using methods is generally more readable, but the code already uses many methods. I don't think we need to break it down further, but that's personal preference.

runners/flink/src/test/java/org/apache/beam/runners/flink/website/PipelineOptionsTableGenerator.java
mxm: `Optional` is used for the return value. It is generally discouraged to use `Optional` on class fields. I agree that we could also return null here, it doesn't matter much though :)

runners/flink/flink_runner.gradle
```diff
@@ -225,3 +225,29 @@ tasks.register('validatesRunner') {
...
 // Generates :runners:flink:1.9:runQuickstartJavaFlinkLocal
 createJavaExamplesArchetypeValidationTask(type: 'Quickstart', runner: 'FlinkLocal')
+
+/**
+ * Updates the documentation with the current pipeline options.
+ */
+def createPipelineOptionsTableTask(String target) {
+  tasks.register("generatePipelineOptionsTable${target}", JavaExec) {
+    group = 'Website'
+    description = "Generates the HTML table for the Flink Runner page"
```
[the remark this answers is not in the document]
mxm: Sure, we can do that. Will do it in a follow-up.

mxm:
Thanks for the review @lgajowy! I agree that we can generify this further. This is the first step to keeping the Flink pipeline options up to date. We can work together with the other runner authors to adapt this to their needs.

mxm:
Unrelated failure in Cassandra test:

```
12:34:36 FAILURE: Build failed with an exception.
12:34:36 
12:34:36 * What went wrong:
12:34:36 Execution failed for task ':sdks:java:io:cassandra:test'.
12:34:36 > There were failing tests. See the report at: file:///home/jenkins/jenkins-slave/workspace/beam_PreCommit_Java_Commit/src/sdks/java/io/cassandra/build/reports/tests/test/index.html
```

runners/flink/flink_runner.gradle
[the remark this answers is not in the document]
mxm: Follow-up: https://github.com/apache/beam/pull/10434

runners/flink/src/test/java/org/apache/beam/runners/flink/website/PipelineOptionsTableGenerator.java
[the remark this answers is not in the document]
lgajowy: I must disagree. Using at least a template and reading it with StringBuilder would simplify the code in this class by separating the view from the generation logic - this alone is a big benefit. Other, previously mentioned points above are still valid too. At the same time, adding the template alone (even only with the StringBuilder reading it) is relatively easy to do.

Right now, future contributors that will need/want to extend/fix this code have to deal with the complexity that otherwise would not be there. It will be therefore harder to generify the solution and adapt it to other runners later and it will probably discourage some to do so.

runners/flink/src/test/java/org/apache/beam/runners/flink/website/PipelineOptionsTableGenerator.java
lgajowy: My take on that is that streams let you separate the consequent steps easily and organize the code well. I don't think the speed of execution is important here. How big the class should be to feel the performance impact? 

Regarding the quantity of the methods - I don't know why is this a problem. I have to disagree one more time. If a method name explains what is done and the method does one thing only then readability increases. IMO, since developers most of the time spent working read the code rather than write it, it's worth investing time and effort to organize the code well. Notice that you wouldn't have to look inside the private methods to see what the public one does when you read the code  (thanks to good naming) so reading and getting to know it is faster.

lgajowy:
@mxm this PR was not accepted but it was merged anyway - we shouldn't act this way. :(

mxm:
I thought I had addressed all your comments. That's why I merged. The only open discussion point seems to resolve about code style, which can be subjective. I definitely always check if any blockers are still open before merging. I merged fast here, though there was sufficient time for the reviewers. 

Please consider, this is not a mission-critical code path. I'm not changing anything related to the Beam runtime. This just generates docs which always became outdated before. I'm essentially doing this on the side and if the review discussion is only about code style, I don't see a reason to block this in case of an isolated class, which is solely used to generate a table in the documentation.

While I'm also guilty of this occasionally, it is worth to take a step back and take a look at the value of a change, instead of insisting on details, which might make it better subjectively, but do not provide more value, or even increase the maintenance costs. I believe I've just recently reviewed some of your PRs, where I tried to live up to that standards.

lgajowy:
Please don't get me the wrong way - I really think this is a good change
and I believe that in my comments I suggested improvements not only to
style but also to functionality (template vs generating code using
system.out.println). I wouldn't say those are subjective. What is more, in
my opinion, after a review round or two we could get even more value from
this pr.

The reason why I think merging this should wait for the reviewer's approval
(not necessarily mine) is that it would prevent missunderstandings like the
one we have right now. I agree this is not "mission critical" code but I
still think we shouldn't judge that on our own.



pt., 20 gru 2019, 14:44 użytkownik Maximilian Michels <
notifications@github.com> napisał:

tweise:
@mxm thanks for working in this, it is a good improvement.

However, regarding the merge if this PR, I would prefer more time is given to allow for comments and consideration of comments. We should wait for an approval, not just assume everything was reviewed and addressed.

There is no reason to rush this type of change.

mxm:
PRs are a means to have a second pair of eyes. An approval is not strictly necessary when a change provides an improvement or feature which is undoubtedly useful or necessary for the success of the project. 

Last time I checked we had close to 200 open pull requests. One reason is that there are simply not enough people reviewing pull requests. Another reason is that the conversation in the PRs progressively slows down because of questions, requests, wishes, or nitpicks. I know some people who have given up contributing to Beam because of too much wishful thinking in the conversations. That is not to say that those comments are not valid, but there is a trade off between in the dimensions of time, complexity, and maintenance that should be carefully thought about when reviewing PRs. 

I always consider all the comments I receive, but I'm only willing to compromise my time to a degree. If a suggestion does not convince me or it simply does not provide enough benefit for the time and mental energy it would to take to implement it, I won't incorporate it, but reject it politely. 

The feature here has already at least twice led to an updated documentation. It works flawlessly and I'm more than happy to see it being improved, generalized, extended to use templating, or xyz. I'm just not going to do this myself at the moment because I would like to focus my mental energy on other work.

Contributing to Beam should be as hard as necessary but not any harder.

PR #10428, opened 2019-12-19
Change access for the TableRowJsonCoder constructor

nahuellofeudo:
The class TableRowJsonCoder is a singleton, which makes it impossible to subclass to tweak the behavior of the Mapper object in cases where non-standard behavior is desired.
Change the constructor of TableRowJsonCoder from private to protected to allow for subclassing.

R: lukecwik

### DIFF
--- a/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/TableRowJsonCoder.java
+++ b/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/TableRowJsonCoder.java
@@ -73,7 +73,7 @@ public class TableRowJsonCoder extends AtomicCoder<TableRow> {
   private static final TableRowJsonCoder INSTANCE = new TableRowJsonCoder();
   private static final TypeDescriptor<TableRow> TYPE_DESCRIPTOR = new TypeDescriptor<TableRow>() {};
 
-  private TableRowJsonCoder() {}
+  protected TableRowJsonCoder() {}
 
   /**
    * {@inheritDoc}


chamikaramj:
LGTM

chamikaramj:
Run Java PreCommit

nahuellofeudo:
Luke, thank you for your comment. We are trying to allow the TableRowJsonCoder to work with non-default settings of ObjectMapper. In particular we had the need to set Feature.ALLOW_UNQUOTED_CONTROL_CHARS. The initial idea was to allow the class to be extended and then customize the constructor to set any relevant flags.
I see your point on the Mapper being private. We wanted to make the fewest possible changes, but if refactoring the mapper reference is OK we can do that.
The only issue I see is that if we introduce withMapper() then the coder can't be a singleton anymore. I can try and make the changes if you are OK with it, and send them for review.

lukecwik:
You can still make the coder returned by `of()` a singleton while `withMapper` would be a unique instance.

You'll need to change TableRowJsonCoder to extend CustomCoder and add a [CloudObjectTranslator](https://github.com/apache/beam/blob/master/runners/google-cloud-dataflow-java/src/main/java/org/apache/beam/runners/dataflow/util/CloudObjectTranslators.java) to maintain pipeline update. For the coder returned by `of()`, you want to maintain the same cloud object representation that `atomic(TableRowJsonCoder.class)` returned while the `withMapper` instances could have the default [CustomCoder mapping](https://github.com/apache/beam/blob/0851849dcfe9b8088d69a5bdd3b55f7806b83ea9/runners/google-cloud-dataflow-java/src/main/java/org/apache/beam/runners/dataflow/util/CloudObjectTranslators.java#L340). Finally, you'll need to add your new translator to the [default registrar](https://github.com/apache/beam/blob/master/runners/google-cloud-dataflow-java/src/main/java/org/apache/beam/runners/dataflow/util/DefaultCoderCloudObjectTranslatorRegistrar.java).

nahuellofeudo:
Thank you. Let me take a stab at it.

PR #10461, opened 2019-12-24
[BEAM-9019] Remove BeamCoderWrapper to avoid extra object allocation

iemejia:
It simplifies #10450 by removing the extra wrapper object allocation to instead reuse the already wrapped Coder and delegate the coding/decoding work to existing methods in `CoderHelpers`.

R: @echauchot

### DIFF
--- a/runners/spark/src/main/java/org/apache/beam/runners/spark/structuredstreaming/translation/helpers/EncoderHelpers.java
+++ b/runners/spark/src/main/java/org/apache/beam/runners/spark/structuredstreaming/translation/helpers/EncoderHelpers.java
@@ -24,7 +24,6 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
-import javax.annotation.Nullable;
 import org.apache.beam.runners.spark.structuredstreaming.translation.SchemaHelpers;
 import org.apache.beam.sdk.coders.Coder;
 import org.apache.spark.sql.Encoder;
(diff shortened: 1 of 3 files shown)


iemejia:

iemejia:

iemejia:

runners/spark/src/main/java/org/apache/beam/runners/spark/structuredstreaming/translation/helpers/EncoderHelpers.java
```diff
@@ -234,30 +194,33 @@ public Expression child() {
...
+      String accessCode = ctx.addReferenceObj("coder", coder, coder.getClass().getName());
       ExprCode input = child.genCode(ctx);
       String javaType = CodeGenerator.javaType(dataType());
+
       List<String> parts = new ArrayList<>();
       List<Object> args = new ArrayList<>();
       /*
             CODE GENERATED:
-            final $javaType ${ev.value} = ($javaType) $beamCoderWrapper.decode(${input.isNull}, ${input.value});
+            final ${javaType} ${ev.value} = (${input.isNull})
```
echauchot: This adds a little more generated code. Out of 4 runs, this PR seems to add from 0.1s to 0.2s to the queries. Maybe doing the ternary expression inside the compiled java code will be better.

iemejia: Ok I moved the ternary expression out of the generated code and now the execution numbers are better (in my machine).

iemejia:
Updated with the requested change. PTAL @echauchot I rebased it accidentally but you can check the requested change in the latest commit.

iemejia:

echauchot:
manual merge for minor final touches.

iemejia:
Thanks, next time please let's discuss about squashing the commits and do not take that decision in your side, I am surprised you did this considering how picky you are about not squashing your own commits.

echauchot:
@iemejia I discussed it in the PR see my comment of 01/06. As you did not comment I thought it was ok.
Regarding my own commits: I apply the same rule remember that in the previous [PR](https://github.com/apache/beam/pull/10450) that you reviewed I also suashed all my refactoring commits into one!

iemejia:
Seems I missed the message. In any case you could have requested it again or remind me to do it, no need to take actions like that without discussing before.

PR #10470, opened 2019-12-27
Revert "[BEAM-8932]  Modify PubsubClient to use the proto message throughout."

boyuanzz:
Changes break PostCommit: https://issues.apache.org/jira/browse/BEAM-9036
R: @udim 
CC: @dpcollins-google

PR #10471, opened 2019-12-27
Perform dependency version update from reverted pr #10470

dpcollins-google:
Perform dependency version update from reverted pr #10470


dpcollins-google:
Run SQL Postcommit

PR #10473, opened 2019-12-27
Add Timestamps to integration test which tries to write into and read out of sql pubsub.

dpcollins-google:
This will fail once timestamp validation is added to the json client.


dpcollins-google:
Run SQL Postcommit

PR #10492, opened 2020-01-03
[BEAM-9041, BEAM-9042] SchemaCoder equals should not rely on from/toRowFunction equality

iemejia:
This PR fixes both issues because (1) one fix depends on the other, and (2) to make it easier to validate/cherry pick into 2.18.0's branch.

BEAM-9041: Don't rely on equality for the from/to functions in `SchemaCoder` because nobody implements equals on `SerializableFunction`. I tried to do this with byte equality, maybe there is a better way to do it but I could not think of another.
BEAM-9042: Since Avro's `Schema` class is not `Serializable` I made it `transient`. Another approach could have been to transform the Schema into a String (not sure if this is needed but I can change it if you think it is worth).

R: @TheNeuralBit

### DIFF
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/SchemaCoder.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/SchemaCoder.java
@@ -102,7 +102,8 @@ public class SchemaCoder<T> extends CustomCoder<T> {
 
   /**
    * Returns a {@link SchemaCoder} for the specified class. If no schema is registered for this
-   * class, then throws {@link NoSuchSchemaException}.
+   * class, then throws {@link NoSuchSchemaException}. The parameter functions to convert from and
+   * to Rows <b>must</b> implement the equals contract.
    */
   public static <T> SchemaCoder<T> of(
       Schema schema,
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/utils/AvroUtils.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/utils/AvroUtils.java
@@ -504,7 +504,10 @@ public class AvroUtils {
 
     private void readObject(ObjectInputStream in) throws IOException, ClassNotFoundException {
       final String avroSchemaAsString = (String) in.readObject();
-      avroSchema = new org.apache.avro.Schema.Parser().parse(avroSchemaAsString);
+      avroSchema =
+          (avroSchemaAsString == null)
+              ? null
+              : new org.apache.avro.Schema.Parser().parse(avroSchemaAsString);
     }
   }
 
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/utils/AvroUtils.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/utils/AvroUtils.java
@@ -20,6 +20,9 @@ package org.apache.beam.sdk.schemas.utils;
 import static org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Preconditions.checkArgument;
 import static org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Preconditions.checkNotNull;
 
+import java.io.IOException;
+import java.io.ObjectInputStream;
+import java.io.ObjectOutputStream;
 import java.lang.reflect.Method;
 import java.math.BigDecimal;
 import java.nio.ByteBuffer;
@@ -466,7 +469,7 @@ public class AvroUtils {
   }
 
   private static class RowToGenericRecordFn implements SerializableFunction<Row, GenericRecord> {
-    private final org.apache.avro.Schema avroSchema;
+    private transient org.apache.avro.Schema avroSchema;
 
     RowToGenericRecordFn(@Nullable org.apache.avro.Schema avroSchema) {
       this.avroSchema = avroSchema;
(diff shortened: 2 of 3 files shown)


sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/utils/AvroUtils.java
```diff
@@ -431,7 +431,20 @@ public static GenericRecord toGenericRecord(
...
+    private final transient org.apache.avro.Schema avroSchema;
+
+    RowToGenericRecordFn(@Nullable org.apache.avro.Schema avroSchema) {
+      this.avroSchema = avroSchema;
+    }
+
+    @Override
+    public GenericRecord apply(Row input) {
+      return toGenericRecord(input, avroSchema);
+    }
```
TheNeuralBit: Couldn't you just implement `equals` here instead of changing to comparing byte equality of the serialized function in SchemaCoder?

iemejia: This is a second different issue about capture of Avro schema on serialization so not really related to equals. As explained above I put both together because I use equality to validate the roundtrip of serialization/deserialization.

iemejia:
This is not a revert. Previous version did not compare from/toRow functions for equality. Do you have any suggestion on how to compare both functions? It is not really clear to me how to do so in particular for functions with no state.

TheNeuralBit:
Sorry revert was a poor choice of words. I just meant a big part of that PR
was making equals work for the functions produced by
GetterBasedSchemaProvider so I would like it if they continued to be used.
For a function without any state I would write the equals function to just
compare this.getClass().equals(other.getClass()) since any two instances
will do the same thing.

kennknowles:
Hmm. I am torn on this. I agree with @TheNeuralBit that mostly these conversions should just own their own `equals`. But I see how `SerializableFunction`, especially lambdas, will usually not implement this.

In the case of the `SchemaCoder` it is often automated, isn't it? And when it isn't, it is fair that it is a one-time up front effort to write a slightly larger class that has a good `equals`. We could provide a convenient base class, I guess, or even now add a `default` implementation in the functional interface maybe?

iemejia:
I see the things more clearly now, thanks for the explanation @TheNeuralBit. I was not aware of the goals of the other PR. I have one question related to that PR. Why did you use the strict conversion (no widening/narrowing) from/to row only on the schemaCoder that generates GenericRecords `AvroUtils#schemaCoder(java.lang.Class<T>)` and not in the others. Or maybe the real question is why we need to do this strictly?

iemejia:
After thinking more on the issue probably we don't have a nice solution for it. We cannot define `equals` (and `hashCode`) on `SerializableFunction` or its parent `ProcessFunction` because Java does not allow to override a member of java.lang.Object.

So the alternative would be to change the signature of SchemaCoder to use a new class that implements SerialiableFunction and equals, but this will require many changes for few returns, maybe the easiest thing to do here is the right one, just to document that the functions to convert from/to rows in SchemaCoder must implement equals/hashCode.

WDYT? I will update the PR with the fix for the Avro case in the meantime.

iemejia:
PR updated I used your equals suggestion to tackle the first part instead of the byte equality change and improved the serialization issue of the internal Avro schema. Now we should be good to go. PTAL @TheNeuralBit

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/SchemaCoder.java
```diff
@@ -102,7 +102,8 @@ protected SchemaCoder(
 
   /**
    * Returns a {@link SchemaCoder} for the specified class. If no schema is registered for this
-   * class, then throws {@link NoSuchSchemaException}.
+   * class, then throws {@link NoSuchSchemaException}. The parameter functions to convert from and
+   * to Rows <b>must</b> implement the equals contract.
```
TheNeuralBit: :+1: thanks!

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/utils/AvroUtils.java
```diff
@@ -431,7 +465,47 @@ public static GenericRecord toGenericRecord(
...
+
+    private void writeObject(ObjectOutputStream out) throws IOException {
+      final String avroSchemaAsString = (avroSchema == null) ? null : avroSchema.toString();
+      out.writeObject(avroSchemaAsString);
+    }
+
+    private void readObject(ObjectInputStream in) throws IOException, ClassNotFoundException {
+      final String avroSchemaAsString = (String) in.readObject();
+      avroSchema = new org.apache.avro.Schema.Parser().parse(avroSchemaAsString);
```
TheNeuralBit: I'm not 100% sure but it looks like the failures are occurring when avroSchema is null. Either way I think you need to check if `avroSchemaAsString` is null here.

iemejia: Yes you are right I forgot to check the nullability of the string before the parse, I will fix that and add a method for this. Hopefully everything will be green at that moment. Thanks for the hint.

iemejia:
Run Java PreCommit

iemejia:
Merging now that the tests are green.

iemejia:
Sorry if I rushed a bit the merge @TheNeuralBit I just wanted to cherry pick it to unblock 2.18.0 release.

TheNeuralBit:
Not at all, sorry I didn't approve sooner.

PR #10494, opened 2020-01-03
typehint fixes to DoOutputsTuple

aaltay:
Changed the output type for `DoOutputsTuple` from PValue to PCollection.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/pvalue.py
+++ b/sdks/python/apache_beam/pvalue.py
@@ -315,7 +315,9 @@ class DoOutputsTuple(object):
         self.producer.add_output(pcoll, tag)
     else:
       # Main output is output of inner ParDo.
-      pcoll = self.producer.parts[0].outputs[None]  # type: PCollection
+      pval = self.producer.parts[0].outputs[None]
+      assert isinstance(pval, PCollection), "DoOutputsTuple should follow a ParDo."
+      pcoll = pval
     self._pcolls[tag] = pcoll
     return pcoll
 
--- a/sdks/python/apache_beam/pvalue.py
+++ b/sdks/python/apache_beam/pvalue.py
@@ -254,7 +254,7 @@ class DoOutputsTuple(object):
     # gets applied.
     self.producer = None  # type: Optional[AppliedPTransform]
     # Dictionary of PCollections already associated with tags.
-    self._pcolls = {}  # type: Dict[Optional[str], PValue]
+    self._pcolls = {}  # type: Dict[Optional[str], PCollection]
 
   def __str__(self):
     return '<%s>' % self._str_internal()
@@ -267,15 +267,15 @@ class DoOutputsTuple(object):
         self.__class__.__name__, self._main_tag, self._tags, self._transform)
 
   def __iter__(self):
-    # type: () -> Iterator[PValue]
-    """Iterates over tags returning for each call a (tag, pvalue) pair."""
+    # type: () -> Iterator[PCollection]
+    """Iterates over tags returning for each call a (tag, pcollection) pair."""
     if self._main_tag is not None:
       yield self[self._main_tag]
     for tag in self._tags:
       yield self[tag]
 
   def __getattr__(self, tag):
-    # type: (str) -> PValue
+    # type: (str) -> PCollection
     # Special methods which may be accessed before the object is
     # fully constructed (e.g. in unpickling).
     if tag[:2] == tag[-2:] == '__':
@@ -283,7 +283,7 @@ class DoOutputsTuple(object):
     return self[tag]
 
   def __getitem__(self, tag):
-    # type: (Union[int, str, None]) -> PValue
+    # type: (Union[int, str, None]) -> PCollection
     # Accept int tags so that we can look at Partition tags with the
     # same ints that we used in the partition function.
     # TODO(gildea): Consider requiring string-based tags everywhere.
@@ -304,7 +304,7 @@ class DoOutputsTuple(object):
     assert self.producer is not None
     if tag is not None:
       self._transform.output_tags.add(tag)
-      pcoll = PCollection(self._pipeline, tag=tag, element_type=typehints.Any)  # type: PValue
+      pcoll = PCollection(self._pipeline, tag=tag, element_type=typehints.Any)
       # Transfer the producer from the DoOutputsTuple to the resulting
       # PCollection.
       pcoll.producer = self.producer.parts[0]
@@ -315,7 +315,7 @@ class DoOutputsTuple(object):
         self.producer.add_output(pcoll, tag)
     else:
       # Main output is output of inner ParDo.
-      pcoll = self.producer.parts[0].outputs[None]
+      pcoll = self.producer.parts[0].outputs[None]  # type: PCollection
     self._pcolls[tag] = pcoll
     return pcoll
 
--- a/sdks/python/apache_beam/pvalue.py
+++ b/sdks/python/apache_beam/pvalue.py
@@ -316,7 +316,8 @@ class DoOutputsTuple(object):
     else:
       # Main output is output of inner ParDo.
       pval = self.producer.parts[0].outputs[None]
-      assert isinstance(pval, PCollection), "DoOutputsTuple should follow a ParDo."
+      assert isinstance(pval, PCollection), (
+          "DoOutputsTuple should follow a ParDo.")
       pcoll = pval
     self._pcolls[tag] = pcoll
     return pcoll
--- a/sdks/python/apache_beam/io/gcp/pubsub.py
+++ b/sdks/python/apache_beam/io/gcp/pubsub.py
@@ -141,11 +141,11 @@ class ReadFromPubSub(PTransform):
   # Implementation note: This ``PTransform`` is overridden by Directrunner.
 
   def __init__(self,
-               topic=None,  # type: Optional[str]
-               subscription=None,  # type: Optional[str]
-               id_label=None,  # type: Optional[str]
+               topic=None,  # type: Union[str, None]
+               subscription=None,  # type: Union[str, None]
+               id_label=None,  # type: Union[str, None]
                with_attributes=False,  # type: bool
-               timestamp_attribute=None  # type: Optional[str]
+               timestamp_attribute=None  # type: Union[str, None]
               ):
     # type: (...) -> None
     """Initializes ``ReadFromPubSub``.
@@ -257,8 +257,8 @@ class WriteToPubSub(PTransform):
   def __init__(self,
                topic,  # type: str
                with_attributes=False,  # type: bool
-               id_label=None,  # type: Optional[str]
-               timestamp_attribute=None  # type: Optional[str]
+               id_label=None,  # type: Union[str, None]
+               timestamp_attribute=None  # type: Union[str, None]
               ):
     # type: (...) -> None
     """Initializes ``WriteToPubSub``.
--- a/sdks/python/apache_beam/io/gcp/pubsub.py
+++ b/sdks/python/apache_beam/io/gcp/pubsub.py
@@ -141,11 +141,11 @@ class ReadFromPubSub(PTransform):
   # Implementation note: This ``PTransform`` is overridden by Directrunner.
 
   def __init__(self,
-               topic=None,  # type: Union[str, None]
-               subscription=None,  # type: Union[str, None]
-               id_label=None,  # type: Union[str, None]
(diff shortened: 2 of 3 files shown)


aaltay:
R: @robertwb could you take a look at this?

- Change to DoOutputsTuple make sense because the bundled outputs of a Flatmap or ParDo should be a PCollection not a PValue.
- Other changes from Optional[X] to Uniuon[X, None] are a bit puzzling. They are supposed to be equivalents but, I noticed some downstream type checkers work with the latter but fail with the former. Still investigating this.

aaltay:
R: @udim -- had an idea about why Optional[X] to Uniuon[X, None]  might be working. (Because Union is not imported and potentially ignored by pytype.)

udim:
> R: @udim -- had an idea about why Optional[X] to Union[X, None] might be working. (Because Union is not imported and potentially ignored by pytype.)

I have a WIP PR (not yet pushed) that runs pytype on Beam. I finally got to pubsub.py and it indeed says `Name 'Union' is not defined` on the changes you made on line 146.

aaltay:
Reverted the `Optional` changes. Could you review the remaining changes?

udim:
95% because I'm not 100% sure that only PCollections should be in there.

aaltay:
> This 95% LGTM. I also scanned the lint logs for what might be related errors.
> 
> CC: @robertwb @chadrik who might have more background/reasoning for or against this change.

Thank you. I will wait for them to comment. DoOutputsTuple bundles outputs of a ParDo, and I beleive that should be alway of PCollection type.

udim:
Yeah, looks like it. (I just took another look at the ciode.)

chadrik:
Hey folks, I'm really glad to see the interest in type hints, but this kind of change is going to make my life a lot more difficult.  I have gotten the Beam source fully passing in mypy in my [other PR](#9056).   This took me at least half a year to complete, and I've been patiently waiting and following up for the last 6 months (#9056 was created on July 12th) to get all of that reviewed and merged.  AFAICT this is just going to break that.  

I would prefer that we get my other PRs completed before we merge this.  I realize that my PRs are pretty big, but there's a good reason for that: typing is a balancing act and it is very difficult to judge the correctness of a few isolated type hints -- you have to reach a critical mass of reinforcing hints across the code base.  So, since I did manage to achieve that critical mass in my PR, if the changes in this PR are different from my final solution (I can't recall if this code gets touched by the next typing PR) there's a good chance that this is not correct, at least not for mypy. 

The current PR is here:  https://github.com/apache/beam/pull/10367

The master PR is here:  #9056

aaltay:
@chadrik - I do not wish to set back your progress. However, there are downstream Beam users that are running into issues due to the typehint changes. Specific example here is that, users that have typechecks enabled in their code that accepts results of DoOutputsTuple and checking that the types are the more accurate PCollection rather than the PValue.

If this PR is proposing an invalid change and outputs could actually be PValue we can drop this change and I can work with downstream consumers to fix their code. However, if this is a valid change I would rather go forward with it and support our users.

IMO we need to prioritize our users. Unfortunately this might be the cost of making this change in master and available to end users in phases. I, @udim, and @robertwb will be happy to continue supporting you. In addition to reviews, if it makes we can also take some of the burden of making the actual changes.

chadrik:
> I do not wish to set back your progress. However, there are downstream Beam users that are running into issues due to the typehint changes. Specific example here is that, users that have typechecks enabled in their code that accepts results of DoOutputsTuple and checking that the types are the more accurate PCollection rather than the PValue.

Are the Beam users using mypy or pytype?

IIUC, Beam type hints should not get picked up by any type checker that properly supports [PEP 561](https://www.python.org/dev/peps/pep-0561/) because we have not added a `py.typed` to the source.  

From the PEP:

>> Package maintainers who wish to support type checking of their code MUST add a marker file named py.typed to their package supporting typing. 

Maybe I'm misunderstanding this, but I thought that preventing this type of problem was part of the motivation behind PEP 561.  

> If this PR is proposing an invalid change and outputs could actually be PValue we can drop this change and I can work with downstream consumers to fix their code. However, if this is a valid change I would rather go forward with it and support our users.

In order to know if this change is valid, I -- or someone else -- will need to look through the remaining changes in my PRs to see how I addressed this.  Unfortunately, I'm pretty busy at the moment.

I certainly understand the need to support users, and I accept that this change may need to go in ASAP, but I would really really love some more support getting my PRs through.

aaltay:
> > I do not wish to set back your progress. However, there are downstream Beam users that are running into issues due to the typehint changes. Specific example here is that, users that have typechecks enabled in their code that accepts results of DoOutputsTuple and checking that the types are the more accurate PCollection rather than the PValue.
> 
> Are the Beam users using mypy or pytype?

I am aware of users using pytype. I do not know about the whole exact setup. Presumably they may have their own py.typed files and linking with Beam at source level (speculating, not sure.)


> 
> IIUC, Beam type hints should not get picked up by any type checker that properly supports [PEP 561](https://www.python.org/dev/peps/pep-0561/) because we have not added a `py.typed` to the source.
> 
> From the PEP:
> 
> > > Package maintainers who wish to support type checking of their code MUST add a marker file named py.typed to their package supporting typing.
> 
> Maybe I'm misunderstanding this, but I thought that preventing this type of problem was part of the motivation behind PEP 561.
> 
> > If this PR is proposing an invalid change and outputs could actually be PValue we can drop this change and I can work with downstream consumers to fix their code. However, if this is a valid change I would rather go forward with it and support our users.
> 
> In order to know if this change is valid, I -- or someone else -- will need to look through the remaining changes in my PRs to see how I addressed this. Unfortunately, I'm pretty busy at the moment.

That is fair. @robertwb and @udim can review for this.

> 
> I certainly understand the need to support users, and I accept that this change may need to go in ASAP, but I would really really love some more support getting my PRs through.

Thank you. What would be particularly helpful? Faster reviews? Splitting some work?

sdks/python/apache_beam/pvalue.py
```diff
@@ -308,7 +308,7 @@ def __getitem__(self, tag):
         self.producer.add_output(pcoll, tag)
     else:
       # Main output is output of inner ParDo.
-      pcoll = self.producer.parts[0].outputs[None]
+      pcoll = self.producer.parts[0].outputs[None]  # type: PCollection
```
chadrik: This doesn't seem safe to me. We should not be promoting the type from `PValue` to a `PCollection` unless we can actually guarantee that it will always be a `PCollection` here.  As far as I can tell, it may actually be `PValue`, but correct me if I'm wrong.

Here's how I came to that conclusion.

The code where `AppliedPTransform.outputs` is assigned values seems to indicate it should be a mapping of `PValue`:

```python
  def add_output(self,
                 output,  # type: Union[pvalue.DoOutputsTuple, pvalue.PValue]
                 tag=None  # type: Union[str, int, None]
                ):
    # type: (...) -> None
    if isinstance(output, pvalue.DoOutputsTuple):
      self.add_output(output[output._main_tag])
    elif isinstance(output, pvalue.PValue):
      # TODO(BEAM-1833): Require tags when calling this method.
      if tag is None and None in self.outputs:
        tag = len(self.outputs)
      assert tag not in self.outputs
      self.outputs[tag] = output
    elif isinstance(output, dict):
      for output_tag, out in output.items():
        self.add_output(out, tag=output_tag)
    else:
      raise TypeError("Unexpected output type: %s" % output)
```

As a result, I have typed `AppliedPTransform.outputs` as `Dict[Union[str, int, None], pvalue.PValue]`.  If it's the case that `AppliedPTransform.outputs` only holds `PCollections` then we should enforce that in the `isinstance` check above (and elsewhere).  

---

Also note that the trailing `# type: PCollection` used here has no effect in mypy, and in fact it generates an error.  In mypy, you can only use this the first time that a variable is defined (e.g. for `pcoll` that's line 300 above this), and you can only use it to broaden a type (as I did originally, going from `PCollection` to `PValue`).   Narrowing the type is basically disregarding/overriding the inspected type, so that requires using `typing.cast()`.

This is why I'm really wary of supporting pytype.

chadrik: If everyone is in agreement that `AppliedPTransform.outputs` can only hold `PCollections` then the typing for `AppliedPTransform` should be fixed.  Doing it this way -- simply overriding the type -- is a superficial fix that leaves the door open for this problem to happen elsewhere in the code where `AppliedPTransform` is used  (and as I mentioned, it doesn't actually work for mypy).

chadrik:
> I am aware of users using pytype. I do not know about the whole exact setup. Presumably they may have their own py.typed files and linking with Beam at source level (speculating, not sure.)

py.typed files are created by maintainers and distributed as part of the package to indicate to type checkers whether the package should be included in the type analysis and if so, in what way.  

It appears that pytype does not support this yet:

https://github.com/google/pytype/issues/151

Can we add `# pytype: skip-file` to this file?

chadrik:
> Thank you. What would be particularly helpful? Faster reviews? Splitting some work?

Faster reviews would be nice, but I understand that this is easier said than done, as everyone has their own tasks they're working on.   I'm happy to adjust how I'm presenting the PRs for reviews, such as including fewer changes per PR.  Feedback welcome!

sdks/python/apache_beam/pvalue.py
aaltay: IIUC, DoOutputsTuple could only follow a ParDo and its outputs should be PCollection's. However this contract is not really captured in the code. You are right that AppliedPTransform.outputs could have types of PValue, but this should not be the case for DoOutputsTuple.

aaltay:
> > I am aware of users using pytype. I do not know about the whole exact setup. Presumably they may have their own py.typed files and linking with Beam at source level (speculating, not sure.)
> 
> py.typed files are created by maintainers and distributed as part of the package to indicate to type checkers whether the package should be included in the type analysis and if so, in what way.
> 
> It appears that pytype does not support this yet:
> 
> [google/pytype#151](https://github.com/google/pytype/issues/151)
> 
> Can we add `# pytype: skip-file` to this file?

@boyuanzz added this comment to all py files for now. So this PR is not very urgent. We do not need to merge it immediately.

aaltay:
> > Thank you. What would be particularly helpful? Faster reviews? Splitting some work?
> 
> Faster reviews would be nice, but I understand that this is easier said than done, as everyone has their own tasks they're working on. I'm happy to adjust how I'm presenting the PRs for reviews, such as including fewer changes per PR. Feedback welcome!

I think making PRs smaller and spreading the load will help. @udim @robertwb @pabloem @tvalentyn  myself, we could all be reviewers on PRs. Especially now we are over the initial phase and if we can get smaller PRs to review.

sdks/python/apache_beam/pvalue.py
chadrik: understood.  If we're confident about that contract I think the proper way to solve this would be something like this:

```python
      pval = self.producer.parts[0].outputs[None]
      assert isinstance(pval, PCollection), "DoOutputsTuple should follow a ParDo"
      pcoll = pval
```
 
If you want to be a bit more formal you could raise a `TypeError` or the like.  

That will naturally narrow the type of `pcoll`.  The reason for the temporary variable `pval` is that you cannot assign of type `PValue` to a variable already defined as `PCollection` (above in the `if tag is not None` block).

robertwb: The types here are correct, `self._pcolls` should be a map to PCollections. This PR looks like an improvement to me. Would you be OK with this going in (with the assertion to do the type restriction)? 

(FWIW, it'd be nice to clean up the whole DoOutputsTuple thing, and how inputs and outputs are represented in general, but that's another issue.)

robertwb:
@chadrik On the note of your typing changes in, yes, I'd like to work with you (and others) to finally get them in. The second pull request should be good to go, it might be worth following up on the list what the status is and what remains (as I, for one, have lost context over the holidays).

sdks/python/apache_beam/pvalue.py
chadrik: yep, looks good to me.

chadrik:
@robertwb Thanks.  Like I said, I completely understand that there's a lot on everyone's plate and that these PRs are both quite large and also non-urgent.  Thanks again for your effort getting these through.

Now that we've done the broad strokes I can start submitting multiple small PRs simultaneously.   I'll post the links to those PRs on the list once they are ready.

aaltay:
Updated this with @chadrik suggestion to assert that pval is of type PCollection

PR #10499, opened 2020-01-03
[BEAM-9049] Add opt-in --shutDownOnThrashing flag

TheNeuralBit:
We've seen several Dataflow pipelines fail when multiple workers are shut down due to thrashing while working on the same work item. I believe instead we should make this an opt-in feature that a user can enable when they are concerned about thrashing and are aware of the risks. 

--- | --- | --- | --- | ---

### DIFF
--- a/runners/google-cloud-dataflow-java/src/main/java/org/apache/beam/runners/dataflow/options/DataflowPipelineDebugOptions.java
+++ b/runners/google-cloud-dataflow-java/src/main/java/org/apache/beam/runners/dataflow/options/DataflowPipelineDebugOptions.java
@@ -206,6 +206,21 @@ public interface DataflowPipelineDebugOptions extends ExperimentalOptions, Pipel
 
   void setSaveHeapDumpsToGcsPath(String gcsPath);
 
+  /**
+   * If {@literal true}, shut down the JVM when sustained periods of GC thrashing are detected. If
+   * {@link #getDumpHeapOnOOM} is also set a heap dump will be saved before shutting down due to
+   * thrashing.
+   *
+   * <p>CAUTION: Enabling this feature can cause a pipeline that would otherwise succeed to fail
+   * when multiple workers are shut down due to thrashing while working on the same work item.
+   */
+  @Description(
+      "If true, shut down the JVM when sustained periods of GC thrashing are detected. If "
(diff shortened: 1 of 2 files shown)


TheNeuralBit:
R: @kennknowles

angoenka:
Do we need to expose this option in python to pass it to JRH?

TheNeuralBit:
Good point, this will also turn off thrashing detection for Python in the JRH, and as its written now there will be no way to opt-in because of `@Hidden`. I think that's acceptable, I wouldn't imagine thrashing in the JVM would be very likely for the JRH?

We could also just remove the `@Hidden` but then that brings in several other flags related to debugging java workers.

angoenka:
> Good point, this will also turn off thrashing detection for Python in the JRH, and as its written now there will be no way to opt-in because of `@Hidden`. I think that's acceptable, I wouldn't imagine thrashing in the JVM would be very likely for the JRH?
> 
> We could also just remove the `@Hidden` but then that brings in several other flags related to debugging java workers.

I agree, as we don't run custom code on JRH in python pipelines, JHR memory thrashing would be unlikely.

angoenka:
Also, is it possible to log `Memory is used/total/max = 5761/8103/8103 MB, GC last/max = 80.97/83.28 %, #pushbacks=0, gc thrashing=true` in warning when the VM is thrashing?

TheNeuralBit:
Good idea, I'll add a change for that.

TheNeuralBit:
After investigating further I think this may not be the best solution. 

In the streaming worker, we call `MemoryMonitor#waitForResources` before grabbing a work item. This has the effect of reducing parallelism when there's memory pressure. This is supposed to be the primary method of dealing with thrashing, and shutting down the JVM when there's sustained thrashing is intended as a last resort, when we are unable to process a single work item without encountering thrashing.

The batch worker never calls `waitForResources`, I'm not sure if this is intentional or if it's an oversight, but it seems that the actual root of our problem is that in batch we aren't using both of these features together as they are intended. Now I'm thinking we should either:
1) add a wait to the batch worker to reduce parallelism in the face of memory pressure, or 
2) disable the JVM shutdown on thrashing, but just for the batch worker.

TheNeuralBit:
Closing this as I think we're better off making changes just for the batch worker.

anuragkyal:
@TheNeuralBit any progress on this?

PR #10515, opened 2020-01-07
[BEAM-8701] Update commons-io to version 2.6

iemejia:
Removed additionally the unused commons-io 1.x dependency
R: @aromanenko-dev


iemejia:

iemejia:

lukecwik:
I would suggest using the linkage checker to verify whether dependency updates will cause linkage errors at runtime. It adds a lot of confidence around whether the change will cause a linkage error during runtime.

See https://github.com/apache/beam/blob/e2f2bf26ff6738841ff03d1bc73ffa475ddfd4a1/build.gradle#L290

PR #10524, opened 2020-01-07
[BEAM-8717] Update commons-lang3 to version 3.9

iemejia:
R: @lukecwik


iemejia:
Run Java PreCommit

iemejia:
Run Java PreCommit

iemejia:
Run Java PreCommit

iemejia:
Run Java PreCommit

iemejia:
Run Java PreCommit

lukecwik:
Should we be vendoring commons-lang3 instead of using the shaded classes from beam-sdks-java-core?

buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy
```diff
@@ -432,7 +432,7 @@ class BeamModulePlugin implements Plugin<Project> {
         commons_compress                            : "org.apache.commons:commons-compress:1.19",
         commons_csv                                 : "org.apache.commons:commons-csv:1.4",
         commons_io                                  : "commons-io:commons-io:2.6",
-        commons_lang3                               : "org.apache.commons:commons-lang3:3.6",
+        commons_lang3                               : "org.apache.commons:commons-lang3:3.9",
```
lukecwik: You could remove this from this list and have the full maven coordinates appear in sdks/java/core/build.gradle since it is no longer used elsewhere.

iemejia: Done

iemejia:
Agree about vendoring I will try to tackle the vendoring one next. Just to clarify the overall goal is to get commons-lang3 out of core in favor of our own vendored version right?
Filled https://issues.apache.org/jira/browse/BEAM-9077 btw.

iemejia:
Wondering how little is our use of commons-lang3 classes in sdks/java/core if we should just try to get rid of the dep there and let the other modules deal with the issue by them selves (in the case of runners the dep is mostly provided). WDYT ?

lukecwik:
It does look like all the uses are trivial and could be replaced by a one or two lines of code that the helper function is doing so that sounds like a much better idea.

PR #10546, opened 2020-01-09
[BEAM-9008] Add CassandraIO readAll method

vmarquez:
Edit: This PR adds a readAll method that returns a PTransform<PCollection<Read<A>>, PCollection<A>> to allow for more fine grained querying of a Cassandra compatible databases (Cassandra, Scylla etc).  Since each Read<A> allows one to specify a Query and/or RingRange, a pipeline does not have to do a full table scan to retrieve data, it can now read a subset using this pattern.  

Because Read<A> allows the user to specify a Set<RingRange>, this allows even more flexibility, in that a user can decide to put N number of RingRanges (which each could be representing individual keys) in a single Read<A>, this allows one to *linearize* queries, which is very important when doing a large number of individual queries so as to not overload single shards or cores.   

_Old comment for posterity, not relevant to the current redesign after working with @iemejia: This PR adds a readAll method that returns a `PTransform<PColl<List<RingRange>>, PCollection<T>>` to allow for parallel querying of a small subset of a Cassandra databse.  In addition, it refactors the current method `read` that returns a `PTransform<PBegin, PColl<T>>` to be a ParDo based one, so we can share code between read and readAll.  
Note: One thing I wasn't sure about was how to abstract over AutoValue classes.  I'm much more familiar with Scala than Java and haven't used AutoValue before, so I wasn't sure if there was a better way than what I did with having a `getCassandraConfig` on both the static Read class as well as the ReadAll class.  Don't think I can use an interface since I don't want the methods public.  Open to suggestions here._ 

Thanks in advance for being patient with a large change and please let me know if I can change/add anything!  

R:@iemejia 

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
+++ b/sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
@@ -21,13 +21,9 @@ import static org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Prec
 import static org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Preconditions.checkState;
 
 import com.datastax.driver.core.Cluster;
-import com.datastax.driver.core.ColumnMetadata;
 import com.datastax.driver.core.ConsistencyLevel;
 import com.datastax.driver.core.PlainTextAuthProvider;
 import com.datastax.driver.core.QueryOptions;
-import com.datastax.driver.core.ResultSet;
-import com.datastax.driver.core.ResultSetFuture;
-import com.datastax.driver.core.Row;
 import com.datastax.driver.core.Session;
 import com.datastax.driver.core.SocketOptions;
 import com.datastax.driver.core.policies.DCAwareRoundRobinPolicy;
@@ -36,32 +32,32 @@ import com.google.auto.value.AutoValue;
 import java.math.BigInteger;
 import java.util.ArrayList;
 import java.util.Collections;
-import java.util.Iterator;
 import java.util.List;
-import java.util.NoSuchElementException;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.function.BiFunction;
 import java.util.stream.Collectors;
+import javax.annotation.Nullable;
 import org.apache.beam.sdk.annotations.Experimental;
 import org.apache.beam.sdk.annotations.Experimental.Kind;
 import org.apache.beam.sdk.coders.Coder;
-import org.apache.beam.sdk.io.BoundedSource;
+import org.apache.beam.sdk.coders.SerializableCoder;
 import org.apache.beam.sdk.options.PipelineOptions;
 import org.apache.beam.sdk.options.ValueProvider;
+import org.apache.beam.sdk.transforms.Create;
 import org.apache.beam.sdk.transforms.DoFn;
 import org.apache.beam.sdk.transforms.PTransform;
 import org.apache.beam.sdk.transforms.ParDo;
+import org.apache.beam.sdk.transforms.Reshuffle;
 import org.apache.beam.sdk.transforms.SerializableFunction;
-import org.apache.beam.sdk.transforms.display.DisplayData;
 import org.apache.beam.sdk.values.PBegin;
 import org.apache.beam.sdk.values.PCollection;
 import org.apache.beam.sdk.values.PDone;
+import org.apache.beam.sdk.values.TypeDescriptor;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.annotations.VisibleForTesting;
-import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Joiner;
-import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.Iterators;
-import org.checkerframework.checker.nullness.qual.Nullable;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.ImmutableSet;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -88,6 +84,11 @@ import org.slf4j.LoggerFactory;
  *     // above options are the minimum set, returns PCollection<Person>
  *
  * }</pre>
+ *
+ * <p>Alternatively, one may use {@code CassandraIO.<Person>readAll()
+ * .withCoder(SerializableCoder.of(Person.class))} to query a subset of the Cassandra database by
+ * creating a PCollection of {@code CassandraIO.Read<Person>} each with their own query or
+ * RingRange.
  *
  * <h3>Writing to Apache Cassandra</h3>
  *
@@ -135,6 +136,11 @@ public class CassandraIO {
   /** Provide a {@link Read} {@link PTransform} to read data from a Cassandra database. */
   public static <T> Read<T> read() {
     return new AutoValue_CassandraIO_Read.Builder<T>().build();
+  }
+
+  /** Provide a {@link ReadAll} {@link PTransform} to read data from a Cassandra database. */
+  public static <T> ReadAll<T> readAll() {
+    return new AutoValue_CassandraIO_ReadAll.Builder<T>().build();
   }
 
   /** Provide a {@link Write} {@link PTransform} to write data to a Cassandra database. */
@@ -185,6 +191,9 @@ public class CassandraIO {
     abstract @Nullable ValueProvider<Integer> readTimeout();
 
     abstract @Nullable SerializableFunction<Session, Mapper> mapperFactoryFn();
+
+    @Nullable
+    abstract ValueProvider<Set<RingRange>> ringRanges();
 
     abstract Builder<T> builder();
 
@@ -371,6 +380,14 @@ public class CassandraIO {
       return builder().setMapperFactoryFn(mapperFactory).build();
     }
 
+    public Read<T> withRingRanges(Set<RingRange> ringRange) {
+      return withRingRanges(ValueProvider.StaticValueProvider.of(ringRange));
+    }
+
+    public Read<T> withRingRanges(ValueProvider<Set<RingRange>> ringRange) {
+      return builder().setRingRanges(ringRange).build();
+    }
+
     @Override
     public PCollection<T> expand(PBegin input) {
       checkArgument((hosts() != null && port() != null), "WithHosts() and withPort() are required");
@@ -379,7 +396,69 @@ public class CassandraIO {
       checkArgument(entity() != null, "withEntity() is required");
       checkArgument(coder() != null, "withCoder() is required");
 
-      return input.apply(org.apache.beam.sdk.io.Read.from(new CassandraSource<>(this, null)));
+      PCollection<Read<T>> splits =
+          input
+              .apply(Create.of(this))
+              .apply("Create Splits", ParDo.of(new SplitFn<T>()))
+              .setCoder(SerializableCoder.of(new TypeDescriptor<Read<T>>() {}));
+
+      return splits.apply("ReadAll", CassandraIO.<T>readAll().withCoder(coder()));
+    }
+
+    private static class SplitFn<T> extends DoFn<Read<T>, Read<T>> {
+      @ProcessElement
+      public void process(
+          @Element CassandraIO.Read<T> read, OutputReceiver<Read<T>> outputReceiver) {
+        Set<RingRange> ringRanges = getRingRanges(read);
+        for (RingRange rr : ringRanges) {
+          Set<RingRange> subset = ImmutableSet.<RingRange>of(rr);
+          outputReceiver.output(read.withRingRanges(ImmutableSet.of(rr)));
+        }
+      }
+
+      private static <T> Set<RingRange> getRingRanges(Read<T> read) {
+        try (Cluster cluster =
+            getCluster(
+                read.hosts(),
+                read.port(),
+                read.username(),
+                read.password(),
+                read.localDc(),
+                read.consistencyLevel(),
+                read.connectTimeout(),
+                read.readTimeout())) {
+          if (isMurmur3Partitioner(cluster)) {
+            LOG.info("Murmur3Partitioner detected, splitting");
+            Integer splitCount;
+            if (read.minNumberOfSplits() != null && read.minNumberOfSplits().get() != null) {
+              splitCount = read.minNumberOfSplits().get();
+            } else {
+              splitCount = cluster.getMetadata().getAllHosts().size();
+            }
+            List<BigInteger> tokens =
+                cluster.getMetadata().getTokenRanges().stream()
+                    .map(tokenRange -> new BigInteger(tokenRange.getEnd().getValue().toString()))
+                    .collect(Collectors.toList());
+            SplitGenerator splitGenerator =
+                new SplitGenerator(cluster.getMetadata().getPartitioner());
+
+            return splitGenerator.generateSplits(splitCount, tokens).stream()
+                .flatMap(List::stream)
+                .collect(Collectors.toSet());
+
+          } else {
+            LOG.warn(
+                "Only Murmur3Partitioner is supported for splitting, using an unique source for "
+                    + "the read");
+            String partitioner = cluster.getMetadata().getPartitioner();
+            RingRange totalRingRange =
+                RingRange.of(
+                    SplitGenerator.getRangeMin(partitioner),
+                    SplitGenerator.getRangeMax(partitioner));
+            return Collections.singleton(totalRingRange);
+          }
+        }
+      }
     }
 
     @AutoValue.Builder
@@ -418,6 +497,8 @@ public class CassandraIO {
 
       abstract Optional<SerializableFunction<Session, Mapper>> mapperFactoryFn();
 
+      abstract Builder<T> setRingRanges(ValueProvider<Set<RingRange>> ringRange);
+
       abstract Read<T> autoBuild();
 
       public Read<T> build() {
@@ -425,390 +506,6 @@ public class CassandraIO {
           setMapperFactoryFn(new DefaultObjectMapperFactory(entity().get()));
         }
         return autoBuild();
-      }
-    }
-  }
-
-  @VisibleForTesting
-  static class CassandraSource<T> extends BoundedSource<T> {
-    final Read<T> spec;
-    final List<String> splitQueries;
-    // split source ached size - can't be calculated when already split
-    Long estimatedSize;
-    private static final String MURMUR3PARTITIONER = "org.apache.cassandra.dht.Murmur3Partitioner";
-
-    CassandraSource(Read<T> spec, List<String> splitQueries) {
-      this(spec, splitQueries, null);
-    }
-
-    private CassandraSource(Read<T> spec, List<String> splitQueries, Long estimatedSize) {
-      this.estimatedSize = estimatedSize;
-      this.spec = spec;
-      this.splitQueries = splitQueries;
-    }
-
-    @Override
-    public Coder<T> getOutputCoder() {
-      return spec.coder();
-    }
-
-    @Override
-    public BoundedReader<T> createReader(PipelineOptions pipelineOptions) {
-      return new CassandraReader(this);
-    }
-
-    @Override
-    public List<BoundedSource<T>> split(
-        long desiredBundleSizeBytes, PipelineOptions pipelineOptions) {
-      try (Cluster cluster =
-          getCluster(
-              spec.hosts(),
-              spec.port(),
-              spec.username(),
-              spec.password(),
-              spec.localDc(),
-              spec.consistencyLevel(),
-              spec.connectTimeout(),
-              spec.readTimeout())) {
-        if (isMurmur3Partitioner(cluster)) {
-          LOG.info("Murmur3Partitioner detected, splitting");
-          return splitWithTokenRanges(
-              spec, desiredBundleSizeBytes, getEstimatedSizeBytes(pipelineOptions), cluster);
-        } else {
-          LOG.warn(
-              "Only Murmur3Partitioner is supported for splitting, using a unique source for "
-                  + "the read");
-          return Collections.singletonList(
-              new CassandraIO.CassandraSource<>(spec, Collections.singletonList(buildQuery(spec))));
-        }
-      }
-    }
-
-    private static String buildQuery(Read spec) {
-      return (spec.query() == null)
-          ? String.format("SELECT * FROM %s.%s", spec.keyspace().get(), spec.table().get())
-          : spec.query().get().toString();
-    }
-
-    /**
-     * Compute the number of splits based on the estimated size and the desired bundle size, and
-     * create several sources.
-     */
-    private List<BoundedSource<T>> splitWithTokenRanges(
-        CassandraIO.Read<T> spec,
-        long desiredBundleSizeBytes,
-        long estimatedSizeBytes,
-        Cluster cluster) {
-      long numSplits =
-          getNumSplits(desiredBundleSizeBytes, estimatedSizeBytes, spec.minNumberOfSplits());
-      LOG.info("Number of desired splits is {}", numSplits);
-
-      SplitGenerator splitGenerator = new SplitGenerator(cluster.getMetadata().getPartitioner());
-      List<BigInteger> tokens =
-          cluster.getMetadata().getTokenRanges().stream()
-              .map(tokenRange -> new BigInteger(tokenRange.getEnd().getValue().toString()))
-              .collect(Collectors.toList());
-      List<List<RingRange>> splits = splitGenerator.generateSplits(numSplits, tokens);
-      LOG.info("{} splits were actually generated", splits.size());
-
-      final String partitionKey =
-          cluster.getMetadata().getKeyspace(spec.keyspace().get()).getTable(spec.table().get())
-              .getPartitionKey().stream()
-              .map(ColumnMetadata::getName)
-              .collect(Collectors.joining(","));
-
-      List<TokenRange> tokenRanges =
-          getTokenRanges(cluster, spec.keyspace().get(), spec.table().get());
-      final long estimatedSize = getEstimatedSizeBytesFromTokenRanges(tokenRanges) / splits.size();
-
-      List<BoundedSource<T>> sources = new ArrayList<>();
-      for (List<RingRange> split : splits) {
-        List<String> queries = new ArrayList<>();
-        for (RingRange range : split) {
-          if (range.isWrapping()) {
-            // A wrapping range is one that overlaps from the end of the partitioner range and its
-            // start (ie : when the start token of the split is greater than the end token)
-            // We need to generate two queries here : one that goes from the start token to the end
-            // of
-            // the partitioner range, and the other from the start of the partitioner range to the
-            // end token of the split.
-            queries.add(generateRangeQuery(spec, partitionKey, range.getStart(), null));
-            // Generation of the second query of the wrapping range
-            queries.add(generateRangeQuery(spec, partitionKey, null, range.getEnd()));
-          } else {
-            queries.add(generateRangeQuery(spec, partitionKey, range.getStart(), range.getEnd()));
-          }
-        }
-        sources.add(new CassandraIO.CassandraSource<>(spec, queries, estimatedSize));
-      }
-      return sources;
-    }
-
-    private static String generateRangeQuery(
-        Read spec, String partitionKey, BigInteger rangeStart, BigInteger rangeEnd) {
-      final String rangeFilter =
-          Joiner.on(" AND ")
-              .skipNulls()
-              .join(
-                  rangeStart == null
-                      ? null
-                      : String.format("(token(%s) >= %d)", partitionKey, rangeStart),
-                  rangeEnd == null
-                      ? null
-                      : String.format("(token(%s) < %d)", partitionKey, rangeEnd));
-      final String query =
-          (spec.query() == null)
-              ? buildQuery(spec) + " WHERE " + rangeFilter
-              : buildQuery(spec) + " AND " + rangeFilter;
-      LOG.debug("CassandraIO generated query : {}", query);
-      return query;
-    }
-
-    private static long getNumSplits(
-        long desiredBundleSizeBytes,
-        long estimatedSizeBytes,
-        @Nullable ValueProvider<Integer> minNumberOfSplits) {
-      long numSplits =
-          desiredBundleSizeBytes > 0 ? (estimatedSizeBytes / desiredBundleSizeBytes) : 1;
-      if (numSplits <= 0) {
-        LOG.warn("Number of splits is less than 0 ({}), fallback to 1", numSplits);
-        numSplits = 1;
-      }
-      return minNumberOfSplits != null ? Math.max(numSplits, minNumberOfSplits.get()) : numSplits;
-    }
-
-    /**
-     * Returns cached estimate for split or if missing calculate size for whole table. Highly
-     * innacurate if query is specified.
-     *
-     * @param pipelineOptions
-     * @return
-     */
-    @Override
-    public long getEstimatedSizeBytes(PipelineOptions pipelineOptions) {
-      if (estimatedSize != null) {
-        return estimatedSize;
-      } else {
-        try (Cluster cluster =
-            getCluster(
-                spec.hosts(),
-                spec.port(),
-                spec.username(),
-                spec.password(),
-                spec.localDc(),
-                spec.consistencyLevel(),
-                spec.connectTimeout(),
-                spec.readTimeout())) {
-          if (isMurmur3Partitioner(cluster)) {
-            try {
-              List<TokenRange> tokenRanges =
-                  getTokenRanges(cluster, spec.keyspace().get(), spec.table().get());
-              this.estimatedSize = getEstimatedSizeBytesFromTokenRanges(tokenRanges);
-              return this.estimatedSize;
-            } catch (Exception e) {
-              LOG.warn("Can't estimate the size", e);
-              return 0L;
-            }
-          } else {
-            LOG.warn("Only Murmur3 partitioner is supported, can't estimate the size");
-            return 0L;
-          }
-        }
-      }
-    }
-
-    @VisibleForTesting
-    static long getEstimatedSizeBytesFromTokenRanges(List<TokenRange> tokenRanges) {
-      long size = 0L;
-      for (TokenRange tokenRange : tokenRanges) {
-        size = size + tokenRange.meanPartitionSize * tokenRange.partitionCount;
-      }
-      return Math.round(size / getRingFraction(tokenRanges));
-    }
-
-    @Override
-    public void populateDisplayData(DisplayData.Builder builder) {
-      super.populateDisplayData(builder);
-      if (spec.hosts() != null) {
-        builder.add(DisplayData.item("hosts", spec.hosts().toString()));
-      }
-      if (spec.port() != null) {
-        builder.add(DisplayData.item("port", spec.port()));
-      }
-      builder.addIfNotNull(DisplayData.item("keyspace", spec.keyspace()));
-      builder.addIfNotNull(DisplayData.item("table", spec.table()));
-      builder.addIfNotNull(DisplayData.item("username", spec.username()));
-      builder.addIfNotNull(DisplayData.item("localDc", spec.localDc()));
-      builder.addIfNotNull(DisplayData.item("consistencyLevel", spec.consistencyLevel()));
-    }
-    // ------------- CASSANDRA SOURCE UTIL METHODS ---------------//
-
-    /**
-     * Gets the list of token ranges that a table occupies on a give Cassandra node.
-     *
-     * <p>NB: This method is compatible with Cassandra 2.1.5 and greater.
-     */
-    private static List<TokenRange> getTokenRanges(Cluster cluster, String keyspace, String table) {
-      try (Session session = cluster.newSession()) {
-        ResultSet resultSet =
-            session.execute(
-                "SELECT range_start, range_end, partitions_count, mean_partition_size FROM "
-                    + "system.size_estimates WHERE keyspace_name = ? AND table_name = ?",
-                keyspace,
-                table);
-
-        ArrayList<TokenRange> tokenRanges = new ArrayList<>();
-        for (Row row : resultSet) {
-          TokenRange tokenRange =
-              new TokenRange(
-                  row.getLong("partitions_count"),
-                  row.getLong("mean_partition_size"),
-                  new BigInteger(row.getString("range_start")),
-                  new BigInteger(row.getString("range_end")));
-          tokenRanges.add(tokenRange);
-        }
-        // The table may not contain the estimates yet
-        // or have partitions_count and mean_partition_size fields = 0
-        // if the data was just inserted and the amount of data in the table was small.
-        // This is very common situation during tests,
-        // when we insert a few rows and immediately query them.
-        // However, for tiny data sets the lack of size estimates is not a problem at all,
-        // because we don't want to split tiny data anyways.
-        // Therefore, we're not issuing a warning if the result set was empty
-        // or mean_partition_size and partitions_count = 0.
-        return tokenRanges;
-      }
-    }
-
-    /** Compute the percentage of token addressed compared with the whole tokens in the cluster. */
-    @VisibleForTesting
-    static double getRingFraction(List<TokenRange> tokenRanges) {
-      double ringFraction = 0;
-      for (TokenRange tokenRange : tokenRanges) {
-        ringFraction =
-            ringFraction
-                + (distance(tokenRange.rangeStart, tokenRange.rangeEnd).doubleValue()
-                    / SplitGenerator.getRangeSize(MURMUR3PARTITIONER).doubleValue());
-      }
-      return ringFraction;
-    }
-
-    /**
-     * Check if the current partitioner is the Murmur3 (default in Cassandra version newer than 2).
-     */
-    @VisibleForTesting
-    static boolean isMurmur3Partitioner(Cluster cluster) {
-      return MURMUR3PARTITIONER.equals(cluster.getMetadata().getPartitioner());
-    }
-
-    /** Measure distance between two tokens. */
-    @VisibleForTesting
-    static BigInteger distance(BigInteger left, BigInteger right) {
-      return (right.compareTo(left) > 0)
-          ? right.subtract(left)
-          : right.subtract(left).add(SplitGenerator.getRangeSize(MURMUR3PARTITIONER));
-    }
-
-    /**
-     * Represent a token range in Cassandra instance, wrapping the partition count, size and token
-     * range.
-     */
-    @VisibleForTesting
-    static class TokenRange {
-      private final long partitionCount;
-      private final long meanPartitionSize;
-      private final BigInteger rangeStart;
-      private final BigInteger rangeEnd;
-
-      TokenRange(
-          long partitionCount, long meanPartitionSize, BigInteger rangeStart, BigInteger rangeEnd) {
-        this.partitionCount = partitionCount;
-        this.meanPartitionSize = meanPartitionSize;
-        this.rangeStart = rangeStart;
-        this.rangeEnd = rangeEnd;
-      }
-    }
-
-    private class CassandraReader extends BoundedSource.BoundedReader<T> {
-      private final CassandraIO.CassandraSource<T> source;
-      private Cluster cluster;
-      private Session session;
-      private Iterator<T> iterator;
-      private T current;
-
-      CassandraReader(CassandraSource<T> source) {
-        this.source = source;
-      }
-
-      @Override
-      public boolean start() {
-        LOG.debug("Starting Cassandra reader");
-        cluster =
-            getCluster(
-                source.spec.hosts(),
-                source.spec.port(),
-                source.spec.username(),
-                source.spec.password(),
-                source.spec.localDc(),
-                source.spec.consistencyLevel(),
-                source.spec.connectTimeout(),
-                source.spec.readTimeout());
-        session = cluster.connect(source.spec.keyspace().get());
-        LOG.debug("Queries: " + source.splitQueries);
-        List<ResultSetFuture> futures = new ArrayList<>();
-        for (String query : source.splitQueries) {
-          futures.add(session.executeAsync(query));
-        }
-
-        final Mapper<T> mapper = getMapper(session, source.spec.entity());
-
-        for (ResultSetFuture result : futures) {
-          if (iterator == null) {
-            iterator = mapper.map(result.getUninterruptibly());
-          } else {
-            iterator = Iterators.concat(iterator, mapper.map(result.getUninterruptibly()));
-          }
-        }
-
-        return advance();
-      }
-
-      @Override
-      public boolean advance() {
-        if (iterator.hasNext()) {
-          current = iterator.next();
-          return true;
-        }
-        current = null;
-        return false;
-      }
-
-      @Override
-      public void close() {
-        LOG.debug("Closing Cassandra reader");
-        if (session != null) {
-          session.close();
-        }
-        if (cluster != null) {
-          cluster.close();
-        }
-      }
-
-      @Override
-      public T getCurrent() throws NoSuchElementException {
-        if (current == null) {
-          throw new NoSuchElementException();
-        }
-        return current;
-      }
-
-      @Override
-      public CassandraIO.CassandraSource<T> getCurrentSource() {
-        return source;
-      }
-
-      private Mapper<T> getMapper(Session session, Class<T> enitity) {
-        return source.spec.mapperFactoryFn().apply(session);
       }
     }
   }
@@ -1179,7 +876,7 @@ public class CassandraIO {
   }
 
   /** Get a Cassandra cluster using hosts and port. */
-  private static Cluster getCluster(
+  static Cluster getCluster(
       ValueProvider<List<String>> hosts,
       ValueProvider<Integer> port,
       ValueProvider<String> username,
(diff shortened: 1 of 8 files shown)


suztomo:
Would you ensure the change passes ` ./gradlew -p sdks/java check`? I see it fails with format violations:

```
suztomo-macbookpro44:vmarquez-beam suztomo$ git log -1
commit d85e6b95d64664af37148c792b35ff02abfd90f5 (HEAD -> feature/BEAM-9008/cassandraio_readall, origin/feature/BEAM-9008/cassandraio_readall)
Author: Vincent Marquez <vincent.marquez@ytel.com>
Date:   Wed Dec 18 16:48:52 2019 -0800

    implementing CassandraIO.readAll
suztomo-macbookpro44:vmarquez-beam suztomo$ ./gradlew -p sdks/java check
Configuration on demand is an incubating feature.

...

* What went wrong:
Execution failed for task ':sdks:java:io:cassandra:spotlessJava'.
> The following files had format violations:
      sdks/java/io/cassandra/src/test/java/org/apache/beam/sdk/io/cassandra/CassandraIOTest.java
          @@ -184,31 +184,33 @@
           
           ····LOGGER.info("Insert·records");
           ····String[][]·scientists·=·{
          -······new·String·[]·{"phys",·"Einstein"},
          -······new·String·[]·{"bio",·"Darwin"},
          -······new·String·[]·{"phys",·"Copernicus"},
          -······new·String·[]·{"bio",·"Pasteur"},
          -······new·String·[]·{"bio",·"Curie"},
          -······new·String·[]·{"phys",·"Faraday"},
          -······new·String·[]·{"math",·"Newton"},
          -······new·String·[]·{"phys",·"Bohr"},
          -······new·String·[]·{"phys",·"Galilei"},
          -······new·String·[]·{·"math",·"Maxwell"},
          +······new·String[]·{"phys",·"Einstein"},
          +······new·String[]·{"bio",·"Darwin"},
          +······new·String[]·{"phys",·"Copernicus"},
          +······new·String[]·{"bio",·"Pasteur"},
          +······new·String[]·{"bio",·"Curie"},
          +······new·String[]·{"phys",·"Faraday"},
          +······new·String[]·{"math",·"Newton"},
          +······new·String[]·{"phys",·"Bohr"},
          +······new·String[]·{"phys",·"Galilei"},
          +······new·String[]·{"math",·"Maxwell"},
           ····};
           ····for·(int·i·=·0;·i·<·NUM_ROWS;·i++)·{
           ······int·index·=·i·%·scientists.length;
          -······String·insertStr·=·String.format(
          -··········"INSERT·INTO·%s.%s(person_department,·person_id,·person_name)·values("
          -··············+·"'"·+·scientists[index][0]·+·"',·"
          -··············+·i
          -··············+·",·'"
          -··············+·scientists[index][1]
          -··············+·"');",
          -··········CASSANDRA_KEYSPACE,
          -··········CASSANDRA_TABLE);
          -········LOGGER.error("Error·with·str·=·"·+·insertStr);
          -······session.execute(insertStr
          -··········);
          +······String·insertStr·=
          +··········String.format(
          +··············"INSERT·INTO·%s.%s(person_department,·person_id,·person_name)·values("
          +··················+·"'"
          +··················+·scientists[index][0]
          +··················+·"',·"
          +··················+·i
          +··················+·",·'"
          +··················+·scientists[index][1]
          +··················+·"');",
      ... (102 more lines that didn't fit)
  Violations also present in:
      sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/QueryFn.java
      sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraConfig.java
      sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
  Run 'gradlew spotlessApply' to fix these violations.
```

iemejia:
Arrghh tests for this one did not start, for that reason you had not seen the failures that @suztomo mentioned.
You can auto fix the code format issues by running `./gradlew :sdks:java:io:cassandra:check spotlessApply`.
I will start letting some review comments in the meantime

vmarquez:
@suztomo @iemejia Thank you for pointing that out, I was not aware of the formatting issues.  Maybe after this MR I can add some of that to the contributing document. :)  I fixed the style and formatting issues, and will await other change requests.

iemejia:
Arrghh unit tests did not run because of the issues with jenkins this week, running again.

iemejia:
Working still in the review sorry for the delay @vmarquez

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
```diff
@@ -325,7 +359,58 @@ private CassandraIO() {}
       checkArgument(entity() != null, "withEntity() is required");
       checkArgument(coder() != null, "withCoder() is required");
 
-      return input.apply(org.apache.beam.sdk.io.Read.from(new CassandraSource<>(this, null)));
+      try (Cluster cluster =
+          getCluster(hosts(), port(), username(), password(), localDc(), consistencyLevel())) {
+        if (isMurmur3Partitioner(cluster)) {
+          LOG.info("Murmur3Partitioner detected, splitting");
+
+          List<BigInteger> tokens =
```
iemejia: For maintenance coherence in DoFn based IOs we follow a pattern where the first part Creates the required object to query (the one I mentioned above), the second splits the query into multiples queries for each partition via a `SplitFn extends DoFn` and then with the result of such splits we do a read equivalent to current `QueryFn`. Can you refactor this segment to behave like that (with the query object I mentioned above).

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
```diff
@@ -608,121 +470,32 @@ static double getRingFraction(List<TokenRange> tokenRanges) {
...
-
-      private Mapper<T> getMapper(Session session, Class<T> enitity) {
-        return source.spec.mapperFactoryFn().apply(session);
-      }
+  /**
+   * Represent a token range in Cassandra instance, wrapping the partition count, size and token
+   * range.
+   */
+  @VisibleForTesting
+  static class TokenRange {
```
iemejia: We don't need this one anymore I think we can remove this class now as well as: `CassandraIO.getRingFraction` and `CassandraIOTest.testRingFraction`. Can you remove these too please.

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/RingRange.java
```diff
@@ -17,10 +17,11 @@
  */
 package org.apache.beam.sdk.io.cassandra;
 
+import java.io.Serializable;
 import java.math.BigInteger;
 
 /** Models a Cassandra token range. */
-final class RingRange {
+final class RingRange implements Serializable {
```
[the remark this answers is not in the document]
iemejia: Actually this should probably be done not for this class but for the `CassandraQuery` object

sdks/java/io/cassandra/src/test/java/org/apache/beam/sdk/io/cassandra/CassandraIOTest.java
```diff
@@ -312,6 +298,52 @@ public void testRead() throws Exception {
     pipeline.run();
   }
 
+  @Test
+  public void testReadAll() {
+
```
iemejia: nit: remove space

sdks/java/io/cassandra/src/test/java/org/apache/beam/sdk/io/cassandra/CassandraIOTest.java
```diff
@@ -387,6 +420,7 @@ public Mapper apply(Session input) {
 
     @Override
     public Iterator map(ResultSet resultSet) {
+
```
iemejia: nit: remove space

sdks/java/io/cassandra/src/test/java/org/apache/beam/sdk/io/cassandra/CassandraIOTest.java
```diff
@@ -466,40 +500,9 @@ public void testCustomMapperImplDelete() {
     assertEquals(1, counter.intValue());
   }
 
-  @Test
-  public void testSplit() throws Exception {
-    PipelineOptions options = PipelineOptionsFactory.create();
```
iemejia: Once we have the SplitFn we should probably test this, actually we had a lot of historic issues with Splitting on this connector so please pay attention to ensure we don't introduce regressions.

iemejia:
Btw, don't hesitate to ask questions or discuss here if you have any doubts before starting.

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/QueryFn.java
```diff
@@ -0,0 +1,89 @@
...
+import com.datastax.driver.core.ResultSet;
+import com.datastax.driver.core.Session;
+import com.datastax.driver.core.Token;
+import java.util.Iterator;
+import java.util.stream.Collectors;
+import org.apache.beam.sdk.transforms.DoFn;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class QueryFn<T> extends DoFn<Iterable<RingRange>, T> {
```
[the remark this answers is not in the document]
vmarquez: I don't think we want this flattened here, the way I'm using readAll currently is I am hashing into groups of queries based on actual shard, which is owned by a core (in ScyllaDB), so grouping queries gives finer grained control over how much load we put on a core.  Someone using cassandra could use a similar split idea for making sure to limit the number of concurrent TokenRanges owned by the same node in Cassandra.

sdks/java/io/cassandra/src/test/java/org/apache/beam/sdk/io/cassandra/CassandraIOTest.java
vmarquez: Hmm, if you know of relevant issues that we previously had, lmk, I will also look through the issue tracker to see if there were previous bugs.

vmarquez:
Thanks for taking a look, I've pushed a new commit (I won't be squashing anymore until we finalize everything) that hopefully addresses all the minor style issues.  LMK if I missed anything. 

I do want to make sure we keep the cassandraIO 'idiomatic' to the rest of the IO connectors, but I don't think modeling this after the SOLR one will work.  For one thing, if we want to share the `ReadFn` class between both Read and ReadAll, it means we have to have some way of having both use it and pass in 'connection' information, which we can't do if the signature of ReadFn is `ReadFn extends DoFn<ReadAll<A>, A>`.  I think another class to look at for something that has both Read and ReadAll PTransforms is the SpannerIO, which is modeled similarly to how I did it here (though not exactly). 

https://github.com/apache/beam/blob/master/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/spanner/SpannerIO.java#L315

They have a configuration class there is public (which we can't do since we want to keep backwards compatibility with the current way `Read` works), it has two different PTransforms, `Read` uses `ReadAll` internally, etc.  

I do think instead of taking a collection of RingRanges, taking some sort of 'Query' object makes sense, and the idea that it doesn't have to tie in to the actual connection means we can split up the CassandraConfig class.  Thoughts on that?

iemejia:
Sorry again for taking a bit to respond, my review queue ended up being longer than usual this week.

I agree on your intuition on letting the component idiomatic, I just took a look at SpannerIO and it is indeed aligned with my first idea `ReadOperation` == `Query`, (I call it query following the naming of `HbaseQuery`). Naming is less important than the fact that this object ‘should have’ only parameters related to the Read/Query and not to the connections. So I agreed 100% on that.

However my intuition changed a bit when I started to think that `ReadOperation/Query` in every case is a subset of Read. Read = Connection + ReadOperation/Query, so why we need ReadOperation at all? We can have just the normal Read but if you set up explicitly the RingRage as an extra parameter well we don’t split and just read from it. SolrIO creates an artificial KV to put that ‘query’ in the side of the normal Read, but for example DynamoDBIO does the read based on an attribute and you could have an ultra simple ReadAll.

```java
  public static class ReadAll<T> extends PTransform<PCollection<Read<T>>, PCollection<T>> {
    @Override
    public PCollection<T> expand(PCollection<Read<T>> input) {
      return input.apply(ParDo.of(new ReadFn<>()));
    }
  }
```

In the end `ReadAll` is more a kind of `Read Split/Partition` (without trying to split), so the only extra thing we need to have for that is the info that allows us to read the specific partition: `ReplicaInfo` in `Solr`,  `segmentId` in `DynamoDB` and `RingRange` in `Cassandra`. Also having the full Read would allow users to read from different systems once the user creates the PCollection<Read> so this is a huge reusability advantage.

Do you see any other tradeoffs, for me it looks quite straight forward and simpler, we won't need to maintain additional objects. Maybe worth trying it.

vmarquez:
@iemejia I think that could work, thanks for your patience as I try to understand what you're thinking.  Some questions:

1.  If our `ReadFn<A> extends DoFn<Read<A>, A>` and the only way we have connection information is from the actual read, that means we need to re-establish a DB connection for each  batch of queries we run?  

2. How would that work for the end user of a `PTransform<PCollection<Read<A>, PCollectoin<A>>`?  Here is what I did in the test and would wand to document how end users could generate 'queries',
https://github.com/apache/beam/pull/10546/files#diff-8ba4ea3b09d563a67e29ff8584269d35R499

Would we instead want to return a `PCollection<Read<A>>` by using something like `return CassandraIO.read().withRingRange(new RingRange(start, finish))`? 

3.  Originally I had wanted to have the ReadFn operate on a *collection* of 'query' objects to ensure a way to enforce linearizability with our queries (mainly so we don't oversaturate a single node/shard).  Currently the groupBy function a user passes in operates on the `RingRange` object, would we keep it that way and just, under the hood, allow for a single `Read<A>` to hold a collection of RingRanges?

iemejia:
> @iemejia I think that could work, thanks for your patience as I try to understand what you're thinking. Some questions:

No, thanks to you who has been the patient one during this discussion.

>     1. If our `ReadFn<A> extends DoFn<Read<A>, A>` and the only way we have connection information is from the `Read<A>` passed in to the processElement, that means we need to re-establish a DB connection for each  batch of queries we run?  As in, the connection would be established in the `processElement` method and could not be in `setup` method?

Yes exactly this will make the method simpler and the cost of starting a connection gets amortized by the processElement producing multiple outputs from a single connection.

>     2. How would that work for the end user of a `PTransform<PCollection<Read<A>, PCollection<A>>`?  Here is what I did in the test and would wand to document how end users could generate 'queries',
>        https://github.com/apache/beam/pull/10546/files#diff-8ba4ea3b09d563a67e29ff8584269d35R499
>        Would we instead want to return a `PCollection<Read<A>>` by using something like `return CassandraIO.read().withRingRange(new RingRange(start, finish))`?  If we do that however, we'd need to do the `withHosts` and all the other connection information, no?  The other option is establishing one `ReadAll` PTransform that maps over the `Read<A>` input and enriches the db connection information?

You have a point here!. We need `class ReadAll extends PTransform<PCollection<Read<A>, PCollection<A>>` and there we read as intended with `ReadFn`. You would have to modify however the `expand` of `Read` to do `input.apply(Create.of(this)).apply(CassandraIO.readAll())` where `ReadAll` should expand into `input.apply(ParDo.of(splitFn)).apply(Reshuffle).apply(Read)` users should deal with building the PCollection of `Reads` before passing that collection to `ReadAll`.

>     3. Originally I had wanted to have the ReadFn operate on a _collection_ of 'query' objects to ensure a way to enforce linearizability with our queries (mainly so we don't oversaturate a single node/shard).  Currently the groupBy function a user passes in operates on the `RingRange` object, would we keep it that way and just, under the hood, allow for a single `Read<A>` to hold a collection of RingRanges?

If I understand this correctly this is covered by following the Create -> Split -> Reshuffle -> Read pattern mentioned above (in the mentioned IOs). So Split is the one who will generate a collection of `Read`s for each given `RingRange` then we use Reshuffle to guarantee that reads are redistributed and finally each read request is read by one worker.

Hope this helps, don't hesitate to ask me more questions if still. I will try to answer quickly this time.
Hope this helps

iemejia:
Forgot to mention that in the above comment that in the Split function you have to split in every case save if the user provided a specific RingRange to read from.

vmarquez:
@iemejia I have started working on a different implementation of ReadAll that extends `PTransform<PCollection<Read<T>>, PCollection<T>>`, however, I have thought about it more and I am thinking that it is bad to not be able to establish a connection in `setup` of the `ReadFn` class.  Just this week at work I had to do some work involving generating hundreds of thousands of queries (`RingRanges`), and the vast majority of them returned no or almost no data!  The data was sparse and never the less I still needed to perform the queries, so having to establish a new connection per `RingRange` in the `processElement` method would have slown things down a lot.  

I understand the desire to have simpler code, but I think the performance impact would be noticable if we only keep connection information in the `Read<T>` that comes into the `processElement` method.   Do you still want to proceed with that?

iemejia:
Cassandra is extremely slow opening new connections however the average case would be that for each `Read` there would be many `T` outputs which in principle would amortize the slow connection time. Couldn't the user in the specific case you mention of many Reads generate queries for RingRanges that match more 'regions' and achieve the same result?

If this is really ends up being a problem one eventual workaround would be to batch requests and create some sort of pool of connections but hopefully this won't be needed.

vmarquez:
> Couldn't the user in the specific case you mention of many Reads generate queries for RingRanges that match more 'regions' and achieve the same result?

 I don't think so, a user might be querying specific partition keys (only one key that returns multiple results) that are not contiguous some of which have data and many of which may have no data.

iemejia:
I think this is needing a rebase. Please ping me when this is ready for review. I am eager to do so hoping this can make it into Beam 2.21.0

iemejia:
I was also thinking about the potential connection issue and I think we can try to tackle it by creating some sort of static connection pool and we id the connections by their server/config and reuse it if possible. We can work on this in this PR or on a follow up, as you prefer.

vmarquez:
Hi @iemejia I've been busy with work and life but was finally able to this just about finished up!  To get around the connection issue (not caching the connection was causing the Cassandra tests to run for ~ten minutes!), by passing in a ReadAll<T> to the ReadFn, we're able to initiate a connection in the setup method, but we can still 'dynamically' use the passed in Read<T> to generate specific queries or query ranges.  

As for a more advanced connection pooling, I'd prefer to get this merged in and then perhaps work on an additional PR.  

There are some minor conflicts but I wasn't sure how you wanted to handle that, do you prefer I rebase against master (? that might then cause issues with Github's UI seeing other folks commits in my changelist), or should I merge master into this then another commit(which makes later squashing a bit harder).  Advice on that would be appreciated!  Thanks again for your guidance and help on this. I like the new design and the flexibility to modify the queries as well as ring ranges is pretty neat.

vmarquez:
@iemejia I've fixed the merge conflicts now with a rebase and I'm ready for further review. Thanks in advance!

iemejia:
retest this please

iemejia:
retest this please

iemejia:
retest this please

iemejia:
bang

iemejia:
retest this please

iemejia:
retest this please

vmarquez:
@iemejia is that a spurious failure or did something I do break the Flink test?  I tested locally and all seems to work... LMK if you need anything from me.

iemejia:
Yes looks like a flaky test there is already a ticket tracking it https://issues.apache.org/jira/browse/BEAM-9534

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
```diff
@@ -318,15 +354,84 @@ private CassandraIO() {}
...
+
+          List<BigInteger> tokens =
+              cluster.getMetadata().getTokenRanges().stream()
+                  .map(tokenRange -> new BigInteger(tokenRange.getEnd().getValue().toString()))
+                  .collect(Collectors.toList());
+
+          SplitGenerator splitGenerator =
+              new SplitGenerator(cluster.getMetadata().getPartitioner());
+
+          List<Read<T>> splits =
```
iemejia, line 405: We should produce a PCollection<Read<T>> here instead and this splitting logic should be moved into the SplitFn function.

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
```diff
@@ -318,15 +354,84 @@ private CassandraIO() {}
...
+                  SplitGenerator.getRangeMin(partitioner), SplitGenerator.getRangeMax(partitioner));
+          return input
+              .apply(Create.of(CassandraIO.<T>read().withRingRange(totalRingRange)))
+              .apply(readAll)
+              .setCoder(coder());
+        }
+      }
+    }
+
+    private static class SplitFn<T> extends DoFn<List<RingRange>, Read<T>> {
```
iemejia, line 428: `private static class SplitFn<T> extends DoFn<Read<T>, Read<T>>`
Notice that in the second Read you should setup the `RingRange` so each individual Read can then be Read by a DoFn that knows how to read `Read` specifications with a RingRange.

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
```diff
@@ -318,15 +354,84 @@ private CassandraIO() {}
...
+      @ProcessElement
+      public void processElement(@Element List<RingRange> input, OutputReceiver<Read<T>> output) {
+        for (RingRange rr : input) {
```
iemejia, line 431: The split logic that is above should be here.

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
```diff
@@ -1170,4 +887,344 @@ private void waitForFuturesToFinish() throws ExecutionException, InterruptedExce
...
+
+  /**
+   * A {@link PTransform} to read data from Apache Cassandra. See {@link CassandraIO} for more
+   * information on usage and configuration.
+   */
+  @AutoValue
+  public abstract static class ReadAll<T> extends PTransform<PCollection<Read<T>>, PCollection<T>> {
+
+    @Nullable
+    abstract ValueProvider<List<String>> hosts();
```
iemejia, line 899: This transform should not have any attribute / with method, those come from the Read in the input PCollection.

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
```diff
@@ -1170,4 +887,344 @@ private void waitForFuturesToFinish() throws ExecutionException, InterruptedExce
...
+
+    @Override
+    public PCollection<T> expand(PCollection<Read<T>> input) {
+      checkArgument((hosts() != null && port() != null), "WithHosts() and withPort() are required");
+      checkArgument(keyspace() != null, "withKeyspace() is required");
+      checkArgument(table() != null, "withTable() is required");
+      checkArgument(entity() != null, "withEntity() is required");
+      checkArgument(coder() != null, "withCoder() is required");
+      checkArgument(groupingFn() != null, "GroupingFn OR splitCount must be set");
+      try (Cluster cluster =
```
iemejia, line 1088: The key point here is to apply for every read the split function and produce subsequent reads with RingRanges that then are Reshuffled and passed to a ParDo with the ReadFn function.

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/ReadFn.java
```diff
@@ -0,0 +1,109 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(ReadFn.class);
+
+  private transient Cluster cluster;
+
+  private transient Session session;
+
+  private final ReadAll<T> readAll;
+
+  public ReadFn(ReadAll<T> readAll) {
+    this.readAll = readAll;
```
iemejia, line 46: You don't need this ReadAll ref here, Read is enough, ReadAll is the one that delegates to a normal Read.

vmarquez: You need this here so we can set up a connection with Setup, this won't work without it.

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
vmarquez: This is more or less leaving the splitting logic the same as it was before, is there a reason we need to change it now?

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
vmarquez: For ReadAll to be useable we need to establish connection information in the `setup` phase of the QueryFn, so we'll need all this before a `Read<T>` comes in.   I'd also like to keep the API between ReadAll and Read similar so it's easy for the end user to switch a `Read<T>` with a `ReadAll<T>`, and just operating on a different type.

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
vmarquez: Not sure I understand what you mean here.  I'm doing a shuffle by key here, `FlattenGrouped` here is what I called it, following the other connectors general pattern.

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/ReadFn.java
vmarquez: If you want I can spend more time getting performance data, but without caching the connection on setup I saw more than two orders of magnitude slowdown on just the `CassandraIOTest`

iemejia:
Just for ref I opened the PR for SolrIO #11357 that introduces ReadAll there.

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/ReadFn.java
iemejia: Caching connections is something we can do in a subsequent PR. Let's get first right the API and then optimize that part. We can cache connections at the class level on the `ReadFn` by creating some size limited pool of connections that we identify by some combination of parameters and we reuse if possible.

vmarquez: @iemejia the tests don't even finish running *locally* after ten minutes when running `CassandraIOTest` if we don't cache the connection.  I have already tried doing it without establishing a connection in `setup`, it's not workable.  The CassandraIOTest returns over *200* partitions to return *10* rows.  That means it has to establish over 20 connections and tear them down per row...

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraConfig.java
```diff
@@ -0,0 +1,88 @@
...
+import org.apache.beam.sdk.options.ValueProvider;
+import org.apache.beam.sdk.transforms.SerializableFunction;
+
+@AutoValue
+abstract class CassandraConfig<T> implements Serializable {
+  @Nullable
+  abstract ValueProvider<List<String>> hosts();
+
+  @Nullable
+  abstract ValueProvider<String> query();
```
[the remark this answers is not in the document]
vmarquez: This class isn't used anymore, I will delete this.  All of the relevant information for a connection is now on the Read being passed in to ReadAll.

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
```diff
@@ -391,36 +393,57 @@ static Cluster getCluster(
...
+            .apply(Create.of(this))
+            .apply(ParDo.of(new SplitFn<T>()))
+            .setCoder(SerializableCoder.of(new TypeDescriptor<Read<T>>() {}))
+            .apply(Reshuffle.viaRandomKey())
+            .apply(readAll);
+      }
+    }
 
-          return input.apply("Creating splits", Create.of(splits)).apply("readAll", readAll);
+    private static class SplitFn<T> extends DoFn<Read<T>, Read<T>> {
```
iemejia, line 405: This looks perfect!

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
```diff
@@ -391,36 +393,57 @@ static Cluster getCluster(
...
-          List<Read<T>> splits =
-              splitGenerator.generateSplits(splitCount, tokens).stream()
-                  .map(rr -> CassandraIO.<T>read().withRingRange(rr))
-                  .collect(Collectors.toList());
+        return input
+            .apply(Create.of(this))
+            .apply(ParDo.of(new SplitFn<T>()))
+            .setCoder(SerializableCoder.of(new TypeDescriptor<Read<T>>() {}))
+            .apply(Reshuffle.viaRandomKey())
+            .apply(readAll);
```
iemejia, line 401: Now we need to tackle this in two parts maybe, one is to implement the read with a `ReadFn` like method  and as a next step to get rid of all the methods on `ReadAll` to simplify it to its core.

iemejia:
retest this please

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
```diff
@@ -1170,4 +898,44 @@ private void waitForFuturesToFinish() throws ExecutionException, InterruptedExce
...
+      return builder().setCoder(coder).build();
+    }
+
+    @Override
+    public PCollection<T> expand(PCollection<Read<T>> input) {
+      checkArgument(coder() != null, "withCoder() is required");
+      return input
+          .apply("shuffle", Reshuffle.viaRandomKey())
+          .apply("read", ParDo.of(new ReadFn<>()))
+          .setCoder(this.coder());
```
iemejia, line 926: coder should come from the input read

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
```diff
@@ -1170,4 +898,44 @@ private void waitForFuturesToFinish() throws ExecutionException, InterruptedExce
...
+  @AutoValue
+  public abstract static class ReadAll<T> extends PTransform<PCollection<Read<T>>, PCollection<T>> {
+
+    @Nullable
+    abstract Coder<T> coder();
+
+    abstract ReadAll.Builder<T> builder();
+
+    /** Specify the {@link Coder} used to serialize the entity in the {@link PCollection}. */
+    public ReadAll<T> withCoder(Coder<T> coder) {
```
iemejia, line 915: I am not sure if we need this, I assume the coder comes form the `Read<T> input.

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
```diff
@@ -326,7 +371,78 @@ private CassandraIO() {}
...
+
+        try (Cluster cluster =
+            getCluster(
+                read.hosts(),
+                read.port(),
+                read.username(),
+                read.password(),
+                read.localDc(),
+                read.consistencyLevel())) {
+          if (isMurmur3Partitioner(cluster)) {
```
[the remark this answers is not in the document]
vmarquez: Oops I did forget that, I will add. Good catch.

sdks/java/io/cassandra/src/test/java/org/apache/beam/sdk/io/cassandra/CassandraIOTest.java
```diff
@@ -480,66 +527,22 @@ public void testCustomMapperImplDelete() {
...
-  }
-
   private List<Row> getRows(String table) {
     ResultSet result =
         session.execute(
             String.format("select person_id,person_name from %s.%s", CASSANDRA_KEYSPACE, table));
     return result.all();
   }
 
+  // TEMP TEST
```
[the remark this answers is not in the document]
vmarquez: Ooops I think this was left in from when I was experimenting with something, I will remove.

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
```diff
@@ -326,7 +371,78 @@ private CassandraIO() {}
...
+            }
+
+            SplitGenerator splitGenerator =
+                new SplitGenerator(cluster.getMetadata().getPartitioner());
+            splitGenerator
+                .generateSplits(splitCount, tokens)
+                .forEach(
+                    rr ->
+                        outputReceiver.output(
+                            CassandraIO.<T>read()
```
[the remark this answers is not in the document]
vmarquez: I think we still need to create a new `Read<T>` because the `SplitGenerator` returns a `List<List<RingRange>>`, so each of the outer list will be a different Read, the inner List will be set to the RingRange.

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
```diff
@@ -370,384 +488,16 @@ private CassandraIO() {}
...
-
-      List<TokenRange> tokenRanges =
-          getTokenRanges(cluster, spec.keyspace().get(), spec.table().get());
-      final long estimatedSize = getEstimatedSizeBytesFromTokenRanges(tokenRanges) / splits.size();
-
-      List<BoundedSource<T>> sources = new ArrayList<>();
-      for (List<RingRange> split : splits) {
-        List<String> queries = new ArrayList<>();
-        for (RingRange range : split) {
-          if (range.isWrapping()) {
```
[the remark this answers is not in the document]
vmarquez: Hm, let me think about this.

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
vmarquez: I think we do, but am not sure.  You have to call `setCoder` on the PCollection itself, so we don't have access to the `Read<T>` at a point when we also have access to a single `Read<T>` (they are only supplied in the `ReadFn` which can't call `setCoder` on the returned PTransform).  Is my thinking correct?  I could be unaware of a different way to do this.

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
```diff
@@ -326,7 +371,78 @@ private CassandraIO() {}
       checkArgument(entity() != null, "withEntity() is required");
       checkArgument(coder() != null, "withCoder() is required");
 
-      return input.apply(org.apache.beam.sdk.io.Read.from(new CassandraSource<>(this, null)));
+      ReadAll<T> readAll = CassandraIO.<T>readAll().withCoder(this.coder());
+
+      return input
+          .apply(Create.of(this))
+          .apply(ParDo.of(new SplitFn()))
```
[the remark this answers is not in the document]
vmarquez: > Move the Split as the first step of the ReadAll expansion so non advanced users (those who do not specify RingRange manually could get their code 'partitioned' correctly.

I am not sure I understand what you mean here, could you explain more?

vmarquez:
@iemejia Apologies for the long delay, I finally had time to deal with the big merge and I believe I have addressed all the issues. Maybe a few nits I missed.  I also realized I needed to make it so readAll could work on a PCollection of reads even without a ringrange, in case I wanted to just have various queries, so now there's even more flexibility in how we can use readAll.  I added an additional test to show that as well.

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
```diff
@@ -1281,4 +1032,44 @@ private void waitForFuturesToFinish() throws ExecutionException, InterruptedExce
...
+    public PCollection<T> expand(PCollection<Read<T>> input) {
+      checkArgument(coder() != null, "withCoder() is required");
+      return input
+          .apply("Reshuffle", Reshuffle.viaRandomKey())
+          .apply("Read", ParDo.of(new ReadFn<>()))
+          .setCoder(this.coder());
+    }
+
+    @AutoValue.Builder
+    abstract static class Builder<T> {
```
iemejia, line 1064: Move this up before the `withCoder` method

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
```diff
@@ -1281,4 +1032,44 @@ private void waitForFuturesToFinish() throws ExecutionException, InterruptedExce
...
+   * A {@link PTransform} to read data from Apache Cassandra. See {@link CassandraIO} for more
+   * information on usage and configuration.
+   */
+  @AutoValue
+  public abstract static class ReadAll<T> extends PTransform<PCollection<Read<T>>, PCollection<T>> {
+
+    @Nullable
+    abstract Coder<T> coder();
+
+    abstract ReadAll.Builder<T> builder();
```
iemejia, line 1046: Remove `ReadAll.`

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
```diff
@@ -373,7 +419,86 @@ private CassandraIO() {}
       checkArgument(entity() != null, "withEntity() is required");
       checkArgument(coder() != null, "withCoder() is required");
 
-      return input.apply(org.apache.beam.sdk.io.Read.from(new CassandraSource<>(this, null)));
+      ReadAll<T> readAll = CassandraIO.readAll();
+
+      return input
+          .apply(Create.of(this))
+          .apply("Split", ParDo.of(new SplitFn()))
```
iemejia, line 426: Can you please move this Split into the ReadAll expand method. When not specified RingRanges like in the `read()` case we should split them on the `ReadAll` expansion. Of course if RingRanges are specified we will probably ignore recalculating them.

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/ReadFn.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+    if (session == null || !reuseSession(lastRead, read)) {
+      this.session = this.cluster.connect(read.keyspace().get());
+    }
+    this.lastRead = read;
+    return this.session;
+  }
+
+  private static <T> boolean reuseCluster(Read<T> readA, Read<T> readB) {
+    return readA != null
```
iemejia, line 111: This is still a bit messy, don't you think we can get something similar by using something like connection pooling and delegate this complexity to the driver (where it should reside)?
https://docs.datastax.com/en/developer/java-driver/2.1/manual/pooling/

Notice that I am not familiar with this but after a quicklook seems to be worth the look.

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/ReadFn.java
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  // TODO: Unit test
+  private static <T> boolean reuseSession(Read<T> readA, Read<T> readB) {
+    return (readA.keyspace() != null && readA.keyspace().equals(readB.keyspace()))
+        || (readA.keyspace() == null && readB.keyspace() == null);
+  }
+
+  /*
+  private static String generateRangeQuery(Read<?> spec, String partitionKey) {
```
iemejia, line 130: Remove if unused

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
```diff
@@ -143,6 +148,36 @@ private CassandraIO() {}
     return Write.<T>builder(MutationType.DELETE).build();
   }
 
+  /** Get a Cassandra cluster using hosts and port. */
+  static Cluster getCluster(
```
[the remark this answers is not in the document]
vmarquez: @iemejia this was done accidentally after the merge.  I will remove it.

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/ReadFn.java
vmarquez: I didn't address this yet, wanted your thoughts on it.  I agree it can be improved. 

The driver will handle Connection *pooling*, (handling N number of connections per cassandra host and sharing them in a thread safe way), but we still need to initiate a `Session` per Cassandra instance, and maintain a reference to the `Session` to keep those connections in scope (under the hood).  

So we could use something as simple as a `Map<?, Cluster>` and `Map<?, Session>`, but remember we have to do both, since each `Read<>` might be hitting an entirely different cassandra cluster.   If we wanted something fancier, we could try a guava LRU or something on the off chance the user is initiating hundreds(thousands/millions?) of different Session objects.

sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java
vmarquez: I don't think it makes sense to move to the readAll...  The readAll takes a Read<>, so on encountering each Read<>, we want to check if there isn't a ringRange or Query set, and then expand it?  I think it is much cleaner here, and I think this is where you originally suggested the split happen?

iemejia:
It is great to know we can still test this in a good scale in practice so a good motivation to finish it. I expect to be a bit more available next week so let's try to push it to the end.

vmarquez:
@iemejia one more thing, regarding the split testing, on the original PR I had removed it, but once we decided on simplifying things to the architecture of `Read<T>` doing the splits and then passing to `ReadAll<T>`, I kept the split logic exactly the same and put the test back.

TheNeuralBit:
Is this ready for @iemejia to review again?

vmarquez:
@TheNeuralBit Yes, thanks for checking.  @iemejia and I were working in his fork with some changes going back and forth, i've merged them here and this should be the up to date PR ready for review again. It is large, so happy to have anyone else take a look as well if you have time.

aromanenko-dev:
Kind ping on this since it hasn't been updated since February. What is a state of this PR?

vmarquez:
@aromanenko-dev I am still waiting any final change requests, but would like to get it merged if there are no further changes.

vmarquez:
@iemejia @aromanenko-dev @TheNeuralBit 

Hello, just checking in to see if this could get reviewed and/or merged anytime soon, it has been open quite a while.  Please let me know if there's anything I can do to help, I don't mind writing more documentation or answering any questions.

vmarquez:
Any reason why that particular test would fail? Passing locally for me.

aromanenko-dev:
Run Java PreCommit

aromanenko-dev:
@vmarquez Sorry for delay with answer, I don't think that this fail is related to your changes. 
I'd leave @iemejia to finish this review since he already was on this.

aromanenko-dev:
Run Java PreCommit

vmarquez:
@aromanenko-dev @iemejia I haven't heard anything in months,  still interested in this contribution?

aromanenko-dev:
@vmarquez Sorry for delay with an answer, this contribution is still very valuable! I hope @iemejia can get back to this PR soon. Otherwise, I'll take it over. 

ping @iemejia

vmarquez:
@aromanenko-dev It's been a couple weeks, is there anything I can do to make this easier to review or help this process out?  LMK if you want more docs/comments/a walk through or anything. 

I can be available on the slack as well to chat in realtime if that would help.

aromanenko-dev:
@vmarquez I'm sorry for delay with a response - I was on holidays and quite busy with Beam Summit in the last weeks. I need some time to refresh my knowledge of this PR and I'll get back to you after.

aaltay:
> @vmarquez I'm sorry for delay with a response - I was on holidays and quite busy with Beam Summit in the last weeks. I need some time to refresh my knowledge of this PR and I'll get back to you after.

@aromanenko-dev - kind ping to see if you have some time.

aromanenko-dev (commented):
@vmarquez Thanks for your work and sorry for delay again. 

I finally managed to take a look on the code and the related discussion to get the main idea of this PR and how it was evolved along the time (it has beed changing pretty well since the beginning). So I expect that @vmarquez and @iemejia came to the final version  of design which was implemented and it should be a good point. Please, let me know if I'm mistaken about this.

Though, I have several questions (sorry if I missed something and it was already discussed before).

1. Can we expect any breaking user API changes?
2. Does this PR contain all necessary tests to test a new functionality and corner cases?
3. @vmarquez Did you try to run it on a real distributed cluster with a real data? If yes, could you provide any details on this (if possible)?
4. Do we expect any performance degradation (or improvement!) for `CassandraIO.read()` since a Read part was mainly rewritten? It would be great to compare.
5. Have been all PR's comments addressed? If not, what is missing?

vmarquez:
> Can we expect any breaking user API changes?

No breaking API changes.

>Does this PR contain all necessary tests to test a new functionality and corner cases?

Yes, the two tests I added test both using RingRanges and Queries.  

>@vmarquez Did you try to run it on a real distributed cluster with a real data? If yes, could you provide any details on this (if possible)?


Sure!  The company I'm working for now  has been running this current version (of my  fork) in production for six months.   We've used it in various dataflow pipelines on a cluster of 5 Scylla servers.   For batch runs we've gotten up to ~3M  elements per second running with hundreds of dataflow worker nodes, for streaming we've only used it on a much smaller sacle, but been doing up to hundreds per second on two nodes. 

>Do we expect any performance degradation (or improvement!) for CassandraIO.read() since a Read part was mainly rewritten? It would be great to compare.

For improvements, I think this basically makes some things possible that just weren't before.  If you have billions of rows it's almost untenable to have to read the entire DB into ram and then filter if you only want a subset.  With readAll you can programatically generate thousands or tens of thousands of queries and send it to readAll instead.  

I don't expect any performance degredation (after I make one final change Ismeal has asked for): Previously the group of ring ranges that were queried were done with async calls.   This doesn't make a lot of sense since in essence it 'undoes' the splitting of the ring ranges (why not just make more splits and have more sync calls?)  

One of the ways we are able to achieve such good performance for large batch jobs with this current change is due to being able to linearize the queries and not overload a specific shard (or core, in Scylla's case).  

>Have been all PR's comments addressed? If not, what is missing?

https://github.com/vmarquez/beam/blob/feature/BEAM-9008/cassandraio_readall/sdks/java/io/cassandra/src/main/java/org/apache/beam/sdk/io/cassandra/CassandraIO.java#L418

@iemejia wanted me to flatten the results here so to mitgiate the change from async to sync DB calls:  The idea being that instead of a single `processElement` call having a number of ring ranges to query (but async), instead we'll have multiple calls to `processElement` each of which is a sync call. 

I agree that is probably most similar to the current behavior and I'm fine making that change.  Happy to hear other thoughts or suggestions.

aromanenko-dev:
> The idea being that instead of a single processElement call having a number of ring ranges to query (but async), instead we'll have multiple calls to processElement each of which is a sync call.

I agree with this change and it looks for me as a more `Beam`-ish way to do such things.

vmarquez:
@aromanenko-dev I've pushed the requested changes.  I also didn't rebase as I thought it'd be easier to review that way, I'm not sure what you prefer.

https://github.com/apache/beam/pull/10546/commits/e72e4d12094aa9fa8d80294b4f9d1617d50a13b3

Let me know and I can squash if you want.

aromanenko-dev:
@vmarquez Thanks! Sounds good - could you squash the commits with a proper commit message, including Jira ID, and rebase it against current `main` branch?

vmarquez:
@aromanenko-dev I squashed then rebased on master, added the JIRA to the commit msg.  Let me know how this looks an anything else I can do.

aromanenko-dev:
Run Java PreCommit

aromanenko-dev:
@vmarquez Please, take a look on `Java PreCommit` [job](https://ci-beam.apache.org/job/beam_PreCommit_Java_Phrase/4017/console) - it fails. `ElasticsearchIOTest.testSplit` fails are caused by flaky tests, so it's not caused by this PR. 

However, there are 2 other failing tasks that we need to fix before merging:
```
15:06:24 1: Task failed with an exception.
15:06:24 -----------
15:06:24 * What went wrong:
15:06:24 Execution failed for task ':sdks:java:io:cassandra:analyzeClassesDependencies'.
15:06:24 > Dependency analysis found issues.
15:06:24   usedUndeclaredArtifacts: 
15:06:24    - com.google.guava:guava:25.1-jre@jar

15:06:24 2: Task failed with an exception.
15:06:24 -----------
15:06:24 * What went wrong:
15:06:24 Execution failed for task ':sdks:java:io:cassandra:checkstyleMain'.
15:06:24 > Checkstyle rule violations were found. See the report at: file:///home/jenkins/jenkins-slave/workspace/beam_PreCommit_Java_Phrase/src/sdks/java/io/cassandra/build/reports/checkstyle/main.html
15:06:24   Checkstyle files with violations: 1
15:06:24   Checkstyle violations by severity: [error:1]
```

aromanenko-dev:
Run Java PreCommit

aromanenko-dev (approved):
@vmarquez Thanks for working on this and finishing this PR! 
LGTM

vmarquez:
@aromanenko-dev thanks so much for your help, I hope I can continue to contribute!

PR #10563, opened 2020-01-13
[BEAM-7310] Add support of Confluent Schema Registry for KafkaIO

aromanenko-dev:
Add `withSchemaRegistry()` which allows to specify Confluent Schema Registry (CSR) url and `withKeySchemaSubject()/withValueSchemaSubject()` to set a schema subject in case if keys and/or values are serialised with Avro schema stored in CSR.

------------------------

--- | --- | --- | --- | ---


aromanenko-dev:

aromanenko-dev:
Run SQL PreCommit

aromanenko-dev:
Run Python PreCommit

aromanenko-dev:

aromanenko-dev:
Run Java PreCommit

aromanenko-dev:

aromanenko-dev:

aromanenko-dev:
@reuvenlax @iemejia could you take a look?

aromanenko-dev:
Run Java PreCommit

iemejia:
CC @rangadi in case you have some time/interest to take a look

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaUnboundedReader.java
```diff
@@ -155,6 +172,25 @@ public boolean start() throws IOException {
     return advance();
   }
 
+  private SchemaRegistryClient getSchemaRegistryClient(Read<K, V> spec) {
+    SchemaRegistryClient registryClient;
+    String schemaRegistryURL =
+        (String)
+            spec.getConsumerConfig().get(AbstractKafkaAvroSerDeConfig.SCHEMA_REGISTRY_URL_CONFIG);
```
iemejia: Can you make `AbstractKafkaAvroSerDeConfig` into `KafkaAvroDeserializerConfig` as below.

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaUnboundedReader.java
```diff
@@ -155,6 +172,25 @@ public boolean start() throws IOException {
...
+
+    if (spec.getSchemaRegistryClientFactoryFn() != null) {
+      registryClient = spec.getSchemaRegistryClientFactoryFn().apply(schemaRegistryURL);
+    } else {
+      registryClient =
+          new CachedSchemaRegistryClient(
+              (String)
```
iemejia: This is `schemaRegistryURL` no?

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
```diff
@@ -817,6 +892,27 @@ public void setValueDeserializer(String valueDeserializer) {
       return new TypedWithoutMetadata<>(this);
     }
 
+    private Schema fetchAvroSchema(String schemaRegistryUrl, String subject) {
+      SchemaRegistryClient registryClient;
```
iemejia: Can you refactor the code of `KafkaUnboundedReader#getSchemaRegistryClient` into a method that you use here to avoid the repeated code to get a client.

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
```diff
@@ -862,19 +958,56 @@ public void setValueDeserializer(String valueDeserializer) {
...
+      Schema avroKeySchema = null;
+      if (schemaRegistryURL != null && getKeySchemaSubject() != null) {
+        avroKeySchema = fetchAvroSchema(schemaRegistryURL, getKeySchemaSubject());
+        checkArgument(avroKeySchema != null, "Avro key schema can't be null");
+      }
+
+      Coder<K> keyCoder;
+      if (avroKeySchema != null) {
+        keyCoder = (Coder<K>) AvroCoder.of(avroKeySchema);
+        checkArgument(
```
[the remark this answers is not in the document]
iemejia: but probably that check should be `checkState` per https://github.com/google/guava/wiki/PreconditionsExplained

sdks/java/io/kafka/src/test/java/org/apache/beam/sdk/io/kafka/KafkaIOTest.java
```diff
@@ -391,6 +451,84 @@ public static void addCountingAsserts(
...
+            "Unable to register schema for subject: " + schemaSubject, e);
+      }
+    }
+  }
+
+  @Test
+  public void testReadAvroKeyValuesWithSchemaRegistry() {
+    int numElements = 100;
+    String topic = "my_topic";
+    String keySubject = topic + "-key";
```
iemejia: Rename to `keySchemaSubject` (same for `valueSchemaSubject`)

sdks/java/io/kafka/src/test/java/org/apache/beam/sdk/io/kafka/KafkaIOTest.java
```diff
@@ -391,6 +451,84 @@ public static void addCountingAsserts(
...
+            + "     {\"name\": \"favorite_color\", \"type\": [\"string\", \"null\"]}\n"
+            + " ]\n"
+            + "}";
+
+    private static final Schema SCHEMA = new Schema.Parser().parse(SCHEMA_STRING);
+
+    private final String keySchemaSubject;
+    private final String valueSchemaSubject;
+
+    private MockSchemaRegistryClientFactoryFn(String keySubject, String valueSubject) {
```
iemejia: `keySchemaSubject` and `valueSchemaSubject`

sdks/java/io/kafka/src/test/java/org/apache/beam/sdk/io/kafka/KafkaIOTest.java
```diff
@@ -160,14 +169,36 @@
...
   // Update mock consumer with records distributed among the given topics, each with given number
   // of partitions. Records are assigned in round-robin order among the partitions.
   private static MockConsumer<byte[], byte[]> mkMockConsumer(
       List<String> topics,
       int partitionsPerTopic,
       int numElements,
       OffsetResetStrategy offsetResetStrategy,
-      Map<String, Object> config) {
+      Map<String, Object> config,
+      boolean isAvro) {
```
iemejia: maybe rename `useAvro` is a more clear name.

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
```diff
@@ -356,6 +366,15 @@
     @Nullable
     abstract Map<String, Object> getOffsetConsumerConfig();
 
+    @Nullable
```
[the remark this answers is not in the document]
iemejia: On the class header (don't know why I pointed it here)

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaUnboundedReader.java
aromanenko-dev: Yes, good catch, will be fixed.

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
```diff
@@ -817,6 +892,27 @@ public void setValueDeserializer(String valueDeserializer) {
...
+      SchemaMetadata latestSchemaMetadata;
+      try {
+        latestSchemaMetadata = registryClient.getLatestSchemaMetadata(subject);
+      } catch (IOException | RestClientException e) {
+        throw new IllegalArgumentException(
+            "Unable to get latest schema metadata for subject: " + subject, e);
+      }
+
+      final Schema avroSchema = new Schema.Parser().parse(latestSchemaMetadata.getSchema());
+      checkArgument(avroSchema != null, "Avro schema can't be null");
```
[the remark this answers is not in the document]
aromanenko-dev: I guess it's almost impossible but I'd prefer to leave this check "just in case" since we can't be sure on 100% and I don't see any disadvantages to have it here.

aromanenko-dev:

aromanenko-dev:

aromanenko-dev:

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
```diff
@@ -586,6 +612,55 @@ public void setValueDeserializer(String valueDeserializer) {
       return toBuilder().setValueDeserializer(valueDeserializer).setValueCoder(valueCoder).build();
     }
 
+    /**
+     * Sets a Confluent Schema Registry URL in case of using Avro schema serialised key/values. It
+     * should be used together with {@link #withKeySchemaSubject(String)} and/or {@link
+     * #withValueSchemaSubject(String)}.
+     */
+    public Read<K, V> withSchemaRegistry(String schemaRegistryUrl) {
```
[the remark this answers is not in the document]
aromanenko-dev: Good question. I think that all different types of authentication can be configured through CSR options and will be used by `CachedSchemaRegistryClient` if any:
https://docs.confluent.io/current/schema-registry/security/index.html

aromanenko-dev:
@iemejia I addressed your comments, PTAL

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
```diff
@@ -175,6 +175,30 @@
...
+ *
+ * <p>For example, below topic values are serialized with Avro schema stored in Schema Registry,
+ * keys are typed as {@link Long}:
+ *
+ * <pre>{@code
+ * PCollection<KafkaRecord<Long, GenericRecord>> input = pipeline
+ *   .apply(KafkaIO.<Long, GenericRecord>read()
+ *      .withBootstrapServers("broker_1:9092,broker_2:9092")
+ *      .withTopic("my_topic")
+ *      .withSchemaRegistry(""http://localhost:8081"") // Confluent Schema Registry URL
```
iemejia: nit: can you make this by default a secure connection in the example e.g. `https`

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
```diff
@@ -913,6 +933,21 @@ private Schema fetchAvroSchema(String schemaRegistryUrl, String subject) {
       return avroSchema;
     }
 
+    static SchemaRegistryClient getSchemaRegistryClient(
+        String schemaRegistryUrl,
+        SerializableFunction<String, SchemaRegistryClient> schemaRegistryClientFactoryFn) {
+      SchemaRegistryClient registryClient;
+      if (schemaRegistryUrl == null) {
```
iemejia: `checkArgument` better, not having a URL to get the client is an Exception condition.

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
```diff
@@ -971,20 +1023,18 @@ private Schema fetchAvroSchema(String schemaRegistryUrl, String subject) {
...
       } else {
         keyCoder =
-            getKeyCoder() != null ? getKeyCoder() : inferCoder(registry, getKeyDeserializer());
-        checkArgument(
-            keyCoder != null,
-            "Key coder could not be inferred from key deserializer. Please provide"
-                + "key coder explicitly using withKeyDeserializerAndCoder()");
+            getKeyCoder() != null ? getKeyCoder() : inferCoder(coderRegistry, getKeyDeserializer());
       }
+      checkArgument(
```
iemejia: `checkState` better, we should use `checkArgument` only for arguments.

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
```diff
@@ -994,34 +1044,17 @@ private Schema fetchAvroSchema(String schemaRegistryUrl, String subject) {
...
-
-      PTransform<PBegin, PCollection<KafkaRecord<K, V>>> transform = unbounded;
-
-      if (getMaxNumRecords() < Long.MAX_VALUE || getMaxReadTime() != null) {
-        transform =
-            unbounded.withMaxReadTime(getMaxReadTime()).withMaxNumRecords(getMaxNumRecords());
-      }
-
-      return input.getPipeline().apply(transform);
+      checkArgument(
```
iemejia: `checkState` better, we should use `checkArgument` only for arguments.

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaUnboundedReader.java
```diff
@@ -91,22 +90,34 @@ public boolean start() throws IOException {
...
+        (String)
+            spec.getConsumerConfig().get(KafkaAvroDeserializerConfig.SCHEMA_REGISTRY_URL_CONFIG);
+    SchemaRegistryClient registryClient =
+        KafkaIO.Read.getSchemaRegistryClient(
+            schemaRegistryURL, spec.getSchemaRegistryClientFactoryFn());
+
     try {
       if (spec.getKeyDeserializer().equals(KafkaAvroDeserializer.class)) {
-        SchemaRegistryClient registryClient = getSchemaRegistryClient(spec);
+        if (registryClient == null) {
```
iemejia: Remove validation per above comment on checkArgument, if we cannot build a client we fail.

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaUnboundedReader.java
```diff
@@ -91,22 +90,34 @@ public boolean start() throws IOException {
...
+              "Schema registry client can't be null when KafkaAvroDeserializer is used for key.");
+        }
         keyDeserializerInstance = (Deserializer) new KafkaAvroDeserializer(registryClient);
       } else {
         keyDeserializerInstance = spec.getKeyDeserializer().getDeclaredConstructor().newInstance();
       }
 
       if (spec.getValueDeserializer().equals(KafkaAvroDeserializer.class)) {
-        SchemaRegistryClient registryClient = getSchemaRegistryClient(spec);
+        if (registryClient == null) {
```
iemejia: Remove too (see above)

rangadi:
Thanks for the PR @aromanenko-dev & @iemejia for thorough review. Overall LGTM. Just one thing to consider in naming: the doc comments always mention 'Confluent Schema Registry', probably implying this is specific to Confluent. Should the API names include `Confluent` or `CSR` in them too?

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
aromanenko-dev: I'd leave this example as simple as possible. All configuration options for CSR can be found in its documentation

aromanenko-dev:
@iemejia I addressed your latest comments , PTAL

aromanenko-dev:
@rangadi Thank you for feedback! I agree with your naming proposal despite that afaik only one Avro Schema Registry exists (Confluent's one). In the same time, to avoid a confusion with Beam schema we should probably name it explicitly `withConfluentSchemaRegistry()`. Do you think we need to name other new public methods in this way, like `getKeySchemaSubject()` and `getValueSchemaSubject()` should be `getKeyConfluentSchemaSubject()` and `getValueConfluentSchemaSubject()` ? 

@rangadi @iemejia Also, now I'm thinking to reduce the number of new public methods, introduced by this PR, and to add only `withConfluentSchemaRegistry(ConfluentSchemaRegistryProvider)` where `ConfluentSchemaRegistryProvider` will incorporate all required members (registry URL, subjects, ClientFactoryFn) to use CSR in `KafkaIO`, like we do for KinesisIO with `AWSClientsProvider`. Wdyt?

iemejia:
Thanks for changes code looks ok for me :+1: 
About your proposals you have a valid point about Beam's schema in the naming (hard to reconcile): 
+1 for `withConfluentSchemaRegistry`
0 for getKeyConfluentSchemaSubject and getValueConfluentSchemaSubject 
The idea of the new object to have only one new signature sounds GORGEOUS to me please PLEASE !
Please tell me when can I take a look again, and thanks for your patience during the review.

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
```diff
@@ -891,6 +1014,50 @@ public void setValueDeserializer(String valueDeserializer) {
...
+            getValueCoder() != null
+                ? getValueCoder()
+                : inferCoder(coderRegistry, getValueDeserializer());
+      }
+      checkState(
+          valueCoder != null,
+          "Value coder could not be inferred from value deserializer. Please provide"
+              + "value coder explicitly using withValueDeserializerAndCoder()");
+      return valueCoder;
+    }
```
reuvenlax: It would be very useful to set a Beam schema based on the Avro schema instead of using AvroCoder. This would allow KafkaIO to be used as a source for Beam SQL

iemejia: Excellent idea, I would suggest to use a more conservative approach to enable Beam Schema Coders similar to the one used for `set/getInferBeamSchema` in `AvroIO`.

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
```diff
@@ -165,6 +175,30 @@
...
+ *
+ * <p>For example, below topic values are serialized with Avro schema stored in Schema Registry,
+ * keys are typed as {@link Long}:
+ *
+ * <pre>{@code
+ * PCollection<KafkaRecord<Long, GenericRecord>> input = pipeline
+ *   .apply(KafkaIO.<Long, GenericRecord>read()
+ *      .withBootstrapServers("broker_1:9092,broker_2:9092")
+ *      .withTopic("my_topic")
+ *      .withSchemaRegistry(""http://localhost:8081"") // Confluent Schema Registry URL
```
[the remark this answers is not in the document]
aromanenko-dev: @reuvenlax In multiple topics case we expect that all topics follow the same schema structure which is stored in Confluent Schema Registry. This is like what we do with deserialisers (and coders) in such case - we expect that all messages from different topics can be deserialised by the same deserialiser. I was thinking to create a kind of map for topic->schema relations and deserialise it accordingly, but there are two objections against that:
- we will need to write a wrapper around `KafkaAvroDeserializer` to support multiple topics with multiple schemas there, which is probably not a big deal but having an output PCollection of `GenericRecord`s created based on different schemas seems quite complex.
- afaik, we can set only one single Beam schema for PCollection. So what we should do in multiple schemas case? 

As for now, I think we need to clear explain it in Javadoc and leave it as a user responsibility to choose a proper schema for all required input topic(s) that satisfy it.

Please, let me know if you have other ideas or examples about that, I'd be happy to discuss them to find a best solution.

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
```diff
@@ -586,6 +636,55 @@ public void setValueDeserializer(String valueDeserializer) {
...
+          ImmutableMap.of(
+              KafkaAvroDeserializerConfig.SCHEMA_REGISTRY_URL_CONFIG, schemaRegistryUrl));
+    }
+
+    /**
+     * Sets a subject name to obtain Avro schema from Confluent Schema Registry used for key
+     * serialisation. Also, sets {@link KafkaAvroDeserializer} as a key deserializer. {@link Coder}
+     * will be inferred based on obtained Avro schema by {@link AvroCoder}. It should be used along
+     * with {@link #withSchemaRegistry(String)}.
+     */
```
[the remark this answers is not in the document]
aromanenko-dev: Please, see my comment above.

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
aromanenko-dev: I planned to do this in the next PR for https://issues.apache.org/jira/browse/BEAM-7336

aromanenko-dev:
As we discussed before, I simplified KafkaIO API to use it with Confluent Schema Registry. I added a `CSRClientProvider` interface and two implementations - `BasicCSRClientProvider` (used by default in KafkaIO) and `TestCSRClientProvider` (for unit tests). The instance of provider should create and return the instance of `SchemaRegistryClient` which is used to fetch Avro schema from CSR.

Also, I added several another tests to test reading of key and values if they Avro serialised in different combinations.

I'm still on the subject if we need to support different schemas from different CSR subjects if user reads from multiple topics. Any ideas will be very welcomed.

aromanenko-dev:
Run Java KafkaIO Performance Test

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
```diff
@@ -586,6 +628,40 @@ public void setValueDeserializer(String valueDeserializer) {
       return toBuilder().setValueDeserializer(valueDeserializer).setValueCoder(valueCoder).build();
     }
 
+    /**
+     * Allows to specify custom {@link CSRClientProvider}. {@link CSRClientProvider} provides {@link
+     * SchemaRegistryClient} instance which is used later for communication with Confluent Schema
+     * Registry. You should use this method if {@link Read#withCSRClientProvider(String, String,
+     * String)} does not suit your needs.
+     */
+    public Read<K, V> withCSRClientProvider(CSRClientProvider csrClientProvider) {
```
iemejia: I would make this one package private until someone really asks for it.

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/BasicCSRClientProvider.java
```diff
@@ -0,0 +1,56 @@
...
+ * limitations under the License.
+ */
+package org.apache.beam.sdk.io.kafka;
+
+import io.confluent.kafka.schemaregistry.client.CachedSchemaRegistryClient;
+import io.confluent.kafka.schemaregistry.client.SchemaRegistryClient;
+import javax.annotation.Nullable;
+
+/** Basic implementation of {@link CSRClientProvider} used by default in {@link KafkaIO}. */
+public class BasicCSRClientProvider implements CSRClientProvider {
```
iemejia: package private, see my comment on `withCSRClientProvider` below

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
```diff
@@ -165,6 +173,35 @@
...
+ * org.apache.avro.generic.GenericRecord}. In this case, users don't need to specify key or/and
+ * value deserializers and coders since they will be set to {@link KafkaAvroDeserializer} and {@link
+ * AvroCoder} by default accordingly.
+ *
+ * <p>For example, below topic values are serialized with Avro schema stored in Schema Registry,
+ * keys are typed as {@link Long}:
+ *
+ * <pre>{@code
+ * // Confluent Schema Registry URL
+ * String schemaRegistryUrl = "http://localhost:8081";
```
iemejia: I would better let the documentation with the 3-arg method, prefer simpler use, we can easily evolve towards the more generic case if there is a need which so far seems not to be the case.

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/BasicCSRClientProvider.java
```diff
@@ -0,0 +1,56 @@
...
+
+  public BasicCSRClientProvider(
+      String schemaRegistryUrl, String keySchemaSubject, String valueSchemaSubject) {
+    this.schemaRegistryUrl = schemaRegistryUrl;
+    this.keySchemaSubject = keySchemaSubject;
+    this.valueSchemaSubject = valueSchemaSubject;
+  }
+
+  @Override
+  public String getSchemaRegistryUrl() {
```
iemejia: Remove, see comment on the interface below

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
```diff
@@ -586,6 +628,40 @@ public void setValueDeserializer(String valueDeserializer) {
...
+          .build()
+          .withConsumerConfigUpdates(
+              ImmutableMap.of(
+                  KafkaAvroDeserializerConfig.SCHEMA_REGISTRY_URL_CONFIG,
+                  csrClientProvider.getSchemaRegistryUrl()));
+    }
+
+    /**
+     * Specify Confluent schema registry url, key and value schema subjects to be used to read
+     * schema from Confluent Schema Registry. If you need more sophisticated credential protocol,
```
iemejia: Remove the comment `"If you need more sophisticated version..."` for consistency with non public exposure of the related method.

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
```diff
@@ -817,6 +893,20 @@ public void setValueDeserializer(String valueDeserializer) {
       return new TypedWithoutMetadata<>(this);
     }
 
+    private Schema fetchAvroSchema(CSRClientProvider csrClientProvider, String subject) {
```
iemejia: static

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
```diff
@@ -817,6 +893,20 @@ public void setValueDeserializer(String valueDeserializer) {
...
+      SchemaMetadata latestSchemaMetadata;
+      try {
+        latestSchemaMetadata = csrClientProvider.getCSRClient().getLatestSchemaMetadata(subject);
+      } catch (IOException | RestClientException e) {
+        throw new IllegalArgumentException(
+            "Unable to get latest schema metadata for subject: " + subject, e);
+      }
+
+      final Schema avroSchema = new Schema.Parser().parse(latestSchemaMetadata.getSchema());
+      checkArgument(avroSchema != null, "Avro schema can't be null");
```
iemejia: `checkState`

sdks/java/io/kafka/src/test/java/org/apache/beam/sdk/io/kafka/KafkaAvroGenericRecordExample.java
```diff
@@ -0,0 +1,53 @@
...
+
+import org.apache.avro.generic.GenericRecord;
+import org.apache.beam.sdk.Pipeline;
+import org.apache.beam.sdk.options.PipelineOptions;
+import org.apache.beam.sdk.options.PipelineOptionsFactory;
+import org.apache.beam.sdk.transforms.DoFn;
+import org.apache.beam.sdk.transforms.ParDo;
+import org.apache.kafka.common.serialization.StringDeserializer;
+
+public class KafkaAvroGenericRecordExample {
```
iemejia: I have mixed feelings about this one with a tendency towards remove it. I like the idea, but at the same time is like more code to maintain (that is not tested). Are we doing this for other IOs or other parts of the codebase.

sdks/java/io/kafka/src/test/java/org/apache/beam/sdk/io/kafka/KafkaIOTest.java
```diff
@@ -1784,4 +1966,75 @@ void shutdown() {
       }
     }
   }
+
+  private static class AvroDeSe {
```
iemejia: better call it `AvroDeSer` if it is a game of words around `AvroSerDe` or even better just call it `AvroSerDe`

sdks/java/io/kafka/src/test/java/org/apache/beam/sdk/io/kafka/KafkaIOTest.java
```diff
@@ -1784,4 +1966,75 @@ void shutdown() {
...
+    static Serializer<AvroGeneratedUser> getSerializer(boolean isKey, String schemaRegistryUrl) {
+      if (serializer == null) {
+        serializer = AvroDeSe.getSerializer(isKey, schemaRegistryUrl);
+      }
+      return serializer;
+    }
+  }
+
+  private static class KeyAvroSerializableFunction extends BaseAvroSerializableFunction {
+    KeyAvroSerializableFunction(String topic, String schemaRegistryUrl, boolean isKey) {
```
iemejia: remove `isKey` and fix it in the call to super to `true`

sdks/java/io/kafka/src/test/java/org/apache/beam/sdk/io/kafka/KafkaIOTest.java
```diff
@@ -1784,4 +1966,75 @@ void shutdown() {
...
+
+    @Override
+    public byte[] apply(Integer i) {
+      return getSerializer(isKey, schemaRegistryUrl)
+          .serialize(topic, new AvroGeneratedUser("KeyName" + i, i, "color" + i));
+    }
+  }
+
+  private static class ValueAvroSerializableFunction extends BaseAvroSerializableFunction {
+    ValueAvroSerializableFunction(String topic, String schemaRegistryUrl, boolean isKey) {
```
iemejia: remove `isKey` and fix it in the call to super to `false`

sdks/java/io/kafka/src/test/java/org/apache/beam/sdk/io/kafka/KafkaIOTest.java
```diff
@@ -391,6 +427,152 @@ public static void addCountingAsserts(
...
+                new TestCSRClientProvider(schemaRegistryUrl, keySchemaSubject, valueSchemaSubject));
+
+    PCollection<KV<GenericRecord, GenericRecord>> input = p.apply(reader.withoutMetadata());
+
+    PAssert.that(input).containsInAnyOrder(inputs);
+    p.run();
+  }
+
+  @Test
+  public void testReadAvroValuesAsPojoRecords() {
```
iemejia: `s/PojoRecord/SpecificRecords` by Avro common parlance

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
```diff
@@ -891,6 +970,52 @@ public void setValueDeserializer(String valueDeserializer) {
...
 
+    private Coder<K> getKeyCoder(CoderRegistry coderRegistry) {
+      Coder<K> keyCoder;
+      CSRClientProvider csrClientProvider = getCSRClientProvider();
+      if (csrClientProvider != null && getKeyDeserializer().equals(KafkaAvroDeserializer.class)) {
+        Schema avroKeySchema = null;
+        String schemaSubject = csrClientProvider.getKeySchemaSubject();
+        if (schemaSubject != null) {
+          avroKeySchema = fetchAvroSchema(csrClientProvider, schemaSubject);
+          checkArgument(avroKeySchema != null, "Avro key schema can't be null");
```
iemejia: `checkState`

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
```diff
@@ -891,6 +970,52 @@ public void setValueDeserializer(String valueDeserializer) {
...
+
+    private Coder<V> getValueCoder(CoderRegistry coderRegistry) {
+      Coder<V> valueCoder;
+      CSRClientProvider csrClientProvider = getCSRClientProvider();
+      if (csrClientProvider != null && getValueDeserializer().equals(KafkaAvroDeserializer.class)) {
+        Schema avroValueSchema = null;
+        String schemaSubject = csrClientProvider.getValueSchemaSubject();
+        if (schemaSubject != null) {
+          avroValueSchema = fetchAvroSchema(csrClientProvider, schemaSubject);
+          checkArgument(avroValueSchema != null, "Avro value schema can't be null");
```
iemejia: `checkState`

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/TestCSRClientProvider.java
```diff
@@ -0,0 +1,64 @@
...
+package org.apache.beam.sdk.io.kafka;
+
+import io.confluent.kafka.schemaregistry.client.MockSchemaRegistryClient;
+import io.confluent.kafka.schemaregistry.client.SchemaRegistryClient;
+import io.confluent.kafka.schemaregistry.client.rest.exceptions.RestClientException;
+import java.io.IOException;
+import org.apache.avro.Schema;
+
+/** Mock implementation of {@link CSRClientProvider} used for testing. */
+public class TestCSRClientProvider extends BasicCSRClientProvider {
```
iemejia: Move this file better `src/test/java` and make the class package private

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/CSRClientProvider.java
```diff
@@ -0,0 +1,38 @@
...
+ * Provides instances of Confluent Schema Registry clients.
+ *
+ * <p>Please note, that any instance of {@link CSRClientProvider} must be {@link Serializable} to
+ * ensure it can be sent to worker machines.
+ */
+public interface CSRClientProvider extends Serializable {
+
+  SchemaRegistryClient getCSRClient();
+
+  String getSchemaRegistryUrl();
```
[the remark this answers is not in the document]
aromanenko-dev: We can't remove it since `SchemaRegistryClient` doesn't return `schemaRegistryUrl`, do it?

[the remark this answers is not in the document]
iemejia: `registryUrl` is an input parameter, not an output one. The user already knows the URL because he is building the client so that side is covered, and KafkaIO does not care about the `registryUrl` because it already has the built client. Or am I missing something?

sdks/java/io/kafka/src/test/java/org/apache/beam/sdk/io/kafka/KafkaAvroGenericRecordExample.java
aromanenko-dev: You are right, I'll remove it. Though, we will need to add this a part of KafkaIO IT later.

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/CSRClientProvider.java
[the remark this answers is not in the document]
aromanenko-dev: We need to set Kafka config option "schema.registry.url" with this value and we do this in `withCSRClientProvider(CSRClientProvider)`. Unfortunately, `SchemaRegistryClient` doesn't have an API method to return `schemaRegistryUrl`, so we use the one stored in `CSRClientProvider`. What is an issue to have `String getSchemaRegistryUrl()` there?

aromanenko-dev:
Run Java KafkaIO Performance Test

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/CSRClientProvider.java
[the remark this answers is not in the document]
iemejia: Oh I see my mistake, sorry. Ugly `Map based APIs`. Ok then, but to the history I want to say that this is information leaking ! :stuck_out_tongue:

iemejia:
Merged manually to squash the review commits.

suztomo:
@aromanenko-dev @iemejia "SQL Postcommit" started failing after this commit. Would you check whether the failure is related to this PR?

image

https://builds.apache.org/job/beam_PostCommit_SQL/3924/

```
org.apache.beam.sdk.extensions.sql.meta.provider.datacatalog.DataCatalogGCSIT.testReadFromGCS


java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.Long
```

udim:
Opened https://issues.apache.org/jira/browse/BEAM-9251 for breakage of updateOfflineRepository

iemejia:
Thanks a lot @suztomo and @udim for taking care of this (time zones did not help for us to look),  a pity we do not have a check to detect such cases earlier.

suztomo:
I don’t know udim’s problem is related to my finding. SQL Postcommit is still red:
https://builds.apache.org/job/beam_PostCommit_SQL/

aromanenko-dev:
@suztomo Thanks, yes, it's still red. Do you know how to run this test locally?

suztomo:
No, I don't. Let me send an email to dev mailing list.

https://gist.github.com/suztomo/43de4ff1f0458e801fef1a16a28da301

aromanenko-dev:
@suztomo Thanks. I don't see a direct relation between this failing test (DataCatalogGCSIT) and KafkaIO. Only guess for now is that this PR could bring some transitive deps and it fails somehow because of this....
Btw, the full stacktrace when I run it locally:
https://gist.github.com/aromanenko-dev/0916b6d4f618a2c1b4f7b8ac5bf320ad
However, it's a different error that we have on Jenkins. Unfortunately, I can't see the details of Jenkins job since it's published into local file.

suztomo:
https://issues.apache.org/jira/browse/BEAM-9253

suztomo:
For BEAM-9253, Brian observes the data in the backend has changed on Feb 4th, and thus this PR is unrelated. Sorry for confusion.

aromanenko-dev:
@suztomo Thank you for taking care of this.

PR #10564, opened 2020-01-13
[BEAM-8740] Remove unused dependency from Spark runner

iemejia:
R: @aromanenko-dev

### DIFF
--- a/sdks/java/core/src/test/java/org/apache/beam/sdk/transforms/ParDoTest.java
+++ b/sdks/java/core/src/test/java/org/apache/beam/sdk/transforms/ParDoTest.java
@@ -3736,6 +3736,7 @@ public class ParDoTest implements Serializable {
       ValidatesRunner.class,
       UsesStatefulParDo.class,
       UsesTimersInParDo.class,
+      UsesTestStream.class,
       UsesTestStreamWithOutputTimestamp.class
     })
     public void testOutputTimestamp() {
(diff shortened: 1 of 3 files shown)


iemejia:

iemejia:
Run Spark ValidatesRunner

iemejia:
Run Spark ValidatesRunner

iemejia:

sdks/java/core/src/test/java/org/apache/beam/sdk/transforms/ParDoTest.java
```diff
@@ -3736,6 +3736,7 @@ public void testTwoTimersSettingEachOtherWithCreateAsInput() {
       ValidatesRunner.class,
       UsesStatefulParDo.class,
       UsesTimersInParDo.class,
+      UsesTestStream.class,
```
iemejia: I had to add this one because it was not correctly categorized and VR were breaking

iemejia: It seems is reported here
https://issues.apache.org/jira/browse/BEAM-9083

lukecwik: Shouldn't the VR tests exclude UsesTestStreamWithOutputTimestamp

iemejia: That's the line exactly below that one. Notice that Spark runner does not implement TestStream so we probably need to make the test belong to both categories.

iemejia: Mind to maybe lgtm / merge this one @lukecwik. Seems most VR are broken because of it.

lukecwik: I see, I was under the impression that we would place things in the most appropriate category but now thinking about it makes sense to place it in this category as well.

PR #10565, opened 2020-01-13
Make SerializableUtils.deserializeFromByteArray generic

iemejia:
Did this little improvement as part of the ongoing work to get rid of commons-lang3 `SerializationUtils` but I prefer to keep it separated because this may introduce much noise in the other review (once finished). 
R: @lukecwik


iemejia:
Oh this seems to be a bad practice http://errorprone.info/bugpattern/TypeParameterUnusedInFormals
What a bummer it avoided so much repetitive casts. Closing then.

PR #10581, opened 2020-01-14
[website] Add security page and update 2.17.0 release blog post to reference it

iemejia:
R: @kennknowles

### DIFF
--- a/website/src/_includes/section-menu/get-started.html
+++ b/website/src/_includes/section-menu/get-started.html
@@ -30,4 +30,4 @@
    </ul>
 </li>
 <li><a href="{{ site.baseurl }}/get-started/downloads">Downloads</a></li>
-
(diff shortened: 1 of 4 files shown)


iemejia:
Review URL https://apache-beam-website-pull-requests.storage.googleapis.com/10581/index.html

iemejia:
Actually they ask for the link in the report to MITRE so we need a 'longstanding' URL. I put it in 'Get Started" just because that's where the download is so people can find the information there, but i can put it in other section if you suggest me one that fits better.

kennknowles:
I see. For a longstanding URL we should set up a super canonical alias.

kennknowles:
Can you hack it to be beam.apache.org/CVE-2020-1929 ?

iemejia:
ok let me try if I can get to do the mapping, but better beam.apache.org/security/CVE-2020-1929

PR #10637, opened 2020-01-19
[BEAM-9186] Allow injection of custom equality function.

sorensenjs:
Testing complex object equality in beam tests is currently hard. This allows constructions like

```
assert_that(p | Create([1, 2, 3]), equal_to(
    ['1', '2', '3'], equals_fn=lambda e, a: int(e) == int(a)))
```

### DIFF
--- a/sdks/python/apache_beam/testing/util.py
+++ b/sdks/python/apache_beam/testing/util.py
@@ -154,41 +154,50 @@ def equal_to_per_window(expected_window_to_elements):
 # Note that equal_to checks if expected and actual are permutations of each
 # other. However, only permutations of the top level are checked. Therefore
 # [1,2] and [2,1] are considered equal and [[1,2]] and [[2,1]] are not.
-def equal_to(expected):
+def equal_to(expected, equals_fn=None):
+  equal_to.equals_fn=equals_fn
 
   def _equal(actual):
     expected_list = list(expected)
+    equals_fn = equal_to.equals_fn
 
     # Try to compare actual and expected by sorting. This fails with a
     # TypeError in Python 3 if different types are present in the same
     # collection. It can also raise false negatives for types that don't have
     # a deterministic sort order, like pyarrow Tables as of 0.14.1
-    try:
-      sorted_expected = sorted(expected)
-      sorted_actual = sorted(actual)
-      if sorted_expected != sorted_actual:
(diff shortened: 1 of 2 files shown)


sorensenjs:
Seems potentially related to https://issues.apache.org/jira/browse/BEAM-2970 although that issue seems already addressed.

tvalentyn:
Run Python PreCommit

sdks/python/apache_beam/testing/util.py
```diff
@@ -154,41 +154,50 @@ def equal_to_per_window(expected_window_to_elements):
 # Note that equal_to checks if expected and actual are permutations of each
 # other. However, only permutations of the top level are checked. Therefore
 # [1,2] and [2,1] are considered equal and [[1,2]] and [[2,1]] are not.
-def equal_to(expected):
+def equal_to(expected, equals_fn=None):
+  equal_to.equals_fn=equals_fn
```
tvalentyn: This creates a side-effect, making `equal_to` not thread-safe. Can we access `equals_fn` from the outer scope as we do with `expected` ?

sorensenjs: Yes, I tried that but it indeed does not work. PTAL and the alternative I just committed.

sdks/python/apache_beam/testing/util.py
```diff
@@ -154,41 +154,48 @@ def equal_to_per_window(expected_window_to_elements):
...
-      if sorted_expected != sorted_actual:
-        raise BeamAssertException(
-            'Failed assert: %r == %r' % (sorted_expected, sorted_actual))
+    if not equals_fn:
+      try:
+        sorted_expected = sorted(expected)
+        sorted_actual = sorted(actual)
+        if sorted_expected == sorted_actual:
+          return
+      except (BeamAssertException, TypeError):
```
aaltay: We can remove `BeamAssertException` from here. The remaining code in the block would not raise that exception.

tvalentyn:
Run Python PreCommit

tvalentyn:
Run PythonLint Precommit

tvalentyn:
Run Python PreCommit

tvalentyn:
Run PythonLint Precommit

tvalentyn:
Thanks, @sorensenjs !

PR #10644, opened 2020-01-21
[BEAM-7427] Refactor JmsCheckpointMark to use SerializableCoder

jbonofre:
This refactoring uses JmsCheckpointMark coder compatible.
------------------------

 - [X] [**Choose reviewer(s)**](https://beam.apache.org/contribute/#make-your-change) and mention them in a comment (`R: @username`).
 - [X] Format the pull request title like `[BEAM-XXX] Fixes bug in ApproximateQuantiles`, where you replace `BEAM-XXX` with the appropriate JIRA issue, if applicable. This will automatically link the pull request to the issue.
 - [X] If this contribution is large, please file an Apache [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/io/jms/src/main/java/org/apache/beam/sdk/io/jms/JmsCheckpointMark.java
+++ b/sdks/java/io/jms/src/main/java/org/apache/beam/sdk/io/jms/JmsCheckpointMark.java
@@ -17,48 +17,55 @@
  */
 package org.apache.beam.sdk.io.jms;
 
+import java.io.IOException;
+import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
-import java.util.function.BiFunction;
-import java.util.function.Supplier;
 import javax.jms.Message;
-import org.apache.beam.sdk.coders.AvroCoder;
-import org.apache.beam.sdk.coders.DefaultCoder;
 import org.apache.beam.sdk.io.UnboundedSource;
-import org.apache.beam.sdk.transforms.windowing.BoundedWindow;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.annotations.VisibleForTesting;
 import org.joda.time.Instant;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * Checkpoint for an unbounded JmsIO.Read. Consists of JMS destination name, and the latest message
- * ID consumed so far.
+ * Checkpoint for an unbounded JMS source. Consists of the JMS messages waiting to be acknowledged
+ * and oldest pending message timestamp.
  */
-@DefaultCoder(AvroCoder.class)
-public class JmsCheckpointMark implements UnboundedSource.CheckpointMark {
+@VisibleForTesting
+public class JmsCheckpointMark implements UnboundedSource.CheckpointMark, Serializable {
 
   private static final Logger LOG = LoggerFactory.getLogger(JmsCheckpointMark.class);
 
-  private final State state = new State();
+  @VisibleForTesting Instant oldestMessageTimestamp = Instant.now();
+  @VisibleForTesting transient List<Message> messages = new ArrayList<>();
+
+  private final ReentrantReadWriteLock lock = new ReentrantReadWriteLock();
 
   public JmsCheckpointMark() {}
 
-  protected List<Message> getMessages() {
-    return state.getMessages();
+  public void add(Message message) throws Exception {
+    lock.writeLock().lock();
+    try {
+      Instant currentMessageTimestamp = new Instant(message.getJMSTimestamp());
+      if (currentMessageTimestamp.isBefore(oldestMessageTimestamp)) {
+        oldestMessageTimestamp = currentMessageTimestamp;
+      }
+      messages.add(message);
+    } finally {
+      lock.writeLock().unlock();
+    }
   }
 
-  protected void addMessage(Message message) throws Exception {
-    Instant currentMessageTimestamp = new Instant(message.getJMSTimestamp());
-    state.atomicWrite(
-        () -> {
-          state.updateOldestPendingTimestampIf(currentMessageTimestamp, Instant::isBefore);
-          state.addMessage(message);
-        });
-  }
-
-  protected Instant getOldestPendingTimestamp() {
-    return state.getOldestPendingTimestamp();
+  public Instant getOldestMessageTimestamp() {
+    lock.readLock().lock();
+    try {
+      return this.oldestMessageTimestamp;
+    } finally {
+      lock.readLock().unlock();
+    }
   }
 
   /**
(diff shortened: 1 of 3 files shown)


jbonofre:
R: @iemejia @aromanenko-dev

sdks/java/io/jms/src/test/java/org/apache/beam/sdk/io/jms/JmsIOTest.java
```diff
@@ -406,6 +408,26 @@ public void testCheckpointMarkSafety() throws Exception {
...
+    JmsIO.JmsCheckpointMark jmsCheckpointMark = new JmsIO.JmsCheckpointMark();
+    jmsCheckpointMark.add(new ActiveMQMessage());
+    Coder coder = new JmsIO.UnboundedJmsSource(null).getCheckpointMarkCoder();
+    CoderProperties.coderSerializable(coder);
+    CoderProperties.coderDecodeEncodeEqual(coder, jmsCheckpointMark);
+  }
+
+  @Test
+  public void testCheckpointMarkSerializableCoder() throws Exception {
+    JmsIO.JmsCheckpointMark jmsCheckpointMark = new JmsIO.JmsCheckpointMark();
```
iemejia: This second test is not needed if you make `UnboundedJmsSource#getCheckpointMarkCoder` return the `SerializableCoder`. Also that is needed to get rid also of `AvroCoder` as part of this fixes.

sdks/java/io/jms/src/main/java/org/apache/beam/sdk/io/jms/JmsIO.java
```diff
@@ -404,9 +408,96 @@ private JmsIO() {}
...
+        throws IOException, ClassNotFoundException {
+      stream.defaultReadObject();
+      messages = new ArrayList<>();
+    }
+
+    @Override
+    public boolean equals(Object other) {
+      if (other instanceof JmsCheckpointMark) {
+        JmsCheckpointMark that = (JmsCheckpointMark) other;
+        return Objects.equals(this.oldestMessageTimestamp, that.oldestMessageTimestamp);
```
iemejia: We need to add messages to the equals too.

sdks/java/io/jms/src/main/java/org/apache/beam/sdk/io/jms/JmsIO.java
```diff
@@ -404,9 +408,96 @@ private JmsIO() {}
...
+   * Checkpoint for an unbounded JMS source. Consists of the JMS messages waiting to be acknowledged
+   * and oldest pending message timestamp.
+   */
+  @VisibleForTesting
+  static class JmsCheckpointMark implements UnboundedSource.CheckpointMark, Serializable {
+
+    private static final Logger LOG = LoggerFactory.getLogger(JmsCheckpointMark.class);
+
+    @VisibleForTesting Instant oldestMessageTimestamp = Instant.now();
+    @VisibleForTesting transient List<Message> messages = new ArrayList<>();
```
iemejia: A `CheckpointMark` is basically an object to save pending unprocessed state, in the case of JmsIO pending state are the messages. If we make the messages `transient` we are losing data. So we need to fix this to make the messages part of the serialized state of the object. However since Jms `Message`s are not `Serializable` probably what we need to store here are `JmsRecord`s.

iemejia: Possible issue after changing the pending state to be `JmsRecord`s is the fact that the records need to recreate Messages to be able to ack on `finalizeCheckpoint`. Doing ack on the CheckpointMark is probably an error because you would need also the connection information, so better to move that logick out of the CheckpointMark.

sdks/java/io/jms/src/main/java/org/apache/beam/sdk/io/jms/JmsCheckpointMark.java
```diff
@@ -1,184 +0,0 @@
...
-import org.joda.time.Instant;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-/**
- * Checkpoint for an unbounded JmsIO.Read. Consists of JMS destination name, and the latest message
- * ID consumed so far.
- */
-@DefaultCoder(AvroCoder.class)
-public class JmsCheckpointMark implements UnboundedSource.CheckpointMark {
```
iemejia: Better leave this as a separate class as it was, no? KafkaIO got splitted because it was almost unmanageable with so much code in one file, or is there a fundamental reason to do this?

sdks/java/io/jms/src/main/java/org/apache/beam/sdk/io/jms/JmsIO.java
iemejia: Other extra bit is that `JmsRecord` equals and hashcode implementation do not take into account the same fields, this should be also addressed otherwise the equals contract might get broken.

sdks/java/io/jms/src/main/java/org/apache/beam/sdk/io/jms/JmsIO.java
jbonofre: Messages list is transient. Do we really need it in the equals ?

sdks/java/io/jms/src/test/java/org/apache/beam/sdk/io/jms/JmsIOTest.java
jbonofre: +1, I'm doing the change.

sdks/java/io/jms/src/main/java/org/apache/beam/sdk/io/jms/JmsIO.java
jbonofre: That's not possible, simply to execute the ack on messages. So, I'm using the same approach as in MqttIO. Just keep the oldest timestamp, and don't ack the messages. Then, the messages stay in the broker and will be consumed by another executor.

sdks/java/io/jms/src/main/java/org/apache/beam/sdk/io/jms/JmsCheckpointMark.java
jbonofre: As the checkpoint mark is small, I used a inner class (like in MqttIO). But I can isolate it if you think it's better for maintenance.

sdks/java/io/jms/src/main/java/org/apache/beam/sdk/io/jms/JmsIO.java
jbonofre: The finalize checkpoint should ack the message. A better impl is:
1. use auto_ack on session
2. convert Message as JmsRecord and store in the checkpoint mark
3. thanks to that, the record will be serializable and we can transport it.

I'm doing the change.

sdks/java/io/jms/src/main/java/org/apache/beam/sdk/io/jms/JmsCheckpointMark.java
aromanenko-dev: I agree with @iemejia - it would more convenient to have `JmsCheckpointMark` as a separate class (as it was before).

sdks/java/io/jms/src/main/java/org/apache/beam/sdk/io/jms/JmsIO.java
aromanenko-dev: @jbonofre  Could this logic be implemented using `JmsCheckpointMark.State` class as it was before? Well, for this case, at least, we need to make `State` serialisable but it should not be a problem with conversion from `Message` to `JmsRecord`.

sdks/java/io/jms/src/main/java/org/apache/beam/sdk/io/jms/JmsCheckpointMark.java
aromanenko-dev: I agree with @iemejia - it would more convenient to have `JmsCheckpointMark` as a separate class (as it was before).

sdks/java/io/jms/src/main/java/org/apache/beam/sdk/io/jms/JmsIO.java
```diff
@@ -404,9 +408,96 @@ private JmsIO() {}
...
+      try {
+        for (Message message : messages) {
+          try {
+            message.acknowledge();
+            Instant currentMessageTimestamp = new Instant(message.getJMSTimestamp());
+            if (currentMessageTimestamp.isAfter(oldestMessageTimestamp)) {
+              oldestMessageTimestamp = currentMessageTimestamp;
+            }
+          } catch (Exception e) {
+            LOG.error("Exception while finalizing message: {}", e);
```
aromanenko-dev: Is it fine that we don't fail here in case of exception? Can it cause a data loss?

sdks/java/io/jms/src/main/java/org/apache/beam/sdk/io/jms/JmsIO.java
aromanenko-dev: I'm also wondering if we can have OOM here if the size of messages is large?

jbonofre: We can limit the size of the messages list and call finalize checkpoint. State is useless IMHO, I would keep it in checkpoint mark as it's more straight forward to understand.

jbonofre:
In the last commit, I'm using a different approach.
Now, the messages are automatically ack on the JMS broker. The pending messages stored in the checkpoint mark. As I don't have to ack the messages "explicitly" anymore, I can now convert the JMS Message as Beam JmsRecord, and so CheckpointMark is now serializable.
This commit needs some polish, but I wanted to share the rough idea with you.
Thoughts ?

jbonofre:
Just to be clear: I strongly think my original implementation is largely better and the only change to do is to remove/transient the messages in the checkpoint mark.
Let me explain. In JMS, the messages stay in destination (topic or queue) up to the ack.
So, when we finalize the checkpoint mark, we ack the pending messages (stored in the checkpoint mark). If, for any reason, the checkpoint mark moves to another executor, and we don't have the pending messages, it's not a big deal as the messages are still in the destination, and will be consumed again.
So, if you don't mind, I will move forward with my previous commit (where I use CLIENT_ACK and transient on messages in the checkpoint mark) adding the following changes:
- checkpoint mark serialization coder (instead of avro coder)
- transient on messages list
Thoughts ?

jbonofre:
I have updated the PR to use serialization coder instead of avro coder by default.

sdks/java/io/jms/src/main/java/org/apache/beam/sdk/io/jms/JmsIO.java
jbonofre: No, it's not a problem as the ack won't be sent.

jbonofre:
FYI, I'm updating the PR to have JmsCheckpointMark in a separate class (but without State inner class which is useless IMHO).

jbonofre:
As discussed, I've keep `JmsCheckpointMark` in a dedicated class.

iemejia:
Sorry for the delay, the extra commit with the fixes looks good. I was thinking that since the stored messages are not needed to restore the progress of the reads on `UnboundedJmsReader` maybe the simplest fix is just to let them transient as you proposed.
About the State changes maybe let's do those in a subsequent PR so we can get this fix out more quickly. WDYT If you agree just let the class as it was before and then I will merge.

iemejia:
Oh you already get rid of state hehe, my bad ok looking again.

jbonofre:
@iemejia thanks ! I will switch to other IOs improvements ;)

aromanenko-dev:
Run Java PreCommit

iemejia:
Merged manually, thanks again JB!

Arthur-RD:
> Just to be clear: I strongly think my original implementation is largely better and the only change to do is to remove/transient the messages in the checkpoint mark.
> Let me explain. In JMS, the messages stay in destination (topic or queue) up to the ack.
> So, when we finalize the checkpoint mark, we ack the pending messages (stored in the checkpoint mark). If, for any reason, the checkpoint mark moves to another executor, and we don't have the pending messages, it's not a big deal as the messages are still in the destination, and will be consumed again.

Hello,

Thank you for this PR.

However, could you please clarify this point? In particular:
   - **When does the finalizeCheckpoint method is used?**
   - **Could modifying the messages list to transient cause message loss?**

My understanding is the following:
We have a Consumer on a node A. It hasn't acknowledged the message 1 and is stopped by the host. The finalizeCheckpoint method is called. As the Consumer hasn't switched to another node, the messages list still contains this message and it's acknowledged.
If unfortunately the Consumer has switched to a node B for any reason when the stop is asked, the message 1 is not transferred as the Checkpoint messages list is transient, so this consumer doesn't have this message anymore **BUT this message is still in the message broker as it hasn't been acked**.
The message 1 will be consumed again by the consumer and isn't lost.

Can you confirm it, please?
**Does the only impact of changing messages to transient in the CheckpointMark is that they will be consumed twice, or can they be lost?**

Thanks a lot for your answer,
Arthur

PR #10655, opened 2020-01-22
[BEAM-8618] Tear down unused DoFns periodically in Python SDK harness.

sunjincheng121:
Tear down unused DoFns periodically in Python SDK harness.

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/runners/worker/sdk_worker.py
+++ b/sdks/python/apache_beam/runners/worker/sdk_worker.py
@@ -289,7 +289,7 @@ class BundleProcessorCache(object):
     }  # type: Dict[str, Tuple[str, bundle_processor.BundleProcessor]]
     self.cached_bundle_processors = collections.defaultdict(
         list)  # type: DefaultDict[str, List[bundle_processor.BundleProcessor]]
-    self.cached_bundle_processors_last_access_time = \
+    self.last_access_times = \
         collections.defaultdict(float)  # type: DefaultDict[str, float]
     self._schedule_periodic_shutdown()
 
@@ -348,7 +348,7 @@ class BundleProcessorCache(object):
     """
     descriptor_id, processor = self.active_bundle_processors.pop(instruction_id)
     processor.reset()
-    self.cached_bundle_processors_last_access_time[descriptor_id] = time.time()
+    self.last_access_times[descriptor_id] = time.time()
     self.cached_bundle_processors[descriptor_id].append(processor)
 
   def shutdown(self):
@@ -369,8 +369,7 @@ class BundleProcessorCache(object):
 
   def _schedule_periodic_shutdown(self):
     def shutdown_inactive_bundle_processors():
-      items = self.cached_bundle_processors_last_access_time.items()
-      for descriptor_id, last_access_time in items:
+      for descriptor_id, last_access_time in self.last_access_times.items():
         if (time.time() - last_access_time >
             DEFAULT_BUNDLE_PROCESSOR_CACHE_SHUTDOWN_THRESHOLD_S):
           BundleProcessorCache._shutdown_cached_bundle_processors(
--- a/sdks/python/apache_beam/runners/worker/sdk_worker.py
+++ b/sdks/python/apache_beam/runners/worker/sdk_worker.py
@@ -30,6 +30,7 @@ import logging
 import queue
 import sys
 import threading
+import time
 import traceback
 from builtins import object
 from concurrent import futures
@@ -55,6 +56,7 @@ from apache_beam.runners.worker import bundle_processor
 from apache_beam.runners.worker import data_plane
 from apache_beam.runners.worker import statesampler
 from apache_beam.runners.worker.channel_factory import GRPCChannelFactory
+from apache_beam.runners.worker.data_plane import PeriodicThread
 from apache_beam.runners.worker.statecache import StateCache
 from apache_beam.runners.worker.worker_id_interceptor import WorkerIdInterceptor
 from apache_beam.runners.worker.worker_status import FnApiWorkerStatusHandler
@@ -70,6 +72,8 @@ _LOGGER = logging.getLogger(__name__)
 # transitions in over 5 minutes.
 # 5 minutes * 60 seconds * 1020 millis * 1000 micros * 1000 nanoseconds
 DEFAULT_LOG_LULL_TIMEOUT_NS = 5 * 60 * 1000 * 1000 * 1000
+
+DEFAULT_BUNDLE_PROCESSOR_CACHE_SHUTDOWN_THRESHOLD_S = 60
 
 
 class SdkHarness(object):
@@ -285,6 +289,9 @@ class BundleProcessorCache(object):
     }  # type: Dict[str, Tuple[str, bundle_processor.BundleProcessor]]
     self.cached_bundle_processors = collections.defaultdict(
         list)  # type: DefaultDict[str, List[bundle_processor.BundleProcessor]]
+    self.cached_bundle_processors_last_access_time = \
+        collections.defaultdict(float)  # type: DefaultDict[str, float]
+    self._schedule_periodic_shutdown()
 
   def register(self, bundle_descriptor):
     # type: (beam_fn_api_pb2.ProcessBundleDescriptor) -> None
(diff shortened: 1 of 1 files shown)


sunjincheng121:
R: @robertwb @lukecwik

sunjincheng121:
R:  @mxm @lukecwik

sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -280,6 +283,7 @@ def get(self, instruction_id, bundle_descriptor_id):
     try:
       # pop() is threadsafe
       processor = self.cached_bundle_processors[bundle_descriptor_id].pop()
+      self.last_access_time[bundle_descriptor_id] = time.time()
     except IndexError:
```
mxm: This won't update the access time when we first create the processor in the except block.

sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -315,18 +319,47 @@ def release(self, instruction_id):
...
     for cached_bundle_processors in self.cached_bundle_processors.values():
-      while len(cached_bundle_processors) > 0:
-        cached_bundle_processors.pop().shutdown()
+      BundleProcessorCache._shutdown_cached_bundle_processors(
+          cached_bundle_processors)
+
+  def _schedule_periodic_shutdown(self):
+    def shutdown_inactive_bundle_processors():
+      for descriptor_id, last_access_time in self.last_access_time.items():
+        if time.time() - last_access_time > 60:
```
mxm: We may want to make this configurable.

sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -315,18 +319,47 @@ def release(self, instruction_id):
...
+          BundleProcessorCache._shutdown_cached_bundle_processors(
+              self.cached_bundle_processors[descriptor_id])
+
+    from apache_beam.runners.worker.data_plane import PeriodicThread
+    self.periodic_shutdown = PeriodicThread(
+        60, shutdown_inactive_bundle_processors)
```
mxm: Same here. Should be configurable or at least be extracted to a variable.

sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -315,18 +319,47 @@ def release(self, instruction_id):
...
+          cached_bundle_processors)
+
+  def _schedule_periodic_shutdown(self):
+    def shutdown_inactive_bundle_processors():
+      for descriptor_id, last_access_time in self.last_access_time.items():
+        if time.time() - last_access_time > 60:
+          BundleProcessorCache._shutdown_cached_bundle_processors(
+              self.cached_bundle_processors[descriptor_id])
+
+    from apache_beam.runners.worker.data_plane import PeriodicThread
```
mxm: I think we should move this to the import section.

sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -315,18 +319,47 @@ def release(self, instruction_id):
...
-        cached_bundle_processors.pop().shutdown()
+      BundleProcessorCache._shutdown_cached_bundle_processors(
+          cached_bundle_processors)
+
+  def _schedule_periodic_shutdown(self):
+    def shutdown_inactive_bundle_processors():
+      for descriptor_id, last_access_time in self.last_access_time.items():
+        if time.time() - last_access_time > 60:
+          BundleProcessorCache._shutdown_cached_bundle_processors(
+              self.cached_bundle_processors[descriptor_id])
```
mxm: Don't we have to remove the bundle processor list from the dictionary? Otherwise we may access a cached shutdown bundle processor.

sdks/python/apache_beam/runners/worker/sdk_worker.py
sunjincheng121: This is expected. The `last_access_time` represents the last time the `cached_bundle_processors` for some `bundle_descriptor` is accessed. If it exceeds the time limit, the remaining bundle processors cached in the `cached_bundle_processors` will be shutdown. What's your thought?

sdks/python/apache_beam/runners/worker/sdk_worker.py
sunjincheng121: I'm not pretty sure if it deserves a configuration. What about extracting it to a variable initially as you suggested below?

sdks/python/apache_beam/runners/worker/sdk_worker.py
sunjincheng121: The bundle processor will be popped from the bundle processor list before shutdown and the pop operation is thread safe and so I think this will not happen. What's your thought?

sdks/python/apache_beam/runners/worker/sdk_worker.py
mxm: In my eyes, the first time it is accessed is when it is created.

sdks/python/apache_beam/runners/worker/sdk_worker.py
mxm: True, since the pop on the list is thread-safe, we can't wind up with a cached bundle processor being in use.

sdks/python/apache_beam/runners/worker/sdk_worker.py
mxm: Sounds good.

sdks/python/apache_beam/runners/worker/sdk_worker.py
sunjincheng121: The strategy here is that the bundle processors which are unused after an amount of time will be shutdown. When a bundle processor is created in the exception block, there are no cached(unused) bundle processors. The bundle processors become unused only when they are added to the cached bundle processors list. What about rename last_access_time to `cached_bundle_processors_last_access_time` to make it more clear?

mxm: I still do not understand, the method is named `get`, so we access the bundle processor independently of whether we create it or not. It is cached, regardless of whether it is created and added to the cache, or retrieved from the cache. 

Logically, you might want to update the time when putting the processor into the cache. That would be in `release`.

What is the advantage of updating the time here? It should be sufficient to update it in `release`, directly before putting it back.

sunjincheng121: Hmm, If a bundle processor is retrieved from the cache, there is a high possibility that the remaining cached bundle processors will be needed in the future and so the last access time is updated.

If the bundle processor is newly created, it means that the cached bundle processor list is empty. This is the main reason that the last access time is only updated when bundle processor is retrieved from the cache. However, I think that it does no harm to update the last access time in both cases if it makes the code more readable.

What do you think?

mxm: >If the bundle processor is newly created, it means that the cached bundle processor list is empty. This is the main reason that the last access time is only updated when bundle processor is retrieved from the cache. 

Consider the case where we just have a single bundle processor. When we call get for the first time, we won't update the last-used time. However, every time we retrieve it afterwards, we will update the time, but the list of cached bundle processors will remain empty.

I think we should either (1) always update the last-used timestamp in `get`, regardless of creation or (2) update it only on `release`.

I'm leaning towards (2) because while a bundle processor is in-use, it can't be removed anyways. We update the timestamp when we put it back in `release`.

sunjincheng121: Regarding the single bundle processor case, it doesn't harm to update the time as the cached bundle processors is empty. However, in cases where there are multiple bundle processors, it will update the time for the remaining cached bundle processors and so improve the cache hit rate. I think this is main difference between solution (1) and (2). However, I'm fine with both solutions as I think both of them work. Will update the PR according to solution (2) if you are favor of it according to your experience. What's your thought?

sunjincheng121:
Thanks for the review and valuable comments. I have address the comments and I appreciate if you can have another look :) @mxm

sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -315,18 +322,49 @@ def release(self, instruction_id):
...
-        cached_bundle_processors.pop().shutdown()
+      BundleProcessorCache._shutdown_cached_bundle_processors(
+          cached_bundle_processors)
+
+  def _schedule_periodic_shutdown(self):
+    def shutdown_inactive_bundle_processors():
+      for descriptor_id, last_access_time in \
+              self.cached_bundle_processors_last_access_time.items():
+        if time.time() - last_access_time > \
+                DEFAULT_BUNDLE_PROCESSOR_CACHE_THRESHOLD_S:
```
mxm: nit: you could surround the condition with paranthesis which would make it easier to read, e.g. 

```python
       if (time.time() - last_access_time >
                DEFAULT_BUNDLE_PROCESSOR_CACHE_THRESHOLD_S):
```

sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -69,6 +71,8 @@
 # 5 minutes * 60 seconds * 1020 millis * 1000 micros * 1000 nanoseconds
 DEFAULT_LOG_LULL_TIMEOUT_NS = 5 * 60 * 1000 * 1000 * 1000
 
+DEFAULT_BUNDLE_PROCESSOR_CACHE_THRESHOLD_S = 60
```
mxm:
```suggestion
DEFAULT_BUNDLE_PROCESSOR_CACHE_SHUTDOWN_THRESHOLD_S = 60
```

sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -315,18 +322,49 @@ def release(self, instruction_id):
...
     for cached_bundle_processors in self.cached_bundle_processors.values():
-      while len(cached_bundle_processors) > 0:
-        cached_bundle_processors.pop().shutdown()
+      BundleProcessorCache._shutdown_cached_bundle_processors(
+          cached_bundle_processors)
+
+  def _schedule_periodic_shutdown(self):
+    def shutdown_inactive_bundle_processors():
+      for descriptor_id, last_access_time in \
+              self.cached_bundle_processors_last_access_time.items():
```
mxm: Same here, use parenthesis.

sunjincheng121: It seems that it doesn't support parenthesis in the for loop？

lukecwik: It should, there are many examples of for loops using parentheses in the code base already.

kamilwu:
Just to let you know that we've just introduced Python autoformatter. Your merge conflict might be a result of this.
Here you can find an instruction on how to run autoformatter: https://cwiki.apache.org/confluence/display/BEAM/Python+Tips, section Formatting.
Sorry for inconvenience.

sunjincheng121:
squash the commits and rebase the code.

sdks/python/apache_beam/runners/worker/sdk_worker.py
sunjincheng121: Ah, I have added a local variable and it should have no this problem now. Is that make sense to you?

sunjincheng121:
R: @lukecwik @mxm

mxm:
retest this please

sunjincheng121:
Run PythonLint PreCommit

mxm:
```
08:17:21 > Task :sdks:python:test-suites:tox:py2:lintPy27_3
08:17:21 ************* Module apache_beam.runners.worker.sdk_worker
08:17:21 W:372,14: dict.items referenced when not iterating (dict-items-not-iterating)
```
https://builds.apache.org/job/beam_PreCommit_PythonLint_Phrase/70/

sunjincheng121:
Thanks for the info @mxm , I have fix the style issue, and update the PR :)

PR #10664, opened 2020-01-22
[BEAM-9172] Add missing parameter to Nexmark CI execution for Flink runner

iemejia:
Due to two misconfigured parameters:

1. The effective runner `TestFlinkRunner` in `CommonTestProperties.groovy`
2. The missing `--runner` parameter in the nexmark arguments.

The CI tests were not producing good execution results for Flink during the last year :S
(1) Made the results to be produced into a different table see for example `nexmark.nexmark_0_sql_TestFlinkRunner_batch` and because of (2) we were producing bad results because if Nexmark does not have the `--runner` parameter assigned it defaults to execute with Direct runner so all the accumulated results in the `TestFlinkRunner` tables DO NOT correspond to Flink runner results.

This PR fixes both.

R: @udim @mxm

### DIFF
--- a/.test-infra/jenkins/CommonTestProperties.groovy
+++ b/.test-infra/jenkins/CommonTestProperties.groovy
@@ -29,7 +29,7 @@ class CommonTestProperties {
         DATAFLOW("DataflowRunner"),
         SPARK("SparkRunner"),
         SPARK_STRUCTURED_STREAMING("SparkStructuredStreamingRunner"),
-        FLINK("TestFlinkRunner"),
+        FLINK("FlinkRunner"),
         DIRECT("DirectRunner"),
         PORTABLE("PortableRunner")
 
--- a/.test-infra/jenkins/job_PostCommit_Java_Nexmark_Flink.groovy
+++ b/.test-infra/jenkins/job_PostCommit_Java_Nexmark_Flink.groovy
@@ -43,6 +43,8 @@ NoPhraseTriggeringPostCommitBuilder.postCommitJob('beam_PostCommit_Java_Nexmark_
       switches('-Pnexmark.runner=":runners:flink:1.9"' +
               ' -Pnexmark.args="' +
               [NexmarkBigqueryProperties.nexmarkBigQueryArgs,
+              '--runner=FlinkRunner',
+              '--shutdownSourcesOnFinalWatermark=true',
               '--streaming=false',
               '--suite=SMOKE',
               '--streamTimeout=60' ,
@@ -58,6 +60,8 @@ NoPhraseTriggeringPostCommitBuilder.postCommitJob('beam_PostCommit_Java_Nexmark_
       switches('-Pnexmark.runner=":runners:flink:1.9"' +
               ' -Pnexmark.args="' +
               [NexmarkBigqueryProperties.nexmarkBigQueryArgs,
+              '--runner=FlinkRunner',
+              '--shutdownSourcesOnFinalWatermark=true',
               '--streaming=true',
               '--suite=SMOKE',
               '--streamTimeout=60' ,
@@ -73,6 +77,8 @@ NoPhraseTriggeringPostCommitBuilder.postCommitJob('beam_PostCommit_Java_Nexmark_
       switches('-Pnexmark.runner=":runners:flink:1.9"' +
               ' -Pnexmark.args="' +
               [NexmarkBigqueryProperties.nexmarkBigQueryArgs,
+              '--runner=FlinkRunner',
+              '--shutdownSourcesOnFinalWatermark=true',
               '--queryLanguage=sql',
               '--streaming=false',
               '--suite=SMOKE',
@@ -88,6 +94,8 @@ NoPhraseTriggeringPostCommitBuilder.postCommitJob('beam_PostCommit_Java_Nexmark_
       switches('-Pnexmark.runner=":runners:flink:1.9"' +
               ' -Pnexmark.args="' +
               [NexmarkBigqueryProperties.nexmarkBigQueryArgs,
+              '--runner=FlinkRunner',
+              '--shutdownSourcesOnFinalWatermark=true',
               '--queryLanguage=sql',
               '--streaming=true',
               '--suite=SMOKE',
@@ -107,8 +115,9 @@ PhraseTriggeringPostCommitBuilder.postCommitJob('beam_PostCommit_Java_Nexmark_Fl
   commonJobProperties.setTopLevelMainJobProperties(delegate, 'master', 240)
 
   def final JOB_SPECIFIC_OPTIONS = [
-          'suite'        : 'SMOKE',
-          'streamTimeout': 60,
+          'suite' : 'SMOKE',
+          'streamTimeout' : 60,
+          'shutdownSourcesOnFinalWatermark' : true,
   ]
 
   Nexmark.standardJob(delegate, Runner.FLINK, SDK.JAVA, JOB_SPECIFIC_OPTIONS, TriggeringContext.PR)


iemejia:
retest this please

iemejia:
retest this please

iemejia:
Run Flink Runner Nexmark Tests

iemejia:
Run Seed Job

iemejia:
Run Flink Runner Nexmark Tests

udim:
I also opened https://issues.apache.org/jira/browse/BEAM-9171 at around the same time (same conclusion), will mark that as a duplicate.

udim:
The BQ table name is also logged (to stdout?): https://github.com/apache/beam/blob/7bea94c59d1ead24659e09b0e467beeb82f4cadd/sdks/java/testing/nexmark/src/main/java/org/apache/beam/sdk/nexmark/NexmarkLauncher.java#L868

Perhaps adding --info to gradlew flags will make it more obvious which table is being written to.

iemejia:
Nice sync thanks. There is definitely an error here but Thomas reported error is happening with the Portable runner + Python so these results won't help, still good to fix anyway.

iemejia:
I am waiting for the exec to finish to see if the PRs table get updated with the new results and then it will be good to review/merge.

.test-infra/jenkins/job_PostCommit_Java_Nexmark_Flink.groovy
```diff
@@ -43,6 +43,7 @@ NoPhraseTriggeringPostCommitBuilder.postCommitJob('beam_PostCommit_Java_Nexmark_
       switches('-Pnexmark.runner=":runners:flink:1.9"' +
               ' -Pnexmark.args="' +
               [NexmarkBigqueryProperties.nexmarkBigQueryArgs,
+               '--runner=FlinkRunner',
```
udim: Does this affect the runner used or just the table name? I'm asking because I believe the runner used is TestFlinkRunner (while the existing table is uses the former runner name).
@mxm

iemejia: It affects both actually and no, TestFlinkRunner is not used for Nexmark since this is not a 'Test' but a real 'execution'

udim:
It seems that the jobs `beam_PostCommit_Java_Nexmark_Flink` and `beam_PostCommit_Java_Nexmark_Flink_PR` are configured somewhat differently. The _PR variant was using `--runner=TestFlinkRunner`, before and after running the seed job here.

iemejia:
Oh I am missing the PRs one, let me fix that.

iemejia:
Oh no wait I misunderstood. Both come from the same script the difference is if it is invoked in the PR or on Postcommit

iemejia:
Updated using now the real runner (TestFlinkRunner is just a wrapper, but it could be producing the name misconfiguration for the tables.

iemejia:
Just to expand on this, if you trigger Nexmark from a PR like here the results go into a different tablespace than the postcommit executions (this is by design), so to validate that this work we need to see that the 'manual' run in this PR produced results in the PRs namespace table:
```sql
SELECT DATE(timestamp) as date, runtimeSec
FROM nexmark_PRs.nexmark_0_FlinkRunner_batch
ORDER BY date DESC;
```

iemejia:
Previous run did not produce results :S, let's see now

iemejia:
Run Seed Job

iemejia:
Thanks for the hint on TestFlinkRunner the results were going to a different table because of that :sa: 
We will have to migrate that data. See in the BigQuery console `nexmark.nexmark_0_sql_TestFlinkRunner_batch`

iemejia:
Run Flink Runner Nexmark Tests

iemejia:
Table produced new data now for the batch version, however it seems that current master is blocking the execution of Nexmark in streaming mode with Flink runner.
This command blocks forever in my machine.

```bash
./gradlew :sdks:java:testing:nexmark:run \
    -Pnexmark.runner=":runners:flink:1.8" \
    -Pnexmark.args="
        --runner=FlinkRunner
        --streaming=true
        --suite=SMOKE
        --manageResources=false
        --monitorJobs=true"
```

 I went back until version 2.9.0 where this similar command worked.
```bash
./gradlew :beam-sdks-java-nexmark:run \
    -Pnexmark.runner=":beam-runners-flink_2.11" \
    -Pnexmark.args="
        --runner=FlinkRunner
        --streaming=true
        --suite=SMOKE
        --manageResources=false
        --monitorJobs=true"
```

Maybe @mxm can help us find what is going on with the streaming run.

.test-infra/jenkins/job_PostCommit_Java_Nexmark_Flink.groovy
mxm: `TestFlinkRunner` is used to set the FlinkRunner's `shutdownSourcesOnFinalWatermark` pipeline option. In the regular `FlinkRunner`, streaming pipelines never shuts down which causes the tests to block here. When sources shut down checkpointing stops working in Flink, that's why we never shut down in production systems. See also https://jira.apache.org/jira/browse/FLINK-2491

mxm: You can use `--shutdownSourcesOnFinalWatermark` here.

mxm: Adding the above pipeline option will unblock streaming.

iemejia:
Thanks Max, adding the parameter let's see if it works.

iemejia:
Run Seed Job

iemejia:
Run Flink Runner Nexmark Tests

iemejia:
Run Seed Job

iemejia:
Run Flink Runner Nexmark Tests

iemejia:
It is green now and data is getting to the right tables. Can you PTAL @mxm 
btw Thanks to @mwalenia for the useful information on the missing parameter for the phrase based version.

.test-infra/jenkins/job_PostCommit_Java_Nexmark_Flink.groovy
```diff
@@ -43,6 +43,8 @@ NoPhraseTriggeringPostCommitBuilder.postCommitJob('beam_PostCommit_Java_Nexmark_
       switches('-Pnexmark.runner=":runners:flink:1.9"' +
               ' -Pnexmark.args="' +
               [NexmarkBigqueryProperties.nexmarkBigQueryArgs,
+              '--runner=FlinkRunner',
+              '--shutdownSourcesOnFinalWatermark=true',
```
mxm: It looks like we still want to use `TestFlinkRunner` here because it auto-sets the `--shutdownSourcesOnFinalWatermark` flag.

.test-infra/jenkins/CommonTestProperties.groovy
```diff
@@ -29,7 +29,7 @@ class CommonTestProperties {
         DATAFLOW("DataflowRunner"),
         SPARK("SparkRunner"),
         SPARK_STRUCTURED_STREAMING("SparkStructuredStreamingRunner"),
-        FLINK("TestFlinkRunner"),
+        FLINK("FlinkRunner"),
```
mxm: This string broke the result reporting because it wrote to the wrong BigQuery table?

.test-infra/jenkins/job_PostCommit_Java_Nexmark_Flink.groovy
mxm: Wouldn't it be enough to just do the [above change](https://github.com/apache/beam/pull/10664/files#diff-3bbf75b566341b4b08d71632cd27e022R32)?

.test-infra/jenkins/job_PostCommit_Java_Nexmark_Flink.groovy
```diff
@@ -107,8 +115,9 @@ PhraseTriggeringPostCommitBuilder.postCommitJob('beam_PostCommit_Java_Nexmark_Fl
   commonJobProperties.setTopLevelMainJobProperties(delegate, 'master', 240)
 
   def final JOB_SPECIFIC_OPTIONS = [
-          'suite'        : 'SMOKE',
-          'streamTimeout': 60,
+          'suite' : 'SMOKE',
+          'streamTimeout' : 60,
+          'shutdownSourcesOnFinalWatermark' : true,
```
mxm: This can also be removed if we use `TestFlinkRunner`.

.test-infra/jenkins/CommonTestProperties.groovy
mxm: Together with setting `--runner=TestFlinkRunner`, this should suffice to make it work again.

.test-infra/jenkins/job_PostCommit_Java_Nexmark_Flink.groovy
iemejia: No, if we use `TestFlinkRunner` the results will be saved in the table with the corrupted data (the runs with the `DirectRunner` results. If we use FlinkRunner we will get the data in the same table with the historical data previous to 2019/01.

.test-infra/jenkins/CommonTestProperties.groovy
iemejia: Yes exactly this wrote to a different table than the historic one. Also having the non Test version will make it uniform with all the other runners + having the parallelism not limited to 1 which gave the impression of worse performance.

.test-infra/jenkins/job_PostCommit_Java_Nexmark_Flink.groovy
iemejia: Same arguments as above. Considering the pros cons I think we have way more extra work letting things back to TestFlinkRunner, also we were supposed to remove this testrunner things no? Please remember that Nexmark is not a UT or a IT, it is a 'user application' so it should be using the real runner.

iemejia:
Yay, thanks Max!

mxm:
You're welcome. Thanks for fixing this!

PR #10611, opened 2020-01-16
[BEAM-9132] De-register state request handler last when closing ActiveBundle

mxm:
We have observed these errors in a state-intense application:

```
Error processing instruction 107. Original traceback is
Traceback (most recent call last):
  File "apache_beam/runners/common.py", line 780, in apache_beam.runners.common.DoFnRunner.process
  File "apache_beam/runners/common.py", line 587, in apache_beam.runners.common.PerWindowInvoker.invoke_process
  File "apache_beam/runners/common.py", line 659, in apache_beam.runners.common.PerWindowInvoker._invoke_process_per_window
  File "apache_beam/runners/common.py", line 880, in apache_beam.runners.common._OutputProcessor.process_outputs
  File "apache_beam/runners/common.py", line 895, in apache_beam.runners.common._OutputProcessor.process_outputs
  File "redacted.py", line 56, in process
    recent_events_map = load_recent_events_map(recent_events_state)
  File "redacted.py", line 128, in _load_recent_events_map
    items_in_recent_events_bag = list(recent_events_state.read())
  File "apache_beam/runners/worker/bundle_processor.py", line 335, in __iter__
    for elem in self.first:
  File "apache_beam/runners/worker/bundle_processor.py", line 214, in __iter__
    self._state_key, self._coder_impl, is_cached=self._is_cached)
  File "apache_beam/runners/worker/sdk_worker.py", line 692, in blocking_get
    self._materialize_iter(state_key, coder))
  File "apache_beam/runners/worker/sdk_worker.py", line 723, in _materialize_iter
    self._underlying.get_raw(state_key, continuation_token)
  File "apache_beam/runners/worker/sdk_worker.py", line 603, in get_raw
    continuation_token=continuation_token)))
  File "apache_beam/runners/worker/sdk_worker.py", line 637, in _blocking_request
    raise RuntimeError(response.error)
RuntimeError: Unknown process bundle instruction id '107'
```

Notice that the error is thrown on the Runner side. The solution is to
de-register the state request handler only after ensuring no more elements are
in-flight, i.e. the input receivers are closed.

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

Lang | SDK | Apex | Dataflow | Flink | Gearpump | Samza | Spark
--- | --- | --- | --- | --- | --- | --- | ---
[...]

PR #10694, opened 2020-01-27
[BEAM-9132] Avoid logging misleading error messages during pipeline failure

mxm:
We have observed these errors in a state-intense application:

```
Error processing instruction 107. Original traceback is
  File "apache_beam/runners/common.py", line 780, in apache_beam.runners.common.DoFnRunner.process
  File "apache_beam/runners/common.py", line 587, in apache_beam.runners.common.PerWindowInvoker.invoke_process
  File "apache_beam/runners/common.py", line 659, in apache_beam.runners.common.PerWindowInvoker._invoke_process_per_window
  File "apache_beam/runners/common.py", line 880, in apache_beam.runners.common._OutputProcessor.process_outputs
  File "apache_beam/runners/common.py", line 895, in apache_beam.runners.common._OutputProcessor.process_outputs
  File "redacted.py", line 56, in process
    recent_events_map = load_recent_events_map(recent_events_state)
  File "redacted.py", line 128, in _load_recent_events_map
    items_in_recent_events_bag = list(recent_events_state.read())
  File "apache_beam/runners/worker/bundle_processor.py", line 335, in __iter__
    for elem in self.first:
  File "apache_beam/runners/worker/bundle_processor.py", line 214, in __iter__
    self._state_key, self._coder_impl, is_cached=self._is_cached)
  File "apache_beam/runners/worker/sdk_worker.py", line 692, in blocking_get
    self._materialize_iter(state_key, coder))
  File "apache_beam/runners/worker/sdk_worker.py", line 723, in _materialize_iter
    self._underlying.get_raw(state_key, continuation_token)
  File "apache_beam/runners/worker/sdk_worker.py", line 603, in get_raw
    continuation_token=continuation_token)))
  File "apache_beam/runners/worker/sdk_worker.py", line 637, in _blocking_request
    raise RuntimeError(response.error)
RuntimeError: Unknown process bundle instruction id '107'
```

Notice that the error is thrown on the Runner side in the GrpcStateService.

Explanation
===========

We experienced a network split of the Flink cluster which caused the job to
fail. The logging implied a failure of the state requests right before the
network split. However, the opposite is true, the failure occured during
shutdown of the pipeline where the state request handler has already been removed.

Solution
========

Ensure that we shutdown the environment and any pending clients which may be
processing pending requests.

--- | --- | --- | --- | ---

### DIFF
--- a/runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
+++ b/runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
@@ -19,7 +19,7 @@ package org.apache.beam.runners.fnexecution.control;
 
 import com.google.auto.value.AutoValue;
 import java.io.IOException;
-import java.util.HashMap;
+import java.util.IdentityHashMap;
 import java.util.Map;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
@@ -321,8 +321,9 @@ public class DefaultJobBundleFactory implements JobBundleFactory {
 
     private final ExecutableStage executableStage;
     private final int environmentIndex;
-    private final HashMap<WrappedSdkHarnessClient, PreparedClient> preparedClients = new HashMap();
-    private PreparedClient currentClient;
+    private final Map<WrappedSdkHarnessClient, PreparedClient> preparedClients =
+        new IdentityHashMap<>();
+    private volatile PreparedClient currentClient;
 
     private SimpleStageBundleFactory(ExecutableStage executableStage) {
       this.executableStage = executableStage;
@@ -342,22 +343,14 @@ public class DefaultJobBundleFactory implements JobBundleFactory {
       // TODO: Consider having BundleProcessor#newBundle take in an OutputReceiverFactory rather
       // than constructing the receiver map here. Every bundle factory will need this.
 
-      if (environmentExpirationMillis == 0 && !loadBalanceBundles) {
-        return currentClient.processor.newBundle(
-            getOutputReceivers(currentClient.processBundleDescriptor, outputReceiverFactory)
-                .build(),
-            stateRequestHandler,
-            progressHandler);
-      }
-
       final LoadingCache<Environment, WrappedSdkHarnessClient> currentCache;
+      final WrappedSdkHarnessClient client;
       if (loadBalanceBundles) {
         // The semaphore is used to ensure fairness, i.e. first stop first go.
         availableCachesSemaphore.acquire();
         // The blocking queue of caches for serving multiple bundles concurrently.
         currentCache = availableCaches.take();
-        WrappedSdkHarnessClient client =
-            currentCache.getUnchecked(executableStage.getEnvironment());
+        client = currentCache.getUnchecked(executableStage.getEnvironment());
         client.ref();
 
         currentClient = preparedClients.get(client);
@@ -370,8 +363,7 @@ public class DefaultJobBundleFactory implements JobBundleFactory {
 
       } else {
         currentCache = environmentCaches.get(environmentIndex);
-        WrappedSdkHarnessClient client =
-            currentCache.getUnchecked(executableStage.getEnvironment());
+        client = currentCache.getUnchecked(executableStage.getEnvironment());
         client.ref();
 
         if (currentClient.wrappedClient != client) {
@@ -406,11 +398,14 @@ public class DefaultJobBundleFactory implements JobBundleFactory {
 
         @Override
         public void close() throws Exception {
-          bundle.close();
-          currentClient.wrappedClient.unref();
-          if (loadBalanceBundles) {
-            availableCaches.offer(currentCache);
-            availableCachesSemaphore.release();
+          try {
+            bundle.close();
+          } finally {
+            client.unref();
+            if (loadBalanceBundles) {
+              availableCaches.offer(currentCache);
+              availableCachesSemaphore.release();
+            }
           }
         }
       };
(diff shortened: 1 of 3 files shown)


mxm:
This is a follow-up to #10611 to fix the root cause of the state handler de-registration. I appreciate if you could have a look @lukecwik.

mxm:
This does not fix the issue because the id generation scheme is effectively the same. I'm still seeing the same errors, but I have a new trace.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
```diff
@@ -464,6 +474,8 @@ ServerInfo getServerInfo() {
     }
 
     public void close() {
```
tweise: Isn't close only called from unref? If so, how does this change the behavior? (Possibly some more explanation needs to be added.)

mxm: It's now also called from here: https://github.com/apache/beam/pull/10694/files#diff-e80c769f0011537cc2b60d3e7898cf5aR260

runners/java-fn-execution/src/test/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactoryTest.java
```diff
@@ -360,6 +360,19 @@ public void closesEnvironmentOnCleanup() throws Exception {
...
 
+  @Test
+  public void closesEnvironmentOnCleanupWithPendingRefs() throws Exception {
+    try (DefaultJobBundleFactory bundleFactory =
+        createDefaultJobBundleFactory(envFactoryProviderMap)) {
+      DefaultJobBundleFactory.SimpleStageBundleFactory stageBundleFactory =
+          (DefaultJobBundleFactory.SimpleStageBundleFactory)
+              bundleFactory.forStage(getExecutableStage(environment));
+      // The client is still being used, e.g. when the pipeline fails and is shut down
+      stageBundleFactory.currentClient.wrappedClient.ref();
```
tweise: What would cause this extra ref in the actual operator lifecycle?

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
```diff
@@ -255,6 +255,14 @@ public void close() throws Exception {
     // Clear the cache. This closes all active environments.
     // note this may cause open calls to be cancelled by the peer
     for (LoadingCache<Environment, WrappedSdkHarnessClient> environmentCache : environmentCaches) {
+      for (WrappedSdkHarnessClient client : environmentCache.asMap().values()) {
+        try {
+          client.close();
```
tweise: Worth mentioning that this is added to close the environments irrespective of open bundles, since this will occur only during shutdown?

runners/java-fn-execution/src/test/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactoryTest.java
mxm: I'm assuming here we have a `ref()` due to the client being used. If we had no ref, we would be idling. The error message we were seeing is only possible if we are currently processing a bundle.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
mxm: I've removed this in favor of using `unref()`. I think I found the case where `unref()` would not be called: https://github.com/apache/beam/pull/10694/files#diff-e80c769f0011537cc2b60d3e7898cf5aR413

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
```diff
@@ -406,11 +407,14 @@ public void split(double fractionOfRemainder) {
...
-            availableCaches.offer(currentCache);
-            availableCachesSemaphore.release();
+          try {
+            bundle.close();
+          } finally {
+            currentClient.wrappedClient.unref();
+            if (loadBalanceBundles) {
+              availableCaches.offer(currentCache);
+              availableCachesSemaphore.release();
+            }
```
mxm: Looks like this was responsible for the cleanup failing. bundle.close() may throw leaving the environment still referenced. My tests do not yield any more errors like in the description.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
```diff
@@ -342,22 +343,14 @@ public RemoteBundle getBundle(
       // TODO: Consider having BundleProcessor#newBundle take in an OutputReceiverFactory rather
       // than constructing the receiver map here. Every bundle factory will need this.
 
-      if (environmentExpirationMillis == 0 && !loadBalanceBundles) {
-        return currentClient.processor.newBundle(
-            getOutputReceivers(currentClient.processBundleDescriptor, outputReceiverFactory)
-                .build(),
-            stateRequestHandler,
-            progressHandler);
-      }
```
mxm: I discovered another issue in this code path. `ref()` is not called on the client in this case. Also, `unref()` will never be called because the code below does not insert the call to the bundle close. `close()` is only ever called on `unref()`.

tweise: This is the original code path that was used when no reference counting is required. What was the problem with it?

mxm: I suppose it was working correctly due to the initial ref() when creating the environment. Still, it doesn't hurt to remove this bit because it just adds a specialized execution path to the generalized version. The ref business is already complex enough. If you think differently feel free to add it back, but IMHO this is easier to understand.

mxm: During the initial review of the bundle load balancing the idea came up to have two implementations, one with refing, one without. I think that would be the best improvement moving forward.

PR #10708, opened 2020-01-29
[BEAM-9213] throw error when flink_submit_uber_jar and not flink_master

ibzib:

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/runners/portability/flink_runner.py
+++ b/sdks/python/apache_beam/runners/portability/flink_runner.py
@@ -51,8 +51,11 @@ class FlinkRunner(portable_runner.PortableRunner):
     flink_master = self.add_http_scheme(
         flink_options.flink_master)
     flink_options.flink_master = flink_master
-    if (flink_options.flink_submit_uber_jar
-        and flink_master not in MAGIC_HOST_NAMES):
+    if flink_options.flink_submit_uber_jar:
+      if flink_master in MAGIC_HOST_NAMES:
+        raise ValueError(
+            'Cannot use flink_submit_uber_jar with flink_master %s'
+            % flink_master)
       if sys.version_info < (3, 6):
         raise ValueError(
             'flink_submit_uber_jar requires Python 3.6+, current version %s'


sdks/python/apache_beam/runners/portability/flink_runner.py
```diff
@@ -51,8 +51,11 @@ def default_job_server(self, options):
     flink_master = self.add_http_scheme(
         flink_options.flink_master)
     flink_options.flink_master = flink_master
-    if (flink_options.flink_submit_uber_jar
-        and flink_master not in MAGIC_HOST_NAMES):
+    if flink_options.flink_submit_uber_jar:
+      if flink_master in MAGIC_HOST_NAMES:
+        raise ValueError(
+            'Cannot use flink_submit_uber_jar with flink_master %s'
+            % flink_master)
```
mxm: +1 I think it is good to be explicit here instead of silently following back to the JarJobServer.

robertwb: On the other hand, I think we may want to make flink_submit_uber_jar the default, in which case this logic wouldn't make sense.

ibzib: > I think we may want to make flink_submit_uber_jar the default

In the long term, I think we should make flink_submit_uber_jar the _only_ option in FlinkRunner. Having two substantially different behaviors under the same "runner" is confusing to users, and in retrospect "flink_submit_uber_jar=false" kind of seems like only an intermediate step. WDYT?

(Incidentally, I've realized all this terminology itself is confusing. "Uber jar" seems to imply java, even though that is the "java-less" option.)

However, in the short term, I'm not yet confident that flink_submit_uber_jar is ready. I've been seeing issues with it hanging in Beam 2.18: [BEAM-9225](https://issues.apache.org/jira/browse/BEAM-9225)

mxm: I understand your rationale @robertwb. Perhaps the alternative to this change would be to print/log a warning that we're falling back to using the normal job server. We could update the `flink_submit_uber_jar` parameter description to state that the parameter is only effective when `flink_master` has been set.

ibzib: > Perhaps the alternative to this change would be to print/log a warning that we're falling back to using the normal job server. We could update the flink_submit_uber_jar parameter description to state that the parameter is only effective when flink_master has been set.

I don't think either of these options would be sufficiently visible to an unfamiliar user.

robertwb: I think the desired end-user behavior is:

(1) If no options are provided, a local flink cluster is started up and is used,
(2) If a master is specified, we submit it to that job in the least intrusive way possible (IMHO, the so-called uberjar method, though I'm curious about BEAM-9225). 

An argument could be made for (1) that we should instead be using [auto] and the parameters would get picked up from wherever flink gets them if they're not explicitly provided, and possibly use the uberjar if that's not local.

mxm: There is also a case for removing the parameter entirely. Then the fallback would be an implementation detail, given that either execution path works reliably. As a user myself, I don't like when parameters do not yield what they promise. I'd rather have the program figure it out itself without any parameters.

kamilwu:
Just to let you know that we've just introduced Python autoformatter. Your merge conflict might be a result of this.
Here you can find an instruction on how to run autoformatter: https://cwiki.apache.org/confluence/display/BEAM/Python+Tips, section Formatting.
Sorry for inconvenience.

sdks/python/apache_beam/runners/portability/flink_runner.py
mxm: How about a `--force-mode` parameter which would enforce a specific mode, e.g. `--force-mode uberjar` to user the uber jar mode. It seems appropriate then to throw an error in case that mode cannot be run.

mxm: (the existing parameter would be removed)

ibzib: Circling back on this now that BEAM-9225 is resolved.

> (1) If no options are provided, a local flink cluster is started up and is used,
> (2) If a master is specified, we submit it to that job in the least intrusive way possible (IMHO, the so-called uberjar method, though I'm curious about BEAM-9225).

This makes sense to me. I'm leaning toward removing `flink_submit_uber_jar` and replacing the condition with `flink_master`.

> How about a --force-mode parameter which would enforce a specific mode, e.g. --force-mode uberjar to user the uber jar mode.

I prefer removing parameters where possible. Ideally we can come up with a solution here that works without the user having to know all about the different implementations.

ibzib:
Closed this in favor of BEAM-9986.

PR #10621, opened 2020-01-17
[BEAM-9056] Staging artifacts from environment

ihji:
Staging artifacts from environment. This PR implements the first part of a cross-language pipeline dependency management. Details: https://docs.google.com/document/d/1L7MJcfyy9mg2Ahfw5XPhUeBe-dyvAPMOYOiFA1-kAog

------------------------

Thank you for your contribution! Follow this checklist to help us incorporate your contribution quickly and easily:

 - [ ] [**Choose reviewer(s)**](https://beam.apache.org/contribute/#make-your-change) and mention them in a comment (`R: @username`).
 - [ ] Format the pull request title like `[BEAM-XXX] Fixes bug in ApproximateQuantiles`, where you replace `BEAM-XXX` with the appropriate JIRA issue, if applicable. This will automatically link the pull request to the issue.
 - [ ] If this contribution is large, please file an Apache [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

See the [Contributor Guide](https://beam.apache.org/contribute) for more tips on [how to make review process smoother](https://beam.apache.org/contribute/#make-reviewers-job-easier).

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

Lang | SDK | Apex | Dataflow | Flink | Gearpump | Samza | Spark
--- | --- | --- | --- | --- | --- | --- | ---
Go | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go/lastCompletedBuild/) | --- | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_VR_Flink/lastCompletedBuild/) | --- | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_VR_Spark/lastCompletedBuild/)
[...]

PR #10733, opened 2020-01-30
[BEAM-9229] Adding dependency information to Environment proto

ihji:
Proto extension for cross-language dependency management. #10621 depends on this PR.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/model/pipeline/src/main/proto/beam_runner_api.proto
+++ b/model/pipeline/src/main/proto/beam_runner_api.proto
@@ -1203,9 +1203,6 @@ message ArtifactUrlPayload {
 message EmbeddedFilePayload {
   // raw data bytes for an embedded artifact
   bytes data = 1;
-
-  // A name of the embedded file.
-  string name = 2;
 }
 
 message PyPIPayload {
--- a/model/pipeline/src/main/proto/beam_runner_api.proto
+++ b/model/pipeline/src/main/proto/beam_runner_api.proto
@@ -1158,6 +1158,90 @@ message SideInput {
   FunctionSpec window_mapping_fn = 3;
 }
 
+message StandardArtifacts {
+  enum Types {
+    // A URN for locally-accessible artifact files.
+    // payload: ArtifactFilePayload
+    FILE     = 0 [(beam_urn) = "beam:artifact:type:file:v1"];
+
+    // A URN for artifacts described by URLs.
+    // payload: ArtifactUrlPayload
+    URL      = 1 [(beam_urn) = "beam:artifact:type:url:v1"];
+
+    // A URN for artifacts embedded in ArtifactInformation proto.
+    // payload: EmbeddedFilePayload.
+    EMBEDDED = 2 [(beam_urn) = "beam:artifact:type:embedded:v1"];
+
+    // A URN for Python artifacts hosted on PYPI.
+    // payload: PypiPayload
+    PYPI     = 3 [(beam_urn) = "beam:artifact:type:pypi:v1"];
+
+    // A URN for Java artifacts hosted on a Maven repository.
+    // payload: MavenPayload
+    MAVEN    = 4 [(beam_urn) = "beam:artifact:type:maven:v1"];
+  }
+  enum Roles {
+    // A URN for staging-to role.
+    // payload: ArtifactStagingToRolePayload
+    STAGING_TO  = 0 [(beam_urn) = "beam:artifact:role:staging_to:v1"];
+  }
+}
+
+message ArtifactFilePayload {
+  // a string for an artifact path e.g. "/tmp/foo.jar"
+  string path = 1;
+
+  // The hex-encoded sha256 checksum of the artifact.
+  string sha256 = 2;
+}
+
+message ArtifactUrlPayload {
+  // a string for an artifact URL e.g. "https://.../foo.jar" or "gs://tmp/foo.jar"
+  string path = 1;
+}
+
+message EmbeddedFilePayload {
+  // raw data bytes for an embedded artifact
+  bytes data = 1;
+
+  // A name of the embedded file.
+  string name = 2;
+}
+
+message PyPIPayload {
+  // Pypi compatible artifact id e.g. "apache-beam"
+  string artifact_id = 1;
+
+  // Pypi compatible version string.
+  string version = 2;
+}
+
+message MavenPayload {
+  // A string specifying Maven artifact.
+  // The standard format is "groupId:artifactId:version[:packaging[:classifier]]"
+  string artifact = 1;
+
+  // (Optional) Repository URL. If not specified, Maven central is used by default.
+  string repository_url = 2;
+}
+
+message ArtifactStagingToRolePayload {
+  // A generated staged name (relative path under staging directory).
+  string staged_name = 1;
+}
+
+message ArtifactInformation {
+  // A URN that describes the type of artifact
+  string type_urn = 1;
+
+  bytes type_payload = 2;
+
+  // A URN that describes the role of artifact
+  string role_urn = 3;
+
+  bytes role_payload = 4;
+}
+
 // An environment for executing UDFs. By default, an SDK container URL, but
 // can also be a process forked by a command, or an externally managed process.
 message Environment {
@@ -1167,6 +1251,7 @@ message Environment {
   // (Optional) The data specifying any parameters to the URN. If
   // the URN does not require any arguments, this may be omitted.
   bytes payload = 3;
+
 
   // (Optional) Static display data for the environment. If there is none,
   // it may be omitted.
@@ -1178,6 +1263,9 @@ message Environment {
   // appearing in the pipeline proto. This may also be used to indicate
   // support of optional protocols not tied to a concrete component.
   repeated string capabilities = 5;
+
+  // (Optional) artifact dependency information used for executing UDFs in this environment.
+  repeated ArtifactInformation dependencies = 6;
 
   reserved 1;
 }


ihji:
R: @robertwb

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -1087,6 +1087,44 @@ message SideInput {
   FunctionSpec window_mapping_fn = 3;
 }
 
+message StandardArtifacts {
+  enum Types {
+    // A URN for artifacts stored in a local directory.
+    // payload: ArtifactFilePayload.
+    FILE     = 0 [(beam_urn) = "beam:artifact:file:v1"];
```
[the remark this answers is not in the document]
ihji: Thanks!

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -1087,6 +1087,44 @@ message SideInput {
...
+    // A URN for artifacts embedded in ArtifactInformation proto.
+    // payload: raw data bytes.
+    EMBEDDED = 1 [(beam_urn) = "beam:artifact:embedded:v1"];
+    // A URN for artifacts described by HTTP links.
+    // payload: a string for an artifact HTTP URL
+    HTTP     = 2 [(beam_urn) = "beam:artifact:http:v1"];
```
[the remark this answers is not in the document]
ihji: Make sense. Renamed to 'remote'.

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -1087,6 +1087,44 @@ message SideInput {
...
+    // payload: None
+    MAVEN    = 4 [(beam_urn) = "beam:artifact:maven:v1"];
+  }
+}
+
+message ArtifactFilePayload {
+  // A path to an artifact file on a local system.
+  string local_path = 1;
+  // A generated staged name (no path).
+  string staged_name = 2;
```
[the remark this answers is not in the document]
ihji: stage_name is specific to files ready to be staged (local files and embedded files).

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -1087,6 +1087,44 @@ message SideInput {
...
+    // artifact_id: [maven group id]:[maven artifact id]
+    // version_range: a Maven compatible version string
+    // payload: None
+    MAVEN    = 4 [(beam_urn) = "beam:artifact:maven:v1"];
+  }
+}
+
+message ArtifactFilePayload {
+  // A path to an artifact file on a local system.
+  string local_path = 1;
```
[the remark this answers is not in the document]
ihji: It's assumed to be used when SDK is saving remote artifacts and preparing to submit them to the staging service from locally downloaded files. `local_path` here is not ubiquitous information. After submitting the artifacts, retrieval_token (and staged_name if necessary) should be used instead of `local_path`. Is there any specific use-cases you're considering?

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -1087,6 +1087,44 @@ message SideInput {
...
+  // A generated staged name (no path).
+  string staged_name = 2;
+}
+
+message ArtifactInformation {
+  string urn = 1;
```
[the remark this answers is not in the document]
ihji: Looks like `urn` and `payload` combination is mostly common in `beam_runner_api.proto` since I couldn't find any other usage of `string type =`. I think it would be better to follow the same convention here. WDYT?

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -1087,6 +1087,44 @@ message SideInput {
...
+  bytes payload = 2;
+  string artifact_id = 3;
+  string version_range = 4;
+}
```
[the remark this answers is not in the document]
ihji: Thanks for a good catch. Added staged_name for embedded payload.

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -1087,6 +1087,44 @@ message SideInput {
...
+  // A path to an artifact file on a local system.
+  string local_path = 1;
+  // A generated staged name (no path).
+  string staged_name = 2;
+}
+
+message ArtifactInformation {
+  string urn = 1;
+  bytes payload = 2;
+  string artifact_id = 3;
```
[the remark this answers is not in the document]
ihji: We also need artifact_id and version_range in local types (local files and embedded files) for possible deduplication. If the field is common for all types of artifacts I thought it should be the top-level attribute. And yes it'd be nice to use a standard format for the top-level version_range arrtibute but I don't have a good idea at this moment.

ihji:
> Thanks. Could you also include the changes to the artifact staging and/or retrieval service? And the expansion service?

I think it would be better to review them separately. The changes to the artifact staging and retrieval service proto require the not-too-small changes to the Java / Python / Go source codes to be compiled. We can submit this PR without breaking the build but not the PR with the service changes.

robertwb:
The expansion service would necessarily need new APIs; worse case we could implement them as not-implemented stubs in the new language. (Agree about avoiding breaking changes for artifact staging/retrieval.)

model/pipeline/src/main/proto/beam_runner_api.proto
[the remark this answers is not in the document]
robertwb: One issue here is that the callee may not know if it is on a shared filesystem with the caller (e.g. when calling the expansion service). And when calling two distinct expansion services, one would like to be able to compare between them. 

Also, perhaps we should not limit ourselves to local paths here, but any path that can be opened with beam filesystems.

model/pipeline/src/main/proto/beam_runner_api.proto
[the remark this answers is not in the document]
robertwb: But eventually we'll have to give it a name, right? (One could argue that dependencies should be a map(name -> artifact). OTOH, for some perhaps we could leave it blank and one could be inferred (e.g. for urls).

model/pipeline/src/main/proto/beam_runner_api.proto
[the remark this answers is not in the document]
robertwb: If we can't come up with a standard format, I think it should be part of the payload. Similarly for artifact_id--they should mean the same thing. (We can always safely dedup on urn+payloads.)

model/pipeline/src/main/proto/beam_runner_api.proto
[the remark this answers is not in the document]
robertwb: But we'll need it for more than just embedded payload. And the name itself may not be enough to determine the role (do we try to install all .tar.gz files in Python? Or are some just data?)

model/pipeline/src/main/proto/beam_runner_api.proto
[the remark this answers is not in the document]
ihji: Makes sense. Moved artifact_id and version_range to each payload.

model/pipeline/src/main/proto/beam_runner_api.proto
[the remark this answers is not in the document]
ihji: I see your point however I wonder what other options we have. It looks hard to describe all possible roles of a given file as a proto field. Maybe we could just delegate how to interpret a given file name to environments? (for instance, predefined file name for installable .tar.gz package for Python)

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -1087,6 +1087,44 @@ message SideInput {
...
 }
 
+message StandardArtifacts {
+  enum Types {
+    // A URN for artifacts stored in a local directory.
+    // payload: ArtifactFilePayload.
+    FILE     = 0 [(beam_urn) = "beam:artifact:file:v1"];
+    // A URN for artifacts embedded in ArtifactInformation proto.
+    // payload: raw data bytes.
+    EMBEDDED = 1 [(beam_urn) = "beam:artifact:embedded:v1"];
```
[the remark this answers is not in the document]
ihji: We could transfer big files via streaming service protocol. Embedded type here is only for files small-enough in a proto size limit.

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -1093,6 +1093,72 @@ message SideInput {
...
+
+    // A URN for artifacts described by remote URLs.
+    // payload: a string for an artifact URL e.g. "https://.../foo.jar"
+    REMOTE     = 2 [(beam_urn) = "beam:artifact:remote:v1"];
+
+    // A URN for Python artifacts hosted on PYPI.
+    // payload: PypiPayload
+    PYPI     = 3 [(beam_urn) = "beam:artifact:pypi:v1"];
+
+    // A URN for Java artifacts hosted on Maven central.
```
lukecwik:
```suggestion
    // A URN for Java artifacts hosted on a Maven repository.
```

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -1093,6 +1093,72 @@ message SideInput {
...
+}
+
+message MavenPayload {
+  // Maven compatible group id e.g. "org.apache.beam"
+  string maven_group_id = 1;
+
+  // Maven compatible artifact id e.g. "beam-sdks-java-core"
+  string maven_artifact_id = 2;
+
+  // Maven compatible version string.
```
lukecwik: Grab the documentation for valid version strings/ranges from:
https://docs.oracle.com/middleware/1212/core/MAVEN/maven_version.htm#MAVEN402

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -1093,6 +1093,72 @@ message SideInput {
...
+message PypiPayload {
+  // Pypi compatible artifact id e.g. "apache-beam"
+  string pypi_artifact_id = 1;
+
+  // Pypi compatible version string.
+  string pypi_version_range = 2;
+}
+
+message MavenPayload {
+  // Maven compatible group id e.g. "org.apache.beam"
```
lukecwik: Please add a URL field since a person may want to specify a repository other then maven central and state that if the field is unspecified then maven central is assumed.

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -1093,6 +1093,72 @@ message SideInput {
...
+  string maven_group_id = 1;
```
lukecwik: Note that group/artifact/version isn't enough to fully specify all types of maven artifacts.

Either use a single string to specify the artifact
`groupId:artifactId:version[:packaging[:classifier]]`
where packaging and classifier are optional and packaging by default is "jar" and classifier is omitted if unspecified.

Or add `classifier` and `type` and rename `maven_group_id` to `group_id`, `maven_artifact_id` to `artifact_id` and `maven_version_range` to `version`.

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -1093,6 +1093,72 @@ message SideInput {
...
+
+message EmbeddedFilePayload {
+  // raw data bytes for an embedded artifact
+  bytes data = 1;
+
+  // A generated staged name (no path).
+  string staged_name = 2;
+}
+
+message PypiPayload {
```
lukecwik: nit: `Pypi` -> `PyPI`?

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -1093,6 +1093,72 @@ message SideInput {
...
+  // A generated staged name (no path).
+  string staged_name = 2;
+}
+
+message EmbeddedFilePayload {
+  // raw data bytes for an embedded artifact
+  bytes data = 1;
+
+  // A generated staged name (no path).
+  string staged_name = 2;
```
lukecwik: `staged_name` -> `name`

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -1093,6 +1093,72 @@ message SideInput {
...
+}
+
+message EmbeddedFilePayload {
+  // raw data bytes for an embedded artifact
+  bytes data = 1;
+
+  // A generated staged name (no path).
```
lukecwik: Should this be an actual path where the user says where in the container this file should live?

model/pipeline/src/main/proto/beam_runner_api.proto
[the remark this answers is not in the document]
lukecwik: How about just `path` and we drop `staged_name`?

OR rename ArtifactFilePayload to ArtifactUrlPayload and then users can reference local and remote artifacts?
Using a `url` field which would always have the scheme specified (e.g. local file paths would always have `file://` prefix instead of the assumed conversion from `/foo/bar` to `file:///foo/bar`

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -1093,6 +1093,72 @@ message SideInput {
...
+    // payload: ArtifactFilePayload.
+    FILE     = 0 [(beam_urn) = "beam:artifact:file:v1"];
+
+    // A URN for artifacts embedded in ArtifactInformation proto.
+    // payload: EmbeddedFilePayload.
+    EMBEDDED = 1 [(beam_urn) = "beam:artifact:embedded:v1"];
+
+    // A URN for artifacts described by remote URLs.
+    // payload: a string for an artifact URL e.g. "https://.../foo.jar"
+    REMOTE     = 2 [(beam_urn) = "beam:artifact:remote:v1"];
```
lukecwik: Not needed it we adopt ArtifactUrlPayload

robertwb: Regarding the file payload, I think we may want to stage this under a different name than its "real" name. However, I think the stage name needs to be lifted to a higher level and apply to all artifact types. 

One reason to keep file separate than URL is that we may want to set the space of files be those understood by (and fetched via) the beam filesystems packages (including remote, non-(public?)-url ones like gcs or aws).

model/pipeline/src/main/proto/beam_runner_api.proto
robertwb: +1 to allowing (relative) paths to some stagingDir root.

model/pipeline/src/main/proto/beam_runner_api.proto
robertwb: If there's a standard string format, let's just use that rather than breaking this up.

model/pipeline/src/main/proto/beam_runner_api.proto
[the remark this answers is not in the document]
robertwb: Yes, I'm saying that we should have a new type here for that.

model/pipeline/src/main/proto/beam_runner_api.proto
robertwb: I would call this one URL rather than REMOTE as most of these are remote.

model/pipeline/src/main/proto/beam_runner_api.proto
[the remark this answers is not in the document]
robertwb: Role would be an opaque string. (Well, likely a URN, but the runner wouldn't do anything about it.) It would be provided by, and consumed by, the SDK.

model/pipeline/src/main/proto/beam_runner_api.proto
ihji: Ack.

model/pipeline/src/main/proto/beam_runner_api.proto
ihji: Done.

model/pipeline/src/main/proto/beam_runner_api.proto
ihji: Done.

model/pipeline/src/main/proto/beam_runner_api.proto
ihji: Done.

model/pipeline/src/main/proto/beam_runner_api.proto
ihji: Merged into a single field `artifact_specifier` which expects the format of `groupId:artifactId:version[:packaging[:classifier]]`

model/pipeline/src/main/proto/beam_runner_api.proto
ihji: Ack.

model/pipeline/src/main/proto/beam_runner_api.proto
ihji: `FILE` and `REMOTE` are merged and renamed to `URL`.

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -1093,6 +1093,66 @@ message SideInput {
...
+  string artifact_id = 1;
+
+  // Pypi compatible version string.
+  string version = 2;
+}
+
+message MavenPayload {
+  // A string specifying Maven artifact.
+  // The standard format is "groupId:artifactId:version[:packaging[:classifier]]"
+  string artifact_specifier = 1;
```
lukecwik: nit: `artifact_specifier` -> `artifact`

model/pipeline/src/main/proto/beam_runner_api.proto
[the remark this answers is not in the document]
lukecwik: Do we really believe we will have a mix and match for "source" descriptions and "destination" descriptions?

model/pipeline/src/main/proto/beam_runner_api.proto
robertwb: Is this powerful enough to designate version ranges?

model/pipeline/src/main/proto/beam_runner_api.proto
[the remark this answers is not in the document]
robertwb: Yes. The .tar.gz is one such example. Another example is that we treat the a manually-specified dev version of apache_beam in a special way (currently it *must* be named python_dataflow_sdk.tar); it would be better to not have such magic names. 

SDKs can choose to leave it blank, but I'd rather this be explicit than implicit in the name.

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -1093,6 +1093,66 @@ message SideInput {
...
+  // a string for an artifact URL e.g. "https://.../foo.jar" or "file:///tmp/foo.jar"
+  string path = 1;
+}
+
+message EmbeddedFilePayload {
+  // raw data bytes for an embedded artifact
+  bytes data = 1;
+
+  // A name of the embedded file.
+  string name = 2;
```
robertwb: This is not redundant with the top-level staged_name.

model/pipeline/src/main/proto/beam_runner_api.proto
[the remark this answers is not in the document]
robertwb: We don't have to do that in this PR.

model/pipeline/src/main/proto/beam_runner_api.proto
lukecwik: Yes.

model/pipeline/src/main/proto/beam_runner_api.proto
chamikaramj: you mean "now redundant " ?

ihji:
@robertwb @lukecwik PTAL. Thanks!

model/pipeline/src/main/proto/beam_runner_api.proto
robertwb: Yes. We can still delete it now due to the role.

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -1158,6 +1158,90 @@ message SideInput {
...
+}
+
+message ArtifactUrlPayload {
+  // a string for an artifact URL e.g. "https://.../foo.jar" or "gs://tmp/foo.jar"
+  string path = 1;
+}
+
+message EmbeddedFilePayload {
+  // raw data bytes for an embedded artifact
+  bytes data = 1;
```
robertwb: We could consider just letting the payload be the raw bytes (and similarly for other one-field payloads).

ihji: Done.

ihji:
@robertwb comments addressed.

PR #10739, opened 2020-01-31
[BEAM-9231] Annotate as Experimental/Internal missing classes in beam-sdks-java-core

iemejia:
Also make `SplittableDoFn` classes in `beam-sdks-java-core` Experimental and make Experimental annotations homogeneous.
For easier review take a look better at the separate commits.
R: @lukecwik

### DIFF
--- a/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/External.java
+++ b/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/External.java
@@ -28,6 +28,8 @@ import org.apache.beam.model.expansion.v1.ExpansionApi;
 import org.apache.beam.model.pipeline.v1.Endpoints;
 import org.apache.beam.model.pipeline.v1.RunnerApi;
 import org.apache.beam.sdk.Pipeline;
+import org.apache.beam.sdk.annotations.Experimental;
+import org.apache.beam.sdk.annotations.Experimental.Kind;
 import org.apache.beam.sdk.runners.AppliedPTransform;
 import org.apache.beam.sdk.transforms.Impulse;
 import org.apache.beam.sdk.transforms.PTransform;
@@ -53,6 +55,7 @@ import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.Iterable
  * service. Note that this is a low-level API and mainly for internal use. A user may want to use
  * high-level wrapper classes rather than this one.
  */
+@Experimental(Kind.PORTABILITY)
 public class External {
   private static final String EXPANDED_TRANSFORM_BASE_NAME = "external";
   private static final String IMPULSE_PREFIX = "IMPULSE";
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/harness/JvmInitializer.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/harness/JvmInitializer.java
@@ -18,6 +18,7 @@
 package org.apache.beam.sdk.harness;
 
 import org.apache.beam.sdk.annotations.Experimental;
+import org.apache.beam.sdk.annotations.Internal;
 import org.apache.beam.sdk.options.PipelineOptions;
 
 /**
@@ -33,6 +34,7 @@ import org.apache.beam.sdk.options.PipelineOptions;
  * automate this.
  */
 @Experimental
+@Internal
 public interface JvmInitializer {
 
   /**
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/metrics/DelegatingCounter.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/metrics/DelegatingCounter.java
@@ -17,12 +17,11 @@
  */
 package org.apache.beam.sdk.metrics;
 
-// Intentionally package private, this is meant to be used with Metrics.java and
-// LabeledMetrics.java
-
 import java.io.Serializable;
+import org.apache.beam.sdk.annotations.Internal;
 
 /** Implementation of {@link Counter} that delegates to the instance for the current context. */
+@Internal
 public class DelegatingCounter implements Metric, Counter, Serializable {
   private final MetricName name;
 
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/util/ApiSurface.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/util/ApiSurface.java
@@ -35,6 +35,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Set;
 import java.util.regex.Pattern;
+import org.apache.beam.sdk.annotations.Internal;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Function;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Joiner;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Predicate;
@@ -84,6 +85,7 @@ import org.slf4j.LoggerFactory;
  * not interesting.
  */
 @SuppressWarnings("rawtypes")
+@Internal
 public class ApiSurface {
   private static final Logger LOG = LoggerFactory.getLogger(ApiSurface.class);
 
--- a/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/testing/FakeBigQueryServices.java
+++ b/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/testing/FakeBigQueryServices.java
@@ -26,6 +26,7 @@ import java.io.IOException;
 import java.util.Iterator;
 import java.util.List;
 import org.apache.beam.sdk.annotations.Experimental;
+import org.apache.beam.sdk.annotations.Experimental.Kind;
 import org.apache.beam.sdk.coders.KvCoder;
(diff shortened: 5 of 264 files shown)


iemejia:
Run Java PreCommit

iemejia:
Run Java PreCommit

iemejia:
Run Java PreCommit

lukecwik:
I'm not sure if we need to mark runner libraries or anything outside of beam-sdks-java-core / beam-sdks-java-extensions-* / beam-sdks-java-io-* as `@Experimental`. WDYT?

iemejia:
Agree 100%, I am going to remove the annotations from other places, e.g. `runners/* save for the `External` transform class that is misplaced there. For ref. [BEAM-8546](https://issues.apache.org/jira/browse/BEAM-8546)

iemejia:
I added in the end 2 extra commits. First one just marks everything out of the modules we discussed out of Experimental/Internal and the second commit refines some `beam-sdks-java-core` classes Internal please take attention in detail to that last one in case you see any issue.

iemejia:

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/External.java
```diff
@@ -53,6 +55,7 @@
  * service. Note that this is a low-level API and mainly for internal use. A user may want to use
  * high-level wrapper classes rather than this one.
  */
+@Experimental(Kind.PORTABILITY)
```
lukecwik: remove?

sdks/java/core/src/main/java/org/apache/beam/sdk/harness/JvmInitializer.java
```diff
@@ -33,6 +34,7 @@
  * automate this.
  */
 @Experimental
+@Internal
```
lukecwik: This isn't internal, we expect people to use this and for **workers** to honor it.

sdks/java/core/src/main/java/org/apache/beam/sdk/metrics/DelegatingCounter.java
```diff
@@ -17,12 +17,11 @@
...
 
-// Intentionally package private, this is meant to be used with Metrics.java and
-// LabeledMetrics.java
-
 import java.io.Serializable;
+import org.apache.beam.sdk.annotations.Internal;
 
 /** Implementation of {@link Counter} that delegates to the instance for the current context. */
+@Internal
 public class DelegatingCounter implements Metric, Counter, Serializable {
```
lukecwik: Can we make this package private like the comment said?

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/CombineWithContext.java
```diff
@@ -28,6 +29,7 @@
  *
  * <p>{@link CombineFnWithContext} is for users to extend.
  */
+@Internal
```
lukecwik: I don't believe this is meant to be marked internal

sdks/java/core/src/main/java/org/apache/beam/sdk/util/ApiSurface.java
```diff
@@ -84,6 +85,7 @@
  * not interesting.
  */
 @SuppressWarnings("rawtypes")
+@Internal
```
lukecwik: I believe the package doc states that all these util/* classes are internal but it might be wise to mark them `@Internal` anyways.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/testing/FakeBigQueryServices.java
```diff
@@ -37,7 +38,7 @@
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.ImmutableList;
 
 /** A fake implementation of BigQuery's query service.. */
-@Experimental(Experimental.Kind.SOURCE_SINK)
+@Experimental(Kind.SOURCE_SINK)
```
lukecwik: `@Internal` ?

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/testing/FakeJobService.java
```diff
@@ -87,7 +88,7 @@
 import org.joda.time.Duration;
 
 /** A fake implementation of BigQuery's job service. */
-@Experimental(Experimental.Kind.SOURCE_SINK)
+@Experimental(Kind.SOURCE_SINK)
```
lukecwik: `@Internal` ?

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/External.java
lukecwik: remove since this is runners core?

sdks/java/core/src/main/java/org/apache/beam/sdk/harness/JvmInitializer.java
iemejia: For this PR I assumed that most things were `@Internal` if the audience was runner/harness authors instead of pipeline authors (final users). Is this class expected to be used by the latter? otherwise it still fits the `@Internal` definition like `DoFnInvokers` and the others, no?

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/CombineWithContext.java
iemejia: ok removing it

sdks/java/core/src/main/java/org/apache/beam/sdk/util/ApiSurface.java
iemejia: It does not but I am going to add that to the package. Probably worth to do that too for the schema package with the Experimental tag too.

sdks/java/core/src/main/java/org/apache/beam/sdk/metrics/DelegatingCounter.java
iemejia: Yes I tried to that I think and something failed, let me try again.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/testing/FakeBigQueryServices.java
iemejia: Definitely!

sdks/java/core/src/main/java/org/apache/beam/sdk/metrics/DelegatingCounter.java
iemejia: Oh I found `runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/LabeledMetrics.java` needs it so that's why it cannot be package private.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/testing/FakeJobService.java
iemejia: Yes

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/External.java
iemejia: I prefer to do an exception for this one because of https://issues.apache.org/jira/browse/BEAM-8546. This class is mislocated but nobody has cared about putting it on `sdks/java/core` where it belongs so this is to prevent than once it is moved there the 'experimental' annotation gets forgotten.

iemejia:
Requested fixes done, I added an extra missing Experimental tag for `DoFn.OnTimersContext`). PTAL again @lukecwik

iemejia:
Oups I have some more additions here (+ need to rebase this), will ping you back when ready again.

iemejia:
Ok, added updates to tag Experimental/Internal at the package level and some other minor adjustments, again I suggest to look at the separate commits for an easier review. PTAL again @lukecwik (thanks).

iemejia:
retest this please

iemejia:
Run Seed Job

iemejia:
Run Java PreCommit

iemejia:
All green again. Can you PTAL @lukecwik

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/External.java
lukecwik: sgtm

sdks/java/core/src/main/java/org/apache/beam/sdk/harness/JvmInitializer.java
lukecwik: We expect pipeline authors to write JVmInitializer and for the worker to use ServiceLoader to find them and use them.

iemejia:
Thanks for the review Luke! I merged the PR manually to remove @Internal from `JvmInitializer`. One final question. What is the reason this class belongs in core instead of in harness?

lukecwik:
`harness` is not meant to be depended on by users as its the main entry point for execution and pipeline authors have regularly requested the ability to be able to modify JVM parameters early on before their DoFn's or other parts of the system initialize.

PR #10741, opened 2020-01-31
[BEAM-9236] Mark missing Schema based classes and methods as Experimental

iemejia:
R: @alexvanboxel 
CC: @reuvenlax (so you'd be aware of the removals, I would revert if there is a reason to still keep those classes).


iemejia:
Run Java PreCommit

alexvanboxel:
Except for the **errors** (public -> package protected) it looks good.

iemejia:
Run Java PreCommit

iemejia:
Run Java PreCommit

iemejia:
Run Java PreCommit

iemejia:
@alexvanboxel thanks for the review. I just fixed the access issues, can you take a second look and approve this PR + LGTM if you agree.

alexvanboxel:
LGTM, it's indeed still experimental as I see a it changing a lot.

reuvenlax:
LGTM. I think we're close to discussing removing the Experimental tag from schemas, as the core APIs haven't really changed much in a while. However there are a few usability improvements I think we can make that are low-hanging fruit, so I think we should keep this tag for now.

alexvanboxel:
@reuvenlax I'm interested in getting into the loop about those discussions about the usability improvements. I've got some remarks as well on the asymmetry of the Row interface and it could be interesting to group those discussion so I can land a hand.

reuvenlax:
@alexvanboxel The Row class is one of the classes that could use the most improvement - especially the Row builder. I also don't like that it's easy to build a Row with the "wrong" schema, and this won't be caught until runtime (usually with very weird and confusing error messages).

iemejia:
+1 on having more awareness on usability improvement ideas where more members of the community contribute on.
For the removal of Experimental we should probably once we decide to remove it let it 'bake still one or two releases where we prove that we did not break backwards compatiblity' and then go, in the meantime better to be clear with the users.

alexvanboxel:
It would be grate if we have the options (see proposal and PR) baked in and the logical types stabilised as well. Would be great if some people could jump in on this. As I'm just a contributor I can 't even merge so it's a slow experience... @iemejia can I add you to schema aware PR's?

iemejia:
Sure, count @aromanenko-dev in too who has been working on Schema related improvements for KafkaIO and is interested in Schema-related subjects.

PR #10763, opened 2020-02-04
[BEAM-9236] Mark missing Schema based classes and methods as Experimental (2)

iemejia:
Looks like I forgot some classes, see this as part 2 of #10741

R: @alexvanboxel
CC: @reuvenlax (so you'd be aware of the removals, I would revert if there is a reason to still keep those classes).


alexvanboxel:
LGTM

iemejia:
Thanks @alexvanboxel !

PR #10764, opened 2020-02-04
[BEAM-9146] Integrate GCP Video Intelligence functionality for Python SDK

EDjur:
This PR relates to https://issues.apache.org/jira/browse/BEAM-9146 and integrates GCP Video Intelligence functionality as part of a PTransform that implicitly accepts a PCollection of GCS URIs or raw bytes.

------------------------

--- | --- | --- | --- | ---


EDjur:
R: @aaltay @kamilwu

aaltay:
Thank you @EDjur. 

I was hoping that we can implement most of the functionality in tfx_bsl, and add cloud AI services by defining new endpoints. We had a discussion on dev list [1] related to this.

[1] https://lists.apache.org/thread.html/r3d9b0e3270668fda989f363eab1f79e4ef0e7fa3e0fcbe9c26344f14%40%3Cdev.beam.apache.org%3E

EDjur:
I see. I signed up to the dev mailing list just last week so must've missed that discussion.

What does that mean in regards to this ticket?

I agree that it would be nice to avoid duplicated behaviour across libraries.

aaltay:
@EDjur - Sorry for the miscommunication. These types of decisions should have reflected on JIRA as well. I feel bad that you spent time on this and we are changing direction after your PR is out.

I think what it means for this ticket is that:
- If tfx_bsl has ability to call into services (which we need to add and TFX team agreed to support us with reviews)
- We can add thin transforms for different things (e.g. AnnotateVideo) and each of these transform will have quite a bit of shared code and behavior.

@kamilwu may have other thoughts, since he was also working on this. I would like to hear what is his opinion.

For this PR, let's try to re-use as much as possible.

EDjur:
@aaltay No worries - I've learnt a lot by creating this PR anyway! Curious to hear @kamilwu's thoughts on it too.

I'm open to modify / work on another PR related to tfx_bsl if needed.

kamilwu:
Thanks @EDjur. 

`tfx_bsl` seems to be useful only in some cases. One of these cases is [AI Platform Prediction](https://cloud.google.com/ai-platform/prediction/docs/overview), which, by the way, we also would like to support in Beam in the nearest future.

Video Intelligence API is pretty much different. We should treat it as a fully-managed machine learning service. There is no machine learning model which could be trained and deployed. Instead, videos are annotated using pre-trained custom model provided by Google. All a user has to do is to enable API and make a request (using REST API or client library). The only input is a GCS path to a video file. As a result, there is little behavior that could be shared.

If you have any questions regarding tfx_bsl and how it connects with Beam, just let me know, as I'm currently working to provide support for remote inference (an ability to call into services) in tfx_bsl.

kamilwu:
@EDjur Because you've added a new dependency, we have to specify it in setup.py file (`sdks/python/setup.py`)

sdks/python/apache_beam/io/gcp/ai/video_intelligence.py
```diff
@@ -0,0 +1,75 @@
...
+from apache_beam.metrics import Metrics
+from apache_beam.transforms import DoFn, ParDo, PTransform
+
+
+class AnnotateVideo(PTransform):
+  """A ``PTransform`` for annotating video using the GCP Video Intelligence API
+  ref: https://cloud.google.com/video-intelligence/docs
+  """
+
+  def __init__(self, features):
```
kamilwu: I think we should also accept other parameters, not only `features`.

A must have is `video_context`, because sometimes it provides additional feature-specific parameters. Others are `location_id` and possibly `metadata`. Take a look at the interface of `annotate_video()`

EDjur: Good idea, thanks! I'll have a look after we've finished the discussion with @aaltay about the usefulness of this PR.

sdks/python/apache_beam/io/gcp/ai/video_intelligence.py
```diff
@@ -0,0 +1,75 @@
...
+      super(AnnotateVideo._VideoAnnotateFn, self).__init__()
+      self._client = None
+      self.features = features
+      self.counter = Metrics.counter(self.__class__, "API Calls")
+
+    def start_bundle(self):
+      self._client = helper.get_videointelligence_client()
+
+    def process(self, element, *args, **kwargs):
+      if isinstance(element, str):  # Is element an URI to a GCS bucket
```
kamilwu: In Python 2 `str` and `bytes` are the same. Since Beam still supports Python 2.7, this code must be compatible with both Python 2 and 3.
Consider using `future.utils` module which provides some methods that can help you to check type of an element.

aaltay:
Thank you @kamilwu for the comment. That makes sense to me.

Related question, what is a good location for these family of transforms? `io/...` does not sounds right in this case. How about a top level `ml/` folder similar to `io`. These transform could be put under `ml/gcp`?

kamilwu:
> Related question, what is a good location for these family of transforms? io/... does not sounds right in this case. How about a top level ml/ folder similar to io. These transform could be put under ml/gcp?

I agree `io/...` can be somewhat misleading. `ml/gcp` sounds good.

EDjur:
Thanks for the feedback! I've adjusted the code to conform to py23 standards as well as added the extra arguments to annotate_video.

I also added the `google-cloud-videointelligence` dependency which I've tested from version `1.8.0` up until `1.12.1`.

sdks/python/apache_beam/ml/gcp/video_intelligence.py
```diff
@@ -0,0 +1,107 @@
...
+            video_context=self.video_context,
+            location_id=self.location_id,
+            metadata=self.metadata)
+      else:
+        raise TypeError(
+            "{}: input element needs to be either {} or {}"
+            " got {} instead".format(
+                self.__class__.__name__, text_type, binary_type, type(element)))
+      self.counter.inc()
+      yield response.result(timeout=120)
```
aaltay: Would it make sense to make 120 configurable?

sdks/python/apache_beam/ml/gcp/video_intelligence.py
```diff
@@ -0,0 +1,107 @@
...
+    return pvalue | ParDo(
+        self._VideoAnnotateFn(
+            features=self.features,
+            video_context=self.video_context,
+            location_id=self.location_id,
+            metadata=self.metadata))
+
+  @typehints.with_input_types(Union[text_type, binary_type])
+  class _VideoAnnotateFn(DoFn):
+    """ A ``DoFn`` that sends every element to the GCP Video Intelligence API
```
aaltay: DoFn does not return a PCollection. Maybe we can update this to something like

A DoFn that sends each input element to the GCP Video Intelligence API service and outputs an element with the return result of the API.

sdks/python/apache_beam/ml/gcp/video_intelligence_helper.py
```diff
@@ -0,0 +1,36 @@
...
+
+For internal use only; no backwards-compatibility guarantees.
+"""
+
+# pytype: skip-file
+
+from __future__ import absolute_import
+
+from cachetools.func import ttl_cache
+from google.cloud import videointelligence
```
aaltay: Should we guard this similar to other optional dependencies (example: https://github.com/apache/beam/blob/master/sdks/python/apache_beam/io/gcp/gcsio.py#L53)

sdks/python/apache_beam/ml/gcp/video_intelligence_helper.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+@ttl_cache(maxsize=128, ttl=3600)
+def get_videointelligence_client():
```
aaltay: Are we expecting to use this for other APIs as well?

sdks/python/apache_beam/ml/gcp/video_intelligence.py
```diff
@@ -0,0 +1,107 @@
...
+      self.video_context = video_context
+      self.location_id = location_id
+      self.metadata = metadata
+      self.counter = Metrics.counter(self.__class__, "API Calls")
+
+    def start_bundle(self):
+      self._client = helper.get_videointelligence_client()
+
+    def process(self, element, *args, **kwargs):
+      if isinstance(element, text_type):  # Is element an URI to a GCS bucket
```
aaltay: Does the API have support for batching? If yes, it might be a good feature to support.

sdks/python/apache_beam/ml/gcp/video_intelligence.py
EDjur: Possibly. I haven't used the API enough to say for sure, but we could make it configurable.

sdks/python/apache_beam/ml/gcp/video_intelligence.py
EDjur: Gotcha, thanks!

sdks/python/apache_beam/ml/gcp/video_intelligence_helper.py
EDjur: I don't think so no, so it would make sense to merge this code into the main videointelligence file.

sdks/python/apache_beam/ml/gcp/video_intelligence.py
EDjur: As far as I can tell, it does not support batching https://cloud.google.com/video-intelligence/docs/reference/rest/v1/videos/annotate. But I may be wrong.

EDjur:
Got one question regarding the naming. Is `video_intelligence` a good name for the module? I'm slightly worried it might cause confusion with the `google.cloud.videointelligence`.

sdks/python/apache_beam/ml/gcp/video_intelligence.py
aaltay: Sounds good. Thank you for checking.

aaltay:
> Got one question regarding the naming. Is `video_intelligence` a good name for the module? I'm slightly worried it might cause confusion with the `google.cloud.videointelligence`.

I think it is a good name reflects the underlying service as it is. Not very different from other gcp io with the similar names to the gcp services.

sdks/python/apache_beam/ml/gcp/video_intelligence.py
aaltay: I think we can make it a default argument to transform with the same default. It would be up to the users to change if they need to.

sdks/python/apache_beam/ml/gcp/video_intelligence.py
```diff
@@ -0,0 +1,131 @@
...
+            metadata=self.metadata,
+            timeout=self.timeout))
+
+  @typehints.with_input_types(Union[text_type, binary_type])
+  class _VideoAnnotateFn(DoFn):
+    """ A DoFn that sends each input element to the GCP Video Intelligence API
+        service and outputs an element with the return result of the API
+        (``google.cloud.videointelligence_v1.types.AnnotateVideoResponse``).
+     """
+    def __init__(self, features, video_context, location_id, metadata, timeout):
```
aaltay: We can default the timeout to 120 here. If you think that is a reasonable default.

kamilwu:
LGTM.

EDjur:
> > Got one question regarding the naming. Is `video_intelligence` a good name for the module? I'm slightly worried it might cause confusion with the `google.cloud.videointelligence`.
> 
> I think it is a good name reflects the underlying service as it is. Not very different from other gcp io with the similar names to the gcp services.

I started work on https://issues.apache.org/jira/browse/BEAM-9247 to integrate the vision API in a similar PTransform. However, here the naming gets more complex. Consider e.g.
```
  from google.cloud import vision
  from apache_beam.ml.gcp import vision
```

One solution for the `vision` transforms is to rename the module `vision_api` instead. But I think we should be consistent across the different GCP ML APIs, so perhaps `video_intelligence` will need renaming too.

What's your take?

aaltay:
> > > Got one question regarding the naming. Is `video_intelligence` a good name for the module? I'm slightly worried it might cause confusion with the `google.cloud.videointelligence`.
> > 
> > 
> > I think it is a good name reflects the underlying service as it is. Not very different from other gcp io with the similar names to the gcp services.
> 
> I started work on https://issues.apache.org/jira/browse/BEAM-9247 to integrate the vision API in a similar PTransform. However, here the naming gets more complex. Consider e.g.
> 
> ```
>   from google.cloud import vision
>   from apache_beam.ml.gcp import vision
> ```
> 
> One solution for the `vision` transforms is to rename the module `vision_api` instead. But I think we should be consistent across the different GCP ML APIs, so perhaps `video_intelligence` will need renaming too.
> 
> What's your take? @aaltay

I do not have a good idea. Users could solve this with something like the following:

```
from google.cloud import vision as gcp_vision_api
from apache_beam.ml.gcp import vision
```

So, maybe it is not a big issue. I agree, it would be good if we can avoid this state.

Related, in the io folder we have modules named `gcsio`, `bigtableio` etc. but we also have modules named `bigquery`, `pubsub`. So it is not very consistent.

We can take the io example and change the names by appending ml, your example will look like:

```
from google.cloud import vision
from apache_beam.ml.gcp import visionml
from apache_beam.ml.gcp import videointelligenceml (Notice I also dropped _ between the words.)
```

What do you think?

EDjur:
I like the idea of appending ml to the modules. Dropping the _ also seems prudent so it's consistent with most of io modules.

aaltay:
LGTM. Triggering tests.

EDjur:
Run Python PreCommit

EDjur:
Seems tests aren't running properly in Jenkins. But they all pass on my local machine. I'll let you re-trigger them and hope that it works.

kamilwu:
> Seems tests aren't running properly in Jenkins.

At the moment, only Beam commiters have permissions to trigger Jenkins jobs. But you can always ask someone on devlist to trigger them for you.

aaltay:
Could you also add a new feature note in https://github.com/apache/beam/blob/master/CHANGES.md

aaltay:
tests, will you trigger?

aaltay:
@EDjur - This LGTM. I would like to merge it but I cannot seem to trigger the tests

@markflyhigh is there a manual way to run the tests?

mwalenia:
retest this please

aaltay:
retest this please

aaltay:
Finally, tests are running.

EDjur:
retest this please

EDjur:
Some minor lint issues should be fixed with that last commit. But it seemed Jenkins was failing on something else: `failed to import pycodestyle` when running the tests. Not sure that is related to this PR.

aaltay:
> Some minor lint issues should be fixed with that last commit. But it seemed Jenkins was failing on something else: `ModuleNotFoundError: No module named 'pycodestyle'` when running the tests. Not sure if that is related to this PR.

This is fixed in (https://github.com/apache/beam/pull/10844). You may need to rebase.

EDjur:
retest this please

aaltay:
retest this please

mwalenia:
retest this please

EDjur:
I cannot seem to make the tests fail locally so this is proving to be quite a struggle 🙄 But hopefully this should fix it.

EDjur:
retest this please

mwalenia:
retest this please

mwalenia:
retest this please

EDjur:
retest this please

mwalenia:
retest this please

mwalenia:
@EDjur your phrase triggers unfortunately won't trigger the tests - there are Jenkins restrictions that won't let you do it

EDjur:
> @EDjur your phrase triggers unfortunately won't trigger the tests - there are Jenkins restrictions that won't let you do it

Gotcha. Thanks for triggering them!

mwalenia:
retest this please

EDjur:
I'm having a small discussion regarding the `video_context` with @kamilwu here: https://issues.apache.org/jira/browse/BEAM-9247

Perhaps this is something we should look at integrating to this PR before (or after considering the use-case seems small) merging too.

mwalenia:
retest this please

EDjur:
Test failures look unrelated to this PR.

```
11:04:57 self = <apache_beam.runners.interactive.caching.streaming_cache_test.InMemoryReader object at 0x7fdcc6b742d0>
11:04:57 watermark = 0, processing_time = 1
11:04:57 
11:04:57     def advance_watermark(self, watermark, processing_time):
11:04:57       record = TestStreamFileRecord(
11:04:57           watermark=Timestamp.of(watermark).to_proto(),
11:04:57 >         processing_time=Timestamp.of(processing_time).to_proto())
11:04:57 E     ValueError: Protocol message TestStreamFileRecord has no "processing_time" field.
```

aaltay:
Test error seems to be related to: https://github.com/apache/beam/pull/10856/files

aaltay:
retest this please

aaltay:
retest this please

aaltay:
retest this please

aaltay:
retest this please

aaltay:
:sdks:python:test-suites:tox:pycommon:docs  task is failing -- There might be pydocs issues in the change.

EDjur:
Seems yapf doesn't format docstrings 🤷‍♂  Should be fixed now.

kkucharc:
retest this please

kkucharc:
retest this please

kkucharc:
retest this please

aaltay:
@EDjur -- Thank you, merged this.

Could we update 'google-cloud-videointelligence>=1.8.0<=1.12.1', to use the recently released 1.13.0 version as well? Do we need anything special?

EDjur:
Looks fine to update! The 1.13.0 essentially added two new enums to the types.Feature, but since these are specified by the user, all should be well.

I also manually reran all the tests without mocking the videointelligence client and they ran as expected.

Cheers!

PR #10766, opened 2020-02-04
[BEAM-4461] Add Selected.flattenedSchema

reuvenlax:
R: @TheNeuralBit


sdks/java/core/src/test/java/org/apache/beam/sdk/schemas/utils/SchemaTestUtils.java
```diff
@@ -90,4 +91,71 @@ public void describeTo(Description description) {
       description.appendText(expected.toString());
     }
   }
+
+  public static class RowFieldMatcher extends BaseMatcher<Row> {
```
TheNeuralBit: This looks nifty, but maybe made it into this PR by accident? I don't see it used anywhere

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/transforms/Unnest.java
```diff
@@ -1,171 +0,0 @@
...
- * <p>UserEvent Schema: userid: INT64 timestamp: DATETIME location.latitude: DOUBLE
- * location.longitude: DOUBLE
- *
- * <p>By default nested names are concatenated to generated the unnested name, however {@link
- * Unnest.Inner#withFieldNameFunction} can be used to specify a custom naming policy.
- *
- * <p>Note that currently array and map values are not unnested.
- */
-@Experimental(Kind.SCHEMAS)
-public class Unnest {
```
TheNeuralBit: I assume there's no concern with removing Unnest outright (rather than deprecating for a period and pointing to `Select.flattenedSchema`) since this is all Experimental?

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/FieldAccessDescriptor.java
```diff
@@ -300,6 +340,16 @@ public FieldAccessDescriptor withNestedField(
...
+  public FieldAccessDescriptor withNestedFieldAs(
+      String nestedFieldName, String nestedFieldRename, FieldAccessDescriptor fieldAccess) {
+    FieldDescriptor field =
+        FieldDescriptor.builder()
+            .setFieldName(nestedFieldName)
+            .setFieldRename(nestedFieldRename)
+            .build();
+    return withNestedField(field, fieldAccess);
+  }
+
```
TheNeuralBit: Consider adding a test for this in `FieldAccessDescriptorTest`, and adding a docstring

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/transforms/Select.java
```diff
@@ -69,55 +75,117 @@
...
+  }
+
+  private static class SelectDoFn<T> extends DoFn<T, Row> {
+    private FieldAccessDescriptor fieldAccessDescriptor;
+    private Schema inputSchema;
+    private Schema outputSchema;
+
+    // TODO: This should be the same as resolved so that Beam knows which fields
+    // are being accessed. Currently Beam only supports wildcard descriptors.
+    // Once BEAM-4457 is fixed, fix this.
```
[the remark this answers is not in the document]
reuvenlax: This is an old comment. I don't know that we've fully decided what to do about BEAM-4457

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/transforms/Select.java
```diff
@@ -69,55 +75,117 @@
...
+      this(CONCAT_FIELD_NAMES);
+    }
+
+    Flattened(SerializableFunction<List<String>, String> nameFn) {
+      this.nameFn = nameFn;
+    }
+
+    /**
+     * For nested fields, concatenate all the names separated by a _ character in the flattened
+     * schema.
```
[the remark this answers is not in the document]
reuvenlax: yeah, though I think we should maybe make underscore a reserved character in fields. We also don't prevent users from putting dots in field names, but all sorts of things will break if they do.

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/transforms/Unnest.java
reuvenlax: yeah - should be fine I think.

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/transforms/Select.java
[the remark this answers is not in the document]
reuvenlax: Added withFieldNameAs to allow the user to rename fields

reuvenlax:
retest this please

reuvenlax:
retest this please

reuvenlax:
retest this please

PR #10767, opened 2020-02-04
Document Beam Schemas

reuvenlax:


reuvenlax:
R: @TheNeuralBit

reuvenlax:
retest this please

reuvenlax:
retest this please

rezarokni:
In 6.5.3. Schemas in ParDo, it would useful to have an example of how you would output a additive property to the object. For example using AddFields followed by convert to a POJO with those new fields, followed by assignment of a value to that field. 
Example use case: Following an aggregation done with Group, we want to enrich the new Row object with other Metadata before output to storage.

reuvenlax:
We already have an AddFields snippet - does that not do what you want?

rezarokni:
Its missing the connection between that and creation of a separate POJO before you can update the field with a non-default value I think...

reuvenlax:
@TheNeuralBit friendly ping

website/src/documentation/programming-guide.md
```diff
@@ -1970,7 +1976,1076 @@ records.apply("WriteToText",
...
+In order to select a field at the top level of a schema, the name of the field is specified. For example, to select just
+the user ids from a `PCollection` of purchases one would write (using the `Select` transform)
+
+```java
+purchases.apply(Select.fieldNames(“userId”));
+```
+
+##### **Nested fields**
+Individual nested fields can be specified using the dot operator. For example, to select just the postal code from the
+ shipping address one would write
```
TheNeuralBit: Might be a good place to address the question of dots in field names. Do we disallow it? Or do we allow it and it just breaks nested field selection?

website/src/documentation/programming-guide.md
```diff
@@ -1970,7 +1976,1076 @@ records.apply("WriteToText",
...
+While the use of  [] brackets in the selector is recommended, to make it clear that array elements are being selected, 
+they can be omitted for brevity. In the future, array slicing will be supported, allowing selection of portions of the 
+array.
+
+##### **Maps**
+A map field, where the value type is a row, can also have subfields of the value type addressed. When selected, the 
+result is a map where the keys are the same as in the original map but the value is the specified type. Similar to 
+arrays, the use of {} curly brackets in the selector is recommended, to make it clear that map value elements are being 
+selected, they can be omitted for brevity. In the future, map key selectors will be supported, allowing selection of 
+specific keys from the map.
```
TheNeuralBit: An example like the one for arrays would be helpful here.

website/src/documentation/programming-guide.md
```diff
@@ -1970,7 +1976,1076 @@ records.apply("WriteToText",
...
+```java
+purchases.apply(Select.fieldNames(“transactions[].bank”));
+```
+
+Will result in a row containing an array field with element-type string, containing the list of banks for each 
+transaction. 
+
+While the use of  [] brackets in the selector is recommended, to make it clear that array elements are being selected, 
+they can be omitted for brevity. In the future, array slicing will be supported, allowing selection of portions of the 
+array.
```
TheNeuralBit: What do you think about linking JIRAs for these "in the future" items?

website/src/documentation/programming-guide.md
TheNeuralBit: Similarly for `[]{}*`, do we consider all of these special characters?

website/src/documentation/programming-guide.md
```diff
@@ -1970,7 +1976,1076 @@ records.apply("WriteToText",
...
+If the source type is a single-field schema, Convert will also convert to the type of the field if asked, effectively
+unboxing the row. For example, give a schema with a single INT64 field, the following will convert it to a
+`PCollection<Long>`
+
+```java
+PCollection<Long> longs = rows.apply(Convert.to(TypeDescriptors.longs()));
+```
+
+In all cases, type checking is done at pipeline graph construction, and if the types do not match the schema then the
+pipeline will fail to launch.
```
TheNeuralBit: What about adding an example of interop with `SqlTransform` mid-pipeline? Or do we not want to encourage that?

website/src/documentation/programming-guide.md
reuvenlax: Added (we should add code to validate that these characters aren't used!)

website/src/documentation/programming-guide.md
```diff
@@ -1970,7 +1976,1076 @@ records.apply("WriteToText",
...
+public abstract class TransactionValue {
+  public abstract String getBank(); 
+  public abstract double getPurchaseAmount();
+}
+```
+
+This is all that’s needed to generate a simple AutoValue class, and the above `@DefaultSchema` annotation tells Beam to
+infer a schema from it. This also allows AutoValue elements to be used inside of `PCollection`s.
+
+`@SchemaFieldName` and `@SchemaIgnore` can be used to alter the schema inferred.
```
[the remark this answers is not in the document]
reuvenlax: yeah, good feedback for docs team?

website/src/documentation/programming-guide.md
```diff
@@ -1970,7 +1976,1076 @@ records.apply("WriteToText",
 See the [Beam-provided I/O Transforms]({{site.baseurl }}/documentation/io/built-in/)
 page for a list of the currently available I/O transforms.
 
-## 6. Data encoding and type safety {#data-encoding-and-type-safety}
+## 6. Schemas {#schemas}
+Often, the type of records being processed have an obvious structure. Common Beam sources produce
+JSON, Avro, Protocol Buffer, or database row objects; all of these types have well defined structures, 
+structures that can often be determined by examining the type. Even within a pipeline, Simple Java POJOs 
+(or  equivalent structures in other languages) are often used as intermediate types, and these also have a
```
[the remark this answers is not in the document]
reuvenlax: I think better to wait until we add Python schema docs

website/src/documentation/programming-guide.md
```diff
@@ -1970,7 +1976,1076 @@ records.apply("WriteToText",
 See the [Beam-provided I/O Transforms]({{site.baseurl }}/documentation/io/built-in/)
 page for a list of the currently available I/O transforms.
 
-## 6. Data encoding and type safety {#data-encoding-and-type-safety}
+## 6. Schemas {#schemas}
+Often, the type of records being processed have an obvious structure. Common Beam sources produce
+JSON, Avro, Protocol Buffer, or database row objects; all of these types have well defined structures, 
+structures that can often be determined by examining the type. Even within a pipeline, Simple Java POJOs
```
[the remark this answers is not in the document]
reuvenlax: I meant within a pipeline as that's where users create these intermediate objects - maybe SDK pipeline?

website/src/documentation/programming-guide.md
```diff
@@ -1970,7 +1976,1076 @@ records.apply("WriteToText",
...
+      <td>cost</td>
+      <td>INT64</td>      
+    </tr>
+    <tr>
+      <td>transactions</td>
+      <td>ARRAY[ROW(Transaction)]</td>      
+    </tr>                  
+  </tbody>
+</table>
+<br/>
```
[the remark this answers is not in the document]
reuvenlax: This is true for other tables in this doc as well (e.g. coders). Maybe we should figure out a way to fix all of them at once?

website/src/documentation/programming-guide.md
```diff
@@ -1970,7 +1976,1076 @@ records.apply("WriteToText",
...
+    </tr>
+    <tr>
+      <td>purchaseAmount</td>
+      <td>DOUBLE</td>      
+    </tr>                  
+  </tbody>
+</table>
+<br/>
+
+Purchase event records are represented by the aove purchase schema. Each purchase event contains a shipping address, which
```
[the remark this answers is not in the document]
reuvenlax: done

website/src/documentation/programming-guide.md
reuvenlax: I think it's a fine example, but let's hold off. we will need a followup PR that adds a bunch of examples anyway

website/src/documentation/programming-guide.md
```diff
@@ -1970,7 +1976,1076 @@ records.apply("WriteToText",
...
+
+Schemas provide us a type-system for Beam records that is independent of any specific programming-language type. There
+might be multiple Java classes that all have the same schema (for example a Protocol-Buffer class or a POJO class),
+and Beam will allow us to seamlessly convert between these types. Schemas also provide a simple way to reason about 
+types across different programming-language APIs.
+
+A `PCollection` with a schema does not need to have a `Coder` specified, as Beam knows how to encode and decode 
+Schema rows.
+
+### 6.2. Schemas for programming language types {#schemas-for-pl-types}
```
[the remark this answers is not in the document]
reuvenlax: hmmm.... this section is trying to get across the basic theory that schemas are _the_ types, and if multiple Pl types have the same schema then they are interchangeable. The goal of this section isn't a reference, though we use some examples to illustrate the point. The other two sections you mention are meant to be more of a reference.

website/src/documentation/programming-guide.md
```diff
@@ -1970,7 +1976,1076 @@ records.apply("WriteToText",
...
+
+and
+```java
+transactionPojos.apply(ParDo.of(new DoFn<...>() {
+   @ProcessElement public void process(@Element Transaction row) {
+    }
+}));
+```
+
+EEven though the in both cases the `@Element` parameter differs from the the `PCollection`'s Java type, since the
```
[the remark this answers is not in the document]
reuvenlax: done

website/src/documentation/programming-guide.md
```diff
@@ -1970,7 +1976,1076 @@ records.apply("WriteToText",
...
+  @SchemaCreate
+  Public TransactionBean(String bank, double purchaseAmount) { … }
+  public String getBank() { … }
+  public double getPurchaseAmount() { … }
+}
+```
+
+`@SchemaFieldName` and `@SchemaIgnore` can be used to alter the schema inferred, just like with POJO classes.
+
+##### AutoValue
```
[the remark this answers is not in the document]
reuvenlax: done

website/src/documentation/programming-guide.md
reuvenlax: unless there's precedent for linking JIRAs in the docs, I'd rather not.

reuvenlax:
@TheNeuralBit any more comments?

website/src/documentation/programming-guide.md
```diff
@@ -1970,7 +1976,1078 @@ records.apply("WriteToText",
...
+
+Three collection types are supported as field types: ARRAY, ITERABLE and MAP:
+* **ARRAY** This represents a repeated value type, where the repeated elements can have any supported type. Arrays of 
+nested rows are supported, as are arrays of arrays.
+* **ITERABLE** This is very similar to the array type, it represents a repeated value, but one in which the full list of 
+items is not known until iterated over. This is intended for the case where an iterable might be larger than the 
+available memory, and backed by external storage (for example, this can happen with the iterable returned by a 
+`GroupByKey`). The repeated elements can have any supported type.
+* **MAP** This represents an associative map from keys to values. All schema types are supported for both keys and values.
+ Values that contain map types cannot be used as keys in any grouping operation.
```
iemejia: Just for curiosity what is the reason for this contraint? Is it because keys order is not guaranteed?

website/src/documentation/programming-guide.md
```diff
@@ -1970,7 +1976,1078 @@ records.apply("WriteToText",
...
+below.
+
+#### 6.4.1. Defining a logical type {#defining-a-logical-type}
+To define a logical type you must specify a Schema type to be used to represent the underlying type as well as a unique
+identifier for that type. A logical type imposes additional semantics on top a schema type. For example, a logical 
+type to represent nanosecond timestamps is represented as a schema containing an INT64 and an INT32 field. This schema
+alone does not say anything about how to interpret this type, however the logical type tells you that this represents
+a nanosecond timestamp, with the INT64 field representing seconds and the INT32 field representing nanoseconds.
+
+Logical types are also specified by an argument, which allows creating a class of related typed. For example, a
```
iemejia: `which allows to create a class of related types` ?

website/src/documentation/programming-guide.md
```diff
@@ -1970,7 +1976,1078 @@ records.apply("WriteToText",
...
+    return Instant.of(row.getInt64("seconds"), row.getInt32("nanos"));
+  }
+
+     ...
+}
+```
+
+#### 6.4.2. Useful logical types {#built-in-logical-types}
+##### **EnumerationType**
+EnumerationType allows creating an enumeration type consisting of a set of named constants.
```
iemejia: `This logical type allows to create .an enum...`?

website/src/documentation/programming-guide.md
```diff
@@ -1970,7 +1976,1078 @@ records.apply("WriteToText",
...
+assuming that constructor parameters have the same names as the field names. `@SchemaCreate` can also be used to annotate
+static factory methods on the class, allowing the constructor to remain private. If there is no `@SchemaCreate`
+ annotation then all the fields must be non-final and the class must have a zero-argument constructor.
+
+There are a couple of other useful annotations that affect how Beam infers schemas. By default the schema field names 
+inferred will match that of the class field names. However `@SchemaFieldName` can be used to specify a different name to
+be used for the schema field. `@SchemaIgnore` can be used to mark specific class fields as excluded from the inferred
+schema. For example, it’s common to have ephemeral fields in a class that should not be included in a schema 
+(e.g. caching the hash value to prevent expensive recomputation of the hash), and `@SchemaIgnore` can be used to
+exclude these fields.
```
iemejia: Maybe is worth to clarify here if `@SchemaIgnore`d values are not trimmed from the records encoding (if so), only from the schema.

website/src/documentation/programming-guide.md
```diff
@@ -1970,7 +1976,1078 @@ records.apply("WriteToText",
...
+  @SchemaCreate
+  Public TransactionBean(String bank, double purchaseAmount) { … }
+  public String getBank() { … }
+  public double getPurchaseAmount() { … }
+}
+```
+
+`@SchemaFieldName` and `@SchemaIgnore` can be used to alter the schema inferred, just like with POJO classes.
+
+##### **AutoValue(())
```
iemejia: This title does not render properly

website/src/documentation/programming-guide.md
```diff
@@ -1970,7 +1976,1078 @@ records.apply("WriteToText",
...
+Java value classes are notoriously difficult to generate correctly. There is a lot of boilerplate you must create in 
+order to properly implement a value class. AutoValue is a popular framework for easily generating such classes by i
```
iemejia: s/framework/library

website/src/documentation/programming-guide.md
```diff
@@ -1970,7 +1976,1078 @@ records.apply("WriteToText",
...
+  public abstract String getBank(); 
+  public abstract double getPurchaseAmount();
+}
+```
+
+This is all that’s needed to generate a simple AutoValue class, and the above `@DefaultSchema` annotation tells Beam to
+infer a schema from it. This also allows AutoValue elements to be used inside of `PCollection`s.
+
+`@SchemaFieldName` and `@SchemaIgnore` can be used to alter the schema inferred.
+
```
iemejia: Can we add Avro here too?

website/src/documentation/programming-guide.md
```diff
@@ -1970,7 +1976,1078 @@ records.apply("WriteToText",
...
+### 6.6. Using Schemas {#using-schemas}
```
iemejia: Using Schema based PTransforms maybe? You are already talking about schema ptransforms here so better inline  6.6.1 as part of 6.6 and make 6.6.1, 6.6.2, ... for each transform type

TheNeuralBit:

website/src/documentation/programming-guide.md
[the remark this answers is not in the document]
TheNeuralBit: Perhaps it's a moot point with the [switch to docsy](https://lists.apache.org/thread.html/r7fa6d710c0a1959cce5108e460d71c306ce5756cf96af818b41cb7ca%40%3Cdev.beam.apache.org%3E) going on

website/src/documentation/programming-guide.md
```diff
@@ -1970,7 +1976,1078 @@ records.apply("WriteToText",
...
+```
+
+Even though the in both cases the `@Element` parameter differs from the the `PCollection`'s Java type, since the
+schemas are the same Beam will automatically make the conversion. The built-in `Convert` transform can also be used
+to translate between Java types of equivalent schemas, as detailed below.
+
+### 6.3. Schema definition {#schema-definition}
+The schema for a `PCollection` defines elements of that `PCollection` as an ordered list of named fields. Each field
+has a name, a type, and possibly a set of user options. The type of a field can be primitive or composite. The following
+are the primitive types currently supported by Beam:
```
TheNeuralBit: Are there any specific limitations for field names we want to document here? Many special characters (`$`, `-`, ...) make it impossible to map schemas to POJO and some other external systems, and others (`.`, `[`, `]`) make field accesses ambiguous and seems to be effectively disallowed in Java today.

Perhaps we _should_ allow all of these special characters in field names, and make that clear here. We would need to make sure our implementation provides ways to work around the issues - For example, allow escaping in field access descriptors.

Maybe this is worth a ML discussion..

aromanenko-dev:
Is "dynamic schemas" already supported in some way? If not, what are the plans for this?

TheNeuralBit:
> Is "dynamic schemas" already supported in some way? If not, what are the plans for this?

I think the only plan that's been discussed is allowing schemas to change (with certain restrictions) during a pipeline update. I don't think there are any plans for supporting schemas that can only be determined at execution time.

website/src/documentation/programming-guide.md
TheNeuralBit: Looks like the consensus from [dev@](https://lists.apache.org/thread.html/re0451b85feb703d8305f7044d17da9cdc494efaa4b7c0fcf152514e4%40%3Cdev.beam.apache.org%3E) is field names should be allowed to contain any character. Can we document that somewhere?

aromanenko-dev:
> I don't think there are any plans for supporting schemas that can only be determined at execution time.

I believe it should be very useful feature for long-running streaming pipeline. As a good example could be Confluent Schema Registry with Avro schemas that supports schema evolution by versioning.

website/src/documentation/programming-guide.md
reuvenlax: Added

website/src/documentation/programming-guide.md
```diff
@@ -1970,7 +1976,1078 @@ records.apply("WriteToText",
 See the [Beam-provided I/O Transforms]({{site.baseurl }}/documentation/io/built-in/)
 page for a list of the currently available I/O transforms.
 
-## 6. Data encoding and type safety {#data-encoding-and-type-safety}
+## 6. Schemas {#schemas}
+Often, the type of records being processed have an obvious structure. Common Beam sources produce
```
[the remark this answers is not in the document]
reuvenlax: fixed

website/src/documentation/programming-guide.md
```diff
@@ -1970,7 +1976,1078 @@ records.apply("WriteToText",
...
+JSON, Avro, Protocol Buffer, or database row objects; all of these types have well defined structures, 
+structures that can often be determined by examining the type. Even within a SDK pipeline, Simple Java POJOs 
+(or  equivalent structures in other languages) are often used as intermediate types, and these also have a
+ clear structure that can be inferred by inspecting the class. By understanding the structure of a pipeline’s 
+ records, we can provide much more concise APIs for data processing.
+ 
+### 6.1. What is a schema {#what-is-a-schema}
+Most structured records share some common characteristics: 
+* They can be subdivided into separate named fields. Fields usually have string names, but sometimes - as in the case of indexed
```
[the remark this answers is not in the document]
reuvenlax: Left it as is because the numerical indices are currently called "ids" in the code.

website/src/documentation/programming-guide.md
```diff
@@ -1970,7 +1976,1078 @@ records.apply("WriteToText",
...
+    <tr>
+      <td>purchaseAmount</td>
+      <td>DOUBLE</td>      
+    </tr>                  
+  </tbody>
+</table>
+<br/>
+
+Purchase event records are represented by the above purchase schema. Each purchase event contains a shipping address, which
+is a nested row containing its own schema. Each purchase also contains a list of credit-card transactions
```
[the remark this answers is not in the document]
reuvenlax: done

website/src/documentation/programming-guide.md
```diff
@@ -1970,7 +1976,1078 @@ records.apply("WriteToText",
...
+
+This provides an abstract description of the types involved, one that is abstracted away from any specific programming 
+language.
+
+Schemas provide us a type-system for Beam records that is independent of any specific programming-language type. There
+might be multiple Java classes that all have the same schema (for example a Protocol-Buffer class or a POJO class),
+and Beam will allow us to seamlessly convert between these types. Schemas also provide a simple way to reason about 
+types across different programming-language APIs.
+
+A `PCollection` with a schema does not need to have a `Coder` specified, as Beam knows how to encode and decode
```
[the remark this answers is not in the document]
reuvenlax: clarified

website/src/documentation/programming-guide.md
```diff
@@ -1970,7 +1976,1078 @@ records.apply("WriteToText",
...
+
+So if we had two `PCollection`s as follows
+
+```java
+PCollection<Transaction> transactionBeans = readTransactionsAsJavaBean();
+PCollection<TransactionPojos> transactionPojos = readTransactionsAsPojo();
+```
+
+Then these two `PCollection`s would have the same schema, even though their Java types would be different. This means
+for example the the following two code snippets are valid:
```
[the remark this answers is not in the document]
reuvenlax: done

website/src/documentation/programming-guide.md
reuvenlax: Yes - because we don't guarantee deterministic encoding of maps.

website/src/documentation/programming-guide.md
reuvenlax: fixed

website/src/documentation/programming-guide.md
reuvenlax: not sure I understand?

website/src/documentation/programming-guide.md
reuvenlax: clarified

website/src/documentation/programming-guide.md
reuvenlax: fixed

website/src/documentation/programming-guide.md
reuvenlax: fixed

website/src/documentation/programming-guide.md
reuvenlax: yes - plan to add Avro and Proto in a followup PR

website/src/documentation/programming-guide.md
reuvenlax: Changed, though this section also talks about how Schema is built into ParDo, so not just about PTransforms.

reuvenlax:
@TheNeuralBit @iemejia thanks for the reviews. Comments addressed.

website/src/documentation/programming-guide.md
reuvenlax: I would prefer to wait until we actually can support that (i.e. until our selection language supports escaping characters) before documenting that.

reuvenlax:
@aromanenko-dev "dynamic" schemas can mean many things. Schemas support dynamic map fields, that can be used to dynamically extend the schema.

aromanenko-dev:
> "dynamic" schemas can mean many things. Schemas support dynamic map fields, that can be used to dynamically extend the schema.

@reuvenlax I meant schema evolution or schema versioning for long running pipelines. For example, full support of Confluent Schema Registry in KafkaIO. Well, I guess the better place to discuss it will be dev@. Sorry for messing things up here.

reuvenlax:
@TheNeuralBit @iemejia friendly ping

iemejia:
@reuvenlax sorry for the slow reaction, I only will be able to take a look a deep look after friday, if you can wait I will do a second round otherwise maybe is better than Brian does it. In the meantime it seems this needs a rebase now

reuvenlax:
@iemejia the conflict appears to be in a file that should not be in this PR - looks like accidental keystrokes.

@TheNeuralBit have you had time to look at this again?

reuvenlax:
friendly ping, as this PR is now quite old

TheNeuralBit:
Sorry, I'll take a look today

sdks/java/core/src/test/java/org/apache/beam/sdk/transforms/ParDoTest.java
```diff
@@ -3427,7 +3427,7 @@ public void processElement(
                 ProcessContext context, BoundedWindow window, @TimerId(timerId) Timer timer) {
               try {
                 timer.set(window.maxTimestamp().plus(1L));
-                fail("Should have failed due to processing time with absolute timer.");
+                fail("Should have failed due ting-guo processing time with absolute timer.");
```
TheNeuralBit, line 3430: This looks like it was brought in by mistake

website/src/documentation/programming-guide.md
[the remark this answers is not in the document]
TheNeuralBit: Ahh I see. Makes sense, that's a very good thing to clear up.

website/src/documentation/programming-guide.md
```diff
@@ -1970,7 +1976,1078 @@ records.apply("WriteToText",
...
+
+A `PCollection` with a schema does not need to have a `Coder` specified, as Beam knows how to encode and decode 
+Schema rows.
+
+### 6.2. Schemas for programming language types {#schemas-for-pl-types}
+While schemas themselves are language independent, they are designed to embed naturally into the programming languages
+of the Beam SDK being used. This allows Beam users to continue using native types while reaping the advantage of 
+having Beam understand their element schemas.
+ 
+ {:.language-java}
```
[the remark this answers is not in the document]
TheNeuralBit: Is this something we need to raise with docs folks?

website/src/documentation/programming-guide.md
```diff
@@ -1970,7 +1976,1108 @@ records.apply("WriteToText",
...
+ 
+  @SchemaCreate
+  public Transaction(String bank, double purchaseAmount) {
+     ...
+  }
+}
+```
+
+Using JavaBean classes as above is one way to map a schema to Java classes. However multiple Java classes might have
+the same schema, in which case the different Java types can often be used interchangeably. For example, the above
```
TheNeuralBit, line 2157: Maybe add a sentence here:

> Beam will add implicit conversions between types that have matching schemas. For example, ...

website/src/documentation/programming-guide.md
```diff
@@ -1970,7 +1976,1108 @@ records.apply("WriteToText",
...
+```java
+transactionPojos.apply(ParDo.of(new DoFn<...>() {
+   @ProcessElement public void process(@Element Transaction row) {
+    }
+}));
+```
+
+Even though the in both cases the `@Element` parameter differs from the the `PCollection`'s Java type, since the
+schemas are the same Beam will automatically make the conversion. The built-in `Convert` transform can also be used
+to translate between Java types of equivalent schemas, as detailed below.
```
TheNeuralBit, line 2196: What about contrasting this idea with logical types? Conceptually, inferred schemas for PL types and a logical type backed by a ROW look very similar, but the critical difference is we will never automatically convert the latter to its base type, or to a type with an equivalent schema.

Maybe just a note linking to logical types like:

> If this behavior is not desirable, then it's likely a logical type backed by a ROW is more appropriate for your programming language type.

website/src/documentation/programming-guide.md
TheNeuralBit: I think the ask is to avoid repeating EnumerationType in the description by referring to it as "This logical type"

iemejia:
This PR is closed at the moment was this intended or just a mistake?

reuvenlax:
@iemejia I'm still not sure what happened. I rebased my client and pushed, and that seemed to revert all my changes and automatically close the PR. This seems to be repeatable, and I'm not quite sure right now what's happening with git.

iemejia:
Mmmm it seems it became bonkers and included all the other commits. Maybe `git pull origin master --rebase` can fix it.

PR #10812, opened 2020-02-10
[BEAM-7092] Fix invalid import of Guava coming from transitive Spark dep

iemejia:
This import is not available in the transitive dependencies of Spark 3
anymore, and we should not be using non-vendored guava.

R: @je-ik


iemejia:
Thanks @je-ik

iemejia:
Run Java PreCommit

PR #10814, opened 2020-02-10
[BEAM-9278] Make HBase client a provided dependency in HBaseIO

iemejia:
R: @aromanenko-dev

### DIFF
--- a/sdks/java/io/hbase/build.gradle
+++ b/sdks/java/io/hbase/build.gradle
@@ -41,7 +41,7 @@ dependencies {
   compile project(path: ":sdks:java:core", configuration: "shadow")
   compile project(":sdks:java:io:hadoop-common")
   compile library.java.slf4j_api
-  compile "org.apache.hbase:hbase-shaded-client:$hbase_version"
+  provided "org.apache.hbase:hbase-shaded-client:$hbase_version"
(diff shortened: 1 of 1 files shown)


iemejia:
Sorry to bother you @suztomo but given your strong knowledge about dependencies I have a question. I cannot get to understand how this can be breaking. Does this make somehow some sense to you?

suztomo:
```
java.lang.NoSuchMethodError: org.apache.hadoop.security.authentication.util.KerberosUtil.hasKerberosTicket(Ljavax/security/auth/Subject;)Z
	at org.apache.hadoop.security.UserGroupInformation.<init>(UserGroupInformation.java:657)
```

'provided' dependency of beam-sdks-java-io-hbase will not appear in other modules that depends on beam-sdks-java-io-hbase. However, I thought the provided dependency should be there when running tests for beam-sdks-java-io-hbase.

`./gradlew -p sdks/java/io/hbase/ dependencies` would clarify what dependencies are included when running tests. There must be another artifact that contains `org.apache.hadoop.security.authentication.util.KerberosUtil` without the `void hasKerberosTicket(Subject arg)` method.

iemejia:
Oh good point, that's what I expected so then it is other class maybe coming from Beam's hadoop-common. I am going to close this one for the moment to try to figure it out. Thanks a lot @suztomo !

PR #10815, opened 2020-02-10
[BEAM-9279] Make HBase.ReadAll based on Reads instead of HBaseQuery

iemejia:
This PR adds an additional commit that is a prerrequisite to improve the serialization of the `HBase.Read` class, as a consequence the 'business' logic became simpler (because the IO related code does not rely on the Serialization wrappers).
Notice that I do not add extra Serialization tests because the pipeline Execution with direct runner already enforces Serialization and even the equals contract (that I had to make the tests happy).

R: @aromanenko-dev


sdks/java/io/hbase/src/main/java/org/apache/beam/sdk/io/hbase/HBaseIO.java
```diff
@@ -240,63 +245,109 @@ private Read(
...
+          && Objects.equals(tableId, read.tableId)
+          && scan.toString().equals(read.scan.toString());
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hash(configuration, tableId, scan);
+    }
+
+    private Object writeReplace() {
```
iemejia: This was required to keep the Read class immutable. This is some kind of magic I did not know before in case you did not too https://lingpipe-blog.com/2009/08/10/serializing-immutable-singletons-serialization-proxy/

sdks/java/io/hbase/src/main/java/org/apache/beam/sdk/io/hbase/HBaseReadSplittableDoFn.java
```diff
@@ -32,65 +31,50 @@
...
   @ProcessElement
-  public void processElement(ProcessContext c, RestrictionTracker<ByteKeyRange, ByteKey> tracker)
+  public void processElement(
+      @Element Read read,
+      OutputReceiver<Result> out,
+      RestrictionTracker<ByteKeyRange, ByteKey> tracker)
       throws Exception {
-    final HBaseQuery query = c.element();
-    TableName tableName = TableName.valueOf(query.getTableId());
+    Connection connection = ConnectionFactory.createConnection(read.getConfiguration());
```
iemejia: This creates a connection for each Read element, this is arguable costly because its nature is 1 to many, but it is an acceptable trade-off to support multiple Configurations and have the multi cluster flexibility.

Connection initialization could be improved in the future via some sort of class based Pool mechanism (as we do for JdbcIO).

iemejia:
Run Java PreCommit

aromanenko-dev:
Thank you for contribution! Sorry for delay with review - I'll be off next week and will get back to this PR after.

sdks/java/io/hbase/src/main/java/org/apache/beam/sdk/io/hbase/HBaseIO.java
```diff
@@ -240,63 +245,109 @@ private Read(
...
-    private ReadAll(SerializableConfiguration serializableConfiguration) {
-      this.serializableConfiguration = serializableConfiguration;
-    }
-
-    /** Reads from the HBase instance indicated by the* given configuration. */
-    public ReadAll withConfiguration(Configuration configuration) {
-      checkArgument(configuration != null, "configuration can not be null");
-      return new ReadAll(new SerializableConfiguration(configuration));
-    }
+  public static class ReadAll extends PTransform<PCollection<Read>, PCollection<Result>> {
```
aromanenko-dev: Is it a breaking change of users API?

iemejia: Yes and it is intended, notice however that the only breaking change is for the SDF based read transform `ReadAll`. Also let's not forget that HBaseIO is still Experimental and Reads based on SDF for HBaseIO are Experimental^2. However after this change I seriously doubt the API will change in the future.

sdks/java/io/hbase/src/main/java/org/apache/beam/sdk/io/hbase/HBaseIO.java
```diff
@@ -173,33 +182,33 @@ public static Read read() {
...
     /** Reads from the specified table. */
     public Read withTableId(String tableId) {
       checkArgument(tableId != null, "tableIdcan not be null");
-      return new Read(serializableConfiguration, tableId, serializableScan);
+      return new Read(configuration, tableId, scan);
     }
 
     /** Filters the rows read from HBase using the given* scan. */
     public Read withScan(Scan scan) {
       checkArgument(scan != null, "scancan not be null");
```
aromanenko-dev: type "scancan"

sdks/java/io/hbase/src/main/java/org/apache/beam/sdk/io/hbase/HBaseIO.java
```diff
@@ -173,33 +182,33 @@ public static Read read() {
     /** Reads from the HBase instance indicated by the* given configuration. */
     public Read withConfiguration(Configuration configuration) {
       checkArgument(configuration != null, "configuration can not be null");
-      return new Read(new SerializableConfiguration(configuration), tableId, serializableScan);
+      return new Read(new Configuration(configuration), tableId, scan);
     }
 
     /** Reads from the specified table. */
     public Read withTableId(String tableId) {
       checkArgument(tableId != null, "tableIdcan not be null");
```
aromanenko-dev: typo "tableIdcan"

sdks/java/io/hbase/src/main/java/org/apache/beam/sdk/io/hbase/HBaseIO.java
aromanenko-dev: Add it into method Javadoc please

sdks/java/io/hbase/src/main/java/org/apache/beam/sdk/io/hbase/HBaseIO.java
aromanenko-dev: Please, add this warning/note to top class Javadoc

sdks/java/io/hbase/src/main/java/org/apache/beam/sdk/io/hbase/HBaseReadSplittableDoFn.java
aromanenko-dev: Please, add TODO about that.  Did you try to run performance test to see how much it's slower?

sdks/java/io/hbase/src/main/java/org/apache/beam/sdk/io/hbase/HBaseIO.java
iemejia: good one, fixing it

sdks/java/io/hbase/src/main/java/org/apache/beam/sdk/io/hbase/HBaseIO.java
iemejia: fixing it, actually they were more ocurrencdes of this 'can' mistake I fixed them all now.

sdks/java/io/hbase/src/main/java/org/apache/beam/sdk/io/hbase/HBaseReadSplittableDoFn.java
iemejia: Better filled https://issues.apache.org/jira/browse/BEAM-9554 to track this. No I did not test performance because this is a quite particular case as I mention, for users doing 1 to n queries when n is big this time would not be considerable, the real issue can manifest mostly in pipelines with streaming where we would like to do reads per window with multiple windows kind of similar to what we found for JdbcIO writes (but this case is waaay more common).

sdks/java/io/hbase/src/main/java/org/apache/beam/sdk/io/hbase/HBaseIO.java
iemejia: good idea, doing it.

sdks/java/io/hbase/src/main/java/org/apache/beam/sdk/io/hbase/HBaseIO.java
iemejia: I better do this in the CHANGES.md release notes file, so this gets announced with the release notes. Java will cover making users aware at the code level :)

iemejia:
Changes done, I rebased since the PR was a bit old. It should be good to go now. PTAL @aromanenko-dev !

sdks/java/io/hbase/src/main/java/org/apache/beam/sdk/io/hbase/HBaseIO.java
aromanenko-dev: thanks

aromanenko-dev:
Thanks! It LGTM, just let's wait for all tests passed before merging.

iemejia:
Thanks Alexey!

PR #10816, opened 2020-02-10
[BEAM-9273] Explicitly disable @RequiresTimeSortedInput on unsupported runners

je-ik:
Fixes [BEAM-9273]
------------------------

--- | --- | --- | --- | ---


je-ik:
Run Flink ValidatesRunner

je-ik:
Run Spark ValidatesRunner

je-ik:

je-ik:
Run Apex ValidatesRunner

je-ik:
Run Dataflow ValidatesRunner

je-ik:
Run Samza ValidatesRunner

je-ik:
Run Flink ValidatesRunner

je-ik:
Run Dataflow ValidatesRunner

je-ik:

je-ik:
Run Apex ValidatesRunner

je-ik:
Run Spark ValidatesRunner

je-ik:
Run Samza ValidatesRunner

je-ik:
Run Gearpump ValidatesRunner

je-ik:
The fail seems not to be related to this PR. First failed post commit seems to be [this one](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Flink/6608/).

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/DoFnFeatures.java
```diff
@@ -0,0 +1,82 @@
...
+import org.apache.beam.sdk.state.SetState;
+import org.apache.beam.sdk.state.State;
+import org.apache.beam.sdk.state.ValueState;
+import org.apache.beam.sdk.state.WatermarkHoldState;
+import org.apache.beam.sdk.transforms.DoFn;
+import org.apache.beam.sdk.transforms.reflect.DoFnSignatures;
+import org.apache.beam.sdk.values.TypeDescriptor;
+
+/**
+ * Features a {@link DoFn} can posses. Each runner might implement a different (sub)set of this
```
kennknowles: The `DoFnSignature` is precisely the list of features already, no? It is OK to have helper methods anyhow.

Style point: for a given class, utility static methods usually go in a companion class like `DoFnSignature` (class) & `DoFnSignatures` (utility methods)

runners/core-construction-java/src/test/java/org/apache/beam/runners/core/construction/DoFnFeaturesTest.java
```diff
@@ -0,0 +1,276 @@
...
+import org.apache.beam.sdk.state.WatermarkHoldState;
+import org.apache.beam.sdk.transforms.DoFn;
+import org.apache.beam.sdk.transforms.splittabledofn.RestrictionTracker;
+import org.apache.beam.sdk.transforms.windowing.TimestampCombiner;
+import org.apache.beam.sdk.values.KV;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.Lists;
+import org.junit.Test;
+
+/** Test suite for {@link DoFnFeatures}. */
+public class DoFnFeaturesTest {
```
kennknowles: Ditto `DoFnSignaturesTest` (utility methods tests)

runners/google-cloud-dataflow-java/src/main/java/org/apache/beam/runners/dataflow/BatchStatefulParDoOverrides.java
```diff
@@ -176,7 +176,7 @@ private MultiOutputOverrideFactory(boolean isFnApi) {
     public PCollection<OutputT> expand(PCollection<KV<K, InputT>> input) {
       DoFn<KV<K, InputT>, OutputT> fn = originalParDo.getFn();
       verifyFnIsStateful(fn);
-      DataflowRunner.verifyStateSupported(fn);
+      DataflowRunner.verifyDoFnSupported(fn, false);
```
kennknowles: Passing a raw bool into a function call is not very readable. I suggest splitting into `verifyDoFnSupportedForStreaming` and `verifyDoFnSupportedForBatch`. These can each call the common code.

runners/google-cloud-dataflow-java/src/main/java/org/apache/beam/runners/dataflow/DataflowPipelineTranslator.java
```diff
@@ -1224,10 +1225,10 @@ private static void translateFn(
       Map<TupleTag<?>, Coder<?>> outputCoders,
       DoFnSchemaInformation doFnSchemaInformation,
       Map<String, PCollectionView<?>> sideInputMapping) {
-    DoFnSignature signature = DoFnSignatures.getSignature(fn.getClass());
 
-    if (signature.usesState() || signature.usesTimers()) {
-      DataflowRunner.verifyStateSupported(fn);
+    boolean isStateful = DoFnFeatures.isStateful(fn);
```
kennknowles: This is a good example of `DoFnFeatures` being useful: since "stateful ParDo" is really a totally separate primitive, it makes sense to just have a single boolean indicating it. Eventually moving so the different primitives are more clearly split with different URNs would be good.

(no action for you here - just a positive comment)

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/DoFnFeatures.java
je-ik: Agree. The reason I created this class is that I wanted to demonstrate the approach of declaring pipeline requirements and validating that runner supports all requirements. 

That would go as follows:
```java
 Set<Features> required = PipelineFeaturs.extract(pipeline);
 Runner.validateAllFeaturesSupported(required);
```
That way, adding a new feature to pipeline would not require any change to runner core, but the runner would reject the pipeline.

I then realized that this change would be too big and really not related to the annotation, so I returned back to the original approach, where runners explicitly reject features they do not support (and adding unsupported feature needs modification in runners code). I will rename the class as you suggest.

runners/core-construction-java/src/test/java/org/apache/beam/runners/core/construction/DoFnFeaturesTest.java
je-ik: :+1:

runners/google-cloud-dataflow-java/src/main/java/org/apache/beam/runners/dataflow/BatchStatefulParDoOverrides.java
je-ik: :+1:

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/DoFnFeatures.java
je-ik: Hmmm ... `DoFnSignatures` already exists. It does something different (creates `DoFnSignature` from `DoFn`). Feels weird to mix these two helper functionalities.

je-ik: @kennknowles would you have any suggestions about naming the class? I think this code really should not go to `DoFnSignatures`, can we agree on some alternative name?

kennknowles: I quite strongly believe it belongs in `DoFnSignatures` and that what you describe in the rest belongs in `Pipelines`. Static-method-only utility classes tend to be disorganized and undiscoverable. It is better to attach them to the thing that they are most related to.

runners/google-cloud-dataflow-java/src/main/java/org/apache/beam/runners/dataflow/BatchStatefulParDoOverrides.java
je-ik: I added both versions, although the second one `verifyDoFnSupportedStreaming` is not used. I added that so that the methods are not imbalanced. The streaming case is called from `DataflowRunner.verifyDoFnSupported(fn, context.getPipelineOptions().isStreaming())`, where it would be weird to do `if (context..isStream()) verifyStreaming() else ...`

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/DoFnFeatures.java
je-ik: Moved the code there. The reason I didn't like it is that `DoFnSignatures` probably when created served the purpose to create signatures from DoFn, while the code I added makes no use of `DoFnSignature` itself. But there seems to be more code like that already, so I'm fine with that.

sdks/java/core/src/test/java/org/apache/beam/sdk/transforms/reflect/DoFnSignaturesTest.java
```diff
@@ -1192,6 +1199,239 @@ public void bar(
...
+          new StatefulWithTimersAndValueState(),
+          new StatefulWithSetState(),
+          new StatefulWithMapState(),
+          new StatefulWithWatermarkHoldState(),
+          new RequiresTimeSortedInput(),
+          new Splittable());
+
+  @Test
+  public void testAllDoFnFeatures() {
+    tests.forEach(FeatureTest::test);
```
kennknowles: Nit: there's a lot of logic in this test class. See https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html

It would be better to simply write out the boilerplate for each test, so each test can be read in a straight line with no context needed.

je-ik: I wanted to keep the declaration of a class and its features encapsulated in single class. Agree that this doesn't play well with the rest of the `DoFnSignaturesTest`. But because there is effort in the direction of superseding all this with "pipeline features", I think it is fine to keep is as is for now and drop it as part of the effort later.

je-ik:
Thanks @kennknowles

PR #10824, opened 2020-02-10
Enable probot autolabeler

alexvanboxel:
Waiting for INFRA enablement

R: iemejia

### DIFF
--- a/.github/autolabeler.yml
+++ b/.github/autolabeler.yml
@@ -1,0 +1,84 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+build: ["buildSrc/*", "release/*", "ownership/*", ".github/*", "gradle/*"]
+docker: ["sdks/python/container/*", "sdks/go/container/*", "sdks/java/container/*", "runners/flink/job-server-container/*", "runners/spark/job-server/container/*"]
+examples: ["examples/*"]
+go: ["sdks/go/*"]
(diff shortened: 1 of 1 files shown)


.github/autolabeler.yml
```diff
@@ -0,0 +1,26 @@
...
+
+infra: [".test-infra/*"]
+build: ["buildSrc/*", "release/*", "ownership/*", ".github/*", "gradle/*"]
+website: ["website/*"]
+go: ["sdks/go/*"]
+java: ["sdks/java/*"]
+python: ["sdks/python/*"]
+runner: ["runners/*"]
+examples: ["examples/*", "learning/*"]
+model: ["model/*"]
```
sunjincheng121: Add newline at the end of file.

alexvanboxel: Added newline

.github/autolabeler.yml
```diff
@@ -0,0 +1,26 @@
...
+# limitations under the License.
+#
+
+infra: [".test-infra/*"]
+build: ["buildSrc/*", "release/*", "ownership/*", ".github/*", "gradle/*"]
+website: ["website/*"]
+go: ["sdks/go/*"]
+java: ["sdks/java/*"]
+python: ["sdks/python/*"]
+runner: ["runners/*"]
```
iemejia: Can we expand this also into the individual runners, e.g.
```
flink: ["runners/flink/*"]
spark: ["runners/spark/*"]
...
```

.github/autolabeler.yml
```diff
@@ -0,0 +1,26 @@
...
+examples: ["examples/*", "learning/*"]
```
iemejia: I would keep learning separate, these are different things and commonly mainted by different people who may be interesting in filtering one or the other.

iemejia:
I had almost the same suggestion than @sunjincheng121 however there is a small refinement to do over that one, do not put prefixes e.g. `extensions-java-sorter` use better `sorter` the other paths are matched by the other tag rules the specific rule for java and for extensions.

iemejia:
Forgot to mention, Can we sort the elements of the file so it can be easier to maintain too.

alexvanboxel:
I've added @iemejia and @sunjincheng121 suggestions. I did use a prefix though to make it usable in maintaining labels. I hope the auto-labeler can handle them. Prefix used are:

- extensions: ext/ (example: ext/jackson)
- io: io/ (example: io/mongodb)
- runners: run/ (example: run/flink)

I also sorted the labels. Squashed commit.

alexvanboxel:
Added the main labels already to the repo (not the sub labels)

iemejia:
Merged manually to do the extra adjustments. I did not want to bother anymore with those.
Since INFRA apparently enabled the action now is time to see those auto labels appearing :)

PR #10845, opened 2020-02-12
Refine labels/categories for PR autolabeling

iemejia:
Added missing spaces after commas too for consistency.
R: @alexvanboxel


.github/autolabeler.yml
```diff
@@ -73,7 +73,7 @@ io:  ["sdks/java/io/*"]
 # Runners
 "runners": ["runners/*"]
 "apex": ["runners/apex/*"]
-"core": ["runners/core-construction-java/*","runners/core-java/*"]
+"core": ["runners/core-construction-java/*", "runners/core-java/*"]
```
alexvanboxel: Because you removed the prefix, I'm not so happy with the resulting core label. You also have an SDK core, it seems confusing. This has nothing todo with this PR though, so LGTM

iemejia: Again it is a question of tradeoffs, we don't have an 'sdks' category so the argument of the second core does not hold (even if we have sdks/core).
Alex I did not do this to annoy you, but with valid arguments on readability and usability, prove me that the prefix version is simpler and better and I don't hesitate to change to it, but I am still unconvinced of what we win with the extra prefix vs the extra noise and reduced readability it brings specially when the prefixes are already covered by the general categories.

iemejia:
Please don't merge this yet, I plan to add other refinements, but that will be tomorrow.

iemejia:
Finally i did it before going to sleep. New refinements added. Can you PTAL again @alexvanboxel

alexvanboxel:
I've add the Kotlinn label to the list

iemejia:
Thanks !

PR #10851, opened 2020-02-13
Pin Avro dependency in Python SDK to be consistent with Avro versioning

iemejia:
Avro does not follow semver so we should not be automatically upgrade on minor versions.
R: @tvalentyn 
CC: @aaltay

### DIFF
--- a/sdks/python/setup.py
+++ b/sdks/python/setup.py
@@ -135,8 +135,12 @@ else:
     cythonize = lambda *args, **kwargs: []
 
 REQUIRED_PACKAGES = [
-    'avro>=1.8.1,<2.0.0; python_version < "3.0"',
-    'avro-python3>=1.8.1,!=1.9.2,<2.0.0; python_version >= "3.0"',
+    # Apache Avro does not follow semantic versioning, so we should not auto
+    # upgrade on minor versions. However, due to an issue on Dataflow we need
+    # to still include the previous version (1.8.x).
+    'avro>=1.8.1,<1.10.0; python_version < "3.0"',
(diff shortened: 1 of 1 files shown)


sdks/python/setup.py
```diff
@@ -135,8 +135,8 @@ def get_version():
     cythonize = lambda *args, **kwargs: []
 
 REQUIRED_PACKAGES = [
-    'avro>=1.8.1,<2.0.0; python_version < "3.0"',
-    'avro-python3>=1.8.1,!=1.9.2,<2.0.0; python_version >= "3.0"',
+    'avro>=1.9.0,<1.10.0; python_version < "3.0"',
```
tvalentyn: Can you please add a comment that Avro does not follow semantic versioning, so second number is major version, and changes in major versions are not backwards-compatible?
Also I would suggest to target 1.8 range in this change, and make 1.8 -> 1.9 as a separate change.
I know that we attempted 1.9 in the past and it did not work well:  https://issues.apache.org/jira/browse/AVRO-2429 (cc: @chamikaramj).

iemejia: done

iemejia:
Comment added + moved lower bound back to 1.8.1. PTAL @tvalentyn

aaltay:
Run Python PreCommit

aaltay:
Run Portable_Python PreCommit

tvalentyn:
Run Python PreCommit

iemejia:
Thanks for the review @aaltay and @tvalentyn

PR #10879, opened 2020-02-17
[BEAM-9326] Make JsonToRow transform input <String> instead of <? extends String>

iemejia:
R: @kennknowles

### DIFF
--- a/sdks/java/core/src/test/java/org/apache/beam/sdk/PipelineTest.java
+++ b/sdks/java/core/src/test/java/org/apache/beam/sdk/PipelineTest.java
@@ -225,7 +225,7 @@ public class PipelineTest {
   @Test
   @Category(ValidatesRunner.class)
   public void testMultipleApply() {
-    PTransform<PCollection<? extends String>, PCollection<String>> myTransform = addSuffix("+");
+    MapElements<String, String> myTransform = addSuffix("+");
 
     PCollection<String> input = pipeline.apply(Create.of(ImmutableList.of("a", "b")));
 
@@ -239,8 +239,7 @@ public class PipelineTest {
     pipeline.run();
   }
 
-  private static PTransform<PCollection<? extends String>, PCollection<String>> addSuffix(
-      final String suffix) {
+  private static MapElements<String, String> addSuffix(final String suffix) {
     return MapElements.via(
         new SimpleFunction<String, String>() {
           @Override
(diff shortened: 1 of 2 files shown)


sdks/java/core/src/test/java/org/apache/beam/sdk/PipelineTest.java
```diff
@@ -225,7 +225,7 @@ public void testPipelineSDKExceptionHandling() {
   @Test
   @Category(ValidatesRunner.class)
   public void testMultipleApply() {
-    PTransform<PCollection<? extends String>, PCollection<String>> myTransform = addSuffix("+");
+    MapElements<String, String> myTransform = addSuffix("+");
```
kennknowles: Suggest changing `? extends String` to just `String` but otherwise leaving the type as `PTransform`. This is a good way to indicate that you don't intend to use any other facts about the transform.

sdks/java/core/src/test/java/org/apache/beam/sdk/PipelineTest.java
```diff
@@ -239,8 +239,7 @@ public void testMultipleApply() {
     pipeline.run();
   }
 
-  private static PTransform<PCollection<? extends String>, PCollection<String>> addSuffix(
-      final String suffix) {
+  private static MapElements<String, String> addSuffix(final String suffix) {
```
kennknowles: Here, too.

sdks/java/core/src/test/java/org/apache/beam/sdk/PipelineTest.java
iemejia: Yes I agree that the PTransform form is more abstract so better, but because of the capricious Java bounds checking this is not possible without doing the `<? extends String>` so I prefer to revert this change and let only  the JsonToRow one, which is the one I care about (and the real goal of this PR).
Thanks for the review Kenn.

iemejia:
Merged manually with the fixes

kennknowles:
Yikes! If removing `<? extends String>` makes it impossible to treat this class as a `PTransform` then I think this change should be reverted. Indeed I wondered if there was something funky going to happy in Java's type checking.

iemejia:
Yes it is already reverted, for the `PipelineTest` class, for `JsonToRow` we are good to go.

PR #10883, opened 2020-02-17
[BEAM-9331] Add better Row builders

reuvenlax:
This PR adds two builders to the Row object. The first allows building a Row by specifying fields by name:

 Row row = Row.withSchema(schema)
               .withFieldValue("userId", "user1)
              .withFieldValue("location.city", "seattle")
              .withFieldValue("location.state", "wa")
               .build();

The second allows building. a Row based on a previous row by specifying only the fields to change:

 Row modifiedRow =
    Row.fromRow(row)
        .withFieldValue("location.city", "tacoma")
        .build();

R: @rezarokni


sdks/java/core/src/main/java/org/apache/beam/sdk/values/Row.java
```diff
@@ -55,8 +59,31 @@
...
+ *
+ * <p>The {@link Row#fromRow} builder can be used to base a row off of another row. The builder can
+ * be used to specify values for specific fields, and all the remaining values will be taken from
+ * the original row. For example, the following produces a row identical to the above row except for
+ * the location.city field.
+ *
+ * <pre>{@code
+ * Row modifiedRow =
+ *     Row.fromRow(row)
+ *        .withFieldValue("location.city", "tacoma")
```
rezarokni: Would this be more readable if it was withAddedFieldValue?

sdks/java/core/src/test/java/org/apache/beam/sdk/values/RowTest.java
```diff
@@ -477,6 +477,172 @@ public void testCreateMapWithRowValue() {
     assertEquals(data, row.getMap("map"));
   }
 
+  @Test
```
rezarokni: Is it worth having a logical type to test as well?

sdks/java/core/src/main/java/org/apache/beam/sdk/values/Row.java
reuvenlax: I don't think so. We're not really adding anything, we're setting something.

sdks/java/core/src/test/java/org/apache/beam/sdk/values/RowTest.java
reuvenlax: yes will add one

sdks/java/core/src/main/java/org/apache/beam/sdk/values/Row.java
```diff
@@ -539,37 +568,118 @@ public String toString() {
...
+
+    private ModifyingBuilder(Row sourceRow) {
+      this.sourceRow = sourceRow;
+    }
+
+    public Schema getSchema() {
+      return sourceRow.getSchema();
+    }
+
+    public ModifyingBuilder withFieldValue(String fieldName, Object value) {
```
alexvanboxel: Wouldn't it be useful to have a `withFieldValue` with an index?

sdks/java/core/src/main/java/org/apache/beam/sdk/values/Row.java
alexvanboxel: Nope, I like the proposed `withFieldValue`.

sdks/java/core/src/test/java/org/apache/beam/sdk/values/RowTest.java
alexvanboxel: What get's people confused on the current API is that you can't loop over the values of a Row with `getValue()` and can't supply them to the Builder as the `addValue()` expects the input type, while the getValue returns a base type.

reuvenlax: added test

reuvenlax: still struggling to figure out a good answer to the API. getValues is supposed to return the base value so that transforms that don't know about the specific logical value can process it.

sdks/java/core/src/main/java/org/apache/beam/sdk/values/Row.java
```diff
@@ -599,272 +709,555 @@ public Builder addArray(Object... values) {
...
-    // values must be specified as the base type.
+    // Values are attached. No verification is done, and no conversions are done. LogicalType values
+    // must be specified
+    // as the base type. This method should be used with great care, as no validation is done. If
+    // incorrect values are
+    // passed in, it could result in strange errors later in the pipeline. This method is largely
+    // used internal
+    // to Beam.
+    @Internal
     public Builder attachValues(List<Object> values) {
```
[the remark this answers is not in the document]
reuvenlax: I like making attachValues return a Row. I thin the same for withFieldValueBuilders. this simplifies the builder code

sdks/java/core/src/main/java/org/apache/beam/sdk/values/Row.java
```diff
@@ -539,37 +568,118 @@ public String toString() {
...
+   * Creates a row builder based on the specified row. Field values in the new row can be explicitly
+   * set using {@link ModifyingBuilder#withFieldValue}. Any values not so overridden will be the
+   * same as the values in the original row.
+   */
+  public static ModifyingBuilder fromRow(Row row) {
+    return new ModifyingBuilder(row);
+  }
+
+  /** Builder for {@link Row} that bases a row on another row. */
+  public static class ModifyingBuilder {
```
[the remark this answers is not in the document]
reuvenlax: Yes, good suggestion. This should simplify the code

sdks/java/core/src/main/java/org/apache/beam/sdk/values/Row.java
```diff
@@ -539,37 +568,118 @@ public String toString() {
...
+    /** Return the schema for the row being built. */
     public Schema getSchema() {
       return schema;
     }
 
+    /**
+     * Set a field value using the field name. Nested values can be set using the field selection
+     * syntax.
+     */
+    public Builder withFieldValue(String fieldName, Object value) {
```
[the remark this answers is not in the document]
reuvenlax: done

sdks/java/core/src/main/java/org/apache/beam/sdk/values/Row.java
reuvenlax: added

sdks/java/core/src/test/java/org/apache/beam/sdk/values/RowTest.java
alexvanboxel: What about this,
- `addValue`, `getValue` and `getValues` return the input type(s) (so the logical type)
- `add`, `get`, `detach` returns the base types (is also inline with the `attach`)
That means a behaviour change only for `getValue` and `getValues`

alexvanboxel: Elaborating on the above comment:

`getValues` is currently not safe (it makes the Row mutable), if the meaning changes that it returns the logical type the implementation will need to make a copy of the content as it needs to translate the basetype to the logicaltype. As it's a copy the Row is immutable again. See this for test:

```  
  @Test
  public void testImmutability() {
    Schema schema = Schema.builder().addInt32Field("a").addInt32Field("b").build();
    Row row1 = Row.withSchema(schema).addValue(1).addValue(2).build();
    List<Object> values = row1.getValues();
    values.set(1, 3);
    assertEquals(2,(int)row1.getValue(1));
  }
```
Then the question is left for `attach` or `detach`: if you make values an immutable array it can be returned with `detach`, as it's immutable that's safe. The `attach` needs then make a copy of the array to an immutable array (or check if it's immutable) and `build` needs to make the immutable array as well.

-

reuvenlax: The fact that getValues returns the underlying objects I think is in line with many other things. e.g. ImmutableList in Java still returns the underlying objects, so you can mutate them. The onus is on the user not to do that.

In particular, the elements of a PCollection are not allowed to be mutated in a ParDo (and the main point of the Row object is to be used in ParDos). The DirectRunner has a mutability checker that will fail your test if you mutate the input elements.

alexvanboxel: Aside from the mutability/. WDYT of the proposed `addValue`, `getValue` and `getValues` vs `add`, `get`, `detach`

alexvanboxel: I propose to take this discussion out of this PR, as I really like this already. For me it's already a LGTM. (Best to wait till the 2.20 is cut though).

alexvanboxel:
Maybe rebase, then we could get this in. This is a useful addition.

reuvenlax:
@alexvanboxel there are still some bugs in this PR related to logical types, which is why I haven't pushed it in yet.

reuvenlax:
Run Java PreCommit

reuvenlax:
@alexvanboxel rebased and fixed bugs. Previously I was blocked on getting logical types to work, but now that we natively store logical types in Row, it's become much easier.

reuvenlax:
run sql postcommit

reuvenlax:
@alexvanboxel I would rather not in this PR, because the RowUtils API wasn't designed for public usage. If we were to make it public, I would prefer to spend a lot more time on the API design, and I would also want to understand the use cases a bit better.

alexvanboxel:
> @alexvanboxel I would rather not in this PR, because the RowUtils API wasn't designed for public usage. If we were to make it public, I would prefer to spend a lot more time on the API design, and I would also want to understand the use cases a bit better.

Understand LGTM

reuvenlax:
run sql postcommit

reuvenlax:
run sql postcommit

PR #10885, opened 2020-02-18
[BEAM-9085] Fix performance regression in SyntheticSource

kamilwu:
Beam was using `numpy.random.RandomState` during generation of Synthetic input, which, starting from NumPy 1.17, is considered as a legacy generator. This caused the slowdown on Python 3. 

This PR fixes the regression.

## Details
The following snippet was used for measuring performance:
```
from timeit import Timer

t = Timer("pardo_test.ParDoTest().run()",
'''
from apache_beam.testing.load_tests import pardo_test
import sys
sys.argv.append('--test-pipeline-options=--input_options=\\'{"num_records": 200000, "value_size": 9, "key_size": 1}\\'')
''')
print(min(t.repeat(number=1)))
```
### Results
Before:
**Python 2.7** 
4.52470111847
**Python 3.7** 
46.89912317299999

After:
**Python 2.7** 
6.37386107445
**Python 3.7** 
5.6981477510000005

### Why not numpy?

It turns out that `numpy.random.bytes()` is a bit slower on Python 3:

**Python 2.7**
`python -m timeit -s 'import numpy as np' 'for i in xrange(20000): np.random.seed(i); np.random.bytes(90)'`
10 loops, best of 3: 57.5 msec per loop

**Python 3.7**
`python -m timeit -s 'import numpy as np' 'for i in range(20000): np.random.seed(i); np.random.bytes(90)'`
1 loop, best of 5: 414 msec per loop

I've achieved better results after switching to `random` standard module. Unfortunately, the performance on Python 2 got a bit worse.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/testing/synthetic_pipeline.py
+++ b/sdks/python/apache_beam/testing/synthetic_pipeline.py
@@ -44,6 +44,7 @@ import math
 import random
 import struct
 import time
+from collections import defaultdict
 
 import apache_beam as beam
 from apache_beam.io import WriteToText
@@ -315,6 +316,7 @@ class SyntheticSource(iobase.BoundedSource):
     def maybe_parse_byte_size(s):
       return parse_byte_size(s) if isinstance(s, str) else int(s)
 
+    self._generators = defaultdict(random.Random)
     self._num_records = input_spec['numRecords']
     self._key_size = maybe_parse_byte_size(input_spec.get('keySizeBytes', 1))
     self._hot_key_fraction = input_spec.get('hotKeyFraction', 0)
@@ -418,29 +420,32 @@ class SyntheticSource(iobase.BoundedSource):
     return tracker
 
   @staticmethod
-  def random_bytes(length):
+  def random_bytes(length, generator):
     """Return random bytes."""
     return b''.join(
-        (struct.pack('B', random.getrandbits(8)) for _ in range(length)))
+        (struct.pack('B', generator.getrandbits(8)) for _ in range(length)))
 
-  def _gen_kv_pair(self, index):
-    random.seed(index)
-    rand = random.random()
+  def _gen_kv_pair(self, generator, index):
+    generator.seed(index)
+    rand = generator.random()
 
     # Determines whether to generate hot key or not.
     if rand < self._hot_key_fraction:
       # Generate hot key.
       # An integer is randomly selected from the range [0, numHotKeys-1]
       # with equal probability.
-      random.seed(index % self._num_hot_keys)
-    return self.random_bytes(self._key_size), self.random_bytes(
-      self._value_size)
+      generator.seed(index % self._num_hot_keys)
+    return self.random_bytes(self._key_size, generator), self.random_bytes(
+      self._value_size, generator)
 
   def read(self, range_tracker):
     index = range_tracker.start_position()
+    # Get an instance of pseudo-random number generator
+    generator = self._generators[(
+        range_tracker.start_position(), range_tracker.stop_position())]
     while range_tracker.try_claim(index):
       time.sleep(self._sleep_per_input_record_sec)
-      yield self._gen_kv_pair(index)
+      yield self._gen_kv_pair(generator, index)
       index += 1
 
   def default_output_coder(self):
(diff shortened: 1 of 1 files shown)


kamilwu:
R: @tvalentyn Please take a look. I've put some test results in the description, they might be interesting.

sdks/python/apache_beam/testing/synthetic_pipeline.py
```diff
@@ -415,19 +418,24 @@ def get_range_tracker(self, start_position, stop_position):
       tracker = range_trackers.UnsplittableRangeTracker(tracker)
     return tracker
 
+  @staticmethod
+  def random_bytes(length):
+    """Return random bytes."""
+    return b''.join(
+        (struct.pack('B', random.getrandbits(8)) for _ in xrange(length)))
```
tvalentyn: Let's use range instead of xrange, and adjust the import accordingly. We try to use idiomatic py3 whenever possible in Beam.

sdks/python/apache_beam/testing/synthetic_pipeline.py
```diff
@@ -415,19 +418,24 @@ def get_range_tracker(self, start_position, stop_position):
...
+
   def _gen_kv_pair(self, index):
-    r = np.random.RandomState(index)
-    rand = r.random_sample()
+    random.seed(index)
```
tvalentyn: How about we instantiate random.Random() instead of modifying the global state? This is important if we execute this generator concurrently. Not sure if we do.

sdks/python/apache_beam/testing/synthetic_pipeline.py
kamilwu: Alright. I was a bit concerned about performance of `range` on Python 2, but the parameter is almost always small so the impact is negligible.

sdks/python/apache_beam/testing/synthetic_pipeline.py
kamilwu: I'm pretty sure it's possible while having multiple range trackers and multiple workers.
I wrote a fix and ran the benchmark. The performance is a bit worse:
```
Python 2: 7.77728295326 (was 6.37386107445)
Python 3: 6.888783617999998 (was 5.6981477510000005)
```
However, we could create a cache for generators (instances of `random.Random`). Then we could create a generator for each range tracker, not for each index. This reduces the offset of instantiating multiple generators:
```
Python 2: 6.42752218246
Python 3: 5.789953311999998
```
WDYT?

kkucharc:
retest this please

kkucharc:
retest this please

sdks/python/apache_beam/testing/synthetic_pipeline.py
tvalentyn: We can add `from builtins import range` to use py3 implementation of range on py2.

sdks/python/apache_beam/testing/synthetic_pipeline.py
```diff
@@ -418,29 +420,32 @@ def get_range_tracker(self, start_position, stop_position):
...
-    return self.random_bytes(self._key_size), self.random_bytes(
-      self._value_size)
+      generator.seed(index % self._num_hot_keys)
+    return self.random_bytes(self._key_size, generator), self.random_bytes(
+      self._value_size, generator)
 
   def read(self, range_tracker):
     index = range_tracker.start_position()
+    # Get an instance of pseudo-random number generator
+    generator = self._generators[(
```
tvalentyn: I don't see a difference in performance if we instantiate the generator in `read`, but don't use the cache. Do we need the cache?

kamilwu: My bad, it seems the cache isn't really needed. It could be helpful if `read` was called many, many times with the same range tracker, but I don't think this is true. Thanks.

mwalenia:
retest this please

mwalenia:
retest this please

mwalenia:
retest this please

mwalenia:
retest this please

tvalentyn:
Thanks, @kamilwu !

kamilwu:
Thanks!

PR #10892, opened 2020-02-18
[BEAM-8335] Make TestStream to/from runner_api include the output_tags property.

rohdesamuel:
The TestStream has the "output_tags" property which keeps track of which events go to which PCollection. A TestStream going through a round-trip to/from proto won't have these fields set. To do this, a modification to the from_runner_api_parameter is needed to include the parent PTransform proto to retrieve the information from the outputs.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/runners/direct/direct_runner.py
+++ b/sdks/python/apache_beam/runners/direct/direct_runner.py
@@ -73,60 +73,12 @@ class SwitchingDirectRunner(PipelineRunner):
   def is_fnapi_compatible(self):
     return BundleBasedDirectRunner.is_fnapi_compatible()
 
-  def apply_TestStream(self, transform, pbegin, options):
-    """Expands the TestStream into the DirectRunner implementation.
-
-    Takes the TestStream transform and creates a _TestStream -> multiplexer ->
-    _WatermarkController.
-    """
-
-    from apache_beam.runners.direct.test_stream_impl import _TestStream
-    from apache_beam.runners.direct.test_stream_impl import _WatermarkController
-    from apache_beam import pvalue
-    assert isinstance(pbegin, pvalue.PBegin)
-
-    # If there is only one tag there is no need to add the multiplexer.
-    if len(transform.output_tags) == 1:
-      return (
-          pbegin
-          | _TestStream(transform.output_tags, events=transform._events)
-          | _WatermarkController())
-
-    # This multiplexing the  multiple output PCollections.
-    def mux(event):
-      if event.tag:
-        yield pvalue.TaggedOutput(event.tag, event)
-      else:
-        yield event
-
-    mux_output = (
-        pbegin
-        | _TestStream(transform.output_tags, events=transform._events)
-        | 'TestStream Multiplexer' >> beam.ParDo(mux).with_outputs())
-
-    # Apply a way to control the watermark per output. It is necessary to
-    # have an individual _WatermarkController per PCollection because the
-    # calculation of the input watermark of a transform is based on the event
-    # timestamp of the elements flowing through it. Meaning, it is impossible
-    # to control the output watermarks of the individual PCollections solely
-    # on the event timestamps.
-    outputs = {}
-    for tag in transform.output_tags:
-      label = '_WatermarkController[{}]'.format(tag)
-      outputs[tag] = (mux_output[tag] | label >> _WatermarkController())
-
-    return outputs
-
-  # We must mark this method as not a test or else its name is a matcher for
-  # nosetest tests.
-  apply_TestStream.__test__ = False
-
   def run_pipeline(self, pipeline, options):
 
     from apache_beam.pipeline import PipelineVisitor
     from apache_beam.runners.dataflow.native_io.iobase import NativeSource
     from apache_beam.runners.dataflow.native_io.iobase import _NativeWrite
-    from apache_beam.runners.direct.test_stream_impl import _TestStream
+    from apache_beam.testing.test_stream import TestStream
 
     class _FnApiRunnerSupportVisitor(PipelineVisitor):
       """Visitor determining if a Pipeline can be run on the FnApiRunner."""
@@ -138,7 +90,7 @@ class SwitchingDirectRunner(PipelineRunner):
       def visit_transform(self, applied_ptransform):
         transform = applied_ptransform.transform
         # The FnApiRunner does not support streaming execution.
-        if isinstance(transform, _TestStream):
+        if isinstance(transform, TestStream):
           self.supported_by_fnapi_runner = False
         # The FnApiRunner does not support reads from NativeSources.
         if (isinstance(transform, beam.io.Read) and
@@ -195,7 +147,8 @@ class _StreamingGroupByKeyOnly(_GroupByKeyOnly):
 
   @staticmethod
   @PTransform.register_urn(urn, None)
-  def from_runner_api_parameter(unused_payload, unused_context):
+  def from_runner_api_parameter(
+      unused_ptransform, unused_payload, unused_context):
     return _StreamingGroupByKeyOnly()
 
 
@@ -214,7 +167,7 @@ class _StreamingGroupAlsoByWindow(_GroupAlsoByWindow):
 
   @staticmethod
   @PTransform.register_urn(urn, wrappers_pb2.BytesValue)
-  def from_runner_api_parameter(payload, context):
+  def from_runner_api_parameter(unused_ptransform, payload, context):
     return _StreamingGroupAlsoByWindow(
         context.windowing_strategies.get_by_id(payload.value))
 
@@ -271,10 +224,21 @@ def _get_transform_overrides(pipeline_options):
       transform = _StreamingGroupAlsoByWindow(transform.dofn.windowing)
       return transform
 
+  class TestStreamOverride(PTransformOverride):
+    def matches(self, applied_ptransform):
+      from apache_beam.testing.test_stream import TestStream
+      self.applied_ptransform = applied_ptransform
+      return isinstance(applied_ptransform.transform, TestStream)
+
+    def get_replacement_transform(self, transform):
+      from apache_beam.runners.direct.test_stream_impl import _ExpandableTestStream
+      return _ExpandableTestStream(transform)
+
   overrides = [
       SplittableParDoOverride(),
       ProcessKeyedElementsViaKeyedWorkItemsOverride(),
-      CombinePerKeyOverride()
+      CombinePerKeyOverride(),
+      TestStreamOverride(),
   ]
 
   # Add streaming overrides, if necessary.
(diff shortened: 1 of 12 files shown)


rohdesamuel:
R: @davidyan74

sdks/python/apache_beam/testing/test_stream.py
```diff
@@ -133,15 +140,17 @@ def __eq__(self, other):
...
   def __lt__(self, other):
     return self.new_watermark < other.new_watermark
 
   def to_runner_api(self, unused_element_coder):
+    tag = 'None' if self.tag is None else self.tag
     return beam_runner_api_pb2.TestStreamPayload.Event(
         watermark_event=beam_runner_api_pb2.TestStreamPayload.Event.
-        AdvanceWatermark(new_watermark=self.new_watermark.micros // 1000))
+        AdvanceWatermark(
+            new_watermark=self.new_watermark.micros // 1000, tag=tag))
```
lukecwik: assert that we aren't losing precision here.

sdks/python/apache_beam/testing/test_stream_test.py
```diff
@@ -528,6 +529,56 @@ def process(
...
+    roundtrip_test_stream = TestStream().from_runner_api(
+        test_stream_proto, context)
+
+    self.assertListEqual(test_stream._events, roundtrip_test_stream._events)
+    self.assertSetEqual(
+        test_stream.output_tags, roundtrip_test_stream.output_tags)
+    self.assertEqual(test_stream.coder, roundtrip_test_stream.coder)
+
+  def test_roundtrip_proto_multi(self):
+    test_stream = (TestStream(output_tags=['a', 'b'])
```
lukecwik: I don't think you should need to specify output_tags here since they will get added via the add/advance calls below.

sdks/python/apache_beam/testing/test_stream.py
```diff
@@ -188,7 +198,16 @@ def _infer_output_coder(self, input_type=None, input_coder=None):
   def expand(self, pbegin):
     assert isinstance(pbegin, pvalue.PBegin)
     self.pipeline = pbegin.pipeline
-    return pvalue.PCollection(self.pipeline, is_bounded=False)
+    if len(self.output_tags) == 0:
```
davidyan74: nit: this can be simplified to just
```
if not self.output_tags:
```
Google's style guide recommends using boolean context whenever possible. Not sure whether there is such a rule in Beam. But since you're already using boolean context elsewhere in the code, it's good to be consistent.

sdks/python/apache_beam/testing/test_stream.py
```diff
@@ -171,13 +180,14 @@ class TestStream(PTransform):
   time. After all of the specified elements are emitted, ceases to produce
   output.
   """
```
lukecwik: Why do we need to declare output_tags here?
Are you trying to allow for outputs that have no events, otherwise shouldn't the tags come from the list of events?

The answer here impacts what we should be doing in expand in the no output_tags case in expand below.

sdks/python/apache_beam/testing/test_stream.py
```diff
@@ -188,7 +198,16 @@ def _infer_output_coder(self, input_type=None, input_coder=None):
   def expand(self, pbegin):
     assert isinstance(pbegin, pvalue.PBegin)
     self.pipeline = pbegin.pipeline
-    return pvalue.PCollection(self.pipeline, is_bounded=False)
+    if len(self.output_tags) == 0:
+      self.output_tags = set([None])
+
+    if len(self.output_tags) == 1:
```
davidyan74: Is there a reason why we treat the case with only one output_tag as a special case (i.e. not returning a dict) ? Please add the reason as a comment in this code.

sdks/python/apache_beam/testing/test_stream.py
```diff
@@ -171,13 +180,14 @@ class TestStream(PTransform):
...
   output.
   """
-  def __init__(self, coder=coders.FastPrimitivesCoder(), events=None):
+  def __init__(
+      self, coder=coders.FastPrimitivesCoder(), events=None, output_tags=None):
     super(TestStream, self).__init__()
     assert coder is not None
     self.coder = coder
     self.watermarks = {None: timestamp.MIN_TIMESTAMP}
     self._events = [] if events is None else list(events)
```
lukecwik: assert that events tags is a subset of output_tags

sdks/python/apache_beam/testing/test_stream.py
```diff
@@ -276,8 +295,11 @@ def to_runner_api_parameter(self, context):
   @PTransform.register_urn(
       common_urns.primitives.TEST_STREAM.urn,
       beam_runner_api_pb2.TestStreamPayload)
-  def from_runner_api_parameter(payload, context):
+  def from_runner_api_parameter(ptransform, payload, context):
     coder = context.coders.get_by_id(payload.coder_id)
+    output_tags = set(
+        None if k == 'None' else k for k in ptransform.outputs.keys())
```
davidyan74: Following up my other comment, what happens if the user tags their output to be the string 'None'?

sdks/python/apache_beam/runners/direct/direct_runner.py
```diff
@@ -73,60 +73,12 @@ class SwitchingDirectRunner(PipelineRunner):
   def is_fnapi_compatible(self):
     return BundleBasedDirectRunner.is_fnapi_compatible()
 
-  def apply_TestStream(self, transform, pbegin, options):
```
lukecwik: nice cleanup

sdks/python/apache_beam/testing/test_stream.py
```diff
@@ -133,15 +140,17 @@ def __eq__(self, other):
...
 
   def __hash__(self):
-    return hash(self.new_watermark)
+    return hash(str(self.new_watermark) + str(self.tag))
 
   def __lt__(self, other):
     return self.new_watermark < other.new_watermark
 
   def to_runner_api(self, unused_element_coder):
+    tag = 'None' if self.tag is None else self.tag
```
davidyan74: Is there a reason why the string 'None' is used as the tag here (instead of just None)? Is there a way to distinguish between a non-existent tag and a tag named "None"?

sdks/python/apache_beam/runners/direct/test_stream_impl.py
```diff
@@ -45,11 +46,69 @@ class _WatermarkController(PTransform):
...
+    if len(self.test_stream.output_tags) == 1:
+      return (
+          pbegin
+          | _TestStream(
+              self.test_stream.output_tags,
+              events=self.test_stream._events,
+              coder=self.test_stream.coder)
+          | _WatermarkController(list(self.test_stream.output_tags)[0]))
+
+    # This multiplexing the  multiple output PCollections.
```
lukecwik:
```suggestion
    # Multiplex to the correct PCollection based upon the event tag.
```

lukecwik:
retest this please

sdks/python/apache_beam/testing/test_stream.py
rohdesamuel: Looking through the codebase, it seems that 'None' is the special keyword used in the Python SDK to represent a tag that is specifically None. This keeps with the current style of the rest of the Python SDK. @lukecwik is this true?

sdks/python/apache_beam/testing/test_stream.py
rohdesamuel: > Are you trying to allow for outputs that have no events, otherwise shouldn't the tags come from the list of events?

Yep! The TestStreamService will allow users to define a TestStream with the output_tags specified at creation time and the events supplied at runtime.

sdks/python/apache_beam/testing/test_stream.py
rohdesamuel: Done.

sdks/python/apache_beam/testing/test_stream.py
rohdesamuel: Gotcha, changed to ```if not self.output_tags```

sdks/python/apache_beam/testing/test_stream.py
rohdesamuel: For backwards compatibility, existing code already relies on a single PCollection being returned.

sdks/python/apache_beam/testing/test_stream.py
rohdesamuel: This behavior isn't unique in the TestStream. This is consistent with the rest of the Python SDK with handling PCollection tags being None. As for what happens, I don't know, there are probably many subtle things that may go wrong.

sdks/python/apache_beam/testing/test_stream_test.py
rohdesamuel: Gotcha, removed the output_tags

sdks/python/apache_beam/testing/test_stream.py
rohdesamuel: done

rohdesamuel:
retest this please

lukecwik:
retest this please

sdks/python/apache_beam/testing/test_stream.py
```diff
@@ -133,15 +140,21 @@ def __eq__(self, other):
...
+
+    # Assert that no prevision is lost.
```
lukecwik:
```suggestion
    # Assert that no precision is lost.
```

sdks/python/apache_beam/testing/test_stream.py
```diff
@@ -133,15 +140,21 @@ def __eq__(self, other):
...
+    assert 1000 * (
+        self.new_watermark.micros // 1000) == self.new_watermark.micros
```
lukecwik: nit: This would be clearer by checking `%` has no remainder

sdks/python/apache_beam/testing/test_stream.py
lukecwik: This should be done for ProcessingTimeEvent as well

sdks/python/apache_beam/testing/test_stream.py
```diff
@@ -171,13 +184,20 @@ class TestStream(PTransform):
   time. After all of the specified elements are emitted, ceases to produce
   output.
   """
-  def __init__(self, coder=coders.FastPrimitivesCoder(), events=None):
+  def __init__(
```
lukecwik: Please add pydoc comments mentioning the few important pieces:
1) specifying the output_tags allows for adding outputs that produce no events
2) output_tags must be a superset of tags found in events if events is specified

sdks/python/apache_beam/testing/test_stream.py
```diff
@@ -171,13 +184,20 @@ class TestStream(PTransform):
   time. After all of the specified elements are emitted, ceases to produce
```
lukecwik: State that if only the default output or only one output tag has been specified, then a PCollection will be returned otherwise a dictionary of output name to PCollection.

PR #10919, opened 2020-02-21
[BEAM-9347] Don't overwrite default runner harness for unified worker

angoenka:

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/runners/dataflow/internal/apiclient.py
+++ b/sdks/python/apache_beam/runners/dataflow/internal/apiclient.py
@@ -186,8 +186,11 @@ class Environment(object):
     if job_type.startswith('FNAPI_'):
       self.debug_options = self.debug_options or DebugOptions()
       self.debug_options.experiments = self.debug_options.experiments or []
+      # Don't add the default image overwrite if user overwrites or
+      # if using unified worker.
       if not self.debug_options.lookup_experiment(
-          'runner_harness_container_image'):
+          'runner_harness_container_image') and not _use_unified_worker(
+              self.debug_options):
         runner_harness_override = (get_runner_harness_container_image())
         if runner_harness_override:
           self.debug_options.add_experiment(
@@ -915,6 +918,9 @@ def _use_unified_worker(pipeline_options):
     return False
   debug_options = pipeline_options.view_as(DebugOptions)
   use_unified_worker_flag = 'use_unified_worker'
+
+  if debug_options.lookup_experiment(use_unified_worker_flag):
+    return debug_options.lookup_experiment(use_unified_worker_flag)
 
   if debug_options.lookup_experiment('use_runner_v2'):
     debug_options.add_experiment(use_unified_worker_flag)
--- a/sdks/python/apache_beam/runners/dataflow/internal/apiclient.py
+++ b/sdks/python/apache_beam/runners/dataflow/internal/apiclient.py
@@ -186,11 +186,13 @@ class Environment(object):
     if job_type.startswith('FNAPI_'):
       self.debug_options = self.debug_options or DebugOptions()
       self.debug_options.experiments = self.debug_options.experiments or []
-      # Don't add the default image overwrite if user overwrites or
-      # if using unified worker.
-      if not self.debug_options.lookup_experiment(
-          'runner_harness_container_image') and not _use_unified_worker(
+      if self.debug_options.lookup_experiment(
+          'runner_harness_container_image') or _use_unified_worker(
               self.debug_options):
+        # Don't add the default image overwrite if user overwrites or
+        # if using unified worker.
+        pass
+      else:
         runner_harness_override = (get_runner_harness_container_image())
         if runner_harness_override:
           self.debug_options.add_experiment(
--- a/sdks/python/apache_beam/runners/dataflow/internal/apiclient.py
+++ b/sdks/python/apache_beam/runners/dataflow/internal/apiclient.py
@@ -184,13 +184,12 @@ class Environment(object):
     ])
     # TODO: Use enumerated type instead of strings for job types.
     if job_type.startswith('FNAPI_'):
-      self.debug_options = self.debug_options or DebugOptions()
       self.debug_options.experiments = self.debug_options.experiments or []
       if self.debug_options.lookup_experiment(
-          'runner_harness_container_image') or _use_unified_worker(
-              self.debug_options):
-        # Don't add the default image overwrite if user overwrites or
-        # if using unified worker.
+          'runner_harness_container_image') or _use_unified_worker(options):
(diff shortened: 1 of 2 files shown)


angoenka:
R: @ananvay @tvalentyn

ananvay:
Thanks Ankur! LGTM.

sdks/python/apache_beam/runners/dataflow/internal/apiclient.py
```diff
@@ -916,6 +919,9 @@ def _use_unified_worker(pipeline_options):
   debug_options = pipeline_options.view_as(DebugOptions)
   use_unified_worker_flag = 'use_unified_worker'
 
+  if debug_options.lookup_experiment(use_unified_worker_flag):
```
tvalentyn: PTAL at this method - we need to remove some of the statements in 922-928.

sdks/python/apache_beam/runners/dataflow/internal/apiclient.py
```diff
@@ -186,8 +186,11 @@ def __init__(self, packages, options, environment_version, pipeline_url):
     if job_type.startswith('FNAPI_'):
       self.debug_options = self.debug_options or DebugOptions()
       self.debug_options.experiments = self.debug_options.experiments or []
+      # Don't add the default image overwrite if user overwrites or
```
tvalentyn: I think it will be easier to read if we restructure this:
```
if (self.debug_options.lookup_experiment('runner_harness_container_image') or 
    _use_unified_worker(self.debug_options)):
   # Comment on WHY we don't want to set the override
   pass
else:
   <...set the override...>
```
WDYT?

tvalentyn: Also, should we be passing pipeline_options instead of debug options? Perhaps it works either way but passing pipeline options would be cleaner since that's what the signature of _use_unified_worker() expects.

sdks/python/apache_beam/runners/dataflow/internal/apiclient.py
tvalentyn: Do we need 922-924?

tvalentyn: nvm, reading this again, looks like this logic is intenional

tvalentyn: Ok, I think I understand the intent but I think it is confusing - can we move lines 925-926 to dataflow_runner.py, see: https://github.com/apache/beam/blob/1117508136ac5d0d66141c0e8f1c0830fb223a3a/sdks/python/apache_beam/runners/dataflow/dataflow_runner.py#L480, and replace  920-928 with line 928?

angoenka: We can move 925-926 to dataflow_runner.py but then i feel the logic to determine unified_worker usage will be more staggered. 
I think it will be better to keep the interpretation of `use_unified_worker` and `use_runner_v2` at a single place.
For now, we support both flag but I think we will transition to `use_runner_v2` over time (not sure by when).

sdks/python/apache_beam/runners/dataflow/internal/apiclient.py
angoenka: > I think it will be easier to read if we restructure this:
> 
> ```
> if (self.debug_options.lookup_experiment('runner_harness_container_image') or 
>     _use_unified_worker(self.debug_options)):
>    # Comment on WHY we don't want to set the override
>    pass
> else:
>    <...set the override...>
> ```
> 
> WDYT?
I think it makes sense. Will update it.




> Also, should we be passing pipeline_options instead of debug options? Perhaps it works either way but passing pipeline options would be cleaner since that's what the signature of _use_unified_worker() expects.

I think the code is structured in a way that a `self.debug_options` is referenced every where and the reference to it can be different from `options` passed in the method argument. So I think it will need a more thoughtful refactoring to move to change it to options in the right manner. Which I think we can differ.

sdks/python/apache_beam/runners/dataflow/internal/apiclient.py
tvalentyn: I think that it would be cleaner to not modify experiments in `_use_unified_worker`, since the name of this method does not suggest that it modifies the state, and it would also be easier to follow the logic. Also, if Dataflow runner relies on the fact that "use_runner_v2" experiment is set, it might be better to set it explicitly at some point, instead of relying on the fact that "_use_unified_worker" will be called. Leaving this up to you.

sdks/python/apache_beam/runners/dataflow/internal/apiclient.py
```diff
@@ -186,8 +186,13 @@ def __init__(self, packages, options, environment_version, pipeline_url):
     if job_type.startswith('FNAPI_'):
       self.debug_options = self.debug_options or DebugOptions()
       self.debug_options.experiments = self.debug_options.experiments or []
-      if not self.debug_options.lookup_experiment(
-          'runner_harness_container_image'):
+      if self.debug_options.lookup_experiment(
```
tvalentyn: Few thoughts here 
1. I think we cannot instantiate `DebugOptions()` in `self.debug_options = self.debug_options or DebugOptions()`. We should create a view of existing `options`.  All views of pipeline options should be views of the same underlying object, and I think that line will break it. Do you know why we added it?
2. For the purpose of this PR it should be sufficient to pass `options` into `_use_unified_worker()` in line 190.

sdks/python/apache_beam/runners/dataflow/internal/apiclient.py
```diff
@@ -186,8 +186,13 @@ def __init__(self, packages, options, environment_version, pipeline_url):
...
+          'runner_harness_container_image') or _use_unified_worker(
+              self.debug_options):
+        # Don't add the default image overwrite if user overwrites or
```
tvalentyn: This comment repeats the logic; I thought you might want to explain why we don't want to pin the image for UW. 
Consider: "In case of using UW, harness container should be defined by DF service unless explicitly overridden."

sdks/python/apache_beam/runners/dataflow/internal/apiclient.py
angoenka: I agree. I think I added it because in certain case it was passed in test. I might be wrong though. 

Makes sense. I will update the PR with both these suggestions.

sdks/python/apache_beam/runners/dataflow/internal/apiclient.py
angoenka: Updates the comment.

angoenka:
Thanks @tvalentyn

PR #10924, opened 2020-02-21
Revert "[BEAM-9085] Fix performance regression in SyntheticSource"

mwalenia:
Reverts apache/beam#10885

Changes in the pseudorandom number generator turned out to impact Python 2 based tests more than expected.

### DIFF
--- a/sdks/python/apache_beam/testing/synthetic_pipeline.py
+++ b/sdks/python/apache_beam/testing/synthetic_pipeline.py
@@ -41,10 +41,7 @@ import argparse
 import json
 import logging
 import math
-import random
-import struct
 import time
-from builtins import range
 
 import apache_beam as beam
 from apache_beam.io import WriteToText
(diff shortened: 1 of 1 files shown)


mwalenia:
R: @tvalentyn @kamilwu

kamilwu:
Some examples of slowdown I've noticed so far:
1. BQ 10GB write test [[link](https://builds.apache.org/view/A-D/view/Beam/view/All/job/beam_BiqQueryIO_Write_Performance_Test_Python_Batch/226/)]. Job execution time went from ~12 min to 42 minutes.
2. CoGBK test [[link](https://builds.apache.org/view/A-D/view/Beam/view/All/job/beam_LoadTests_Python_CoGBK_Dataflow_Batch/224/)] failed completely with a message "The worker lost contact with the service"

I think the best we can do is to revert changes to keep current py2 based tests working. After that, I'm going to take a look at the fix once again and find a better solution that doesn't harm Python 2 performance.

tvalentyn:
Thank you; some of the tests we have internally at Google also started timing out with this change.

PR #10938, opened 2020-02-22
[BEAM-9352] Align version of transitive jackson dependencies with Beam version

iemejia:
Linkage checker analysis
https://gist.github.com/iemejia/9680bd161eeaf36e7a0963c7853f0cbc

R: @lukecwik


iemejia:
retest this please

iemejia:

iemejia:
Run SQL PostCommit

iemejia:
Run PythonLint PreCommit

iemejia:

iemejia:
Run Python PreCommit

iemejia:
Run Portable_Python PreCommit

iemejia:
Run Python PreCommit

lukecwik:
Adding the deps to runtimeOnly doesn't fix the generated pom our users will have.

Would it make sense to mark them as provided or better yet make them part of the Apache Beam BOM?
Should we make all our "fixed" dependency versions part of the Apache Beam BOM?

iemejia:
You are right, fixed now. I am not really that familiar with Beam's BOM, but can try to tackle this afterwards if we agree that it is a good idea. Opinions on that maybe also @suztomo.

PR #10939, opened 2020-02-22
[BEAM-9342] Upgrade vendored bytebuddy to version 1.10.8

iemejia:
The version of bytebuddy Beam is using does not officially support Java 11 and it is more than 1 year old. We should upgrade to a more recent version so we can benefit of the new improvements of the library as well as of ASM.
R: @lukecwik


iemejia:

iemejia:
Run PythonLint PreCommit

iemejia:

iemejia:
Run Python PreCommit

iemejia:
Run Portable_Python PreCommit

PR #10946, opened 2020-02-24
[BEAM-9363] TUMBLE as TVF

amaliujia:
Relevant doc: https://s.apache.org/streaming-beam-sql 

Support 
```
"select * from TUMBLE((select * from KeyValue), descriptor(ts), 'INTERVAL 1 SECOND')
```

Note that ZetaSQL needs an improvement to fully support TVF windowing. This PR adds a skeleton as a starter for moving streaming windowing to `FROM` clause by using TVF. 

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/planner/BeamRuleSets.java
+++ b/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/planner/BeamRuleSets.java
@@ -34,6 +34,7 @@ import org.apache.beam.sdk.extensions.sql.impl.rule.BeamMinusRule;
 import org.apache.beam.sdk.extensions.sql.impl.rule.BeamSideInputJoinRule;
 import org.apache.beam.sdk.extensions.sql.impl.rule.BeamSideInputLookupJoinRule;
 import org.apache.beam.sdk.extensions.sql.impl.rule.BeamSortRule;
+import org.apache.beam.sdk.extensions.sql.impl.rule.BeamTableFunctionScanRule;
 import org.apache.beam.sdk.extensions.sql.impl.rule.BeamUncollectRule;
 import org.apache.beam.sdk.extensions.sql.impl.rule.BeamUnionRule;
 import org.apache.beam.sdk.extensions.sql.impl.rule.BeamUnnestRule;
@@ -133,7 +134,7 @@ public class BeamRuleSets {
           // remove unnecessary sort rule
           // https://issues.apache.org/jira/browse/BEAM-5073
           // SortRemoveRule.INSTANCE,
-
+          BeamTableFunctionScanRule.INSTANCE,
           // prune empty results rules
           PruneEmptyRules.AGGREGATE_INSTANCE,
           PruneEmptyRules.FILTER_INSTANCE,
--- a/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamTableFunctionScanRel.java
+++ b/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamTableFunctionScanRel.java
@@ -48,6 +48,10 @@ import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rex.RexLiteral;
 import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rex.RexNode;
 import org.joda.time.Duration;
 
+/**
+ * BeamRelNode to replace {@code TableFunctionScan}. Currently this class limits to support
+ * table-valued function for streaming windowing.
+ */
 public class BeamTableFunctionScanRel extends TableFunctionScan implements BeamRelNode {
   public BeamTableFunctionScanRel(
       RelOptCluster cluster,
@@ -83,12 +87,14 @@ public class BeamTableFunctionScanRel extends TableFunctionScan implements BeamR
     public PCollection<Row> expand(PCollectionList<Row> input) {
       checkArgument(
           input.size() == 1,
-          "Wrong number of inputs for %s: %s",
+          "Wrong number of inputs for %s, expected 1 input but received: %s",
           BeamTableFunctionScanRel.class.getSimpleName(),
           input);
+      String operatorName = ((RexCall) getCall()).getOperator().getName();
       checkArgument(
-          ((RexCall) getCall()).getOperator().getName().equals("TUMBLE"),
-          "Only support TUMBLE table-valued function");
+          operatorName.equals("TUMBLE"),
+          "Only support TUMBLE table-valued function. Current operator: %s",
+          operatorName);
       RexCall call = ((RexCall) getCall());
       RexInputRef wmCol = (RexInputRef) call.getOperands().get(1);
       PCollection<Row> upstream = input.get(0);
@@ -133,7 +139,6 @@ public class BeamTableFunctionScanRel extends TableFunctionScan implements BeamR
       IntervalWindow window = windowFn.assignWindow(row.getDateTime(windowFieldIndex).toInstant());
       Row.Builder builder = Row.withSchema(outputSchema);
       builder.addValues(row.getValues());
-      // TODO: add window_start and window_end;
       builder.addValue(window.start());
       builder.addValue(window.end());
       c.output(builder.build());
--- a/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamTableFunctionScanRel.java
+++ b/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamTableFunctionScanRel.java
@@ -73,7 +73,7 @@ public class BeamTableFunctionScanRel extends TableFunctionScan implements BeamR
       RelDataType relDataType,
       Set<RelColumnMapping> set) {
     return new BeamTableFunctionScanRel(
-        getCluster(), traitSet, list, rexNode, type, relDataType, columnMappings);
+        getCluster(), traitSet, list, rexNode, type, relDataType, set);
   }
 
   @Override
@@ -123,9 +123,9 @@ public class BeamTableFunctionScanRel extends TableFunctionScan implements BeamR
   }
 
   private static class FixedWindowDoFn extends DoFn<Row, Row> {
-    private int windowFieldIndex;
-    private FixedWindows windowFn;
-    private Schema outputSchema;
+    private final int windowFieldIndex;
+    private final FixedWindows windowFn;
+    private final Schema outputSchema;
 
     public FixedWindowDoFn(FixedWindows windowFn, int windowFieldIndex, Schema schema) {
       this.windowFn = windowFn;
--- a/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamTableFunctionScanRel.java
+++ b/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamTableFunctionScanRel.java
@@ -1,0 +1,157 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.beam.sdk.extensions.sql.impl.rel;
+
+import static org.apache.beam.vendor.calcite.v1_20_0.com.google.common.base.Preconditions.checkArgument;
+
+import java.lang.reflect.Type;
+import java.util.List;
+import java.util.Set;
+import org.apache.beam.sdk.extensions.sql.impl.planner.BeamCostModel;
+import org.apache.beam.sdk.extensions.sql.impl.planner.NodeStats;
+import org.apache.beam.sdk.extensions.sql.impl.utils.CalciteUtils;
+import org.apache.beam.sdk.schemas.Schema;
+import org.apache.beam.sdk.transforms.DoFn;
+import org.apache.beam.sdk.transforms.PTransform;
+import org.apache.beam.sdk.transforms.ParDo;
+import org.apache.beam.sdk.transforms.windowing.FixedWindows;
+import org.apache.beam.sdk.transforms.windowing.IntervalWindow;
+import org.apache.beam.sdk.values.PCollection;
+import org.apache.beam.sdk.values.PCollectionList;
+import org.apache.beam.sdk.values.Row;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.plan.RelOptCluster;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.plan.RelOptPlanner;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.plan.RelTraitSet;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rel.RelNode;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rel.core.TableFunctionScan;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rel.metadata.RelColumnMapping;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rel.metadata.RelMetadataQuery;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rel.type.RelDataType;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rex.RexCall;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rex.RexInputRef;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rex.RexLiteral;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rex.RexNode;
+import org.joda.time.Duration;
+
+public class BeamTableFunctionScanRel extends TableFunctionScan implements BeamRelNode {
+  public BeamTableFunctionScanRel(
+      RelOptCluster cluster,
+      RelTraitSet traitSet,
+      List<RelNode> inputs,
+      RexNode rexCall,
+      Type elementType,
+      RelDataType rowType,
+      Set<RelColumnMapping> columnMappings) {
+    super(cluster, traitSet, inputs, rexCall, elementType, rowType, columnMappings);
+  }
+
+  @Override
+  public TableFunctionScan copy(
+      RelTraitSet traitSet,
+      List<RelNode> list,
+      RexNode rexNode,
+      Type type,
+      RelDataType relDataType,
+      Set<RelColumnMapping> set) {
+    return new BeamTableFunctionScanRel(
+        getCluster(), traitSet, list, rexNode, type, relDataType, columnMappings);
+  }
+
+  @Override
+  public PTransform<PCollectionList<Row>, PCollection<Row>> buildPTransform() {
+    return new Transform();
+  }
+
+  private class Transform extends PTransform<PCollectionList<Row>, PCollection<Row>> {
+
+    @Override
+    public PCollection<Row> expand(PCollectionList<Row> input) {
+      checkArgument(
+          input.size() == 1,
+          "Wrong number of inputs for %s: %s",
+          BeamTableFunctionScanRel.class.getSimpleName(),
+          input);
+      checkArgument(
+          ((RexCall) getCall()).getOperator().getName().equals("TUMBLE"),
+          "Only support TUMBLE table-valued function");
+      RexCall call = ((RexCall) getCall());
+      RexInputRef wmCol = (RexInputRef) call.getOperands().get(1);
+      PCollection<Row> upstream = input.get(0);
+      Schema outputSchema = CalciteUtils.toSchema(getRowType());
+      return upstream
+          .apply(
+              ParDo.of(
+                  new FixedWindowDoFn(
+                      FixedWindows.of(durationParameter(call.getOperands().get(2))),
+                      wmCol.getIndex(),
+                      outputSchema)))
+          .setRowSchema(outputSchema);
+    }
+  }
+
+  private Duration durationParameter(RexNode node) {
+    return Duration.millis(longValue(node));
+  }
+
+  private long longValue(RexNode operand) {
+    if (operand instanceof RexLiteral) {
+      return ((Number) RexLiteral.value(operand)).longValue();
+    } else {
+      throw new IllegalArgumentException(String.format("[%s] is not valid.", operand));
+    }
+  }
+
+  private static class FixedWindowDoFn extends DoFn<Row, Row> {
+    private int windowFieldIndex;
+    private FixedWindows windowFn;
+    private Schema outputSchema;
+
+    public FixedWindowDoFn(FixedWindows windowFn, int windowFieldIndex, Schema schema) {
+      this.windowFn = windowFn;
+      this.windowFieldIndex = windowFieldIndex;
+      this.outputSchema = schema;
+    }
+
+    @ProcessElement
+    public void processElement(ProcessContext c) {
+      Row row = c.element();
+      IntervalWindow window = windowFn.assignWindow(row.getDateTime(windowFieldIndex).toInstant());
+      Row.Builder builder = Row.withSchema(outputSchema);
+      builder.addValues(row.getValues());
+      // TODO: add window_start and window_end;
+      builder.addValue(window.start());
+      builder.addValue(window.end());
+      c.output(builder.build());
+    }
+  }
+
+  @Override
+  public NodeStats estimateNodeStats(RelMetadataQuery mq) {
+    return BeamSqlRelUtils.getNodeStats(getInput(0), mq);
+  }
+
+  @Override
+  public BeamCostModel beamComputeSelfCost(RelOptPlanner planner, RelMetadataQuery mq) {
+    NodeStats inputEstimates = BeamSqlRelUtils.getNodeStats(getInput(0), mq);
+
+    final double rowSize = getRowType().getFieldCount();
+    final double cpu = inputEstimates.getRowCount() * inputEstimates.getRowCount() * rowSize;
+    final double cpuRate = inputEstimates.getRate() * inputEstimates.getWindow() * rowSize;
+    return BeamCostModel.FACTORY.makeCost(cpu, cpuRate);
+  }
+}
--- a/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamTableFunctionScanRel.java
+++ b/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamTableFunctionScanRel.java
@@ -155,7 +155,7 @@ public class BeamTableFunctionScanRel extends TableFunctionScan implements BeamR
     NodeStats inputEstimates = BeamSqlRelUtils.getNodeStats(getInput(0), mq);
 
     final double rowSize = getRowType().getFieldCount();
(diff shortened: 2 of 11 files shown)


amaliujia:
R: @kennknowles

amaliujia:
Run Java PreCommit

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamTableFunctionScanRel.java
```diff
@@ -0,0 +1,157 @@
...
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rel.metadata.RelColumnMapping;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rel.metadata.RelMetadataQuery;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rel.type.RelDataType;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rex.RexCall;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rex.RexInputRef;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rex.RexLiteral;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rex.RexNode;
+import org.joda.time.Duration;
+
+public class BeamTableFunctionScanRel extends TableFunctionScan implements BeamRelNode {
```
kennknowles, line 51: This is named like it is a generic table function scan, but it seems to implement just a very specialized table function. Shouldn't the function be a parameter to the rel?

amaliujia: I agree your parameter.

Given that only windowing TVF is the only TVF usage, having or not having a parameter does not matter, because I don't have a concrete idea what other table-valued function we will support and how.

I would like to leave parametrization to the future when there is a need to support other TVF.  Either we limit windowing to `BeamWindowingTableFunctionScanRel` and have another Rel for other usage, or still have a general Rel but support all. It's not clear to me which approach is better, and that might can be answered when there is a need to support other  usage.

Also to clarify: I do have a plan to support not only UDF, UDAF, but also user-defined tabled function, and that will happen in near term.

amaliujia:
Run Java PreCommit

amaliujia:
cc @tysonjh in case you are interested in.

amaliujia:
A friendly ping on this PR.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamTableFunctionScanRel.java
tysonjh, line 51: Can you add a class comment please?

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/planner/BeamRuleSets.java
```diff
@@ -133,7 +134,7 @@
           // remove unnecessary sort rule
           // https://issues.apache.org/jira/browse/BEAM-5073
```
tysonjh, line 135: This is now closed. Maybe it is worth cleaning up the comment now?

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamTableFunctionScanRel.java
```diff
@@ -0,0 +1,157 @@
...
+    return new Transform();
+  }
+
+  private class Transform extends PTransform<PCollectionList<Row>, PCollection<Row>> {
+
+    @Override
+    public PCollection<Row> expand(PCollectionList<Row> input) {
+      checkArgument(
+          input.size() == 1,
+          "Wrong number of inputs for %s: %s",
```
tysonjh, line 86: I don't know checkArgument too well, does it output the expected size? E.g. it would be nice if this error message said something like,

%s expected 1 input but received %s

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamTableFunctionScanRel.java
```diff
@@ -0,0 +1,157 @@
...
+          BeamTableFunctionScanRel.class.getSimpleName(),
+          input);
+      checkArgument(
+          ((RexCall) getCall()).getOperator().getName().equals("TUMBLE"),
+          "Only support TUMBLE table-valued function");
```
tysonjh, line 91: Here too. It would be nice if the error message included the offending operator name.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamTableFunctionScanRel.java
```diff
@@ -0,0 +1,157 @@
...
+      this.outputSchema = schema;
+    }
+
+    @ProcessElement
+    public void processElement(ProcessContext c) {
+      Row row = c.element();
+      IntervalWindow window = windowFn.assignWindow(row.getDateTime(windowFieldIndex).toInstant());
+      Row.Builder builder = Row.withSchema(outputSchema);
+      builder.addValues(row.getValues());
+      // TODO: add window_start and window_end;
```
tysonjh, line 136: If this isn't to be resolved in this PR, would having an associated Jira be appropriate?

Or... is this TODO out of date? I see window start and end are used below.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamTableFunctionScanRel.java
```diff
@@ -0,0 +1,157 @@
...
+    return BeamSqlRelUtils.getNodeStats(getInput(0), mq);
+  }
+
+  @Override
+  public BeamCostModel beamComputeSelfCost(RelOptPlanner planner, RelMetadataQuery mq) {
+    NodeStats inputEstimates = BeamSqlRelUtils.getNodeStats(getInput(0), mq);
+
+    final double rowSize = getRowType().getFieldCount();
+    final double cpu = inputEstimates.getRowCount() * inputEstimates.getRowCount() * rowSize;
+    final double cpuRate = inputEstimates.getRate() * inputEstimates.getWindow() * rowSize;
```
tysonjh, line 154: Can you explain the logic here? It may be worth adding a comment to describe the calculation.

For example, how does cpu = row_count^2 * row_size ? It isn't obvious to me.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rule/BeamTableFunctionScanRule.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.List;
+import org.apache.beam.sdk.extensions.sql.impl.rel.BeamLogicalConvention;
+import org.apache.beam.sdk.extensions.sql.impl.rel.BeamTableFunctionScanRel;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.plan.Convention;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rel.RelNode;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rel.convert.ConverterRule;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rel.core.TableFunctionScan;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rel.logical.LogicalTableFunctionScan;
+
+public class BeamTableFunctionScanRule extends ConverterRule {
```
tysonjh, line 31: Please add a class comment.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rule/BeamTableFunctionScanRule.java
```diff
@@ -0,0 +1,60 @@
...
+        LogicalTableFunctionScan.class,
+        Convention.NONE,
+        BeamLogicalConvention.INSTANCE,
+        "BeamTableFunctionScanRule");
+  }
+
+  @Override
+  public RelNode convert(RelNode relNode) {
+    TableFunctionScan tableFunctionScan = (TableFunctionScan) relNode;
+    // only support one input for table function scan.
```
tysonjh, line 45: Should this be a checkArgument?

sdks/java/extensions/sql/zetasql/src/main/java/org/apache/beam/sdk/extensions/sql/zetasql/SqlWindowTableFunction.java
```diff
@@ -0,0 +1,113 @@
...
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.sql.SqlFunctionCategory;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.sql.SqlKind;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.sql.SqlNode;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.sql.SqlOperandCountRange;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.sql.type.SqlOperandCountRanges;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.sql.type.SqlReturnTypeInference;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.sql.type.SqlTypeName;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.sql.validate.SqlValidator;
+
+public class SqlWindowTableFunction extends SqlFunction {
```
tysonjh, line 37: Please add class comment.

sdks/java/extensions/sql/zetasql/src/main/java/org/apache/beam/sdk/extensions/sql/zetasql/SqlWindowTableFunction.java
```diff
@@ -0,0 +1,113 @@
...
+
+  @Override
+  public SqlOperandCountRange getOperandCountRange() {
+    return SqlOperandCountRanges.of(3);
+  }
+
+  @Override
+  public boolean checkOperandTypes(SqlCallBinding callBinding, boolean throwOnFailure) {
+    // There should only be three operands, and number of operands are checked before
+    // this call.
```
tysonjh, line 56: For my own understanding, where are they checked?

sdks/java/extensions/sql/zetasql/src/main/java/org/apache/beam/sdk/extensions/sql/zetasql/translation/ExpressionConverter.java
```diff
@@ -551,6 +561,60 @@ public RexNode convertResolvedLiteral(ResolvedLiteral resolvedLiteral) {
     return ret;
   }
 
+  public RexCall convertTableValuedFunction(
```
tysonjh, line 564: Please add a method comment since it is public and not 'simple' like a getter or setter.

sdks/java/extensions/sql/zetasql/src/main/java/org/apache/beam/sdk/extensions/sql/zetasql/translation/ExpressionConverter.java
```diff
@@ -551,6 +561,60 @@ public RexNode convertResolvedLiteral(ResolvedLiteral resolvedLiteral) {
...
+      List<ResolvedColumn> inputTableColumns) {
+    switch (tvf.getName()) {
+      case "TUMBLE":
+        // TUMBLE tvf's second argument is descriptor.
+        ResolvedColumn wmCol =
+            extractWatermarkColumnFromDescriptor(argumentList.get(1).getDescriptorArg());
+        if (wmCol.getType().getKind() != TYPE_TIMESTAMP) {
+          throw new IllegalArgumentException(
+              "Watermarked column should be TIMESTAMP type: "
+                  + extractWatermarkColumnNameFromDescriptor(
```
tysonjh, line 577: Can you do wmCol.getName() or is this different?

sdks/java/extensions/sql/zetasql/src/main/java/org/apache/beam/sdk/extensions/sql/zetasql/translation/ExpressionConverter.java
```diff
@@ -551,6 +561,60 @@ public RexNode convertResolvedLiteral(ResolvedLiteral resolvedLiteral) {
...
+    for (int i = 0; i < inputTableColumns.size(); i++) {
+      if (inputTableColumns.get(i).equals(wmCol)) {
+        return rexBuilder()
+            .makeInputRef(TypeUtils.toRelDataType(rexBuilder(), wmCol.getType(), false), i);
+      }
+    }
+
+    // ZetaSQL parser guarantees that wmCol can be found from inputTableColumns.
+    // so it shouldn't reach here.
+    throw new IllegalArgumentException();
```
tysonjh, line 605: Would this be more descriptive as an IllegalStateException? Either way having a message would be useful here.

sdks/java/extensions/sql/zetasql/src/main/java/org/apache/beam/sdk/extensions/sql/zetasql/translation/TVFScanConverter.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.List;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rel.RelNode;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rel.logical.LogicalTableFunctionScan;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rel.type.RelDataType;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rel.type.RelDataTypeField;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rel.type.RelDataTypeFieldImpl;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.rel.type.RelRecordType;
+import org.apache.beam.vendor.calcite.v1_20_0.org.apache.calcite.sql.type.SqlTypeName;
+
+class TVFScanConverter extends RelConverter<ResolvedTVFScan> {
```
tysonjh, line 34: Please add a class comment. They're really useful and should be on all classes and non-trivial public methods.

sdks/java/extensions/sql/zetasql/src/main/java/org/apache/beam/sdk/extensions/sql/zetasql/translation/TVFScanConverter.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+    return inputs;
+  }
+
+  private RelDataType createRowTypeWithWindowStartAndEnd(RelDataType inputRowType) {
+    List<RelDataTypeField> newFields = new ArrayList<>(inputRowType.getFieldList());
+    RelDataType timestampType = getCluster().getTypeFactory().createSqlType(SqlTypeName.TIMESTAMP);
+
+    RelDataTypeField windowStartField =
+        new RelDataTypeFieldImpl("window_start", newFields.size(), timestampType);
```
tysonjh, line 76: I've seen window_start and window_end come up a few times in this PR. Is there a reasonable place to have these as constants and would it make sense to do this?

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamTableFunctionScanRel.java
amaliujia: Done

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/planner/BeamRuleSets.java
amaliujia: Nice catch. I decided to re-open that Jira (so keep the work for re-enabling SortRemoveRule). 

Keep comments as what it is now.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamTableFunctionScanRel.java
amaliujia: Nice idea! Done!

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamTableFunctionScanRel.java
amaliujia: Ah it is out-of-date... Removed

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rule/BeamTableFunctionScanRule.java
amaliujia: Done

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rule/BeamTableFunctionScanRule.java
amaliujia: Nice catch. This is a ConverterRule whose `canConvert` returns always `true`. I mis-remembered it is `RelOptRule` in which the input size check should be in `canConvert` (and in fact there is no `canConvert` override).

Added the checkArgument.

sdks/java/extensions/sql/zetasql/src/main/java/org/apache/beam/sdk/extensions/sql/zetasql/SqlWindowTableFunction.java
amaliujia: Done

sdks/java/extensions/sql/zetasql/src/main/java/org/apache/beam/sdk/extensions/sql/zetasql/SqlWindowTableFunction.java
amaliujia: There is a method in this class:
```
  @Override
  public SqlOperandCountRange getOperandCountRange() {
    return SqlOperandCountRanges.of(3);
  }
```

Calcite will call this method to validate the size of arguments.

sdks/java/extensions/sql/zetasql/src/main/java/org/apache/beam/sdk/extensions/sql/zetasql/translation/ExpressionConverter.java
amaliujia: Done.

sdks/java/extensions/sql/zetasql/src/main/java/org/apache/beam/sdk/extensions/sql/zetasql/translation/ExpressionConverter.java
amaliujia: I implemented `descriptor` in ZetaSQL. Per my implementation, a descriptor argument contains two class members:
a. `ImmutableList<ResolvedColumn>`
b. `ImmutableList<String>`

The `ImmutableList<String>` contains exact column names where `ImmutableList<ResolvedColumn>` contains a list of `ColumnRef` that connects these exact columns names with columns from input table. 

I am not 100% sure whether names in those `ColumnRef` always match exact names. So use `extractWatermarkColumnNameFromDescriptor` method to have a 100% guarantee on getting columns names that users typed into their queries.

sdks/java/extensions/sql/zetasql/src/main/java/org/apache/beam/sdk/extensions/sql/zetasql/translation/ExpressionConverter.java
amaliujia: I moved comments here to exception as the exception message. I don't have a concert description for in which normal case it will reach this step. It could reach this step when ZetaSQL has bugs, but putting something like "ZetaSQL has bugs to make you reach this exception" sounds not useful neither.

sdks/java/extensions/sql/zetasql/src/main/java/org/apache/beam/sdk/extensions/sql/zetasql/translation/TVFScanConverter.java
amaliujia: Done.

sdks/java/extensions/sql/zetasql/src/main/java/org/apache/beam/sdk/extensions/sql/zetasql/translation/TVFScanConverter.java
amaliujia: Good idea. I added a util class named `TVFStreamingUtils` to encapsulate common constants and util functions for TVF streaming.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamTableFunctionScanRel.java
amaliujia: I believe it is carried over from a class in BeamSQL. Such computation was a part of a intern project before about SQL cost based optimization.

Let me do some research and see how it should work.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamTableFunctionScanRel.java
amaliujia: Done.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamTableFunctionScanRel.java
```diff
@@ -0,0 +1,162 @@
...
+          "Only support TUMBLE table-valued function. Current operator: %s",
+          operatorName);
+      RexCall call = ((RexCall) getCall());
+      RexInputRef wmCol = (RexInputRef) call.getOperands().get(1);
+      PCollection<Row> upstream = input.get(0);
+      Schema outputSchema = CalciteUtils.toSchema(getRowType());
+      return upstream
+          .apply(
+              ParDo.of(
+                  new FixedWindowDoFn(
```
kennknowles, line 105: This still appears to actually be a "FixedWindowsTableFunctionRel".

amaliujia: Interesting point. Let me check if there are benefits to split TUMBLE/HOP/SESSION at Rel level, especially what is the impact on JOIN.

amaliujia: Ok, I have thought about this. Can we keep what this code is since it will still fast evolving. 

There are a few dimensions to consider:
1. if TUMBLE, SESSION, HOP have their own Rels
2. How many Rels needed for JOIN with windowing 


I believe we will have better answer after we see HOP and SESSION implementation.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamTableFunctionScanRel.java
amaliujia: Yeah the cpu computation is wrong (I copied it from SortRel).

The reasonable computation is `cpu = inputEstimates.getRowCount() * rowSize * function_per_row_cost_factor;`. Because we don't have a reasonable stats for `function_per_row_cost_factor`, the formula is simplified as row count times row size.

amaliujia:
Run Java PreCommit

amaliujia:
Have tried to address comments. Please take another look.

amaliujia:
Run Java PreCommit

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamTableFunctionScanRel.java
kennknowles: My point is really that committing the code like this makes code with names that don't make sense. If you get interrupted to do something else after this PR, then the codebase will be in an incorrect state.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamTableFunctionScanRel.java
```diff
@@ -0,0 +1,162 @@
...
+  @Override
+  public TableFunctionScan copy(
+      RelTraitSet traitSet,
+      List<RelNode> list,
+      RexNode rexNode,
+      Type type,
+      RelDataType relDataType,
+      Set<RelColumnMapping> set) {
+    return new BeamTableFunctionScanRel(
+        getCluster(), traitSet, list, rexNode, type, relDataType, columnMappings);
```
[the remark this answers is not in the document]
amaliujia: Good catch! Replaced with the right parameter.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/utils/TVFStreamingUtils.java
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.beam.sdk.extensions.sql.impl.utils;
+
+/** Provides static constants or utils for TVF streaming. */
+public class TVFStreamingUtils {
```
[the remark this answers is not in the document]
amaliujia: `SqlWindowTableFunction` is a class in Calcite (since 1.22.0). After we successfully upgrade to newer version of Calcite (I hope), we can remove `SqlWindowTableFunction`, thus there is a need to keep a `TVFStreamingUtils`.

There could be an argument though that such constants can be put into  `SqlWindowTableFunction` in Calcite. We can leave such discussion in the future.

sdks/java/extensions/sql/zetasql/src/main/java/org/apache/beam/sdk/extensions/sql/zetasql/SqlAnalyzer.java
```diff
@@ -175,6 +185,37 @@ private void addBuiltinFunctionsToCatalog(SimpleCatalog catalog, AnalyzerOptions
...
+                .setDescriptorResolutionTableOffset(0)
+                .build(),
+            1);
+
+    FunctionArgumentType stringType =
+        new FunctionArgumentType(TypeFactory.createSimpleType(TypeKind.TYPE_STRING));
+
+    // TUMBLE
+    catalog.addTableValuedFunction(
+        new TableValuedFunction.ForwardInputSchemaToOutputSchemaWithAppendedColumnTVF(
```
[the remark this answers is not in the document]
amaliujia: Yeah. And at least I am glad this name itself does not exceed 80 chars so it at least fits into a single line..

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamTableFunctionScanRel.java
amaliujia: I guess I got what you meant: you think it should be a PTransform, not a DoFn for this `FixedWindow` assignment? (I was confused on why this transforming should be a Rel)

If so, either PTransform or a DoFn seems work to me.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/utils/TVFStreamingUtils.java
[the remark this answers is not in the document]
apilloud: In that case, drop the constants all together. You can't reference a class in Beam from Calcite, and these constants are used in that class.

amaliujia:
Run Java PreCommit

PR #10947, opened 2020-02-24
[BEAM-9364] Refactor KafkaIO to use DeserializerProviders

iemejia:
This builds on your previous PR with some API improvements. Since this adjusts the public API that will be out as part of 2.20.0 is probably better if we include it as part of 2.20.0 too.

R: @aromanenko-dev


aromanenko-dev:
Run Java KafkaIO Performance Test

iemejia:
Weird it did not fail, but it is referrencing and incorrect (older) run. See here for the actual run
https://builds.apache.org/job/beam_PerformanceTests_Kafka_IO/452/

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
```diff
@@ -600,7 +599,20 @@ public void setValueDeserializer(String valueDeserializer) {
...
         Class<? extends Deserializer<K>> keyDeserializer, Coder<K> keyCoder) {
-      return toBuilder().setKeyDeserializer(keyDeserializer).setKeyCoder(keyCoder).build();
+      return withKeyDeserializer(keyDeserializer).toBuilder().setKeyCoder(keyCoder).build();
+    }
+
+    /**
+     * Specify Confluent schema registry url and subject to get a deserializer (and schema) from
+     * Confluent Schema Registry.
+     */
+    public Read<K, V> withKeyDeserializer(String schemaRegistryUrl, String subject) {
```
aromanenko-dev: I'd suggest to extract these two string arguments into dedicated Confluent Schema Registry input arguments object to avoid potential collisions with other `withKeyDeserializer(String, String)` methods that can be added in the future. Also, it can be extended (if needed) and it will make clear that this is used only for CSR's schema.

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
```diff
@@ -612,7 +624,10 @@ public void setValueDeserializer(String valueDeserializer) {
      * Coder)} to provide the value coder explicitly.
      */
     public Read<K, V> withValueDeserializer(Class<? extends Deserializer<V>> valueDeserializer) {
-      return toBuilder().setValueDeserializer(valueDeserializer).build();
+      return toBuilder()
+          .setValueDeserializer(valueDeserializer)
```
aromanenko-dev: The same question as for `withKeyDeserializer`

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
```diff
@@ -624,40 +639,20 @@ public void setValueDeserializer(String valueDeserializer) {
...
-        builder.setValueDeserializer((Class) KafkaAvroDeserializer.class);
-      }
-      return builder
-          .setCSRClientProvider(csrClientProvider)
-          .build()
-          .withConsumerConfigUpdates(
-              ImmutableMap.of(
-                  KafkaAvroDeserializerConfig.SCHEMA_REGISTRY_URL_CONFIG,
-                  csrClientProvider.getSchemaRegistryUrl()));
+    public Read<K, V> withValueDeserializer(String schemaRegistryUrl, String subject) {
```
aromanenko-dev: The same note as for `withKeyDeserializer(String, String)`

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/LocalDeserializerProvider.java
```diff
@@ -0,0 +1,88 @@
...
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.util.Map;
+import org.apache.beam.sdk.coders.CannotProvideCoderException;
+import org.apache.beam.sdk.coders.CoderRegistry;
+import org.apache.beam.sdk.coders.NullableCoder;
+import org.apache.kafka.common.serialization.Deserializer;
+
+class LocalDeserializerProvider<T> implements DeserializerProvider<T> {
```
aromanenko-dev: Please, add Javadoc for this class

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/LocalDeserializerProvider.java
```diff
@@ -0,0 +1,88 @@
...
+    } catch (InstantiationException
+        | IllegalAccessException
+        | InvocationTargetException
+        | NoSuchMethodException e) {
+      throw new RuntimeException("Could not instantiate deserializers", e);
+    }
+  }
+
+  @Override
+  public NullableCoder<T> getCoder(CoderRegistry coderRegistry) {
```
aromanenko-dev: Please, add a Javadoc from "old" `inferCoder()` method here.

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
```diff
@@ -588,7 +584,10 @@ public void setValueDeserializer(String valueDeserializer) {
      * provide the key coder explicitly.
      */
     public Read<K, V> withKeyDeserializer(Class<? extends Deserializer<K>> keyDeserializer) {
-      return toBuilder().setKeyDeserializer(keyDeserializer).build();
+      return toBuilder()
+          .setKeyDeserializer(keyDeserializer)
```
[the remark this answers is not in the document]
iemejia: I preferred to let it as it was because existing users can rely on `getDeserializer` which would be broken if so. I think we can do this clean up as a follow up after this PR, with the respective deprecation in place.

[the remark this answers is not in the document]
iemejia: This was to save the state of the object in case someone call the `getDeserializer` method, but it is now inconsistent so probably better to remove it.

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
iemejia: I will remove it as mentioned above.

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/LocalDeserializerProvider.java
iemejia: done

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/LocalDeserializerProvider.java
iemejia: done

sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
iemejia: Let's remove completely the method signature and let users use the `DeserializerProvider` verion that way they will put explicitly the Confluent variant with their arguments.

iemejia:
Run Java KafkaIO Performance Test

iemejia:
Fixes done PTAL @aromanenko-dev

aromanenko-dev:
Run Java PreCommit

aromanenko-dev:
Run Java KafkaIO Performance Test

iemejia:
Thanks !

PR #10959, opened 2020-02-25
[BEAM-9247] Integrate GCP Vision API functionality

EDjur:
This PR refers to https://issues.apache.org/jira/browse/BEAM-9247: [Python] PTransform that integrates Cloud Vision functionality.

Small comments:
The synchronous annotation is very similar to the videointelligence implementation.

I opted to only support async (offline) batch annotation:
* Synchronous batch annotation only supports batching <=5 elements at a time, so is not very useful (https://cloud.google.com/vision/docs/batch).
* Synchronous batch annotation also doesn't make a lot of sense for Beam, as the CPU/workers would just sit around and wait until a response is received from the Vision API, which seems inefficient.

I'm still working on a few kinks and issues with the batch annotation as well as making sure all tests run properly, so consider this a draft PR for now.

------------------------

--- | --- | --- | --- | ---


EDjur:
R: @aaltay @kamilwu

robertwb:
Has there been a discussion of whether this (rather specialized) kind of transform belongs in beam vs. as a separate library built on top of beam? 

Also, is the synchronous/batching pattern something that would be generally useful, and could be pulled out, or is it really integral to these transforms and contexts?

EDjur:
Most of the discussion was done in this PR: https://github.com/apache/beam/pull/10764 and the over-arching Jira ticket is here https://issues.apache.org/jira/browse/BEAM-9145

I agree that these transforms are slightly different as they call an external API and lets that do most of the heavy-lifting processing.

sdks/python/apache_beam/ml/gcp/visionml.py
```diff
@@ -0,0 +1,526 @@
...
+
+  def _flush_batch(self):
+    operation = self._client.async_batch_annotate_images(
+        requests=self._batch_elements,
+        output_config=self.output_config,
+        retry=self.retry,
+        timeout=self.timeout,
+        metadata=self.metadata)
+    self._batch_elements = []
+    return operation
```
EDjur: On closer examination, it may be that we need to wait for the return result of the operation anyway, like so: `return operation.results()` as without it, the function returns a `google.api_core.operation.Operation object` and then Beam runs into the following error:
```
AttributeError: Can't pickle local object 'if_exception_type.<locals>.if_exception_type_predicate' [while running 'Annotate image/ParDo(_AsyncBatchImageAnnotateFn)']
```

EDjur: That is, if this PTransform should return anything at all.

Perhaps it makes more sense to not return anything and just send off the Vision API call. The "downside" is that this PTransform (AsyncBatchAnnotateImage) would essentially become a sink).

kamilwu:
> Has there been a discussion of whether this (rather specialized) kind of transform belongs in beam vs. as a separate library built on top of beam?

There's been a discussion on devlist some time ago: [here](https://lists.apache.org/thread.html/rdc9647d00c4bb173f00ca433a2cdea3efe176aea19e631ab9723a80c%40%3Cdev.beam.apache.org%3E). General conclusion was that it's fine to put Cloud AI transforms directly into Beam. Also, we have similar transforms already implemented: https://github.com/apache/beam/tree/master/sdks/python/apache_beam/ml/gcp

kamilwu:
I wonder whether it makes sense to support async (offline) annotation from Beam's perspective. Let's suppose that we don't return anything and `AsyncBatchAnnotateImage` essentially become a sink. In that case, Beam, as a data processing framework, doesn't provide much value. If all the transform does is just sending a request, there is no point in executing it on multiple Dataflow or Flink workers. We'd rather use a task orchestration tool, like Apache Airflow.

On the other hand, the main advantage of sync (online) annotation is that it returns results relatively fast for further processing.

kamilwu:
> Synchronous batch annotation only supports batching <=5 elements at a time, so is not very useful

From my perspective, there should be no problem in sending a request containing up to 5 files to be annotated, then waiting for the result and sending another request.

sdks/python/apache_beam/ml/gcp/visionml.py
kamilwu: I've answered to this in the main thread.

EDjur:
> I wonder whether it makes sense to support async (offline) annotation from Beam's perspective. Let's suppose that we don't return anything and `AsyncBatchAnnotateImage` essentially become a sink. In that case, Beam, as a data processing framework, doesn't provide much value. If all the transform does is just sending a request, there is no point in executing it on multiple Dataflow or Flink workers. We'd rather use a task orchestration tool, like Apache Airflow.
> 
> On the other hand, the main advantage of sync (online) annotation is that it returns results relatively fast for further processing.

I agree that those sort of tasks are more suited for e.g. Airflow, and that was something that crossed my mind too. 

If you think that batches of <=5 items are useful,  then perhaps we should go with sync (online) batch annotation rather than async.

kamilwu:
> If you think that batches of <=5 items are useful, then perhaps we should go with sync (online) batch annotation rather than async.

I don't see any obstacles.

Also take a look at `BatchElements` built-in PTransform (to be found at `sdks/python/apache_beam/transforms/util.py`), which could be useful in our case. This transform consumes a PCollection of element type T and produces a PCollection of type List[T]. The maximum size of the list can be configured by setting a `max_batch_size` parameter. This would speed things up a bit: it's much better to send one request with 5 files than 5 separate requests.

EDjur:
Cheers for the tip about `BatchElements`. Using that reduces code duplication quite a bit as we can offload the creation of the `AnnotateImageRequest` to an earlier step in the `PTransform`, leaving us with only one `DoFn` for the two Batch transforms.

kamilwu:
Cool, nice to hear that!

One more thing that came to my mind.
Would it be possible to merge `AnnotateImage` and `BatchAnnotateImage` transforms? I think the user has little interest in configuring `min_batch_size` and `max_batch_size`, because, at the end, responses are flatten (output PCollection is of type `AnnotateImageResponse`, not `List[AnnotateImageResponse]`). Also, I expect that `batch_annotate_images` with `len(requests) == 1` works the same way as `annotate_images`. But maybe I'm wrong.

aaltay:
Overall direction LGTM. One comment on the setup.py

1. Do you need a lower limit?
2. <0.43.0 ? if we dont want to automatically pick up the next version.

EDjur:
> Overall direction LGTM. One comment on the setup.py
> 
> 1. Do you need a lower limit?
> 2. <0.43.0 ? if we dont want to automatically pick up the next version.

Yes, sorry, I'm meaning to test out a few different versions below 0.43.0 before finalising the PR. But wanted to make sure that the PR was heading in the correct general direction before doing that.

EDjur:
> Cool, nice to hear that!
> 
> One more thing that came to my mind.
> Would it be possible to merge `AnnotateImage` and `BatchAnnotateImage` transforms? I think the user has little interest in configuring `min_batch_size` and `max_batch_size`, because, at the end, responses are flatten (output PCollection is of type `AnnotateImageResponse`, not `List[AnnotateImageResponse]`). Also, I expect that `batch_annotate_images` with `len(requests) == 1` works the same way as `annotate_images`. But maybe I'm wrong.

Indeed, looking at the source code for `annotate_image`, it is essentially just a wrapper around `batch_annotate_images` like so `r = self.batch_annotate_images([request], retry=retry, timeout=timeout)`.

I'll need to keep the min_batch_size and max_batch_size parameters for testing purposes, in order to get a valid count of the API calls as otherwise the number of calls is dependent on the BatchElements transform. But I'll make sure to document these two parameters as such.

EDjur:
@aaltay I noticed an issue in `setup.py` for the videointelligence library: it was missing a comma. I added that fix in this PR if that's okay.

I also updated `CHANGES.md` to specify that the videointelligence features are for the Python SDK.

sdks/python/apache_beam/ml/gcp/visionml_test.py
```diff
@@ -0,0 +1,252 @@
...
+except ImportError:
+  ImageAnnotatorClient = None
+
+
+@unittest.skipIf(
+    ImageAnnotatorClient is None, 'Vision dependencies are not installed')
+class VisionTest(unittest.TestCase):
+  def setUp(self):
+    self._mock_client = mock.Mock()
+    self.m2 = mock.Mock()
```
kamilwu: This mock seems to be unused

sdks/python/apache_beam/ml/gcp/visionml_test.py
```diff
@@ -0,0 +1,252 @@
...
+              min_batch_size=self.min_batch_size,
+              max_batch_size=self.max_batch_size))
+      result = p.run()
+      result.wait_until_finish()
+
+      read_filter = MetricsFilter().with_name('API Calls')
+      query_result = result.metrics().query(read_filter)
+      if query_result['counters']:
+        read_counter = query_result['counters'][0]
+        self.assertTrue(read_counter.committed == expected_counter)
```
kamilwu: It's safer to use `result` property rather than `committed`, because when `committed` is empty, `attempted` is being used. See the code:
https://github.com/apache/beam/blob/f9f3159ef6da97b41227d5fa52ac1c176179eb09/sdks/python/apache_beam/metrics/execution.py#L129-L134

kamilwu:
LGTM. I've left two small comments, but consider them as non-blockers.

Thanks for merging those PTransform I mentioned and for documenting batch_size parameters.

EDjur:
Thanks for the thorough review!

mwalenia:
retest this please

mwalenia:
retest this please

PR #10970, opened 2020-02-25
[BEAM-9342] Exclude module-info.class from vendored Byte Buddy 1.10.8

iemejia:
R: @lukecwik

### DIFF
--- a/vendor/bytebuddy-1_10_8/build.gradle
+++ b/vendor/bytebuddy-1_10_8/build.gradle
@@ -28,6 +28,9 @@ vendorJava(
   relocations: [
     "net.bytebuddy": "org.apache.beam.vendor.bytebuddy.v1_10_8.net.bytebuddy"
   ],
+  exclusions: [
+    "META-INF/versions/9/module-info.class"
+  ],
   groupId: group,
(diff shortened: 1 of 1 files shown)


vendor/bytebuddy-1_10_8/build.gradle
```diff
@@ -28,6 +28,9 @@ vendorJava(
   relocations: [
     "net.bytebuddy": "org.apache.beam.vendor.bytebuddy.v1_10_8.net.bytebuddy"
   ],
+  exclusions: [
+    "META-INF/versions/9/module-info.class"
```
lukecwik: We should exclude it regardless where it appears. This also matches how its being excluded in vendored gRPC.
```suggestion
    "module-info.class"
```

iemejia: will do, I tried the version you proposed and it did not match it, but `**/module-info.class` does.

iemejia:
Merged manually thanks @lukecwik and @vectorijk .
Filled https://issues.apache.org/jira/browse/BEAM-9385 to tackle the verification issue. I will work on that.

PR #10974, opened 2020-02-26
[BEAM-9384] Add SchemaRegistry.getSchemaCoder to get SchemaCoders for registered types

iemejia:
R: @aromanenko-dev @reuvenlax

### DIFF
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/SchemaCoder.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/SchemaCoder.java
@@ -22,6 +22,7 @@ import static org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Prec
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.util.Map;
 import java.util.Objects;
 import java.util.UUID;
 import javax.annotation.Nullable;
@@ -59,7 +60,7 @@ import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.Immutabl
 @Experimental(Kind.SCHEMAS)
 public class SchemaCoder<T> extends CustomCoder<T> {
   // This contains a map of primitive types to their coders.
-  public static final ImmutableMap<TypeName, Coder> CODER_MAP =
+  public static final Map<TypeName, Coder> CODER_MAP =
       ImmutableMap.<TypeName, Coder>builder()
           .put(TypeName.BYTE, ByteCoder.of())
           .put(TypeName.BYTES, ByteArrayCoder.of())
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/SchemaRegistry.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/SchemaRegistry.java
@@ -226,27 +226,16 @@ public class SchemaRegistry {
   }
 
   /**
-   * Get a schema for a given {@link Class} type. If no schema exists, throws {@link
+   * Retrieve a {@link Schema} for a given {@link Class} type. If no schema exists, throws {@link
    * NoSuchSchemaException}.
    */
   public <T> Schema getSchema(Class<T> clazz) throws NoSuchSchemaException {
     return getSchema(TypeDescriptor.of(clazz));
   }
 
-  private <ReturnT> ReturnT getProviderResult(Function<SchemaProvider, ReturnT> f)
-      throws NoSuchSchemaException {
-    for (SchemaProvider provider : providers) {
-      ReturnT result = f.apply(provider);
-      if (result != null) {
-        return result;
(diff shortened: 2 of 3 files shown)


iemejia:
Run Java PreCommit

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/SchemaRegistry.java
```diff
@@ -288,6 +281,38 @@ public void registerSchemaProvider(SchemaProvider schemaProvider) {
...
+   * Retrieve a {@link SchemaCoder} for a given {@link TypeDescriptor} type. If no schema exists,
+   * throws {@link * NoSuchSchemaException}.
+   */
+  public <T> SchemaCoder<T> getSchemaCoder(TypeDescriptor<T> typeDescriptor)
+      throws NoSuchSchemaException {
+    return SchemaCoder.of(
+        getSchema(typeDescriptor),
+        typeDescriptor,
+        getToRowFunction(typeDescriptor),
+        getFromRowFunction(typeDescriptor));
```
reuvenlax: This looks fine, but can you explain the use case? The goal was for users to never have to deal with SchemaCoder (the fact that schemas are implemented via a special coder should be a Beam implementation details), but I'd understand if we have cases where the coder is still needed.

iemejia: See #10978 for a concrete case. PTransform authors may benefit of being able to infer a `SchemaCoder` for a given type from the `SchemaRegistry`.
We can even add that too to PubsubIO to read JavaBeans and be able to query them with Beam's SQL.

iemejia: I agree with you in the fact that `SchemaCoder` is an internal detail that regular users (authors of Pipelines) should not care about. I felt tempted to mark this method as `@Internal` however PTransform authors (e.g. IO authors) will find this useful (as I did for the PR I mention above for KafkaIO schema support), so probably worth to let it available, also I cannot think of a better place to put this method than here.

iemejia:
Run Java PreCommit

iemejia:

mwalenia:
Run seed job

mwalenia:
retest this please

mwalenia:
I thought retesting will clear the nonexistent job :(

iemejia:
@mwalenia I just rebased and push forced to see the unavailable Java11 job dissapear, thanks for the tip!

iemejia:
Run Java PreCommit

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/SchemaCoder.java
```diff
@@ -59,7 +60,7 @@
 @Experimental(Kind.SCHEMAS)
 public class SchemaCoder<T> extends CustomCoder<T> {
   // This contains a map of primitive types to their coders.
-  public static final ImmutableMap<TypeName, Coder> CODER_MAP =
+  public static final Map<TypeName, Coder> CODER_MAP =
```
iemejia: This change is required because we should not be exposing Guava (even our own vendored version) in the public API and it was surfaced by `GcpCoreApiSurfaceTest` break

iemejia:
retest this please

iemejia:
Run Java PreCommit

iemejia:
Run Java PreCommit

iemejia:
Run Java PreCommit

iemejia:
Finally the tests are green. Can you PTAL @reuvenlax .

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/SchemaRegistry.java
reuvenlax: It's useful when integrating schema code with code that does not yet understand schemas.

In the KafkaIO example I think that the ideal solution would be to allow a Schema on KafkaRecord (this probably requires us to add Java generic type awareness to schema inference though), in which case they keyCoder and valueCoder isn't needed. I agree that allowing easy inference of SchemaCoder allows for lower-effort integration of schemas in code like this, though hopefully this is just a short-term solution.

reuvenlax:
lgtm

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/SchemaRegistry.java
iemejia: Interesting. I have not thought about making KafkaRecord 'schema' like good point. There are some consequences on that that are still not clear to me (like how will we deal with the runtime resolution part of Schemas for KV that we do now with the Confluent Schema Registry support. I am going to give it a try and ping you once I have something in the other PR #10978. Let's continue that discussion there.

PR #10966, opened 2020-02-25
[WIP][BEAM-7336] Add schema inferring for KafkaIO when reading Avro values

aromanenko-dev:
Adds `withAvroSchemaValues()` to read `PCollection<GenericRecord>` from Kafka topics where values are serialised by Avro schemas . Also, it sets Beam schema inferred on Avro schema used for values serialisation, so, it can be used as an input for `SqlTransform` downstream of pipeline. It has to be used with `ConfluentSchemaRegistryDeserializerProvider`.

------------------------

Thank you for your contribution! Follow this checklist to help us incorporate your contribution quickly and easily:

 - [x] [**Choose reviewer(s)**](https://beam.apache.org/contribute/#make-your-change) and mention them in a comment (`R: @username`).
 - [x] Format the pull request title like `[BEAM-XXX] Fixes bug in ApproximateQuantiles`, where you replace `BEAM-XXX` with the appropriate JIRA issue, if applicable. This will automatically link the pull request to the issue.
 - [ ] Update `CHANGES.md` with noteworthy changes.
 - [ ] If this contribution is large, please file an Apache [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

See the [Contributor Guide](https://beam.apache.org/contribute) for more tips on [how to make review process smoother](https://beam.apache.org/contribute/#make-reviewers-job-easier).

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

Lang | SDK | Apex | Dataflow | Flink | Gearpump | Samza | Spark
--- | --- | --- | --- | --- | --- | --- | ---
Go | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go/lastCompletedBuild/) | --- | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_VR_Flink/lastCompletedBuild/) | --- | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_VR_Spark/lastCompletedBuild/)
[...]

PR #10978, opened 2020-02-26
[WIP][BEAM-7336] Add Beam schema inferring for KafkaIO

iemejia:
As discussed offline, here it is a version of my proposed changes to #10966 
This one assumes that the user will extract the Keys and Values as it already the case in KafkaIO and allows to configure Beam Schema inference separately.
A nice improvement is the resolution using `SchemaRegistry` but I found some points there that probably should be addressed in a different issue.

R: @aromanenko-dev @reuvenlax

### DIFF
--- a/sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
+++ b/sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
@@ -50,6 +50,10 @@ import org.apache.beam.sdk.io.UnboundedSource;
 import org.apache.beam.sdk.io.UnboundedSource.CheckpointMark;
 import org.apache.beam.sdk.options.PipelineOptions;
 import org.apache.beam.sdk.options.ValueProvider;
+import org.apache.beam.sdk.schemas.NoSuchSchemaException;
+import org.apache.beam.sdk.schemas.SchemaCoder;
+import org.apache.beam.sdk.schemas.SchemaRegistry;
+import org.apache.beam.sdk.schemas.utils.AvroUtils;
 import org.apache.beam.sdk.transforms.DoFn;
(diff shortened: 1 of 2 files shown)


sdks/java/io/kafka/src/main/java/org/apache/beam/sdk/io/kafka/KafkaIO.java
```diff
@@ -913,16 +985,55 @@ public void setValueDeserializer(String valueDeserializer) {
...
+              : getValueDeserializerProvider().getCoder(coderRegistry);
+      if (isValueInferBeamSchema()) {
+        return inferSchemaCoder(valueCoder, schemaRegistry);
+      }
+      return valueCoder;
+    }
+
+    @VisibleForTesting
+    static <T> SchemaCoder<T> inferSchemaCoder(Coder<T> coder, SchemaRegistry schemaRegistry) {
+      // At the moment there is not a general way to infer SchemaCoders from non-schema Coders (that
```
iemejia: @reuvenlax I tried to get the Avro resolution by registering `AvroRecordSchema` as a a new `SchemaRegistryProvider` but this produced a lot of issues. 

I have the impression we need to do something like what I mention to register which basic Coders we can support also as SchemaCoders e.g. Avro, Protobuf and others that contain a `Schema` and the `from/toRowFunction`s

iemejia:
Run Java PreCommit

aromanenko-dev:
Run Spotless PreCommit

aromanenko-dev:
@reuvenlax Could you take a look on this?

pabloem:
hi Alexey. I'm sorry this was left to languish... is this something you think can be used?

aromanenko-dev:
@pabloem I'm pretty sure that this feature is still worth to have and these changes seems feasible at the first glance but I don't remember why it was not merged. 
@iemejia Do you recall why it was closed?

PR #10983, opened 2020-02-27
[BEAM-9393] Support schemas in state API

reuvenlax:
Add schema inference for types used in the state API.

Add state overrides for Row types.

Disable Coder inference for Row types, as we should only use Row with schemas.

R: @dpmills


reuvenlax:
@dpmills any comments on this PR?

dpmills:
LGTM

PR #10988, opened 2020-02-27
[BEAM-9382] Clean up of TestStreamTranscriptTests.

acrites:
Removes check for final field in PaneInfo since it is not reported properly by direct runner. Updates index for session panes and changes test with Count(2) early trigger to only inject two elements since behavior with 3 is non-deterministic.

Added changes to trigger driver to delete pane info index and nonspeculative index when merging since otherwise the newly created window would inherit it from the merged away window resulting in incorrect values.

R: @robertwb 

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/testing/data/trigger_transcripts.yaml
+++ b/sdks/python/apache_beam/testing/data/trigger_transcripts.yaml
@@ -53,7 +53,7 @@ transcript:
   - input: [1, 2, 10, 11, 80, 81]
   - watermark: 100
   - expect:
-      - {window: [0, 9], values: [1, 2], timestamp: 9, final: false}
+      - {window: [0, 9], values: [1, 2], timestamp: 9}
       - {window: [10, 19], values: [10, 11], timestamp: 19}
       - {window: [80, 89], values: [80, 81], timestamp: 89, late: false}
   - input: [7, 8] # no output
@@ -71,7 +71,7 @@ transcript:
   - input: [1, 2, 3, 10, 11, 25]
   - watermark: 100
   - expect:
-      - {window: [0, 9], values: [1, 2, 3], timestamp: 1, final: false}
+      - {window: [0, 9], values: [1, 2, 3], timestamp: 1}
       - {window: [10, 19], values: [10, 11], timestamp: 10}
       - {window: [20, 29], values: [25], timestamp: 25, late: false}
 
@@ -84,7 +84,7 @@ transcript:
   - input: [1, 2, 3, 10, 11, 25]
   - watermark: 100
   - expect:
-      - {window: [0, 9], values: [1, 2, 3], timestamp: 3, final: false}
+      - {window: [0, 9], values: [1, 2, 3], timestamp: 3}
       - {window: [10, 19], values: [10, 11], timestamp: 11}
       - {window: [20, 29], values: [25], timestamp: 25, late: false}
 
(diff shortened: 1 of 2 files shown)


lukecwik:
Drive by comment:
Shouldn't we disable the test for the direct runner if it isn't supported and file a bug instead of editing what the test does?
Alternatively, could we fix the direct runner?

kennknowles:
Since we are doing drive by comments, perhaps a clear comment describing exactly what the test is trying to verify would allow us to be sure to have all the necessary verifications but no extraneous info.

Also I assume you mean the Python direct runner (each language has a direct runner for that language) since I think the Java direct runner does support pane info.

acrites:
Along the lines of what Kenn is saying, I had originally thought that these tests weren't really testing whether or not PaneInfo.final gets set correctly in the various triggering strategies, but more so that it was outputting the correct elements. Maybe we want to test all aspects in a single test though.

I'm fine with sickbaying the tests for Python direct runner if you're ok with having a large number of tests inactive until we getting around to fixing direct runner. Maybe that'll motivate us to get it fixed faster...

robertwb:
The Python direct runner does support PaneInfo, but it doesn't (yet) support allowed lateness, so the "final" bit is wrong. Yes, we should fix this, though hopefully on the FnApiRunner once it supports streaming rather than invest too much into the old direct runner (unless it's cheap to do). 

I would rather disable the "final" bit check for these tests (and add a new, disabled test that checks this bit) than disable these tests entirely for the direct runner. So as is, this change LGTM.

We could also consider expanding the API to indicate certain features are broken on certain runners, rather than entire tests, which would allow us to fully specify the expected results and still produce partial validation on incomplete/in-progress runners.

acrites:
RE fixing the Python direct runner: right now we just use the return value
from on_fire for each trigger to set the final bit. This doesn't take into
account allowed lateness. So we'd probably need to update all of the
triggers.

Maybe just a single test for now that is disabled and has the final bit
check? Or did you want to duplicate all the tests? I'm thinking something
like this one:

name: fixed_allowed_lateness_final_set
broken_on:
  - SwitchingDirectRunner  # Doesn't set final field properly with late
data.
window_fn: FixedWindows(10)
trigger_fn: AfterWatermark(early=AfterCount(2), late=AfterCount(1))
timestamp_combiner: OUTPUT_AT_EOW
allowed_lateness: 10
transcript:
  - input: [1, 2]
  - expect:
      - {window: [0, 9], values: [1, 2], timestamp: 9, final: false}
  - input: [3]
  - watermark: 100
  - expect:
      - {window: [0, 9], values: [1, 2, 3], timestamp: 9, final: true}

acrites:
Actually, that last example doesn't fail like I would expect. So I guess the Python direct runner sets final = true sometimes...

acrites:
Run Python PreCommit

acrites:
It looks like TriggerDriverTranscriptTest.test_early_late_sessions is failing (since there's no way to sickbay that test). Am I understanding the semantics of PaneInfo index correctly? The first firing of each window should have index 0 even if that window was the result of merging another window, right?

kennknowles:
Yea, that's right.

acrites:
Run Python PreCommit

acrites:
Run Python PreCommit

lukecwik:
retest this please

robertwb:
I'm not convinced that reseting the indices to 0 on merging windows is the desired semantics, but that discussion is probably out of scope for this PR. Other than that, LGTM.

kennknowles:
Yea I don't think anyone is happy about it. But it has always been that way and also is the only way we meet the spec of panes being (EARLY* ON_TIME? LATE*)

acrites:
Apparently the failing test in the PreCommit has been "Failing for the past 11,721 builds (Since #0 )"

PR #10990, opened 2020-02-27
[BEAM-9569] disable coder inference for rows

reuvenlax:


sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamSetOperatorRelBase.java
```diff
@@ -65,6 +66,16 @@ public BeamSetOperatorRelBase(BeamRelNode beamRelNode, OpType opType, boolean al
         inputs);
     PCollection<Row> leftRows = inputs.get(0);
     PCollection<Row> rightRows = inputs.get(1);
+    Schema leftSchema = leftRows.getSchema();
+    Schema rightSchema = rightRows.getSchema();
+    if (!leftSchema.typesEqual(rightSchema)) {
+      throw new IllegalArgumentException(
+          "Can't intersect two tables with different schemas."
```
kennknowles: This is the base class, so shouldn't be calling it "intersect". Incidentally I don't think having a base class adds much value here, so inlining or inverting would be a-ok.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/transform/BeamSetOperatorsTransforms.java
```diff
@@ -38,61 +35,62 @@
...
-            // output both left & right
-            Iterator<Row> iter = leftRows.iterator();
-            while (iter.hasNext()) {
-              ctx.output(iter.next());
-            }
-            iter = rightRows.iterator();
-            while (iter.hasNext()) {
-              ctx.output(iter.next());
+            for (int i = 0; i < numLeftRows + numRightRows; i++) {
+              o.output(key);
```
kennknowles: This optimization seems to have nothing to do with disabling coder inference. Please split into separate commit so git history shows it.

reuvenlax:
@kennknowles I've removed the premature optimization. I decided to leave the base class for now - would prefer to leave that as a refactoring cleanup for another PR.

Note: support for nullable fields in this PR (as this PR exposed bugs in the Schema transforms) is also pulled out into pr/11046. I plan on merging 11046 first and rebasing this one on top, so those changes will not be included in this merge.

reuvenlax:
@kennknowles the Select changes have now been rebased away

reuvenlax:
R: @dpmills

reuvenlax:
Run Java PreCommit

dpmills:
LGTM

PR #11011, opened 2020-03-02
[website] Update link to environment_type (SDK harness configuration)

iemejia:
R: @ibzib

### DIFF
--- a/sdks/python/apache_beam/io/mongodbio.py
+++ b/sdks/python/apache_beam/io/mongodbio.py
@@ -92,7 +92,7 @@ __all__ = ['ReadFromMongoDB', 'WriteToMongoDB']
 
 @experimental()
 class ReadFromMongoDB(PTransform):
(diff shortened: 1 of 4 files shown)


website/src/roadmap/portability.md
```diff
@@ -165,7 +165,3 @@ Please see the [Spark Runner page]({{ site.baseurl }}/documentation/runners/spar
 how to run portable pipelines on top of Spark.
 
 Python streaming mode is not yet supported on Spark.
-
-## SDK Harness Configuration {#sdk-harness-config}
```
ibzib: Can we leave this?
a) Links to this header could be left floating around out there in my emails etc.
b) I also think it's useful to link these two pages.

iemejia: Yes makes sense, will rollback that change.

iemejia:
Suggested fixes done PTAL @ibzib

iemejia:
Thanks Kyle!

PR #11025, opened 2020-03-03
[BEAM-6428] Improve select performance with codegen

reuvenlax:
The schema select functionality is currently very inefficient - it walks through the schema resolution on each element selected in order to figure out how to properly distribute the select. Select is used by the select transform, but it is also used by the Group, Join, and any conversion transforms.

This PR analyzes the select predicate (e.g. a.b[].c.*), and generates bytecode to perform the select. All the schema resolution code that previously was done per element is now done once in order to determine what bytecode to generate.


reuvenlax:
run dataflow validatesrunner

reuvenlax:
run sql postcommit

reuvenlax:
run dataflow validatesrunner

reuvenlax:
run sql postcommit

reuvenlax:
run dataflow validatesrunner

reuvenlax:
run sql postcommit

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/transforms/Select.java
```diff
@@ -28,8 +28,8 @@
 import org.apache.beam.sdk.schemas.FieldAccessDescriptor;
 import org.apache.beam.sdk.schemas.Schema;
 import org.apache.beam.sdk.schemas.utils.RowSelector;
-import org.apache.beam.sdk.schemas.utils.SelectByteBuddyHelpers;
```
alexvanboxel: I was wondering if SelectByteBuddyHelpers is still used after this removal (and would be able to be deleted). But I don't find it (org.apache.beam.sdk.schemas.utils.SelectByteBuddyHelpers) in the current master.

reuvenlax: This PR introduces SelectByteBuddyHelpers.

reuvenlax: I think you are looking at an individual commit.

reuvenlax:
Run Java PreCommit

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/transforms/Select.java
alexvanboxel: oeps my bad. You are right (early morning). Resoving, looking further.

PR #11034, opened 2020-03-03
[BEAM-9424] Allow grouping by LogicalType

kanterov:
Add LogicalTypeCoder.

Support LogicalTypes in Group transform and SchemaCoder generation.

Related to: https://github.com/apache/beam/pull/11015

------------------------

--- | --- | --- | --- | ---


kanterov:
@reuvenlax it looks like there is a couple of places where LogicalType don't work since we introduced to/fromBaseType functions. Looking into the code we would need to support it in RowCoderGenerator that doesn't look straightforward because LogicalType doesn't have a static factory we can call.

Do you have thoughts on how we can fix that?

sdks/java/core/src/main/java/org/apache/beam/sdk/coders/RowCoderGenerator.java
```diff
@@ -401,6 +404,20 @@ private static StackManipulation getCoder(Schema.FieldType fieldType) {
     }
   }
 
+  private static StackManipulation logicalTypeCoder(
+      Schema.LogicalType logicalType, StackManipulation baseCoder) {
+    throw new UnsupportedOperationException("not implemented");
```
kanterov: This section is left unimplemented, and that's why the test is failing

reuvenlax:
I'm not sure I understand the problem or the solution. Why don't logical types work in Group? Why does creating a new LogicalTypeCoder solve this?

kanterov:
@reuvenlax because `RowCoderGenerator` uses `Coder<BaseT>`, not `Coder<InputT>`. We know how to produce `Coder<BaseT>`, and we can make `Coder<InputT>` out of it using `LogicalType`. So the test fails because Row builder expected `EnumerationType.Value` but got `String`. What is special about Group is that it constructs

reuvenlax:
The whole point of Logical types is that the representation should be that of the base type (the schema returned by getBaseType). This means that the logical type can be processed by code that doesn't know about InputT, and also makes it possible to use it cross language. 

Where in Group do we need Coder<InputT>? I feel like I'm still missing something.

kanterov:
@reuvenlax I cleaned the code and left only test that reproduces this issue.

The reason for that is because it uses `RowCoderGenerator` that would use a `Coder<BaseT>` instead of `Coder<InputT>` (see `RowCoderGenerator.java:374`). While it doesn't make any difference for binary representation, SDK code fails because of that. In the stack trace we can see how Combine should get `InputT=EnumerationType$Value`, but it gets `BaseT=Integer`.

```
Caused by: java.lang.ClassCastException: java.lang.Integer cannot be cast to org.apache.beam.sdk.schemas.logicaltypes.EnumerationType$Value
	at org.apache.beam.sdk.schemas.logicaltypes.EnumerationType.toBaseType(EnumerationType.java:37)
	at org.apache.beam.sdk.values.Row$Builder.verifyLogicalType(Row.java:659)
	at org.apache.beam.sdk.values.Row$Builder.verify(Row.java:652)
	at org.apache.beam.sdk.values.Row$Builder.verify(Row.java:635)
	at org.apache.beam.sdk.values.Row$Builder.build(Row.java:840)
	at org.apache.beam.sdk.schemas.utils.SelectHelpers.selectRow(SelectHelpers.java:194)
	at org.apache.beam.sdk.schemas.transforms.SchemaAggregateFn$Inner$ExtractSingleFieldFunction.apply(SchemaAggregateFn.java:241)
	at org.apache.beam.sdk.schemas.transforms.SchemaAggregateFn$Inner$ExtractSingleFieldFunction.apply(SchemaAggregateFn.java:231)
	at org.apache.beam.sdk.transforms.CombineFns$ComposedCombineFn.addInput(CombineFns.java:377)
	at org.apache.beam.sdk.schemas.transforms.SchemaAggregateFn$Inner.addInput(SchemaAggregateFn.java:283)
	at org.apache.beam.sdk.schemas.transforms.SchemaAggregateFn$Inner.addInput(SchemaAggregateFn.java:55)
	at org.apache.beam.sdk.transforms.Combine$CombineFn.apply(Combine.java:446)
	at org.apache.beam.sdk.transforms.Combine$GroupedValues$1.processElement(Combine.java:2171)
```

reuvenlax:
@kanterov ok, I see the bug now. the problem is because we have an unfortunate asymmetry between Row.Builder.addValues and Row.getValues. getValues returns the base type for logical types (because that's how it's stored, but addValues expects to get the input type.

I'm still thinking of the best way to resolve this inconsistency without breaking other use cases. However a simple fix for now is to fix SelectHelpers to use attachValues instead of addValues. Coincidentally that change is already made in https://github.com/apache/beam/pull/11025. Can you check to see whether that PR fixes the problem you're seeing?

reuvenlax:
BTW I think I need to prioritize fixing the inconsistency between getValues and addValues, as this is not the first time it's caused a bug.

kanterov:
@reuvenlax I've rebased on the latest master that includes https://github.com/apache/beam/pull/11025. It didn't seem to help, because the test is failing with the same exception:

https://builds.apache.org/job/beam_PreCommit_Java_Commit/10259/testReport/junit/org.apache.beam.sdk.schemas.transforms/GroupTest/testAggregateLogicalType/

reuvenlax:
@kanterov it's failing differently now - no longer in. SelectHelpers.

reuvenlax:
@kanterov this is because (as I mentioned before) Row.getValue today always returns the base type (in this case int), because that's how we store it. This has the side effect of passing the base value into your combiner, which is currently expecting the value type.

This inconsistency keeps causing strange effects like this. I think we need getValue to return the same types passed into addValues (the value type). I'm trying to think of a good way to do this without breaking other semantics and impacting performance.

kanterov:
@reuvenlax is there an option to have alternative to attachValues, but for getting values to be used in performance-critical code?

reuvenlax:
IT's not just performance critical code. We wanted these transforms to be
usable by code that doesn't understand the logical type.

For example: you should be able to write a SQL expression over a type that
contains a OneOf. SQL doesn't understand the OneOf logical type of course,
however the base type is a Row with nullable fields, Since today getValue
returns the base type, you should be able to query it. Similar story with
Enumerations - you should be able to treat them as integers in a SQL
statement or in one of the schema transforms. (note: I"m not sure if this
has been tested with SQL, so it may not work today).

We need to figure out a good API that allows these transformation to use
the base type by default but _also_ allows you to write a combiner that
accepts the appropriate value type.

Reuven

PR #11040, opened 2020-03-04
[BEAM-9305] Allow value provider query strings in _CustomBigQuerySource

EDjur:
This PR relates to https://issues.apache.org/jira/browse/BEAM-9305 and allows for the usage of ValueProvider query strings in _ReadFromBigQuery.

------------------------

--- | --- | --- | --- | ---


EDjur:
R: @pabloem 

As I have little experience with the BQ module, I'm a bit unsure if the test was placed in the correct module.

EDjur:
Since `estimate_size` is called before `split` and the uses `self.query`, I realised the value provider evaluation actually needs to take place before `split` is called. Does this make sense?

sdks/python/apache_beam/io/gcp/bigquery_read_it_test.py
```diff
@@ -162,6 +163,15 @@ def test_iobase_source(self):
...
       assert_that(result, equal_to(self.TABLE_DATA))
 
+  @attr('IT')
+  def test_valueprovider_query_string(self):
+    query = StaticValueProvider(str, self.query)
+    with beam.Pipeline(argv=self.args) as p:
+      result = (
+          p | 'read' >> beam.io._ReadFromBigQuery(
+              query=query, use_standard_sql=True, project=self.project))
+      assert_that(result, equal_to(self.TABLE_DATA))
```
pabloem: could you use the value provider in another one of the tests in this file? Since each test in this file takes ~5min to run, I'd prefer to avoid adding a new test.

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -644,6 +644,8 @@ def estimate_size(self):
           self.table_reference.tableId)
       return int(table.numBytes)
     else:
+      if isinstance(self.query, ValueProvider):
+        self.query = self.query.get()
```
pabloem: I would probably do this without changing the class variable.

I'd do a local variable: `query = self.query.get()`, and do it on every instance where `self.query` is accessed.

The reason is that the Source objects may be recreated in different workers, and the  `self.query` variable would contain the original ValueProvider in all of those.

EDjur:
How about something like this instead @pabloem ? I noticed that `gcs_location`supports ValueProviders and is wrapped & initialised in the `_ReadFromBigQuery` transform. This reduces clutter in the `_CustomBigQuerySource`.

EDjur:
Adding @kamilwu as you were mentioned in the Jira ticket too.

sdks/python/apache_beam/io/gcp/bigquery_read_it_test.py
```diff
@@ -156,12 +157,19 @@ def test_native_source(self):
...
   @attr('IT')
   def test_iobase_source(self):
+    query = StaticValueProvider(str, self.query)
     with beam.Pipeline(argv=self.args) as p:
       result = (
           p | 'read' >> beam.io._ReadFromBigQuery(
               query=self.query, use_standard_sql=True, project=self.project))
       assert_that(result, equal_to(self.TABLE_DATA))
 
+    with beam.Pipeline(argv=self.args) as p:
```
kamilwu: Do we need an integration test to cover your value provider's logic? Take a look at `TestReadFromBigQuery` test case in `bigquery_test.py`. Those are unit tests for `_ReadFromBigQuery` that actually run very fast, because pipeline is not executed.

Adding new tests in `bigquery_read_it_test.py` causes huge overhead, since they are all executed by three runners (Direct, Dataflow and Flink) on every supported Python version.

EDjur: To actually evaluate the value providers, the pipeline needs to be executed right? Cause these are realistically only 'known' at runtime. The tests in `TestReadFromBigQuery` just validate the input data to the initialisation.

We can of course just validate that query is of type `ValueProvider` after initialising a pipeline, but I figured we want to check that the value provider query is actually evaluated to a string type on pipeline execution.

Am I perhaps missing something?

pabloem: What I would say is use the value provider in the first pipeline of this method, and don't add a second pipeline. I think that should be fine. WDYT @kamilwu ?

kamilwu: @EDjur Thanks for explanation. That's right, if we want to check that query is successfully evaluated to a string, we have to execute the pipeline.
@pabloem I think this is good idea. Let's do it this way.

EDjur:
@kamilwu Looks like the `AnnotateText` PTransform mention in `CHANGES.md` was removed in https://github.com/apache/beam/commit/0cc577134b9e98b32020efab2ca07ccd1665f7cb#diff-8b1c3fd0d4a6765c16dfd18509182f9d.

So I added it back again.

pabloem:
I'm srry @EDjur - the `expand` method runs at template construction, so the change currently will not work with parameters provided at execution time. You'd need to call `get` on the valueprovider in estimate_size/split/etc like we had discussed.

EDjur:
The `gcs_location` value provider is evaluated in the `expand` method: https://github.com/apache/beam/pull/11040/files#diff-683ba04a0e750cf0456b9b119393ad64R1587. Is that a bug then?

pabloem:
ah good catch. Yes, that would make it a bug..

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -694,14 +699,16 @@ def read(self, range_tracker):
     raise NotImplementedError('BigQuery source must be split before being read')
 
   def _setup_temporary_dataset(self, bq):
+    query = self.evaluate_query_valueprovider()
```
kamilwu: There's a `check_accessible` decorator that could be used for this function (as well as for `estimate_size`).
I think you could also test if `self.query` is a ValueProvider or not in the constructor. If it's not, you can then create a StaticValueProvider object. `self.evaluate_query_valueprovider()` would be unnecessary.

Here'a en example:
https://github.com/apache/beam/blob/f75838f306fc1d851b0dea11c2cb861b43430568/sdks/python/apache_beam/io/filebasedsource.py#L115-L117

kamilwu:
>  Is that a bug then?

Yes, this might be a bug. I don't know a good solution at the moment though... this is more complicated, because we have to fallback to temp_location from pipeline options if gcs_location is empty

sdks/python/apache_beam/io/gcp/bigquery.py
EDjur: Thanks for the tip re: `check_accessible`, I didn't know about that. I've updated the PR based on these suggestions.

kamilwu:
@EDjur Could you create a JIRA ticket regarding `gcs_location` value provider evaluation and assign it to me afterwards? Thanks.

EDjur:
Sure, JIRA ticket is at https://issues.apache.org/jira/browse/BEAM-9506

kamilwu:
LGTM, let's wait for @pabloem's final review as well

pabloem:
Thanks @EDjur for the contribution! Thanks @kamilwu for reviewing

TheNeuralBit:
It looks like this change may have broken `apache_beam.io.gcp.bigquery_io_read_it_test.BigqueryIOReadIT.test_bigquery_read_1M_python` in Python postcommits, e.g. https://builds.apache.org/job/beam_PostCommit_Python2/2036/

Could you take a look @EDjur and/or @kamilwu?

pabloem:
Thanks for catching this @TheNeuralBit . I have https://github.com/apache/beam/pull/11206 out to fix this.

TheNeuralBit:
Thank _you_ for catching it @pabloem :)

PR #11052, opened 2020-03-05
[BEAM-9446] Add missing parallelism and execution mode args.

ibzib:
I'm not thrilled about manually copying these over, later I might look into a long-term solution to this problem. But this works for now.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/options/pipeline_options.py
+++ b/sdks/python/apache_beam/options/pipeline_options.py
@@ -1075,6 +1075,22 @@ class FlinkRunnerOptions(PipelineOptions):
         ' directly, rather than starting up a job server.'
         ' Only applies when flink_master is set to a'
         ' cluster address.  Requires Python 3.6+.')
+    parser.add_argument(
+        '--parallelism',
+        default=-1,
+        type=int,
+        help='The degree of parallelism to be used when distributing '
+             'operations onto workers. If the parallelism is not set, the '
+             'configured Flink default is used, or 1 if none can be found.'
+    )
+    parser.add_argument(
+        '--execution_mode_for_batch',
+        default='PIPELINED',
+        help='Flink mode for data exchange of batch pipelines. '
+             'Reference org.apache.flink.api.common.ExecutionMode. '
+             'Set this to BATCH_FORCED if pipelines get blocked, see '
+             'https://issues.apache.org/jira/browse/FLINK-10672'
+    )
 
 
 class SparkRunnerOptions(PipelineOptions):


angoenka:
If I understand correctly, this is needed because we don't go though job api for submission. 
In job api based submission, runner options are automatically pulled from the runner?

sdks/python/apache_beam/options/pipeline_options.py
```diff
@@ -1075,6 +1075,22 @@ def _add_argparse_args(cls, parser):
...
+        default=-1,
+        type=int,
+        help='The degree of parallelism to be used when distributing '
+             'operations onto workers. If the parallelism is not set, the '
+             'configured Flink default is used, or 1 if none can be found.'
+    )
+    parser.add_argument(
+        '--execution_mode_for_batch',
+        default='PIPELINED',
+        help='Flink mode for data exchange of batch pipelines. '
```
iemejia: (slightly unrelated to the PR comment) Do we have a way to mark pipelineoptions as `@Experimental` in Python? PipelineOptions are critical from the point of view of backwards compatibility, so we should probably be marking non stable options (if we do not).

sdks/python/apache_beam/options/pipeline_options.py
```diff
@@ -1075,6 +1075,22 @@ def _add_argparse_args(cls, parser):
...
+    parser.add_argument(
+        '--parallelism',
+        default=-1,
+        type=int,
+        help='The degree of parallelism to be used when distributing '
+             'operations onto workers. If the parallelism is not set, the '
+             'configured Flink default is used, or 1 if none can be found.'
+    )
+    parser.add_argument(
+        '--execution_mode_for_batch',
```
mxm: I'm not sure we should add these here because that's what we used to do and it was inconsistent and hard to maintain. What we want to do, is to not discard those options but still warn about them not being parsed in the Python SDK. This will allow us to still use them in the Runner code.

robertwb:
I'm OK with this as a short-term fix, but we should still pursue fixing this properly as this is not scalable.

ibzib:
> I'm OK with this as a short-term fix, but we should still pursue fixing this properly as this is not scalable.

@robertwb for a short term fix, I think we can get by with setting these options in the Flink cluster configuration file instead.

mxm:
Have you tried working around this by not discarding these options? AFAIK the json parser is smart enough to read the stringified verison of all option values.

ibzib:
> Have you tried working around this by not discarding these options? AFAIK the json parser is smart enough to read the stringified verison of all option values.

I think this may be the best strategy for the uber jar job server, however I don't think we should change this behavior for other runners. (Not sure if that's what you were proposing, just organizing my thoughts here:)
- In Dataflow, we seem to duplicate every runner option for each SDK, perhaps because there is no better choice due to the runner architecture. In that case, since all the args are presumably known by the SDK, it makes more sense to drop them (status quo) or maybe even error when arguments are unknown, because it usually means the user made a mistake.
- With the "old" Flink job server, retrieving args from the job server is an adequate workaround, so again, there should be no need for unrecognized arguments.

I discussed this with @angoenka today and he suggested that we consider the runner-Flink boundary as well -- i.e., if we should have some way of enabling _all_ Flink environment options to be set through Beam pipeline options instead of just adding the ones we need as we go. This would potentially save users from having to wait for a new release just for us to add a pipeline option that trivially maps 1:1 to Flink (of course, they can always change Flink's conf files, which was going to be my proposed workaround here, but AFAIK that requires a restart of the cluster and would affect all jobs run on the cluster). WDYT?

+cc @tweise

mxm:
What I meant is keeping "unknown" pipeline options instead of discarding them. If we pass the unknown ones as strings I thought this could work. For example, if we pass `"123"` the parser on the Java side is smart enough to figure it is the integer `123`. I haven't tried this out but I'd suggest we do.

Of course this is not very user-friendly. That's why I'd suggest to write pipeline options in a language-agnostic format which can be read by all SDKs.

Considering the Flink pipeline options, what we usually want to set is what is called `ExecutionConfig` in Flink. With Flink they are configured directly in the API. For Beam if the parameters are just strings/ints we could come up with something reflection based to set arbitrary new settings. I agree this would be helpful.

sdks/python/apache_beam/options/pipeline_options.py
tweise: I think we should continue to discard invalid options when the client is aware of the full set of options and can perform the validation - as is the case with portable runner and job server.

ibzib: > I think we should continue to discard invalid options when the client is aware of the full set of options and can perform the validation

Agreed, we would keep options only when we can't be sure (so just the uber jar job server for now).

tweise:
Integer options need to be converted to string and that is done in the portable runner: https://github.com/apache/beam/blob/6fce2528b0ff011ff3d416ac9e70c8d7264b1f3c/sdks/python/apache_beam/runners/portability/portable_runner.py#L162

ibzib:
> Integer options need to be converted to string and that is done in the portable runner:

All the unknown options will be strings already, so that shouldn't be needed.

sdks/python/apache_beam/options/pipeline_options.py
robertwb: While experiments are preferred, I think we can document options such as this as experimental as well.

mxm:
@ibzib Could we try to keep the Runner options in case we are not reading them from the job server? We wouldn't have to manually add the options then.

ibzib:
> @ibzib Could we try to keep the Runner options in case we are not reading them from the job server? We wouldn't have to manually add the options then.

Yeah, that's the plan. Just haven't gotten around to it yet. Looking now.

ibzib:
This PR is obsolete since #11189 is merged.

PR #11060, opened 2020-03-05
[BEAM-9454] Create Deduplication transform based on user timer/state

boyuanzz:
Create a Deduplication PTransform by using user timer/state in python SDK.

R: @lukecwik @robertwb 

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/transforms/deduplicate.py
+++ b/sdks/python/apache_beam/transforms/deduplicate.py
@@ -1,0 +1,134 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# pytype: skip-file
+
+"""a collection of ptransforms for deduplicating elements."""
+
+from __future__ import absolute_import
+from __future__ import division
+
+import typing
+
+from apache_beam import typehints
+from apache_beam.coders.coders import BooleanCoder
+from apache_beam.transforms import core
+from apache_beam.transforms import ptransform
+from apache_beam.transforms import userstate
+from apache_beam.transforms.timeutil import TimeDomain
+from apache_beam.utils import timestamp
+
+__all__ = [
+    'Deduplicate',
+    'DeduplicatePerKey',
+]
+
+K = typing.TypeVar('K')
+V = typing.TypeVar('V')
+
+
+@typehints.with_input_types(typing.Tuple[K, V])
+@typehints.with_output_types(typing.Tuple[K, V])
+class DeduplicatePerKey(ptransform.PTransform):
+  """ A PTransform which deduplicates <key, value> pair over a time domain and
+  threshold. Values in different windows will NOT be considered duplicates of
+  each other. Deduplication is guaranteed with respect of time domain and
+  duration.
+
+  Time durations are required so as to avoid unbounded memory and/or storage
+  requirements within a runner and care might need to be used to ensure that the
+  deduplication time limit is long enough to remove duplicates but short enough
+  to not cause performance problems within a runner. Each runner may provide an
+  optimized implementation of their choice using the deduplication time domain
+  and threshold specified.
+
+  Does not preserve any order the input PCollection might have had.
+  """
+  def __init__(self, processing_time_duration=None, event_time_duration=None):
+    if processing_time_duration is None and event_time_duration is None:
+      raise ValueError(
+          'DeduplicatePerKey requires at lease provide either'
+          'processing_time_duration or event_time_duration.')
+    self.processing_time_duration = processing_time_duration
+    self.event_time_duration = event_time_duration
+
+  def _create_deduplicate_fn(self):
+    processing_timer_spec = userstate.TimerSpec(
+        'processing_timer', TimeDomain.REAL_TIME)
+    event_timer_spec = userstate.TimerSpec('event_timer', TimeDomain.WATERMARK)
+    state_spec = userstate.BagStateSpec('seen', BooleanCoder())
+    processing_time_duration = self.processing_time_duration
+    event_time_duration = self.event_time_duration
+
+    class DeduplicationFn(core.DoFn):
+      def process(
+          self,
+          kv,
+          ts=core.DoFn.TimestampParam,
+          seen_state=core.DoFn.StateParam(state_spec),
+          processing_timer=core.DoFn.TimerParam(processing_timer_spec),
+          event_timer=core.DoFn.TimerParam(event_timer_spec)):
+        if True in seen_state.read():
+          return
+
+        if processing_time_duration is not None:
+          processing_timer.set(
+              timestamp.Timestamp.now() + processing_time_duration)
+        if event_time_duration is not None:
+          event_timer.set(ts + event_time_duration)
+        seen_state.add(True)
+        yield kv
+
+      @userstate.on_timer(processing_timer_spec)
+      def process_processing_timer(
+          self, seen_state=core.DoFn.StateParam(state_spec)):
+        seen_state.clear()
+
+      @userstate.on_timer(event_timer_spec)
+      def process_event_timer(
+          self, seen_state=core.DoFn.StateParam(state_spec)):
+        seen_state.clear()
+
+    return DeduplicationFn()
+
+  def expand(self, pcoll):
+    return (
+        pcoll
+        | 'DeduplicateFn' >> core.ParDo(self._create_deduplicate_fn()))
+
+
+class Deduplicate(ptransform.PTransform):
+  """Similar to DeduplicatePerKey, the Deduplicate transform takes any arbitray
+  value as input and uses value as key to deduplicate among certain amount of
+  time duration.
+  """
+  def __init__(self, processing_time_duration=None, event_time_duration=None):
+    if processing_time_duration is None and event_time_duration is None:
+      raise ValueError(
+          'Deduplicate requires at lease provide either'
+          'processing_time_duration or event_time_duration.')
+    self.processing_time_duration = processing_time_duration
+    self.event_time_duration = event_time_duration
+
+  def expand(self, pcoll):
+    return (
+        pcoll
+        | 'Use Value as Key' >> core.Map(lambda x: (x, None))
+        | 'DeduplicatePerKey' >> DeduplicatePerKey(
+            processing_time_duration=self.processing_time_duration,
+            event_time_duration=self.event_time_duration)
+        | 'Output Value' >> core.Map(lambda kv: kv[0]))
--- a/sdks/python/apache_beam/transforms/deduplicate_test.py
+++ b/sdks/python/apache_beam/transforms/deduplicate_test.py
@@ -1,0 +1,167 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# pytype: skip-file
+
+"""Unit tests for deduplicate transform by using TestStream."""
+
+from __future__ import absolute_import
+
+import unittest
+
+from nose.plugins.attrib import attr
+
+import apache_beam as beam
+from apache_beam.coders import coders
+from apache_beam.testing.test_pipeline import TestPipeline
+from apache_beam.testing.test_stream import TestStream
+from apache_beam.testing.util import assert_that
+from apache_beam.testing.util import equal_to
+from apache_beam.testing.util import equal_to_per_window
+from apache_beam.transforms import deduplicate
+from apache_beam.transforms import window
+from apache_beam.utils.timestamp import Duration
+from apache_beam.utils.timestamp import Timestamp
+
+
+# TestStream is only supported in streaming pipeline. The Deduplicate transform
+# also requires Timer support. Sickbaying this testsuite until dataflow runner
+# supports both TestStream and user timer.
+@attr('ValidatesRunner', 'sickbay-batch', 'sickbay-streaming')
+class DeduplicateTest(unittest.TestCase):
+  def __init__(self, *args, **kwargs):
+    self.runner = None
+    self.options = None
+    super(DeduplicateTest, self).__init__(*args, **kwargs)
+
+  def set_runner(self, runner):
+    self.runner = runner
+
+  def set_options(self, options):
+    self.options = options
+
+  def create_pipeline(self):
+    if self.runner and self.options:
+      return TestPipeline(runner=self.runner, options=self.options)
+    elif self.runner:
+      return TestPipeline(runner=self.runner)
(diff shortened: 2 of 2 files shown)


sdks/python/apache_beam/runners/sdf_utils.py
```diff
@@ -244,3 +251,63 @@ def get_estimator_state(self):
...
+
+    class DeduplicationFn(DoFn):
+      def process(
+          self,
+          element,
+          ts=DoFn.TimestampParam,
+          seen_state=DoFn.StateParam(state_spec),
+          expiry_timer=DoFn.TimerParam(timer_spec)):
+        if True not in seen_state.read():
+          if domain is userstate.TimeDomain.REAL_TIME:
```
lukecwik: The Python timer API seems primitive compared to the Java SDK

Use a switch case for the three types:
DEPENDENT_REAL_TIME
REAL_TIME
WATERMARK
and throw an error otherwise since you won't know what to do.

sdks/python/apache_beam/runners/sdf_utils.py
```diff
@@ -244,3 +251,63 @@ def get_estimator_state(self):
...
+            expiry_timer.set(Timestamp.now() + duration)
+          else:
+            expiry_timer.set(ts + duration)
+          seen_state.add(True)
+          value, _ = element
+          yield value
```
lukecwik: I should have used a guard style statement in the Java implementation instead of nesting.

```suggestion
        if True in seen_state.read():
          return

        if domain is userstate.TimeDomain.REAL_TIME:
          expiry_timer.set(Timestamp.now() + duration)
        else:
          expiry_timer.set(ts + duration)
        seen_state.add(True)
        value, _ = element
        yield value
```

sdks/python/apache_beam/runners/sdf_utils.py
```diff
@@ -244,3 +251,63 @@ def get_estimator_state(self):
         return None
 
     return _NoOpWatermarkEstimator()
+
+
+class DeduplictaionWithinDuration(ptransform.PTransform):
```
lukecwik: It would be useful to expose a keyed deduplication transform as the common implementation that all use internally so in the future we can turn into a well known URN and then runners could provide optimized deduplication transform implementations.

We want pipeline authors to use this transform and I think it should go into sdks/python/apache_beam/transforms/util.py or into a dedicated file such as sdks/python/apache_beam/transforms/ such as deduplicate.py.
CC: @udim What do you think?

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
```diff
@@ -1250,6 +1313,9 @@ def test_sdf_with_sdf_initiated_checkpointing(self):
   def test_sdf_with_watermark_tracking(self):
     raise unittest.SkipTest("This test is for a single worker only.")
 
+  def test_deduplication_transform_with_processing_time(self):
+    raise unittest.SkipTest("This test is for a single worker only.")
```
lukecwik: Is test stream unsupported in these other setups?

sdks/python/apache_beam/runners/sdf_utils.py
```diff
@@ -244,3 +251,63 @@ def get_estimator_state(self):
...
+  """ A PTransform which deduplicate input records over a time domain and
```
lukecwik:
```suggestion
  """ A PTransform which deduplicates input records over a time domain and
```

sdks/python/apache_beam/runners/sdf_utils.py
boyuanzz: I would prefer the second one: `sdks/python/apache_beam/transforms/deduplicate.py`

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
boyuanzz: No, I don't think so.

sdks/python/apache_beam/runners/sdf_utils.py
boyuanzz: Python timer only supports `REAL_TIME ` and `WATERMARK `. I changed it to checking the type of `TimeDomain` when constructing `DeduplictaionWithinDuration`.
The type is string so it's not switchable, and python doesn't have switch.

sdks/python/apache_beam/runners/sdf_utils.py
boyuanzz: Done.

sdks/python/apache_beam/runners/sdf_utils.py
boyuanzz: Moved transform to `sdks/python/apache_beam/transforms/deduplicate.py`

sdks/python/apache_beam/runners/sdf_utils.py
boyuanzz: When writing this part, I think we should update user-face timer API in python to do similar thing as Java, which makes life easier a lot.

sdks/python/apache_beam/runners/sdf_utils.py
udim: Unlike Java, we typically don't put transforms in their own module in Python. I would recommend putting it under an existing file such as transforms/util.py.

boyuanzz: Thanks Udi! Similar to `watermark_estimators.py`, I'm thinking about providing a collection of transforms for deduplicating purpose. Any reason that we don't put a group of similar transforms into one module?

udim: No, no reason. Go ahead with using a separate file

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
```diff
@@ -448,6 +449,42 @@ def is_buffered_correctly(actual):
 
       assert_that(actual, is_buffered_correctly)
 
+  def test_deduplication_transform_with_processing_time(self):
```
robertwb: fn_api_runner_test primarily for testing primitives. I would put this in apache_beam.transforms.deduplicate_test (or util_test, if that's where we plan to move it).

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
```diff
@@ -65,6 +65,7 @@
 from apache_beam.transforms import environments
 from apache_beam.transforms import userstate
 from apache_beam.transforms import window
+from apache_beam.transforms.deduplicate import DeduplictaionWithinDuration
```
robertwb: Generally we've been following the convention of importing modules rather than objects from modules. http://google.github.io/styleguide/pyguide.html#22-imports

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
```diff
@@ -448,6 +449,42 @@ def is_buffered_correctly(actual):
 
       assert_that(actual, is_buffered_correctly)
 
+  def test_deduplication_transform_with_processing_time(self):
+    # Note that current FnApiRunner doesn't respect either real timestamp.
```
robertwb: IIRC, the FnApiRunner does respect timestamps. (Not sure about the Create transform though.)

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
```diff
@@ -448,6 +449,42 @@ def is_buffered_correctly(actual):
...
+                         ]).advance_watermark_to_infinity())
+
+    with self.create_pipeline() as p:
+      actual = (
+          p
+          | test_stream
+          | DeduplictaionWithinDuration(
+              duration=timestamp.Duration(10),
+              time_domain=userstate.TimeDomain.WATERMARK))
+      assert_that(actual, equal_to(['value_1', 'value_1', 'value_2']))
```
robertwb: The duplication of value_1 probably merits a comment.

sdks/python/apache_beam/transforms/deduplicate.py
```diff
@@ -0,0 +1,108 @@
...
+from apache_beam.typehints import trivial_inference
+from apache_beam.utils.timestamp import Duration
+from apache_beam.utils.timestamp import Timestamp
+
+__all__ = [
+    'DeduplictaionWithinDuration',
+]
+
+
+class DeduplictaionWithinDuration(ptransform.PTransform):
```
robertwb: PTransform names are generally verbs. Perhaps just call this `Deduplicate`. The duration argument would typically be part of the constructor, e.g. `Deduplicate(duration=...)`, so wouldn't need to be repeated in the name.

sdks/python/apache_beam/transforms/deduplicate.py
```diff
@@ -0,0 +1,108 @@
...
+  a runner and care might need to be used to ensure that the deduplication time
+  limit is long enough to remove duplicates but short enough to not cause
+  performance problems within a runner. Each runner may provide an optimized
+  implementation of their choice using the deduplication time domain and
+  threshold specified.
+
+  Does not preserve any order the input PCollection might have had.
+  """
+  def __init__(
+      self, time_domain=TimeDomain.REAL_TIME, duration=Duration(10 * 60)):
```
robertwb: I would make duration a required parameter; I don't think there's a good default we can pick for all pipelines. (This also allows us to have a more intelligent default in the future, e.g. dynamically based on the actual watermark/propagation delay in the pipeline.)

Also, is it not possible to have both a real and watermark delay? This API forces you to do one or the other.

sdks/python/apache_beam/transforms/deduplicate.py
```diff
@@ -0,0 +1,108 @@
...
+      def process_timer(self, seen_state=DoFn.StateParam(state_spec)):
+        seen_state.clear()
+
+    return DeduplicationFn()
+
+  def expand(self, pcoll):
+    return (
+        pcoll
+        | 'KeyByElement' >> Map(lambda x: (x, None))
+        | 'DeduplicateFn' >> ParDo(self._create_deduplicate_fn()))
```
robertwb: I would expose a DeduplicatePerKey transform which emits a single KV per key, and then have the generic Deduplicate build on that.

sdks/python/apache_beam/transforms/deduplicate.py
```diff
@@ -0,0 +1,108 @@
...
+          ts=DoFn.TimestampParam,
+          seen_state=DoFn.StateParam(state_spec),
+          expiry_timer=DoFn.TimerParam(timer_spec)):
+        if True in seen_state.read():
+          return
+
+        if domain == TimeDomain.REAL_TIME:
+          expiry_timer.set(Timestamp.now() + duration)
+        elif domain == TimeDomain.WATERMARK:
+          expiry_timer.set(ts + duration)
```
robertwb: Suppose one has values with timestamps 10, 1, 12, and duration 10. This will not properly deduplicate because the timer will fire at 11 clearing state before 12 gets seen.

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
```diff
@@ -448,6 +449,42 @@ def is_buffered_correctly(actual):
...
+          | DeduplictaionWithinDuration(duration=timestamp.Duration(5)))
+      assert_that(actual, equal_to(['value_1']))
+
+  @unittest.skip('TestStream not yet supported')
+  def test_deduplication_transform_with_event_time(self):
+    test_stream = (
+        TestStream().advance_watermark_to(0).add_elements([
+            window.TimestampedValue('value_1', 1),
+            window.TimestampedValue('value_1', 2),
+            window.TimestampedValue('value_1', 10)
```
robertwb: It's be a stronger test to add elements one at a time.

sdks/python/apache_beam/transforms/deduplicate.py
boyuanzz: > Also, is it not possible to have both a real and watermark delay? This API forces you to do one or the other.

Currently the API only allows either watermark based or walltime based.  I'm not sure what kind of strategy we can do if allowing both. Do we want to do no matter which fires first, just clear another?

sdks/python/apache_beam/transforms/deduplicate.py
boyuanzz: Are you suggesting that we provide both `DeduplicatePerKey` and `Deduplicate` which uses value as key?

sdks/python/apache_beam/transforms/deduplicate.py
boyuanzz: Are you talking about the watermark case? yes the timer will be fires when the watermark advance to 11 if the timer is set by the value with timestamp 1

boyuanzz:

boyuanzz:
Run Python 3.5 Flink ValidatesRunner

sdks/python/apache_beam/runners/portability/fn_api_runner_test.py
boyuanzz: Thanks for the pointer!

boyuanzz:

boyuanzz:
Run Python 3.5 Flink ValidatesRunner

boyuanzz:
@robertwb Please take another look. Thanks for your help : )

sdks/python/apache_beam/transforms/deduplicate.py
```diff
@@ -0,0 +1,133 @@
...
+          'DeduplicatePerKey requires at lease provide either'
+          'processing_time_duration or event_time_duration.')
+    self.processing_time_duration = processing_time_duration
+    self.event_time_duration = event_time_duration
+
+  def _create_deduplicate_fn(self):
+    processing_timer_spec = userstate.TimerSpec(
+        'processing_timer', TimeDomain.REAL_TIME)
+    event_timer_spec = userstate.TimerSpec('event_timer', TimeDomain.WATERMARK)
+    state_spec = userstate.BagStateSpec('seen', BooleanCoder())
```
robertwb: Perhaps this should be a combining state?

sdks/python/apache_beam/transforms/deduplicate_test.py
```diff
@@ -0,0 +1,168 @@
...
+from apache_beam.transforms import deduplicate
+from apache_beam.transforms import window
+from apache_beam.utils.timestamp import Duration
+from apache_beam.utils.timestamp import Timestamp
+
+
+# TestStream is only supported in streaming pipeline. The Deduplicate transform
+# also requires Timer support. Sickbaying this testsuite until dataflow runner
+# supports both TestStream and user timer.
+@attr('ValidatesRunner', 'sickbay-batch', 'sickbay-streaming')
```
robertwb: Is it not possible to sickbay it for only dataflow?

sdks/python/apache_beam/transforms/deduplicate.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+@typehints.with_input_types(typing.Tuple[K, V])
+@typehints.with_output_types(typing.Tuple[K, V])
+class DeduplicatePerKey(ptransform.PTransform):
+  """ A PTransform which deduplicates <key, value> pair over a time domain and
+  threshold. Values in different windows will NOT be considered duplicates of
+  each other. Deduplication is best effort.
+
+  The durations specified may impose memory and/or storage requirements within
```
robertwb: I might phrase this as "Time durations are required so as to avoid unbounded memory and/or storage requirements on the runner..."

sdks/python/apache_beam/transforms/deduplicate.py
```diff
@@ -0,0 +1,133 @@
...
+K = typing.TypeVar('K')
+V = typing.TypeVar('V')
+
+
+@typehints.with_input_types(typing.Tuple[K, V])
+@typehints.with_output_types(typing.Tuple[K, V])
+class DeduplicatePerKey(ptransform.PTransform):
+  """ A PTransform which deduplicates <key, value> pair over a time domain and
+  threshold. Values in different windows will NOT be considered duplicates of
+  each other. Deduplication is best effort.
```
robertwb: It's not really best effort, rather it is only respected within the time domains (and there you can count on it).

sdks/python/apache_beam/transforms/deduplicate_test.py
boyuanzz: The  'sickbay-batch' and 'sickbay-streaming' is only used by dataflow suite now. And unfortunately, I don't we have runners supporting these python test now.

sdks/python/apache_beam/transforms/deduplicate.py
boyuanzz: The `seen_state` is only set once per key during that duration. I'm not sure whether combining state is more suitable here. What do you think?

robertwb: To follow up from what we discussed off-line, I think combining state semantically makes slightly more sense here, but as we're not adding to the state unless it's the first element there's no performance concerns, and I'm also fine with this as is.

boyuanzz:
Run Python PreCommit

boyuanzz:
Run Python PreCommit

PR #11061, opened 2020-03-06
[BEAM-9459] Revert "[BEAM-6374] Emit PCollection metrics from GoSDK"

lostluck:
Reverts apache/beam#10942 

Seems to be breaking the post commit. Since I'm going on vacation tonight, I'm rolling to back, and will look into it when I get back.


lostluck:
Run Go Postcommit

PR #11074, opened 2020-03-09
Store logical type values in Row instead of base values

reuvenlax:
@kanterov @alexvanboxel I believe this PR will fix the issues you've both had with logical types.

After some thought, I think we're better off storing the logical type value in the Row object. It potentially makes SchemaCoder a tiny bit slower, but unlikely this is noticeable. One bonus: it means that storage for OneOf types is far more memory efficient, as we no longer need to store the entire row.

### DIFF
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/coders/RowCoderGenerator.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/coders/RowCoderGenerator.java
@@ -23,6 +23,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Modifier;
 import java.util.BitSet;
 import java.util.List;
 import java.util.Map;
@@ -32,7 +33,6 @@ import org.apache.beam.sdk.annotations.Experimental.Kind;
 import org.apache.beam.sdk.schemas.Schema;
 import org.apache.beam.sdk.schemas.Schema.Field;
 import org.apache.beam.sdk.schemas.Schema.FieldType;
-import org.apache.beam.sdk.schemas.Schema.TypeName;
 import org.apache.beam.sdk.schemas.SchemaCoder;
 import org.apache.beam.sdk.values.Row;
 import org.apache.beam.vendor.bytebuddy.v1_10_8.net.bytebuddy.ByteBuddy;
@@ -47,11 +47,9 @@ import org.apache.beam.vendor.bytebuddy.v1_10_8.net.bytebuddy.dynamic.scaffold.I
 import org.apache.beam.vendor.bytebuddy.v1_10_8.net.bytebuddy.implementation.FixedValue;
 import org.apache.beam.vendor.bytebuddy.v1_10_8.net.bytebuddy.implementation.Implementation;
 import org.apache.beam.vendor.bytebuddy.v1_10_8.net.bytebuddy.implementation.bytecode.ByteCodeAppender;
+import org.apache.beam.vendor.bytebuddy.v1_10_8.net.bytebuddy.implementation.bytecode.ByteCodeAppender.Size;
 import org.apache.beam.vendor.bytebuddy.v1_10_8.net.bytebuddy.implementation.bytecode.Duplication;
 import org.apache.beam.vendor.bytebuddy.v1_10_8.net.bytebuddy.implementation.bytecode.StackManipulation;
-import org.apache.beam.vendor.bytebuddy.v1_10_8.net.bytebuddy.implementation.bytecode.StackManipulation.Compound;
-import org.apache.beam.vendor.bytebuddy.v1_10_8.net.bytebuddy.implementation.bytecode.TypeCreation;
-import org.apache.beam.vendor.bytebuddy.v1_10_8.net.bytebuddy.implementation.bytecode.collection.ArrayFactory;
 import org.apache.beam.vendor.bytebuddy.v1_10_8.net.bytebuddy.implementation.bytecode.member.FieldAccess;
 import org.apache.beam.vendor.bytebuddy.v1_10_8.net.bytebuddy.implementation.bytecode.member.MethodInvocation;
 import org.apache.beam.vendor.bytebuddy.v1_10_8.net.bytebuddy.implementation.bytecode.member.MethodReturn;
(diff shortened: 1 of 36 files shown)


reuvenlax:
R: @kanterov

alexvanboxel:
I can review the PR only this evening, but the first question that pops into my mind is:
What traits does the logical type need to have to play nice with Beam:
- Does it need to be comparable?
- Does the equal need to have proper behaviour?
These things we have control over when storing and working with the base type.

sdks/java/core/src/test/java/org/apache/beam/sdk/schemas/transforms/GroupTest.java
```diff
@@ -620,6 +531,88 @@ public void testGloballyWithSchemaAggregateFnNestedFields() {
...
+    Collection<BasicEnum> elements =
+        Lists.newArrayList(
+            BasicEnum.of("a", BasicEnum.Test.ONE), BasicEnum.of("a", BasicEnum.Test.TWO));
+
+    PCollection<Row> aggregate =
+        pipeline
+            .apply(Create.of(elements))
+            .apply(
+                Group.<BasicEnum>globally()
+                    .aggregateFieldBaseValue("enumeration", Sum.ofIntegers(), "enum_sum"));
```
kanterov: This looks rather unexpected that Combiner accepts `int` as an argument instead of `EnumerationType.Value`

sdks/java/core/src/test/java/org/apache/beam/sdk/schemas/transforms/GroupTest.java
```diff
@@ -620,6 +531,88 @@ public void testGloballyWithSchemaAggregateFnNestedFields() {
...
+  }
+
+  @Test
+  @Category(NeedsRunner.class)
+  public void testAggregateLogicalValuesGlobally() {
+    Collection<BasicEnum> elements =
+        Lists.newArrayList(
+            BasicEnum.of("a", BasicEnum.Test.ONE), BasicEnum.of("a", BasicEnum.Test.TWO));
+
+    SampleAnyCombineFn<EnumerationType.Value> sampleAnyCombineFn = new SampleAnyCombineFn<>(100);
```
kanterov: Does it only work because of type erasure, or is it actually aggregating `EnumerationType.Value` (not `Integer`)?

sdks/java/core/src/test/java/org/apache/beam/sdk/schemas/JavaFieldSchemaTest.java
```diff
@@ -592,41 +593,66 @@ public void testEnumFieldToRow() throws NoSuchSchemaException {
...
+            .addValues(enumerationType.valueOf("BLUE"), allColors)
+            .build();
+
+    List<Color> allColorsJava = Lists.newArrayList(Color.RED, Color.GREEN, Color.BLUE);
 
     SerializableFunction<PojoWithEnum, Row> toRow = registry.getToRowFunction(PojoWithEnum.class);
-    assertEquals(redRow, toRow.apply(new PojoWithEnum(Color.RED)));
-    assertEquals(greenRow, toRow.apply(new PojoWithEnum(Color.GREEN)));
-    assertEquals(blueRow, toRow.apply(new PojoWithEnum(Color.BLUE)));
+    System.err.println("CONVERTED " + toRow.apply(new PojoWithEnum(Color.RED, allColorsJava)));
```
kanterov: nit: we probably don't want `System.err.println` in tests

sdks/java/core/src/main/java/org/apache/beam/sdk/values/Row.java
```diff
@@ -362,7 +371,24 @@ public Boolean getBoolean(int idx) {
...
+  /**
+   * Returns the Logical Type base type for this field. If there are multiple nested logical types,
+   * they are all resolved to the first non-logical type. {@link IllegalStateException} is thrown if
+   * schema doesn't match.
+   */
+  @Nullable
+  public <T> T getLogicalTypeBaseValue(int idx, Class<T> clazz) {
+    Object value = getValue(idx);
+    FieldType fieldType = getSchema().getField(idx).getType();
+    checkArgument(fieldType.getTypeName().isLogicalType());
```
kanterov: nit: is it necessary to enforce? if yes, probably error message can be improved

sdks/java/core/src/main/java/org/apache/beam/sdk/values/Row.java
```diff
@@ -362,7 +371,24 @@ public Boolean getBoolean(int idx) {
...
+    while (fieldType.getTypeName().isLogicalType()) {
+      value = fieldType.getLogicalType().toBaseType(value);
```
kanterov: Should `toBaseType` handle null values for nullable fields or not? Can we assume that `toBaseType(null) == null`?

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/Sample.java
```diff
@@ -202,10 +202,10 @@ public void populateDisplayData(DisplayData.Builder builder) {
   }
 
   /** A {@link CombineFn} that combines into a {@link List} of up to limit elements. */
-  private static class SampleAnyCombineFn<T> extends CombineFn<T, List<T>, Iterable<T>> {
+  public static class SampleAnyCombineFn<T> extends CombineFn<T, List<T>, Iterable<T>> {
```
kanterov: nit: doesn't look necessary, we can probably replicate this combiner in tests, instead of extending public API

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/SchemaCoder.java
```diff
@@ -119,27 +86,6 @@ protected SchemaCoder(
     return RowCoder.of(schema);
   }
 
-  /** Returns the coder used for a given primitive type. */
-  public static <T> Coder<T> coderForFieldType(FieldType fieldType) {
```
kanterov: I find this was useful to have, and it was public. Instead, I would suggest `SchemaCoderHelpers` to be package-private, and this method to delegate to `SchemaCoderHelpers`.

sdks/java/core/src/main/java/org/apache/beam/sdk/coders/RowCoderGenerator.java
```diff
@@ -99,37 +97,14 @@
...
   private static final VarIntCoder VAR_INT_CODER = VarIntCoder.of();
-  private static final ForLoadedType NULLABLE_CODER = new ForLoadedType(NullableCoder.class);
 
   private static final String CODERS_FIELD_NAME = "FIELD_CODERS";
 
-  // A map of primitive types -> StackManipulations to create their coders.
-  private static final Map<TypeName, StackManipulation> CODER_MAP;
-
   // Cache for Coder class that are already generated.
   private static Map<UUID, Coder<Row>> generatedCoders = Maps.newConcurrentMap();
```
kanterov: nit: not directly relevant to this PR, but I guess this field should be final, and name should follow convention

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/FromRowUsingCreator.java
```diff
@@ -126,21 +127,25 @@ public T apply(Row row) {
               valueType,
               typeFactory);
     } else {
-      if (type.getTypeName().isLogicalType()
-          && OneOfType.IDENTIFIER.equals(type.getLogicalType().getIdentifier())) {
+      if (type.isLogicalType(OneOfType.IDENTIFIER)) {
```
kanterov: Probably I need to look into the code more. Why is `OneOfType` is a special case, while there is a generic branch for other logical types?

reuvenlax:
@alexvanboxel Yes, that is exactly true. This means that the logical type value must now implement equals and hashCode, whereas before it didn't (because we stored only the base value).

We could work around this by always converting to the base type inside of equals and hashCode, though that might be a bit expensive. However value types without a proper equals are generally discouraged in Java, so I think we can simply document that equals is required.

alexvanboxel:
> We could work around this by always converting to the base type inside of equals and hashCode, though that might be a bit expensive. However value types without a proper equals are generally discouraged in Java, so I think we can simply document that equals is required.

Documenting the requirement seems reasonable.

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/SchemaCoderHelpers.java
```diff
@@ -0,0 +1,132 @@
...
+                    coderForFieldType(fieldType.getMapValueType()));
+        break;
+      case LOGICAL_TYPE:
+        coder =
+            new LogicalTypeCoder(
+                fieldType.getLogicalType(),
+                coderForFieldType(fieldType.getLogicalType().getBaseType()));
+        break;
+      default:
+        coder = (Coder<T>) CODER_MAP.get(fieldType.getTypeName());
```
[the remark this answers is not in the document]
reuvenlax: Added  a null check.

sdks/java/core/src/test/java/org/apache/beam/sdk/schemas/transforms/GroupTest.java
reuvenlax: it's because aggregateFieldBaseValue is called instead of aggregateField. Would a different name make it more clear?

sdks/java/core/src/test/java/org/apache/beam/sdk/schemas/transforms/GroupTest.java
reuvenlax: No, it's actually aggregation EnumerationType.Value

sdks/java/core/src/main/java/org/apache/beam/sdk/values/Row.java
reuvenlax: good catch, we should.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/Sample.java
reuvenlax: Sample.anyCombineFn() exists, so I reverted this change

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/SchemaCoder.java
reuvenlax: done

sdks/java/core/src/main/java/org/apache/beam/sdk/coders/RowCoderGenerator.java
reuvenlax: done

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/FromRowUsingCreator.java
reuvenlax: because the actual conversion type changes depending on the oneOf case type. there might be a better way to handle this, but I'm not sure what it is yet.

reuvenlax:
Run Java PreCommit

sdks/java/core/src/test/java/org/apache/beam/sdk/schemas/transforms/GroupTest.java
kanterov: I see I didn't get it in the beginning, but now it all makes sense.

sdks/java/core/src/test/java/org/apache/beam/sdk/schemas/transforms/GroupTest.java
kanterov: I see, there are two methods: `aggregateFieldBaseValue` and `aggregateField`

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/SchemaCoderHelpers.java
```diff
@@ -0,0 +1,132 @@
...
+          .put(TypeName.INT64, VarLongCoder.of())
+          .put(TypeName.DECIMAL, BigDecimalCoder.of())
+          .put(TypeName.FLOAT, FloatCoder.of())
+          .put(TypeName.DOUBLE, DoubleCoder.of())
+          .put(TypeName.STRING, StringUtf8Coder.of())
+          .put(TypeName.DATETIME, InstantCoder.of())
+          .put(TypeName.BOOLEAN, BooleanCoder.of())
+          .build();
+
+  private static class LogicalTypeCoder extends Coder {
```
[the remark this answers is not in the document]
reuvenlax: Made these changes.

We should probably update SchemaCoder as well to do this properly, but that can be in a different PR IMO.

reuvenlax:
@alexvanboxel @kanterov any more comments?

@alexvanboxel this will conflict badly with your PR I think (as you moved code into a new file, which makes merges tricky) so we need to be quite careful with how we merge these PRs.

alexvanboxel:
> @alexvanboxel this will conflict badly with your PR I think (as you moved code into a new file, which makes merges tricky) so we need to be quite careful with how we merge these PRs.

No more comment, I like the consistency. We're using some logical types into our pipelines, so I will make a custom build before the release to see if all goes well.

I've merged my PR onto master as soon as I saw the LGTM, so I think the easiest thing is to rebase this branch onto master and handle the schema conflicts. I'll review as soon as this is done.

reuvenlax:
@alexvanboxel I've rebased on top of your PR now. the only conflicts were in Row.java, which I've now handled.

reuvenlax:
Run SQL Postcommit

reuvenlax:
Run SQL Postcommit

reuvenlax:
Run Java PreCommit

reuvenlax:
Run Java PreCommit

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/SchemaCoderHelpers.java
```diff
@@ -0,0 +1,168 @@
...
+      this.logicalType = logicalType;
+      this.baseTypeCoder = baseTypeCoder;
+      this.isDateTime = logicalType.getBaseType().equals(FieldType.DATETIME);
+    }
+
+    @Override
+    public void encode(InputT value, OutputStream outStream) throws CoderException, IOException {
+      BaseT baseType = logicalType.toBaseType(value);
+      if (isDateTime) {
+        baseType = (BaseT) ((ReadableInstant) baseType).toInstant();
```
kanterov: Why is this check needed?

reuvenlax:
Run Java PreCommit

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/SchemaCoderHelpers.java
reuvenlax: It's to maintain the current invariant that any ReadableInstant can be passed in, while the current InstantCoder requires an Instant. This used to be enforced in the Row builder because we would call toBaseType there.

We need to redo DateTime types I think (really we have a timestamp type, not a datetime type), but until then I wanted to maintain the existing behavior.

kanterov: Thanks. Makes sense. Agree that DateTime types need to be redone.

reuvenlax:
Run Java PreCommit

reuvenlax:
Run Java PreCommit

reuvenlax:
Run Java PreCommit

reuvenlax:
Run Java PreCommit

reuvenlax:
Run Java PreCommit

reuvenlax:
After 8 runs, the only Java Precommit failures have been random flakes (e.g. in Flink tests).

sdks/java/core/src/main/java/org/apache/beam/sdk/values/RowWithGetters.java
```diff
@@ -130,10 +130,11 @@ private Iterable getIterableValue(FieldType elementType, Object fieldValue) {
...
         Object convertedOneOfField =
-            getValue(oneOfValue.getFieldType(), oneOfValue.getValue(), null);
-        return (T)
-            oneOfType.toBaseType(
-                oneOfType.createValue(oneOfValue.getCaseType(), convertedOneOfField));
+            getValue(oneOfType.getFieldType(oneOfValue), oneOfValue.getValue(), null);
+        return (T) oneOfType.createValue(oneOfValue.getCaseType(), convertedOneOfField);
+      } else if (type.getTypeName().isLogicalType()) {
+        // Getters are assumed to return the base type.
+        return (T) type.getLogicalType().toInputType(fieldValue);
```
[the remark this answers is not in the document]
TheNeuralBit: Alternatively, I wonder if we could we add a way to register getters that are assumed to return the input type?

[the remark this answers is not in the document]
apilloud: There was a past world where the getters supported returning either logical type or base type. This was suppose to be whatever the previous transform provided (or the base type if coming from a coder). The goal was to save a conversion from base type to logical type for values that were just passed along. I believe the only use case was SQL and I removed that in #13930 because it is broken and like you notice here more expensive.

It should be possible to do more simplification such that we always work with logical types instead of base types. Eventually we will want a pass-through optimization again, but it will need to be at a lower level than logical types. (The only expensive type today is String, which isn't a logical type.)

PR #11084, opened 2020-03-09
[BEAM-9474] Improve robustness of BundleFactory and ProcessEnvironment

mxm:
The cleanup code in DefaultJobBundleFactory and its RemoteEnvironments may leak
resources. This is especially a concern when the execution engines reuses the
same JVM or underlying machines for multiple runs of a pipeline.

Exceptions encountered during cleanup should not lead to aborting the cleanup
procedure. Not all code handles this correctly. We should also ensure that the
cleanup succeeds even if the runner does not properly close the bundle,
e.g. when a exception occurs during closing the bundle.

--- | --- | --- | --- | ---

### DIFF
--- a/runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
+++ b/runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
@@ -21,6 +21,7 @@ import com.google.auto.value.AutoValue;
 import java.io.IOException;
 import java.util.IdentityHashMap;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.LinkedBlockingDeque;
@@ -76,6 +77,7 @@ import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.cache.RemovalNot
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.ImmutableList;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.ImmutableMap;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.Iterables;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.Sets;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -103,6 +105,10 @@ public class DefaultJobBundleFactory implements JobBundleFactory {
   private final LinkedBlockingDeque<LoadingCache<Environment, WrappedSdkHarnessClient>>
       availableCaches;
   private final boolean loadBalanceBundles;
+  /** Clients which were evicted due to environment expiration but still had pending references. */
+  private final Set<WrappedSdkHarnessClient> evictedActiveClients;
+
+  private boolean closed;
 
   public static DefaultJobBundleFactory create(JobInfo jobInfo) {
     PipelineOptions pipelineOptions =
@@ -140,6 +146,7 @@ public class DefaultJobBundleFactory implements JobBundleFactory {
             getMaxEnvironmentClients(jobInfo));
     this.availableCachesSemaphore = new Semaphore(environmentCaches.size(), true);
     this.availableCaches = new LinkedBlockingDeque<>(environmentCaches);
+    this.evictedActiveClients = Sets.newIdentityHashSet();
   }
 
   @VisibleForTesting
@@ -158,23 +165,28 @@ public class DefaultJobBundleFactory implements JobBundleFactory {
         createEnvironmentCaches(serverFactory -> serverInfo, getMaxEnvironmentClients(jobInfo));
     this.availableCachesSemaphore = new Semaphore(environmentCaches.size(), true);
     this.availableCaches = new LinkedBlockingDeque<>(environmentCaches);
+    this.evictedActiveClients = Sets.newIdentityHashSet();
   }
 
   private ImmutableList<LoadingCache<Environment, WrappedSdkHarnessClient>> createEnvironmentCaches(
       ThrowingFunction<ServerFactory, ServerInfo> serverInfoCreator, int count) {
-    CacheBuilder builder =
+    CacheBuilder<Environment, WrappedSdkHarnessClient> builder =
         CacheBuilder.newBuilder()
             .removalListener(
                 (RemovalNotification<Environment, WrappedSdkHarnessClient> notification) -> {
-                  int refCount = notification.getValue().unref();
-                  LOG.debug(
-                      "Removed environment {} with {} remaining bundle references.",
-                      notification.getKey(),
-                      refCount);
+                  WrappedSdkHarnessClient client = notification.getValue();
+                  int refCount = client.unref();
+                  if (refCount > 0) {
+                    LOG.warn(
+                        "Expiring environment {} with {} remaining bundle references. Taking note to clean it up during shutdown if the references are not removed by then.",
+                        notification.getKey(),
+                        refCount);
+                    evictedActiveClients.add(client);
+                  }
                 });
 
     if (environmentExpirationMillis > 0) {
-      builder = builder.expireAfterWrite(environmentExpirationMillis, TimeUnit.MILLISECONDS);
+      builder.expireAfterWrite(environmentExpirationMillis, TimeUnit.MILLISECONDS);
     }
 
     ImmutableList.Builder<LoadingCache<Environment, WrappedSdkHarnessClient>> caches =
@@ -252,14 +264,57 @@ public class DefaultJobBundleFactory implements JobBundleFactory {
   }
 
   @Override
-  public void close() throws Exception {
-    // Clear the cache. This closes all active environments.
-    // note this may cause open calls to be cancelled by the peer
-    for (LoadingCache<Environment, WrappedSdkHarnessClient> environmentCache : environmentCaches) {
-      environmentCache.invalidateAll();
-      environmentCache.cleanUp();
+  public synchronized void close() throws Exception {
+    if (closed) {
+      return;
     }
-    executor.shutdown();
+    // The following code is written defensively to guard against any exceptions occurring
+    // during shutdown. It is not visually appealing but unless it can be written more
+    // defensively, there is no reason to change it.
+    Exception exception = null;
+    for (LoadingCache<Environment, WrappedSdkHarnessClient> environmentCache : environmentCaches) {
+      try {
+        // Clear the cache. This closes all active environments.
+        // note this may cause open calls to be cancelled by the peer
+        environmentCache.invalidateAll();
+        environmentCache.cleanUp();
+      } catch (Exception e) {
+        if (exception != null) {
+          exception.addSuppressed(e);
+        } else {
+          exception = e;
+        }
+      }
+    }
+    // Cleanup any left-over environments which were not properly dereferenced by the Runner, e.g.
+    // when the bundle was not closed properly. This ensures we do not leak resources.
+    for (WrappedSdkHarnessClient client : evictedActiveClients) {
+      try {
+        //noinspection StatementWithEmptyBody
+        while (client.unref() > 0) {
+          // Remove any pending references from the client to force closing the environment
+        }
+      } catch (Exception e) {
+        if (exception != null) {
+          exception.addSuppressed(e);
+        } else {
+          exception = e;
+        }
+      }
+    }
+    try {
+      executor.shutdown();
+    } catch (Exception e) {
+      if (exception != null) {
+        exception.addSuppressed(e);
+      } else {
+        exception = e;
+      }
+    }
+    closed = true;
+    if (exception != null) {
+      throw exception;
+    }
   }
 
   private static ImmutableMap.Builder<String, RemoteOutputReceiver<?>> getOutputReceivers(
@@ -375,6 +430,11 @@ public class DefaultJobBundleFactory implements JobBundleFactory {
         }
       }
 
+      if (environmentExpirationMillis > 0) {
+        // Cleanup list of clients which were active during eviction but now do not hold references
+        evictedActiveClients.removeIf(c -> c.bundleRefCount.get() == 0);
+      }
+
       final RemoteBundle bundle =
           currentClient.processor.newBundle(
               getOutputReceivers(currentClient.processBundleDescriptor, outputReceiverFactory)
@@ -436,6 +496,8 @@ public class DefaultJobBundleFactory implements JobBundleFactory {
     private final ServerInfo serverInfo;
     private final AtomicInteger bundleRefCount = new AtomicInteger();
 
+    private boolean closed;
+
     static WrappedSdkHarnessClient wrapping(RemoteEnvironment environment, ServerInfo serverInfo) {
       SdkHarnessClient client =
           SdkHarnessClient.usingFnApiClient(
@@ -459,7 +521,10 @@ public class DefaultJobBundleFactory implements JobBundleFactory {
       return serverInfo;
     }
 
-    public void close() {
+    public synchronized void close() {
+      if (closed) {
+        return;
+      }
       // DO NOT ADD ANYTHING HERE WHICH MIGHT CAUSE THE BLOCK BELOW TO NOT BE EXECUTED.
       // If we exit prematurely (e.g. due to an exception), resources won't be cleaned up properly.
       // Please make an AutoCloseable and add it to the try statement below.
@@ -476,6 +541,7 @@ public class DefaultJobBundleFactory implements JobBundleFactory {
         // This will close _all_ of these even in the presence of exceptions.
         // The first exception encountered will be the base exception,
         // the next one will be added via Throwable#addSuppressed.
+        closed = true;
       } catch (Exception e) {
         LOG.warn("Error cleaning up servers {}", environment.getEnvironment(), e);
       }
@@ -487,13 +553,14 @@ public class DefaultJobBundleFactory implements JobBundleFactory {
     }
 
     private int unref() {
-      int count = bundleRefCount.decrementAndGet();
-      if (count == 0) {
+      int refCount = bundleRefCount.decrementAndGet();
+      Preconditions.checkState(refCount >= 0, "Reference count must not be negative.");
+      if (refCount == 0) {
         // Close environment after it was removed from cache and all bundles finished.
         LOG.info("Closing environment {}", environment.getEnvironment());
         close();
       }
-      return count;
+      return refCount;
     }
   }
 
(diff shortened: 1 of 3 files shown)


mxm:
retest this please

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
```diff
@@ -166,11 +168,20 @@ public static DefaultJobBundleFactory create(
         CacheBuilder.newBuilder()
             .removalListener(
                 (RemovalNotification<Environment, WrappedSdkHarnessClient> notification) -> {
-                  int refCount = notification.getValue().unref();
-                  LOG.debug(
-                      "Removed environment {} with {} remaining bundle references.",
-                      notification.getKey(),
-                      refCount);
+                  WrappedSdkHarnessClient client = notification.getValue();
```
tweise: Having this as part of the removal listener would prematurely close an environment that is still referenced. The purpose of the refcount is to be able to remove the environment from the cache when it expires but close it only after all references are gone.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/environment/ProcessManager.java
```diff
@@ -186,31 +186,28 @@ private void stopProcess(String id, Process process) {
       LOG.debug("Attempting to stop process with id {}", id);
       // first try to kill gracefully
       process.destroy();
-      long maxTimeToWait = 2000;
-      if (waitForProcessToDie(process, maxTimeToWait)) {
-        LOG.debug("Process for worker {} shut down gracefully.", id);
-      } else {
-        LOG.info("Process for worker {} still running. Killing.", id);
-        process.destroyForcibly();
+      long maxTimeToWait = 500;
```
tweise: Why the timeout change?

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/environment/ProcessManager.java
```diff
@@ -234,9 +231,7 @@ public void run() {
               // Graceful shutdown period
               Thread.sleep(200);
               break;
-            } catch (InterruptedException e) {
-              Thread.currentThread().interrupt();
-              throw new RuntimeException(e);
+            } catch (InterruptedException ignored) {
```
tweise: Add comment on why the exception is ignored.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
```diff
@@ -252,14 +263,47 @@ public StageBundleFactory forStage(ExecutableStage executableStage) {
...
-    // Clear the cache. This closes all active environments.
-    // note this may cause open calls to be cancelled by the peer
-    for (LoadingCache<Environment, WrappedSdkHarnessClient> environmentCache : environmentCaches) {
-      environmentCache.invalidateAll();
-      environmentCache.cleanUp();
+  public synchronized void close() throws Exception {
+    if (closed) {
+      return;
+    }
+    Exception exception = null;
```
tweise: It may be good to add an explanation here for the following elaborate cleanup logic as otherwise there is the danger it will be "simplified" in the future.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/environment/ProcessEnvironment.java
```diff
@@ -64,13 +63,32 @@ public InstructionRequestHandler getInstructionRequestHandler() {
...
-        isClosed = true;
+  public synchronized void close() throws Exception {
+    if (isClosed) {
+      return;
+    }
+    Exception exception = null;
+    try {
+      processManager.stopProcess(workerId);
+    } catch (Exception e) {
+      if (exception != null) {
```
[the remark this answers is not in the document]
mxm: I intentionally didn't do that. If you move around the the try block to change the order, this will break. I'd prefer to be defensive.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
```diff
@@ -252,14 +263,47 @@ public StageBundleFactory forStage(ExecutableStage executableStage) {
...
+      try {
+        executor.shutdown();
+      } catch (Exception e) {
+        if (exception != null) {
+          exception.addSuppressed(e);
+        } else {
+          exception = e;
+        }
+      }
+    } catch (Exception e) {
```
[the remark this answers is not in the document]
mxm: Not strictly, but the iterator could throw. I'm trying to be extra defensive here.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
```diff
@@ -166,11 +168,20 @@ public static DefaultJobBundleFactory create(
...
+                  // Double-check to trigger closing of all environments in case the "refing" does
+                  // not clean them up during operator shutdown. This is necessary in some
+                  // situations, e.g when the bundle cannot be closed and thus the ref cannot be
+                  // released. All environment types ensure they can only be closed once.
+                  if (refCount > 0) {
+                    LOG.warn(
+                        "Clearing remaining {} bundle references from environment {} to ensure it shuts down.",
+                        refCount,
+                        notification.getKey());
+                    //noinspection StatementWithEmptyBody
```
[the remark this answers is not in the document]
mxm: Just to inform when there were still references. The comment is to suppress a warning on the while loop which counts down the refs.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/environment/ProcessManager.java
mxm: The old timeout was too long.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/environment/ProcessManager.java
mxm: Will do. Ignoring it because there is not point to throw it here, we want to shutdown the processes after the grace period expired here.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
mxm: Good point.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
mxm: It doesn't work though if we do not ensure dereferencing under all circumstances. We need a safeguard here, also considering other runners may not dereference correctly. Generally, it is hard to guarantee dereferencing due to the nesting of DoFnRunners which may not even allow closing the bundle in error cases. I considered not doing this but I think it is the safer route.

If you take a step back, when would the reference counting really be useful? Every restarted job will anyways run in a new classloader, so the environment will never be recycled. When we call close we should tear down everything. 

Taking back another step, the reference counting should really be removed entirely. It was error prone from the beginning leading to subtle problems with dereferencing. If you don't mind, I'd remove it. What do you think?

mxm: I've removed the reference counting, PTAL.

mxm:
retest this please

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
tweise: As stated above, reference counting is required for environment expiration. The environment can only be closed when all bundles that reference it have finished. I would prefer we limit this PR scope strictly to the original purpose and discuss other changes separately.

tweise: The correct solution might be to enforce this from close, not from the cache removal listener.

mxm: You are right that we can't close the environment during removal due to pending references hold by still-processing bundles. Please take a look at the follow-up. I'm now quarantining clients which still hold references, to remove them during shutdown if they have not been dereferenced before.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
```diff
@@ -352,20 +407,18 @@ public RemoteBundle getBundle(
         // The blocking queue of caches for serving multiple bundles concurrently.
         currentCache = availableCaches.take();
         client = currentCache.getUnchecked(executableStage.getEnvironment());
-        client.ref();
```
tweise: `client.ref` needs to remain here, the lines below rely on that and it is also more readable.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
```diff
@@ -375,6 +428,11 @@ public RemoteBundle getBundle(
         }
       }
 
+      // Ensure client is referenced for this bundle, unref in close()
+      client.ref();
+      // Cleanup list of clients which were active during eviction but now do not hold references
+      evictedActiveClients.removeIf(c -> c.bundleRefCount.get() <= 0);
```
tweise: I'm not sure I like this in the path for every bundle. We could probably move it below line 416 (`preparedClients.keySet().removeIf(c -> c.bundleRefCount.get() <= 0);`)

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
```diff
@@ -352,20 +407,18 @@ public RemoteBundle getBundle(
...
         client = currentCache.getUnchecked(executableStage.getEnvironment());
-        client.ref();
 
         currentClient = preparedClients.get(client);
         if (currentClient == null) {
           // we are using this client for the first time
           preparedClients.put(client, currentClient = prepare(client, executableStage));
           // cleanup any expired clients
-          preparedClients.keySet().removeIf(c -> c.bundleRefCount.get() == 0);
+          preparedClients.keySet().removeIf(c -> c.bundleRefCount.get() <= 0);
```
tweise: Why the change from `==` to `<=`?

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/environment/ProcessManager.java
tweise: How did that manifest? 500ms timeout for process termination seems too aggressive. I would prefer a longer timeout to allow for connections to be closed gracefully.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
mxm: Please explain which lines rely on it.

Readability is highly highly subjective, I find it more readable to not duplicate the same code for two branches, as it is currently the case (which is potentially error-prone).

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
mxm: I prefer including the entire range whenever I write such checks. Just a way to guard against future changes where we might countdown after the value has reached zero.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
mxm: This is a cheap operation. I don't think we can move it because we need to run this cleanup for all branches.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/environment/ProcessManager.java
mxm: The shutdown code is synchronous, so I'd prefer a shorter timeout here. It would be a good improvement to make it async.

I've tested this on a cluster with many runs and I've not seen a single instance lingering. Also I did not notice any difference in the logs. The environment will be torn down last, after all connections have been closed. So failures would not be visible anymore.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
mxm: I checked again and couldn't see how this change alters the behavior. In any case, I don't mind to move it back to unblock this PR.

mxm:
Thanks for your comments @tweise. Let me know if you have more questions.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/environment/ProcessManager.java
mxm: If you want I can restore the old timeout but I would then also change the code to make the stopping async or at least stop all the processes at once and then wait (instead of tearing down the process one-by-one).

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
tweise: Below, `preparedClients.keySet().removeIf(c -> c.bundleRefCount.get() <= 0);` removes everything that isn't referenced. These two statements are logically one unit and hence I prefer to not scatter them:
```
client = currentCache.getUnchecked(executableStage.getEnvironment());
        client.ref();
client.ref();
```

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
tweise: On the flip side, this could hide a bug in the reference counting. That could be covered with tests though. Not important for this PR.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/environment/ProcessManager.java
tweise: If you have verified that the graceful shutdown works (in the happy path), then we are good. Maybe add a comment to the code, since all of this isn't very obvious.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
mxm: It's not true that a later `ref()` introduces a bug for `preparedClients.keySet().removeIf(c -> c.bundleRefCount.get() <= 0);` because the refcount will be >0, otherwise we wouldn't be able to retrieve the client from the cache.

In any case, I will revert this change.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
mxm: I'll also revert this change to not distract from the main objective of this PR.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
tweise: > It's not true that a later `ref()` introduces a bug for `preparedClients.keySet().removeIf(c -> c.bundleRefCount.get() <= 0);` because the refcount will be >0, otherwise we wouldn't be able to retrieve the client from the cache.

Cache and environment are shared between executable stages. So the refcount can become 0 with concurrent eviction and release. That actually raises the question if these 2 statements should be atomic.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
tweise: Not important, can stay as is. What would be good to check is that the environment expiration still works.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/environment/ProcessManager.java
mxm: It's not always shutting down gracefully but that's what the change is about: removing processes and ensuring a quick recovery time. It's a trade-off. Ideally we would want to allow more time but if we wait 2 seconds with an SDK parallelism of 16, that's already more than half a minute waiting time. We really want to do the process removal in parallel. I'll look into this.

I'm not sure the ProcessManager is a good place to document the shutdown behavior. If you have any suggestions though, I'll add them here.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
mxm: I've improved this to only remove when environment expiration is used.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
mxm: Yes, that makes sense. I've already reverted the change.

I suppose there is a race condition where we retrieve an environment X and before we can call `ref()` on it, we evict the environment X, close all its references, and shut it down. This will result in a job restart.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/DefaultJobBundleFactory.java
```diff
@@ -487,13 +553,14 @@ private int ref() {
     }
 
     private int unref() {
-      int count = bundleRefCount.decrementAndGet();
-      if (count == 0) {
+      int refCount = bundleRefCount.decrementAndGet();
+      Preconditions.checkState(refCount >= 0, "Reference count must not be negative.");
```
mxm: FYI, I've added this check instead to check for correct bounds.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/environment/ProcessManager.java
tweise: Won't delay the PR for it!

PR #11086, opened 2020-03-09
[BEAM-8910] Make custom BQ source read from Avro

pabloem:
This adds support for reading data from BigQuery using FastAvro in the new BigQuery source transform: `_ReadFromBigQuery`.

This creates a few subtle changes in the way that data is returned by BigQuery. These changes are the following:

- All the time-specific data types in BigQuery (`DATE`, `DATETIME`, `TIMESTAMP`) are converted into Python-specific types (`datetime.date`, `datetime.datetime`, `datetime.datetime`).
  - This is different from `BigQuerySource`, where we export JSON data, and all of these datatypes are returned as strings.

- The BigQuery `BYTES` type is returned as Python `bytes`.
  - This is different from `BigQuerySource`, where we export JSON data. For JSON, BigQuery (and beam's BigQuerySource) returns base64-encoded bytes.

Note that in tests, the behavior of BigQuerySource **remains the same**, while _ReadFromBigQuery has different behavior.

This transform does not need to honor backwards compatibility, as it is a new transform, and **it will not replace** BigQuerySource automatically. Nonetheless, we should definitely note the subtle API differences for it.

r: @chamikaramj @tvalentyn @kamilwu 

cc: @mwalenia 

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -59,6 +59,10 @@
 * Support for X source added (Java/Python) ([BEAM-X](https://issues.apache.org/jira/browse/BEAM-X)).
 * Support for reading from Snowflake added (Java) ([BEAM-9722](https://issues.apache.org/jira/browse/BEAM-9722)).
 * Support for writing to Splunk added (Java) ([BEAM-8596](https://issues.apache.org/jira/browse/BEAM-8596)).
+* A new transform to read from BigQuery has been added: `apache_beam.io.gcp.bigquery.ReadFromBigQuery`. This transform
+  is experimental. It reads data from BigQuery by exporting data to Avro files, and reading those files. It also supports
+  reading data by exporting to JSON files. This has small differences in behavior for Time and Date-related fields. See
+  Pydoc for more information.
 
 ## New Features / Improvements
 
--- a/sdks/python/apache_beam/io/gcp/big_query_query_to_table_it_test.py
+++ b/sdks/python/apache_beam/io/gcp/big_query_query_to_table_it_test.py
@@ -173,7 +173,8 @@ class BigQueryQueryToTableIT(unittest.TestCase):
         'output_schema': DIALECT_OUTPUT_SCHEMA,
         'use_standard_sql': False,
         'wait_until_finish_duration': WAIT_UNTIL_FINISH_DURATION_MS,
-        'on_success_matcher': all_of(*pipeline_verifiers)
+        'on_success_matcher': all_of(*pipeline_verifiers),
+        'experiments': 'use_beam_bq_sink',
     }
     options = self.test_pipeline.get_full_options_as_args(**extra_opts)
     big_query_query_to_table_pipeline.run_bq_pipeline(options)
@@ -196,7 +197,8 @@ class BigQueryQueryToTableIT(unittest.TestCase):
         'output_schema': DIALECT_OUTPUT_SCHEMA,
         'use_standard_sql': True,
         'wait_until_finish_duration': WAIT_UNTIL_FINISH_DURATION_MS,
-        'on_success_matcher': all_of(*pipeline_verifiers)
+        'on_success_matcher': all_of(*pipeline_verifiers),
+        'experiments': 'use_beam_bq_sink',
     }
     options = self.test_pipeline.get_full_options_as_args(**extra_opts)
     big_query_query_to_table_pipeline.run_bq_pipeline(options)
@@ -254,7 +256,8 @@ class BigQueryQueryToTableIT(unittest.TestCase):
         'output_schema': NEW_TYPES_OUTPUT_SCHEMA,
         'use_standard_sql': False,
         'wait_until_finish_duration': WAIT_UNTIL_FINISH_DURATION_MS,
-        'on_success_matcher': all_of(*pipeline_verifiers)
+        'on_success_matcher': all_of(*pipeline_verifiers),
+        'experiments': 'use_beam_bq_sink',
     }
     options = self.test_pipeline.get_full_options_as_args(**extra_opts)
     big_query_query_to_table_pipeline.run_bq_pipeline(options)
--- a/sdks/python/apache_beam/io/gcp/big_query_query_to_table_it_test.py
+++ b/sdks/python/apache_beam/io/gcp/big_query_query_to_table_it_test.py
@@ -239,7 +239,7 @@ class BigQueryQueryToTableIT(unittest.TestCase):
     self.assertEqual(kms_key, table.encryptionConfiguration.kmsKeyName)
 
   @attr('IT')
-  def test_big_query_new_types_json(self):
+  def test_big_query_new_types(self):
     expected_checksum = test_utils.compute_hash(NEW_TYPES_OUTPUT_EXPECTED)
     verify_query = NEW_TYPES_OUTPUT_VERIFY_QUERY % self.output_table
     pipeline_verifiers = [
@@ -263,7 +263,7 @@ class BigQueryQueryToTableIT(unittest.TestCase):
     big_query_query_to_table_pipeline.run_bq_pipeline(options)
 
   @attr('IT')
-  def test_big_query_new_types(self):
+  def test_big_query_new_types_avro(self):
     expected_checksum = test_utils.compute_hash(NEW_TYPES_OUTPUT_EXPECTED)
     verify_query = NEW_TYPES_OUTPUT_VERIFY_QUERY % self.output_table
     pipeline_verifiers = [
--- a/sdks/python/apache_beam/io/gcp/big_query_query_to_table_it_test.py
+++ b/sdks/python/apache_beam/io/gcp/big_query_query_to_table_it_test.py
@@ -239,6 +239,30 @@ class BigQueryQueryToTableIT(unittest.TestCase):
     self.assertEqual(kms_key, table.encryptionConfiguration.kmsKeyName)
 
   @attr('IT')
+  def test_big_query_new_types_json(self):
+    expected_checksum = test_utils.compute_hash(NEW_TYPES_OUTPUT_EXPECTED)
+    verify_query = NEW_TYPES_OUTPUT_VERIFY_QUERY % self.output_table
+    pipeline_verifiers = [
+        PipelineStateMatcher(),
+        BigqueryMatcher(
+            project=self.project,
+            query=verify_query,
+            checksum=expected_checksum)
+    ]
+    self._setup_new_types_env()
+    extra_opts = {
+        'query': NEW_TYPES_QUERY % (self.dataset_id, NEW_TYPES_INPUT_TABLE),
+        'output': self.output_table,
+        'output_schema': NEW_TYPES_OUTPUT_SCHEMA,
+        'use_standard_sql': False,
+        'wait_until_finish_duration': WAIT_UNTIL_FINISH_DURATION_MS,
+        'use_json_exports': True,
+        'on_success_matcher': all_of(*pipeline_verifiers)
+    }
+    options = self.test_pipeline.get_full_options_as_args(**extra_opts)
+    big_query_query_to_table_pipeline.run_bq_pipeline(options)
+
+  @attr('IT')
   def test_big_query_new_types(self):
     expected_checksum = test_utils.compute_hash(NEW_TYPES_OUTPUT_EXPECTED)
     verify_query = NEW_TYPES_OUTPUT_VERIFY_QUERY % self.output_table
--- a/sdks/python/apache_beam/io/gcp/bigquery.py
+++ b/sdks/python/apache_beam/io/gcp/bigquery.py
@@ -700,7 +700,7 @@ class _CustomBigQuerySource(BoundedSource):
 
       unused_schema, metadata_list = self._export_files(bq)
       self.split_result = [
-          create_avro_source(metadata.path, 0, True)
+          create_avro_source(metadata.path, 0, True, use_fastavro=True)
           for metadata in metadata_list
       ]
 
--- a/sdks/python/apache_beam/io/gcp/bigquery.py
+++ b/sdks/python/apache_beam/io/gcp/bigquery.py
@@ -611,7 +611,7 @@ class _CustomBigQuerySource(BoundedSource):
       flatten_results=True,
       kms_key=None,
       bigquery_job_labels=None,
-      backwards_compatible_data_format=False):
+      use_json_exports=False):
     if table is not None and query is not None:
       raise ValueError(
           'Both a BigQuery table and a query were specified.'
@@ -642,12 +642,14 @@ class _CustomBigQuerySource(BoundedSource):
     self.bigquery_job_labels = bigquery_job_labels or {}
 
   def display_data(self):
+    export_format = 'JSON' if self.use_json_exports else 'AVRO'
     return {
         'table': str(self.table_reference),
         'query': str(self.query),
         'project': str(self.project),
         'use_legacy_sql': self.use_legacy_sql,
         'bigquery_job_labels': json.dumps(self.bigquery_job_labels),
+        'export_file_format': export_format,
     }
 
   def estimate_size(self):
@@ -694,7 +696,7 @@ class _CustomBigQuerySource(BoundedSource):
     return project
 
   def _create_source(self, path, schema):
-    if not self.backwards_compatible:
+    if not self.use_json_exports:
       return create_avro_source(path, 0, True, use_fastavro=True)
     else:
       return TextSource(
@@ -761,7 +763,7 @@ class _CustomBigQuerySource(BoundedSource):
       bigquery.TableSchema instance, a list of FileMetadata instances
     """
     job_id = uuid.uuid4().hex
-    if self.backwards_compatible:
+    if self.use_json_exports:
       job_ref = bq.perform_extract_job([self.gcs_location],
                                        job_id,
                                        self.table_reference,
@@ -1687,10 +1689,13 @@ class ReadFromBigQuery(PTransform):
       to BigQuery export and query jobs created by this transform. See:
       https://cloud.google.com/bigquery/docs/reference/rest/v2/\
               Job#JobConfiguration
-    backwards_compatible_data_format (bool): By default, this transform reads
+    use_json_exports (bool): By default, this transform reads
       data in native Python types that come from Avro-exports of BigQuery. By
       setting this flag to True, the transform will return the types from
       exports to JSON files, much like the older BigQuerySource.
+      To learn more about type conversions between BigQuery and Avro, see:
+      https://cloud.google.com/bigquery/docs/loading-data-cloud-storage-avro\
+              #avro_conversions
    """
   def __init__(self, gcs_location=None, validate=False, *args, **kwargs):
     if gcs_location:
--- a/sdks/python/apache_beam/io/gcp/bigquery.py
+++ b/sdks/python/apache_beam/io/gcp/bigquery.py
@@ -45,8 +45,8 @@ call *one* row of the main table and *all* rows of the side table. The runner
 may use some caching techniques to share the side inputs between calls in order
 to avoid excessive reading:::
 
-  main_table = pipeline | 'VeryBig' >> beam.io.Read(beam.io.BigQuerySource()
-  side_table = pipeline | 'NotBig' >> beam.io.Read(beam.io.BigQuerySource()
+  main_table = pipeline | 'VeryBig' >> beam.io.ReadFroBigQuery(...)
+  side_table = pipeline | 'NotBig' >> beam.io.ReadFromBigQuery(...)
   results = (
       main_table
       | 'ProcessData' >> beam.Map(
@@ -58,14 +58,8 @@ as a parameter to the Map transform. AsList signals to the execution framework
 that its input should be made available whole.
 
 The main and side inputs are implemented differently. Reading a BigQuery table
-as main input entails exporting the table to a set of GCS files (currently in
-JSON format) and then processing those files. Reading the same table as a side
-input entails querying the table for all its rows. The coder argument on
-BigQuerySource controls the reading of the lines in the export files (i.e.,
-transform a JSON object into a PCollection element). The coder is not involved
-when the same table is read as a side input since there is no intermediate
-format involved. We get the table rows directly from the BigQuery service with
-a query.
+as main input entails exporting the table to a set of GCS files (in AVRO or in
+JSON format) and then processing those files.
 
 Users may provide a query to read from rather than reading all of a BigQuery
 table. If specified, the result obtained by executing the specified query will
@@ -77,6 +71,12 @@ be used as the data of the input transform.::
 When creating a BigQuery input transform, users should provide either a query
 or a table. Pipeline construction will fail with a validation error if neither
 or both are specified.
+
+When reading from BigQuery using `BigQuerySource`, bytes are returned as
+base64-encoded bytes. When reading via `ReadFromBigQuery`, bytes are returned
+as bytes without base64 encoding. This is due to the fact that ReadFromBigQuery
+uses Avro expors by default. To get base64-encoded bytes, you can use the flag
+`use_json_exports` to export data as JSON, and receive base64-encoded bytes.
 
 Writing Data to BigQuery
 ========================
@@ -225,8 +225,7 @@ The GEOGRAPHY data type works with Well-Known Text (See
 https://en.wikipedia.org/wiki/Well-known_text) format for reading and writing
 to BigQuery.
 BigQuery IO requires values of BYTES datatype to be encoded using base64
-encoding when writing to BigQuery. When bytes are read from BigQuery they are
-returned as base64-encoded bytes.
+encoding when writing to BigQuery.
 """
 
 # pytype: skip-file
@@ -640,6 +639,7 @@ class _CustomBigQuerySource(BoundedSource):
     self.split_result = None
     self.options = pipeline_options
     self.bigquery_job_labels = bigquery_job_labels or {}
+    self.use_json_exports = use_json_exports
 
   def display_data(self):
     export_format = 'JSON' if self.use_json_exports else 'AVRO'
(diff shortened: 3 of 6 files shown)


pabloem:
Run Python 3.7 PostCommit

pabloem:
Run Python 3.7 PostCommit

pabloem:

pabloem:
Run Python PreCommit

pabloem:
Run Python 2 PostCommit

pabloem:
Run Python 2 PostCommit

pabloem:
Run Python 3.7 PostCommit

pabloem:
@chamikaramj @tvalentyn 

Hey there, I'm working on this improvement. There's some interesting outcomes here:

- For datetime types (date,time,datetime,timestamp), AvroSource returns python types (datetime.date, datetime.time,datetime.datetime, datetime.datetime) - along with timezone info. I feel that in this case, Python types are the most desirable. Thoughts? _CustomBQSource is experimental, so I wouldn't think it's subject to backwards compatibility concerns.

- For bytes, AvroSource is currently returning base64 encoded bytes. I am trying to figure out how to handle this. For py2, strings and bytes may not be differentiable, so we may not be able to handle this properly...

tvalentyn:
> * For bytes, AvroSource is currently returning base64 encoded bytes. I am trying to figure out how to handle this. For py2, strings and bytes may not be differentiable, so we may not be able to handle this properly...
Do you have a summary of Py2 vs Py3  x Current Behavior vs New behavior? What is returned on Py3 - is it also base64 encoded bytes?
This is a prior discussion on Py3 migration issues in Bigquery: https://lists.apache.org/thread.html/f35c836887014e059527ed1a806e730321e2f9726164a3030575f455%40%3Cdev.beam.apache.org%3E. It took us some time to understand what was going on (https://lists.apache.org/thread.html/0c2178cf8e5d9e77c4f233f05a0b87b6011a1daa1a5ae47b41463af5%40%3Cdev.beam.apache.org%3E)

pabloem:
Run Python 3.7 PostCommit

pabloem:
Run Python 2 PostCommit

pabloem:
Run Python 3.7 PostCommit

pabloem:
Run Python 3.7 PostCommit

pabloem:
Run Python PreCommit

pabloem:
Run Python 2 PostCommit

pabloem:
Run Python PreCommit

pabloem:
@kamilwu @chamikaramj PTAL : ) -- see first comment for analysis on typing, and changes

sdks/python/apache_beam/io/gcp/bigquery_read_it_test.py
```diff
@@ -236,11 +251,12 @@ def create_table(cls, table_name):
...
         cls.project, cls.dataset_id, table_name, table_data)
 
-  def get_expected_data(self):
+  def get_expected_data(self, native=True):
+    byts = b'\xab\xac'
     expected_row = {
         'float': 0.33,
         'numeric': Decimal('10'),
-        'bytes': base64.b64encode(b'\xab\xac'),
+        'bytes': base64.b64encode(byts) if native else byts,
```
tvalentyn: How intuitive will this behavior be to users? should't the output be consistent? 
If we decide to keep it, how will this be documented?

pabloem: The behavior will be different for different transforms. Users will need to explicitly change the transform in their code. We can make them aware of the new transform, and its typing differences in release notes, and possibly in Pydoc for the new transform as well. Thoughts?

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -663,14 +662,10 @@ def split(self, desired_bundle_size, start_position=None, stop_position=None):
         self._setup_temporary_dataset(bq)
         self.table_reference = self._execute_query(bq)
 
-      schema, metadata_list = self._export_files(bq)
+      unused_schema, metadata_list = self._export_files(bq)
```
kamilwu: Does it mean that coder is no longer needed for `_CustomBigQuerySource`?

sdks/python/apache_beam/io/gcp/bigquery_read_it_test.py
```diff
@@ -69,6 +70,20 @@ def wrapped(self):
   return inner
 
 
+def datetime_to_utc(element):
+  for k, v in element.items():
+    if isinstance(v, datetime.time):
```
kamilwu: We can combine this condition with the one below:
`if isinstance(v, datetime.time) or isinstance(v, datetime.date):`

sdks/python/apache_beam/io/gcp/bigquery_read_it_test.py
tvalentyn: Bytes treatment should be called out in IO doc, we do mention it: https://github.com/apache/beam/blob/8bc2880cca40c00a96623b3ce96ea0b856af76c9/sdks/python/apache_beam/io/gcp/bigquery.py#L227. 

b64encoding may be unnecessary, and less efficient. I think the reason native IO encodes bytes with b64 is because that was the behavior in Java SDK. We can argue that's not necessary. However I am concerned about the consistency of the UX here. Different UX for two transforms means the transforms will not be interchangeable, and users might overlook this.  This might also cause friction in cross-language pipelines.

sdks/python/apache_beam/io/gcp/bigquery.py
pabloem: You're correct!

sdks/python/apache_beam/io/gcp/bigquery_read_it_test.py
pabloem: I think that's reasonable. But I'd think that after slight confusion, users would figure out the differences.

Another option is to provide a parameter to the transform to have backwards-compatible types at a performance cost.

@chamikaramj @tvalentyn  wdyt?

sdks/python/apache_beam/io/gcp/bigquery.py
kamilwu: This is great! That would mean that we can safely remove `_JsonToDictCoder`. Do you think you can do this as a part of this PR?

sdks/python/apache_beam/io/gcp/bigquery_read_it_test.py
tvalentyn: Having a parameter for backwards-compatibility would make sense to me, but will defer to @chamikaramj  on this. Since new transform comes with a new name, whoever will be using it, will have to read the API, so will be aware of the issue. But that would mean we add a knob, potentially indefinitely.

sdks/python/apache_beam/io/gcp/bigquery.py
pabloem: We may do that, but if we end up keeping a backwards compatibility flag, we'll need to keep the coder as-is.

kamilwu: I see. Then let's wait for the decision.

sdks/python/apache_beam/io/gcp/bigquery_read_it_test.py
pabloem: So the proposal right now is: To have a parameter for backwards compatibility in the transform, but by default return Avro-returned types. We will need to have clear Pydoc on _ReadFromBigQuery, and on CHANGES.md.
@chamikaramj thoughts?

pabloem:
Run Python 3.5 PostCommit

pabloem:
Py35 postcommit run: https://builds.apache.org/job/beam_PostCommit_Python35_PR/114/

pabloem:
Run Python 3.5 PostCommit

pabloem:
Run Python 3.5 PostCommit

pabloem:
Run Python 3.5 PostCommit

pabloem:
Run Python 3.5 PostCommit

pabloem:
Run Python 3.5 PostCommit

pabloem:
Run Python 3.5 PostCommit

pabloem:
Passing py35 pc: https://builds.apache.org/job/beam_PostCommit_Python35_PR/118/

pabloem:
Lint failures unrelated

pabloem:
Precommit failures unrelated.

sdks/python/apache_beam/io/gcp/bigquery_read_it_test.py
pabloem: @chamikaramj @tvalentyn @kamilwu - added a backward compatibility flag. Can you review?

pabloem: If the changes are reasonable, I'll add to release notes

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -609,7 +611,8 @@ def __init__(
       coder=None,
       use_standard_sql=False,
       flatten_results=True,
-      kms_key=None):
+      kms_key=None,
+      backwards_compatible_data_format=None):
```
tvalentyn, line 615: 1. Can we use a bool as a default value instead of `None`?
2. Can we use rename this flag to something that describes the course of action, for example: `use_json_data_representation=False` - you can come up with a better name.

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -1570,6 +1584,10 @@ class _ReadFromBigQuery(PTransform):
       bucket where the extracted table should be written as a string or
       a :class:`~apache_beam.options.value_provider.ValueProvider`. If
       :data:`None`, then the temp_location parameter is used.
+    backwards_compatible_data_format (bool): By default, this transform reads
+      data in native Python types that come from Avro-exports of BigQuery. By
+      setting this flag to True, the transform will return JSON-types, like
+      the older BigQuerySource.
```
tvalentyn, line 1604: 1. Do all BQ types have a native python type?
2. What are JSON-types?

sdks/python/apache_beam/io/gcp/bigquery.py
pabloem: 1. Done.
2. I've called it `use_json_exports` - does that help?

sdks/python/apache_beam/io/gcp/bigquery.py
pabloem: I've added a link to https://cloud.google.com/bigquery/docs/loading-data-cloud-storage-avro#avro_conversions for details on the conversions. I've also rephrased the doc. LMK what you think.

sdks/python/apache_beam/io/gcp/bigquery_read_it_test.py
chamikaramj: I think having a single "backwards compatibility" parameter can be confusing and error prone. We might still be backwards incompatible in some unknown ways. I think it's better to document exact differences between the transforms and give clear documentation/guidelines for anyone who is upgrading.

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -610,7 +611,8 @@ def __init__(
       coder=None,
       use_standard_sql=False,
       flatten_results=True,
-      kms_key=None):
+      kms_key=None,
+      use_json_exports=False):
```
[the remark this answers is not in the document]
pabloem: not supporting JSON would increase the amount of changes that users need to make to migrate to the new source, as the data output types would change. I don't think this would be ideal. Wouldn't it be better to allow users to migrate to the new source without having to change the rest of their pipeline?

pabloem:
@chamikaramj @tvalentyn @kamilwu 
I've documented the data types that are different, and provided links to BQ documentation to see differences in representation. I've also addressed Cham's comments. I think it's worth keeping JSON export capability.

Finally, I've added the change to CHANGES.md. This should help us move the source forward towards full support. Thoughts?

pabloem:
Run Python 3.5 PostCommit

pabloem:

pabloem:
Run Python 3.5 PostCommit

pabloem:
Run Python PreCommit

pabloem:
@chamikaramj @tvalentyn @kamilwu PTAL - see comment https://github.com/apache/beam/pull/11086#issuecomment-615517971

tvalentyn:
Thanks Pablo.
1. Can we take another look at a big comment at the beginning of bigquery.py and see if it needs an update. It sounds like we now have two ways of reading and writing to BigQuery. Can we add a guidance to users when to use which, and what they will need to change when they switch from one to another?

2. The same comment says: BigQuery IO requires values of BYTES datatype to be encoded using base64
encoding when writing to BigQuery. When bytes are read from BigQuery they are
returned as base64-encoded bytes. Does this apply to the new source? Sounds like not, so please clarify.
3. What is the test story for BQ IO? I see several branches here: natve vs non-native IO, Avro vs Json. Which combinations are tested?
4. Big Query tests are one of the major source of [postcommit flakiness](https://issues.apache.org/jira/issues/?jql=text~BigQuery%20AND%20summary~flaky%20AND%20status%3DOpen%20AND%20summary~Test)? What are the plans to address that?

pabloem:
> Thanks Pablo.
> 
> 1. Can we take another look at a big comment at the beginning of bigquery.py and see if it needs an update. It sounds like we now have two ways of reading and writing to BigQuery. Can we add a guidance to users when to use which, and what they will need to change when they switch from one to another?
> 2. The same comment says: BigQuery IO requires values of BYTES datatype to be encoded using base64
>    encoding when writing to BigQuery. When bytes are read from BigQuery they are
>    returned as base64-encoded bytes. Does this apply to the new source? Sounds like not, so please clarify.

Ah great observation. Thanks Valentyn. I've rephrased that. LMK what you think now

> 3. What is the test story for BQ IO? I see several branches here: natve vs non-native IO, Avro vs Json. Which combinations are tested?

There are tests for pretty much all the variants. There are:
- Native: `BigQuerySource`. This is widely tested and used.
- Custom (AVRO) `ReadFromBigQuery`. This is tested by almost all of the BQ-related PostCommits.
- Custom (JSON) `ReadFromBigQuery(use_json_exports=True)`. This is tested in a couple of the BQ-related PostCommits (note `BigQueryQueryToTableIT.test_new_types_json` checks all the special types including bytes, and datetime-related types).

> 4. Big Query tests are one of the major source of [postcommit flakiness](https://issues.apache.org/jira/issues/?jql=text~BigQuery%20AND%20summary~flaky%20AND%20status%3DOpen%20AND%20summary~Test)? What are the plans to address that?

These are all Write tests. The Read from BQ tests are less troublesome, likely due to the fact that the transform is less complex, and asserts are kept in the pipeline (vs BigQuery matchers). I'm happy to own all breakages related to thistest - though I expect few compared to the WriteToBQ issues.

pabloem:
Run Python 3.7 PostCommit

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -45,8 +45,8 @@
 may use some caching techniques to share the side inputs between calls in order
 to avoid excessive reading:::
 
-  main_table = pipeline | 'VeryBig' >> beam.io.Read(beam.io.BigQuerySource()
-  side_table = pipeline | 'NotBig' >> beam.io.Read(beam.io.BigQuerySource()
+  main_table = pipeline | 'VeryBig' >> beam.io.ReadFroBigQuery(...)
```
tvalentyn, line 48: Typo in From

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -78,6 +72,12 @@
 or a table. Pipeline construction will fail with a validation error if neither
 or both are specified.
 
+When reading from BigQuery using `BigQuerySource`, bytes are returned as
```
tvalentyn, line 75: Can we add some guidance when to use which?

tvalentyn:
Run Python 2 PostCommit

pabloem:
Run Python 2 PostCommit

pabloem:
Run Python 3.7 PostCommit

pabloem:
Run Python 3.7 PostCommit

pabloem:
Run Python 2 PostCommit

pabloem:
Run Python 2 PostCommit

pabloem:
Run Python 3.7 PostCommit

pabloem:
Run Python 3.7 PostCommit

pabloem:
Run Python 2 PostCommit

pabloem:
Run Python PreCommit

pabloem:
Run PythonDocker PreCommit

pabloem:

pabloem:
Run Python 2 PostCommit

pabloem:
Run Python 2 PostCommit

pabloem:
Run Python 2 PostCommit

pabloem:
Run Python 3.6 PostCommit

pabloem:
retest this please

pabloem:
retest this please

pabloem:
Run Python 3.7 PostCommit

pabloem:
Run Python 2 PostCommit

pabloem:
Run Python 3.7 PostCommit

pabloem:
Run Python PreCommit

pabloem:
Run Python 3.7 PostCommit

pabloem:
Run Python 2 PostCommit

pabloem:
Run Python 3.7 PostCommit

sdks/python/apache_beam/io/gcp/bigquery_read_it_test.py
robertwb: or `isinstance(v, (datetime.time, datetime.date))`

sdks/python/apache_beam/io/gcp/bigquery.py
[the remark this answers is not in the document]
robertwb: Assuming Avro is a superior format, I think we should drop support for json (but not right away). My preference would be to have this choice be orthogonal to the choice of types (and having the bytes default to be base64 encoded feels really wrong) so do think having a backwards compatibility parameter is preferable. (If it's just the date and bytes types, I would split this into two parameters: `use_base64_encoded_bytes` and `datedtime_type`.)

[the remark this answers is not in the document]
tvalentyn: > "having the bytes default to be base64 encoded feels really wrong" .
I think we changed bytes to be base64-encoded several years back, because Java SDK used that representation: https://github.com/apache/beam/blob/master/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryIO.java#L181-L182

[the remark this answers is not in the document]
lukecwik: Yes, the joys of matching what was done in Java years ago for the Dataflow service API.

pabloem:

pabloem:
Run Python 3.7 PostCommit

pabloem:
Run Python 3.7 PostCommit

pabloem:
Run Python 3.7 PostCommit

pabloem:
Run Python 2 PostCommit

pabloem:

sdks/python/apache_beam/io/gcp/bigquery.py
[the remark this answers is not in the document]
pabloem: Restarting this discussion after some rebasing and preparing...

I think I'd prefer to allow users to choose the export file format. This is what we do for writing to BQ in Java and Python. Allowing users to choose output type for specific column types would add overhead of (if avro and json_format_bytes: formatbytes; if avro and json_datetime_format: formatdatetime; if not avro and not json_format_bytes: formatbytes; ...).

We can discourage users from using JSON (AVRO is already the default) - and eventually stop supporting it if that makes sense. Thoughts?

tvalentyn:
@pabloem, would you mind squashing the commits that happened after the last
review iteration (but please don't squash earlier commits until review is
finalized) and resolving any comments that are already addressed? You can
also delete the "Run postcommit..." which would make  it easier to follow
up on remaining comment threads.

pabloem:
Run Python 3.5 PostCommit

pabloem:
Run Python 2 PostCommit

pabloem:
py2 failure is hdfs integration test

pabloem:
Run Python PreCommit

pabloem:
Run Python 2 PostCommit

pabloem:

sdks/python/apache_beam/io/gcp/big_query_query_to_table_it_test.py
```diff
@@ -254,11 +256,36 @@ def test_big_query_new_types(self):
...
         'use_standard_sql': False,
         'wait_until_finish_duration': WAIT_UNTIL_FINISH_DURATION_MS,
+        'use_json_exports': True,
         'on_success_matcher': all_of(*pipeline_verifiers)
     }
     options = self.test_pipeline.get_full_options_as_args(**extra_opts)
     big_query_query_to_table_pipeline.run_bq_pipeline(options)
 
+  @attr('IT')
+  def test_big_query_new_types(self):
```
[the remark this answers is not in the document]
pabloem: Thanks for pointing that out. I've renamed the test cases to keep coverage for the feature on the same test.

pabloem:
So, I think the last remaining issue was whether to have the `use_json_exports` flag or individual flags per columns. Are there other topics to discuss?

pabloem:
Run Python PreCommit

pabloem:
Run Python 3.7 PostCommit

pabloem:
Run Python 2 PostCommit

pabloem:
Run Python 3.7 PostCommit

pabloem:
@tvalentyn @chamikaramj thoughts?

tvalentyn:
I will defer to Cham on use_json_exports.

chamikaramj:
Can you provide some more context :) ?

chamikaramj:
Based on some offline discussion, I think it's enough to add a single flag here. We should try to minimize the number of flags we add (add flags is easy removing is hard).

The main question here I think is regarding preserving backwards compatibility for users who rely on old behavior. For this purpose a single flag is enough. Users who need to convert a given type from one form to the other can easily do that by adding a ParDo. If adding support for a certain type is beneficial to a lot of users and if there's significant user demand we can always add that later.

pabloem:
okay, in that case, the main points have been addressed. Any objections from merging? fyi @chamikaramj @tvalentyn @robertwb @kamilwu 

If no objections, I'll merge in a day or two.

sdks/python/apache_beam/io/gcp/bigquery.py
tvalentyn: Ping on this

sdks/python/apache_beam/io/gcp/bigquery.py
tvalentyn: Ping on this - do we need to guide users in any way which transform they should use? Is some of them experimental or to be deprecated?

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -732,11 +748,20 @@ def _export_files(self, bq):
...
-                                     bigquery_tools.FileFormat.JSON,
-                                     include_header=False)
+    if self.use_json_exports:
+      job_ref = bq.perform_extract_job([self.gcs_location],
+                                       job_id,
+                                       self.table_reference,
+                                       bigquery_tools.FileFormat.JSON,
+                                       include_header=False)
+    else:
+      job_ref = bq.perform_extract_job([self.gcs_location],
```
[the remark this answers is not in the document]
tvalentyn: Is this resolved?

sdks/python/apache_beam/io/gcp/bigquery.py
```diff
@@ -78,6 +72,12 @@
 or a table. Pipeline construction will fail with a validation error if neither
 or both are specified.
 
+When reading from BigQuery using `BigQuerySource`, bytes are returned as
+base64-encoded bytes. When reading via `ReadFromBigQuery`, bytes are returned
+as bytes without base64 encoding. This is due to the fact that ReadFromBigQuery
+uses Avro expors by default. To get base64-encoded bytes, you can use the flag
```
tvalentyn, line 78: nit: exports

CHANGES.md
```diff
@@ -58,6 +58,10 @@
 
 * Support for X source added (Java/Python) ([BEAM-X](https://issues.apache.org/jira/browse/BEAM-X)).
 * Support for reading from Snowflake added (Java) ([BEAM-9722](https://issues.apache.org/jira/browse/BEAM-9722)).
+* A new transform to read from BigQuery has been added: `apache_beam.io.gcp.bigquery.ReadFromBigQuery`. This transform
```
tvalentyn, line 61: To confirm - there is no changes in behavior for old sync, right?

sdks/python/apache_beam/io/gcp/bigquery.py
pabloem: done

CHANGES.md
pabloem: that's correct.

pabloem:
Run Python PreCommit

pabloem:
Run Portable_Python PreCommit

pabloem:

pabloem:
Run Python PreCommit

pabloem:
Run Python PreCommit

pabloem:
Run Python PreCommit

pabloem:
Run Python PreCommit

pabloem:
Run Python PreCommit

pabloem:

pabloem:
Run Python 2 PostCommit

pabloem:
Run Python 3.6 PostCommit

PR #11090, opened 2020-03-10
[BEAM-9470] :sdks:java:io:kinesis:test is flaky

jfarr:
This PR:
- Fixes a race condition in ShardReadersPoolTest.shouldCallRateLimitPolicy() that was causing intermittent failures.
- Modifies ShardReadersPool to ensure that rateLimitPolicy.onSuccess() will always be called when it should be.
r: @aromanenko-dev 

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/ShardReadersPool.java
+++ b/sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/ShardReadersPool.java
@@ -126,9 +126,21 @@ class ShardReadersPool {
   private void readLoop(ShardRecordsIterator shardRecordsIterator, RateLimitPolicy rateLimiter) {
     while (poolOpened.get()) {
       try {
-        List<KinesisRecord> kinesisRecords;
         try {
-          kinesisRecords = shardRecordsIterator.readNextBatch();
+          List<KinesisRecord> kinesisRecords = shardRecordsIterator.readNextBatch();
+          try {
+            for (KinesisRecord kinesisRecord : kinesisRecords) {
+              recordsQueue.put(kinesisRecord);
+              numberOfRecordsInAQueueByShard.get(kinesisRecord.getShardId()).incrementAndGet();
+            }
+          } finally {
+            // One of the paths into this finally block is recordsQueue.put() throwing
+            // InterruptedException so we should check the thread's interrupted status before
+            // calling onSuccess().
+            if (!Thread.currentThread().isInterrupted()) {
+              rateLimiter.onSuccess(kinesisRecords);
+            }
+          }
         } catch (KinesisShardClosedException e) {
           LOG.info(
               "Shard iterator for {} shard is closed, finishing the read loop",
@@ -143,11 +155,6 @@ class ShardReadersPool {
           readFromSuccessiveShards(shardRecordsIterator);
           break;
         }
-        for (KinesisRecord kinesisRecord : kinesisRecords) {
-          recordsQueue.put(kinesisRecord);
-          numberOfRecordsInAQueueByShard.get(kinesisRecord.getShardId()).incrementAndGet();
-        }
-        rateLimiter.onSuccess(kinesisRecords);
       } catch (KinesisClientThrottledException e) {
         try {
           rateLimiter.onThrottle(e);
--- a/sdks/java/io/kinesis/src/test/java/org/apache/beam/sdk/io/kinesis/ShardReadersPoolTest.java
+++ b/sdks/java/io/kinesis/src/test/java/org/apache/beam/sdk/io/kinesis/ShardReadersPoolTest.java
@@ -22,7 +22,6 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Matchers.any;
 import static org.mockito.Matchers.eq;
 import static org.mockito.Matchers.same;
-import static org.mockito.Mockito.atLeastOnce;
 import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.timeout;
 import static org.mockito.Mockito.times;
@@ -324,10 +323,11 @@ public class ShardReadersPoolTest {
       }
     }
 
-    verify(customRateLimitPolicy).onThrottle(same(e));
-    verify(customRateLimitPolicy).onSuccess(eq(ImmutableList.of(a, b)));
-    verify(customRateLimitPolicy).onSuccess(eq(singletonList(c)));
-    verify(customRateLimitPolicy).onSuccess(eq(singletonList(d)));
-    verify(customRateLimitPolicy, atLeastOnce()).onSuccess(eq(Collections.emptyList()));
+    verify(customRateLimitPolicy, timeout(TIMEOUT_IN_MILLIS)).onThrottle(same(e));
+    verify(customRateLimitPolicy, timeout(TIMEOUT_IN_MILLIS)).onSuccess(eq(ImmutableList.of(a, b)));
+    verify(customRateLimitPolicy, timeout(TIMEOUT_IN_MILLIS)).onSuccess(eq(singletonList(c)));
+    verify(customRateLimitPolicy, timeout(TIMEOUT_IN_MILLIS)).onSuccess(eq(singletonList(d)));
+    verify(customRateLimitPolicy, timeout(TIMEOUT_IN_MILLIS).atLeastOnce())
+        .onSuccess(eq(Collections.emptyList()));
   }
 }
--- a/sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/RateLimitPolicyFactory.java
+++ b/sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/RateLimitPolicyFactory.java
@@ -24,7 +24,9 @@ import org.joda.time.Duration;
 
 /**
  * Implement this interface to create a {@code RateLimitPolicy}. Used to create a rate limiter for
- * each shard.
+ * each shard. The factory will be called from multiple threads, so if it returns a singleton
+ * instance of RateLimitPolicy then that instance should be thread-safe, otherwise it should return
+ * separate RateLimitPolicy instances.
  */
 public interface RateLimitPolicyFactory extends Serializable {
 


aromanenko-dev:
retest this please

aromanenko-dev:
Please, fix Spotless issue

sdks/java/io/kinesis/src/test/java/org/apache/beam/sdk/io/kinesis/ShardReadersPoolTest.java
```diff
@@ -324,10 +325,15 @@ public void shouldCallRateLimitPolicy()
...
+    assertThat(capturedRecords).contains(
+        ImmutableList.of(a, b),
+        singletonList(c),
+        singletonList(d),
+        Collections.emptyList()
+    );
     verify(customRateLimitPolicy).onThrottle(same(e));
-    verify(customRateLimitPolicy).onSuccess(eq(ImmutableList.of(a, b)));
-    verify(customRateLimitPolicy).onSuccess(eq(singletonList(c)));
-    verify(customRateLimitPolicy).onSuccess(eq(singletonList(d)));
```
suztomo: Would you share your thought on in what situation (the order of concurrent operations) customRateLimitPolicy did not take `[d]` even when it takes `[a, b]` and `[c]`?

https://builds.apache.org/job/beam_PreCommit_Java_Commit/10277/testReport/junit/org.apache.beam.sdk.io.kinesis/ShardReadersPoolTest/shouldCallRateLimitPolicy/

> Argument(s) are different! Wanted:
> customRateLimitPolicy.onSuccess(
>    [d]
> );
> -> at org.apache.beam.sdk.io.kinesis.ShardReadersPoolTest.shouldCallRateLimitPolicy(ShardReadersPoolTest.java:330)

jfarr: @suztomo My thought is that Mockito is expecting to see these parameters in this order, which may not always be the case due to the 2 ShardReadersPool worker threads. My suggested fix is to rewrite the Mockito verifications so that they are definitely not dependent on order.

jfarr:
> Thanks! Please, fix Spotless issue and add `[BEAM-9470]` as a prefix to commit message.

@aromanenko-dev Sure, sorry about that. It's fixed now.

sdks/java/io/kinesis/src/test/java/org/apache/beam/sdk/io/kinesis/ShardReadersPoolTest.java
suztomo: Thank you for response. I still need your help here. Mockito is not supposed to check the order of the method calls (unless [configured to use inOrde](https://stackoverflow.com/questions/21901368/mockito-verify-order-sequence-of-method-calls)r). I checked its behavior with the following code:
Screen Shot 2020-03-11 at 12 03 34 AM

Do you know something special in this test case that makes mockito's `verify` order sensitive?

jfarr: Thanks, that's how I thought Mockito.verify() worked too but this made me doubt it. I didn't do any detailed testing, though, so thanks for taking the time to validate. I'm honestly at a loss for what could be causing intermittent failures then. All I can say is this version of the test is semantically equivalent and might not be subject to the same failure mode. I have not been able to reproduce the failure myself.

suztomo: How about writing one-off code that invokes the test, say, 100 times and count test failures?  We can verify whether your version fixes the flakiness.

I don’t expect that one-off code to be checked in, but attaching the observation and the code snippet to PR is helpful.

aromanenko-dev: I did run this test thousands of times locally (with `@RepeatedTest(n)` and some hacks to run it on Junit5) and I've managed to reproduce it only twice with the similar but a bit different error as it was on Jenkins.  

Jenkins error: https://gist.github.com/aromanenko-dev/f9714c3dd5a47a9439b9245f3ef7b04b
My local error (old test version): https://gist.github.com/aromanenko-dev/9a8acf265b96e010ad0ad1d09bb3ab92

The common thing there is that it starts to emit empty records list after it returned a list with `c` value.

Also, I did the same thing for new version of test and it's also reproducible (2 fails over 1000 runs):
My local error (new test version) 1st fail: https://gist.github.com/aromanenko-dev/faa4229a486117d7504e486282503009
My local error (new test version) 2nd fail: https://gist.github.com/aromanenko-dev/751cd007bb89a4f6e67f1ce9f9ef3d25

As you can see the actually the errors with old and new versions tests are similar. So, it seems that it's not related to `verify` using and more related to how every mock `ShardRecordsIterator` returns the records. For some reasons it starts to return just empty list.

suztomo: @aromanenko-dev Thank you for sharing your result.

jfarr: Yes, thanks Alexey. I was just working on something similar. Maybe the test is just overspecified then. I propose we change the validations to just:
```
verify(customRateLimitPolicy).onThrottle(any(KinesisClientThrottledException.class));
verify(customRateLimitPolicy, atLeastOnce()).onSuccess(any(List.class));
```
If you both agree I'll make that change.

suztomo: I disagree (I’m not a committer). We just revealed a bug, which is very hard to reproduce.

@aromanenko-dev 

> @RepeatedTest(10000) and some hacks to run it on Junit5

Would you share this?

jfarr: > Would you share this?
 +1, please

@suztomo What bug are you suggesting this exposes?

suztomo: bug: sometimes customRateLimitPolicy does not receive items, "[d]" in the test case.

jfarr: @suztomo You have a good point. The intention of this test is to validate that every time readNextBatch() returns a list of records onSuccess() gets called and every time it throws `KinesisClientThrottledException` onThrottle() gets called. I can see a couple of code paths where readNextBatch() could return records without onSuccess() getting called, namely if any exceptions are thrown along in here: https://github.com/apache/beam/blob/d62521f69ead4b58924043f041978d49e9beeb62/sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/ShardReadersPool.java#L146-L149

I haven't seen evidence that this is what's happening, but let's try my latest changes in ShardReadersPool and see if that fixes it. With this change it should not be possible for readNextBatch() to return without onSuccess() being called.

jfarr: @aromanenko-dev If you could share with us how you got @RepeatedTest working I'd be happy to retest.

aromanenko-dev: Sure, this patch is enough to run only this test `shouldCallRateLimitPolicy()` with Junit 5:
https://gist.github.com/aromanenko-dev/02c4b206bd7c64bec3ee7a4e3588a5b6
I didn't test other tests, so plz don't blame if something doesn't work =) Sorry for delay, I'm in a different time zone.

aromanenko-dev:
retest this please

aromanenko-dev:
retest this please

aromanenko-dev:
Run Java PreCommit

sdks/java/io/kinesis/src/test/java/org/apache/beam/sdk/io/kinesis/ShardReadersPoolTest.java
suztomo: @aromanenko-dev Thank you very much. I can now reproduce the problem.

image

Everyone is in different time zones.

aromanenko-dev: What I can guess for now is that `firstIterator` and `secondIterator` are executed in different threads (since `readLoop()` is called for every topic shard in the separate thread) and for some reasons **sometimes** one thread takes more priority over the second while running and the `secondIterator` doesn't have a chance to return "d" record since `shardReadersPool` got closed.
PS: Interesting that on Jenkins it's much easier reproducible than locally.

aromanenko-dev: I did some testing on my side and I think that the problem is with using the same mock of `RateLimitPolicy` for both `ShardRecordsIterator`s. For example, when failed assertion says that "d" record is absent in `capturedRecords`, actually I see (using debug output) that it was read with `shardRecordsIterator.readNextBatch()` in `readLoop()`.
Since, in case of using real `RateLimitPolicy` they will be created per `readLoop()` thread (right?), then we need to test it in the same way with mock policies.

jfarr: OK, I see what's happening here. It's basically a race condition. This loop will exit as soon as nextRecord() returns the last record:
https://github.com/apache/beam/blob/b0e7afbc3baf4ac17f7159cb8aa1fabe326be3e2/sdks/java/io/kinesis/src/test/java/org/apache/beam/sdk/io/kinesis/ShardReadersPoolTest.java#L320-L325

I believe that sometimes the test is hitting the verify() call for that last record in the split second between shardReadersPool putting the record into the blocking queue and calling onSuccess(). The fix is to use Mockito's timeout() to ensure that we wait long enough for that last onSuccess() call. It took me about 30,000 iterations to reproduce with the original code but I have run the new test code over 100,000 times without a failure.

I do still think it's a good idea to make these changes to ShardReadersPool to ensure that onSuccess() always gets called (as long as the thread wasn't interrupted).

jfarr: > Since, in case of using the real RateLimitPolicy in normal code, the instances of them will be created per every readLoop() thread (right?). So, we need to test it in the same way with mock policies.

It's possible for a custom RateLimitPolicyFactory to return a singleton instance of RateLimitPolicy, and I think that should be fine (as long as it's thread-safe). So maybe we shouldn't assume that they will always be different instances.

suztomo:
In the test case, the two iterators return the following data:

- firstIterator: `throw(e)`;  `[a, b]`;   `[]`;   `[]`;   `[]`;   `[]` ...
- secondIterator: `[c]`;   `[d]`;   `[]`;   `[]`;   `[]` ...

The test case waits for 5 elements from shardReadersPool:

```
    List<KinesisRecord> fetchedRecords = new ArrayList<>();
    while (fetchedRecords.size() < 4) {
      CustomOptional<KinesisRecord> nextRecord = shardReadersPool.nextRecord();
      if (nextRecord.isPresent()) {
        fetchedRecords.add(nextRecord.get());
      }
    }
```

The test's expectation was:

```
    verify(customRateLimitPolicy).onThrottle(same(e));
    verify(customRateLimitPolicy).onSuccess(eq(ImmutableList.of(a, b)));
    verify(customRateLimitPolicy).onSuccess(eq(singletonList(c)));
    verify(customRateLimitPolicy).onSuccess(eq(singletonList(d)));
```

My assumption: there's no guarantee that 2 elements from firstIterator and secondIterator are served evenly when 5 elements are consumed.

My suggestion: make the while loop until it confirms the expected 5 elements are in `fetchedRecords` variable.

aromanenko-dev:
retest this please

aromanenko-dev:
> The test case waits for 5 elements from shardReadersPool:

I think I disagree here - `fetchedRecords` will contain only 4 elements maximum. And they will always only "a", "b", "c" and "d" but maybe in different order. Empty records won't be returned by `shardReadersPool.nextRecord()`, however they will be passed to `rateLimiter.onSuccess()` as an argument (which is unnecessary I think). 
Let me explain how I understand this test. @jfarr , please correct me if I'm wrong. Actually, we need this loop running until `fetchedRecords.size() < 4` only to make sure that 2 reader threads did read all these 4 records. Than we assert that `rateLimiter` worked correctly during this since this is a real purpose of this test.

I think that new version of this test with timeout (last commit) should be fine. I tested it locally several times with 10k runs and no fails for the moment.

sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/ShardReadersPool.java
```diff
@@ -126,9 +126,21 @@ void startReadingShards(Iterable<ShardRecordsIterator> shardRecordsIterators) {
...
+            for (KinesisRecord kinesisRecord : kinesisRecords) {
+              recordsQueue.put(kinesisRecord);
+              numberOfRecordsInAQueueByShard.get(kinesisRecord.getShardId()).incrementAndGet();
+            }
+          } finally {
+            // One of the paths into this finally block is recordsQueue.put() throwing
+            // InterruptedException so we should check the thread's interrupted status before
+            // calling onSuccess().
+            if (!Thread.currentThread().isInterrupted()) {
+              rateLimiter.onSuccess(kinesisRecords);
```
aromanenko-dev: Do we really need to pass an empty list of records if it's happen to receive from `shardRecordsIterator.readNextBatch()`?

sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/ShardReadersPool.java
```diff
@@ -126,9 +126,21 @@ void startReadingShards(Iterable<ShardRecordsIterator> shardRecordsIterators) {
   private void readLoop(ShardRecordsIterator shardRecordsIterator, RateLimitPolicy rateLimiter) {
```
aromanenko-dev: Don't you think that `rateLimiter` should not be shared across different reader threads?

suztomo:
@aromanenko-dev You're right. `fetchedRecords` only contains 4 elements. My understanding for the test was incorrect.

sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/ShardReadersPool.java
jfarr: Yes, in order to rate limit correctly onSuccess() needs to be called after every successful getRecords() call even if it returned no records. Otherwise we would be right back to exceeding our API rate limit when the shard has no more records to process.

sdks/java/io/kinesis/src/main/java/org/apache/beam/sdk/io/kinesis/ShardReadersPool.java
jfarr: I don't see an inherent problem with it, as long as the rate limiter is thread-safe it can be shared. If you want I could add a note in the Javadoc for RateLimitPolicyFactory to indicate that it will be used from multiple threads, so if it returns a singleton instance of RateLimitPolicy then that instance should be thread-safe, otherwise it should return separate RateLimitPolicy instances.

aromanenko-dev: Yes, please add this into Javadoc to make it clear for users. Other than that we are fine, I think.

aromanenko-dev:
I think it's almost LGTM, just requires some Javadoc updates on using rate limiter in multiple threads.

jfarr:
> I think it's almost LGTM, just requires some Javadoc updates on using rate limiter in multiple threads.

Thanks @aromanenko-dev, that's done now. And thank you @suztomo for pushing me on this and keeping me honest.

aromanenko-dev:
retest this please

aromanenko-dev:
retest this please

aromanenko-dev:

aromanenko-dev:

aromanenko-dev:

aromanenko-dev:

aromanenko-dev:

aromanenko-dev:

PR #11092, opened 2020-03-10
[BEAM-9085] Fix performance regression in SyntheticSource on Python 3

kamilwu:
Beam was using `numpy.random.RandomState` during generation of Synthetic input, which, starting from NumPy 1.17, is considered as a legacy generator. This caused the slowdown on Python 3. The solution is using `random` module instead of Numpy on Python 3, and keep using Numpy on Python 2 until we deprecate Python 2 support.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/testing/synthetic_pipeline.py
+++ b/sdks/python/apache_beam/testing/synthetic_pipeline.py
@@ -42,9 +42,10 @@ import itertools
 import json
 import logging
 import math
-import random
 import struct
+import sys
 import time
+from random import Random
 
 import apache_beam as beam
 from apache_beam.io import WriteToText
@@ -62,6 +63,35 @@ try:
   import numpy as np
 except ImportError:
   np = None
+
+
+class _Random(Random):
+  """A subclass of `random.Random` from the Python Standard Library that
+  provides a method returning random bytes of arbitrary length.
+  """
+
+  # `numpy.random.RandomState` does not provide `random()` method, we keep this
+  # for compatibility reasons.
+  random_sample = Random.random
+
+  def bytes(self, length):
+    """Returns random bytes.
+
+    Args:
+      length (int): Number of random bytes.
+    """
+    n = length // 8 + 1
+    # pylint: disable=map-builtin-not-iterating
+    return struct.pack(
+        '{}Q'.format(n),
+        *map(self.getrandbits, itertools.repeat(64, n)))[:length]
+
+
+Generator = _Random
+
+# TODO: Remove this when Beam drops Python 2.
+if np is not None and sys.version_info.major == 2:
+  Generator = np.random.RandomState
 
 
 def parse_byte_size(s):
@@ -418,31 +448,26 @@ class SyntheticSource(iobase.BoundedSource):
       tracker = range_trackers.UnsplittableRangeTracker(tracker)
     return tracker
 
-  @staticmethod
-  def random_bytes(length, generator):
-    """Return random bytes."""
-    n = length // 8 + 1
-    # pylint: disable=map-builtin-not-iterating
-    return struct.pack(
-        '{}Q'.format(n),
-        *map(generator.getrandbits, itertools.repeat(64, n)))[:length]
-
   def _gen_kv_pair(self, generator, index):
     generator.seed(index)
-    rand = generator.random()
+    rand = generator.random_sample()
 
     # Determines whether to generate hot key or not.
     if rand < self._hot_key_fraction:
       # Generate hot key.
       # An integer is randomly selected from the range [0, numHotKeys-1]
       # with equal probability.
-      generator.seed(index % self._num_hot_keys)
-    bytes_ = self.random_bytes(self.element_size, generator)
-    return bytes_[:self._key_size], bytes_[self._key_size:]
+      generator_hot = Generator(index % self._num_hot_keys)
+      bytes_ = generator_hot.bytes(self._key_size), generator.bytes(
+        self._value_size)
+    else:
+      bytes_ = generator.bytes(self.element_size)
+      bytes_ = bytes_[:self._key_size], bytes_[self._key_size:]
+    return bytes_
 
   def read(self, range_tracker):
     index = range_tracker.start_position()
-    generator = random.Random()
+    generator = Generator()
     while range_tracker.try_claim(index):
       time.sleep(self._sleep_per_input_record_sec)
       yield self._gen_kv_pair(generator, index)
@@ -563,7 +588,8 @@ class SyntheticSDFAsSource(beam.DoFn):
           SyntheticSDFSourceRestrictionProvider())):
     cur = restriction_tracker.current_restriction().start
     while restriction_tracker.try_claim(cur):
-      r = np.random.RandomState(cur)
+      r = Generator()
+      r.seed(cur)
       time.sleep(element['sleep_per_input_record_sec'])
       yield r.bytes(element['key_size']), r.bytes(element['value_size'])
       cur += 1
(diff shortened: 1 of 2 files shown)


kamilwu:
R: @tvalentyn 

I did many tests and gathered the results in one single document: https://docs.google.com/document/d/1AegjUCc5w4B90_rvR8WAfIeL65PHUi4oGYkodgPS0o0/edit?usp=sharing. The previous solution (https://github.com/apache/beam/pull/10885) didn't work very well on big `element_size`, which was most probably the cause of failures and slowdowns.

Although this PR is still a bit slower than we have now using numpy 1.16 and Python 2.7, I'd rather avoid downgrading numpy via supplying an additional requirements.txt file to a Dataflow job. According to this document (https://numpy.org/neps/nep-0029-deprecation_policy.html), the support for numpy 1.16 will be dropped on Jan 13, 2021 - and this looks like a temporary solution.

kamilwu:
pylint complains about using `map`, I'll try to fix this tomorrow

tvalentyn:
Thanks, @kamilwu. I'll have to take a closer look - I see that this pr causes a 8x slowdown in one of the benchmarks we run internally. I don't know yet where it comes from.

kamilwu:
8x slowdown is huge. Can you share synthetic source parameters used in this benchmark?

kamilwu:
Looks like we have problem with "no space left" on apache-beam-jenkins-7 worker once again...

kamilwu:

kamilwu:
Run Python PreCommit

kamilwu:

kamilwu:
Run Python PreCommit

tvalentyn:
It is testing a pipeline with a side input:
 
 '--sideinput_num_records=1',
 '--sideinput_key_size=100000',
 '--maininput_key_size=100000',
 '--maininput_num_records=100000',

I reran the test, and the regression was on the order of 3x; running the tests again 5 times to see the difference.

tvalentyn:
Yes, looks like it is 3x slowdown on Py2, not 8x.

kamilwu:
Alright. This is still quite huge slowdown. @tvalentyn Do you have any other ideas how to resolve the issue? Our primary goal is to move all performance tests to Python 3 eventually.

tvalentyn:
Yes, I agree; we should be migrating the tests to Python 3. I think we'll pick the option that works well on Py3 and move or branch the generation on Py2 vs Py3, if we don't find one-fit-all solution. Afterall, the input generation should not be the slowest part of the pipeline.

tvalentyn:
What do you think about branching the generation based on Py version as a workaround until we deprecate Py2 support in Beam?

kamilwu:
I think branching the generation makes the most sense, since Py2 deprecation in Beam is closer than ever. I'll try to go back with a solution in the next week.

tvalentyn:
Also encountered a test internally that is slower on Py3 due to Synthetic source issue. Choosing an optimal implementation per python version should unblock this, so looking forward to your change.

tvalentyn:
Hi @kamilwu , do you still have time to work on this? Let me know if you need help. Thank you!

kamilwu:
Hi @tvalentyn, sorry for the delay. Changes are ready, I will push them in a second.

sdks/python/apache_beam/testing/synthetic_pipeline.py
```diff
@@ -61,6 +65,35 @@
...
+    n = length // 8 + 1
+    # pylint: disable=map-builtin-not-iterating
+    return struct.pack(
+        '{}Q'.format(n), *map(self.getrandbits, itertools.repeat(64,
+                                                                 n)))[:length]
+
+
+Generator = _Random
+
+# TODO: Remove this when Beam drops Python 2.
```
kamilwu, line 92: @tvalentyn Do we have a JIRA issue that accumulates all things that have to be done when Beam drops py2? I know there is a https://issues.apache.org/jira/browse/BEAM-5949, but this refers only to `__ne__`

kamilwu:
I reran some tests and it seems the performance on py2 is almost the same as before. @tvalentyn Could you check if slowdown on one of your internal tests still exists?

sdks/python/apache_beam/testing/synthetic_pipeline.py
tvalentyn: yes - https://issues.apache.org/jira/browse/BEAM-7372

tvalentyn: Commented on BEAM-5459, thank you!

sdks/python/apache_beam/testing/synthetic_pipeline.py
```diff
@@ -61,6 +65,35 @@
...
+    """Returns random bytes.
+
+    Args:
+      length (int): Number of random bytes.
+    """
+    n = length // 8 + 1
+    # pylint: disable=map-builtin-not-iterating
+    return struct.pack(
+        '{}Q'.format(n),
+        *map(self.getrandbits, itertools.repeat(64, n)))[:length]
```
tvalentyn, line 87: Since we don't need py2 compatibility anymore, consider using `to_bytes`. Here's an equivalent for chunk_size=8. It seems to be somewhat slower than current method (perhaps since I'm not using `map+repeat()`), but with larger `chuck_size`, seems to be more efficient. Large chunk size may be less efficient for short bytesequences.

```
chunk_size_bytes = 8 // TBD - larger chunks seem to improve performance.
chunk_size_bits = chunk_size_bytes * 8
num_chunks = length // chunk_size_bytes + 1

return b''.join([self.getrandbits(chunk_size_bits).to_bytes(chunk_size_bytes, sys.byteorder) for _ in range(num_chunks)])[:length]
```

If you decide to keep current implementation - please add a comment explaining the mechanics for readers not familiar with this code (we generate 8-byte stings, and then fit them into a representation of C++'s long-long, which also takes up 8 bytes).

sdks/python/apache_beam/testing/synthetic_pipeline.py
kamilwu: Thanks, this is what I was looking for

sdks/python/apache_beam/testing/synthetic_pipeline.py
kamilwu: How about not using chunks at all? I did some tests again and, surprisingly, it looks like we don't need them. 
Here's a test with a set of different chunk sizes:
`for CHUNK_SIZE in {4,8,32,64}; do python -m timeit -s "import random; import sys; chunk_size=$CHUNK_SIZE; len=10; num_chunks=len//chunk_size+1" 'b"".join([random.getrandbits(chunk_size * 8).to_bytes(chunk_size, sys.byteorder) for _ in range(num_chunks)])[:len]'; done
`

Results:
for len==10:
200000 loops, best of 5: 1.62 usec per loop
200000 loops, best of 5: 1.34 usec per loop
200000 loops, best of 5: 1.02 usec per loop
200000 loops, best of 5: 1.19 usec per loop

for len==1000:
5000 loops, best of 5: 87.7 usec per loop
5000 loops, best of 5: 50.7 usec per loop
20000 loops, best of 5: 16.7 usec per loop
20000 loops, best of 5: 11 usec per loop

And without chunks:
`python -m timeit -s "import random; import sys; len=10" 'random.getrandbits(len * 8).to_bytes(len, sys.byteorder)'`

for len==10:
1000000 loops, best of 5: 358 nsec per loop

for len==1000:
50000 loops, best of 5: 4.5 usec per loop

kamilwu: I'll push a commit with that so you can see how it looks in the code.

kamilwu:

kamilwu:

sdks/python/apache_beam/testing/synthetic_pipeline.py
tvalentyn: This is great. I really like how we simplified this code.

tvalentyn:
Thanks, @kamilwu!

PR #11122, opened 2020-03-13
[BEAM-9346] Improve the efficiency of TFRecordIO

piotr-szuberski:
I added the code requested in the Jira issue.
R: @mwalenia @kamilwu

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/io/WriteFiles.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/io/WriteFiles.java
@@ -22,7 +22,9 @@ import static org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Prec
 
 import com.google.auto.value.AutoValue;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
@@ -48,15 +50,15 @@ import org.apache.beam.sdk.io.fs.ResourceId;
 import org.apache.beam.sdk.options.PipelineOptions;
 import org.apache.beam.sdk.options.ValueProvider;
 import org.apache.beam.sdk.options.ValueProvider.StaticValueProvider;
+import org.apache.beam.sdk.transforms.Combine;
+import org.apache.beam.sdk.transforms.Combine.CombineFn;
 import org.apache.beam.sdk.transforms.DoFn;
 import org.apache.beam.sdk.transforms.Flatten;
 import org.apache.beam.sdk.transforms.GroupByKey;
 import org.apache.beam.sdk.transforms.PTransform;
 import org.apache.beam.sdk.transforms.ParDo;
-import org.apache.beam.sdk.transforms.Reify;
 import org.apache.beam.sdk.transforms.Reshuffle;
 import org.apache.beam.sdk.transforms.Values;
-import org.apache.beam.sdk.transforms.View;
 import org.apache.beam.sdk.transforms.WithKeys;
 import org.apache.beam.sdk.transforms.display.DisplayData;
 import org.apache.beam.sdk.transforms.windowing.BoundedWindow;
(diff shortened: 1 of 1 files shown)


mwalenia:
retest this please

piotr-szuberski:
retest this please

kamilwu:
@piotr-szuberski As a non-commiter, you are not allowed to trigger Jenkins tests on your own, unfortunately. Let me do this for you.

kamilwu:
retest this please

piotr-szuberski:
Ok, thank you - I was confused that the job has not started yet. Good to know!

mwalenia:
Run Java Precommig

mwalenia:
run java precommit

mwalenia:
Run TFRecord IOIT

mwalenia:
Run Java TFRecordIO Performance Test

mwalenia:
LGTM, thanks for the contribution!

aaltay:
If this is a common pattern (another one from: https://github.com/apache/beam/blob/master/sdks/java/core/src/main/java/org/apache/beam/sdk/io/WriteFiles.java#L412), should this be abstracted out to happen in a single place?

/cc @chamikaramj @kennknowles

sdks/java/core/src/main/java/org/apache/beam/sdk/io/WriteFiles.java
```diff
@@ -410,13 +412,44 @@ private GatherResults(Coder<ResultT> resultCoder) {
       } else {
         // Pass results via a side input rather than reshuffle, because we need to get an empty
         // iterable to finalize if there are no results.
-        return input
-            .getPipeline()
-            .apply(Reify.viewInGlobalWindow(input.apply(View.asList()), ListCoder.of(resultCoder)));
+        return input.apply("ToList", Combine.globally(new ToListCombineFn<>()));
```
chamikaramj: Seems like this was applied to all file-based IO not just TFRecordIO as mentioned in the PR description. Have we done enough experiments to make sure this won't have unintended adverse performance consequences in batch/streaming ? For example, what if the list does not fit in memory ?

Apologies if this was already discussed in the dev list.

cc: @robertwb @lukecwik @iemejia

lukecwik: I believe this could be a regression based upon the above comment a few lines higher then where the code was changed:
```
// Pass results via a side input rather than reshuffle, because we need to get an empty
// iterable to finalize if there are no results.
```
The combine globally in the global window will produce a default output in the global window but will throw an exception if used against non-global window PCollections.

Performance wise this is likely an improvement overall since runners typically do a better job with GBK then with side inputs.

sdks/java/core/src/main/java/org/apache/beam/sdk/io/WriteFiles.java
```diff
@@ -410,13 +412,44 @@ private GatherResults(Coder<ResultT> resultCoder) {
...
       }
     }
   }
 
+  public static class ToListCombineFn<ResultT>
```
lukecwik: nit: This should have been private class since its an implementation detail and didn't need to expose this to users of the WriteFiles transform.

piotr-szuberski: You're right, I just copy pasted the code from Jira. It was my first PR in Beam and I wanted to get the workflow without digging the code proposed.

sdks/java/core/src/main/java/org/apache/beam/sdk/io/WriteFiles.java
chamikaramj: Thanks. Agree with Luke. Combine globally has shuffle (GBK) inside hence breaks above statements. This could be a regression at least for Dataflow when there are no outputs. We should try following cases.

(1) Dataflow (and possibly other runners may have similar regressions ?) with an empty output. 
(2) Writing using a non-global window while WriteFiles.withWindowedWrites() not set.

Luke, does that make sense ? Anything else to try out to make sure there's no regression here ?

chamikaramj: What's the verdict here ? I suggest we revert this and introduce as an optional change since this is a potential regression for Dataflow runner.

chamikaramj: Created https://issues.apache.org/jira/browse/BEAM-9734 to make sure that this does not get into 2.21.0

piotr-szuberski: Imo just a revert would be better - giving an option to something that is not validated and can have unexpected adverse effects would result in user's confusion

mwalenia: I'm not sure what's the verdict right now. I think we should revert this and discuss further steps with repo free of regressions. Tagging @lukecwik for input about leaving the optimization for global windows. @chamikaramj how do you think the optional change could be applied? A runtime flag for the runner?

lukecwik: You could always apply the trigger if it is guaranteed to produce output by checking what the trigger definition is. I don't think you need a flag.

I'm not sure how much of a holistic picture of the whole WriteFiles transform was done to see if there is a better way to structure the transforms.

chamikaramj: Thanks for reverting.

My suggestion was to fork the transform expansion to use GlobalCombine when specifically requested through a transform parameter (for example see how we fork TextIO.Read expansion based on withHintMatchesManyFiles()).

https://github.com/apache/beam/blob/master/sdks/java/core/src/main/java/org/apache/beam/sdk/io/TextIO.java#L371
https://github.com/apache/beam/blob/master/sdks/java/core/src/main/java/org/apache/beam/sdk/io/TextIO.java#L401

I agree that the more flags we add the more confusing it becomes for users. So we should make sure that there are significant advantages in adding this option through performance benchmarking.

PR #11151, opened 2020-03-18
[BEAM-9468]  Hl7v2 io

jaketf:
Add HL7v2IO connector for the Google Cloud Healthcare API.
Initial work for these connectors is being prioritized based to support various 
customer use cases for HL7v2 -> FHIR ETL pipelines.

Healthcare clients are extremely concerned with not losing data in their pipelines and want dead letter queues for these sources / sinks.

With this in mind the `HL7v2IO.Read` is designed as a mini pipeline that:
  - Starts with an arbitrary `PTransform` to populate a `PCollection` of message IDs. This was chosen to give flexibility between reading the HL7v2 pubsub notifications, reading an entire HL7v2 store (with optional filter) or using some other method (e.g. a hand prepared subset of message IDs uploaded to GCS and read w/ TextIO). In the future we can gain efficiency here when the Healthcare API exposes a batch read method where we can batch up requests rather than making a separate call per element.
  - Then a DoFn that attempts to fetch the actual message contents from the HL7v2 store. Successfully fetched messages will be written to `HL7v2IO.Read.OUT` tag and Failures will be written to a `HLv2IO.Read.DEAD_LETTER` tag. The "expected" failure case would be attempting to fetch a message ID that doesn't exist.

Similarly, the `HL7v2IO.Write` is designed as a mini pipeline that 
  - starts with a `PTransform` that attempts to ingest a `PCollection` of `Messages`
  - then applies the `this.getDeadLetterPTransform()` to the `PCollection` of messages that failed to ingest. This gives the customer the ability to capture dead letters in a system of their chosing (e.g. gcs, bq, pubsub) based on their needs.

- [HL7v2](https://cloud.google.com/healthcare/docs/how-tos/hl7v2-messages):
  - Unbounded Read: Uses PubsubIO to read [notification subscription](https://cloud.google.com/healthcare/docs/how-tos/pubsub#hl7v2_messages)
  - Bounded Read: DoFn to get message IDs using the [Messages.List](https://cloud.google.com/healthcare/docs/reference/rest/v1beta1/projects.locations.datasets.hl7V2Stores.messages/list) REST API method

Future Work will include a very similar IO transform for FHIR store:
- [FhirIO](https://cloud.google.com/healthcare/docs/how-tos/fhir-bundles)
  - Write use the [executeBundle](https://cloud.google.com/healthcare/docs/reference/rest/v1beta1/projects.locations.datasets.fhirStores.fhir/executeBundle) REST API method to execute transactions on the FHIR Store
  - Read with [read](https://cloud.google.com/healthcare/docs/reference/rest/v1beta1/projects.locations.datasets.fhirStores.fhir/read) / [search](https://cloud.google.com/healthcare/docs/reference/rest/v1beta1/projects.locations.datasets.fhirStores.fhir/search) / from [notification subscription](https://cloud.google.com/healthcare/docs/how-tos/pubsub#fhir_resources)

# Linkage Check
([log](https://gist.github.com/jaketf/2f5363a00f4469d066a482a9fc0f77e2))

------------------------

--- | --- | --- | --- | ---


jaketf:
@lastomato FYI Moved this to it's own PR.
Once we get feedback and iterate on this with beam reviewer, I can open a PR for FHIRIO.

jaketf:
R: @pabloem

pabloem:
retest this please

pabloem:
("retest this please" is a code phrase for jenkins to run precommits on the PR)

pabloem:
since your change is a Java change, you don't need to worry about the Python errors for now. They seem to come from upstream.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
```diff
@@ -0,0 +1,620 @@
...
+ *
+ * pipeline.run();
+ *
+ * }***
+ * </pre>
+ */
+public class HL7v2IO {
+  // TODO add metrics for failed records.
+
+  private static Read.Builder read(PTransform<PBegin, PCollection<String>> messageIDTransform) {
```
pabloem: Under this interface, you would do something like this:
```
Pipeline p = ....
PCollectionTuple pct = p.apply(HL7v2IO.read(PubSubIO.read().fromTopic(....)));
```

This is a little odd. A more natural way of doing this would be something like this:

```
Pipeline p = ....
PCollection<String> messageIds = p.apply(PubSubIO.read().fromTopic(....));

HL7v2IO.Read.Result result = messageIds.apply(HL7v2IO.readAll());
```

This way, `HL7v2IO.Read` takes in a `PCollection<String>` containing the message IDs, instead of having a PTransform be a parameter, which is not commonly done elsewhere.

This is similar to how FileIO transforms work. Check out the functions `match`, `matchAll`, `readMatches`. https://beam.apache.org/releases/javadoc/2.19.0/index.html?org/apache/beam/sdk/io/FileIO.html

jaketf:
@lastomato
Thoughts on how we could rework the HCAPI client to get around this failing CI Check?
https://builds.apache.org/job/beam_PreCommit_Java_Commit/10418/testReport/org.apache.beam.sdk.io.gcp/GcpApiSurfaceTest/testGcpApiSurface/

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
```diff
@@ -0,0 +1,620 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * {@link HL7v2IO} provides an API for reading from and writing to <a
+ * href="https://cloud.google.com/healthcare/docs/concepts/hl7v2">Google Cloud Healthcare HL7v2 API.
+ * </a>
+ *
+ * <p>Read HL7v2 Messages are fetched from the HL7v2 store based on the {@link PCollection} of of
+ * message IDs {@link String}s produced by the {@link AutoValue_HL7v2IO_Read#getMessageIDTransform}
+ * as {@link PCollectionTuple}*** containing an {@link HL7v2IO.Read#OUT} tag for successfully
```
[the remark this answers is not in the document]
jaketf: Agreed. This is cleaner, I never thought of extending `POutput`, thanks for the tip!

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
jaketf: refactored to implement both  `HL7v2IO.[Read,Write].Result`

pabloem:
let me know if this is ready for review or not

pabloem:
retest this please

pabloem:
retest this please

jaketf:
@pabloem sorry for the rage commits today.
This is now ready for review.

jaketf:
Open Questions:
1. Should we remove adaptive throttling?
    - Seems that we're using retries in the client request initializer and right now a "bad record" will slow down the Read / Write (even though the error has nothing to do with the HL7v2 store being overwhelmed). Originally we wanted to be safe with overwhelming QPS on the HL7v2 store in batch scenarios.
1. Should we add more to the `HealthcareIOError`?
    - Add (processing time) Timestamp?
    - Add a convenience DoFn `HealthcareIOErrrorToTableRowFn` to ease writing deadletter queue to BigQuery.
1. Would it me more useful to expose an error rate metric than an error count?

pabloem:
retest this please

pabloem:
retest this please

pabloem:
retest this please

pabloem:
retest this please

jaketf:
Future improvements:
- Currently this uses the alpha API because the motivating use case of HL7v2 -> FHIR mapping which requires the `schematizedData` field (not yet available in beta API).
- For backwards compatibility the `ListHL7v2MessageIDs` always uses Messages.List to get message IDs and relies on the the `HL7v2IO.Read` to fetch the message contents. This provides a consistent flow for real-time and batch and provides backwards compatibility. However, in the alpha API Messages.List returns the actual message contents (rather than just the message IDs). This leads to us reading the messages contents twice for alpha HL7v2 stores. I suggest we address optimizing this "double fetch" in a future PR as the alpha API stabilizes.

pabloem:
retest this please

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
```diff
@@ -0,0 +1,658 @@
...
+    private final List<String> hl7v2Stores;
+    private final String filter;
+
+    /**
+     * Instantiates a new List HL7v2 message IDs with filter.
+     *
+     * @param hl7v2Stores the HL7v2 stores
+     * @param filter the filter
+     */
+    ListHL7v2MessageIDs(List<String> hl7v2Stores, String filter) {
```
brianlucier: It might be worth supporting a the extra options on the new list API (view, and orderBy).  In particular, the view can return the full message payload right in the list, thereby saving a future retrieval step.

jaketf: Thanks for bringing this up.
Agreed on supporting getting the messages from the new list API to save the extra retrieval.

IMO this DoFn should remain as is to support backwards compatibility with the beta list api (which only has message IDs in the response) through a similar code path to the real time use cases which would subscribe to pubsub notifications (which also only have the message IDs and require retrieval).

@pabloem @lastomato thoughts on how we should best handle the two different APIs minimizing calls where possible?

I see two possible strategies:
1. Rely on user to use a different PTransform depending on which API version they are trying to hit.
   - For old api requiring "get" after list: `HL7v2IO.ListHL7v2MessageIDs` -> `HL7v2IO.Read`(current path).
   - Add a new PTransform for new api (where message contents are in the list): `HL7v2IO.V2.Read extends PTransform<PBegin, HL7v2Message>`
1. Restructuring so we can support both API responses with a single composite PTransform that follows this logic:
   - `ListMessageFn` makes the message list call and outputs to two `PCollection`s with the following logic:
        - IF response contains the `hl7V2Messages` key and either output the messages to the `PCollection<HL7v2Message>` with a `V2` tuple tag
        - ELSE  output the messageIDs `PCollection<String>` with a `V1` tuple tag
   - On the `V1` tagged PCollection apply `HL7v2IO.Read` to produce a `PCollection<HL7v2Message>`
   - `Flatten` the two PCollection<HL7v2Message> (one of which will always in theory be empty) to form your ultimate output collection

My thoughts of trade-offs are:
1. Pros: Simpler execution path, simpler to implement (strictly additive) given work already done. Cons: User has to know what they're doing and need to document / maintain both paths
1. Pros: Simpler interface for users Cons: more complicated execution path, heavier lift.

I'm torn on which would be preferable and would love input on which approach others think is cleanest before implementing. (Or suggestions to something I haven't thought of!)

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HttpHealthcareApiClient.java
```diff
@@ -0,0 +1,452 @@
...
+      }
+
+      @Override
+      public List<String> next() throws NoSuchElementException {
+        try {
+          ListMessagesResponse response = makeListRequest(client, hl7v2Store, filter, pageToken);
+          this.isFirstRequest = false;
+          this.pageToken = response.getNextPageToken();
+          List<String> msgs = response.getMessages();
+          if (msgs == null && response.get("hl7V2Messages") != null) {
```
jaketf: Note, this `ArrayMap<String, String>` is basically the `Message` model.
If you use the alpha API to query an beta HL7v2 store, there will be a single key `name` with the message ID value.
If you use the alpha API to query an alpha HL7v2 store, there will be all of the keys in the `Message` model.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HttpHealthcareApiClient.java
```diff
@@ -0,0 +1,452 @@
...
+    ingestMessageRequest.setMessage(msg);
+    return client
+        .projects()
+        .locations()
+        .datasets()
+        .hl7V2Stores()
+        .messages()
+        .ingest(hl7v2Store, ingestMessageRequest)
+        .execute();
+  }
```
brianlucier: Does the client handle quota-based retries on 429 responses?  Or is left to the caller to handle retries?

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HttpHealthcareApiClient.java
brianlucier: This abstract modeling of the Message would work just fine.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
brianlucier: As a user -- I'll know what version of the V2 store I am using and likely am never switching back and forth.  For instance, I would just ignore the beta PTransform and use the alpha one.  Seems like that is not a large user burden.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
```diff
@@ -0,0 +1,658 @@
...
+       */
+      @ProcessElement
+      public void writeMessages(ProcessContext context) {
+        HL7v2Message msg = context.element();
+        final int throttleWaitSeconds = 5;
+        long startTime = System.currentTimeMillis();
+        Sleeper sleeper = Sleeper.DEFAULT;
+        switch (writeMethod) {
+          case BATCH_IMPORT:
+            throw new UnsupportedOperationException("The Batch import API is not available yet");
```
brianlucier: I like that the switch is here... looking fwd to it :)

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
```diff
@@ -0,0 +1,658 @@
...
+              if (throttler.throttleRequest(startTime)) {
+                LOG.info("Delaying request due to previous failures.");
+                this.throttledSeconds.inc(throttleWaitSeconds);
+                sleeper.sleep(throttleWaitSeconds * 1000);
+                this.throttler.successfulRequest(startTime);
+                this.successfulHL7v2MessageWrites.inc();
+              }
+              client.ingestHL7v2Message(hl7v2Store, msg.toModel());
+            } catch (Exception e) {
+              failedMessageWrites.inc();
```
brianlucier: I believe the reason for failure may be relevant.  a 429 would want a certain number of retries, where a 401 or 403 may need to refresh the credentials, where another error is likely a hard stop.

jaketf:
Ok an updates here from an internal thread w/ API team.

1. Message.List returning message contents is available in beta API with the view parameter.
1. Schematized Data should be in next beta release roughly in ~2 weeks.
1. right now the sink is outputting schematized data json wrapped in "{data=<actual_valid_json>}" 

In light of these I will do the following refactors: 
1. how we batch read from to always avoid the double get. This will make it a completely parallel code path than the real-time path but I think that's ok.
1. refactor to use beta client library
1. I'll strip out that `{data=}` wrapper to make this easier for users.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
jaketf: all retryable errors will be retried by the request initializer in the http client.

I think adding the throttling was a premature optimization on my part. We should remove adaptive throttling entirely and test this at load and see if throttling is necessary.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HttpHealthcareApiClient.java
jaketf: yes [here](https://github.com/apache/beam/pull/11151/files/cac03ec30c77e6bd45954e659b875d0378370892#diff-3ab3eba677d52f20f1a688b1017adf84R274) we extend [`RetryHttpRequestInitializer`](https://beam.apache.org/releases/javadoc/2.13.0/org/apache/beam/sdk/extensions/gcp/util/RetryHttpRequestInitializer.html)

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
```diff
@@ -0,0 +1,636 @@
...
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.ImmutableMap;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * {@link HL7v2IO} provides an API for reading from and writing to <a
+ * href="https://cloud.google.com/healthcare/docs/concepts/hl7v2">Google Cloud Healthcare HL7v2 API.
+ * </a>
+ *
+ * <p>Read
```
jaketf: @brianlucier PTAL at this updated doc string. 
it describes my latest change in ba9d023 to avoid the double get whenever we reading a whole HL7v2Store with Messages.List API by adding the view=FULL param.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
jaketf: This can be resolved because beta API also supports view now.
Refactored to serve this use case w/ a separate PTransform and updated docstring.

pabloem:
retest this please

pabloem:
retest this please

lastomato:
Sorry for the delay. Please see my comments inline.

> Open Questions:
> 
> 1. Should we remove adaptive throttling?
I think it is fine to remove it since the API has quota enabled by default, and retry logic is in place.
>    
>    * Seems that we're using retries in the client request initializer and right now a "bad record" will slow down the Read / Write (even though the error has nothing to do with the HL7v2 store being overwhelmed). Originally we wanted to be safe with overwhelming QPS on the HL7v2 store in batch scenarios.
> 2. Should we add more to the `HealthcareIOError`?
The second point below would be very helpful, but I am fine with adding it to next PR (if that's easier).
>    
>    * Add (processing time) Timestamp?
>    * Add a convenience DoFn `HealthcareIOErrrorToTableRowFn` to ease writing deadletter queue to BigQuery.
> 3. Would it be more useful to expose an error rate metric than an error count?
This functionality is probably already provided by services like stackdriver, we might want to wait until there is a concrete use case.

pabloem:
retest this please

pabloem:
I'm starting to take a good look at this now.

pabloem:
in case it's not obvious: feel free to ignore failures in unrelated test suites (e.g. communitymetrics)

jaketf:
Next Steps (based on offline feed):

- [x] Improve API for users:
    - [x] Add static methods for common patterns with `ListHL7v2Messages` 
    - [x] Add `ValueProvider<String>` support to ease use in the DataflowTemplates
        - [x] `ListHL7v2Messages` (hl7v2Store and filter)
        - [x] `Write` (hl7v2Store)
- [ ] "standardize" integration tests
    - [ ] Remove hard coding of my HL7v2Store / project in integration tests.
    - [ ] Add HL7v2 store to Beam GCP IT project

jaketf:
@pabloem PTAL.
I've addressed all outstanding feedback and setup integration testing infrastructure

Note on integration tests:
Integration tests seem to be passing.
Created the following resources
- [`gs://temp-storage-for-healthcare-io-tests`](https://pantheon.corp.google.com/storage/browser/temp-storage-for-healthcare-io-tests/?forceOnBucketsSortingFiltering=false&project=apache-beam-testing)
    - will house temp files for FHIR Import in upcoming FhirIO PR
- [`projects/apache-beam-testing/locations/us-central1/datasets/apache-beam-integration-testing`](https://pantheon.corp.google.com/healthcare/browser/locations/us-central1/datasets/apache-beam-integration-testing/datastores?project=apache-beam-testing) which houses ephemeral HL7v2 store per test class (and later ephemeral FHIR store per class.)

pabloem:
retest this please

pabloem:
retest this please

jaketf:
@pabloem 
Questions:

1. How could / should we add this to this [page for built in connectors](https://beam.apache.org/documentation/io/built-in/)? Are there other documentation things that I might have missed? 
1. Should we file jira issues or buganizer  for future work, notably:
    - Using Batch import / export for read when that api is available
    - Updating to newer version of client library (coming soon, which will simplify the hack-y handling of schematized data in this PR)
1. What is the feasibility / process for back-porting this to prior beam releases for expediting use by users who want to use an already released version e.g. 2.19 ?

pabloem:
Run Java PostCommit

pabloem:
Run Java PostCommit

pabloem:
Java PostCommit is probably our slowest suite of all, but once it runs we should be good to go

brianlucier:
> 3\. What is the feasibility / process for back-porting this to prior beam releases for expediting use by users who want to use an already released version e.g. 2.19 ?

bumping this question.  would like to know timing on when I can see it in a released version... either backport, or a new one.

lukecwik:
> > 3. What is the feasibility / process for back-porting this to prior beam releases for expediting use by users who want to use an already released version e.g. 2.19 ?
> 
> bumping this question. would like to know timing on when I can see it in a released version... either backport, or a new one.

You would need to get support from the community to do the backport and release which seems unlikely based upon this being a new feature (vs a critical bug/security issue) and people are constrained time wise due to COVID-19.

Once this is merged it would be part of the subsequent release, the branch cut dates are here:  https://calendar.google.com/calendar/embed?src=0p73sl034k80oob7seouanigd0%40group.calendar.google.com (with 2.22.0 scheduled on May 20th)

buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy
```diff
@@ -454,6 +455,7 @@ class BeamModulePlugin implements Plugin<Project> {
         google_api_services_clouddebugger           : "com.google.apis:google-api-services-clouddebugger:v2-rev20191003-$google_clients_version",
         google_api_services_cloudresourcemanager    : "com.google.apis:google-api-services-cloudresourcemanager:v1-rev20191206-$google_clients_version",
         google_api_services_dataflow                : "com.google.apis:google-api-services-dataflow:v1b3-rev20190927-$google_clients_version",
+        google_api_services_healthcare              : "com.google.apis:google-api-services-healthcare:v1alpha2-rev20190901-$google_cloud_healthcare_version",
```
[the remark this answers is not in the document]
jaketf: Good question!
Please let me know if I'm misunderstanding but I believe that is [set to 1.30.3](https://github.com/apache/beam/blob/master/buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy#L379)
And looking at this [repo for healthcare client](https://github.com/googleapis/google-api-java-client-services/tree/master/clients/google-api-services-healthcare/v1beta1) there doesn't seem to be anything past 1.30.1 (which was recently regenerated).

Also FYI you're commenting on an outdated commit we are now using beta client lib.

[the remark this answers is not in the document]
jaketf: However the generated README has version listed as v1beta1-rev20200327-1.30.9
so maybe I am misunderstanding.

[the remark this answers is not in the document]
jaketf: It'd make this PR cleaner to use this most recent version because [this commit adds `SchematizedData`](https://github.com/googleapis/google-api-java-client-services/commit/48db535eb30fea17c615f9951d7afad6f3ea72b4#diff-89048b80fe6bcee9b0aade84024aea66).

which I am currently around [here](https://github.com/apache/beam/pull/11151/files#diff-2173a334dc15f8b7dc3257efeec64504R27).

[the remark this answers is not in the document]
jaketf: @lukecwik Is there a reason this needs to be `google_clients_version` / 1.30.3?
I've just tried testing with that version and it breaks most usage of HCAPI.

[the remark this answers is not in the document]
lukecwik: It doesn't have to be 1.30.3, it could be any later version. The purpose of them being the same is to ensure that GCP dependencies are compatible with each other so the healthcare dep doesn't break other GCP deps and vice versa.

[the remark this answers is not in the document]
jaketf: I'm rather new to these generated clients and beam.
Is there any process this community has followed in the past for finding a suitable version in situations like this?
Seems like a lot for this PR to make this rather global upgrade.
Naturally if I just flip google_clients_version to 1.30.9 it breaks  a lot:
```
Could not determine the dependencies of task ':sdks:java:harness:shadowJar'.
> Could not resolve all dependencies for configuration ':sdks:java:harness:runtimeClasspath'.
   > Could not find com.google.apis:google-api-services-cloudresourcemanager:v1-rev20191206-1.30.9.
     Required by:
         project :sdks:java:harness > project :sdks:java:extensions:google-cloud-platform-core
   > Could not find com.google.apis:google-api-services-storage:v1-rev20191011-1.30.9.
     Required by:
         project :sdks:java:harness > project :sdks:java:extensions:google-cloud-platform-core
   > Could not find com.google.oauth-client:google-oauth-client:1.30.9.
   > Could not find com.google.apis:google-api-services-storage:v1-rev20191011-1.30.9.
     Required by:
         project :sdks:java:harness > project :sdks:java:extensions:google-cloud-platform-core > com.google.cloud.bigdataoss:gcsio:2.0.0
         project :sdks:java:harness > project :sdks:java:extensions:google-cloud-platform-core > com.google.cloud.bigdataoss:util:2.0.0
   > Could not find com.google.oauth-client:google-oauth-client:1.30.9.
     Required by:
         project :sdks:java:harness > project :sdks:java:extensions:google-cloud-platform-core > com.google.cloud.bigdataoss:gcsio:2.0.0
         project :sdks:java:harness > project :sdks:java:extensions:google-cloud-platform-core > com.google.cloud.bigdataoss:util:2.0.0
   > Could not find com.google.oauth-client:google-oauth-client-java6:1.30.9.
     Required by:
         project :sdks:java:harness > project :sdks:java:extensions:google-cloud-platform-core > com.google.cloud.bigdataoss:gcsio:2.0.0
         project :sdks:java:harness > project :sdks:java:extensions:google-cloud-platform-core > com.google.cloud.bigdataoss:util:2.0.0
   > Could not find com.google.oauth-client:google-oauth-client-java6:1.30.9.
     Required by:
         project :sdks:java:harness > project :sdks:java:extensions:google-cloud-platform-core > com.google.cloud.bigdataoss:gcsio:2.0.0 > com.google.api-client:google-api-client-java6:1.30.9
* 
```

seems like our culprits are 
- [ ] `google-oauth-client`
- [ ] `google-oauth-client-java6`
- [ ] `google-api-services-storage`

If I go in and try to massage the revision numbers to make this work how will I ensure that all those reliant on those deps are still working? Is that covered in pre/post commit ITs?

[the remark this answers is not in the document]
lukecwik: https://cwiki.apache.org/confluence/display/BEAM/Dependency+Upgrades

Combining the linkage checker with a bunch of precommits and postcommits gives us the confidence that the change is safe.

The linkage checker is a tool which finds linkage errors by scanning the classpath and finding missing classes/methods. You run the linkage checker on all modules before and after your change and show that the linkage errors reduced or stayed the same. Any new linkage errors need to be followed up and explained why they are safe.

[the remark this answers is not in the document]
lukecwik: Typically knowing which versions you can try is very hard without a lot of experience with a specific set of dependencies (e.g. knowing a lot about jackson echo system vs gRPC vs ...).

You should find a google-cloud-bom and matching libraries-bom that contain the healthcare dep you want to add. That should give you hints as to what are all the versions of things that work together. You will still need to perform the linkage checker analysis and run the tests as described above.

pabloem:
retest this please

pabloem:
retest this please

jaketf:
@lukecwik I ran the linkeage check "as-is" ([log](https://gist.github.com/suztomo/3178f298bd315e2cc3897b5528a961b8)) and IIUC this has no issues.

Is this acceptable or is it a hard requirement that I update google_clients_version?

lukecwik:
> @lukecwik I ran the linkeage check "as-is" ([log](https://gist.github.com/suztomo/3178f298bd315e2cc3897b5528a961b8)) and IIUC this has no issues.
> 
> Is this acceptable or is it a hard requirement that I update google_clients_version?

This is a hard requirement. Either we are using the google_clients_version of shared deps or google_healthcare_clients_version for deps. We can't be using both.

lukecwik:
Note that your dependency report seems to be showing that you should be able to update to the version needed by the healthcare lib.

jaketf:
@lukecwik that's what I thought to but when I do that I get the output in [this comment](https://github.com/apache/beam/pull/11151#discussion_r406551322)

lukecwik:
> @lukecwik that's what I thought to but when I do that I get the output in [this comment](https://github.com/apache/beam/pull/11151#discussion_r406551322)

A lot of google libraries have three parts to the version string (API major version, API minor version, shared lib version)

You have to find and upgrade the client libraries to versions that contain the 1.30.9 shared lib version like:
google-api-services-cloudresourcemanager:v1-rev20191206-1.30.4 becomes google-api-services-cloudresourcemanager:v1-rev20200311-1.30.9

Best way to find matching versions is to look at the index of the maven repo for each artifact (e.g.):
https://repo1.maven.org/maven2/com/google/apis/google-api-services-cloudresourcemanager/

Find ones that have the same API major version and any newer API minor version with the shard lib version being 1.30.9.

Major version changes are breaking API changes, minor version increases are meant to be safe to upgrade and shared lib version just needs to be the same across all versions.

jaketf:
Thanks for the pointer!
I will give that a try.

jaketf:
@lukecwik 
some progress I was able to update storage / cloudresourcemanager that way.

However, in
https://repo1.maven.org/maven2/com/google/oauth-client/google-oauth-client/
https://repo1.maven.org/maven2/com/google/oauth-client/google-oauth-client-java6/

They stop at 1.30.6

jaketf:
@lukecwik would it be a reasonable compromise to introduce a `google_oauth_clients_verion`? If not I believe we are at an impass.
Reasoning / Research / Background:
- features in healthcare API making this PR possible require >=1.30.9 (healthcare api absolute minimum version is 1.30.8)
- all deps under `com.google.apis` are updatable to 1.30.9
- the [two](https://repo1.maven.org/maven2/com/google/oauth-client/google-oauth-client/) [deps](https://repo1.maven.org/maven2/com/google/oauth-client/google-oauth-client-java6/) under `com.google.oauth-client` have versions that stop at 1.30.6
- I have a suspicion that this 1.30.9 for `com.google.apis` is not even really related to the 1.30.6 for `com.google.oauth-clients` but an unfortunate naming convention similarity
    - see that 1.30.6 [oauth client updated fairly recently on 2020-03-05](https://repo1.maven.org/maven2/com/google/oauth-client/google-oauth-client/1.30.6/)

I've prepared this in the latest commit to make this a concrete proposal to review / run pre/post commits on.

pabloem:
retest this please

pabloem:
retest this please

pabloem:
retest this please

jaketf:
now linkage checker is giving me no output:
```
jferriero@shadow-gallery:~/VersionControl/beam$ sdks/java/build-tools/beam-linkage-check.sh
Fri 10 Apr 2020 04:42:32 PM PDT: Installing artifacts of HL7v2IO(785a5937) to Maven local repository.
jferriero@shadow-gallery:~/VersionControl/beam$ 
```

pabloem:
Run Java PostCommit

pabloem:
Run Java PreCommit

pabloem:
Run Java PreCommit

pabloem:
Run Java PreCommit

pabloem:
Run Java PostCommit

pabloem:
Run Java PostCommit

buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy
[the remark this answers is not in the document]
lukecwik: It turns out that someone did the work to upgrade google_clients_version to 1.30.9 in #11208

pabloem:
retest this please

pabloem:
Run Java PostCommit

pabloem:
Run Java PostCommit

pabloem:
retest this please

pabloem:
retest this please

pabloem:
Run Java PostCommit

pabloem:
oops. looks like a merge issue
```
13:11:33 /home/jenkins/jenkins-slave/workspace/beam_PreCommit_Java_Commit/src/buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy: 386: The current scope already contains a variable of the name google_oauth_clients_version
13:11:33  @ line 386, column 9.
13:11:33        def google_oauth_clients_version = "1.30.6"
13:11:33            ^
13:11:33 
```

pabloem:
retest this please

pabloem:
retest this please

pabloem:
Run Java PostCommit

pabloem:
Run Java PostCommit

pabloem:
this looks fine to me as long as the dependency changes look fine to @lukecwik

pabloem:
@lukecwik ptal so @jaketf won't have to rebase if changes look fine

jaketf:
@lukecwik please let me know if there's anything else I can do to give confidence in the dependency changes.

lukecwik:
> @lukecwik please let me know if there's anything else I can do to give confidence in the dependency changes.

I have a lot of reviews on my radar and will get to this one as I can. If you want, you can reach out to the community for another reviewer to double check the dependency work and/or have Pablo take it up.

pabloem:
@suztomo - Tomo are you able to review the dependency changes for this PR?

jaketf:
@suztomo would you be able take a look at the dependency changes? otherwise I will drop a note to the dev list.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HttpHealthcareApiClient.java
```diff
@@ -0,0 +1,459 @@
...
+    // Create a HttpRequestInitializer, which will provide a baseline configuration to all requests.
+    // HttpRequestInitializer requestInitializer = new RetryHttpRequestInitializer();
+    // GoogleCredentials credentials = GoogleCredentials.getApplicationDefault();
+    HttpRequestInitializer requestInitializer =
+        new AuthenticatedRetryInitializer(
+            GoogleCredentials.getApplicationDefault()
+                .createScoped(CloudHealthcareScopes.CLOUD_PLATFORM));
+
+    client =
+        new CloudHealthcare.Builder(new NetHttpTransport(), new GsonFactory(), requestInitializer)
```
suztomo, line 335: Memo: this uses `com.google.api.client.json.gson.GsonFactory`.

suztomo: I see CloudHealthcare.Builder accepts JacksonFactory.
https://github.com/googleapis/google-api-java-client-services/blob/master/clients/google-api-services-healthcare/v1beta1/1.30.1/com/google/api/services/healthcare/v1beta1/CloudHealthcare.java#L16349

buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy
```diff
@@ -469,6 +470,7 @@ class BeamModulePlugin implements Plugin<Project> {
         google_cloud_datastore_v1_proto_client      : "com.google.cloud.datastore:datastore-v1-proto-client:1.6.3",
         google_cloud_spanner                        : "com.google.cloud:google-cloud-spanner:$google_cloud_spanner_version",
         google_http_client                          : "com.google.http-client:google-http-client:$google_http_clients_version",
+        google_http_client_gson                     : "com.google.http-client:google-http-client-gson:$google_http_clients_version",
```
suztomo, line 473: Do you really need google_http_client_gson here? Beam already has google_http_client_jackson2 and it seems CloudHealthcare.Builder accepts Jackson.

jaketf: good question.
I used Gson to be consistent w/ how we do things in the healthcare example docs and a CDAP plugin that was developed for the same purpose.
It does not seem strictly necessary.
I will prepare a commit to remove gson dep and migrate to jackson.

pabloem:
retest this please

pabloem:
retest this please

pabloem:
retest this please

pabloem:
retest this please

pabloem:
retest this please

pabloem:
retest this please

pabloem:
retest this please

pabloem:
Run Java PostCommit

pabloem:
retest this please

pabloem:
retest this please

pabloem:
Run Java PreCommit

pabloem:
Run Java PostCommit

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
```diff
@@ -0,0 +1,597 @@
...
+     * List messages.
+     *
+     * @param context the context
+     * @throws IOException the io exception
+     */
+    @ProcessElement
+    public void listMessages(ProcessContext context) throws IOException {
+      String hl7v2Store = context.element();
+      // Output all elements of all pages.
+      this.client.getHL7v2MessageStream(hl7v2Store, this.filter).forEach(context::output);
```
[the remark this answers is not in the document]
chamikaramj: Yeah, this can be a performance bottleneck and this whole operation will be limited to a single machine. Usually sources need an additional level of parallelism due to being high fanout. BTW it might sense to add a Reshuffle at the end here just to allow any subsequent steps to parallelize.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
```diff
@@ -0,0 +1,597 @@
...
+        HL7v2Message msg = context.element();
+        long startTime = System.currentTimeMillis();
+        Sleeper sleeper = Sleeper.DEFAULT;
+        switch (writeMethod) {
+          case BATCH_IMPORT:
+            throw new UnsupportedOperationException("The Batch import API is not available yet");
+          case INGEST:
+          default:
+            try {
+              client.ingestHL7v2Message(hl7v2Store, msg.toModel());
```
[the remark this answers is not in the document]
chamikaramj: Please add a TODO for this. I think this is a critical performance bottleneck.

jaketf:
>Yeah, this can be a performance bottleneck and this whole operation will be limited to a single machine. Usually sources need an additional level of parallelism due to being high fanout. BTW it might sense to add a Reshuffle at the end here just to allow any subsequent steps to parallelize.

@chamikaramj that sounds like a good idea. However, I get deprecation working on `org.apache.beam.sdk.transforms.Reshuffle` is there a new blessed way of doing this?

chamikaramj:
You can ignore the deprecation warning for Reshuffle.viaRandomKey(). The deprecation was just because the behavior across runners for Reshuffle is not well defined. Transform is not going away. Many runners add a fusion break after GBK. So this will allow subsequent steps to parallelize better.

chamikaramj:
You can also consider adding an option to not add the Reshuffle to avoid adding any additional shuffle for anyone who already will have a GBK downstream. See here: https://github.com/apache/beam/blob/master/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcIO.java#L823

jaketf:
@pabloem please do not retest this until I say so. 
reshuffle is messing up something to do w/ coders in my ITs.
will investigate and let you know when it's safe to re run tests

jaketf:
@pabloem reshuffle added and ITs passing locally as of c50df5f

pabloem:
retest this please

pabloem:
retest this please

pabloem:
Run Java PostCommit

pabloem:
Run Java PostCommit

pabloem:
And liftoff

chamikaramj:
Seems like one of the new tests is flaky.

https://builds.apache.org/job/beam_PostCommit_Java/5947/
https://builds.apache.org/job/beam_PostCommit_Java/5943/
https://builds.apache.org/job/beam_PostCommit_Java/5942/

jaketf:
@chamikaramj looking into this now.
Struggling to reproduce locally.

PR #11174, opened 2020-03-19
[BEAM-7923] Pop failed transform when error is raised

KevinGG:
1. The problem: when the CombineGlobally raises the ValueError, it
breaks out the process of transform `apply` where the current transform
has been pushed into a transform stack, to be expanded and then popped out in the end.
Without the popping out, the pipeline will hold a current transform that
is referring the failed CombineGlobally transform and never changes,
meaning the pipeline will not accept any new transforms after.
2. It did not affect non-interactive usages because when the pipeline
construction runs into an error, the user has to restart the whole
program and nothing has been executed at the point the error is raised,
so no side-effects.
3. It did leave the pipeline in a dead state when the user constructs
pipelines in an interactive environment.
4. The pop before raising the error allows the transform stack to
recover from the failed state.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/pipeline.py
+++ b/sdks/python/apache_beam/pipeline.py
@@ -307,58 +307,61 @@ class Pipeline(object):
           elif len(inputs) == 0:
             input_node = pvalue.PBegin(self.pipeline)
 
-          # We have to add the new AppliedTransform to the stack before expand()
-          # and pop it out later to make sure that parts get added correctly.
-          self.pipeline.transforms_stack.append(replacement_transform_node)
+          try:
+            # We have to add the new AppliedTransform to the stack before
+            # expand() and pop it out later to make sure that parts get added
+            # correctly.
+            self.pipeline.transforms_stack.append(replacement_transform_node)
 
-          # Keeping the same label for the replaced node but recursively
-          # removing labels of child transforms of original transform since they
-          # will be replaced during the expand below. This is needed in case
-          # the replacement contains children that have labels that conflicts
-          # with labels of the children of the original.
-          self.pipeline._remove_labels_recursively(original_transform_node)
+            # Keeping the same label for the replaced node but recursively
+            # removing labels of child transforms of original transform since
+            # they will be replaced during the expand below. This is needed in
+            # case the replacement contains children that have labels that
+            # conflicts with labels of the children of the original.
+            self.pipeline._remove_labels_recursively(original_transform_node)
 
-          new_output = replacement_transform.expand(input_node)
-          assert isinstance(
-              new_output, (dict, pvalue.PValue, pvalue.DoOutputsTuple))
+            new_output = replacement_transform.expand(input_node)
+            assert isinstance(
+                new_output, (dict, pvalue.PValue, pvalue.DoOutputsTuple))
 
-          if isinstance(new_output, pvalue.PValue):
-            new_output.element_type = None
-            self.pipeline._infer_result_type(
-                replacement_transform, inputs, new_output)
+            if isinstance(new_output, pvalue.PValue):
+              new_output.element_type = None
+              self.pipeline._infer_result_type(
+                  replacement_transform, inputs, new_output)
 
-          if isinstance(new_output, dict):
-            for new_tag, new_pcoll in new_output.items():
-              replacement_transform_node.add_output(new_pcoll, new_tag)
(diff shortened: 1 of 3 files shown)


KevinGG:
yapf formatted.
Lint passed locally.

R: @aaltay 
R: @davidyan74 
R: @rohdesamuel 

PTAL, thx!

aaltay:
`expand` for any other transform could also fail. Do we expect implementers of each of those to do the same thing? Can we apply this somewhere at a stack, where if expand fails, we clean up this stack?

@robertwb

sdks/python/apache_beam/transforms/core.py
```diff
@@ -1811,6 +1811,8 @@ def add_input_types(transform):
       return view
     else:
       if pcoll.windowing.windowfn != GlobalWindows():
+        # Remove the broken transform when running into value error.
+        pcoll.pipeline.transforms_stack.pop()
```
robertwb: This is not the right place to pop this (internal) stack. Instead, we should popping from the stack in a finally clause of a try block that pushes to the stack. (Alternatively, we could manage the stack with a Python context, but that might be overkill.)

robertwb: E.g. put this https://github.com/apache/beam/blob/release-2.19.0/sdks/python/apache_beam/pipeline.py#L330 in a finally clause of a try block that starts where it's pushed.

KevinGG: Yes, agree with it! I'll make the change.

aaltay:
retest this please

sdks/python/apache_beam/pipeline.py
```diff
@@ -307,58 +307,61 @@ def _replace_if_needed(self, original_transform_node):
...
-              if not pcoll.producer:
-                pcoll.producer = replacement_transform_node
-              output_map[original_transform_node.outputs[tag]] = pcoll
-
-          self.pipeline.transforms_stack.pop()
+          try:
+            # We have to add the new AppliedTransform to the stack before
+            # expand() and pop it out later to make sure that parts get added
+            # correctly.
+            self.pipeline.transforms_stack.append(replacement_transform_node)
```
davidyan74: This is a python newbie question. Does list.append() ever throw an exception? If so, should we move this out of the try block so that we don't pop() if list.append() fails?

KevinGG: It wouldn't raise an error if it's still a list on the line, and if it's not a list (becomes a None or some other object) at the moment it causes an error, the finally block would error out too. So it's not necessary to exclude it from the try block.

Putting it inside the try makes it a little bit more self-explained.

aaltay:
retest this please

KevinGG:
Fixed a failed test.

TL;DR, the test was wrong and passed in the past because it got lucky.

The test used to pass and started failing with the try-append-finally-pop change, because:
1. The test tried to append two no name / label transforms into a same pipeline, both of them will generate the same label and raise errors while the tests asserted for errors, thus allowing pipeline construction to continue even if it ran into fatal errors;
2. In the past, the pipeline was in a broken state where the current_transform was not popped when the 1st error was raised. The full label generated was `WriteToBigQuery`;
3. In the past, the second transform appending wrongly added parts into the broken current_transform `WriteToBigQuery`, causing the transform node to be appended after a wrong failed parent node and luckily got a unique transform full label `WriteToBigQuery/WriteToBigQuery`, thus the test didn't fail due to duplicated labels. But the pipeline was constructed as `WriteToBigQuery`->`WriteToBigQuery` which was completely messed up;
4. Still the test didn't fail because it was testing for errors even though it built a broken pipeline with broken usages.

Once the try-append-finally-pop change is applied, the test functions as:
1. First transform still fails to be appended and has side-effects including an applied transform label, but it would not leave the pipeline in a broken state where current transform is still the right node;
2. Second transform still fails but now appended to the correct node;
3. As long as the 2 transforms have different labels (or executed in different cells if in an interactive environment), the test still passes, pipeline is not broken and can be used for future development, side-effects are ruled out when data-centric APIs such as `show` and `collect` are invoked.

The reason we keep the applied label is that we never know what side effects are when the error is raised.

aaltay:
retest this please

aaltay:
retest this please

PR #11179, opened 2020-03-20
[BEAM-3301] Bugfix in DoFn validation.

youngoli:
Accidentally left a bug in DoFn validation. It was treating main inputs
as if they could only be FnValues, when they could also be FnIter or
FnReIter as well, such as in the case of a GBK or CoGBK, which meant
valid pipelines were failing validation. This fixes that.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/go/pkg/beam/core/graph/fn.go
+++ b/sdks/go/pkg/beam/core/graph/fn.go
@@ -327,11 +327,9 @@ func AsDoFn(fn *Fn, numMainIn mainInputs) (*DoFn, error) {
 		return nil, addContext(err, fn)
 	}
 
-	// If numMainIn is unknown, we can try inferring it from the second input in ProcessElement.
-	// If there is none, or it's not a FnValue type, then we can safely infer that there's only
-	// one main input.
+	// If numMainIn is unknown, we can try inferring it from the number of inputs in ProcessElement.
 	pos, num, _ := processFn.Inputs()
-	if numMainIn == MainUnknown && (num == 1 || processFn.Param[pos+1].Kind != funcx.FnValue) {
+	if numMainIn == MainUnknown && num == 1 {
 		numMainIn = MainSingle
 	}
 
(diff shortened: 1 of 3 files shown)


youngoli:
R: @lostluck

sdks/go/pkg/beam/pcollection.go
```diff
@@ -60,6 +60,12 @@ func (p PCollection) Type() FullType {
 	return p.n.Type()
 }
 
+// OutputsKV returns whether the output of this PCollection are single value
+// elements or KV pairs.
+func (p PCollection) OutputsKV() bool {
```
lostluck: 1. No need to have this exported right now, since we can't make a breaking change later. It's only used in the same package. Let's not expand the user surface unless there are good user usages, as a broad API get confusing.

2. IsKV would be a more precise name, since PCollections are a logical representation of all their data, not actually a source or a sink. They can represent KV type or they aren't.

3. This isn't checking if it's a KV type, it's checking if it's a Keyed type or not, since it's also checking if it's a CoGBK.

4. Since this is only used in the one place, it's reasonable to move the conditional there instead of adding the one off helper method.

I'm always willing to hear other opinions!

youngoli: I was originally picturing this as a helper function for callers of NewDoFn. It seems easy for future callers to make a mistake and only check if the PCollection is a KV and forget to check for CoGBK, so I thought a helper method would be useful in the future.

1. I missed that pardo.go is in the same package as pcollection.go. I'm also leaning to not expanding the user surface if it's not necessary.

2 & 3. Yeah I was unsure about the name, since it's not technically checking for KVs, I just couldn't think of anything better. IsKeyed sounds good though.

4. That's the other part I was debating. My goal was to make it easy to avoid the mistake in the future, but thinking about it... It seems unlikely that anyone else would even be using this code, and I would expect that if they were they were an advanced user doing something tricky.

I think I'll go with just putting the conditional in pardo.go and adding a comment. We can always turn it into a helper later if it does get used in multiple places.

lostluck: That's my usual guideline. If I use it once, keep it in place; twice, copy it; three times, helper function.

sdks/go/pkg/beam/core/graph/fn.go
```diff
@@ -446,23 +444,16 @@ func validateMainInputs(fn *Fn, method *funcx.Fn, methodName string, numMainIn m
...
-					"main input was expected.",
-				methodName, fn.Name(), pos+i)
-			err = errors.WithContextf(err, "method %v", methodName)
-			return err
-		}
+	// Check that the first input is not an Iter or ReIter (those aren't valid
+	// as the first main input).
+	first := method.Param[pos].Kind
+	if first != funcx.FnValue {
+		err := errors.New("first main input parameter must be value type")
```
lostluck: ...must be a value type..

youngoli: I'll just add it in real quick while squashing the commits.

PR #11184, opened 2020-03-20
[BEAM-4374] Update protos related to MonitoringInfo.

lukecwik:
Provide documentation and update all URNs and type URNs to v1.
Remove all but the payload field for MonitoringInfo for transmitting data.
Update all SDKs to encode the payload field.

------------------------

--- | --- | --- | --- | ---


lukecwik:
CC: @ajamato

model/pipeline/src/main/proto/metrics.proto
```diff
@@ -139,7 +137,7 @@ message MonitoringInfoSpecs {
 
     USER_DISTRIBUTION_COUNTER = 6 [(monitoring_info_spec) = {
       urn: "beam:metric:user_distribution",
-      type_urn: "beam:metrics:distribution_int_64",
+      type_urn: "beam:metrics:distribution_int64",
```
ajamato: Add a :v1 here

model/pipeline/src/main/proto/metrics.proto
```diff
@@ -229,101 +215,127 @@ message MonitoringInfo {
...
+    //   - valueX: beam:coder:double:v1
+    PROGRESS_TYPE = 10 [(org.apache.beam.model.pipeline.v1.beam_urn) =
+                       "beam:metrics:progress:v1"];
+
+    // General MonitoredState information which contains
+    // structured information which does not fit into a typical
+    // metric format. See MonitoringTableData for more details.
+    //
+    // Encoding: MonitoringTableData proto encoded as bytes
+    MONITORING_TABLE_TYPE = 11 [(org.apache.beam.model.pipeline.v1.beam_urn) =
```
ajamato: Hopefully you can "flatten" out the proto a bit. A common complaint was that it was complicated to go through the nested hierarchy here.

model/pipeline/src/main/proto/metrics.proto
```diff
@@ -229,101 +215,127 @@ message MonitoringInfo {
...
+  map<string, string> labels = 4;
 }
 
 message MonitoringInfoTypeUrns {
   enum Enum {
+    // Represents an integer counter where values are summed across bundles.
+    //
+    // Encoding: <value>
+    //   - value: beam:coder:varint:v1
     SUM_INT64_TYPE = 0 [(org.apache.beam.model.pipeline.v1.beam_urn) =
```
ajamato: I think there are only a few of these types being used now. SUM_INT64_TYPE and DISTRIBUTION_INT64_TYPE. I hope we can make it very simple to add new ones of these with minimal changes (Adding MonitoringInfoSpec and reusing existing framework/libraries in the SDK, runners can mostly pass them through to a service to aggregate across multiple workers)

model/pipeline/src/main/proto/metrics.proto
```diff
@@ -194,33 +188,25 @@ extend google.protobuf.EnumValueOptions {
...
   string type = 2;
 
-  // The Metric or monitored state.
-  oneof data {
-    MonitoringTableData monitoring_table_data = 3;
-    Metric metric = 4;
-    bytes payload = 7;
-  }
+  // The monitored state encoded as per the specification defined by the type.
+  bytes payload = 3;
```
ajamato: Probably you are already planning on doing this. But having helper functions to easily encode/decode these would be great.

ajamato: My biggest concern is losing the ability to print debug strings, which are helpful when people are trying to learn how these are populated. But maybe we can just add a few obvious places to dump debug logs, debug files, etc with the MonitoringInfors parses properly.

model/pipeline/src/main/proto/metrics.proto
```diff
@@ -229,101 +215,127 @@ message MonitoringInfo {
...
   // provided labels
-  map<string, string> labels = 5;
-
-  // The walltime of the most recent update.
-  // Useful for aggregation for latest types such as LatestInt64.
-  google.protobuf.Timestamp timestamp = 6;
+  map<string, string> labels = 4;
 }
 
 message MonitoringInfoTypeUrns {
```
ajamato: Let's add some comments to make it clear the type is referring to what is collected in each MonitoringInfo update, and how they should be aggregated together

model/pipeline/src/main/proto/metrics.proto
```diff
@@ -229,101 +215,127 @@ message MonitoringInfo {
...
+    // Represents the smallest set of double values seen across bundles.
+    //
+    // Encoding: <value1><value2>...<valueN>
+    //   - valueX: beam:coder:double:v1
+    BOTTOM_N_DOUBLE_TYPE = 9 [(org.apache.beam.model.pipeline.v1.beam_urn) =
+                           "beam:metrics:bottom_n_int64:v1"];
+
+    // Encoding: <value1><value2>...<valueN>
+    //   - valueX: beam:coder:double:v1
+    PROGRESS_TYPE = 10 [(org.apache.beam.model.pipeline.v1.beam_urn) =
```
ajamato: I think progress should be a MonitoringInfoSpec URN and it should use either SUM_DOUBLE_TYPE or X_DOUBLE_TYPE.

It should be something which defscribes what each update is supposed to be, and how they should be aggregated together. I think that's required for each type, in order to all the use case of "worker just forwards these to a service to aggregate" As it may need to go pre aggregation of its own?

GAUGE_DOUBLE_TYPE? I am not sure what's appropriate here. But it probably doesn't make sense for progress to be its own type.

model/pipeline/src/main/proto/metrics.proto
lukecwik: I'll see what I can do when I'm mucking around with this in the code.

model/pipeline/src/main/proto/metrics.proto
lukecwik: Yup

model/pipeline/src/main/proto/metrics.proto
lukecwik: Done

model/pipeline/src/main/proto/metrics.proto
lukecwik: This is the individual progress for each element and restriction pair for a splittable DoFn. The aggregation is interesting since if you sum across all **latest** values then that tells you the amount of completed work globally and the amount of known but not yet finished work globally. Note that this won't cover the amount of unknown work (e.g. elements and restrictions that have yet to be processed).

@robertwb / @ajamato What do you think about adding `ELEMENT` as labels (storing the encoded element and restriction) and then we could use `LATEST_DOUBLE_TYPE`?

robertwb: We don't want to report progress for each processed element, as this would become an unreasonably large value. We could report aggregated progress of all finished ones, plus partial progress of the in-flight ones, in which case the sum becomes a refinement of done-element-count. There's still no useful way to define cross-bundle aggregation (without resulting in an absolutely huge value), so I would lean towards this being its own type.

model/pipeline/src/main/proto/metrics.proto
lukecwik: The problematic type is `string`:
Is `sum` the same as concat? (if so, this will grow to be huge very quickly also making distribution pointless)
Is min/max/top_n/bottom_n based upon lexicographical ordering? (if so, case sensitive/insensitive/...)

`string` only makes sense with latest and if we wanted to define any of the other ones I think it would be better to use terms like `case_insensitive_max`.

model/pipeline/src/main/proto/metrics.proto
lukecwik: Done

model/pipeline/src/main/proto/metrics.proto
```diff
@@ -229,121 +330,148 @@ message MonitoringInfo {
...
+    //
+    // Encoding: <timestamp><value>
+    //   - timestamp: beam:coder:varint:v1     (milliseconds since epoch)
+    //   - value:     beam:coder:double:v1
+    LATEST_DOUBLE_TYPE = 5 [(org.apache.beam.model.pipeline.v1.beam_urn) =
+                           "beam:metrics:latest_double:v1"];
+
+    // Represents the largest set of integer values seen across bundles.
+    //
+    // Encoding: <value1><value2>...<valueN>
```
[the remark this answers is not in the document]
lukecwik: Done

model/pipeline/src/main/proto/metrics.proto
```diff
@@ -52,38 +55,157 @@ message Annotation {
...
         key: "description",
-        value: "URN utilized to report user numeric counters."
+        value: "URN utilized to report user metric."
       }]
     }];
 
-    ELEMENT_COUNT = 1 [(monitoring_info_spec) = {
+    // Represents a double counter where values are summed across bundles.
+    USER_SUM_DOUBLE = 1 [(monitoring_info_spec) = {
+      urn: "beam:metric:user:v1",
```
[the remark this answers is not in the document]
lukecwik: The URN represents the unique identifier in the "semantic meaning" space of the counter. So saying that this is a user counter and that there are many types makes sense to me.

model/pipeline/src/main/proto/metrics.proto
robertwb: Agree that string summation (and those metrics requiring it) could be omitted, but min, max, top_n, bottom_n are meaningful. (And distribution could still be meaningful giving a count and range.)

model/pipeline/src/main/proto/metrics.proto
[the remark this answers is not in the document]
robertwb: I suppose I still haven't bought into the idea that the whole set of "user counters" belongs to a single semantic identifier on par with each individual system counter like msecs or element counts--all we really need is a way to not have names between the two collide--but if we're going to go this route I don't think there's value in enumerating them here (vs. simply saying "user counter" can use any type).

sdks/go/pkg/beam/core/runtime/harness/monitoring.go
```diff
@@ -126,15 +134,18 @@ func monitoring(p *exec.Plan) (*fnpb.Metrics, []*ppb.MonitoringInfo) {
...
-			&ppb.MonitoringInfo{
-				Urn:  "beam:metric:element_count:v1",
-				Type: "beam:metrics:sum_int_64",
-				Labels: map[string]string{
-					"PCOLLECTION": snapshot.PID,
-				},
-				Data: int64Counter(snapshot.Count),
-			})
+		payload, err := int64Counter(snapshot.Count)
+		if err != nil {
```
lostluck: Error case reversal. 
Consider 
```
if err != nil {
  panic(err)
}
// ...actual contents of the block. 
```

sdks/go/pkg/beam/core/runtime/harness/monitoring.go
```diff
@@ -80,36 +84,40 @@ func monitoring(p *exec.Plan) (*fnpb.Metrics, []*ppb.MonitoringInfo) {
...
+						Type:    "beam:metrics:distribution_int64:v1",
+						Labels:  userLabels(l),
+						Payload: payload,
+					})
+			}
 		},
 		GaugeInt64: func(l metrics.Labels, v int64, t time.Time) {
-			ts, err := ptypes.TimestampProto(t)
+			payload, err := int64Latest(t, v)
 			if err != nil {
```
lostluck: Error case reversal. 
Consider 
```
if err != nil {
  panic(err)
}
// ...actual contents of the block. 
```

sdks/go/pkg/beam/core/runtime/harness/monitoring.go
```diff
@@ -80,36 +84,40 @@ func monitoring(p *exec.Plan) (*fnpb.Metrics, []*ppb.MonitoringInfo) {
...
 		DistributionInt64: func(l metrics.Labels, count, sum, min, max int64) {
-			monitoringInfo = append(monitoringInfo,
-				&ppb.MonitoringInfo{
-					Urn:    "beam:metric:user_distribution",
-					Type:   "beam:metrics:distribution_int_64",
-					Labels: userLabels(l),
-					Data:   int64Distribution(count, sum, min, max),
-				})
+			payload, err := int64Distribution(count, sum, min, max)
+			if err != nil {
```
lostluck: Error case reversal. 
Consider 
```
if err != nil {
  panic(err)
}
// ...actual contents of the block. 
```

sdks/go/pkg/beam/core/runtime/harness/monitoring.go
```diff
@@ -80,36 +84,40 @@ func monitoring(p *exec.Plan) (*fnpb.Metrics, []*ppb.MonitoringInfo) {
...
 		SumInt64: func(l metrics.Labels, v int64) {
-			monitoringInfo = append(monitoringInfo,
-				&ppb.MonitoringInfo{
-					Urn:    "beam:metric:user",
-					Type:   "beam:metrics:sum_int_64",
-					Labels: userLabels(l),
-					Data:   int64Counter(v),
-				})
+			payload, err := int64Counter(v)
+			if err != nil {
```
lostluck: Error case reversal. 
Consider 
```
if err != nil {
  panic(err)
}
// ...actual contents of the block. 
```

model/pipeline/src/main/proto/metrics.proto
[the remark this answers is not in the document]
lostluck: For the Go short id generation, I'm relying on the label set to be unique, and map to the semantic URN and Type. That is, I'm assuming for a counter with a given namespace, name, and ptransform, will always be the same URN and Type, and should use the same short id.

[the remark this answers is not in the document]
lukecwik: I enumerate them so that its easier to write validation logic in the future that uses the monitoring specs.

Also I don't believe user counters can have any type as the progress type doesn't make much sense.

sdks/go/pkg/beam/core/runtime/harness/monitoring.go
lukecwik: Done

sdks/go/pkg/beam/core/runtime/harness/monitoring.go
lukecwik: Done

sdks/go/pkg/beam/core/runtime/harness/monitoring.go
lukecwik: Done

sdks/go/pkg/beam/core/runtime/harness/monitoring.go
lukecwik: Done

lukecwik:
R: @ajamato for Java/Python
R: @lostluck for Go

lukecwik:
While this is getting reviewed, I'm going to start working on testing/updating the Google codebase to reflect this backwards incompatible change.

lukecwik:

lukecwik:

lukecwik:
Run PythonLint PreCommit

lukecwik:

lukecwik:

lukecwik:
Run Java PreCommit

lukecwik:
Run Java PreCommit

model/pipeline/src/main/proto/metrics.proto
```diff
@@ -229,121 +330,152 @@ message MonitoringInfo {
...
+                                 "beam:metrics:distribution_double:v1"];
+
+    // Represents the latest seen integer value. The timestamp is used to
+    // provide an "ordering" over multiple values to determine which is the
+    // latest.
+    //
+    // Encoding: <timestamp><value>
+    //   - timestamp: beam:coder:varint:v1     (milliseconds since epoch)
+    //   - value:     beam:coder:varint:v1
+    LATEST_INT64_TYPE = 4 [(org.apache.beam.model.pipeline.v1.beam_urn) =
```
lostluck: Should we consider explicitly having the ms as part of the URN, to keep 'us' and 'ns' granularity definitions open for the future?

sdks/go/pkg/beam/core/runtime/harness/monitoring.go
```diff
@@ -80,36 +84,43 @@ func monitoring(p *exec.Plan) (*fnpb.Metrics, []*ppb.MonitoringInfo) {
...
+			  panic(err)
+			}
+      monitoringInfo = append(monitoringInfo,
```
lostluck: Need to run `go fmt` again.

lukecwik: done

lukecwik: Done

model/pipeline/src/main/proto/metrics.proto
lukecwik: Would need to do this in other places such as the timer coder, windowed value coder if we wanted to keep the URN naming to be consistent so I'll leave as is and added this to the protocol changes tracking sheet.

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/SimpleMonitoringInfoBuilder.java
```diff
@@ -86,46 +89,97 @@ public SimpleMonitoringInfoBuilder setUrn(String urn) {
...
   }
 
-  /** Sets the the appropriate type URN for sum int64 counters. */
-  public SimpleMonitoringInfoBuilder setInt64TypeUrn() {
-    this.builder.setType(MonitoringInfoConstants.TypeUrns.SUM_INT64);
+  /**
+   * Encodes the value and sets the type to {@link
+   * MonitoringInfoConstants.TypeUrns#DISTRIBUTION_INT64_TYPE}.
+   */
+  public SimpleMonitoringInfoBuilder setDoubleDistributionValue(
```
ajamato: Looks like SimpleMonitoringInfoBuilder helped a lot with this refactor :)

runners/core-java/src/test/java/org/apache/beam/runners/core/metrics/MonitoringInfoMatchers.java
```diff
@@ -28,30 +31,17 @@
...
-        }
-
-        if (mi.getMetric().hasCounterData()) {
-          long valueToMatch = mi.getMetric().getCounterData().getInt64Value();
-          if (valueToMatch != item.getMetric().getCounterData().getInt64Value()) {
-            return false;
-          }
-        }
-        return true;
+        return (mi.getUrn().isEmpty() || mi.getUrn().equals(item.getUrn()))
```
ajamato: Why did you add the isEmpty() checks, this looks like the old matcher didn't have this logic, and it would match if you initialized the matcher with an empty monitoring info

runners/core-java/src/test/java/org/apache/beam/runners/core/metrics/MonitoringInfoMatchers.java
```diff
@@ -76,17 +63,20 @@ public void describeTo(Description description) {
    * Matches a {@link MonitoringInfo} with that has the set fields in the provide MonitoringInfo.
    *
    * <p>This is useful for tests which do not want to match the specific value (execution times).
-   * Currently this will only check for URNs, labels, type URNs and int64Values.
+   * Currently this will only check for URNs, labels, type URNs and {@code beam:coder:varint:v1}
+   * encoded values.
    */
-  public static TypeSafeMatcher<MonitoringInfo> valueGreaterThan(final long value) {
+  public static TypeSafeMatcher<MonitoringInfo> valueGreaterThanOrEqualTo(final long value) {
```
ajamato: would this only work with certain types? Should this be renamed?

runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/fn/control/ElementCountMonitoringInfoToCounterUpdateTransformer.java
```diff
@@ -95,7 +99,12 @@ public CounterUpdate transform(MonitoringInfo monitoringInfo) {
       return null;
     }
 
-    long value = monitoringInfo.getMetric().getCounterData().getInt64Value();
+    long value;
+    try {
+      value = VARINT_CODER.decode(monitoringInfo.getPayload().newInput());
```
ajamato: Can we encapsulte this so that its still one line to get the value out? I think that this would be a nicer interface. For consumers. Ideally something as easy to use on the receiving side as SimpleMonitoringInfoBuilder.

i.e.
long value = MonitoringInfos.getSumInt64Value(monitoringInfo)

runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/fn/control/FnApiMonitoringInfoToCounterUpdateTransformer.java
```diff
@@ -35,50 +35,82 @@
 public class FnApiMonitoringInfoToCounterUpdateTransformer
     implements MonitoringInfoToCounterUpdateTransformer {
 
-  final Map<String, MonitoringInfoToCounterUpdateTransformer> counterTransformers = new HashMap<>();
+  final Map<String, Map<String, MonitoringInfoToCounterUpdateTransformer>>
```
ajamato: Can you key this by a tuple instead of a 2 layer dictionary? That might reduce some of the boilerplate below. I believe the Pair class can work here

runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/fn/control/MSecMonitoringInfoToCounterUpdateTransformer.java
```diff
@@ -120,14 +110,19 @@ public CounterUpdate transform(MonitoringInfo monitoringInfo) {
       return null;
     }
 
-    long value = monitoringInfo.getMetric().getCounterData().getInt64Value();
+    long value;
+    try {
+      value = VARINT_CODER.decode(monitoringInfo.getPayload().newInput());
```
ajamato: Ditto here, please use a one liner helper method to extract the value, here and in all the transformer classes.

runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/fn/control/MeanByteCountMonitoringInfoToCounterUpdateTransformer.java
```diff
@@ -100,8 +101,18 @@ public CounterUpdate transform(MonitoringInfo monitoringInfo) {
       return null;
     }
 
-    IntDistributionData value =
-        monitoringInfo.getMetric().getDistributionData().getIntDistributionData();
+    long count;
+    long sum;
+    try {
+      InputStream payload = monitoringInfo.getPayload().newInput();
```
ajamato: ditto

runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/fn/control/RegisterAndProcessBundleOperation.java
```diff
@@ -446,7 +447,11 @@ long getInputElementsConsumed(final Iterable<MonitoringInfo> monitoringInfos) {
         String pcollection =
             mi.getLabelsOrDefault(MonitoringInfoConstants.Labels.PCOLLECTION, null);
         if (pcollection != null && pcollection.equals(grpcReadTransformOutputPCollectionName)) {
-          return mi.getMetric().getCounterData().getInt64Value();
+          try {
+            return VARINT_CODER.decode(mi.getPayload().newInput());
```
ajamato: ditto, this is also too low level, please use a helper here. Let's try to encapsulate the bytes payload format of MonitoringInfos.

runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/fn/control/UserMonitoringInfoToCounterUpdateTransformer.java
```diff
@@ -95,7 +100,12 @@ public CounterUpdate transform(MonitoringInfo monitoringInfo) {
       return null;
     }
 
-    long value = monitoringInfo.getMetric().getCounterData().getInt64Value();
+    long value;
+    try {
+      value = VARINT_CODER.decode(monitoringInfo.getPayload().newInput());
```
ajamato: ditto please use one liner here

runners/google-cloud-dataflow-java/worker/src/test/java/org/apache/beam/runners/dataflow/worker/fn/control/MSecMonitoringInfoToCounterUpdateTransformerTest.java
```diff
@@ -54,39 +58,33 @@ public void setUp() throws Exception {
 
   @Test
   public void testTransformReturnsNullIfSpecValidationFails() {
-    Map<String, String> counterNameMapping = new HashMap<>();
-    counterNameMapping.put("beam:counter:supported", "supportedCounter");
-
     Map<String, DataflowStepContext> stepContextMapping = new HashMap<>();
 
     MSecMonitoringInfoToCounterUpdateTransformer testObject =
```
ajamato: Can you add a TODO to rename this to ExecutionTimeMonitoringInfoToCounterUpdateTransformer

runners/portability/java/src/main/java/org/apache/beam/runners/portability/PortableMetrics.java
```diff
@@ -115,17 +113,15 @@ private static PortableMetrics convertMonitoringInfosToMetricResults(
...
-        monitoringInfo.getMetric().getExtremaData().getIntExtremaData();
-    // Get only last value of the extrema table
-    Instant timestamp = Instant.ofEpochSecond(monitoringInfo.getTimestamp().getSeconds());
-    if (extremaData.getIntValuesCount() > 0) {
-      GaugeResult result =
-          GaugeResult.create(
-              extremaData.getIntValues(extremaData.getIntValuesCount() - 1), timestamp);
+    try {
+      InputStream input = monitoringInfo.getPayload().newInput();
+      long timestamp = VARINT_CODER.decode(input);
```
ajamato: ditto, please use a helper here to make this a one liner to extract the value

runners/portability/java/src/main/java/org/apache/beam/runners/portability/PortableMetrics.java
```diff
@@ -135,21 +131,23 @@ private static PortableMetrics convertMonitoringInfosToMetricResults(
...
-    DistributionResult result =
-        DistributionResult.create(
-            intDistributionData.getSum(),
-            intDistributionData.getCount(),
-            intDistributionData.getMin(),
-            intDistributionData.getMax());
-    return MetricResult.create(key, false, result);
+    try {
+      InputStream input = monitoringInfo.getPayload().newInput();
+      long count = VARINT_CODER.decode(input);
```
ajamato: ditto, please use helper here

model/pipeline/src/main/proto/metrics.proto
```diff
@@ -33,14 +33,17 @@ import "beam_runner_api.proto";
...
-// ProcessBundleResponse reporting.
+// A specification for describing a well known MonitoringInfo.
 message MonitoringInfoSpec {
+  // Defines the semantic meaning of the metric or monitored state.
   string urn = 1;
-  string type_urn = 2;
-  // The list of required
+
+  // Defines the encoding and aggregation method for the payload.
+  string type = 2;
```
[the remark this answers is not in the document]
lukecwik: Done

model/pipeline/src/main/proto/metrics.proto
```diff
@@ -33,14 +33,17 @@ import "beam_runner_api.proto";
 import "google/protobuf/descriptor.proto";
 import "google/protobuf/timestamp.proto";
 
-// A specification containing required set of fields and labels required
-// to be set on a MonitoringInfo for the specific URN for SDK->RunnerHarness
-// ProcessBundleResponse reporting.
+// A specification for describing a well known MonitoringInfo.
 message MonitoringInfoSpec {
+  // Defines the semantic meaning of the metric or monitored state.
```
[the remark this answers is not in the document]
lukecwik: Done

model/pipeline/src/main/proto/metrics.proto
```diff
@@ -52,38 +55,157 @@ message Annotation {
...
+      annotations: [{
+        key: "description",
+        value: "URN utilized to report user metric."
+      }]
+    }];
+
+    // General monitored state information which contains structured information
+    // which does not fit into a typical metric format. See MonitoringTableData
+    // for more details.
+    USER_MONITORING_TABLE = 10 [(monitoring_info_spec) = {
```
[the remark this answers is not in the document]
lukecwik: Commented it out with a TODO and filed BEAM-9617

runners/google-cloud-dataflow-java/worker/src/test/java/org/apache/beam/runners/dataflow/worker/fn/control/MSecMonitoringInfoToCounterUpdateTransformerTest.java
lukecwik: Did the rename.

runners/core-java/src/main/java/org/apache/beam/runners/core/metrics/SimpleMonitoringInfoBuilder.java
lukecwik: Yup

runners/core-java/src/test/java/org/apache/beam/runners/core/metrics/MonitoringInfoMatchers.java
lukecwik: The contract of the matcher states that it only compares **set** fields and it seemed it was incorrectly implemented. I can change it back to the existing implementation where `urn`, `labels`, and `type` must always be specified and fix the comment.

runners/core-java/src/test/java/org/apache/beam/runners/core/metrics/MonitoringInfoMatchers.java
lukecwik: Renamed to `counterValueGreaterThanOrEqualTo`

Also fixed comment.

runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/fn/control/FnApiMonitoringInfoToCounterUpdateTransformer.java
lukecwik: Done.

runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/fn/control/ElementCountMonitoringInfoToCounterUpdateTransformer.java
lukecwik: Done here an elsewhere. I introduced a MonitoringInfoEncodings class with the convenience methods for the currently used encodings.

model/pipeline/src/main/proto/metrics.proto
```diff
@@ -52,38 +61,160 @@ message Annotation {
...
+      type: "beam:metrics:sum_double:v1",
+      required_labels: ["PTRANSFORM", "NAMESPACE", "NAME"],
+      annotations: [{
+        key: "description",
+        value: "URN utilized to report user metric."
+      }]
+    }];
+
+    // Represents a distribution of an integer value where:
+    //   - count: represents the number of values seen across all bundles
```
ajamato: Now it seems like there technically aren't any fields named "count", "sum", "min", "max". Just 4 encoded varints in that specific order. There is no longer a proto or anything which defines this format.

If we are going to keep type urns, I think that there should be somewhere in this file where you could a "TypeSpec", which describes how to encode each opaque bytes payload. i.e. the coders used for each value, the order they must be encoded. Or a proto that should be serialized into that bytes field, etc. A description that will work for all languages. Right now you can only know that from looking at your encoding code.

I think it would be best if SDK implemented could look at a reference file like this and know how to populate the MonitoringInfo. That was the original intention behind MonitoringInfoSpec, and I believe that is a bit lost now with this change.

model/pipeline/src/main/proto/metrics.proto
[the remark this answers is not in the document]
lukecwik: Made all the URNs unique and added a test to make sure that they remain unique.

lukecwik:
This is ready for the next review.

model/pipeline/src/main/proto/metrics.proto
robertwb: It is explicit in the type field which is a URN denoting exactly how the values are encoded? Did we need more?

model/pipeline/src/main/proto/metrics.proto
```diff
@@ -52,38 +62,160 @@ message Annotation {
...
+    //   urn: "beam:metric:user:v1",
+    //   type: "beam:metrics:monitoring_table:v1",
+    //   required_labels: ["PTRANSFORM", "NAMESPACE", "NAME"],
+    //   annotations: [{
+    //     key: "description",
+    //     value: "URN utilized to report user monitoring data."
+    //   }]
+    // }];
+
+    ELEMENT_COUNT = 10 [(monitoring_info_spec) = {
```
robertwb: Nit: Should we jump to 100 (or alternatively have the user types be 1xxx) just to have nice consecutive numbers are more types get added?

lukecwik: Having the actual monitoring specs that are related grouped together is all that matters, the numbers shouldn't matter.

model/pipeline/src/main/proto/metrics.proto
lukecwik: I chatted with Alex about this and the TypeUrns describing the encoding was enough.

lukecwik:
Run Python PreCommit

lukecwik:
Run Java PreCommit

lukecwik:
Run Java PreCommit

lukecwik:
Run Java PreCommit

lukecwik:
Run Java PreCommit

lukecwik:
https://builds.apache.org/job/beam_PreCommit_Java_Examples_Dataflow_Commit/7318/ passed, merging

PR #11186, opened 2020-03-20
[BEAM-9564] Remove insecure ssl options from MongoDBIO

iemejia:
These changes are not backwards compatible but this is intended to solve the potential security issues and also because MongoDBIO does not have strong backwards compatibility yet (aka it is still tagged as `@Experimental`).

R: @alexvanboxel


iemejia:
Run Java PreCommit

iemejia:
Run Java PreCommit

alexvanboxel:
Do users have an alternative when using self signed certificates?

iemejia:
They will once we allow them to provide their own MongoClient via a provider. Filled [BEAM-9571](https://issues.apache.org/jira/browse/BEAM-9571) for this. This PR is about avoiding at the Beam side to suggest them to shoot in their feet. Up to them if they want to explicitly do :)
I can inverse the order of the PRs if you prefer. Providing first the valid replacement alternative (BEAM-9571) and then the removal of the method, we can even deprecate it and be more 'user friendly' if you prefer. (I wanted to go ahead faster because this is a potential security issue for users). WDYT?

alexvanboxel:
> I can inverse the order of the PRs if you prefer. Providing first the valid replacement alternative ([BEAM-9571](https://issues.apache.org/jira/browse/BEAM-9571)) and then the removal of the method, we can even deprecate it and be more 'user friendly' if you prefer. (I wanted to go ahead faster because this is a potential security issue for users). WDYT?

I'm afraid that when the second ticket doesn't make the release cut, this one could block users from upgrading... So if we can reverse the order this is more user friendly (they still need to change something)... and the security issue is not because it's a code vulnerability.

iemejia:
Ok, i will work on the other PR and ping you there, and then we come back to this one. Thanks for the question and comment, that's a safer approach.

iemejia:
Lies this is not stale!!! well maybe a bit, but I plan to come back to it soonish.

aromanenko-dev:
Any plans to work on this? Do we need to label it with `[WIP]`?

PR #11189, opened 2020-03-20
[BEAM-9446] Retain unknown arguments when using uber jar job server.

ibzib:
See #11052 for context.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/options/pipeline_options.py
+++ b/sdks/python/apache_beam/options/pipeline_options.py
@@ -296,8 +296,12 @@ class PipelineOptions(HasDisplayData):
       while i < len(unknown_args):
         # Treat all unary flags as booleans, and all binary argument values as
         # strings.
-        if i + 1 >= len(unknown_args) or unknown_args[i + 1].startswith('--'):
-          parser.add_argument(unknown_args[i], action='store_true')
+        if i + 1 >= len(unknown_args) or unknown_args[i + 1].startswith('-'):
+          split = unknown_args[i].rsplit('=')
+          if len(split) == 1:
+            parser.add_argument(unknown_args[i], action='store_true')
+          else:
+            parser.add_argument(split[0], type=str)
           i += 1
         else:
           parser.add_argument(unknown_args[i], type=str)
--- a/sdks/python/apache_beam/options/pipeline_options.py
+++ b/sdks/python/apache_beam/options/pipeline_options.py
@@ -297,7 +297,7 @@ class PipelineOptions(HasDisplayData):
         # Treat all unary flags as booleans, and all binary argument values as
         # strings.
         if i + 1 >= len(unknown_args) or unknown_args[i + 1].startswith('-'):
-          split = unknown_args[i].rsplit('=')
+          split = unknown_args[i].split('=', 1)
           if len(split) == 1:
             parser.add_argument(unknown_args[i], action='store_true')
           else:
--- a/sdks/python/apache_beam/options/pipeline_options.py
+++ b/sdks/python/apache_beam/options/pipeline_options.py
@@ -256,7 +256,8 @@ class PipelineOptions(HasDisplayData):
   def get_all_options(
       self,
       drop_default=False,
-      add_extra_args_fn=None  # type: Optional[Callable[[_BeamArgumentParser], None]]
+      add_extra_args_fn=None,  # type: Optional[Callable[[_BeamArgumentParser], None]]
+      retain_unknown_options=False
   ):
     # type: (...) -> Dict[str, Any]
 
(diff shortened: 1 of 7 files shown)


ibzib:
Run PortableJar_Flink PostCommit

sdks/python/apache_beam/options/pipeline_options.py
```diff
@@ -285,10 +289,25 @@ def get_all_options(
...
+      i = 0
+      while i < len(unknown_args):
+        # Treat all unary flags as booleans, and all binary argument values as
+        # strings.
+        if i + 1 >= len(unknown_args) or unknown_args[i + 1].startswith('--'):
+          parser.add_argument(unknown_args[i], action='store_true')
+          i += 1
+        else:
+          parser.add_argument(unknown_args[i], type=str)
+          i += 2
```
mxm: What about the `--arg=value` format?

sdks/python/apache_beam/options/pipeline_options.py
```diff
@@ -285,10 +289,25 @@ def get_all_options(
...
     known_args, unknown_args = parser.parse_known_args(self._flags)
-    if unknown_args:
-      _LOGGER.warning("Discarding unparseable args: %s", unknown_args)
-    result = vars(known_args)
+    if retain_unknown_options:
+      i = 0
+      while i < len(unknown_args):
+        # Treat all unary flags as booleans, and all binary argument values as
+        # strings.
+        if i + 1 >= len(unknown_args) or unknown_args[i + 1].startswith('--'):
```
mxm: Should we always check a parameter name starts with `--`, like done here for boolean flags?

sdks/python/apache_beam/options/pipeline_options.py
ibzib: Good catch. I will have to add additional logic to handle that.

sdks/python/apache_beam/options/pipeline_options.py
ibzib: I guess I should change it to `-`, not `--`. `-` is the default prefix for argparse, and AFAICT that is true for Beam as well, even though most (all?) pipeline options use `--`.
https://docs.python.org/3/library/argparse.html#prefix-chars

So then the question becomes "Should we always check a parameter name starts with `-`?" And the answer is we don't have to, because the call to parse_args will do that for us. `test_retain_unknown_options_unary_missing_prefix` tests that.

ibzib:
@mxm thanks for the feedback, I addressed your comments.

Unrelated: I wasn't aware that the Flink portable jar postcommit has been failing. I filed and self-assigned BEAM-9575.

mxm:
Thanks for filing the bug. Unfortunately we have some post commits failing after adding Flink 1.10.

sdks/python/apache_beam/options/pipeline_options.py
mxm: That sounds good :)

sdks/python/apache_beam/options/pipeline_options.py
```diff
@@ -285,10 +289,29 @@ def get_all_options(
...
-    if unknown_args:
-      _LOGGER.warning("Discarding unparseable args: %s", unknown_args)
-    result = vars(known_args)
+    if retain_unknown_options:
+      i = 0
+      while i < len(unknown_args):
+        # Treat all unary flags as booleans, and all binary argument values as
+        # strings.
+        if i + 1 >= len(unknown_args) or unknown_args[i + 1].startswith('-'):
+          split = unknown_args[i].rsplit('=')
```
mxm: Why `rsplit`? Shouldn't this be: 
```suggestion
          split = unknown_args[i].split('=', 1)
```

?

mxm: Otherwise this will break options like `experiments=state_cache_size=1`.

ibzib: The split is only used to get the argument name, and the rest of the split isn't used.

mxm: Comment still applies. Splitting should be performed from the left, not the right. At most one split has to be performed.

ibzib: Oh, I see what you mean now. I will commit your suggestion to avoid confusion. However, I don't think we should support unrecognized `append` arguments. I don't want to have to infer complex argument types. As for the specific case of `experiments`, that will not be a problem because `experiments` is already recognized by the parser.

ibzib: I will merge this now. If we need to support `append` arguments for some reason, we should be able to add that later.

mxm: I was only referring to options values containing `=` which can be the case independently of append options. That is fixed now. So all good.

PR #11191, opened 2020-03-23
[BEAM-9570] Harden deserialization in SerializableCoder

coheigea:
It's possible to harden the deserialization process in SerializableCoder by enforcing that deserialization will only take place if the root object matches the type that the SerializableCoder is associated with.
R: @kennknowles


coheigea:
I'll close this PR due to the comments above, and update the JIRA to reflect instead the proposals on the mailing list.

PR #11205, opened 2020-03-24
[BEAM-9578] Enumerating artifacts is too expensive in Java

ihji:
------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/model/pipeline/src/main/proto/beam_runner_api.proto
+++ b/model/pipeline/src/main/proto/beam_runner_api.proto
@@ -1146,23 +1146,27 @@ message StandardArtifacts {
   enum Types {
     // A URN for locally-accessible artifact files.
     // payload: ArtifactFilePayload
-    FILE     = 0 [(beam_urn) = "beam:artifact:type:file:v1"];
+    FILE      = 0 [(beam_urn) = "beam:artifact:type:file:v1"];
 
     // A URN for artifacts described by URLs.
     // payload: ArtifactUrlPayload
-    URL      = 1 [(beam_urn) = "beam:artifact:type:url:v1"];
+    URL       = 1 [(beam_urn) = "beam:artifact:type:url:v1"];
 
     // A URN for artifacts embedded in ArtifactInformation proto.
     // payload: EmbeddedFilePayload.
-    EMBEDDED = 2 [(beam_urn) = "beam:artifact:type:embedded:v1"];
+    EMBEDDED  = 2 [(beam_urn) = "beam:artifact:type:embedded:v1"];
 
     // A URN for Python artifacts hosted on PYPI.
     // payload: PypiPayload
-    PYPI     = 3 [(beam_urn) = "beam:artifact:type:pypi:v1"];
+    PYPI      = 3 [(beam_urn) = "beam:artifact:type:pypi:v1"];
 
     // A URN for Java artifacts hosted on a Maven repository.
     // payload: MavenPayload
-    MAVEN    = 4 [(beam_urn) = "beam:artifact:type:maven:v1"];
+    MAVEN     = 4 [(beam_urn) = "beam:artifact:type:maven:v1"];
+
+    // A URN for deferred artifacts.
+    // payload: DeferredArtifactPayload
+    DEFERRED = 5 [(beam_urn) = "beam:artifact:type:deferred:v1"];
   }
   enum Roles {
     // A URN for staging-to role.
@@ -1172,7 +1176,7 @@ message StandardArtifacts {
 }
 
 message ArtifactFilePayload {
-  // a string for an artifact path e.g. "/tmp/foo.jar"
+  // a string for an artifact file path e.g. "/tmp/foo.jar"
   string path = 1;
 
   // The hex-encoded sha256 checksum of the artifact.
@@ -1204,6 +1208,15 @@ message MavenPayload {
 
   // (Optional) Repository URL. If not specified, Maven central is used by default.
   string repository_url = 2;
+}
+
+message DeferredArtifactPayload {
+  // A unique string identifier assigned by the creator of this payload. The creator may use this key to confirm
+  // whether they can parse the data.
+  string key = 1;
+
+  // Data for deferred artifacts. Interpretation of bytes is delegated to the creator of this payload.
+  bytes data = 2;
 }
 
 message ArtifactStagingToRolePayload {
--- a/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/ArtifactResolver.java
+++ b/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/ArtifactResolver.java
@@ -1,0 +1,50 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.beam.runners.core.construction;
+
+import java.util.List;
+import org.apache.beam.model.pipeline.v1.RunnerApi;
+
+/**
+ * An interface for artifact resolvers. Artifact resolvers transform {@link
+ * RunnerApi.ArtifactInformation} from one type to another typically from remote or deferred
+ * artifacts to locally accessible file artifacts.
+ */
+public interface ArtifactResolver {
+
+  /**
+   * Register {@link ResolutionFn}. When multiple {@link ResolutionFn}s are capable of resolving the
+   * same {@link RunnerApi.ArtifactInformation}, the application order is up to the implementing
+   * class of this interface.
+   *
+   * @param fn a resolution function to be registered
+   */
+  void register(ResolutionFn fn);
+
+  /** Updating pipeline proto by applying registered {@link ResolutionFn}s. */
+  RunnerApi.Pipeline resolveArtifacts(RunnerApi.Pipeline pipeline);
+
+  /**
+   * A lazy transformer for resolving {@link RunnerApi.ArtifactInformation}. Note that {@link
+   * ResolutionFn} postpones any side-effect until {@link
+   * ArtifactResolver#resolveArtifacts(RunnerApi.Pipeline)} is called.
+   */
+  interface ResolutionFn {
+    List<RunnerApi.ArtifactInformation> resolve(RunnerApi.ArtifactInformation info);
+  }
+}
--- a/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/ArtifactResolver.java
+++ b/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/ArtifactResolver.java
@@ -18,6 +18,7 @@
 package org.apache.beam.runners.core.construction;
 
 import java.util.List;
+import java.util.Optional;
 import org.apache.beam.model.pipeline.v1.RunnerApi;
 
 /**
@@ -45,6 +46,6 @@ public interface ArtifactResolver {
    * ArtifactResolver#resolveArtifacts(RunnerApi.Pipeline)} is called.
    */
   interface ResolutionFn {
-    List<RunnerApi.ArtifactInformation> resolve(RunnerApi.ArtifactInformation info);
+    Optional<List<RunnerApi.ArtifactInformation>> resolve(RunnerApi.ArtifactInformation info);
   }
 }
--- a/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/DefaultArtifactResolver.java
+++ b/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/DefaultArtifactResolver.java
@@ -19,6 +19,7 @@ package org.apache.beam.runners.core.construction;
 
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -39,18 +40,18 @@ public class DefaultArtifactResolver implements ArtifactResolver {
       Lists.newArrayList(
           (info) -> {
             if (BeamUrns.getUrn(RunnerApi.StandardArtifacts.Types.FILE).equals(info.getTypeUrn())) {
-              return ImmutableList.of(info);
+              return Optional.of(ImmutableList.of(info));
             } else {
-              return ImmutableList.of();
+              return Optional.empty();
             }
           });
 
   private Function<RunnerApi.ArtifactInformation, Stream<RunnerApi.ArtifactInformation>> resolver =
       (info) -> {
         for (ResolutionFn fn : Lists.reverse(fns)) {
-          List<RunnerApi.ArtifactInformation> resolved = fn.resolve(info);
-          if (!resolved.isEmpty()) {
-            return resolved.stream();
+          Optional<List<RunnerApi.ArtifactInformation>> resolved = fn.resolve(info);
+          if (resolved.isPresent()) {
+            return resolved.get().stream();
           }
         }
         throw new RuntimeException(String.format("Cannot resolve artifact information: %s", info));
--- a/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/DefaultArtifactResolver.java
+++ b/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/DefaultArtifactResolver.java
@@ -1,0 +1,92 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.beam.runners.core.construction;
+
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.apache.beam.model.pipeline.v1.RunnerApi;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.ImmutableList;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.ImmutableMap;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.Lists;
+
+/**
+ * A default artifact resolver. This resolver applies {@link ResolutionFn} in the reversed order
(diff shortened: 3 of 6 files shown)


iemejia:
retest this please

lukecwik:
Run Java PreCommit

lukecwik:
Should we add a `directory` option and role that would be to `unzip`?

robertwb:
In this case, these are directories to be added to the classpath, so it may be preferable to keep them zipped, though such a role does make sense. (This is also an example of a new "directory" type that could be "resolved" to a zipfile.)

ihji:
directory type and unzip_to role added. didn't create a interface for resolving artifacts since we only have one (resolving directory) now. maybe refactor later for supporting more resolvers.

ihji:
R: @lukecwik

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -1146,33 +1146,41 @@ message StandardArtifacts {
...
+    UNZIP_TO    = 1 [(beam_urn) = "beam:artifact:role:unzip_to:v1"];
   }
 }
 
 message ArtifactFilePayload {
-  // a string for an artifact path e.g. "/tmp/foo.jar"
+  // a string for an artifact file path e.g. "/tmp/foo.jar"
   string path = 1;
 
   // The hex-encoded sha256 checksum of the artifact.
```
lukecwik: It doesn't seem like we are populating sha256 here.

ihji: adding deferred artifacts and populating sha256.

ihji:
CC: @robertwb

ihji:
CC: @chamikaramj

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -1206,6 +1210,11 @@ message MavenPayload {
   string repository_url = 2;
 }
 
+message DeferredArtifactPayload {
+  // A id for deferred artifacts.
+  string id = 1;
```
lukecwik, line 1215: I was under the impression we were going to make this a `bytes` field so that any deferred information can get passed through and then back to the **creator** whether it be an id or a serialized blob of objects or ...

Allowing for `bytes` enables for solutions beyond in memory maps.

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/ArtifactResolver.java
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.beam.runners.core.construction;
+
+import java.util.Optional;
+import org.apache.beam.model.pipeline.v1.RunnerApi;
+
+public interface ArtifactResolver {
+  void register(ResolutionFn fn);
```
lukecwik, line 24: We won't need to rely on using Optional if we make registration take a URN and ResolutionFn and then the resolver can be found by URN.

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/ArtifactResolver.java
```diff
@@ -0,0 +1,31 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.beam.runners.core.construction;
+
+import java.util.Optional;
+import org.apache.beam.model.pipeline.v1.RunnerApi;
+
+public interface ArtifactResolver {
```
lukecwik, line 23: Please add comments to this class and methods.

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/DefaultArtifactResolver.java
```diff
@@ -0,0 +1,83 @@
...
+ * limitations under the License.
+ */
+package org.apache.beam.runners.core.construction;
+
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import org.apache.beam.model.pipeline.v1.RunnerApi;
+
+public class DefaultArtifactResolver implements ArtifactResolver {
```
lukecwik, line 25: Class comment

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/DefaultArtifactResolver.java
```diff
@@ -0,0 +1,83 @@
...
+          if (resolved.isPresent()) {
+            return resolved;
+          } else {
+            return resolver.resolve(info);
+          }
+        };
+  }
+
+  @Override
+  public RunnerApi.Pipeline resolveArtifacts(RunnerApi.Pipeline pipeline) {
```
lukecwik, line 51: The level of nesting in this method is getting a little silly.

Use local variables to logically describe what your doing and consider dropping using stream

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/Environments.java
```diff
@@ -215,22 +221,76 @@ public static Environment createProcessEnvironment(
...
-            filesToStage.add(createArtifactInformation(zipDirectory(file)));
-          } catch (IOException e) {
-            throw new RuntimeException(e);
-          }
-        } else {
-          filesToStage.add(createArtifactInformation(file));
-        }
+      // Spurious items get added to the classpath. Filter by just those that exist.
+      if (file.exists()) {
+        String id = UUID.randomUUID().toString();
```
lukecwik, line 230: Note that you could put one ID into the map for the entire list of files if you allowed ResolutionFn to return a List/Collection of artifacts.

model/pipeline/src/main/proto/beam_runner_api.proto
robertwb: +1

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/ArtifactResolver.java
robertwb: I agree. However, this might be more limited. E.g. one could have a more than one resolver per type, each of which can only resolve a subset, or alternatively one could want to resolve more than one type (or any type), e.g. a proxying resolver.

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/ArtifactResolver.java
```diff
@@ -0,0 +1,31 @@
...
+  void register(ResolutionFn fn);
+
+  RunnerApi.Pipeline resolveArtifacts(RunnerApi.Pipeline pipeline);
+
+  interface ResolutionFn {
+    Optional<RunnerApi.ArtifactInformation> resolve(RunnerApi.ArtifactInformation info);
```
robertwb, line 29: As with the resolution API, one may want to attempt to resolve multiple artifacts (e.g. maven dependencies) simultaneously. One may also need to return multiple artifacts as the resolution of a single artifact (e.g. the deferred "ambient environment" one).

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/DefaultArtifactResolver.java
```diff
@@ -0,0 +1,83 @@
...
+        if (BeamUrns.getUrn(RunnerApi.StandardArtifacts.Types.FILE).equals(info.getTypeUrn())) {
+          return Optional.of(info);
+        } else {
+          return Optional.empty();
+        }
+      };
+
+  @Override
+  public void register(ResolutionFn fn) {
+    resolver =
```
robertwb, line 39: I wonder if having an explicit List<ResolutionFn> would be easier to understand than the implicit chaining in these abstract classes.

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/Environments.java
robertwb: Yes, I think the goal is to avoid even this enumeration until we actually need it.

reuvenlax:
Friendly question: is this PR close? I'm still having to trigger Java precommits 3-4 times in order to get a green run, partially due to these timeouts.

lukecwik:
> Friendly question: is this PR close? I'm still having to trigger Java precommits 3-4 times in order to get a green run, partially due to these timeouts.

We have agreed on the solution and this PR represents an early version of it so I would say that we are close.

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/ArtifactResolver.java
ihji: done.

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/DefaultArtifactResolver.java
ihji: done.

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/DefaultArtifactResolver.java
ihji: done :smile:

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/Environments.java
ihji: done.

model/pipeline/src/main/proto/beam_runner_api.proto
ihji: Don't we need at least a key field (or urn or identifier) which can be used to check whether the payload is parsable by the creator? Otherwise, it would be pretty hard to know where the bytes payload originally came from.

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/ArtifactResolver.java
ihji: changed to List.

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -1206,6 +1210,15 @@ message MavenPayload {
   string repository_url = 2;
 }
 
+message DeferredArtifactPayload {
+  // A unique string identifier assigned by the creator of this payload. The creator may use this key to confirm
+  // whether they can parse the data.
+  string key = 1;
+
+  // A data for deferred artifacts. Interpretation of bytes is delegated to the creator of this payload.
+  bytes data = 2;
```
lukecwik, line 1219:
```suggestion
  bytes payload = 2;
```

model/pipeline/src/main/proto/beam_runner_api.proto
lukecwik: Thats a good point about the key and in general this will become a problem for all artifacts since none of them have unique keys associated with them unless an intermediary resolves the artifact immediately and possibly "renames" the contents to make it unique.

If we ever want to support multiple layers of expansion for XLang we'll want to proxy any artifact resolution/retrieval calls through the layers.

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -1206,6 +1210,15 @@ message MavenPayload {
   string repository_url = 2;
 }
 
+message DeferredArtifactPayload {
+  // A unique string identifier assigned by the creator of this payload. The creator may use this key to confirm
+  // whether they can parse the data.
+  string key = 1;
+
+  // A data for deferred artifacts. Interpretation of bytes is delegated to the creator of this payload.
```
lukecwik, line 1218:
```suggestion
  // Data for deferred artifacts. Interpretation of bytes is delegated to the creator of this payload.
```

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/DefaultArtifactResolver.java
```diff
@@ -0,0 +1,93 @@
...
+          entry
+              .getValue()
+              .getDependenciesList()
+              .parallelStream()
+              .flatMap(
+                  (info) -> {
+                    List<RunnerApi.ArtifactInformation> resolved = resolver.resolve(info);
+                    if (resolved.isEmpty()) {
+                      throw new RuntimeException(
+                          String.format("cannot resolve artifact information: %s", info));
```
lukecwik, line 72:
```suggestion
                          String.format("Cannot resolve artifact information: %s", info));
```

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/Environments.java
```diff
@@ -214,24 +220,87 @@ public static Environment createProcessEnvironment(
...
+          } else {
+            return ImmutableList.of();
+          }
+        });
+
+    return ImmutableList.of(
+        ArtifactInformation.newBuilder()
+            .setTypeUrn(BeamUrns.getUrn(StandardArtifacts.Types.DEFERRED))
+            .setTypePayload(
+                RunnerApi.DeferredArtifactPayload.newBuilder().setKey(id).build().toByteString())
```
lukecwik, line 302:
```suggestion
    String key = UUID.randomUUID().toString();
    DefaultArtifactResolver.INSTANCE.register(
        (info) -> {
          if (BeamUrns.getUrn(StandardArtifacts.Types.DEFERRED).equals(info.getTypeUrn())) {
            RunnerApi.DeferredArtifactPayload deferredArtifactPayload;
            try {
              deferredArtifactPayload =
                  RunnerApi.DeferredArtifactPayload.parseFrom(info.getTypePayload());
            } catch (InvalidProtocolBufferException e) {
              throw new RuntimeException("Error parsing deferred artifact payload.", e);
            }
            if (key.equals(deferredArtifactPayload.getKey())) {
              return lazyArtifacts.stream().map(Supplier::get).collect(Collectors.toList());
            } else {
              return ImmutableList.of();
            }
          } else {
            return ImmutableList.of();
          }
        });

    return ImmutableList.of(
        ArtifactInformation.newBuilder()
            .setTypeUrn(BeamUrns.getUrn(StandardArtifacts.Types.DEFERRED))
            .setTypePayload(
                RunnerApi.DeferredArtifactPayload.newBuilder().setKey(key).build().toByteString())
```

lukecwik:
retest this please

ihji:
@lukecwik Ready to merge. Please trigger the test and take a final look.

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/DefaultArtifactResolver.java
ihji: done.

model/pipeline/src/main/proto/beam_runner_api.proto
robertwb: As discussed (but putting here for the record) having a proxy artifact type could solve this issue.

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/DefaultArtifactResolver.java
```diff
@@ -0,0 +1,92 @@
...
+public class DefaultArtifactResolver implements ArtifactResolver {
+  public static final ArtifactResolver INSTANCE = new DefaultArtifactResolver();
+
+  private List<ResolutionFn> fns =
+      Lists.newArrayList(
+          (info) -> {
+            if (BeamUrns.getUrn(RunnerApi.StandardArtifacts.Types.FILE).equals(info.getTypeUrn())) {
+              return ImmutableList.of(info);
+            } else {
+              return ImmutableList.of();
```
robertwb, line 44: Is the empty list special? In particular sometimes a deferred artifact may resolve to nothing, which is different than not being able to be resolved... I think we still need optional or null or an exception to denote unresolveable by this resolver.

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -1206,6 +1210,15 @@ message MavenPayload {
   string repository_url = 2;
 }
 
+message DeferredArtifactPayload {
+  // A unique string identifier assigned by the creator of this payload. The creator may use this key to confirm
+  // whether they can parse the data.
+  string key = 1;
```
robertwb, line 1216: Should this be uid? Any collisions here could be bad...

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/Environments.java
```diff
@@ -214,24 +220,90 @@ public static Environment createProcessEnvironment(
       pathsToStage.addAll(stagingFiles);
     }
 
-    ImmutableList.Builder<ArtifactInformation> filesToStage = ImmutableList.builder();
+    ImmutableList.Builder<Supplier<ArtifactInformation>> lazyArtifactsBuilder =
+        ImmutableList.builder();
     for (String path : pathsToStage) {
```
robertwb, line 225: Don't we want this for loop to be lazy? 

Rather than introducing intermediate streams of Suppliers, I think we could just rename the existing `getArtifacts()` something like `getNonDeferredArtifacts()` and then call it during resolution. 

```
if (key.equals(deferredArtifactPayload.getKey())) {
      return getNonDeferredArtifacts(options);
}
```

lukecwik:
retest this please

model/pipeline/src/main/proto/beam_runner_api.proto
lukecwik: This is going to have to get revamped for XLang and since it isn't being exported outside of the SDK for portable runners we can easily change it.

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/Environments.java
ihji: This for loop is fairly cheap and from stream of Suppliers we can easily get additional performance benefits by creating parallelStream. When we consider parallelizing expensive computations, some boilerplate codes are needed anyway in `getNonDeferredArtifacts()`. I think building a stream is a nice way to abstract them out.

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/DefaultArtifactResolver.java
ihji: Done. Optional list makes three choices: Failure, Success with empty output and Success with a list of artifacts.

lukecwik:
retest this please

mxm:
Since this continues to affect many Beam developers with test timeouts, do you think we could commit an intermediate fix for this? Or perhaps revert the experimental feature until a fix has been developed?

chamikaramj:
Retest this please

chamikaramj:
Retest this please

chamikaramj:
Retest this please

ihji:
@robertwb All comments are addressed. PTAL!

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/Environments.java
```diff
@@ -207,31 +210,94 @@ public static Environment createProcessEnvironment(
...
+        artifactBuilder.setTypeUrn(BeamUrns.getUrn(StandardArtifacts.Types.FILE));
+        artifactBuilder.setRoleUrn(BeamUrns.getUrn(StandardArtifacts.Roles.STAGING_TO));
+        artifactBuilder.setRolePayload(
+            RunnerApi.ArtifactStagingToRolePayload.newBuilder()
+                .setStagedName(createStagingFileName(file))
+                .build()
+                .toByteString());
         if (file.isDirectory()) {
-          // Zip up directories so we can upload them to the artifact service.
+          File zippedFile;
```
robertwb, line 229: Nit, there seems to be a fair amount of duplication between these two case.

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/Environments.java
```diff
@@ -207,31 +210,94 @@ public static Environment createProcessEnvironment(
...
+                          .build()
+                          .toByteString())
+                  .build());
         } else {
-          filesToStage.add(createArtifactInformation(file));
+          HashCode hashCode;
+          try {
+            hashCode = Files.asByteSource(file).hash(Hashing.sha256());
+          } catch (IOException e) {
+            throw new RuntimeException(e);
```
robertwb, line 251: Or would it be better to let the method throw an IOException?

PR #11218, opened 2020-03-25
[BEAM-9605] BIP-1: Rename setRowOption to setOption on Option builder

alexvanboxel:
Rename setRowOption to setOption on Option builder as setRowOption
name is too confusing.

It sets an option as a Row, not an option on a Row. Using setOption
is better and doesn't conflict with the other setOption with 3
parameters and explicit type.

This commit also removed left over @Nullable annotations.


alexvanboxel:
Run Java PreCommit

PR #11231, opened 2020-03-26
[BEAM-4374] Shortids for the Go SDK

lostluck:
A quick unblocker for current proto changes. Adds Short ID handling for the Go SDK.

There are likely some performance improvements that can go in around the single lock, either moving them to a RW lock or a sync.Map, but this should be fine for Batch cases at present, until we measure differently.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/go/pkg/beam/core/runtime/harness/monitoring.go
+++ b/sdks/go/pkg/beam/core/runtime/harness/monitoring.go
@@ -17,6 +17,9 @@ package harness
 
 import (
 	"bytes"
+	"strconv"
+	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/apache/beam/sdks/go/pkg/beam/core/graph/coder"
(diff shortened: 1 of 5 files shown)


lostluck:
Run Go Postcommit

lostluck:
R: @lukecwik

lostluck:
This code drifts slightly from https://github.com/apache/beam/pull/11184, which removes the OneOf in favour of always using the payload format. This version still uses the oneof, but only populates payload.

sdks/go/pkg/beam/core/runtime/harness/monitoring.go
```diff
@@ -16,20 +16,71 @@
...
 	ppb "github.com/apache/beam/sdks/go/pkg/beam/model/pipeline_v1"
 	"github.com/golang/protobuf/ptypes"
 )
 
-func monitoring(p *exec.Plan) (*fnpb.Metrics, []*ppb.MonitoringInfo) {
+// TODO: 2020/03/26 - measure mutex overhead vs sync.Map for this case.
+// sync.Map might have lower contention for this read heavy load.
+var (
+	shortMu         sync.Mutex
+	labels2ShortIds map[metrics.Labels]string
```
lukecwik: This won't be enough as the key. There are things that have the same labels but different URNs, e.g start/process/finish msecs

I would suggest using the MonitoringInfo with the payload field "blanked" as the key of the map.

sdks/go/pkg/beam/core/runtime/harness/monitoring.go
```diff
@@ -16,20 +16,71 @@
...
+	shortIds2Infos  map[string]*ppb.MonitoringInfo
+
+	lastShortID int64
+)
+
+func getNextShortID() string {
+	id := atomic.AddInt64(&lastShortID, 1)
+	// No reason not to use the smallest string short ids possible.
```
lukecwik: +1

sdks/go/pkg/beam/core/runtime/harness/monitoring.go
lostluck: Ah good to know. Thanks!

sdks/go/pkg/beam/core/runtime/harness/monitoring.go
lostluck: Base36: It's what efficiency craves.

lostluck:
A rather large refactoring occurred. For one, I wasn't initializing the maps properly which correctly caused postcommit failures.

Now I've moved that verification to tests, and made the cache easier to test.

sdks/go/pkg/beam/core/runtime/harness/monitoring.go
lostluck: Ah good point. 
Can't use protos as Go Map keys, because of all the magic fields they have, but I can use other things.

I've put in aligned constants, types, and string arrays for the proto specified strings, so these lookups don't end up hashing the strings every time (and instead use a uint32, which is very fast for go maps to deal with.) There's still the hashing of the fields in metrics.Labels, but we can do the same hashing in the metrics code at a later time, to allow for faster lookups for those instead.

lostluck:
Run Go Postcommit

lostluck:
Just as a note, I'll wait until https://github.com/apache/beam/pull/11184 is in, and resolve the merge conflicts on my end before we merge this one.

sdks/go/pkg/beam/core/runtime/harness/monitoring.go
```diff
@@ -16,20 +16,165 @@
...
 	"github.com/golang/protobuf/ptypes"
 )
 
-func monitoring(p *exec.Plan) (*fnpb.Metrics, []*ppb.MonitoringInfo) {
+type mUrn uint32
+type mType uint32
+
+// TODO: Pull these from the protos.
+var sUrns = []string{
+	"beam:metric:user:v1",
```
lukecwik: heads up that this has now been exploded so that each MonitoringInfoSpec has a unique urn meaning that you'll see:
beam:metric:user:sum_int64:v1, beam:metric:user:sum_double:v1, ...

sdks/go/pkg/beam/core/runtime/harness/monitoring.go
```diff
@@ -16,20 +16,165 @@
...
+	"beam:metrics:top_n_double:v1",
+	"beam:metrics:bottom_n_int64:v1",
+	"beam:metrics:bottom_n_double:v1",
+	"beam:metrics:monitoring_table:v1",
+	"beam:metrics:progress:v1",
+
+	"TestingSentinelType", // Must remain last.
+}
+
+const (
```
lukecwik: Since the urns uniquely identify the type now, you don't need this anymore and a monitoring info is uniquely described by urn + labels.

sdks/go/pkg/beam/core/runtime/harness/monitoring_test.go
```diff
@@ -0,0 +1,122 @@
...
+	tests := []struct {
+		id           string
+		urn          mUrn
+		typ          mType
+		expectedUrn  string
+		expectedType string
+	}{
+		{
+			id:           "1",
+			urn:          urnUser,
```
lukecwik: Can you add the case where the same urn but unique labels are used gets a different short id?

sdks/go/pkg/beam/core/runtime/harness/monitoring.go
lostluck: Ack. I've kept the extra marker type around for the moment, but I might collapse things into the function to simplify some thing.

lostluck:
Run Go Postcommit

lostluck:
Run Go Postcommit

lostluck:
R: @youngoli @lukecwik 

I can't wait to get rid of both of the legacy stuff and the older monitoring info listing. That function is becoming crufty.

sdks/go/pkg/beam/core/runtime/harness/monitoring.go
```diff
@@ -28,10 +31,180 @@ import (
...
+		Urn:    sUrns[urn],
+		Type:   urnToType(urn),
+		Labels: userLabels(l),
+	}
+	return s
+}
+
+func (c *shortIDCache) shortIdsToInfos(shortids []string) map[string]*ppb.MonitoringInfo {
+	c.mu.Lock()
+	defer c.mu.Unlock()
```
TheNeuralBit, line 181: Is the locking correct here? I'd think you'd want to acquire the lock in `getShortId` since that's where the race condition could occur. This function is only reading `shortIds2Infos` which should only gain more entries over time and entries are never modified, so I'd think it would be thread-safe even without acquiring the lock. Am I missing something?

lukecwik: CC: @lostluck
Thats a good point, it looks like we should also acquire the lock in getShortID after we generate the short id.

lostluck: Also, I did comment that getShortId assumes the lock is held.

lukecwik: Thanks for the explanation again, I knew this before but blanked when @TheNeuralBit brought it up.

TheNeuralBit: Oops sorry for the noise @lostluck, thanks for the explanation

PR #11235, opened 2020-03-26
[BEAM-9618] Pull bundle descriptors.

robertwb:
------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/model/fn-execution/src/main/proto/beam_fn_api.proto
+++ b/model/fn-execution/src/main/proto/beam_fn_api.proto
@@ -81,6 +81,16 @@ service BeamFnControl {
       returns (
           // A stream of instructions requested of the SDK to be performed.
           stream InstructionRequest) {}
+
+  // Used to get the full process bundle descriptors for bundles one
+  // is asked to process.
+  rpc GetProcessBundleDescriptor(GetProcessBundleDescriptorRequest) returns (
+      ProcessBundleDescriptor) {}
+}
+
+// Requests the ProcessBundleDescriptor with the given id.
+message GetProcessBundleDescriptorRequest {
+  string process_bundle_descriptor_id = 1;
 }
 
 // A request sent by a runner which the SDK is asked to fulfill.
--- a/runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/FnApiControlClientPoolService.java
+++ b/runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/FnApiControlClientPoolService.java
@@ -19,13 +19,17 @@ package org.apache.beam.runners.fnexecution.control;
 
 import static org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Preconditions.checkState;
 
-import java.util.ArrayList;
-import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import javax.annotation.concurrent.GuardedBy;
 import org.apache.beam.model.fnexecution.v1.BeamFnApi;
 import org.apache.beam.model.fnexecution.v1.BeamFnControlGrpc;
 import org.apache.beam.runners.fnexecution.FnService;
 import org.apache.beam.runners.fnexecution.HeaderAccessor;
+import org.apache.beam.vendor.grpc.v1p26p0.io.grpc.Status;
+import org.apache.beam.vendor.grpc.v1p26p0.io.grpc.StatusException;
 import org.apache.beam.vendor.grpc.v1p26p0.io.grpc.stub.StreamObserver;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Strings;
 import org.slf4j.Logger;
@@ -39,9 +43,11 @@ public class FnApiControlClientPoolService extends BeamFnControlGrpc.BeamFnContr
   private final Object lock = new Object();
   private final ControlClientPool.Sink clientSink;
   private final HeaderAccessor headerAccessor;
+  private final ConcurrentMap<String, BeamFnApi.ProcessBundleDescriptor> processBundleDescriptors =
+      new ConcurrentHashMap<>();
 
   @GuardedBy("lock")
-  private final Collection<FnApiControlClient> vendedClients = new ArrayList<>();
+  private final Map<String, FnApiControlClient> vendedClients = new HashMap<>();
 
   @GuardedBy("lock")
   private boolean closed = false;
@@ -82,7 +88,8 @@ public class FnApiControlClientPoolService extends BeamFnControlGrpc.BeamFnContr
     }
 
     LOG.info("Beam Fn Control client connected with id {}", workerId);
-    FnApiControlClient newClient = FnApiControlClient.forRequestObserver(workerId, requestObserver);
+    FnApiControlClient newClient =
+        FnApiControlClient.forRequestObserver(workerId, requestObserver, processBundleDescriptors);
     try {
       // Add the client to the pool of vended clients before making it available - we should close
       // the client when we close even if no one has picked it up yet. This can occur after the
@@ -94,7 +101,7 @@ public class FnApiControlClientPoolService extends BeamFnControlGrpc.BeamFnContr
             !closed, "%s already closed", FnApiControlClientPoolService.class.getSimpleName());
         // TODO: https://issues.apache.org/jira/browse/BEAM-4151: Prevent stale client references
         // from leaking.
-        vendedClients.add(newClient);
+        vendedClients.put(workerId, newClient);
       }
       // We do not attempt to transactionally add the client to our internal list and offer it to
       // the sink.
@@ -106,11 +113,29 @@ public class FnApiControlClientPoolService extends BeamFnControlGrpc.BeamFnContr
   }
 
   @Override
+  public void getProcessBundleDescriptor(
+      BeamFnApi.GetProcessBundleDescriptorRequest request,
+      StreamObserver<BeamFnApi.ProcessBundleDescriptor> responseObserver) {
+    String bundleDescriptorId = request.getProcessBundleDescriptorId();
+    LOG.info("getProcessBundleDescriptor request with id {}", bundleDescriptorId);
+    BeamFnApi.ProcessBundleDescriptor descriptor = processBundleDescriptors.get(bundleDescriptorId);
+    if (descriptor == null) {
+      String msg =
+          String.format("ProcessBundleDescriptor with id %s not found", bundleDescriptorId);
+      responseObserver.onError(new StatusException(Status.NOT_FOUND.withDescription(msg)));
+      LOG.error(msg);
+    } else {
+      responseObserver.onNext(descriptor);
+      responseObserver.onCompleted();
+    }
+  }
+
+  @Override
   public void close() {
     synchronized (lock) {
       if (!closed) {
         closed = true;
-        for (FnApiControlClient vended : vendedClients) {
+        for (FnApiControlClient vended : vendedClients.values()) {
           vended.close();
         }
       }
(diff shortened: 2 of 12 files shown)


model/fn-execution/src/main/proto/beam_fn_api.proto
```diff
@@ -81,6 +81,13 @@ service BeamFnControl {
       returns (
           // A stream of instructions requested of the SDK to be performed.
           stream InstructionRequest) {}
+
+  rpc GetProcessBundleDescriptor(GetProcessBundleDescriptorRequest) returns (
```
lukecwik: I would have expected this to be a separate stream containing SDK requests similar to how Control works or made Control bidirectional for requests and responses by making a singular `Instruction` field that is streamed both ways that contains both Request and Response fields.

The advantage of using a stream is that it may be important to provide ordering over SDK requests/responses. Putting it into the Control stream allows for a total ordering between all Runner and SDK requests/responses (to solve the instruction id collision issue, we can assign servers the "s" prefix and clients the "c" prefix).

robertwb: The current Control must be a stream because (1) it is server-initiated requests on a client-initiated connection and (2) in streaming pipelines, it we want it to be high-volume, low-latency, possibly pipelined. However, using streams in this way vs. "standard" RPC methods comes at a cost (in both code and conceptual complexity). 

As far as ordering, it is preferable to have a protocol that is not sensitive to this kind of thing. (In fact, this eliminates the one need for ordering in our Control stream, where bundles needed to be registered before they were used (and this didn't fit well with pattern of dispatching instructions to threadpool-backed handlers in the SDKs which required extra locking/waiting anyway). Progress and splitting requests were already asynchronous.)

lukecwik: sg

robertwb:
R: @lukecwik this is rebased and should be ready for review. I will remove commit cebab89 before merging.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/FnApiControlClientPoolService.java
```diff
@@ -105,12 +105,38 @@ public static FnApiControlClientPoolService offeringClientsToPool(
...
   }
 
+  @Override
+  public void getProcessBundleDescriptor(
+      BeamFnApi.GetProcessBundleDescriptorRequest request,
+      StreamObserver<BeamFnApi.ProcessBundleDescriptor> responseObserver) {
+    final String workerId = headerAccessor.getSdkWorkerId();
+    if (Strings.isNullOrEmpty(workerId)) {
+      // TODO(BEAM-4149): Enforce proper worker id.
+      LOG.warn("No worker_id header provided in getProcessBundleDescriptor request");
```
lukecwik, line 115: Since this is a new method, we should force all users to supply it.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/FnApiControlClientPoolService.java
```diff
@@ -105,12 +105,38 @@ public static FnApiControlClientPoolService offeringClientsToPool(
...
+    synchronized (lock) {
+      client = vendedClients.get(workerId);
+    }
+    LOG.warn("getProcessBundleDescriptor client {}", client);
+    LOG.warn("getProcessBundleDescriptor id {}", request.getProcessBundleDescriptorId());
+    LOG.warn(
+        "getProcessBundleDescriptor result {}",
+        client.getProcessBundleDescriptor(request.getProcessBundleDescriptorId()));
+    responseObserver.onNext(
+        client.getProcessBundleDescriptor(request.getProcessBundleDescriptorId()));
```
lukecwik, line 130: Please return the NOT_FOUND status error if the process bundle descriptor is null.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/FnApiControlClientPoolService.java
```diff
@@ -105,12 +105,38 @@ public static FnApiControlClientPoolService offeringClientsToPool(
...
+  public void getProcessBundleDescriptor(
+      BeamFnApi.GetProcessBundleDescriptorRequest request,
+      StreamObserver<BeamFnApi.ProcessBundleDescriptor> responseObserver) {
+    final String workerId = headerAccessor.getSdkWorkerId();
+    if (Strings.isNullOrEmpty(workerId)) {
+      // TODO(BEAM-4149): Enforce proper worker id.
+      LOG.warn("No worker_id header provided in getProcessBundleDescriptor request");
+    }
+
+    LOG.info("getProcessBundleDescriptor request with id {}", workerId);
```
lukecwik, line 118:
```suggestion
    LOG.info("getProcessBundleDescriptor: descriptor {} worker id {}", request.getProcessBundleDescriptorId(), workerId);
```

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/FnApiControlClientPoolService.java
```diff
@@ -105,12 +105,38 @@ public static FnApiControlClientPoolService offeringClientsToPool(
...
+
+    FnApiControlClient client;
+    synchronized (lock) {
+      client = vendedClients.get(workerId);
+    }
+    LOG.warn("getProcessBundleDescriptor client {}", client);
+    LOG.warn("getProcessBundleDescriptor id {}", request.getProcessBundleDescriptorId());
+    LOG.warn(
+        "getProcessBundleDescriptor result {}",
+        client.getProcessBundleDescriptor(request.getProcessBundleDescriptorId()));
```
lukecwik, line 128: Did you intend to keep all these logs?

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/FnApiControlClientPoolService.java
```diff
@@ -105,12 +105,38 @@ public static FnApiControlClientPoolService offeringClientsToPool(
     return newClient.asResponseObserver();
   }
 
+  @Override
+  public void getProcessBundleDescriptor(
+      BeamFnApi.GetProcessBundleDescriptorRequest request,
+      StreamObserver<BeamFnApi.ProcessBundleDescriptor> responseObserver) {
+    final String workerId = headerAccessor.getSdkWorkerId();
+    if (Strings.isNullOrEmpty(workerId)) {
```
lukecwik, line 113: We should check that we are not closed.

sdks/python/apache_beam/runners/worker/sdk_worker_test.py
```diff
@@ -115,8 +115,9 @@ def _check_fn_registration_multi_request(self, *args):
           {item.id: item
            for item in process_bundle_descriptors})
 
-  def test_fn_registration(self):
-    self._check_fn_registration_multi_request((1, 4), (4, 4))
+
+#   def test_fn_registration(self):
+#     self._check_fn_registration_multi_request((1, 4), (4, 4))
```
lukecwik, line 120: if we can, we should test the registration lookup failure case.

The success case will be easily be covered by other tests.

sdks/python/apache_beam/runners/worker/worker_id_interceptor.py
```diff
@@ -48,17 +49,23 @@ def __init__(self, worker_id=None):
     if worker_id:
       self._worker_id = worker_id
 
+  def intercept_unary_unary(self, continuation, client_call_details, request):
```
lukecwik, line 52: You might as well and implement the UnaryStreamClientInterceptor and StreamUnaryClientInterceptor now.

sdks/python/apache_beam/runners/worker/worker_id_interceptor.py
```diff
@@ -48,17 +49,23 @@ def __init__(self, worker_id=None):
...
+    return self._intercept(continuation, client_call_details, request)
```
lukecwik, line 53: Hmm, each RPC will need to pass the worker_id which is a minor argument for bidi stream but not worth the complexity of it compared to the unary RPC. We also don't expect this to happen a lot.

runners/java-fn-execution/src/main/java/org/apache/beam/runners/fnexecution/control/FnApiControlClientPoolService.java
```diff
@@ -105,12 +105,38 @@ public static FnApiControlClientPoolService offeringClientsToPool(
     return newClient.asResponseObserver();
   }
 
+  @Override
+  public void getProcessBundleDescriptor(
```
lukecwik, line 109: Please add failure tests to FnApiControlClientPoolService for NOT_FOUND and missing worker id.

sdks/python/apache_beam/runners/worker/sdk_worker.py
```diff
@@ -112,7 +113,11 @@ def __init__(self,
     self._state_handler_factory = GrpcStateHandlerFactory(
         self._state_cache, credentials)
     self._profiler_factory = profiler_factory
-    self._fns = {}  # type: Dict[str, beam_fn_api_pb2.ProcessBundleDescriptor]
+    self._fns = KeyedDefaultDict(
+        lambda id: self._control_stub.GetProcessBundleDescriptor(
+            beam_fn_api_pb2.GetProcessBundleDescriptorRequest(
```
lukecwik, line 118: What appears in the default dict if GetProcessBundleDescriptor raises an error because the call fails?
Does it get propagated to the caller and it fails processing the bundle (is anything left in the map)?

sdks/python/apache_beam/runners/worker/worker_id_interceptor.py
```diff
@@ -35,7 +35,8 @@ class _ClientCallDetails(collections.namedtuple(
   pass
 
 
-class WorkerIdInterceptor(grpc.StreamStreamClientInterceptor):
+class WorkerIdInterceptor(grpc.UnaryUnaryClientInterceptor,
```
lukecwik, line 38: test?

sdks/python/apache_beam/runners/worker/sdk_worker.py
robertwb: The error gets raised and nothing gets into the map.

robertwb:
@lukecwik Was there anything more you wanted me to do on this?

sdks/python/apache_beam/runners/worker/worker_id_interceptor.py
lukecwik: Removing the worker id means that process bundle descriptors need to be unique per control server instance.

runners/java-fn-execution/src/test/java/org/apache/beam/runners/fnexecution/control/FnApiControlClientPoolServiceTest.java
```diff
@@ -120,4 +121,16 @@ public void onCompleted() {
...
+  @Test
+  public void testUnknownBundle() throws Exception {
+    BeamFnApi.GetProcessBundleDescriptorRequest request =
+        BeamFnApi.GetProcessBundleDescriptorRequest.newBuilder()
+            .setProcessBundleDescriptorId("missing")
+            .build();
+    StreamObserver<BeamFnApi.ProcessBundleDescriptor> responseObserver = mock(StreamObserver.class);
+    controlService.getProcessBundleDescriptor(request, responseObserver);
+
+    verify(responseObserver).onError(any(StatusException.class));
```
lukecwik, line 134: Validate the "not found" shows up in the status message and/or status code to ensure that we got the right StatusException

runners/java-fn-execution/src/test/java/org/apache/beam/runners/fnexecution/control/FnApiControlClientTest.java
```diff
@@ -175,4 +179,23 @@ public void testOnCloseMultipleListener() throws Exception {
     verify(mockConsumer1).accept(client);
     verify(mockConsumer2).accept(client);
   }
+
+  @Test
+  public void testRegisterNundle() {
```
lukecwik, line 184:
```suggestion
  public void testRegisterBundle() {
```

robertwb:
Fixed typo and strengthened test. 

I actually did think about GCing based on worker ID, but don't think this'll be a problem in practice as the lifetimes tend to be the same (and it is still possible to do if the server wants to keep track of what bundle processor descriptors are used by which clients, even if they don't send worker IDs, which is also being fixed).

PR #11257, opened 2020-03-28
[BEAM-9642] Create runtime invokers for SDF methods.

youngoli:
This is just a first pass, and I now realize that I can do a bunch
more to optimize this for execution time, but that will come in a
subsequent commit probably. For now, this commit adds SDF method
invokers that have functioning unit tests. Next stage will be trying to
get it to return a function, so we can avoid a lot of the unnecessary
code that's currently getting executed once per element, like the switch
statements.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/go/pkg/beam/core/runtime/exec/sdf_invokers.go
+++ b/sdks/go/pkg/beam/core/runtime/exec/sdf_invokers.go
@@ -1,0 +1,301 @@
+// Licensed to the Apache Software Foundation (ASF) under one or more
+// contributor license agreements.  See the NOTICE file distributed with
+// this work for additional information regarding copyright ownership.
+// The ASF licenses this file to You under the Apache License, Version 2.0
+// (the "License"); you may not use this file except in compliance with
+// the License.  You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package exec
+
+import (
+	"github.com/apache/beam/sdks/go/pkg/beam/core/funcx"
+	"github.com/apache/beam/sdks/go/pkg/beam/core/sdf"
+	"github.com/apache/beam/sdks/go/pkg/beam/core/util/reflectx"
+	"github.com/apache/beam/sdks/go/pkg/beam/internal/errors"
+	"reflect"
+)
+
+// This file contains invokers for SDF methods. These invokers are based off
+// exec.invoker which is used for regular DoFns. Since exec.invoker is
+// specialized for DoFns it cannot be used for SDF methods. Instead, these
+// invokers pare down the functionality to only what is essential for
+// executing SDF methods, including per-element optimizations.
+//
+// Each SDF method invoker in this file is specific to a certain method, but
+// they are all used the same way. Create an invoker with new[Method]Invoker
+// in the Up method of an exec.Unit, and then invoke it with Invoke. Finally,
+// call Reset on it when the bundle ends in FinishBundle.
+//
+// These invokers are not thread-safe.
+
+// cirInvoker is an invoker for CreateInitialRestriction.
+type cirInvoker struct {
+	fn   *funcx.Fn
+	args []interface{} // Cache to avoid allocating new slices per-element.
+	call func(elms *FullValue) (rest interface{})
+}
+
+func newCreateInitialRestrictionInvoker(fn *funcx.Fn) (*cirInvoker, error) {
+	n := &cirInvoker{
+		fn:   fn,
+		args: make([]interface{}, len(fn.Param)),
+	}
+	if err := n.initCallFn(); err != nil {
+		return nil, errors.WithContext(err, "sdf CreateInitialRestriction invoker")
+	}
+	return n, nil
+}
+
+func (n *cirInvoker) initCallFn() error {
+	// Expects a signature of the form:
+	// (key?, value) restriction
+	// TODO(BEAM-9643): Link to full documentation.
+	switch fnT := n.fn.Fn.(type) {
+	case reflectx.Func1x1:
+		n.call = func(elms *FullValue) interface{} {
+			return fnT.Call1x1(elms.Elm)
+		}
+	case reflectx.Func2x1:
+		n.call = func(elms *FullValue) interface{} {
+			return fnT.Call2x1(elms.Elm, elms.Elm2)
+		}
+	default:
+		switch len(n.fn.Param) {
+		case 1:
+			n.call = func(elms *FullValue) interface{} {
+				n.args[0] = elms.Elm
+				return n.fn.Fn.Call(n.args)[0]
+			}
+		case 2:
+			n.call = func(elms *FullValue) interface{} {
+				n.args[0] = elms.Elm
+				n.args[1] = elms.Elm2
+				return n.fn.Fn.Call(n.args)[0]
+			}
+		default:
+			return errors.Errorf("CreateInitialRestriction fn %v has unexpected number of parameters: %v",
+				n.fn.Fn.Name(), len(n.fn.Param))
+		}
+	}
+
+	return nil
+}
+
+// Invoke calls CreateInitialRestriction with the given FullValue as the element
+// and returns the resulting restriction.
+func (n *cirInvoker) Invoke(elms *FullValue) (rest interface{}) {
+	return n.call(elms)
+}
+
+// Reset zeroes argument entries in the cached slice to allow values to be
+// garbage collected after the bundle ends.
+func (n *cirInvoker) Reset() {
+	for i := range n.args {
+		n.args[i] = nil
+	}
+}
+
+// srInvoker is an invoker for SplitRestriction.
+type srInvoker struct {
+	fn   *funcx.Fn
+	args []interface{} // Cache to avoid allocating new slices per-element.
+	call func(elms *FullValue, rest interface{}) (splits interface{})
+}
+
+func newSplitRestrictionInvoker(fn *funcx.Fn) (*srInvoker, error) {
+	n := &srInvoker{
+		fn:   fn,
+		args: make([]interface{}, len(fn.Param)),
+	}
+	if err := n.initCallFn(); err != nil {
+		return nil, errors.WithContext(err, "sdf SplitRestriction invoker")
+	}
+	return n, nil
+}
+
+func (n *srInvoker) initCallFn() error {
+	// Expects a signature of the form:
+	// (key?, value, restriction) []restriction
+	// TODO(BEAM-9643): Link to full documentation.
+	switch fnT := n.fn.Fn.(type) {
+	case reflectx.Func2x1:
+		n.call = func(elms *FullValue, rest interface{}) interface{} {
+			return fnT.Call2x1(elms.Elm, rest)
+		}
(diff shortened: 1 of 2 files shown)


youngoli:
CC: @lostluck 

Like I mentioned in the commit message above, this is lacking runtime optimizations. I got the testing working and wanted to get this in a PR so you can give some general feedback so I can know if I'm on the right track. But as-is, the change isn't done yet.

My current plan for optimization is to follow the regular invoker's example and make this a method we can call in Setup, have it output a CallFn, and then we can use that in ProcessElement to actually invoke the method, without the extra boilerplate that doesn't need to be done per-element. I avoided all that for the first go-round because I didn't understand why it was being done that way, and I just wanted the simplest implementation possible to get it working.

sdks/go/pkg/beam/core/runtime/exec/sdf_invokers.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+		ret = fn.Fn.Call(inputs)[0]
+	}
+
+	// Return value is an interface{}, but we need to convert it to a []interface{}.
+	val := reflect.ValueOf(ret)
+	s := make([]interface{}, 0, val.Len())
+	for i := 0; i < val.Len(); i++ {
+		s = append(s, val.Index(i).Interface())
+	}
```
lostluck: This feels like the only bit that doesn't fit into the existing invokers. It would just need an additional wrapper to handle this part (since it's equivalent to the single return value kind), though handling the *FullValue we get back might be less clean. Hmmm.

sdks/go/pkg/beam/core/runtime/exec/sdf_invokers.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func invokeSplitRestriction(fn *funcx.Fn, elms *FullValue, rest interface{}) ([]interface{}, error) {
+	if fn == nil {
+		return nil, nil
+	}
+
+	var ret interface{}
+	switch fnT := fn.Fn.(type) {
```
lostluck: Just the concrete note that the reason the existing invokers are written the way they are, is so every time they're called, they don't need to make these choices over again. This is a WIP PR, so no worries.

sdks/go/pkg/beam/core/runtime/exec/sdf_invokers.go
```diff
@@ -0,0 +1,131 @@
...
+		default:
+			err := errors.Errorf("fn %v has unexpected number of inputs: %v", fn.Fn.Name(), fn.Fn.Type().NumIn())
+			return nil, errors.WithContext(err, "sdf CreateInitialRestriction invoker")
+		}
+		rets := fn.Fn.Call(inputs)
+		return rets[0], nil
+	}
+}
+
+func invokeSplitRestriction(fn *funcx.Fn, elms *FullValue, rest interface{}) ([]interface{}, error) {
```
lostluck: It occurs to me that we don't seem to have the abstract signatures documented for the "Provider" methods like SplitRestriction. IIRC we avoided doing that for now to avoid accidental usage? It might be good to document the expectation here at least, for this and the other SDF methods. eg. SplitRestriction(K?, V, RT) []RT

sdks/go/pkg/beam/core/runtime/exec/sdf_invokers.go
lostluck: I won't worry about it until after optimization, where it might be clearer what we can re-use. This is presently nicely factored, so unless we can make it shorter/clearer with a reuse of DoFn invoker (possibly by adding other helpers/generated methods, etc or otherwise), this is fine.

sdks/go/pkg/beam/core/runtime/exec/sdf_invokers_test.go
```diff
@@ -0,0 +1,262 @@
...
+
+type Sdf struct {
+}
+
+// Creates a restriction with the given value.
+func (fn *Sdf) CreateInitialRestriction(i int) Restriction {
+	return Restriction{i}
+}
+
+// Outputs two restrictions, the first containing the sum of i and rest.Val, the
```
lostluck: While this type and methods are only for test purposes, please adhere to the Go documentation idiom of having the method name in full to start the comment.

sdks/go/pkg/beam/core/runtime/exec/sdf_invokers.go
youngoli: Yeah, I was going to document the abstract signatures in the frontend comment for SDF, which will most likely be in the same comment that currently documents DoFns (I think it's in beam/pardo.go).

That said, I can definitely leave a comment here, with a TODO to replace it when the frontend documentation is in.

youngoli:
R: @lostluck 

Ready for an actual review. It should be much more optimized now. I have a call function so the code won't repeatedly hit the switch statements, and I have a cache for input elements (which is skipped altogether if the user generated code anyway).

I did consider ways I could use the existing invoker code, but there were so many places that assumed a normal DoFn was being used that it would be very difficult to use it without doing a good amount of refactoring. I think if the regular DoFn methods and the SDF methods ever get a standardized format for their signatures that covers both of them, then we can easily combine it, but until then I think this is preferable.

Only other part I was debating was whether it was worth trying to find a way to have one single sdfInvoker instead of a separate invoker for each method. Ultimately I felt like it was a lot of extra work to avoid a little bit of duplication (only the Invoke and Reset methods could be deduplicated, the initCallFn methods would still be 4 different methods).

sdks/go/pkg/beam/core/runtime/exec/sdf_invokers.go
```diff
@@ -0,0 +1,302 @@
...
+			n.args[0] = rest
+			return n.fn.Fn.Call(n.args)[0].(sdf.RTracker)
+		}
+	}
+	return nil
+}
+
+// Invoke calls CreateTracker given a restriction and returns an sdf.RTracker.
+// the associated restriction, and returns a size.
+func (n *ctInvoker) Invoke(rest interface{}) (tracker sdf.RTracker) {
```
lostluck, line 292: As a rule, naming return values where the meaning is already clear, and you aren't needing to modify the return in a defer is considered an anti pattern, unless it's a primitive type like string or a number, where the meaning of the type isn't necessarily clear.

It's also not necessary to do just because other functions/methods in the same file are doing it. There's no consistency requirement for it, and it can make things more confusing.

In short, drop the tracker, it's cleaner. :)

sdks/go/pkg/beam/core/runtime/exec/sdf_invokers.go
```diff
@@ -0,0 +1,302 @@
...
+		n.call = func(rest interface{}) sdf.RTracker {
+			n.args[0] = rest
+			return n.fn.Fn.Call(n.args)[0].(sdf.RTracker)
+		}
+	}
+	return nil
+}
+
+// Invoke calls CreateTracker given a restriction and returns an sdf.RTracker.
+// the associated restriction, and returns a size.
```
lostluck, line 291: This documentation line is copypasta I think.

sdks/go/pkg/beam/core/runtime/exec/sdf_invokers.go
```diff
@@ -0,0 +1,302 @@
...
+			return errors.Errorf("SplitRestriction fn %v has unexpected number of parameters: %v",
+				n.fn.Fn.Name(), len(n.fn.Param))
+		}
+	}
+	return nil
+}
+
+// Invoke calls SplitRestriction given a FullValue containing an element and
+// the associated restriction, and returns a slice of split restrictions.
+func (n *srInvoker) Invoke(elms *FullValue, rest interface{}) (split []interface{}) {
```
lostluck, line 162: eg. This is a great use of a named return value to add clarity. 

I'd call it "splits" though, since it's a slice and there could be more than one.

sdks/go/pkg/beam/core/runtime/exec/sdf_invokers_test.go
```diff
@@ -0,0 +1,282 @@
...
+
+package exec
+
+import (
+	"github.com/apache/beam/sdks/go/pkg/beam/core/graph"
+	"github.com/google/go-cmp/cmp"
+	"testing"
+)
+
+func init() {
```
lostluck, line 24: remove empty init block?

sdks/go/pkg/beam/core/runtime/exec/sdf_invokers.go
```diff
@@ -23,29 +23,36 @@ import (
 	"reflect"
 )
 
-func invokeCreateInitialRestriction(fn *funcx.Fn, elms *FullValue) (rest interface{}, err error) {
+func createInitialRestrictionCallFn(fn *funcx.Fn) (callFn func(*FullValue) interface{}, err error) {
```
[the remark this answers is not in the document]
youngoli: It's actually returning a func (which returns an interface{}), and an error. I can see the confusion though, might be worth wrapping the return value in parentheses just for clarity.

sdks/go/pkg/beam/core/runtime/exec/sdf_invokers.go
```diff
@@ -0,0 +1,302 @@
...
+	"github.com/apache/beam/sdks/go/pkg/beam/core/funcx"
+	"github.com/apache/beam/sdks/go/pkg/beam/core/sdf"
+	"github.com/apache/beam/sdks/go/pkg/beam/core/util/reflectx"
+	"github.com/apache/beam/sdks/go/pkg/beam/internal/errors"
+	"reflect"
+)
+
+// This file contains invokers for SDF methods. These invokers are based off
+// exec.invoker which is used for regular DoFns. Since exec.invoker is
+// specialized for DoFns it cannot be used for SDF methods. Instead, these
```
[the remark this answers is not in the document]
youngoli: I'd support that if we can refactor the DoFn-specific behavior out of the core invoker code so the SDF and DoFn invokers can both share that code, while adding on their own unique behavior via ret functions and such.

(Just to be clear, by core invoker code I mean the switch statements in fn_arity, and the cached arg slice. I.E. both things that I had to replicate and could potentially be shared, but right now have DoFn-specific logic associated with them.)

sdks/go/pkg/beam/core/runtime/exec/sdf_invokers.go
```diff
@@ -0,0 +1,302 @@
...
+func (n *cirInvoker) Reset() {
+	for i := range n.args {
+		n.args[i] = nil
+	}
+}
+
+// srInvoker is an invoker for SplitRestriction.
+type srInvoker struct {
+	fn   *funcx.Fn
+	args []interface{} // Cache to avoid allocating new slices per-element.
```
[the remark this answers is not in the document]
youngoli: Ack. Seems like a good idea to get in at some point

sdks/go/pkg/beam/core/runtime/exec/sdf_invokers.go
youngoli: Done.

sdks/go/pkg/beam/core/runtime/exec/sdf_invokers.go
youngoli: Yep, removed.

sdks/go/pkg/beam/core/runtime/exec/sdf_invokers.go
youngoli: Agreed, "splits" makes more sense.

sdks/go/pkg/beam/core/runtime/exec/sdf_invokers_test.go
youngoli: Whoops, yeah. Removed.

PR #11258, opened 2020-03-29
[BEAM-9571] Add withMongoClientProvider to MongoDbIO

iemejia:
This introduces the option so users can configure their MongoClients at will, as discussed in #11186 PTAL
R: @alexvanboxel

### DIFF
--- a/sdks/java/io/mongodb/src/main/java/org/apache/beam/sdk/io/mongodb/MongoDbIO.java
+++ b/sdks/java/io/mongodb/src/main/java/org/apache/beam/sdk/io/mongodb/MongoDbIO.java
@@ -24,8 +24,6 @@ import com.google.auto.value.AutoValue;
 import com.mongodb.BasicDBObject;
 import com.mongodb.MongoBulkWriteException;
 import com.mongodb.MongoClient;
-import com.mongodb.MongoClientOptions;
-import com.mongodb.MongoClientURI;
 import com.mongodb.client.AggregateIterable;
 import com.mongodb.client.MongoCollection;
 import com.mongodb.client.MongoCursor;
@@ -39,7 +37,6 @@ import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
-import javax.net.ssl.SSLContext;
 import org.apache.beam.sdk.annotations.Experimental;
 import org.apache.beam.sdk.annotations.Experimental.Kind;
 import org.apache.beam.sdk.coders.Coder;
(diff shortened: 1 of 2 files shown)


sdks/java/io/mongodb/src/main/java/org/apache/beam/sdk/io/mongodb/MongoDbIO.java
```diff
@@ -166,7 +165,7 @@ private MongoDbIO() {}
 
     abstract SerializableFunction<MongoCollection<Document>, MongoCursor<Document>> queryFn();
 
-    abstract Builder builder();
+    abstract Builder toBuilder();
```
iemejia: This rename is slightly unreated to this PR but an improvement to be consistent with [AutoValue conventions](https://github.com/google/auto/blob/master/value/userguide/builders-howto.md#to_builder).

iemejia:
Run Java PreCommit

sdks/java/io/mongodb/src/main/java/org/apache/beam/sdk/io/mongodb/MongoDbIO.java
```diff
@@ -191,7 +193,29 @@ private MongoDbIO() {}
       abstract Builder setQueryFn(
           SerializableFunction<MongoCollection<Document>, MongoCursor<Document>> queryBuilder);
 
-      abstract Read build();
```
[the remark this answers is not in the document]
iemejia: This one was not renamed, it changed because now it has a default implementation (see below in line 208). The default implementation sets a provider given the respective parameters.

iemejia:
Thanks for the review @alexvanboxel You have a point about the javadoc. I will add that.

sdks/java/io/mongodb/src/main/java/org/apache/beam/sdk/io/mongodb/MongoDbIO.java
```diff
@@ -843,22 +843,39 @@ private MongoClient createClient(Read spec) {
...
     }
 
     /** Enable invalidHostNameAllowed for ssl for connection. */
     public Write withSSLInvalidHostNameAllowed(boolean invalidHostNameAllowed) {
-      return builder().setSslInvalidHostNameAllowed(invalidHostNameAllowed).build();
+      return toBuilder().setSslInvalidHostNameAllowed(invalidHostNameAllowed).build();
+    }
+
+    /** Sets a MongoClient provider. */
+    public Write withMongoClientProvider(
```
alexvanboxel, line 875: Maybe we should rename this to withClientProvider, the we can use this pattern in other IO's and make it as part of a IO style guide to make them more consistent. WDYT?

PR #11259, opened 2020-03-29
Use attachValues in SQL

reuvenlax:
Use the more efficient attachValues when building rows in SQL. addValues does a deep copy of the parameter, while addValue does not. We have also recently made API improvements that makes addValues slightly less efficient, so this allows SQL to bypass that.

We are not using attachValues on any of the IOs. This is strictly for internal rows generated by SQL.

### DIFF
--- a/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamAggregationRel.java
+++ b/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamAggregationRel.java
@@ -24,6 +24,7 @@ import static org.apache.beam.vendor.calcite.v1_20_0.com.google.common.base.Prec
 import java.io.Serializable;
 import java.util.List;
 import javax.annotation.Nullable;
+import org.apache.beam.sdk.extensions.sql.impl.BeamSqlPipelineOptions;
 import org.apache.beam.sdk.extensions.sql.impl.planner.BeamCostModel;
 import org.apache.beam.sdk.extensions.sql.impl.planner.NodeStats;
 import org.apache.beam.sdk.extensions.sql.impl.transform.agg.AggregationCombineFnAdapter;
@@ -280,9 +281,13 @@ public class BeamAggregationRel extends Aggregate implements BeamRelNode {
         ignoreValues = true;
       }
 
+      boolean verifyRowValues =
+          pinput.getPipeline().getOptions().as(BeamSqlPipelineOptions.class).getVerifyRowValues();
       return windowedStream
           .apply(combiner)
-          .apply("mergeRecord", ParDo.of(mergeRecord(outputSchema, windowFieldIndex, ignoreValues)))
+          .apply(
+              "mergeRecord",
+              ParDo.of(mergeRecord(outputSchema, windowFieldIndex, ignoreValues, verifyRowValues)))
           .setRowSchema(outputSchema);
     }
 
@@ -324,7 +329,10 @@ public class BeamAggregationRel extends Aggregate implements BeamRelNode {
     }
 
     static DoFn<Row, Row> mergeRecord(
-        Schema outputSchema, int windowStartFieldIndex, boolean ignoreValues) {
+        Schema outputSchema,
+        int windowStartFieldIndex,
+        boolean ignoreValues,
+        boolean verifyRowValues) {
       return new DoFn<Row, Row>() {
         @ProcessElement
         public void processElement(
@@ -343,7 +351,11 @@ public class BeamAggregationRel extends Aggregate implements BeamRelNode {
             fieldValues.add(windowStartFieldIndex, ((IntervalWindow) window).start());
           }
 
-          o.output(Row.withSchema(outputSchema).attachValues(fieldValues));
+          Row row =
+              verifyRowValues
+                  ? Row.withSchema(outputSchema).addValues(fieldValues).build()
+                  : Row.withSchema(outputSchema).attachValues(fieldValues);
+          o.output(row);
         }
       };
     }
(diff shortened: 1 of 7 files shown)


reuvenlax:
run sql postcommit

alexvanboxel:
> Use the more efficient attachValues when building rows in SQL. addValues does a deep copy of the parameter, while addValue does not. We have also recently made API improvements that makes addValues slightly less efficient, so this allows SQL to bypass that.
> 
> We are not using attachValues on any of the IOs. This is strictly for internal rows generated by SQL.

I've changed from `while attachValues does not` too `while addValue does not`. (typo)

apilloud:
Actually it looks like this completely broke BeamCalcRel. You should fix that too.

reuvenlax:
run sql postcommit

apilloud:
FYI, I was going through our internal test failures and found a bunch that will probably be much more difficult to debug after this change. I included a sample of the current failures below, it would be nice to have a way to turn `SchemaVerification` on for testing.

```
Caused by: java.lang.IllegalArgumentException: FieldType{typeName=INT64, nullable=false, logicalType=null, collectionElementType=null, mapKeyType=null, mapValueType=null, rowSchema=null, metadata={}} is not nullable in Array field $arrayx
	at org.apache.beam.sdk.values.SchemaVerification.verifyArray(SchemaVerification.java:101)
	at org.apache.beam.sdk.values.SchemaVerification.verifyFieldValue(SchemaVerification.java:65)
	at org.apache.beam.sdk.values.SchemaVerification.verifyRowValues(SchemaVerification.java:57)
	at org.apache.beam.sdk.values.Row$Builder.build(Row.java:673)
	at org.apache.beam.sdk.extensions.sql.zetasql.BeamZetaSqlCalcRel$CalcFn.processElement(BeamZetaSqlCalcRel.java:195)
```

reuvenlax:
@apilloud I added a pipeline option that can be used to enable more verbose Row checking. Dataflow can enable this on its internal tests.

apilloud:
LGTM

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamAggregationRel.java
```diff
@@ -343,7 +351,11 @@ public void processElement(
             fieldValues.add(windowStartFieldIndex, ((IntervalWindow) window).start());
           }
 
-          o.output(Row.withSchema(outputSchema).addValues(fieldValues).build());
+          Row row =
+              verifyRowValues
+                  ? Row.withSchema(outputSchema).addValues(fieldValues).build()
+                  : Row.withSchema(outputSchema).attachValues(fieldValues);
```
TheNeuralBit, line 357: I think this is missing a `.build()`

TheNeuralBit:
Run SQL PostCommit

sdks/java/extensions/sql/zetasql/src/main/java/org/apache/beam/sdk/extensions/sql/zetasql/BeamZetaSqlCalcRel.java
```diff
@@ -185,14 +193,19 @@ public void processElement(ProcessContext c) {
...
-            ZetaSqlUtils.zetaSqlValueToJavaObject(v, outputSchema.getField(i).getType()));
+        values.add(
+            ZetaSqlUtils.zetaSqlValueToJavaObject(
+                v, outputSchema.getField(i).getType(), verifyRowValues));
       }
-      c.output(output.build());
+      Row outputRow =
+          verifyRowValues
+              ? Row.withSchema(outputSchema).addValues(values).build()
+              : Row.withSchema(outputSchema).attachValues(values);
```
TheNeuralBit, line 207: and here

sdks/java/extensions/sql/zetasql/src/main/java/org/apache/beam/sdk/extensions/sql/zetasql/ZetaSqlUtils.java
```diff
@@ -227,18 +229,25 @@ private static Instant zetaSqlTimestampValueToJodaInstant(Value timestampValue)
...
     for (int i = 0; i < values.size(); i++) {
-      objects.add(zetaSqlValueToJavaObject(values.get(i), schema.getField(i).getType()));
+      objects.add(
+          zetaSqlValueToJavaObject(values.get(i), schema.getField(i).getType(), verifyValues));
     }
-    return Row.withSchema(schema).addValues(objects).build();
+    Row row =
+        verifyValues
+            ? Row.withSchema(schema).addValues(objects).build()
+            : Row.withSchema(schema).attachValues(objects);
```
TheNeuralBit, line 250: and here

TheNeuralBit:
PostCommit doesn't get it either..

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamAggregationRel.java
reuvenlax: pr/10883 changed attachValues() to return a Row, so build() not needed.

sdks/java/extensions/sql/zetasql/src/main/java/org/apache/beam/sdk/extensions/sql/zetasql/BeamZetaSqlCalcRel.java
reuvenlax: ditto

sdks/java/extensions/sql/zetasql/src/main/java/org/apache/beam/sdk/extensions/sql/zetasql/ZetaSqlUtils.java
reuvenlax: ditto

PR #11229, opened 2020-03-25
[BEAM-9608] Increasing scope of context managers for FnApiRunner

pabloem:
**Please** add a meaningful description for your change here

------------------------

Thank you for your contribution! Follow this checklist to help us incorporate your contribution quickly and easily:

 - [ ] [**Choose reviewer(s)**](https://beam.apache.org/contribute/#make-your-change) and mention them in a comment (`R: @username`).
 - [ ] Format the pull request title like `[BEAM-XXX] Fixes bug in ApproximateQuantiles`, where you replace `BEAM-XXX` with the appropriate JIRA issue, if applicable. This will automatically link the pull request to the issue.
 - [ ] Update `CHANGES.md` with noteworthy changes.
 - [ ] If this contribution is large, please file an Apache [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

See the [Contributor Guide](https://beam.apache.org/contribute) for more tips on [how to make review process smoother](https://beam.apache.org/contribute/#make-reviewers-job-easier).

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

Lang | SDK | Apex | Dataflow | Flink | Gearpump | Samza | Spark
--- | --- | --- | --- | --- | --- | --- | ---
Go | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go/lastCompletedBuild/) | --- | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_VR_Flink/lastCompletedBuild/) | --- | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_VR_Spark/lastCompletedBuild/)
[...]

PR #11270, opened 2020-03-30
[BEAM-9639][BEAM-9608] Improvements for FnApiRunner

pabloem:
r: @robertwb 

These changes are mostly reshuffling of code.

Each commit is a logical unit, so each commit can be reviewed separately. The commit message explains what each commit does:

- **commit**[BEAM-9608] BundleManagers use BundleContextManager for configuration - (this commit modifies the BundleManagers to receive only a BundleContextManager with most of their configuration. It also adds a `dry_run` option for processing bundles without writing to `pcoll_buffers`)
- **commit**[BEAM-9639] Storing side inputs after producer execution, not before consumption - (this commit ensures that side inputs are stored in state right after they are computed - not before they are consumed. this will be useful in streaming so each bundle's inputs are eagerly available)
  - **commit**Ensuring downstream side inputs are calculated on fully expanded graph - (this commit is more of an accessory to the previous one. It ensures that during graph translations, downstream_side_inputs are annotated after SDFs are expanded)
- **commit**[BEAM-9639] Separate Stage and Bundle execution. Improve typing. - (this commit separates the sections of the code for executing a stage such as context creation, committing of side inputs, scheduling of all bundles until no deferred inputs vs the sections of executing a bundle for that stage such as pushing data to worker, collecting bundle deferred inputs)

---
## Notes/todos from https://github.com/apache/beam/pull/11229

- [ ] It's odd that _run_stage no longer takes as a parameter the stage to run. Perhaps bundle_context_manager (and its class?) should be named stage_context or similar?
- [ ] On this note, perhaps it makes sense to break FnApiRunner into the (mostly stateless) runner that can execute multiple pipelines and an executor (that has methods like run_stage) that might be stateful and is initialized with and tasked with running a single pipeline. Much of what is on context(s) would become state of self of this new object.
- [ ] I was bitten by the fact that it is an error to access the process_bundle_descriptor before _extract_outputs is called. This should be clearly documented (and similarly for the other lazy attributes(s) in this class). Given that that's called external to this class and can't easily be checked, makes me wonder if the boundary of encapsulation needs to be adjusted here.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/runners/portability/fn_api_runner/execution.py
+++ b/sdks/python/apache_beam/runners/portability/fn_api_runner/execution.py
@@ -22,6 +22,8 @@ from __future__ import absolute_import
 import collections
 import itertools
 from typing import TYPE_CHECKING
+from typing import Dict
+from typing import List
 from typing import MutableMapping
 
 from typing_extensions import Protocol
@@ -29,10 +31,15 @@ from typing_extensions import Protocol
 from apache_beam import coders
 from apache_beam.coders.coder_impl import create_InputStream
 from apache_beam.coders.coder_impl import create_OutputStream
+from apache_beam.coders.coders import BytesCoder
+from apache_beam.coders.coders import GlobalWindowCoder
+from apache_beam.coders.coders import WindowedValueCoder
 from apache_beam.portability import common_urns
 from apache_beam.portability.api import beam_fn_api_pb2
 from apache_beam.portability.api import beam_runner_api_pb2
 from apache_beam.runners import pipeline_context
+from apache_beam.runners.portability.fn_api_runner import translations
+from apache_beam.runners.portability.fn_api_runner.translations import create_buffer_id
 from apache_beam.runners.portability.fn_api_runner.translations import only_element
 from apache_beam.runners.portability.fn_api_runner.translations import split_buffer_id
 from apache_beam.runners.portability.fn_api_runner.translations import unique_name
@@ -45,8 +52,15 @@ from apache_beam.utils import windowed_value
 
 if TYPE_CHECKING:
   from apache_beam.coders.coder_impl import CoderImpl
-  from apache_beam.runners.portability.fn_api_runner import translations
+  from apache_beam.portability.api import beam_runner_api_pb2
   from apache_beam.runners.portability.fn_api_runner import worker_handlers
+
+ENCODED_IMPULSE_VALUE = WindowedValueCoder(
+    BytesCoder(), GlobalWindowCoder()).get_impl().encode_nested(
+        GlobalWindows.windowed_value(b''))
+
+DataOutput = Dict[str, bytes]
+SideInputAccessPattern = beam_runner_api_pb2.FunctionSpec
 
 
 class Buffer(Protocol):
@@ -256,6 +270,7 @@ class FnApiRunnerExecutionContext(object):
        ``beam.PCollection``.
  """
   def __init__(self,
+      stages,  # type: List[translations.Stage]
       worker_handler_manager,  # type: worker_handlers.WorkerHandlerManager
       pipeline_components,  # type: beam_runner_api_pb2.Components
       safe_coders,
@@ -268,6 +283,9 @@ class FnApiRunnerExecutionContext(object):
     :param safe_coders:
     :param data_channel_coders:
     """
+    self.stages = stages
+    self.side_input_descriptors_by_stage = (
+        self._build_data_side_inputs_map(stages))
     self.pcoll_buffers = {}  # type: MutableMapping[bytes, PartitionableBuffer]
     self.timer_buffers = {}  # type: MutableMapping[bytes, ListBuffer]
     self.worker_handler_manager = worker_handler_manager
@@ -279,6 +297,36 @@ class FnApiRunnerExecutionContext(object):
         self.pipeline_components,
         iterable_state_write=self._iterable_state_write)
     self._last_uid = -1
+
+  @staticmethod
+  def _build_data_side_inputs_map(stages):
+    # type: (Iterable[translations.Stage]) -> MutableMapping[str, DataSideInput]
+    """Builds an index mapping stages to side input descriptors.
+
+    A side input descriptor is a map of side input IDs to side input access
+    patterns for all of the outputs of a stage that will be consumed as a
+    side input.
+    """
+    data_side_inputs_by_stage = {}
+    all_side_inputs, stage_consumers, transform_consumers = (
+        translations.get_all_side_inputs_and_consumers(stages))
+
+    for s in stages:
+      data_side_inputs_by_stage[s.name] = {}
+      for transform in s.transforms:
+        for output in transform.outputs.values():
+          if output in all_side_inputs:
+            for consuming_transform in transform_consumers[output]:
+              payload = proto_utils.parse_Bytes(consuming_transform.spec.payload,
+                                                beam_runner_api_pb2.ParDoPayload)
+              for si_tag in payload.side_inputs:
+                side_input_id = (consuming_transform.unique_name, si_tag)
+                data_side_inputs_by_stage[s.name][side_input_id] = (
+                    consuming_transform.inputs[si_tag],
+                    payload.side_inputs[si_tag].access_pattern)
+
+    return data_side_inputs_by_stage
+
 
   @property
   def state_servicer(self):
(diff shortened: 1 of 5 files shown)


pabloem:
Run Portable_Python PreCommit

pabloem:
Run Python PreCommit

pabloem:
this latest commit is purely aesthetic

pabloem:
Run Python PreCommit

pabloem:
@robertwb ptal

pabloem:
attempting to rebase. let's see how far that takes me...

pabloem:
Run Python PreCommit

pabloem:
Run Python PreCommit

pabloem:
Run Python PreCommit

sdks/python/apache_beam/runners/portability/fn_api_runner/execution.py
```diff
@@ -367,6 +413,73 @@ def _build_process_bundle_descriptor(self):
...
+    Only the outputs that should be side inputs are committed to state.
+    """
+    data_side_input = {}  # type: DataSideInput
+    for pcoll, si_ids in self.stage.downstream_side_inputs.items():
+      for (consumer_transform_name, tag), access_pattern in si_ids.items():
+        data_side_input[consumer_transform_name, tag] = (
+            translations.create_buffer_id(pcoll), access_pattern)
+    self.execution_context.commit_side_inputs_to_state(data_side_input)
+
+  def extract_bundle_inputs(self):
```
robertwb, line 440: extract_bundle_inputs_and_outputs?

sdks/python/apache_beam/runners/portability/fn_api_runner/execution.py
```diff
@@ -367,6 +413,73 @@ def _build_process_bundle_descriptor(self):
...
+          self.data_api_service_descriptor()
+        if data_api_service_descriptor:
+          data_spec.api_service_descriptor.url = (
+              data_api_service_descriptor.url)
+        transform.spec.payload = data_spec.SerializeToString()
+      elif transform.spec.urn in translations.PAR_DO_URNS:
+        for timer_family_id in payload.timer_family_specs.keys():
+          expected_timer_output[(transform.unique_name, timer_family_id)] = (
+              create_buffer_id(timer_family_id, 'timers'))
+    return data_input, data_output, expected_timer_output
```
robertwb, line 494: Update docs to match.

sdks/python/apache_beam/runners/portability/fn_api_runner/execution.py
```diff
@@ -367,6 +413,73 @@ def _build_process_bundle_descriptor(self):
         state_api_service_descriptor=self.state_api_service_descriptor(),
         timer_api_service_descriptor=self.data_api_service_descriptor())
 
+  def commit_output_views_to_state(self):
```
robertwb, line 428: Not sure what "output views" means. Maybe call this commit_side_inputs_to_state as well?

sdks/python/apache_beam/runners/portability/fn_api_runner/translations.py
```diff
@@ -149,15 +155,26 @@ def no_overlap(a, b):
...
+        res[si] = other_si_ids
+    return res
 
   def fuse(self, other):
     # type: (Stage) -> Stage
     return Stage(
         "(%s)+(%s)" % (self.name, other.name),
         self.transforms + other.transforms,
-        union(self.downstream_side_inputs, other.downstream_side_inputs),
+        self._fuse_downstream_side_inputs(other),
```
robertwb, line 175: Nit: this sounds like it mutates self.

sdks/python/apache_beam/runners/portability/fn_api_runner/translations.py
```diff
@@ -75,21 +75,27 @@
 
 IMPULSE_BUFFER = b'impulse'
 
+# SideInputId is identified by a consumer ParDo + tag.
+SideInputId = Tuple[str, str]
+
+DataSideInput = Dict[SideInputId,
```
robertwb: What does the value represent?

sdks/python/apache_beam/runners/portability/fn_api_runner/translations.py
```diff
@@ -149,15 +155,26 @@ def no_overlap(a, b):
...
+        not consumer.is_runner_urn(context) and no_overlap(
+            set(self.downstream_side_inputs.keys()),
+            {i
+             for i, _, _ in consumer.side_inputs()}))
+
+  def _fuse_downstream_side_inputs(self, other):
+    res = dict(self.downstream_side_inputs)
+    for si, other_si_ids in other.downstream_side_inputs.items():
+      if si in res:
+        res[si] = union(res[si], other_si_ids)
```
robertwb, line 165: So this is actually a dict mapping to sets?

sdks/python/apache_beam/runners/portability/fn_api_runner/translations.py
```diff
@@ -75,21 +75,27 @@
...
+                     Tuple[bytes, beam_runner_api_pb2.FunctionSpec]]
+
 
 class Stage(object):
   """A set of Transforms that can be sent to the worker for processing."""
   def __init__(self,
                name,  # type: str
                transforms,  # type: List[beam_runner_api_pb2.PTransform]
-               downstream_side_inputs=None,  # type: Optional[FrozenSet[str]]
+               downstream_side_inputs=None,  # type: Optional[Dict[str, SideInputId]]
```
robertwb, line 88: The goal of this (which, yes, should have been better documented) is to quickly be able to prohibit fusion. But the reason we defined our own union was so that memory didn't grow as O(n^2) in the common case because many stages were able to share this set (rather than have their own copy). These changes seem to break that. 

Also, could you clarify why this was made into a dict?

sdks/python/apache_beam/runners/portability/fn_api_runner/fn_runner.py
```diff
@@ -914,14 +898,16 @@ def process_bundle(self,
                      expected_outputs,  # type: DataOutput
                      fired_timers,  # type: Mapping[Tuple[str, str], PartitionableBuffer]
                      expected_output_timers  # type: Dict[str, Dict[str, str]]
+                     dry_run=False
```
robertwb: This should be the default, we shouldn't have to pass it.

sdks/python/apache_beam/runners/portability/fn_api_runner/translations.py
pabloem: woah this is a bug.
downstream side input is a dictionary mapping to dictionaries.

Dict[Output Pcollection, Dict[Side input ID, Access pattern]]
Where Side input ID is Tuple[consumer ptransform, input index]. Added appropriate annotations, and fixed the bug.

sdks/python/apache_beam/runners/portability/fn_api_runner/translations.py
pabloem: The value is a tuple with the encoded data. I've moved these to translations.py, and updated the comment

sdks/python/apache_beam/runners/portability/fn_api_runner/translations.py
pabloem: renamed to `_get_fused_downstream_side_inputs` thoughts?

sdks/python/apache_beam/runners/portability/fn_api_runner/execution.py
pabloem: done

sdks/python/apache_beam/runners/portability/fn_api_runner/execution.py
pabloem: done

sdks/python/apache_beam/runners/portability/fn_api_runner/execution.py
pabloem: done

sdks/python/apache_beam/runners/portability/fn_api_runner/execution.py
```diff
@@ -326,8 +327,8 @@ def commit_side_inputs_to_state(
       data_side_input,  # type: DataSideInput
   ):
     # type: (...) -> None
-    for (consuming_transform_id, tag), (buffer_id, func_spec) \
-        in data_side_input.items():
+    for (consuming_transform_id, tag), (buffer_id,
```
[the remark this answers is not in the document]
pabloem: it did not : ( hehe

sdks/python/apache_beam/runners/portability/fn_api_runner/fn_runner.py
pabloem: Done

sdks/python/apache_beam/runners/portability/fn_api_runner/translations.py
pabloem: Hm so this change breaks that, so the memory requirements would be larger. I would think that they would not be too bad, since most graphs don't have many side inputs going many places. What do you think? I'm willing to find a better solution for this, but I wonder if it's worth the extra time.

The reason that this is made into a dict is to contain more information about downstream side inputs. specifically, it contains which transforms will consume the side inputs. this is used to commit the side inputs to state after they are calculated (rather than before they are consumed). This will be necessary for streaming, because side inputs will need to be added to state as they are computed.

pabloem:
Run Python PreCommit

pabloem:
Run Python PreCommit

pabloem:
failed test is streaming wordcount test

pabloem:
Run Python PreCommit

pabloem:
Run Python PreCommit

pabloem:
Building the side input index elsewhere. LMK what you think.

pabloem:
Run Python PreCommit

pabloem:
Run Python PreCommit

sdks/python/apache_beam/runners/portability/fn_api_runner/translations.py
robertwb: Discussed offline, but capturing here for the record. These sets contain the transitive collection of everything downstream of any side-input consuming transform, and as such can be large even if the total number of side inputs is small. (The number of distinct such sets is about the same as the number of side inputs, so we keep the total memory use down by re-using them--to give each transform its own copy would easily be O(n^2).)

Your change of computing the side input mapping after the graph has been fused is good (and arguably better, as you only need the immediate consumers, and don't have to re-compute each time a stage is fused).

pabloem:

sdks/python/apache_beam/runners/portability/fn_api_runner/fn_runner_test.py
```diff
@@ -240,6 +240,30 @@ def test_multimap_side_input(self):
               lambda k, d: (k, sorted(d[k])), beam.pvalue.AsMultiMap(side)),
           equal_to([('a', [1, 3]), ('b', [2])]))
 
+  def test_multimap_multiside_input(self):
```
boyuanzz, line 243: This test breaks Spark VR test: https://issues.apache.org/jira/browse/BEAM-9862. Please either support the same function for Spark or sickbay it.

ibzib: Thanks for reporting Boyuan, this was a flaw with the Spark runner. Fix: #11644

PR #11290, opened 2020-04-02
[BEAM-9670] Fix nullability widening in CoGroup key resolution

reuvenlax:
This PR also adds some zetasql unit tests that expose the previous bug.

### DIFF
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/SchemaUtils.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/SchemaUtils.java
@@ -1,0 +1,101 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.beam.sdk.schemas;
+
+import org.apache.beam.sdk.schemas.Schema.FieldType;
+
+/** A set of utility functions for schemas. */
+public class SchemaUtils {
+  /**
+   * Given two schema that have matching types, return a nullable-widened schema.
+   *
+   * <p>The schemas must have matching types, except for field names which can differ. The returned
+   * schema will contain the field names in the first schema. All field types will be nullable if
+   * the corresponding field type is nullable in either of the input schemas.
+   */
+  public static Schema mergeWideningNullable(Schema schema1, Schema schema2) {
+    if (schema1.getFieldCount() != schema2.getFieldCount()) {
+      throw new IllegalArgumentException(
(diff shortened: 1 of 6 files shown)


reuvenlax:
run sql postcommit

reuvenlax:
Run Java PreCommit

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamSetOperatorRelBase.java
```diff
@@ -63,16 +62,6 @@ public BeamSetOperatorRelBase(BeamRelNode beamRelNode, OpType opType, boolean al
...
-    Schema leftSchema = leftRows.getSchema();
-    Schema rightSchema = rightRows.getSchema();
-    if (!leftSchema.typesEqual(rightSchema)) {
-      throw new IllegalArgumentException(
-          "Can't intersect two tables with different schemas."
-              + "lhsSchema: "
-              + leftSchema
-              + "  rhsSchema: "
-              + rightSchema);
-    }
```
TheNeuralBit, line 75: Shouldn't we still verify that the schemas have equivalent types without nullability?

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/SchemaUtils.java
```diff
@@ -0,0 +1,101 @@
...
+              "Logical types don't match and cannot be merged: "
+                  + fieldType1.getLogicalType().getIdentifier()
+                  + ".v.s"
+                  + fieldType2.getLogicalType().getIdentifier());
+        }
+        // fall through
+      default:
+        result = fieldType1;
+    }
+    return result.withNullable(fieldType1.getNullable() || fieldType2.getNullable());
```
TheNeuralBit, line 99: Just a general comment: do you think we should add a visitor pattern for this sort of thing?

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamSetOperatorRelBase.java
reuvenlax: The merge function itself now verifies this.

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/SchemaUtils.java
reuvenlax: Maybe? I'm not sure visitor per se, as we don't require double dispatch. However we might consider generalizing the recursive tree walk.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamSetOperatorRelBase.java
TheNeuralBit: Ah I see. If there aren't any tests verifying this already could you add one?

reuvenlax: There is a test in CoGroupTest.java that verifies this. Should I add an additional one here?

TheNeuralBit: It could be nice to have something in ZetaSQLDialectSpecTest to verify the set operation just in case we break the logic there and not in CoGroup, but I guess it's fine either way

reuvenlax:
@TheNeuralBit thanks. I'll merge once tests complete

reuvenlax:
Run Java PreCommit

reuvenlax:
Run Java PreCommit

PR #11305, opened 2020-04-03
[BEAM-9577] Update container boot code to stage from dependencies, if present.

robertwb:
Falls back to legacy staging if no dependencies are present and the
retrieval token is non-trivial.

Currently the dependencies must be of role staging_to. Future work
to handle more expressive roles.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/go/pkg/beam/artifact/materialize_test.go
+++ b/sdks/go/pkg/beam/artifact/materialize_test.go
@@ -19,14 +19,19 @@ import (
 	"context"
 	"crypto/sha256"
 	"encoding/hex"
+	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"testing"
 
+	"github.com/apache/beam/sdks/go/pkg/beam/internal/errors"
 	pb "github.com/apache/beam/sdks/go/pkg/beam/model/jobmanagement_v1"
+	"github.com/apache/beam/sdks/go/pkg/beam/model/pipeline_v1"
 	"github.com/apache/beam/sdks/go/pkg/beam/util/grpcx"
+	"github.com/golang/protobuf/proto"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
 )
 
 // TestRetrieve tests that we can successfully retrieve fresh files.
@@ -68,7 +73,7 @@ func TestMultiRetrieve(t *testing.T) {
 	defer os.RemoveAll(dst)
 
 	client := pb.NewLegacyArtifactRetrievalServiceClient(cc)
-	if err := MultiRetrieve(ctx, client, 10, artifacts, rt, dst); err != nil {
+	if err := LegacyMultiRetrieve(ctx, client, 10, artifacts, rt, dst); err != nil {
 		t.Errorf("failed to retrieve: %v", err)
 	}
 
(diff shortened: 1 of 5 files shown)


lostluck:
Run Go Postcommit

sdks/go/pkg/beam/artifact/materialize_test.go
```diff
@@ -148,6 +153,156 @@ func stage(ctx context.Context, scl pb.LegacyArtifactStagingServiceClient, t *te
...
+	}
+}
+
+func (fake fakeGetArtifactResponse) Recv() (*pb.GetArtifactResponse, error) {
+	if *fake.index < len(fake.data) {
+		*fake.index += 1
+		return &pb.GetArtifactResponse{Data: fake.data[*fake.index-1 : *fake.index]}, nil
+	} else {
+		return nil, io.EOF
+	}
```
lostluck, line 274:
```suggestion
	} 
	return nil, io.EOF
```

sdks/go/pkg/beam/artifact/materialize_test.go
```diff
@@ -148,6 +153,156 @@ func stage(ctx context.Context, scl pb.LegacyArtifactStagingServiceClient, t *te
...
+	return &response, nil
+}
+
+func (fake fakeRetrievalService) GetArtifact(ctx context.Context, request *pb.GetArtifactRequest, opts ...grpc.CallOption) (pb.ArtifactRetrievalService_GetArtifactClient, error) {
+	var index int
+	if request.Artifact.TypeUrn == "resolved" {
+		return fakeGetArtifactResponse{data: request.Artifact.TypePayload, index: &index}, nil
+	} else {
+		return nil, errors.Errorf("Unsupported artifact %v", request.Artifact)
+	}
```
lostluck, line 265:
```suggestion
	}
	return nil, errors.Errorf("Unsupported artifact %v", request.Artifact)
```

sdks/go/pkg/beam/artifact/materialize_test.go
```diff
@@ -148,6 +153,156 @@ func stage(ctx context.Context, scl pb.LegacyArtifactStagingServiceClient, t *te
...
+}
+
+func (fake fakeGetArtifactResponse) Recv() (*pb.GetArtifactResponse, error) {
```
lostluck, line 267: I think It's a bit confusing that this is called a response when it's a fake Client which returns responses.

sdks/go/pkg/beam/artifact/materialize_test.go
```diff
@@ -148,6 +153,156 @@ func stage(ctx context.Context, scl pb.LegacyArtifactStagingServiceClient, t *te
...
+		*fake.index += 1
+		return &pb.GetArtifactResponse{Data: fake.data[*fake.index-1 : *fake.index]}, nil
+	} else {
+		return nil, io.EOF
+	}
+}
+
+type fakeGetArtifactResponse struct {
+	data  []byte
+	index *int
```
lostluck, line 277: Why a pointer to the index rather than make the methods on the pointer to the struct? The implementation seems to be trying to avoid the implicit value method copying, instead of just using pointer methods. 

See https://golang.org/doc/effective_go.html#methods for more information.

tldr; replacing all the (fake fakeGetArtifactResponse) with (fake *fakeGetArtifactResponse) puts the method on the pointer, which means that repeated calls have access to the full state, and you don't need the indirections to have index increment as expected.

This also applies to the other fakes or interfaces you've written, but given that you're only modifying reference state if at all (eg. maps, or pointers), then it doesn't matter so much that a value method is being used.

The general rule is that when in doubt, use a pointer receiver.

robertwb: That is what I tried at first, but it complained that it didn't satisfy the interface.

sdks/go/pkg/beam/artifact/materialize_test.go
robertwb: It's the object that represents the response stream (and is returned by the actual client). Would naming it `fakeGetArtifactResponseStream` be better?

lostluck: Yes.

sdks/go/pkg/beam/artifact/materialize_test.go
```diff
@@ -148,6 +153,154 @@ func stage(ctx context.Context, scl pb.LegacyArtifactStagingServiceClient, t *te
...
+			response.Replacements = append(response.Replacements, dep)
+		}
+	}
+	return &response, nil
+}
+
+func (fake fakeRetrievalService) GetArtifact(ctx context.Context, request *pb.GetArtifactRequest, opts ...grpc.CallOption) (pb.ArtifactRetrievalService_GetArtifactClient, error) {
+	var index int
+	if request.Artifact.TypeUrn == "resolved" {
+		return fakeGetArtifactResponse{data: request.Artifact.TypePayload, index: &index}, nil
```
lostluck, line 262: You need to return the pointer instance here so that the pointer methods can satisfy the interface.

Value methods are "promoted" to the derived pointer type, but not the other way around.

From Effective Go:
 The rule about pointers vs. values for receivers is that value methods can be invoked on pointers and values, but pointer methods can only be invoked on pointers.

sdks/go/pkg/beam/artifact/materialize_test.go
lostluck: Answered at the definition of the instance.

sdks/go/pkg/beam/artifact/materialize_test.go
robertwb: Thanks for the helpful explication.

lostluck:
Run Go PostCommit

lostluck:
In the post commit, it doesn't look the dataflow jobs still don't like the current situation. They claim "no artifacts staged".

lostluck:
That appears to have been the case since https://github.com/apache/beam/pull/11305

robertwb:
Yeah, I'm tracking that bug. This doesn't make things worse.

PR #11314, opened 2020-04-04
[BEAM-9562] Send Timers over Data Channel as Elements

boyuanzz:
For commit: https://github.com/apache/beam/pull/11314/commits/3ff8c3eb28944cbe7c757fdbae49947b57c3774f
r: @robertwb for data_plane.py and bundle_processor.py
r: @pabloem for fn_runner related part.

For commit: https://github.com/apache/beam/pull/11314/commits/a2a716455edb8d1a80ae4d302a79b19ef6053484
r: @TheNeuralBit 

cc: @y1chi for reference

------------------------

--- | --- | --- | --- | ---


sdks/python/apache_beam/pipeline.py
```diff
@@ -1071,6 +1071,14 @@ def named_inputs(self):
     }
     return dict(main_inputs, **side_inputs)
 
+  def main_inputs(self):
```
robertwb, line 1074: Generic transforms don't have the notion of main inputs, let's filter things out in the implementation in ParDo.

sdks/python/apache_beam/transforms/core.py
```diff
@@ -1294,16 +1295,43 @@ def _pardo_fn_data(self):
     windowing = None
     return self.fn, self.args, self.kwargs, si_tags_and_types, windowing
 
-  def to_runner_api_parameter(self, context):
+  def to_runner_api(self, context, main_inputs, has_parts=False):
```
robertwb, line 1298: This is starting to look like a lot of code duplication. How about we pass (all) inputs as a keyword argument, and let `PTransform.to_runner_api` take an `**extra_kwargs` that it passes on to `to_runner_api_parameter`.

sdks/python/apache_beam/runners/worker/bundle_processor.py
```diff
@@ -562,45 +562,45 @@ class OutputTimer(object):
   def __init__(self,
                key,
                window,  # type: windowed_value.BoundedWindow
-               receiver  # type: operations.ConsumerSet
+               paneinfo,
+               timer_family_id,
+               timer_coder_impl,
+               output_stream
```
robertwb, line 568: A type on this parameter would be useful.

sdks/python/apache_beam/runners/worker/bundle_processor.py
```diff
@@ -562,45 +562,45 @@ class OutputTimer(object):
...
-                fire_timestamp=timestamp.Timestamp.of(clear_ts),
-                hold_timestamp=timestamp.Timestamp.of(0),
-                paneinfo=windowed_value.PANE_INFO_UNKNOWN)),
-                                     0, (self._window, )))
+    timer = userstate.Timer(
+        user_key=self._key,
+        dynamic_timer_tag='',
+        windows=(self._window, ),
+        clear_bit=False,
+        fire_timestamp=clear_ts,
```
robertwb, line 599: Don't bother setting these timestamps, or paneinfo.

robertwb: (Should the coder be ignoring them as well?)

sdks/python/apache_beam/runners/worker/bundle_processor.py
```diff
@@ -611,7 +611,7 @@ def __init__(self,
                transform_id,  # type: str
                key_coder,  # type: coders.Coder
                window_coder,  # type: coders.Coder
-               timer_family_specs  # type: Mapping[str, beam_runner_api_pb2.TimerFamilySpec]
+               timer_coders
```
robertwb, line 614: type?

sdks/python/apache_beam/runners/worker/bundle_processor.py
```diff
@@ -1088,6 +1142,30 @@ def create_operation(self,
         transform_proto.spec.payload, parameter_type)
     return creator(self, transform_id, transform_proto, payload, consumers)
 
+  def get_timer_coders(self):
+    timer_coder = {}
+    for transform_id, transform_proto in self.descriptor.transforms.items():
```
robertwb, line 1147: I see us doing this loop three times now. Perhaps it would be more useful to do the loop once to set everything up, creating a single dictionary (transform_id, timer_family_id) -> (all info about that timer we need to dispatch them).

sdks/python/apache_beam/runners/worker/bundle_processor.py
```diff
@@ -846,6 +870,30 @@ def process_bundle(self, instruction_id):
...
+      timer_output_streams = {}
+      for transform_id, timer_list in self.timer_ids.items():
+        output_streams = {}
+        for timer_id in timer_list:
+          output_streams[
+              timer_id] = self.timer_data_channel.output_timer_stream(
+                  instruction_id, transform_id, timer_id)
+          timer_output_streams[transform_id] = output_streams
+        self.process_timer_ops[
+            transform_id].user_state_context.update_timer_output_streams(
```
robertwb, line 883: Nit: rather than this double nesting, it might simplify things to have an `update_timer_output_streams(timer_id, output_stream)` method that could be called repeatedly.

sdks/python/apache_beam/runners/worker/bundle_processor.py
```diff
@@ -846,6 +870,30 @@ def process_bundle(self, instruction_id):
...
+                output_streams)
+
+      # Process timers
+      if self.timer_data_channel:
```
robertwb, line 887: We can't safely assume the runner will finish sending all timers before sending any of the data (and the buffer may get full, resulting in a deadlock). I think we need to have a data_channel.inputs() that returns both data and timers and then branch in the loop.

pabloem:
I'm sorry. I am havinhg a heavy headache. I'll bow out. @robertwb can you review fn_runner.py and siblings?

runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/BatchDataflowWorker.java
```diff
@@ -238,9 +238,14 @@ protected BatchDataflowWorker(
         sdkFusedStage =
             pipeline == null
                 ? RegisterNodeFunction.withoutPipeline(
-                    idGenerator, sdkHarnessRegistry.beamFnStateApiServiceDescriptor())
+                    idGenerator,
+                    sdkHarnessRegistry.beamFnStateApiServiceDescriptor(),
+                    sdkHarnessRegistry.beamFnDataApiServiceDescriptor())
```
TheNeuralBit, line 243: Isn't the timer API service descriptor different from the data API service descriptor? Does that need to be plumbed through SdkHarnessRegistry and used here instead of the data API descriptor? (same question below and in streaming worker)

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/ParDoTranslation.java
```diff
@@ -257,6 +230,19 @@ public static ParDoPayload translateParDo(
...
     }
 
+    Coder<BoundedWindow> windowCoder =
+        (Coder<BoundedWindow>) mainInput.getWindowingStrategy().getWindowFn().windowCoder();
+    Coder<?> keyCoder;
+    if (signature.usesState() || signature.usesTimers()) {
+      checkArgument(
+          mainInput.getCoder() instanceof KvCoder,
+          "DoFn's that use state or timers must have an input PCollection with a KvCoder but received %s",
+          mainInput.getCoder());
```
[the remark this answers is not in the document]
lukecwik: It was being covered by validation in DoFnSignatures but it is being repeated here for defense in depth reasons.

runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/BatchDataflowWorker.java
lukecwik: They both use the Data API so no. All were saying here is that we will re-use the same gRPC channel for both timers and data.

sdks/java/harness/src/test/java/org/apache/beam/fn/harness/control/ProcessBundleHandlerTest.java
```diff
@@ -914,6 +923,7 @@ public Object createRunnerForPTransform(
                       PipelineOptions pipelineOptions,
                       BeamFnDataClient beamFnDataClient,
                       BeamFnStateClient beamFnStateClient,
+                      BeamFnTimerClient beamFnTimerClient,
                       String pTransformId,
                       PTransform pTransform,
                       Supplier<String> processBundleInstructionId,
```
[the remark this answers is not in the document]
lukecwik: for completeness yes

runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/BatchDataflowWorker.java
TheNeuralBit: I see. So we only have a separate timer_api_service_descriptor in the protos so that a runner has the option to make it separate, but it doesn't need to be separate?

runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/graph/RegisterNodeFunction.java
```diff
@@ -134,14 +136,18 @@ public static RegisterNodeFunction forPipeline(
...
+    return new RegisterNodeFunction(
+        null, idGenerator, stateApiServiceDescriptor, timerApiServiceDescriptor);
   }
 
   private RegisterNodeFunction(
       @Nullable RunnerApi.Pipeline pipeline,
       IdGenerator idGenerator,
-      Endpoints.ApiServiceDescriptor stateApiServiceDescriptor) {
+      Endpoints.ApiServiceDescriptor stateApiServiceDescriptor,
+      Endpoints.ApiServiceDescriptor timerApiServiceDescriptor) {
```
TheNeuralBit, line 150: timerApiServiceDescriptor isn't used? Should it be stored and written to the ProcessBundleDescrioptor?

sdks/python/apache_beam/runners/worker/bundle_processor.py
boyuanzz: > (Should the coder be ignoring them as well?)
No, the timer coder is encoding all of these info now.

> Don't bother setting these timestamps, or paneinfo.
Could you please explain more about this?

runners/google-cloud-dataflow-java/worker/src/main/java/org/apache/beam/runners/dataflow/worker/BatchDataflowWorker.java
lukecwik: That is correct.

boyuanzz:
The [test_pardo_timers_clear](https://github.com/apache/beam/blob/master/sdks/python/apache_beam/runners/portability/fn_api_runner/fn_runner_test.py#L351-L389) fails with streaming Flink. The python sdk sends all timers(hold_timestamp=-INF with python default behavior) but only gets the timer with timestamp=20 back. Given the test only fails when streaming, it seems like something not correct with watermark(?). @lukecwik

sdks/python/apache_beam/runners/worker/bundle_processor.py
```diff
@@ -1088,6 +1145,21 @@ def create_operation(self,
         transform_proto.spec.payload, parameter_type)
     return creator(self, transform_id, transform_proto, payload, consumers)
 
+  def extract_timers_info(self):
```
robertwb, line 1157: I would populate output_stream here as well rather than above.

sdks/python/apache_beam/runners/worker/data_plane.py
```diff
@@ -274,20 +301,47 @@ def inverse(self):
     return self._inverse
 
   def input_elements(self,
-                     instruction_id,  # type: str
-                     unused_expected_transforms=None,  # type: Optional[Collection[str]]
-                     abort_callback=None  # type: Optional[Callable[[], bool]]
-                    ):
-    # type: (...) -> Iterator[beam_fn_api_pb2.Elements.Data]
```
robertwb, line 281: It'd be good to not lose the typing information. You can make an alias at the top of the file `DataOrTimers = Union[beam_fn_api_pb2.Elements.Data, beam_fn_api_pb2.Elements.Timer]` to cut down on verbosity, here and elsewhere.

sdks/python/apache_beam/runners/worker/data_plane.py
```diff
@@ -274,20 +301,47 @@ def inverse(self):
...
+      instruction_id,  # type: str
+      unused_expected_inputes=None,   # type: Collection[str]
```
robertwb, line 305: inputes -> inputs

sdks/python/apache_beam/runners/worker/data_plane.py
```diff
@@ -354,15 +410,15 @@ def input_elements(self,
 
     Args:
       instruction_id(str): instruction_id for which data is read
-      expected_transforms(collection): expected transforms
+      expected_inputs(collection): expected inputs, include both data and timer.
     """
     received = self._receiving_queue(instruction_id)
-    done_transforms = set()  # type: Set[str]
+    done_inputs = set()  # type: Set[str]
```
robertwb, line 416: update type hint

sdks/python/apache_beam/runners/worker/data_plane.py
```diff
@@ -372,12 +428,18 @@ def input_elements(self,
...
-            done_transforms.add(data.transform_id)
-          else:
-            assert data.transform_id not in done_transforms
-            yield data
+          if isinstance(element, beam_fn_api_pb2.Elements.Timer):
+            if element.is_last:
+              done_inputs.add((element.transform_id, element.timer_family_id))
+            else:
+              yield element
+          if isinstance(element, beam_fn_api_pb2.Elements.Data):
```
robertwb, line 436: elif

sdks/python/apache_beam/runners/worker/data_plane.py
```diff
@@ -408,27 +470,67 @@ def close_callback(data):
...
-        pass
-      if data[-1] is self._WRITES_FINISHED:
-        done = True
-        data.pop()
-      if data:
-        yield beam_fn_api_pb2.Elements(data=data)
+    stream_done = False
+    while not stream_done:
+      streams = None
+      if not stream_done:
```
robertwb, line 502: This will always be true (given the loop condition).

sdks/python/apache_beam/runners/worker/data_plane.py
```diff
@@ -408,27 +470,67 @@ def close_callback(data):
...
+      if streams:
+        elements = beam_fn_api_pb2.Elements()
+        data_stream = []
+        timer_stream = []
+        for stream in streams:
+          if isinstance(stream, beam_fn_api_pb2.Elements.Timer):
+            timer_stream.append(stream)
+          if isinstance(stream, beam_fn_api_pb2.Elements.Data):
+            data_stream.append(stream)
+        if data_stream:
```
robertwb, line 522: No need to have these conditionals, you can just write

`yield beam_fn_api_pb2.Elements(data=data_stream, timer=timer_stream)`

sdks/python/apache_beam/runners/worker/data_plane.py
```diff
@@ -408,27 +470,67 @@ def close_callback(data):
...
+          stream_done = True
+          streams.pop()
+      if streams:
+        elements = beam_fn_api_pb2.Elements()
+        data_stream = []
+        timer_stream = []
+        for stream in streams:
+          if isinstance(stream, beam_fn_api_pb2.Elements.Timer):
+            timer_stream.append(stream)
+          if isinstance(stream, beam_fn_api_pb2.Elements.Data):
```
robertwb, line 520: else

sdks/python/apache_beam/runners/worker/operations.pxd
```diff
@@ -92,7 +92,7 @@ cdef class DoOperation(Operation):
   cdef DoFnRunner dofn_runner
   cdef object tagged_receivers
   cdef object side_input_maps
-  cdef object user_state_context
+  cpdef public object user_state_context
```
robertwb, line 95: Rather than making this public, I would add an `add_timer_info` method to this operation.

lukecwik:
Run Python PreCommit

lukecwik:

sdks/python/apache_beam/runners/worker/bundle_processor.py
boyuanzz: Correct, when `clear_bit` is `True`, the coder ignores these fields. I think we should have a better `Timer` with API `of` and `clear` like in Java as a follow up.

sdks/python/apache_beam/runners/worker/bundle_processor.py
boyuanzz: We can only populate output_stream when processing bundle since instruction_id is required.

robertwb: Ack, expanding the whole diff I see that this is happening in different methods now (in which case here two separate maps, as you had originally, might be preferable). But not a big deal.

boyuanzz: The keys of maps should still be the same though(tuple of (transform_id, timer_family_id)). That's why I make the value as a map{coder_impl, output_stream}

sdks/python/apache_beam/transforms/core.py
```diff
@@ -1323,12 +1325,47 @@ def _pardo_fn_data(self):
...
+      key_coder = coder.key_coder()
+    else:
+      key_coder = coders.registry.get_coder(typehints.Any)
+    window_coder = input_pcoll.windowing.windowfn.get_window_coder()
+    return key_coder, window_coder
+
+  def to_runner_api(self, context, **extra_kwargs):
+    # type: (PipelineContext, bool) -> beam_runner_api_pb2.FunctionSpec
+    has_parts = extra_kwargs.get('has_part', False)
+    urn, typed_param = self.to_runner_api_parameter(context, **extra_kwargs)
```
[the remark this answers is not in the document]
robertwb: Nevermind, I see what's going on here.

sdks/python/apache_beam/runners/portability/fn_api_runner/execution.py
```diff
@@ -355,20 +364,41 @@ def _build_process_bundle_descriptor(self):
...
+
+  def get_coder_impl(self, coder_id):
     if coder_id in self.execution_context.safe_coders:
       return self.execution_context.pipeline_context.coders[
           self.execution_context.safe_coders[coder_id]].get_impl()
     else:
       return self.execution_context.pipeline_context.coders[coder_id].get_impl()
 
+  def get_timer_coder_impl(self, transform_id, timer_family_id):
+    assert (transform_id, timer_family_id) in self._timer_coder_ids
```
robertwb, line 398: The key error if it's not present below will be sufficient.

sdks/python/apache_beam/transforms/core.py
```diff
@@ -1323,12 +1325,47 @@ def _pardo_fn_data(self):
...
+        raise ValueError(
+            'Input elements to the transform %s with stateful DoFn must be '
+            'key-value pairs.' % self)
+      key_coder = coder.key_coder()
+    else:
+      key_coder = coders.registry.get_coder(typehints.Any)
+    window_coder = input_pcoll.windowing.windowfn.get_window_coder()
+    return key_coder, window_coder
+
+  def to_runner_api(self, context, **extra_kwargs):
```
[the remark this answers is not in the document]
boyuanzz: We can delete this override since we pass `extra_kwargs` from `PTransform` now.

sdks/python/apache_beam/transforms/core.py
```diff
@@ -1272,6 +1272,8 @@ def expand(self, pcoll):
         key_coder = coder.key_coder()
       else:
         key_coder = coders.registry.get_coder(typehints.Any)
+      self.window_coder = pcoll.windowing.windowfn.get_window_coder()
```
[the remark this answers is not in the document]
boyuanzz: No. Will removed.

sdks/python/apache_beam/runners/worker/bundle_processor.py
```diff
@@ -837,25 +869,59 @@ def process_bundle(self, instruction_id):
...
-      # Inject inputs from data plane.
+      # Each data_channel is mapped to a list of expected inputs which includes
+      # both data input and timer input. The data input is identied by
+      # transform_id. The data input is identified by
+      # (transform_id, timer_family_id).
       data_channels = collections.defaultdict(
           list
       )  # type: DefaultDict[data_plane.GrpcClientDataChannel, List[str]]
+
+      # Inject data inputs from data plane.
```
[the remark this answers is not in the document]
boyuanzz: Updated the comment.

sdks/python/apache_beam/runners/portability/fn_api_runner/fn_runner.py
```diff
@@ -987,8 +1019,10 @@ def __init__(
 
   def process_bundle(self,
                      inputs,  # type: Mapping[str, PartitionableBuffer]
-                     expected_outputs  # type: DataOutput
-                    ):
+                     expected_outputs,  # type: DataOutput
+                     fired_timers,  # type: Mapping[str, Mapping[str, PartitionableBuffer]]
```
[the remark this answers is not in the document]
boyuanzz: I updated the `fired_timers` implementation but forgot to update the typing here. Thanks!

sdks/python/apache_beam/runners/portability/fn_api_runner/fn_runner.py
```diff
@@ -536,7 +525,8 @@ def _run_stage(self,
         runner_execution_context,
         bundle_context_manager,
         data_input,
-        data_output,
+        data_output, {},
```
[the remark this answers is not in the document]
boyuanzz: yapf helps me put the {} here.

mxm:
This is a big change which also affects the runners. Would it have made sense to notify Runner authors, especially since post commit tests are broken? It took me a bit to figure out what caused the regression.

ibzib:
@mxm Which post commits are you referring to? & Can you please mark the jira(s) with fix version 2.21.0 so we can fix the regression in the release?

boyuanzz:
> This is a big change which also affects the runners. Would it have made sense to notify Runner authors, especially since post commit tests are broken? It took me a bit to figure out what caused the regression.

Thanks, Max! Sorry for the inconvenience. It seems like currently both Spark and Flink fail on the same test: org.apache.beam.sdk.transforms.ParDoTest$TimerTests.testEventTimeTimerAlignBounded. The failure pattern is also the same: the pipeline only produces the output from timer, not from the ProcessElement fn. I think there should be something wrong in the java runner shared library code. Have you worked on it? Or do you want me to follow up fixing this issue?

lukecwik:
The problem is with the Timer implementation inside the FnApiDoFnRunner. The spec for Timer wasn't clear as to what the defaults were when withOutputTimestamp was added and hence some critical logic was deleted during the migration.

See #11402 for the fix.

mxm:
I was actually working on something related to timers in #11362 and was surprised to see that the test failed when I opened the PR, since I had run tests locally. Then figured something must have changed on master in the meantime. Thanks for following up with this!

PR #11357, opened 2020-04-09
[BEAM-7770] Add ReadAll transform for SolrIO

iemejia:
This follows the ReadAll pattern style introduced on HBaseIO but for the case of non SDF based DoFn.

R: @aromanenko-dev

### DIFF
--- a/sdks/java/io/solr/src/main/java/org/apache/beam/sdk/io/solr/SolrIO.java
+++ b/sdks/java/io/solr/src/main/java/org/apache/beam/sdk/io/solr/SolrIO.java
@@ -41,7 +41,6 @@ import org.apache.beam.sdk.util.BackOff;
 import org.apache.beam.sdk.util.BackOffUtils;
 import org.apache.beam.sdk.util.FluentBackoff;
 import org.apache.beam.sdk.util.Sleeper;
-import org.apache.beam.sdk.values.KV;
 import org.apache.beam.sdk.values.PBegin;
 import org.apache.beam.sdk.values.PCollection;
 import org.apache.beam.sdk.values.PDone;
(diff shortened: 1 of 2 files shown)


sdks/java/io/solr/src/main/java/org/apache/beam/sdk/io/solr/SolrIO.java
```diff
@@ -364,36 +372,41 @@ Read withBatchSize(int batchSize) {
...
       checkArgument(
           getConnectionConfiguration() != null, "withConnectionConfiguration() is required");
       checkArgument(getCollection() != null, "from() is required");
-
-      return input
-          .apply("Create", Create.of(this))
-          .apply("Split", ParDo.of(new SplitFn()))
-          .apply("Reshuffle", Reshuffle.viaRandomKey())
-          .apply("Read", ParDo.of(new ReadFn()));
+      return input.apply("Create", Create.of(this)).apply("ReadAll", readAll());
```
aromanenko-dev, line 392: Is `withReplicaInfo()` required here? If yes, please, add a check here.

aromanenko-dev:
retest this please

sdks/java/io/solr/src/main/java/org/apache/beam/sdk/io/solr/SolrIO.java
aromanenko-dev: My bad, I think we don't it here since anyway it will pass through `SplitFn` downstream

aromanenko-dev:
Since `readAll()` is a new public API method, please ,add test(s) for it. Apart from that, it's fine for me.

iemejia:
`readAll()` is already tested. Notice that `read()` was refactored to rely on `readAll()` in line 392. I can write an extra test if you prefer but it is just duplication because it is already covered. Should I write the extra test?

aromanenko-dev:
@iemejia You are right - it's clear that `readAll()` is tested implicitly through calling `read()`. However, tests are not (and shouldn't be!) aware  about internal implementation of tested method, so we have to have tests for all public API methods. So, yes, please add an extra test for that.

iemejia:
`testReadAll` added can you PTAL again @aromanenko-dev. Thanks!

sdks/java/io/solr/src/test/java/org/apache/beam/sdk/io/solr/SolrIOTest.java
```diff
@@ -155,6 +155,23 @@ public void testRead() throws Exception {
...
+    PCollection<SolrDocument> output =
+        pipeline
+            .apply(
+                Create.of(
+                    SolrIO.read()
+                        .withConnectionConfiguration(connectionConfiguration)
+                        .from(SOLR_COLLECTION)
+                        .withBatchSize(101)))
+            .apply(SolrIO.readAll());
+    PAssert.thatSingleton(output.apply("Count", Count.globally())).isEqualTo(NUM_DOCS);
```
aromanenko-dev, line 171: I believe we can improve this test in the future by reading with multiple `Read`s (the goal of adding `readAll()`) and check the content of read messages (not only count).

PR #11360, opened 2020-04-09
[BEAM-9722] added SnowflakeIO with Read operation

DariuszAniszewski:
This PR is adding `SnowflakeIO` with `Read` operation as part of [BEAM-9722](https://issues.apache.org/jira/browse/BEAM-9722)

Snowflake is an analytic data warehouse provided as Software-as-a-Service (SaaS). It uses a new SQL database engine with a unique architecture designed for the cloud. To read more details please check [here](https://www.snowflake.com ) and [here](https://docs.snowflake.net/manuals/user-guide/intro-key-concepts.html).

The `SnowflakeIO.Read` uses [Snowflake's JDBC driver](https://docs.snowflake.com/en/user-guide/jdbc.html) to run [COPY INTO <location>](https://docs.snowflake.com/en/sql-reference/sql/copy-into-location.html) statement to move data on GCS as CSV files that are then read via `FileIO`.

The `SnowflakeIO` allows to use three authentication methods against Snowflake:
* username and password
* key-pair
* pre-obtained OAuth token

This PR is first of series, once merged, subsequent PRs will come - with Write, integration tests etc. We're also working on cross-language to support Python SDK.

------------------------

--- | --- | --- | --- | ---


sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/SnowflakeIO.java
```diff
@@ -0,0 +1,691 @@
...
+ *
+ * <p>For example:
+ *
+ * <pre>{@code
+ * SnowflakeIO.DataSourceConfiguration dataSourceConfiguration =
+ *     SnowflakeIO.DataSourceConfiguration.create(SnowflakeCredentialsFactory.of(options))
+ *         .withServerName(options.getServerName())
+ *         .withWarehouse(options.getWarehouse())
+ *         .withDatabase(options.getDatabase())
+ *         .withSchema(options.getSchema);
```
takidau, line 93: options.getSchema()

DariuszAniszewski: thanks!

DariuszAniszewski:
R: @aromanenko-dev 
on dev-list you were interested in reviewing PRs related to SnowflakeIO - could you please take a look or point other reviewer? thanks!

aromanenko-dev:
@DariuszAniszewski Thank you for contribution! I'm a bit busy for now but I'll try to take a look on this asap.

DariuszAniszewski:
@aromanenko-dev I don't want to push you but do you have any ETA on the review? I'm OK with waiting but it would be nice to roughly know how long it may take ;) 

I also have a generic question - as mentioned in description of this very PR, it's first from a series. We have more pieces of SnowflakeIO done, but we wanted to make as small and atomic PRs as possible to ease the review process and merge them one by one. But maybe this approach is misfired and it's better to put all pieces of the IO (in atomic commits of course) into this one PR?

WDYT?

aromanenko-dev:
@iemejia Would you have some time to take a look as well?

sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/SnowflakeIO.java
```diff
@@ -0,0 +1,643 @@
...
+          "fromTable() and fromQuery() is not allowed together");
+      checkArgument(getCsvMapper() != null, "withCsvMapper() is required");
+      checkArgument(getCoder() != null, "withCoder() is required");
+      checkArgument(getIntegrationName() != null, "withIntegrationName() is required");
+      checkArgument(getStagingBucketName() != null, "withStagingBucketName() is required");
+      checkArgument(
+          (getDataSourceProviderFn() != null),
+          "withDataSourceConfiguration() or withDataSourceProviderFn() is required");
+
+      String gcpTmpDirName = makeTmpDirName();
```
aromanenko-dev, line 268: I believe, in general, COPY INTO can use the different external locations, like Amazon S3, Google Cloud Storage, or Microsoft Azure. So, please make it configurable for users even if this IO now  supports only GCS.

sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/SnowflakeServiceImpl.java
```diff
@@ -0,0 +1,83 @@
...
+
+    String from;
+    if (query != null) {
+      // Query must be surrounded with brackets
+      from = String.format("(%s)", query);
+    } else {
+      from = table;
+    }
+
+    String externalLocation = String.format("gcs://%s/%s/", stagingBucketName, tmpDirName);
```
aromanenko-dev, line 51: Is it GC specific implementation? I think we need to make it configurable which type of external location to use.

sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/credentials/package-info.java
```diff
@@ -0,0 +1,2 @@
+/** Credentials for SnowflakeIO. */
```
aromanenko-dev, line 1: Add ASF license header

sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/package-info.java
```diff
@@ -0,0 +1,2 @@
+/** Snowflake IO transforms. */
```
aromanenko-dev, line 1: Add ASF license header

sdks/java/io/snowflake/src/test/java/org/apache/beam/sdk/io/snowflake/test/FakeSnowflakeDatabase.java
```diff
@@ -0,0 +1,73 @@
...
+import java.io.Serializable;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import net.snowflake.client.jdbc.SnowflakeSQLException;
+
+/** Fake implementation of SnowFlake warehouse used in test code. */
+public class FakeSnowflakeDatabase implements Serializable {
```
aromanenko-dev, line 29: Are there any mock/fake already implemented snowflake databases that can be used for testing?

sdks/java/io/snowflake/src/test/java/org/apache/beam/sdk/io/snowflake/test/package-info.java
```diff
@@ -0,0 +1,2 @@
+/** Snowflake IO tests. */
```
aromanenko-dev, line 1: Add ASF license header

sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/SnowflakeCloudProvider.java
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.beam.sdk.io.snowflake;
+
+/** Interface which defines common methods for cloud providers. */
+public interface SnowflakeCloudProvider {
+  void removeFiles(String bucketName, String pathOnBucket);
```
lukecwik, line 22: Why not use Apache Beam's notion of a FileSystem?

sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/GCSProvider.java
```diff
@@ -0,0 +1,36 @@
...
+import com.google.api.gax.paging.Page;
+import com.google.cloud.storage.Blob;
+import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.StorageOptions;
+import java.io.Serializable;
+
+public class GCSProvider implements SnowflakeCloudProvider, Serializable {
+
+  @Override
+  public void removeFiles(String bucketName, String pathOnBucket) {
```
lukecwik, line 29: Why not use the Apache Beam GCS filesystem?

sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/SnowflakeIO.java
```diff
@@ -0,0 +1,643 @@
...
+
+      abstract Builder<T> setSnowflakeCloudProvider(SnowflakeCloudProvider snowflakeCloudProvider);
+
+      abstract Read<T> build();
+    }
+
+    public Read<T> withDataSourceConfiguration(final DataSourceConfiguration config) {
+
+      try {
+        Connection connection = config.buildDatasource().getConnection();
```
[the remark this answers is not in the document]
DariuszAniszewski: Yes, it does. The reason is to early validate if connection parameters are valid

sdks/java/io/snowflake/src/test/java/org/apache/beam/sdk/io/snowflake/test/FakeSnowflakeDatabase.java
DariuszAniszewski: There is no such thing AFAIK - we had to prepare fake ourselves

DariuszAniszewski:
Thanks @aromanenko-dev and @lukecwik for review. 
I rebased onto (current) master and applied simplest changes already - I'll be back in a few days with rest of them.

DariuszAniszewski:
I'm dealing with some private issues this week and I'll be unavailable. 
@kkucharc and @purbanow will continue working on this PR

sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/SnowflakeIO.java
```diff
@@ -0,0 +1,643 @@
...
+              .apply(ParDo.of(new MapCsvToStringArrayFn()))
+              .apply(ParDo.of(new MapStringArrayToUserDataFn<>(getCsvMapper())));
+
+      output.setCoder(getCoder());
+
+      emptyCollection
+          .apply(Wait.on(output))
+          .apply(
+              ParDo.of(
+                  new CleanTmpFilesFromGcsFn(
```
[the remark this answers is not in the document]
DariuszAniszewski: It seems that it doesn't at the moment. @kkucharc will share more info here

DariuszAniszewski:
@aromanenko-dev @lukecwik thanks again for the review. We've applied the changes and all your comments/questions are addressed: either changed or answered. 

Can you please re-review? :)

sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/SnowflakeIO.java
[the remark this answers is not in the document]
kkucharc: Yes, we checked and it doesn't. @aromanenko-dev do you think it should be provided? In case of testing, probably tests should take care of cleanup.

sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/SnowflakeCloudProvider.java
kkucharc: Thanks for suggestion @lukecwik . I changed it to `GCSFileSystem`. 

I also tried to change removeFiles in `Fake` implementation to use `LocalFileSystem` but I am a little bit concerned - `LocalFileSystem` doesn't match nested directories and fails on deleting not empty directory. That can cause that testing directory won't be cleaned and tests will become flaky.

sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/SnowflakeIO.java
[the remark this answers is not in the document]
aromanenko-dev: I'm afraid that many failed pipelines could lead to wasting of used dick space in this case. It would be better to avoid such behavior, if possible.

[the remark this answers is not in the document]
DariuszAniszewski: While I generally agree that failing pipelines will eventually lead to lots of garbage on GCS (or other storage in future) I'm not sure how to ensure those files are deleted regardless of pipeline status. 

AFAIK there is nothing like `@After` or `@Teardown` for `PTransform` which we have in our IO. Using `Wait` transform gives us ability to remove files once data is read and seems reasonable choice.

I've been checking how i.e. `BigQueryIO` is handling that case as it also needs to cleanup and they also have a cleanup transform that is called once all rows are read. I assume in their case cleanup also won't be run if something in-between fails.

How about filing JIRA issue for this case and try to look at solution in parallel to delivering other pieces of the Snowflake connector? @aromanenko-dev @kkucharc  WDYT?

sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/SnowflakeIO.java
```diff
@@ -0,0 +1,768 @@
...
+     * User-defined function mapping CSV lines into user data.
+     *
+     * @param csvMapper - an instance of {@link CsvMapper}.
+     */
+    public Read<T> withCsvMapper(CsvMapper<T> csvMapper) {
+      return toBuilder().setCsvMapper(csvMapper).build();
+    }
+
+    /**
+     * A coder to use.
```
chamikaramj, line 283: Coder to be used by the output PCollection generated by the source.

BTW that's the default if a coder is not provided ? This does not have to be always configurable.

sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/SnowflakeIO.java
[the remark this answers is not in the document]
chamikaramj: This is a known issue with BQ source as well. Failed pipelines can leave temporary files behind. I'm afraid there is no good solution today. I think we need to introduce some sort of a generalized cleanup step to address this.

sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/SnowflakeIO.java
```diff
@@ -0,0 +1,768 @@
...
+                integrationName,
+                stagingBucketDir,
+                tmpDirName,
+                this.cloudProvider);
+
+        context.output(output);
+      }
+    }
+
+    public static class MapCsvToStringArrayFn extends DoFn<String, String[]> {
```
chamikaramj, line 396: Is there a reason to introduce two transforms here instead of having a single transform that converts Strings to type T ?

sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/SnowflakeIO.java
```diff
@@ -0,0 +1,768 @@
...
+          .setDataSource(dataSource)
+          .build();
+    }
+
+    /**
+     * Creates {@link DataSourceConfiguration} from instance of {@link SnowflakeCredentials}.
+     *
+     * @param credentials - an instance of {@link SnowflakeCredentials}.
+     */
+    public static DataSourceConfiguration create(SnowflakeCredentials credentials) {
```
chamikaramj, line 556: Can we introduce just one create(SnowflakeCredentials credentials) method and fork based on the type of object passed in ?

sdks/java/io/snowflake/src/test/java/org/apache/beam/sdk/io/snowflake/test/unit/read/SnowflakeIOReadTest.java
```diff
@@ -0,0 +1,251 @@
...
+
+    SnowflakeIO.Read<GenericRecord> read =
+        SnowflakeIO.<GenericRecord>read(snowflakeService, snowflakeCloudProvider)
+            .withDataSourceConfiguration(dataSourceConfiguration)
+            .fromTable(FAKE_TABLE)
+            .withStagingBucketName(stagingBucketName)
+            .withCsvMapper(getCsvMapper())
+            .withCoder(AvroCoder.of(AvroGeneratedUser.getClassSchema()));
+
+    read.expand(null);
```
chamikaramj, line 123: Can we introduce tests to actually read data (from SnowFlakeFakeServiceImpl) and verify results. Please see DirectRunner-based BQ pipeline read tests.

https://github.com/apache/beam/blob/master/sdks/java/io/google-cloud-platform/src/test/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryIOReadTest.java#L285

Also there's no need to manually call 'expand' in tests. We can just execute regular Beam pipelines that will perform the expansion.

sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/SnowflakeIO.java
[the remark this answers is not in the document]
aromanenko-dev: In this case, I'd suggest to add this into IO class Java doc to make users aware that such situation is possible and it will require manual procedure to clean temp dirs up.

aromanenko-dev:
retest this please

aromanenko-dev:
retest this please

sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/SnowflakeIO.java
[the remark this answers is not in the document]
DariuszAniszewski: I've added a note [here](https://github.com/apache/beam/pull/11360/commits/6f3ff11d758a45619ff25b6e5f270c2c970fce05#diff-de3ace437cd24772382c19f0c5d9a062R115) - would it be enough?

sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/SnowflakeIO.java
```diff
@@ -0,0 +1,768 @@
...
+     * A table name to be read in Snowflake.
+     *
+     * @param table - String with the name of the table.
+     */
+    public Read<T> fromTable(String table) {
+      return toBuilder().setTable(table).build();
+    }
+
+    /**
+     * Name of the bucket to use as tmp location of CSVs during COPY statement.
```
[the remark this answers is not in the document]
DariuszAniszewski: GCS currently. Any of the Snowflake-supported cloud storages are possible in the future.

sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/SnowflakeIO.java
```diff
@@ -0,0 +1,768 @@
...
+import org.apache.beam.sdk.values.PCollection;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * IO to read and write data on Snowflake.
+ *
+ * <p>SnowflakeIO uses <a href="https://docs.snowflake.net/manuals/user-guide/jdbc.html">Snowflake
+ * JDBC</a> driver under the hood, but data isn't read/written using JDBC directly. Instead,
+ * SnowflakeIO uses dedicated <b>COPY</b> operations to read/write data from/to Google Cloud
```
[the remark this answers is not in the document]
DariuszAniszewski: It's GCS currently, but in the future other cloud providers can be added. I've made changes to the docs - PTAL

sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/SnowflakeIO.java
DariuszAniszewski: Thanks for the better description. 
There is no default coder, user must always set coder - is this a problem?

sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/SnowflakeIO.java
DariuszAniszewski: Yes, the story behind it is to make connector more user-friendly. 

We wanted to handle the CSV parsing for the user and left them only mapping from `Object[]` to `T`. 
Otherwise the user would have to do CSV line parsing on their own and face the same issues that we could fix for them (like having a comma in string) by setting proper delimiters and quoting when calling COPY statement. 

Do you think it's OK like this?

kkucharc:
retest this please

sdks/java/io/snowflake/src/test/java/org/apache/beam/sdk/io/snowflake/test/unit/read/SnowflakeIOReadTest.java
kkucharc: Thank you @chamikaramj for pointing this. I changed  this tests so now it uses `pipeline.run()` instead of `expand`. The interesting thing is that I had to use `TestRule` similarly as in BQ tests (otherwise tests were failing with `Outputs for non-root node SnowflakeIO.Read are null`), do you know why is that? Is it something connected with "clearing" the pipeline for each test?

In case of reading data by `SnowFlakeFakeServiceImpl` there is a test `testReadWithConfigIsProper` (previously `testConfigIsProper`) [L248](https://github.com/apache/beam/pull/11360/commits/c5abe5ab21d040e45ab75f95dd93cfa1490b4bf9#diff-8b70435e69bb0a87cf801ea7c4ad3c4e) which does that. Do you think there should be more tests like this?

sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/SnowflakeCloudProvider.java
```diff
@@ -0,0 +1,29 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.beam.sdk.io.snowflake;
+
+import java.io.IOException;
+
+/** Interface which defines common methods for cloud providers. */
```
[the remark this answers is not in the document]
purbanow: Yes, you're right :) , I got rid of `SnowfalkeCloudProvider` with result of removing 200 lines of code. Thanks a lot @chamikaramj for spotting this

DariuszAniszewski:
@chamikaramj Thanks for review. We've applied some of them already and answered/commented to rest of them - can you please have another look on this PR?

aromanenko-dev:
retest this please

aromanenko-dev:
retest this please

sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/SnowflakeIO.java
```diff
@@ -0,0 +1,735 @@
...
+ *    .withStagingBucketName(stagingBucketName)
+ *    .withIntegrationName(integrationName)
+ *    .withCsvMapper(...)
+ *    .withCoder(...));
+ * }</pre>
+ *
+ * <p><b>Important</b> When reading data from Snowflake, temporary CSV files are created on the
+ * specified stagingBucketName in directory named `sf_copy_csv_[RANDOM CHARS]_[TIMESTAMP]`. This
+ * directory and all the files are deleted automatically by default, but in case of failed pipeline
+ * they will remain and have to be removed manually.
```
chamikaramj, line 122: they may remain and will have to ...

sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/SnowflakeIO.java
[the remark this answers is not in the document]
chamikaramj: Probably you should provide a withoutValidation() option so that users who submit a pipeline form a node that do not have access to the server can disable this.

sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/SnowflakeIO.java
```diff
@@ -0,0 +1,735 @@
...
+      String gcpTmpDirName = makeTmpDirName();
+      String stagingBucketDir = String.format("%s/%s/", getStagingBucketName(), gcpTmpDirName);
+
+      PCollection<Void> emptyCollection = input.apply(Create.of((Void) null));
+
+      PCollection<T> output =
+          emptyCollection
+              .apply(
+                  ParDo.of(
+                      new CopyIntoStageFn(
```
chamikaramj, line 320: What will happen in case of a failure and a retry of a bundle ?

I think you should add a Reshuffle step after CopyIntoStageFn() step to checkpoint the set of files (supported by most runners).

Also, make sure you have considered what would happen if CopyIntoStageFn() itself fails and retries. For example, BQ source tries to look for an already started export job before starting a new one.

sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/SnowflakeIO.java
chamikaramj: Sounds good.

sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/SnowflakeIO.java
```diff
@@ -0,0 +1,735 @@
...
+
+      @ProcessElement
+      public void processElement(ProcessContext c) throws IOException {
+        String combinedPath = stagingBucketDir + "*";
+        List<ResourceId> paths =
+            FileSystems.match(combinedPath).metadata().stream()
+                .map(metadata -> metadata.resourceId())
+                .collect(Collectors.toList());
+
+        FileSystems.delete(paths);
```
chamikaramj, line 421: Probably specify MoveOptions.IGNORE_MISSING_FILES to not fail incase this bundle is retries after a failure (either the source or the destination should exist).
https://github.com/apache/beam/blob/master/sdks/java/core/src/main/java/org/apache/beam/sdk/io/fs/MoveOptions.java#L30

sdks/java/io/snowflake/src/test/java/org/apache/beam/sdk/io/snowflake/test/unit/read/SnowflakeIOReadTest.java
```diff
@@ -0,0 +1,268 @@
...
+            .withIntegrationName(integrationName)
+            .withCsvMapper(getCsvMapper())
+            .withCoder(AvroCoder.of(AvroGeneratedUser.getClassSchema())));
+
+    pipeline.run();
+  }
+
+  @Test
+  public void testReadWithConfigIsProper() {
+    PCollection<GenericRecord> items =
```
chamikaramj, line 245: Please make sure that all read paths are covered by tests.

sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/SnowflakeIO.java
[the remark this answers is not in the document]
DariuszAniszewski: That's good idea, added

sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/SnowflakeIO.java
DariuszAniszewski: If running `COPY` fails and it will be retried it will fail on any retry. 
It's because the first attempt will put some files into staging bucket and another try will find out that staging location is not empty and it will fail.

There is a `OVERWRITE` option in Snowflake, telling that if there are files on cloud storage they will be overwritten, but we don't use it. The reason is simple - we're reading all data from this temporary location. If for any reason, on this location there will be garbage files, data will be malformed. I imagine it might happen that first attempt will put some files, and another attempt might have slightly different filenames and files will be appended. In this case we have partial results from first try and full results from second - we don't want that.

There is also no such thing as "job" in this case. It's just, simple, standalone "plain old good sql query" that is executed against snowflake. It is parallelised, but internally, on Snowflake side - from pipeline perspective it's one-off task. It doesn't have a job_id which status can be checked and resumed or waited for finish.

I'd suggest to keep it as is - failed pipeline in this case might be better solution than broken data. WDYT?

sdks/java/io/snowflake/src/test/java/org/apache/beam/sdk/io/snowflake/test/unit/read/SnowflakeIOReadTest.java
DariuszAniszewski: I've just added tests for `withQuery` as it was missing and fixed tests in general

DariuszAniszewski:
@chamikaramj All done, PTAL ;)

sdks/java/io/snowflake/src/main/java/org/apache/beam/sdk/io/snowflake/SnowflakeIO.java
chamikaramj: It's better but I don't think it's a good user experience to fail the pipeline whenever a bundle is retried. Bundle retries are a usual part of the runner execution and source/sink transforms should be able to handle that. 

What is usually done in this case is to write data to a new temporary location each time the copy step is run. Then you add a Reshuffle write after the copy operation to checkpoint the results. This way you can guarantee that the set of files that are output to the next step are from a single execution and the job does not have to fail simply because a bundle is retried.

When you cleanup you have to clean all temporary locations including retries.

I'm fine if you don't want to handle this in the first try but let's at least add a TODO and a JIRA to fix this later.

DariuszAniszewski: Thanks @chamikaramj for insisting on this. The approach you suggested works perfectly and was easy to incorporate into our case.

I've made changes that are creating a new sub-directory for each attempt of `CopyIntoStageFn`. If this Fn is retried, another subdirectory is created and if step succeeds, exactly this subdirectory will be passed to `FileIO` to further processing. In the end, all sub-directories are cleaned-up.

----------------

I'm however not sure about `Reshuffle`. In our case output from `CopyIntoStageFn` is single String containing cloud storage path from where files should be read by FileIO. Does it make sense to reshuffle it? I've tested it and it works well with retries without reshuffle.
Also, Reshuffle is deprecated and might not work well in portability (if I understand docs properly) and we're aiming to use cross-language to support Python SDK.
Do you think the `Reshuffle` step is needed here?

chamikaramj: So you have following two steps that will get fused by the runner and run together.
ParDo(CopyIntoStageFn)+ParDo(ReadFiles)

Now assume there's a failure when reading files. This will result in expensive CopyIntoStageFn from being executed again.

Reshuffle (for most runners) has the affect of breaking fusion and hence checkpointing results of ParDo(CopyIntoStageFn). So if there's a failure when reading, the ParDo(CopyIntoStageFn) step will not have to be re-run. Also this is a very inexpensive shuffle (we only send file-names through it). 

Reshuffle is deprecated since its behavior across runners is not well defined. Some runners may consider it to be an identity operation while most runners will checkpoint results. I don't think it can be removed from the SDK anytime soon and many widely used connectors depend on it.

DariuszAniszewski: Thanks @chamikaramj for detailed explanation. I've added `Reshuffle` step in [latest commit](https://github.com/apache/beam/pull/11360/commits/3ba192a57217a9b251f71f77771b68ceb2882b02)

chamikaramj:
LGTM. Thanks.

chamikaramj:
Retest tthis please

chamikaramj:
Retest this please

chamikaramj:
I can "squash and merge" after tests pass.
Feel free to squash/fixup commits if you need more than one commit.

chamikaramj:
Retest this please

DariuszAniszewski:
Thanks!
I'm OK with squash-and-merge - it was intended to go as single commit

chamikaramj:
Retest this please

chamikaramj:
Retest this please

chamikaramj:
Having trouble re-triggering tests.
Can someone else try ?

chamikaramj:
Retest this please

chamikaramj:
Run Python PreCommit

chamikaramj:
Run Python PreCommit

chamikaramj:

chamikaramj:

chamikaramj:

kkucharc:

DariuszAniszewski:
Just a small comment about the force-push from above - it was mistakenly done, then reverted. HEAD of this branch is still on **3ba192a** and comment is a leftover.

kkucharc:
I retested failing test - probably the previous one was timeouting.

chamikaramj:
Seems like this was not included in Beam 2.22.0 cut. So I'll remove it from CHANGES.md entry for Beam 2.22.0.

PR #11362, opened 2020-04-09
[BEAM-9733] Always let ImpulseSourceFunction emit a final watermark

mxm:
This contains two fixes which are tight to one another:

    [BEAM-9733] Always let ImpulseSourceFunction emit a final watermark

    The Flink Runner's ImpulseSourceFunction does not emit a final watermark, unless
    the `--shutdownSourcesOnFinalWatermark` flag has been specified (the flag is
    used in tests to shutdown the pipeline after reading all data). Most pipelines
    will be long-running and thus do not specify the flag.

    Not sending out the final watermark causes GroupByKey to hold back the data of
    event time windows until the pipeline is shut down (the final watermark is
    always emitted on pipeline shutdown which is why using the above flag works).

###

    [BEAM-9733] Make up for timers set while processing the bundle

    Due to the asynchronous nature of the bundle processing which includes access to
    state and timers, the output watermark can only be advanced after the current
    bundle has finished. Furthermore, we have to take care to only advance it to the
    lowest newly set timer in the finished bundle. Otherwise we would be violating
    the watermark contract.

    Besides the problem of advancing the output watermark too soon, the current
    model was also overly restrictive with handling the input watermark. The input
    watermark would be pushed back until the end of the bundle. This is not
    necessary as we can advance the input watermark freely during bundle processing
    which can trigger timers as soon as they become eligible for firing. Special
    care has to be taken to handle cleanup timers which may already fire when we
    haven't collected all the timers from the SDK Harness. In case we end up with
    new timers for a window for which the cleanup timer scheduled a cleanup at the
    end of the bundle, we have to defer the cleanup and reschedule the cleanup
    timer. Otherwise, we would lose all state for a timer.

--- | --- | --- | --- | ---

### DIFF
--- a/runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/DoFnOperator.java
+++ b/runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/DoFnOperator.java
@@ -116,6 +116,8 @@ import org.apache.flink.streaming.runtime.streamrecord.StreamRecord;
 import org.apache.flink.streaming.runtime.tasks.StreamTask;
 import org.apache.flink.util.OutputTag;
 import org.joda.time.Instant;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Flink operator for executing {@link DoFn DoFns}.
@@ -123,10 +125,14 @@ import org.joda.time.Instant;
  * @param <InputT> the input type of the {@link DoFn}
  * @param <OutputT> the output type of the {@link DoFn}
  */
+// We use Flink's lifecycle methods to initialize transient fields
+@SuppressFBWarnings("SE_TRANSIENT_FIELD_NOT_RESTORED")
 public class DoFnOperator<InputT, OutputT> extends AbstractStreamOperator<WindowedValue<OutputT>>
     implements OneInputStreamOperator<WindowedValue<InputT>, WindowedValue<OutputT>>,
         TwoInputStreamOperator<WindowedValue<InputT>, RawUnionValue, WindowedValue<OutputT>>,
         Triggerable<ByteBuffer, TimerData> {
+
+  private static final Logger LOG = LoggerFactory.getLogger(DoFnOperator.class);
 
   protected DoFn<InputT, OutputT> doFn;
 
@@ -206,14 +212,16 @@ public class DoFnOperator<InputT, OutputT> extends AbstractStreamOperator<Window
   private transient volatile long elementCount;
   /** Time that the last bundle was finished (to set the timer). */
   private transient volatile long lastFinishBundleTime;
+  /** Callback to be executed before the current bundle is started. */
+  private transient volatile Runnable preBundleCallback;
   /** Callback to be executed after the current bundle was finished. */
   private transient volatile Runnable bundleFinishedCallback;
 
   // Watermark state.
   // Volatile because these can be set in two mutually exclusive threads (see above).
-  protected transient volatile long currentInputWatermark;
-  protected transient volatile long currentSideInputWatermark;
-  protected transient volatile long currentOutputWatermark;
+  private transient volatile long currentInputWatermark;
+  private transient volatile long currentSideInputWatermark;
+  private transient volatile long currentOutputWatermark;
   private transient volatile long pushedBackWatermark;
 
   /** Constructor for DoFnOperator. */
@@ -353,9 +361,9 @@ public class DoFnOperator<InputT, OutputT> extends AbstractStreamOperator<Window
       pushedBackElementsHandler = NonKeyedPushedBackElementsHandler.create(listState);
     }
 
-    setCurrentInputWatermark(BoundedWindow.TIMESTAMP_MIN_VALUE.getMillis());
-    setCurrentSideInputWatermark(BoundedWindow.TIMESTAMP_MIN_VALUE.getMillis());
-    setCurrentOutputWatermark(BoundedWindow.TIMESTAMP_MIN_VALUE.getMillis());
+    currentInputWatermark = BoundedWindow.TIMESTAMP_MIN_VALUE.getMillis();
+    currentSideInputWatermark = BoundedWindow.TIMESTAMP_MIN_VALUE.getMillis();
+    currentOutputWatermark = BoundedWindow.TIMESTAMP_MIN_VALUE.getMillis();
 
     sideInputReader = NullSideInputReader.of(sideInputs);
 
@@ -371,9 +379,9 @@ public class DoFnOperator<InputT, OutputT> extends AbstractStreamOperator<Window
       Stream<WindowedValue<InputT>> pushedBack = pushedBackElementsHandler.getElements();
       long min =
           pushedBack.map(v -> v.getTimestamp().getMillis()).reduce(Long.MAX_VALUE, Math::min);
-      setPushedBackWatermark(min);
+      pushedBackWatermark = min;
     } else {
-      setPushedBackWatermark(Long.MAX_VALUE);
+      pushedBackWatermark = Long.MAX_VALUE;
     }
 
     // StatefulPardo or WindowDoFn
@@ -548,15 +556,20 @@ public class DoFnOperator<InputT, OutputT> extends AbstractStreamOperator<Window
     }
   }
 
-  protected long getPushbackWatermarkHold() {
-    return pushedBackWatermark;
+  public long getEffectiveInputWatermark() {
+    // hold back by the pushed back values waiting for side inputs
+    return Math.min(pushedBackWatermark, currentInputWatermark);
   }
 
-  protected void setPushedBackWatermark(long watermark) {
-    pushedBackWatermark = watermark;
+  public long getCurrentOutputWatermark() {
+    return currentOutputWatermark;
   }
 
-  protected void setBundleFinishedCallback(Runnable callback) {
+  protected final void setPreBundleCallback(Runnable callback) {
+    this.preBundleCallback = callback;
+  }
+
+  protected final void setBundleFinishedCallback(Runnable callback) {
     this.bundleFinishedCallback = callback;
   }
 
@@ -580,7 +593,7 @@ public class DoFnOperator<InputT, OutputT> extends AbstractStreamOperator<Window
       min = Math.min(min, pushedBackValue.getTimestamp().getMillis());
       pushedBackElementsHandler.pushBack(pushedBackValue);
     }
-    setPushedBackWatermark(min);
+    pushedBackWatermark = min;
 
     checkInvokeFinishBundleByCount();
   }
@@ -634,7 +647,7 @@ public class DoFnOperator<InputT, OutputT> extends AbstractStreamOperator<Window
       min = Math.min(min, pushedBackValue.getTimestamp().getMillis());
       pushedBackElementsHandler.pushBack(pushedBackValue);
     }
-    setPushedBackWatermark(min);
+    pushedBackWatermark = min;
 
     checkInvokeFinishBundleByCount();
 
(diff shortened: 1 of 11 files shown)


mxm:

tweise:
Just to confirm: Emitting the final watermark won't terminate the source function and therefore checkpointing will still succeed?

mxm:
> Just to confirm: Emitting the final watermark won't terminate the source function and therefore checkpointing will still succeed?

Yes, the two are unrelated to each other.

mxm:
The flag `shutdownSourcesOnFinalWatermark` might be confusing because it does not actually shutdown on final watermark but instead shuts down the source function which causes Flink to emit the final watermark.

iemejia:
Any plans on how to make the flag less confusing?

mxm:
I need to rewrite the FlinkSavepointTest since it assumed different semantics. I think we can use the recently introduced timer output timestamp feature. However, I realized the portable operator needs a slight adjustment to fully support holding back the output timestamp correctly at all times. This is trickier than in the non-portable operator with respect to timers setting new timers; we do not have a direct feedback loop as we have in the non-portable operator. We need an additional check when a timer sets a new timer with a timer output timestamp because we only get to fire that timer after we started a new bundle. Thus, we can't always advance the watermark after we finish bundle execution.

mxm:
> Any plans on how to make the flag less confusing?

We could deprecate the flag and add an alias for backwards-compatibility.

mxm:

mxm:
Test case failing has been failing before these changes. I'll look into the cause. https://builds.apache.org/job/beam_PostCommit_Java_PVR_Flink_Streaming_PR/191/

mxm:
Failure was caused by https://github.com/apache/beam/pull/11314.

mxm:

mxm:
Tests should be passing now, though I'll give this another thought over night.

mxm:

mxm:
I've had another go at the solution because I wasn't entirely satisfied. This allowed me to remove dependencies on the base DoFnOperator, e.g. the abuse of the watermark hold intended for side inputs for holding back the watermark in portable bundles.

mxm:
Unrelated test failure:
```
15:55:21 FAILURE: Build failed with an exception.
15:55:21 
15:55:21 * What went wrong:
15:55:21 Execution failed for task ':sdks:java:container:generateThirdPartyLicenses'.
15:55:21 > Process 'command './sdks/java/container/license_scripts/license_script.sh'' finished with non-zero exit value 1
```

mxm:
Run Java PreCommit

mxm:
> @mxm these are significant changes after the PR was approved. Questions:

I'm aware of that. I've tried to keep the necessary changes at a minimum. I'm happy to hear your feedback.

> 1. What test coverage do we have for ensuring that watermarks don't bypass in-progress elements?

We have tests in `ExecutableStageDoFnOperatorTest` (`outputsAreTaggedCorrectly`, `testEnsureDeferredStateCleanupTimerFiring`) and through our integration tests. Admittedly, it would be good to dedicate a test specifically to watermark behavior.

> 2. Do these changes affect how the main input watermark interacts with the side input watermark?

Effectively, side inputs in portability were broken before because (a) the side input watermark hold was abused by the the portable operator (b) only `processWatermark` was overridden but for proper support for side inputs we have to override `processWatermark1` (1 is the main input when we have side inputs, `DoFnoperator#processWatermark` calls `processWatermark1` when we do not have side inputs, but `processWatermark` is not called when we have side input, only `processWatermark1`).
 
> 3. Will the added watermark logging affect the usefulness of debug logging for other investigations (I had in the past removed it after done debugging issues)

It greatly helped me to debug the current behavior and develop the solution. I find it immensely helpful and would like to keep it if further debugging is necessary. It is very useful to have debug information already built-in, instead of having to add it manually every time (which in any case will be possible).

tweise:
> We have tests in `ExecutableStageDoFnOperatorTest` (`outputsAreTaggedCorrectly`, `testEnsureDeferredStateCleanupTimerFiring`) and through our integration tests. Admittedly, it would be good to dedicate a test specifically to watermark behavior.

There is some coverage here: https://github.com/apache/beam/blob/8db19a4645b8588ce9e046637b7619815169bdb1/runners/flink/src/test/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperatorTest.java#L376

It covers only finish bundle on close though. But a similarly fine-grained test should do (integration tests generally don't provide a signal for this).

ibzib:
What remains to be done on this PR? Test coverage? I'd like to get this merged as soon as possible to unblock the 2.21 release.

mxm:
I apologize for the delay here. Turns out we had a bit of work to do with regards to handling watermark and timers in portability. Please see the PR description for an update on what has been changed. @tweise @iemejia Please have a look. I'm sorry that there are more changes to review now but I'm convinced the implementation is now correct and robust. I'm also expecting considerable performance gains for portability.

mxm:
Run Flink ValidatesRunner

mxm:

mxm:
Run Nexmark Flink

mxm:
Flink Runner Nexmark Tests

mxm:
Run Flink Runner Nexmark Tests

mxm:
Third party licensing checks fail in: https://builds.apache.org/job/beam_PreCommit_Python2_PVR_Flink_Commit/4209/

mxm:

mxm:
Run Java PreCommit

mxm:
Run Java PreCommit

mxm:
There is still an issue with the bundle timeout timer which only became visible on Jenkins. Need to look into this but the principle approach holds here.

mxm:
Run Flink ValidatesRunner

mxm:

mxm:
Run Flink Runner Nexmark Tests

mxm:
Run Flink ValidatesRunner

mxm:
Run Java PreCommit

mxm:

mxm:
Run Flink ValidatesRunner

mxm:
Run Flink Runner Nexmark Tests

mxm:
Unrelated test failure, batch test for SDF fails:

```
21:40:10 org.apache.beam.sdk.transforms.SplittableDoFnTest > testPairWithIndexWindowedTimestampedBounded FAILED
21:40:10     org.apache.beam.sdk.Pipeline$PipelineExecutionException at SplittableDoFnTest.java:224
21:40:10         Caused by: java.lang.NullPointerException
21:40:34 
21:40:34 org.apache.beam.sdk.transforms.SplittableDoFnTest > testPairWithIndexBasicBounded FAILED
21:40:34     org.apache.beam.sdk.Pipeline$PipelineExecutionException at SplittableDoFnTest.java:157
21:40:34         Caused by: java.lang.NullPointerException
21:41:25 
21:41:25 242 tests completed, 2 failed, 2 skipped
21:41:32 
21:41:32 > Task :runners:flink:1.10:validatesRunnerBatch FAILED
```
https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Flink_PR/239/

tweise:
I will take another pass shortly, but overall the changes look good. With the improved test coverage, we will be more confident that future modifications don't break the watermark handling. Given the scope of changes, it would be good to put this through regression testing with a real application.

mxm:
Thanks @tweise. I started running tests with a "real application". I haven't observed any regressions. Throughput/latency look identical. The application doesn't make use of user timers though but it does exercise most code paths of the changes.

FYI the SDF issues are caused by https://github.com/apache/beam/pull/11418#issuecomment-617021782

ibzib:
@tweise I'm trying to clear out remaining release blockers, so please give a final review for this soon.

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/DoFnOperator.java
```diff
@@ -658,46 +671,69 @@ public void processWatermark1(Watermark mark) throws Exception {
...
     }
-    output.emitWatermark(new Watermark(watermark));
   }
 
   @Override
-  public void processWatermark2(Watermark mark) throws Exception {
-
-    setCurrentSideInputWatermark(mark.getTimestamp());
+  public final void processWatermark2(Watermark mark) throws Exception {
+    currentSideInputWatermark = mark.getTimestamp();
```
tweise, line 736: Don't we have to emit pushed back data at this point (up to the side input watermark), even when it is not the final watermark?

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/ExecutableStageDoFnOperator.java
```diff
@@ -544,30 +601,49 @@ public void processWatermark(Watermark mark) throws Exception {
...
+    minEventTimeTimerTimestampInLastBundle = minEventTimeTimerTimestampInCurrentBundle;
+    minEventTimeTimerTimestampInCurrentBundle = Long.MAX_VALUE;
+    try {
+      if (!closed
+          && minEventTimeTimerTimestampInLastBundle < Long.MAX_VALUE
+          && minEventTimeTimerTimestampInLastBundle <= getEffectiveInputWatermark()) {
+        ProcessingTimeService processingTimeService = getProcessingTimeService();
+        // We are scheduling a timer for advancing the watermark. Otherwise we
+        // could potentially loop forever here when a timer keeps scheduling a timer
+        // for the same timestamp. This in itself would not be an issue. However,
```
tweise, line 634: The comment is a bit misleading. How about: "We are scheduling a timer for advancing the watermark, to not delay finishing the bundle and temporarily release the checkpoint lock. Otherwise, we could potentially loop when a timer keeps scheduling a timer for the same timestamp."

mxm: Sure. Changed.

tweise:
@mxm I looked at the metrics of the internal test run and see an increase of ~50% in task manager CPU utilization. That would be significant, but the test also had low throughput and overall low utilization. It would be good to run a benchmark with higher base throughput and CPU utilization to ensure this isn't a performance regression.

mxm:
I don't see how this change can affect the CPU performance much. Apart from enumerating the timers while executing the user state cleanup there is no CPU intense code in the changes. But even the timer lookup should be very efficient because we only operate on a specific key in the keyed state backend. The test results are likely skewed due to the infrastructure, e.g. pod deployment or utilization of the underlying machine. I'll run another test to verify though.

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/DoFnOperator.java
tweise: Yes, it is independent of this PR. The pre-existing side input related issue still exists. When no side input arrives, elements will never be emitted unless the final watermark occurs.

tweise:
It might very well be a test execution issue. From looking at the changes I would not expect a significant utilization change. The only contributing piece in this changeset appears the timer loop at the end of the bundle through `hasPendingEventTimeTimers`.

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/DoFnOperator.java
mxm: I agree, that is something we should address.

mxm:
> The only contributing piece in this changeset appears the timer loop at the end of the bundle through `hasPendingEventTimeTimers`.

Exactly, but that should not be expensive because we are iterating over timers only for that particular key.

tweise:
The additional test run shows only a negligible CPU utilization difference. LGTM

mxm:
Thanks! I also ran another test and found that depending on chance sometimes the old, sometimes the new implementation would be slightly more CPU expensive.

PR #11367, opened 2020-04-09
Revert "[BEAM-9651] Prevent StreamPool and stream initialization livelock"

scwhittle:
Reverts apache/beam#11364 because the tests were not run before merging.


scwhittle:
R: @reuvenlax

scwhittle:
R: @lukecwik

PR #11405, opened 2020-04-13
[BEAM-9747] Deprecate RedisIO.readAll() and add RedisIO.readKeyPatterns as a replacement

iemejia:
The idea is to let this deprecation for two releases before introducing the ReadAll pattern based on Reads for RedisIO to like proposed in [BEAM-9403](https://issues.apache.org/jira/browse/BEAM-9403).

R: @aromanenko-dev

### DIFF
--- a/sdks/java/io/redis/src/main/java/org/apache/beam/sdk/io/redis/RedisIO.java
+++ b/sdks/java/io/redis/src/main/java/org/apache/beam/sdk/io/redis/RedisIO.java
@@ -80,13 +80,13 @@ import redis.clients.jedis.ScanResult;
  *
  * }</pre>
  *
- * <p>{@link #readAll()} can be used to request Redis server using input PCollection elements as key
- * pattern (as String).
+ * <p>{@link #readKeyPatterns()} can be used to request Redis server using input PCollection
+ * elements as key pattern (as String).
  *
  * <pre>{@code
  * pipeline.apply(...)
  *    // here we have a PCollection<String> with the key patterns
- *    .apply(RedisIO.readAll().withEndpoint("::1", 6379))
+ *    .apply(RedisIO.readKeyPatterns().withEndpoint("::1", 6379))
  *   // here we have a PCollection<KV<String,String>>
  *
  * }</pre>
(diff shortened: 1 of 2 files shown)


iemejia:
Run Java PreCommit

sdks/java/io/redis/src/main/java/org/apache/beam/sdk/io/redis/RedisIO.java
```diff
@@ -123,8 +123,24 @@ public static Read read() {
...
+
+  /**
+   * Like {@link #read()} but executes multiple instances of the Redis query substituting each
+   * element of a {@link PCollection} as key pattern.
+   *
+   * @deprecated This method is not consistent with the readAll pattern of other transforms and will
+   *     be remove soon. Please update you code to use {@link #readKeyPatterns()} instead.
+   */
+  @Deprecated
+  public static ReadKeyPatterns readAll() {
```
aromanenko-dev, line 142: There is API breakage here since some users could rely on the fact that `readAll()` returns `ReadAll`. I think we should keep the same signature for now but just deprecate it. Then remove it after 3 releases or change it to make consistent with all other `ReadAll` patterns that take `PCollection<Read>` as an input.

In the same time, we still can add new transform `ReadKeyPatterns`, available through `readKeyPatterns()`, and ask users to use it instead.

iemejia: Yes I will adjust it. Thanks

iemejia:
PR updated with the suggested changes can you PTAL @aromanenko-dev

aromanenko-dev:
Run Java PreCommit

iemejia:
Thanks Alexey!

sdks/java/io/redis/src/main/java/org/apache/beam/sdk/io/redis/RedisIO.java
n3lsok: Hello! how can I to use this version ?? I cant' see how to compile and to use with an development which I have problems using RedisIO.readAll(), as I see this versión solved this issue. Please your help, All the best!

PR #11406, opened 2020-04-13
[BEAM-9748] Refactor Reparallelize as an alternative Reshuffle implementation

iemejia:
Some DoFn based IOs like JdbcIO and RedisIO rely on the Reparallelize transform, a combination of a an empty PCollectionView and Reshuffle to force the materialization and reparallelize a PCollection. The idea of this issue is to extract this transform and expose it as part of the internal Reshuffle transform to avoid repeating the code for transforms (notably IOs) that require to reparallelize its output.

R: @lukecwik

### DIFF
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/Reshuffle.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/Reshuffle.java
@@ -140,7 +140,7 @@ public class Reshuffle<K, V> extends PTransform<PCollection<KV<K, V>>, PCollecti
         // and ensures that data to be shuffled can be generated in parallel, while reshuffling
         // provides perfect parallelism.
         // In most cases where a "fusion break" is needed, a simple reshuffle would be sufficient.
-        // The current approach is necessary only to support the particular case of JdbcIO where
+        // The current approach is necessary to support use cases such as JdbcIO where
         // a single query may produce many gigabytes of query results.
         PCollectionView<Iterable<T>> empty =
             input
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/Reshuffle.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/Reshuffle.java
@@ -27,6 +27,7 @@ import org.apache.beam.sdk.transforms.windowing.Window;
 import org.apache.beam.sdk.util.IdentityWindowFn;
 import org.apache.beam.sdk.values.KV;
 import org.apache.beam.sdk.values.PCollection;
+import org.apache.beam.sdk.values.PCollectionView;
 import org.apache.beam.sdk.values.TimestampedValue;
 import org.apache.beam.sdk.values.WindowingStrategy;
 import org.joda.time.Duration;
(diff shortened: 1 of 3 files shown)


sdks/java/io/redis/src/main/java/org/apache/beam/sdk/io/redis/RedisIO.java
```diff
@@ -309,7 +305,7 @@ public ReadAll withOutputParallelization(boolean outputParallelization) {
               .apply(ParDo.of(new ReadFn(connectionConfiguration(), batchSize())))
               .setCoder(KvCoder.of(StringUtf8Coder.of(), StringUtf8Coder.of()));
       if (outputParallelization()) {
-        output = output.apply(new Reparallelize());
+        output = (PCollection<KV<String, String>>) output.apply(Reshuffle.reparallelize());
```
lukecwik, line 308: nit: Why is the cast necessary here?

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/Reshuffle.java
```diff
@@ -65,6 +66,11 @@ private Reshuffle() {}
     return new ViaRandomKey<>();
   }
 
+  @Experimental
+  public static <T> Reparallelize<T> reparallelize() {
```
lukecwik, line 70: It is unclear to a user why they would choose this implementation over viaRandomKeys

iemejia:
> Why do we want to do this over Reshuffle.viaRandomKeys which should get us the output parallelization we want?
I think that's the case, maybe @jkff who created that code may confirm.

sdks/java/io/redis/src/main/java/org/apache/beam/sdk/io/redis/RedisIO.java
iemejia: Not necessary, I will remove it.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/Reshuffle.java
iemejia: Yes I will improve the javadoc, maybe move some of the details in the impl. comment below.

jkff:
The comments inside Reparallelize explain how this transform differs from Reshuffle.viaRandomKey(): it performs dramatically better on Dataflow in case the input PCollection is generated highly sequentially, as in the case of reading several GB of JDBC results. It almost certainly performs somewhat worse if the input PCollection is generated in a well-parallelized way, but I haven't measured that; I haven't measured the former case for non-Dataflow runners either.

I think it's reasonable to move this to Reshuffle, but rename it to something more clear: maybe Reshuffle.forSequentiallyGeneratedInput()?

iemejia:
Changes addressed and renamed to the suggested name by Eugene. PTAL again @lukecwik

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/Reshuffle.java
```diff
@@ -65,6 +66,16 @@ private Reshuffle() {}
     return new ViaRandomKey<>();
   }
 
+  @Experimental
```
jkff, line 69: Annotation should probably be below the comment.

Also suggest rephrasing a bit, to explain when one should use one or the other, and what are the consequences of choosing wrong. I'm not sure that "sequentially generated" is clear enough to a casual user. Maybe something like this:

```
Materializes the input and prepares it to be consumed in a highly parallel fashion.

This version is tailored to the case when input was produced in an extremely sequential way - typically by a ParDo that emits millions of outputs _per input element_, e.g., executing a large database query or a large simulation and emitting all of their results.

Internally, this version first materializes the input at a moderate cost before reshuffling it internally using viaRandomKey(), making the reshuffling itself significantly cheaper in these extreme cases on some runners. Use this over viaRandomKey() only if your benchmarks show an improvement.
```

And mention this at the class-level documentation of Reshuffle for visibility.

lukecwik:
> The comments inside Reparallelize explain how this transform differs from Reshuffle.viaRandomKey(): it performs dramatically better on Dataflow in case the input PCollection is generated highly sequentially, as in the case of reading several GB of JDBC results. It almost certainly performs somewhat worse if the input PCollection is generated in a well-parallelized way, but I haven't measured that; I haven't measured the former case for non-Dataflow runners either.
> 
> I think it's reasonable to move this to Reshuffle, but rename it to something more clear: maybe Reshuffle.forSequentiallyGeneratedInput()?

This was true at some point in time for Dataflow but I don't think we have rerun benchmarks to see how it performs now.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/Reshuffle.java
lukecwik: I think the key is that the input was generated with limited parallelism and suggest that we update `ViaRandomKey` to perform both expansions where the side input fusion break is added if the `withHintLimitedInputParallelism()` is specified.

jkff: I think this (`withHint...`) is a great idea, thanks Luke. It also wouldn't hurt to rerun the benchmarks (and to check them in - sadly I didn't do it back then, and probably don't have them anymore, but it only took like 15 min to write).

lukecwik:
retest this please

iemejia:
I had forgotten about this one. Suggestions made, can you PTAL again @lukecwik @jkff . Thanks!

iemejia:
Run Java PreCommit

iemejia:
Run Java PreCommit

iemejia:
Run Java PreCommit

lukecwik:
Run Java PreCommit

lukecwik:
Run Java PreCommit

lukecwik:
Run Java PreCommit

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/Reshuffle.java
```diff
@@ -107,10 +108,57 @@ public void processElement(
...
+     * <p>It is tailored to the case when input was produced in an extremely sequential way -
+     * typically by a ParDo that emits millions of outputs _per input element_, e.g., executing a
+     * large database query or a large simulation and emitting all of their results.
+     *
+     * <p>Internally, it materializes the input at a moderate cost before reshuffling it, making the
+     * reshuffling itself significantly cheaper in these extreme cases on some runners. Use this
+     * only if your benchmarks show an improvement.
+     */
+    public ViaRandomKey<T> withHintHighFanoutAndLimitedInputParallelism() {
+      this.isHighFanoutAndLimitedInputParallelism = true;
```
lukecwik, line 128: Lets return a new PTransform like our other PTransform builder patterns instead of mutating the state of the current transform.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/Reshuffle.java
```diff
@@ -107,10 +108,57 @@ public void processElement(
...
+      if (isHighFanoutAndLimitedInputParallelism) {
+        // See https://issues.apache.org/jira/browse/BEAM-2803
+        // We use a combined approach to "break fusion" here:
+        // (see https://cloud.google.com/dataflow/service/dataflow-service-desc#preventing-fusion)
+        // 1) force the data to be materialized by passing it as a side input to an identity fn,
+        // then 2) reshuffle it with a random key. Initial materialization provides some parallelism
+        // and ensures that data to be shuffled can be generated in parallel, while reshuffling
+        // provides perfect parallelism.
+        // In most cases where a "fusion break" is needed, a simple reshuffle would be sufficient.
+        // The current approach is necessary only to support the particular case of JdbcIO where
```
lukecwik, line 143:
```suggestion
        // The current approach is necessary to support use cases such as JdbcIO where
```

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/Reshuffle.java
```diff
@@ -107,10 +108,57 @@ public void processElement(
...
+        // a single query may produce many gigabytes of query results.
+        PCollectionView<Iterable<T>> empty =
+            input
+                .apply("Consume", Filter.by(SerializableFunctions.constant(false)))
+                .apply(View.asIterable());
+        PCollection<T> materialized =
+            input.apply(
+                "Identity",
+                ParDo.of(
```
jkff, line 152: nit: here you could use a `MapElements.via(Contextful.of(t -> t, requiresSideInputs(empty)))`

lukecwik:
ping?

PR #11454, opened 2020-04-17
[BEAM-8871] Support trySplit for ByteKeyRangeTracker

boyuanzz:
r: @lukecwik 

cc: @youngoli 

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/ByteKeyRangeTracker.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/ByteKeyRangeTracker.java
@@ -65,31 +65,61 @@ public class ByteKeyRangeTracker extends RestrictionTracker<ByteKeyRange, ByteKe
 
   @Override
   public SplitResult<ByteKeyRange> trySplit(double fractionOfRemainder) {
-    // TODO(BEAM-8871): Add support for splitting off a fixed amount of work for this restriction
-    // instead of only supporting checkpointing.
-
-    // If we haven't done any work, we should return the original range we were processing
-    // as the checkpoint.
-    if (lastAttemptedKey == null) {
-      ByteKeyRange rval = range;
-      // We update our current range to an interval that contains no elements.
-      range = NO_KEYS;
-      return SplitResult.of(range, rval);
+    // No split on an empty range.
+    if (NO_KEYS.equals(range)
+        || (!range.getEndKey().isEmpty() && range.getStartKey().equals(range.getEndKey()))) {
+      return null;
+    }
+    // There is no more remaining work after the entire range has been claimed.
+    if (lastAttemptedKey != null && lastAttemptedKey.isEmpty()) {
+      return null;
     }
 
-    // Return an empty range if the current range is done.
-    if (lastAttemptedKey.isEmpty()
-        || !(range.getEndKey().isEmpty() || range.getEndKey().compareTo(lastAttemptedKey) > 0)) {
-      return SplitResult.of(range, NO_KEYS);
+    ByteKey unprocessedRangeStartKey =
+        (lastAttemptedKey == null) ? range.getStartKey() : next(lastAttemptedKey);
+    ByteKey endKey = range.getEndKey();
+    // There is no more space for split.
+    if (!endKey.isEmpty() && unprocessedRangeStartKey.compareTo(endKey) >= 0) {
+      return null;
     }
 
-    // Otherwise we compute the "remainder" of the range from the last key.
-    assert lastAttemptedKey.equals(lastClaimedKey)
-        : "Expect both keys to be equal since the last key attempted was a valid key in the range.";
-    ByteKey nextKey = next(lastAttemptedKey);
-    ByteKeyRange res = ByteKeyRange.of(nextKey, range.getEndKey());
-    this.range = ByteKeyRange.of(range.getStartKey(), nextKey);
-    return SplitResult.of(range, res);
+    // Treat checkpoint specially because {@link ByteKeyRange#interpolateKey} computes a key with
+    // trailing zeros when fraction is 0.
+    if (fractionOfRemainder == 0.0) {
+      // If we haven't done any work, we should return the original range we were processing
+      // as the checkpoint.
+      if (lastAttemptedKey == null) {
+        // We update our current range to an interval that contains no elements.
+        ByteKeyRange rval = range;
+        range =
+            range.getStartKey().isEmpty()
+                ? NO_KEYS
+                : ByteKeyRange.of(range.getStartKey(), range.getStartKey());
+        return SplitResult.of(range, rval);
+      } else {
+        range = ByteKeyRange.of(range.getStartKey(), unprocessedRangeStartKey);
+        return SplitResult.of(range, ByteKeyRange.of(unprocessedRangeStartKey, endKey));
+      }
+    }
+
+    ByteKeyRange unprocessedRange = ByteKeyRange.of(unprocessedRangeStartKey, range.getEndKey());
+    ByteKey splitPos;
+    try {
+      // The interpolateKey shouldn't return empty key. Please refer to {@link
+      // ByteKeyRange#interpolateKey}.
+      splitPos = unprocessedRange.interpolateKey(fractionOfRemainder);
+      checkState(!splitPos.isEmpty());
+    } catch (Exception e) {
+      // There is no way to interpolate a key based on provided fraction.
+      return null;
+    }
+    // Computed splitPos is out of current tracking restriction.
+    if (!range.getEndKey().isEmpty() && splitPos.compareTo(range.getEndKey()) >= 0) {
+      return null;
+    }
+
+    range = ByteKeyRange.of(range.getStartKey(), splitPos);
+    return SplitResult.of(range, ByteKeyRange.of(splitPos, endKey));
   }
 
   /**
@@ -142,7 +172,8 @@ public class ByteKeyRangeTracker extends RestrictionTracker<ByteKeyRange, ByteKe
     // Handle checking the empty range which is implicitly done.
     // This case can occur if the range tracker is checkpointed before any keys have been claimed
     // or if the range tracker is checkpointed once the range is done.
-    if (NO_KEYS.equals(range)) {
+    if (NO_KEYS.equals(range)
+        || (!range.getEndKey().isEmpty() && range.getStartKey().equals(range.getEndKey()))) {
       return;
     }
 
@@ -154,6 +185,11 @@ public class ByteKeyRangeTracker extends RestrictionTracker<ByteKeyRange, ByteKe
     // Return if the last attempted key was the empty key representing the end of range for
     // all ranges.
     if (lastAttemptedKey.isEmpty()) {
+      return;
+    }
+
+    // The lastAttemptedKey is the last key of current restriction.
+    if (!range.getEndKey().isEmpty() && next(lastAttemptedKey).compareTo(range.getEndKey()) >= 0) {
       return;
     }
 
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/RestrictionTracker.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/RestrictionTracker.java
@@ -18,6 +18,7 @@
 package org.apache.beam.sdk.transforms.splittabledofn;
 
 import com.google.auto.value.AutoValue;
+import javax.annotation.Nullable;
 import org.apache.beam.sdk.annotations.Experimental;
 import org.apache.beam.sdk.annotations.Experimental.Kind;
 import org.apache.beam.sdk.transforms.DoFn;
@@ -34,7 +35,7 @@ import org.apache.beam.sdk.transforms.DoFn;
 public abstract class RestrictionTracker<RestrictionT, PositionT> {
   /**
    * Attempts to claim the block of work in the current restriction identified by the given
-   * position.
+   * position. Each claimed position MUST be a valid split point.
    *
    * <p>If this succeeds, the DoFn MUST execute the entire block of work. If this fails:
    *
@@ -83,8 +84,11 @@ public abstract class RestrictionTracker<RestrictionT, PositionT> {
    *
    * @param fractionOfRemainder A hint as to the fraction of work the primary restriction should
    *     represent based upon the current known remaining amount of work.
-   * @return a {@link SplitResult} if a split was possible, otherwise returns {@code null}.
+   * @return a {@link SplitResult} if a split was possible, otherwise returns {@code null}. If the
+   *     {@code fractionOfRemainder == 0}, a {@code null} result MUST imply that the restriction
+   *     tracker is done and there is no more work left to do.
    */
+  @Nullable
   public abstract SplitResult<RestrictionT> trySplit(double fractionOfRemainder);
 
   /**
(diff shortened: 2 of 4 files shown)


sdks/java/core/src/test/java/org/apache/beam/sdk/transforms/splittabledofn/ByteKeyRangeTrackerTest.java
```diff
@@ -287,4 +318,30 @@ public void testBacklogPartiallyCompleted() {
     tracker.tryClaim(ByteKey.of(0xa0));
     assertThat(tracker.getSize(), allOf(greaterThan(0.), lessThan(1.)));
   }
+
+  /** Asserts the two ByteKey are equal except trailing zeros. */
+  private static void assertByteKeyEqualExceptPadding(ByteKey expected, ByteKey key) {
```
boyuanzz, line 323: This is taken from https://github.com/apache/beam/blob/master/sdks/java/core/src/test/java/org/apache/beam/sdk/io/range/ByteKeyRangeTest.java#L365-L382

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/ByteKeyRangeTracker.java
```diff
@@ -64,31 +64,38 @@ public ByteKeyRange currentRestriction() {
...
-    // If we haven't done any work, we should return the original range we were processing
-    // as the checkpoint.
-    if (lastAttemptedKey == null) {
-      ByteKeyRange rval = range;
-      // We update our current range to an interval that contains no elements.
-      range = NO_KEYS;
-      return SplitResult.of(range, rval);
+    // No split on an empty range.
+    if (NO_KEYS.equals(range)) {
+      return null;
```
iemejia, line 69: Should this method then be `@Nullable`  and/or in the parent class?

sdks/java/core/src/main/java/org/apache/beam/sdk/io/range/ByteKeyRange.java
```diff
@@ -218,6 +218,10 @@ public double estimateFractionForKey(ByteKey key) {
   public ByteKey interpolateKey(double fraction) {
     checkArgument(
         fraction >= 0.0 && fraction < 1.0, "Fraction %s must be in the range [0, 1)", fraction);
+    // Return starKey when fraction is 0 in order to avoid adding trailing zeros during computation.
```
iemejia, line 221: s/starKey/startKey

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/ByteKeyRangeTracker.java
lukecwik: It should be in the parent class.

@robertwb was going to take a look at fixing up the docs and method signatures to ensure that null, None, nil were allowed and what they meant under certain scenarios.

boyuanzz: I can update  java `RestrictionTracker` within this PR if that's preferable.

boyuanzz:
Run Java PreCommit

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/RestrictionTracker.java
```diff
@@ -29,7 +30,7 @@
 public abstract class RestrictionTracker<RestrictionT, PositionT> {
   /**
    * Attempts to claim the block of work in the current restriction identified by the given
-   * position.
+   * position. Each claimed position should be a valid split point.
```
lukecwik, line 33:
```suggestion
   * position. Each claimed position MUST be a valid split point.
```

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/RestrictionTracker.java
```diff
@@ -80,6 +81,7 @@
    *     represent based upon the current known remaining amount of work.
    * @return a {@link SplitResult} if a split was possible, otherwise returns {@code null}.
```
lukecwik, line 82: We should make it clear that a checkpoint must return a valid split result or null where null represents that there is no residual.

sdks/java/core/src/test/java/org/apache/beam/sdk/transforms/splittabledofn/ByteKeyRangeTrackerTest.java
lukecwik: Make that method public and call it from this test since they are both in the same module.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/ByteKeyRangeTracker.java
lukecwik: That works for me.

iemejia: I am ok with `null` but couldn't we have modelled this with something more type friendly like `UnsplittableResult`

lukecwik: Since restriction trackers return `RestrictionT`, we could only do this if we updated the `trySplit` method to return `SplitResult<RestrictionT>` where `SplitResult` would be something like `Optional`.

That could be useful if we wanted to plumb the failure reason for the split back to the runner.

boyuanzz: The `trySplit` has returned `SplitResult<RestrictionT>` as result. If we want the runner understand the failure reason, we also need to create a set of well known literals. If just for logging purpose, we can also log the msg at SDK level.

robertwb: Semantically, what we want to return here is the empty split. It is important that the infrastructure be able to recognize and empty split (of type RestrictionT) so that it won't try to further split/reschedule it. One could create a new SplitResult type for Java, but I'm unconvinced it's worth its weight. (Certainly None/nil feel more appropriate for Python and Go respectively.)

sdks/java/core/src/test/java/org/apache/beam/sdk/transforms/splittabledofn/ByteKeyRangeTrackerTest.java
```diff
@@ -152,9 +158,35 @@ public void testCheckpointAfterLastUsingEmptyKey() throws Exception {
...
+        ByteKeyRangeTracker.of(ByteKeyRange.of(ByteKey.EMPTY, ByteKey.of(0x80)));
+    SplitResult<ByteKeyRange> res = tracker.trySplit(0.5);
+    assertKeyRangeEqualExceptPadding(
+        ByteKeyRange.of(ByteKey.EMPTY, ByteKey.of(0x40)), res.getPrimary());
+    assertKeyRangeEqualExceptPadding(
+        ByteKeyRange.of(ByteKey.of(0x40), ByteKey.of(0x80)), res.getResidual());
+    tracker.tryClaim(ByteKey.of(0x00));
+    res = tracker.trySplit(0.5);
+    assertKeyRangeEqualExceptPadding(
+        ByteKeyRange.of(ByteKey.of(0x00), ByteKey.of(0x20)), res.getPrimary());
```
lukecwik, line 178: Why is the primary here 0x00 for the start key and not ByteKey.EMPTY?

sdks/java/core/src/test/java/org/apache/beam/sdk/transforms/splittabledofn/ByteKeyRangeTrackerTest.java
```diff
@@ -152,9 +158,35 @@ public void testCheckpointAfterLastUsingEmptyKey() throws Exception {
...
+    assertNull(tracker.trySplit(0));
     assertEquals(ByteKeyRange.of(ByteKey.of(0x10), ByteKey.of(0xc0)), tracker.currentRestriction());
-    assertEquals(ByteKeyRangeTracker.NO_KEYS, checkpoint);
+    tracker.checkDone();
+  }
+
+  @Test
+  public void testTrySplit() throws Exception {
+    ByteKeyRangeTracker tracker =
+        ByteKeyRangeTracker.of(ByteKeyRange.of(ByteKey.EMPTY, ByteKey.of(0x80)));
```
lukecwik, line 169: Start with the ALL_KEYS range.

sdks/java/core/src/test/java/org/apache/beam/sdk/transforms/splittabledofn/ByteKeyRangeTrackerTest.java
```diff
@@ -152,9 +158,35 @@ public void testCheckpointAfterLastUsingEmptyKey() throws Exception {
...
+    assertKeyRangeEqualExceptPadding(
+        ByteKeyRange.of(ByteKey.of(0x00), ByteKey.of(0x20)), res.getPrimary());
+    assertKeyRangeEqualExceptPadding(
+        ByteKeyRange.of(ByteKey.of(0x20), ByteKey.of(0x40)), res.getResidual());
+    assertNull(tracker.trySplit(1));
+  }
+
+  @Test
+  public void testTrySplitAtEmptyRange() throws Exception {
+    ByteKeyRangeTracker tracker = ByteKeyRangeTracker.of(ByteKeyRangeTracker.NO_KEYS);
```
lukecwik, line 186: Cover the case where startKey == endKey and also add a call for trySplit at 0.5

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/ByteKeyRangeTracker.java
```diff
@@ -64,31 +64,56 @@ public ByteKeyRange currentRestriction() {
 
   @Override
   public SplitResult<ByteKeyRange> trySplit(double fractionOfRemainder) {
-    // TODO(BEAM-8871): Add support for splitting off a fixed amount of work for this restriction
-    // instead of only supporting checkpointing.
+    // No split on an empty range.
+    if (NO_KEYS.equals(range)) {
```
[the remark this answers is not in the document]
boyuanzz: I think the major question is about `startKey == endKey` or `NO_KEYS`. There are 2 options in my mind:
1. Use `startKey == endKey` and `NO_KEYS` to represent an empty key range. 
2. Only use `NO_KEYS` as a notion of empty key range. In order to do this, we need to change the current range to `NO_KEYS` after checkpointing if necessary.
I chose option 2 as my current implementation, that's why I only exam `NO_KEYS` in `tryClaim` and `checkDone`. 
What do you think?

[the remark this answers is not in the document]
lukecwik: A user can still create a restriction where startkey == endkey, have no tryClaim calls and be checkpointed which is why we need to support both cases.

iemejia:
(Sorry for adding this but cannot find a better moment ot ask). Curious question related to fractional splits (but not to this PR). Would testing trySplit for a RestrictionTracker would be enough to test that a SDF based IO that uses this kind of Restriction supports DWR correctly?

I am asking because in HBaseIO based on SDF it is one of the missing tests compared with the tests we had in the past for `splitAtFranction` https://github.com/apache/beam/blob/ec67a9374671ea9ae670fb0f3935ead2ebed7981/sdks/java/io/hbase/src/test/java/org/apache/beam/sdk/io/hbase/HBaseIOTest.java#L359

If it is not the case, any suggestions on how to test this?

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/ByteKeyRangeTracker.java
```diff
@@ -64,31 +64,56 @@ public ByteKeyRange currentRestriction() {
...
 
-    // If we haven't done any work, we should return the original range we were processing
-    // as the checkpoint.
-    if (lastAttemptedKey == null) {
-      ByteKeyRange rval = range;
-      // We update our current range to an interval that contains no elements.
-      range = NO_KEYS;
-      return SplitResult.of(range, rval);
+    ByteKey startKey = (lastAttemptedKey == null) ? range.getStartKey() : next(lastAttemptedKey);
+    ByteKey endKey = range.getEndKey();
```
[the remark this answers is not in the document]
boyuanzz: When there is a split, the `range` is updated first before constructing residual. Thus stashing `endKey` is needed unless we create another local var to stash residual.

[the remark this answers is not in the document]
lukecwik: I see. Thanks

lukecwik:
> (Sorry for adding this but cannot find a better moment ot ask). Curious question related to fractional splits (but not to this PR). Would testing trySplit for a RestrictionTracker would be enough to test that a SDF based IO that uses this kind of Restriction supports DWR correctly?
> 
> I am asking because in HBaseIO based on SDF it is one of the missing tests compared with the tests we had in the past for `splitAtFranction`
> 
> https://github.com/apache/beam/blob/ec67a9374671ea9ae670fb0f3935ead2ebed7981/sdks/java/io/hbase/src/test/java/org/apache/beam/sdk/io/hbase/HBaseIOTest.java#L359-L384
> 
> If it is not the case, any suggestions on how to test this?

Testing the restriction tracker is almost all that is needed for the most part for testing DWR.

The last bit would be to make sure that the DoFn doesn't assume its processing the initial restriction and interacts with the restriction tracker correctly. An example of a splittable DoFn that would work a lot of the time but be incorrect when splitting occurs would be something like:
```
@ProcessElement
public void process(RestrictionTracker<OffsetRange, Long> foo) {
  int start = foo.getCurrentRestriction().getFrom();
  int end = foo.getCurrentRestriction().getTo();
  from (int i = start; i < end; ++i) {
    ... produce output ...
  }
  tracker.tryClaim(end);
}
```
All the examples I know of stem from producing output before claiming. Also, moving the tryClaim before the loop would make the splittable DoFn correct but would prevent it from being able to effectively split.

sdks/java/core/src/test/java/org/apache/beam/sdk/transforms/splittabledofn/ByteKeyRangeTrackerTest.java
boyuanzz: It should be `EMPTY`. Thanks for catching this!

boyuanzz:
Run Java PreCommit

boyuanzz:
Run Java PreCommit

iemejia:
> Testing that output is only produced after the DoFn has done the claim call makes a lot of sense and ensuring that only the part that has been claimed was produced as output would cover correctness.

Thanks for the answer (again). It does not seem trivial to be able to test this from the outside of the DoFn without having to play with DoFnRunners and friends. Or is there something that would help to do this?

lukecwik:
> > Testing that output is only produced after the DoFn has done the claim call makes a lot of sense and ensuring that only the part that has been claimed was produced as output would cover correctness.
> 
> Thanks for the answer (again). It does not seem trivial to be able to test this from the outside of the DoFn without having to play with DoFnRunners and friends. Or is there something that would help to do this?

Users should rely on PAssert/TestStream for correctness testing. As a simpler test, if the user doesn't use StartBundleContext/FinishBundleContext/ProcessContext, it should be easy for the user to instantiate their DoFn and call the methods on it directly if needs be.

Runners that support dynamic work rebalancing can tell users that their SDFs aren't behaving well by seeing that the SDF is not splitting or making a lot of unbalanced splits.

PR #11456, opened 2020-04-17
[BEAM-7554] Add MillisInstant logical type to replace DATETIME 

TheNeuralBit:
This PR adds a `MillisInstant` logical type, and replaces all usages of the primitive `DATETIME` type with it.

`MillisInstant` mirrors the `NanosInstant` type. It uses `org.joda.time.Instant` rather than `java.time.Instant` for consistency with `DATETIME`, and it is backed by an `INT64` representing the number of milliseconds since the epoch.

The majority of the changes in this PR are relatively trivial, but there are a few significant (and some potentially controversial) ones:
- Updates the datetime getter/setter functions generated in `ByteBuddyUtils` to produce and consume instances of `Long` rather than `Instant`. This is necessary because these functions are expected to produce/consume the base type for logical types.
- Updates all joda time literals in schema inference tests so they specify a timezone.
- Throughout the SQL module I've replaced references to `getBaseValue` with `getValue`. Previously the only logical types used in SQL were `PassThroughLogicalType` instances, where the input type and base type were the same.

Changes that are not currently included in this PR but we might consider:
- Replace all references to `DateTime` (`addDateTimeField`, `addDateTimeValue`, ..) with `Instant`, and deprecate or remove the `DateTime` references.

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/Schema.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/Schema.java
@@ -420,7 +420,6 @@ public class Schema implements Serializable {
     FLOAT,
     DOUBLE,
     STRING, // String.
-    DATETIME, // Date and time.
     BOOLEAN, // Boolean.
     BYTES, // Byte array.
     ARRAY,
@@ -432,7 +431,6 @@ public class Schema implements Serializable {
     public static final Set<TypeName> NUMERIC_TYPES =
         ImmutableSet.of(BYTE, INT16, INT32, INT64, DECIMAL, FLOAT, DOUBLE);
     public static final Set<TypeName> STRING_TYPES = ImmutableSet.of(STRING);
-    public static final Set<TypeName> DATE_TYPES = ImmutableSet.of(DATETIME);
     public static final Set<TypeName> COLLECTION_TYPES = ImmutableSet.of(ARRAY, ITERABLE);
     public static final Set<TypeName> MAP_TYPES = ImmutableSet.of(MAP);
     public static final Set<TypeName> COMPOSITE_TYPES = ImmutableSet.of(ROW);
@@ -447,10 +445,6 @@ public class Schema implements Serializable {
 
     public boolean isStringType() {
       return STRING_TYPES.contains(this);
-    }
-
-    public boolean isDateType() {
-      return DATE_TYPES.contains(this);
     }
 
     public boolean isCollectionType() {
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/Schema.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/Schema.java
@@ -39,6 +39,7 @@ import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 import org.apache.beam.sdk.annotations.Experimental;
 import org.apache.beam.sdk.annotations.Experimental.Kind;
+import org.apache.beam.sdk.schemas.logicaltypes.MillisInstant;
 import org.apache.beam.sdk.values.Row;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Preconditions;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.BiMap;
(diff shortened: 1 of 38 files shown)


TheNeuralBit:
Run SQL PostCommit

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/Schema.java
```diff
@@ -419,7 +420,6 @@ public int hashCode() {
     FLOAT,
     DOUBLE,
     STRING, // String.
-    DATETIME, // Date and time.
```
reuvenlax, line 422: I'm a little worried about this. Empirically many users are using schemas. Maybe we should start off by leaving DATETIME around and remove it later in another PR?

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamAggregationRel.java
```diff
@@ -257,10 +257,8 @@ private Transform(
           // Combining over a single field, so extract just that field.
           combined =
               (combined == null)
-                  ? byFields.aggregateFieldBaseValue(
-                      inputs.get(0), combineFn, fieldAggregation.outputField)
-                  : combined.aggregateFieldBaseValue(
-                      inputs.get(0), combineFn, fieldAggregation.outputField);
+                  ? byFields.aggregateField(inputs.get(0), combineFn, fieldAggregation.outputField)
+                  : combined.aggregateField(inputs.get(0), combineFn, fieldAggregation.outputField);
```
reuvenlax, line 261: I'm not sure I agree with this change. I think it's important that SQL work over user logical types by interpreting it as the base value. The user writing the SQL statement usually understands the base type of their logical type, and can write the SQL statement appropriately. This will break that.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamCalcRel.java
```diff
@@ -416,36 +417,33 @@ private static Expression value(
 
       Expression value =
           Expressions.convert_(
-              Expressions.call(
-                  expression,
-                  "getBaseValue",
-                  Expressions.constant(index),
-                  Expressions.constant(convertTo)),
-              convertTo);
+              Expressions.call(expression, "getValue", Expressions.constant(index)), convertTo);
```
reuvenlax, line 420: I have the same concern with this change.

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/Schema.java
TheNeuralBit: Yeah we could deprecate it for a release or two before removing. Would we also need to keep support for the primitive DATETIME in IOs and SQL? I'd need to think about how that transition would work.

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamAggregationRel.java
TheNeuralBit: What about adding the option for SqlTransform to convert unknown logical types to their base type at the input? I think that behavior would be effectively the same

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/logicaltypes/MillisInstant.java
```diff
@@ -0,0 +1,44 @@
...
+ * limitations under the License.
+ */
+package org.apache.beam.sdk.schemas.logicaltypes;
+
+import org.joda.time.Instant;
+import org.joda.time.ReadableInstant;
+
+/** A timestamp represented as milliseconds since the epoch. */
+public class MillisInstant extends MillisType<ReadableInstant> {
+  public static final String IDENTIFIER = "beam:logical_type:millis_instant:v1";
```
TheNeuralBit, line 25: @reuvenlax what do you think about making this type (and maybe NanosInstant as well) parameterized by timezone?

The [arrow approach](https://github.com/apache/arrow/blob/master/format/Schema.fbs#L178) seems useful: un-specified timezone indicates time-zone naive (e.g. joda time Instant), otherwise time zone parameter should reference a value in tzdata (and would map to joda time DateTime).

cc: @alexvanboxel

reuvenlax: A timestamp type seems like it's by definition time-zone agnostic. If we want a datetime type, that should be a different type.

robinyqiu:
cc: @robinyqiu @ZijieSong946

PR #11517, opened 2020-04-24
[BEAM-9643] Adding Go SDF Documentation.

youngoli:
There are still a few small changes I'm thinking of making to the user API frontend before I merge this in, so I may hold off on submitting this until I make some changes and can rebase this. But still sending it out for review since the vast majority of it will be unchanged.

(I'm thinking of changing TrySplit to return two restrictions instead of one restriction tracker. I think both should theoretically work, but that approach is probably more intuitive. And I need to change GetProgress to return two floats, for finished and remaining work.)

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/go/examples/stringsplit/offsetrange/offsetrange.go
+++ b/sdks/go/examples/stringsplit/offsetrange/offsetrange.go
@@ -88,22 +88,25 @@ func (tracker *Tracker) GetError() error {
 	return tracker.Err
 }
 
-// TrySplit splits at the nearest integer greater than the given fraction of the remainder.
-func (tracker *Tracker) TrySplit(fraction float64) (interface{}, error) {
+// TrySplit splits at the nearest integer greater than the given fraction of the remainder. If the
+// fraction given is outside of the [0, 1] range, it is clamped to 0 or 1.
+func (tracker *Tracker) TrySplit(fraction float64) (primary, residual interface{}, err error) {
 	if tracker.Stopped || tracker.IsDone() {
-		return nil, nil
+		return tracker.Rest, nil, nil
 	}
-	if fraction < 0 || fraction > 1 {
-		return nil, errors.New("fraction must be in range [0, 1]")
+	if fraction < 0 {
+		fraction = 0
+	} else if fraction > 1 {
+		fraction = 1
 	}
 
-	splitPt := tracker.Rest.Start + int64(fraction*float64(tracker.Rest.End-tracker.Rest.Start))
-	if splitPt == tracker.Rest.End {
-		return nil, nil
+	splitPt := tracker.Claimed + int64(fraction*float64(tracker.Rest.End-tracker.Claimed))
+	if splitPt >= tracker.Rest.End {
+		return tracker.Rest, nil, nil
 	}
-	residual := Restriction{splitPt, tracker.Rest.End}
+	residual = Restriction{splitPt, tracker.Rest.End}
 	tracker.Rest.End = splitPt
-	return residual, nil
+	return tracker.Rest, residual, nil
 }
 
 // GetProgress reports progress based on the claimed size and unclaimed sizes of the restriction.
--- a/sdks/go/pkg/beam/core/sdf/sdf.go
+++ b/sdks/go/pkg/beam/core/sdf/sdf.go
@@ -13,42 +13,36 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package sdf is experimental, incomplete, and not yet meant for general usage.
+// Package sdf contains interfaces used specifically for splittable DoFns.
 package sdf
 
 // RTracker is an interface used to interact with restrictions while processing elements in
-// SplittableDoFns. Each implementation of RTracker is expected to be used for tracking a single
-// restriction type, which is the type that should be used to create the RTracker, and output by
-// TrySplit.
+// splittable DoFns (specifically, in the ProcessElement method). Each RTracker tracks the progress
+// of a single restriction.
 type RTracker interface {
-	// TryClaim attempts to claim the block of work in the current restriction located at a given
-	// position. This method must be used in the ProcessElement method of Splittable DoFns to claim
-	// work before performing it. If no work is claimed, the ProcessElement is not allowed to perform
-	// work or emit outputs. If the claim is successful, the DoFn must process the entire block. If
-	// the claim is unsuccessful the ProcessElement method of the DoFn must return without performing
+	// TryClaim attempts to claim the block of work located in the given position of the
+	// restriction. This method must be called in ProcessElement to claim work before it can be
+	// processed. Processing work without claiming it first can lead to incorrect output.
+	//
+	// If the claim is successful, the DoFn must process the entire block. If the claim is
+	// unsuccessful ProcessElement method of the DoFn must return without performing
 	// any additional work or emitting any outputs.
 	//
-	// TryClaim accepts an arbitrary value that can be interpreted as the position of a block, and
-	// returns a boolean indicating whether the claim succeeded.
-	//
-	// If the claim fails due to an error, that error can be retrieved with GetError.
-	//
-	// For SDFs to work properly, claims must always be monotonically increasing in reference to the
-	// restriction's start and end points, and every block of work in a restriction must be claimed.
+	// If the claim fails due to an error, that error is stored and will be automatically emitted
+	// when the RTracker is validated, or can be manually retrieved with GetError.
 	//
 	// This pseudocode example illustrates the typical usage of TryClaim:
 	//
-	// 	pos = position of first block after restriction.start
+	// 	pos = position of first block within the restriction
 	// 	for TryClaim(pos) == true {
 	// 		// Do all work in the claimed block and emit outputs.
-	// 		pos = position of next block
+	// 		pos = position of next block within the restriction
 	// 	}
 	// 	return
 	TryClaim(pos interface{}) (ok bool)
 
-	// GetError returns the error that made this RTracker stop executing, and it returns nil if no
-	// error occurred. If IsDone fails while validating this RTracker, this method will be
-	// called to log the error.
+	// GetError returns the error that made this RTracker stop executing, and returns nil if no
+	// error occurred. This is the error that is emitted if automated validation fails.
 	GetError() error
 
 	// TrySplit splits the current restriction into a primary and residual based on a fraction of the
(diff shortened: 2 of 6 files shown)


youngoli:
R: @lostluck

sdks/go/pkg/beam/core/sdf/sdf.go
```diff
@@ -13,42 +13,36 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package sdf is experimental, incomplete, and not yet meant for general usage.
+// Package sdf contains interfaces used specifically for splittable DoFns.
```
lostluck, line 16: Note, even if we submit this now-ish, we can still call it experimental until we have testing and some IO usage to bolster confidence.

sdks/go/pkg/beam/pardo.go
```diff
@@ -222,6 +222,81 @@ func ParDo0(s Scope, dofn interface{}, col PCollection, opts ...Option) {
...
+//
+// Dynamic splitting occurs during the processing of a restriction in runners
+// that have implemented it. If there are available workers, runners may split
+// the unprocessed portion of work from a busy worker and shard it to available
+// workers in order to better distribute work. With unsplittable DoFns this can
+// only occur on element boundaries, but for splittable DoFns this split
+// can land within a restriction and will require splitting that restriction.
+//
+// * Note: The Go SDK currently does not support dynamic splitting for SDFs,
+//   only initial splitting. Work can only be split at element boundaries.
```
lostluck, line 261: I think in this case you mean 
```
Work can only be split by the initial restriction boundaries.
```
Since technically the initial split could turn every element into three restrictions or something, but those restrictions are at the "sub element" boundaries rather than in every element (eg. Each initial element is a filename, and the Initial split stats each file's size, and just provides restrictions once every 1000 bytes), and handles actually "empty restrictions" during processing.

sdks/go/pkg/beam/pardo.go
```diff
@@ -222,6 +222,81 @@ func ParDo0(s Scope, dofn interface{}, col PCollection, opts ...Option) {
...
+//     containing only the original restriction. This method will always be
+//     called on each newly created restriction before they are processed.
+// * `RestrictionSize(elem, restriction) float64`
+//     RestrictionSize returns a cheap size estimation for a restriction. This
+//     size is an abstract scalar value that represents how much work a
+//     restriction takes compared to other restrictions in the same DoFn. For
+//     example, a size of 200 represents twice as much work as a size of
+//     100, but the numbers do not represent anything on their own. Size is
+//     used by runners to estimate work for liquid sharding.
+// * `CreateTracker(restriction) sdf.RTracker`
```
lostluck, line 289: Is this supposed to be a concrete type or the literal sdf.RTracker interface?

sdks/go/pkg/beam/pardo.go
youngoli: I was actually trying to say that during execution, work is only split along element boundaries. But even that is a bit misleading, because the elements in question are the initially split restrictions. So I'll probably either word it something like "This means that for runners that support liquid sharding, only the initially split restrictions can be distributed among workers, and stragglers cannot be split into smaller restrictions". Or maybe I'll keep it simple and just leave the first sentence.

sdks/go/pkg/beam/pardo.go
youngoli: Ah, good point. It's supposed to be a concrete type implementing the sdf.RTracker interface. I'll update this to mention that explicitly.

sdks/go/pkg/beam/core/sdf/sdf.go
youngoli: Done, I prefer to call it experimental at the moment.

sdks/go/pkg/beam/pardo.go
youngoli: Done.

sdks/go/pkg/beam/pardo.go
youngoli: Done.

sdks/go/examples/stringsplit/offsetrange/offsetrange.go
```diff
@@ -89,21 +89,23 @@ func (tracker *Tracker) GetError() error {
...
-func (tracker *Tracker) TrySplit(fraction float64) (interface{}, error) {
+func (tracker *Tracker) TrySplit(fraction float64) (interface{}, interface{}, error) {
 	if tracker.Stopped || tracker.IsDone() {
-		return nil, nil
+		return tracker.Rest, nil, nil
 	}
-	if fraction < 0 || fraction > 1 {
-		return nil, errors.New("fraction must be in range [0, 1]")
+	if fraction < 0 {
+		fraction = 0
```
lostluck, line 97: Might be worth documenting this behavior in the comment.

sdks/go/examples/stringsplit/offsetrange/offsetrange.go
```diff
@@ -89,21 +89,23 @@ func (tracker *Tracker) GetError() error {
 }
 
 // TrySplit splits at the nearest integer greater than the given fraction of the remainder.
-func (tracker *Tracker) TrySplit(fraction float64) (interface{}, error) {
+func (tracker *Tracker) TrySplit(fraction float64) (interface{}, interface{}, error) {
```
lostluck, line 92: Given this is the example, using named return values (primary, residual ...) is appropriate here for documentation purposes (but not so one can use an empty return.)

sdks/go/pkg/beam/core/sdf/sdf.go
```diff
@@ -13,63 +13,64 @@
...
-	// TrySplit splits the current restriction into a primary and residual based on a fraction of the
-	// work remaining. The split is performed along the first valid split point located after the
-	// given fraction of the remainder. This method is called by the SDK harness when receiving a
-	// split request by the runner.
+	// TrySplit splits the current restriction into a primary (currently executing work) and
+	// residual (work to be split off) based on a fraction of work remaining. The split is performed
+	// at the first valid split point located after the given fraction of remaining work.
+	//
+	// For example, a fraction of 0.5 means to split at the halfway point of remaining work only. If
+	// 50% of work is done and 50% remaining, then a fraction of 0.5 would split after 75% of work.
```
lostluck, line 56: +1 to this concrete example.

sdks/go/pkg/beam/pardo.go
```diff
@@ -222,6 +222,87 @@ func ParDo0(s Scope, dofn interface{}, col PCollection, opts ...Option) {
...
+// * `RestrictionSize(elem, restriction) float64`
+//     RestrictionSize returns a cheap size estimation for a restriction. This
+//     size is an abstract scalar value that represents how much work a
+//     restriction takes compared to other restrictions in the same DoFn. For
+//     example, a size of 200 represents twice as much work as a size of
+//     100, but the numbers do not represent anything on their own. Size is
+//     used by runners to estimate work for liquid sharding.
+// * `CreateTracker(restriction) restrictionTracker`
+//     CreateTracker creates and returns a restriction tracker (a concrete type
+//     implementing `sdf.RTracker`) given a restriction. The restriction tracker
```
lostluck, line 296: Consider being explicit about sdf.RTracker being an interface.
eg... implementing the `sdf.RTracker` interface.

sdks/go/pkg/beam/runners/universal/runnerlib/stage.go
```diff
@@ -43,7 +43,7 @@ func Stage(ctx context.Context, id, endpoint, binary, st string) (retrievalToken
 		return "", nil
 	}
 
-  return StageViaLegacyApi(ctx, cc, binary, st)
```
lostluck, line 46: This cleanup is  separate from the rest of the change.

sdks/go/pkg/beam/core/sdf/sdf.go
```diff
@@ -13,63 +13,64 @@
...
 	//
-	// 	pos = position of first block after restriction.start
+	// 	pos = position of first block within the restriction
 	// 	for TryClaim(pos) == true {
 	// 		// Do all work in the claimed block and emit outputs.
-	// 		pos = position of next block
+	// 		pos = position of next block within the restriction
 	// 	}
 	// 	return
 	TryClaim(pos interface{}) (ok bool)
```
lostluck, line 45: It would be worth explicitly saying that the position type is related directly to the type of Restriction being handled. 

Eg.
A linear offset restriction could use a single int64 value to represent a position. Similarly a multi dimensional restriction space could use a more complex type to represent the area claimed as a position.

sdks/go/pkg/beam/core/sdf/sdf.go
```diff
@@ -13,63 +13,64 @@
...
+	// split point is the end of the restriction), then the returned residual is nil.
 	//
-	// The current restriction is split into two by modifying the current restriction's endpoint to
-	// turn it into the primary, and returning a new restriction tracker representing the residual.
-	// If no valid split point exists, this method returns nil instead of a residual, but does not
-	// return an error. If this method is unable to split due to some error then it returns nil and
-	// an error.
-	TrySplit(fraction float64) (residual interface{}, err error)
+	// If an error is returned, some catastrophic failure occurred and the entire bundle will fail.
+	TrySplit(fraction float64) (primary, residual interface{}, err error)
```
lostluck, line 64: IIRC for dynamic splitting, this is the one that requires the Tracker to be concurrency safe? Do we want to declared ahead of time that implementations of RTracker must be threadsafe?

sdks/go/examples/stringsplit/offsetrange/offsetrange.go
youngoli: Done.

sdks/go/examples/stringsplit/offsetrange/offsetrange.go
youngoli: Good point, done.

sdks/go/pkg/beam/pardo.go
youngoli: Don't see why not. Done.

sdks/go/pkg/beam/runners/universal/runnerlib/stage.go
youngoli: Came bundled in with `go fmt` but I'll move it to a separate PR.

sdks/go/pkg/beam/core/sdf/sdf.go
youngoli: Good idea, done.

sdks/go/pkg/beam/core/sdf/sdf.go
youngoli: Hmm, good point. It's definitely something a use should know, especially for writing future-proof SDFs. It would need to be in all the methods I think, except *maybe* GetError. I'm not completely sure off the top of my head though, I feel like it'll depend on the details of how we do dynamic splitting.

For now I'll add a comment saying to make all methods thread-safe. Later we should add the thread-safe wrapper and recommend using that, and if we can dodge mutexes in some methods then we can implement it there.

lostluck: Technically since GetError is going to be reading from a location that might be concurrently modified, it's required to be thread safe as well.

PR #11533, opened 2020-04-27
[BEAM-9827] Ensure minimum watermark hold is computed across all keys

mxm:
For computing the current output watermark we need to compute the minimum
watermark hold across all watermark holds. We maintain an in-memory view of all
watermark holds because iterating over all keys is an expensive operation which
we only want to do when restoring from a checkpoint, not on every watermark.

The watermark view may not be in sync with its backing managed state. This
change corrects this by maintaining a priority queue with all watermark holds
and adding additional tests.

--- | --- | --- | --- | ---


mxm:
Run Flink ValidatesRunner

mxm:

mxm:
Run Nexmark Flink

mxm:
This fixes the flaky test in BEAM-9827.

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/state/FlinkStateInternals.java
```diff
@@ -76,8 +76,8 @@
   private final KeyedStateBackend<ByteBuffer> flinkStateBackend;
   private Coder<K> keyCoder;
 
-  // Combined watermark holds for all keys of this partition
-  private final Map<String, Instant> watermarkHolds = new HashMap<>();
+  // Watermark holds for all keys/windows of this partition
+  private final PriorityQueue<Long> watermarkHolds = new PriorityQueue<>();
```
dmvk, line 80: To sum up if I understand this correctly
- We only need to keep a single value (the minimum)
- We can not use scalar value that would hold the minimum, because we need a mutable wrapper

in this case can we only keep the minimum value in order to have O(log N) complexity? eg.:
```
void insertAndKeepMinimum(PriorityQueue<T> pq, T element) {
  pq.insert(element)
  while(pq.size() > 1) {
    pq.remove();
  }
}
```

`pq.remove(T ...)` has linear time complexity

dmvk: Hmm, I see it now, `FlinkWatermarkHoldState` is per window :/

mxm: Only keeping the minimum doesn't work. When the minimum hold is cleared, we will have to recompute the minimum by iterating over all the keys in the state backend which will defeat the purpose of having the cache.

dmvk: 👍 that makes sense

It would be nice if we could get rid of `pq.remove(...)` calls as these are O(n) and number of keys may be fairly large. How about using `TreeMap<Long, Integer>` instead, where value would be number of references to that particular offset?

This should have O(log n) characteristic and may hopefully deduplicate some of the entries.

dmvk:
👍 This change makes sense. I'm little bit unsure about memory requirements here, but hopefully storing an offset per key / window should be ok.

runners/flink/src/main/java/org/apache/beam/runners/flink/translation/wrappers/streaming/state/FlinkStateInternals.java
mxm: Good idea. I've changed it to use a TreeMap.

mxm:
Run Flink ValidatesRunner

mxm:

mxm:
Run Nexmark Flink

mxm:
Run Portable_Python PreCommit

mxm:
Thanks for the review @dmvk!

mxm:
CC @tweise

PR #11581, opened 2020-04-30
[BEAM-8307] NPE in Calcite dialect when input PCollection has logical…

rahul8383:
… type in schema, from JdbcIO Transform

When SqlTransform.query() PTransform is used with JdbcIO, where JdbcIO has logical types, an NPE is thrown in BeamSql when converting Beam Schema to Calcite RelDataType.

This PR adds standard logical types with URNs in `schemas.logicaltypes` package and adds Calcite's RelDataType Mapping for these logical types. 

`JdbcIO` is modified to use these standard logical types to convert JDBC Schema to Beam Schema.

Standard logical types for DATE, TIME, TIME_WITH_TIMEZONE, TIMESTAMP_WITH_TIMEZONE are yet to be handled. 

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/logicaltypes/FixedLengthString.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/logicaltypes/FixedLengthString.java
@@ -20,37 +20,36 @@ package org.apache.beam.sdk.schemas.logicaltypes;
 import static org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Preconditions.checkArgument;
 
 import org.apache.beam.sdk.annotations.Experimental;
-import org.apache.beam.sdk.annotations.Experimental.Kind;
 import org.apache.beam.sdk.schemas.Schema;
 
-/** A LogicalType representing a fixed-size byte array. */
-@Experimental(Kind.SCHEMAS)
-public class FixedBytes extends IdenticalBaseTAndInputTLogicalType<byte[]> {
-  public static final String IDENTIFIER = "beam:logical_type:fixed_length_bytes:v1";
-  private final int byteArraySize;
+/** A LogicalType representing a fixed-length string. */
+@Experimental(Experimental.Kind.SCHEMAS)
+public class FixedLengthString extends IdenticalBaseTAndInputTLogicalType<String> {
+  public static final String IDENTIFIER = "beam:logical_type:fixed_length_string:v1";
+  private final int length;
 
-  private FixedBytes(int byteArraySize) {
-    super(IDENTIFIER, Schema.FieldType.INT32, byteArraySize, Schema.FieldType.BYTES);
-    this.byteArraySize = byteArraySize;
-  }
-
-  public static FixedBytes of(int byteArraySize) {
-    return new FixedBytes(byteArraySize);
+  private FixedLengthString(int length) {
+    super(IDENTIFIER, Schema.FieldType.INT32, length, Schema.FieldType.STRING);
+    this.length = length;
   }
 
   public int getLength() {
-    return byteArraySize;
+    return length;
+  }
+
+  public static FixedLengthString of(int length) {
+    return new FixedLengthString(length);
   }
 
   @Override
-  public byte[] toBaseType(byte[] input) {
-    checkArgument(input == null || input.length == byteArraySize);
+  public String toBaseType(String input) {
+    checkArgument(input == null || input.length() == length);
     return input;
   }
 
   @Override
-  public byte[] toInputType(byte[] base) {
-    checkArgument(base == null || base.length == byteArraySize);
+  public String toInputType(String base) {
+    checkArgument(base == null || base.length() == length);
     return base;
   }
 }
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/logicaltypes/IdenticalBaseTAndInputTLogicalType.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/logicaltypes/IdenticalBaseTAndInputTLogicalType.java
@@ -1,0 +1,82 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.beam.sdk.schemas.logicaltypes;
+
+import java.util.Objects;
+import org.apache.beam.sdk.annotations.Experimental;
+import org.apache.beam.sdk.schemas.Schema;
+
+/** A base class for LogicalTypes that use the same input type as the underlying base type. */
+@Experimental(Experimental.Kind.SCHEMAS)
+public abstract class IdenticalBaseTAndInputTLogicalType<T> implements Schema.LogicalType<T, T> {
+  protected final String identifier;
+  protected final Schema.FieldType argumentType;
+  protected final Object argument;
+  protected final Schema.FieldType baseType;
+
+  protected IdenticalBaseTAndInputTLogicalType(
+      String identifier,
+      Schema.FieldType argumentType,
+      Object argument,
+      Schema.FieldType baseType) {
+    this.identifier = identifier;
+    this.argumentType = argumentType;
+    this.argument = argument;
+    this.baseType = baseType;
+  }
+
+  @Override
+  public String getIdentifier() {
+    return identifier;
+  }
+
+  @Override
+  public Schema.FieldType getArgumentType() {
+    return argumentType;
+  }
+
+  @Override
+  @SuppressWarnings("TypeParameterUnusedInFormals")
+  public <ArgumentT> ArgumentT getArgument() {
+    return (ArgumentT) argument;
+  }
+
+  @Override
+  public Schema.FieldType getBaseType() {
+    return baseType;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    IdenticalBaseTAndInputTLogicalType<?> that = (IdenticalBaseTAndInputTLogicalType<?>) o;
+    return identifier.equals(that.identifier)
+        && argument.equals(that.argument)
+        && baseType.equals(that.baseType);
+  }
+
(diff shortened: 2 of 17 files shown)


rahul8383:
R: @reuvenlax @amaliujia @iemejia

aaltay:
retest this please

reuvenlax:
I'm wondering if this is a scalable solution. 

In general SQL is supposed to be able to handle unknown logical types, and simply treat them as the base type. If you'r seeing a NPE, maybe we need to fix that. Where do you see the NPE?

rahul8383:
@reuvenlax 
I have provided the source table schema and attached the NPE that I have faced in [BEAM-8307](https://issues.apache.org/jira/browse/BEAM-8307)

NPE is thrown as Calcite's RelDataType cannot be found for the JdbcIO Logical Type.

rahul8383:
I agree that this is not a scalable solution. Providing a Calcite RelDataType Mapping for every Logical Type defined(which is the solution presented in this PR) by every IO is not scalable.

Another approach to solving the problem is: 
Provide Calcite RelDataType Mapping depending on the Base Type defined in the logical type. 
But, as CHAR and VARCHAR logical types have Base Type as STRING, we have to choose a default Calcite RelDataType Mapping. The same thing applies to BINARY and VARBINARY logical types.
We also have to choose a default RelDataType Mapping if the Base Type is DATETIME.
By using this approach, I think we might be missing some built-in Aggregation functions provided by Calcite for specific types.

I was also thinking if we can use the IDENTIFIER of the logical type to determine the corresponding Calcite RelDataType. But, as the IDENTIFIER type is String and not an enum, it cannot be used. For example, all the logical types defined by JdbcIO use java.sql.JDBCType name as the IDENTIFIER.

Please correct me if my understanding is incorrect.

reuvenlax:
Where do you see the NPE?

rahul8383:
JdbcIO.Read -> SqlTransform.query(SELECT COUNT(*) FROM PCOLLECTION //Any query ) would throw NPE if the input PCollection to SqlTransform JdbcIO specific Logical Types in its Schema.

I have provided the Source Table Schema and attached the Exception stack trace in the JIRA ticket.

rahul8383:
I will make the necessary changes as suggested by @TheNeuralBit in https://lists.apache.org/thread.html/r281e2913379c9733f6ac5baa08f361cc4ebe880a9880b2d54d6095b0%40%3Cdev.beam.apache.org%3E

amaliujia:
retest this please

rahul8383:
R: @TheNeuralBit 

I found a bug while implementing this feature and raised PR #11609 to fix the bug as I thought that the bug fix could be cherry-picked in 2.21.0 release.

iemejia:
retest this please

iemejia:
retest this please

rahul8383:
The failing test `org.apache.beam.runners.fnexecution.control.RemoteExecutionTest.testSplit` should be unrelated to this change.

TheNeuralBit:
Run SQL PostCommit

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/logicaltypes/LogicalDecimal.java
```diff
@@ -0,0 +1,69 @@
...
+      Schema.builder().addInt32Field("precision").addInt32Field("scale").build();
+  private final int precision;
+  private final int scale;
+
+  private LogicalDecimal(int precision, int scale) {
+    super(
+        IDENTIFIER,
+        Schema.FieldType.row(schema),
+        Row.withSchema(schema).addValues(precision, scale).build(),
+        Schema.FieldType.STRING);
```
TheNeuralBit, line 41: I'm a little confused about how this works. This argument indicates the base type is String, but this is an `IdenticalBaseTAndInputTLogicalType`, and `toBaseType` just returns a `BigDecimal`, which shouldn't be compatible.

It's possible this is a bug, it's just not turning up in any tests because we don't have a schema with LogicalDecimal being used in a coder. Could you add these new logical types to a test that exercises RowCoder? There are a variety of them: `SchemaCoderTest`, `RowCoderTest`, and `CloudObjectsTest` (a Dataflow specific test). Probably `RowCoderTest` would be the best place to test these types.

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/logicaltypes/FixedBytes.java
```diff
@@ -43,44 +42,15 @@ public int getLength() {
...
 
   @Override
   public byte[] toInputType(byte[] base) {
-    checkArgument(base.length <= byteArraySize);
-    if (base.length == byteArraySize) {
-      return base;
-    } else {
-      return Arrays.copyOf(base, byteArraySize);
-    }
-  }
```
TheNeuralBit, line 80: FYI @reuvenlax this PR is removing the ability to convert smaller byte arrays. As noted in https://github.com/apache/beam/pull/11609#issuecomment-624810068 it seems this logic is inaccessible anyway.

TheNeuralBit:
Run SQL PostCommit

rahul8383:
@apilloud @amaliujia @robinyqiu   
`JdbcIO` was using logical types which were specifically defined in JdbcIO package. When `SqlTransform.query()` was used with JdbcIO, an NPE is thrown as the Calcite's RelDataType Mapping could not be found for the logical types.
As all of these are standard logical types, this PR moved these logical types from JdbcIO package to `schema.logicaltypes` package so that other IOs which need these logical types can make use of it. 
If BeamSql is supported for these standard logical types, BeamSql can be used with all the IOs that use these types.
 
To close this bug, some functional tests need to be added which uses the standard logical types defined in `schema.logicaltypes` package. Can you please provide some pointers as to what type of queries can be used to test these types?
Also `CalciteUtils.CALCITE_TO_BEAM_TYPE_MAPPING` doesn't specify these logical types. Does this mean the conversion of types is lossy? ex: will we lose the information about the maximum length of the String if `VariableLengthString` logical type is used in BeamSql?

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/logicaltypes/LogicalDecimal.java
rahul8383: Ah! This is a blunder. 
Sure, Brian. I will modify/add tests.

apilloud:
Beam SQL should be rejecting unknown logical types. We expect users to put their data into supported logical types before passing it in. This conversion is up to the user to implement and will likely be lossy. It would be significant work to make it do something different.

TheNeuralBit:
> Beam SQL should be rejecting unknown logical types.

The types that @rahul8383 is making into standard logical types here are well known SQL types. I don't think Beam SQL should consider them unknown.

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/logicaltypes/LogicalDecimal.java
rahul8383: @TheNeuralBit Modified `RowCoderTest.testLogicalType()` which exercises coder for the standard logical types.

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/logicaltypes/IdenticalBaseTAndInputTLogicalType.java
```diff
@@ -0,0 +1,82 @@
...
+ */
+package org.apache.beam.sdk.schemas.logicaltypes;
+
+import java.util.Objects;
+import org.apache.beam.sdk.annotations.Experimental;
+import org.apache.beam.sdk.schemas.Schema;
+
+/** A base class for LogicalTypes that use the same input type as the underlying base type. */
+@Experimental(Experimental.Kind.SCHEMAS)
+public abstract class IdenticalBaseTAndInputTLogicalType<T> implements Schema.LogicalType<T, T> {
```
TheNeuralBit, line 26: Can you make this package-private?

sdks/java/extensions/sql/src/test/java/org/apache/beam/sdk/extensions/sql/impl/utils/CalciteUtilsTest.java
```diff
@@ -59,68 +64,121 @@ public void setUp() {
...
+            .addField("date", CalciteUtils.DATE)
+            .addField("time", CalciteUtils.TIME)
+            .addField("timestamp", CalciteUtils.TIMESTAMP)
+            .addField("timeWithLocalTZ", CalciteUtils.TIME_WITH_LOCAL_TZ)
+            .addField("timestampWithLocalTZ", CalciteUtils.TIMESTAMP_WITH_LOCAL_TZ)
+            .addField("fixedBytes", Schema.FieldType.logicalType(FixedBytes.of(10)))
+            .addField("variableBytes", Schema.FieldType.logicalType(VariableLengthBytes.of(100)))
+            .addField("fixedString", Schema.FieldType.logicalType(FixedLengthString.of(10)))
+            .addField("variableString", Schema.FieldType.logicalType(VariableLengthString.of(100)))
+            .addField("customDecimal", Schema.FieldType.logicalType(LogicalDecimal.of(10, 5)))
```
TheNeuralBit, line 87: Could you add tests that use some or all of these new types in a SQL query? Right now I think this is the only test that exercises the new code in CalciteUtils, and it's not actually verifying the types will work in a SQL statement.

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/logicaltypes/FixedLengthString.java
```diff
@@ -0,0 +1,55 @@
...
+    checkArgument(input == null || input.length() == length);
+    return input;
+  }
+
+  @Override
+  public String toInputType(String base) {
+    checkArgument(base == null || base.length() == length);
+    return base;
+  }
+}
```
TheNeuralBit, line 55: Could you add constants for these new logical types in [SqlTypes](https://github.com/apache/beam/blob/master/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/logicaltypes/SqlTypes.java)?

cc: @robinyqiu

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/logicaltypes/IdenticalBaseTAndInputTLogicalType.java
TheNeuralBit, line 26: Could you make this package-private? I think its only used in `schemas.logicaltypes` right now

sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/logicaltypes/LogicalDecimal.java
```diff
@@ -0,0 +1,69 @@
...
+import static org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Preconditions.checkArgument;
+
+import java.math.BigDecimal;
+import org.apache.beam.sdk.annotations.Experimental;
+import org.apache.beam.sdk.schemas.Schema;
+import org.apache.beam.sdk.values.Row;
+
+/** A LogicalType representing a Decimal type with custom precision and scale. */
+@Experimental(Experimental.Kind.SCHEMAS)
+public class LogicalDecimal extends IdenticalBaseTAndInputTLogicalType<BigDecimal> {
```
TheNeuralBit, line 29: Can we just call this `Decimal`?

PR #11538, opened 2020-04-27
[BEAM-9831] Improve UX for HL7v2IO

jaketf:
This implements the following improvements for HL7v2IO based on initial testing with customer
1. emit early panes rather than blocking on pagination in ListHL7v2Messages. Note, this is because the architecture of this PTransform is such that it fans out from a PCollection of HL7v2 stores (typically just 1) to many, many messages. By default, beam will not emit panes until it estimates all the data has arrived. Because one element is fanning out to (potentially) millions of messages that must be paginated through. Beam only knows "I'm not done outputting all the values for this single element, so my output must be incomplete". Unfortunately there is no splitable restriction due to the pagination model not knowing the page tokens ahead of time, this must  happen in a single thread. This change just allows us to start processed the early pages sooner, allowing for a smoother throughput curve of output elements rather than a massive step function when pagination is finally complete.
1. allow end users to use HL7v2MessageCoder in their pipelines by making constructor public.
1. drop output only fields (keeping only data and labels) before calling ingestMessages in HL7v2IO.Write rather than expecting the user to do this. (e.g. in a read from one HL7v2 store write directly to another use case)

R: @pabloem 
CC: @lastomato
------------------------

Thank you for your contribution! Follow this checklist to help us incorporate your contribution quickly and easily:

 - [x] [**Choose reviewer(s)**](https://beam.apache.org/contribute/#make-your-change) and mention them in a comment (`R: @username`).
 - [x] Format the pull request title like `[BEAM-XXX] Fixes bug in ApproximateQuantiles`, where you replace `BEAM-XXX` with the appropriate JIRA issue, if applicable. This will automatically link the pull request to the issue.
 - [x] Update `CHANGES.md` with noteworthy changes.
 - [x] If this contribution is large, please file an Apache [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

See the [Contributor Guide](https://beam.apache.org/contribute) for more tips on [how to make review process smoother](https://beam.apache.org/contribute/#make-reviewers-job-easier).

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

Lang | SDK | Apex | Dataflow | Flink | Gearpump | Samza | Spark
--- | --- | --- | --- | --- | --- | --- | ---
[...]

PR #11596, opened 2020-05-02
[BEAM-9856] Optimization/hl7v2 io list messages

jaketf:
This PR is an experiment to illustrate a potential strategy for implementing HL7v2IO.ListMessages as a splittable DoFn that carves the sendTime dimension up into time ranges and uses the Messages.List API with filter on sendTime and orderBy sendTime (to make lists monotonically increasing).

Currently this initial split is daily and the hope is to leave dynamic splits are left up to the existing implementaion in [OffsetRangeTracker](https://beam.apache.org/releases/javadoc/2.20.0/org/apache/beam/sdk/transforms/splittabledofn/OffsetRangeTracker.html)

## DISCLAIMER
This PR is currently a discussion piece to help drive the decision of the future of HL7v2IO.ListMessages. 

It has excessive logging for demonstrating behavior / debugging and is currently broken.

CC: @pabloem, @lukecwik 

## Questions
- What are the prototypical tests for validating splitable DoFn's various methods (e.g. SplitRestriction) stability?
- Are there best practices in optimizing inital split? In this case it's impossible to know a head of time how many messages.
- Is there a max recommended initial splitting? For example 100s or 1000s of splits preferrable? For IO connectors I imagine it depends what the QPS limits are of your source/target system are.
- Should I add adaptive throttling similar to DatastoreIO as we are introducing much more parallelism? Are there other methods in splitable dofns for rate limiting requests or indicating that runner can split until you have n currently active splits?

## Current Status: Working
Stole a lot from OffsetRangetTracker to implement OrderedTimeRageTracker a restriction tracker that assumes that local results sets are ordered by timestamp.

--- | --- | --- | --- | ---


sdks/java/io/google-cloud-platform/src/test/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IOTestUtil.java
```diff
@@ -59,7 +59,7 @@
               + "AL1|2|allergy|Z91.013^Personal history of allergy to sea food^ZAL|SEVERE|Swollen face|\r"
               + "AL1|3|allergy|Z91.040^Latex allergy^ZAL|MODERATE|Raised, itchy, red rash|",
           // Another ADT Message
-          "MSH|^~\\&|hl7Integration|hl7Integration|||||ADT^A08|||2.5|\r"
+          "MSH|^~\\&|hl7Integration|hl7Integration|||20190309132544||ADT^A08|||2.5|\r"
```
jaketf, line 62: this time stamp will always be present in valid HL7v2 Messages.

jaketf:
The NPE is caused by `OffsetRangeTracker.lastAttemptedOffset` being unboxed in `OffsetRangeTracker::checkDone` [here](https://github.com/apache/beam/blob/0291976d6029b4cf4d72caa03bd3836900fb6328/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/OffsetRangeTracker.java#L98). 

[all](https://github.com/apache/beam/blob/0291976d6029b4cf4d72caa03bd3836900fb6328/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/OffsetRangeTracker.java#L53) [other](https://github.com/apache/beam/blob/0291976d6029b4cf4d72caa03bd3836900fb6328/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/OffsetRangeTracker.java#L77) [uses](https://github.com/apache/beam/blob/0291976d6029b4cf4d72caa03bd3836900fb6328/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/OffsetRangeTracker.java#L119) check that `lastAttemptedOffset`  is not null.

I'm not sure if this was intentional in the implementation of OffsetRangeTracker.

pabloem:
@chamikaramj can you take a look at this?

jaketf:
> The NPE is caused by `OffsetRangeTracker.lastAttemptedOffset` being unboxed in `OffsetRangeTracker::checkDone` [here](https://github.com/apache/beam/blob/0291976d6029b4cf4d72caa03bd3836900fb6328/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/OffsetRangeTracker.java#L98).
> 
> [all](https://github.com/apache/beam/blob/0291976d6029b4cf4d72caa03bd3836900fb6328/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/OffsetRangeTracker.java#L53) [other](https://github.com/apache/beam/blob/0291976d6029b4cf4d72caa03bd3836900fb6328/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/OffsetRangeTracker.java#L77) [uses](https://github.com/apache/beam/blob/0291976d6029b4cf4d72caa03bd3836900fb6328/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/OffsetRangeTracker.java#L119) check that `lastAttemptedOffset` is not null.
> 
> I'm not sure if this was intentional in the implementation of OffsetRangeTracker.

@chamikaramj pablo said you might know about this.

should check done have some conditional on `lastAttemptedOffset != null`
e.g.
```java
  @Override
  public void checkDone() throws IllegalStateException {
    if (range.getFrom() == range.getTo()) {
      return;
    }

    if (lastAttemptedOffset != null) {
      checkState(
          lastAttemptedOffset >= range.getTo() - 1,
          "Last attempted offset was %s in range %s, claiming work in [%s, %s) was not attempted",
          lastAttemptedOffset,
          range,
          lastAttemptedOffset + 1,
          range.getTo());
    }
  }
```

I'm not really familiar with what checkDone should do in the case that lastAttemptedOffset was null.

chamikaramj:
Is using SplittableDoFn API here intentional ?

I think this API is being updated. cc: @lukecwik 

If you need dynamic work rebalancing, consider using the BoundedSource interface. Otherwise we can just implement the source as a regular and wait for SplittableDoFn API to stabilize before adding support for dynamic work rebalancing.

jaketf:
@chamikaramj thanks for the suggestion. I will look into using BoundedSource API.

Unfortunately, regular DoFns don't cut it because a single elements outputs are committed atomically (see this [conversation](https://github.com/apache/beam/pull/11538#discussion_r416927740)).
Basically we have one input element (HL7v2 store) exploding to many, many output elements (all the messages in that store) in a single ProcessElement call. I'm trying to explore strategies for splitting up this listing.

I originally chose splittable DoFn over BoundedSource based off the sentiment of this statement:
> **Coding against the Source API involves a lot of boilerplate and is error-prone**, and it does not compose well with the rest of the Beam model because a Source can appear only at the root of a pipeline. - https://beam.apache.org/blog/2017/08/16/splittable-do-fn.html

The blog also mentions 
- A Source can not emit an additional output (for example, records that failed to parse).
    - Healthcare customers feeding requirements for this plugin want DLQ on all sinks and sources. To be consistent with the streaming API provided in `HL7v2IO.Read` I wanted to provide DLQ in `HLv2IO.ListMessages`. However, I believe this is more of a nice to have for batch use cases.

jaketf:
@chamikaramj @lukecwik 
Apologies, I believe the the NPE was a user error on my part. 
I've been able to revert my changes to to OffsetRangeTracker without reintroducing the NPE.

To help future users as foolish as me,  to get a less confusing NPE, I suggest we do something like I put in 8891292. But this is definitely not a core issue in OffsetRangeTracker.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
```diff
@@ -472,24 +548,120 @@ public void initClient() throws IOException {
...
+
+    @NewTracker
+    public OrderedTimeRangeTracker newTracker(@Restriction OrderedTimeRange timeRange) {
+      return timeRange.newTracker();
+    }
+
+    @SplitRestriction
+    public void split(
+        @Restriction OrderedTimeRange timeRange, OutputReceiver<OrderedTimeRange> out) {
+      // TODO(jaketf) How to pick optimal values for desiredNumOffsetsPerSplit ?
```
lukecwik, line 569: Typically, doing one split for every 64mbs of output has been our guidance here in the past.

Dynamic splitting is meant to fill in the gap if there is too little splitting or a specific restriction has a lot more data then other restrictions.

jaketf: Unfortunately, in this use case dynamic splitting would be crucial because we can't know the distribution of data in the restriction dimension (sendTime). 

If you imagine a hospital might be much busier during daytime / weekdays than night times weekends (though never dormant due to ICU and emergency services). "Day time" might change base on hospital location, week days are subject to holidays, etc.

Data distribution in sendTime may be subject to significant spikes if one of the upstream systems populating sendTime has to backfill after a maintenance period and doesn't responsibly set this field to event time but sets all of the sendTimes to a short range of  backfill 
processing time (this is sub optimal behavior of that system but sometimes a reality).

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/OrderedTimeRange.java
```diff
@@ -0,0 +1,155 @@
...
+import org.apache.beam.sdk.coders.CoderException;
+import org.apache.beam.sdk.coders.InstantCoder;
+import org.apache.beam.sdk.util.VarInt;
+import org.apache.beam.sdk.values.TypeDescriptor;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Objects;
+import org.joda.time.Duration;
+import org.joda.time.Instant;
+
+/** A restriction represented by a range of Instants [from, to). */
+public class OrderedTimeRange
```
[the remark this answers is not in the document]
jaketf: TBH, the main value add here is more readable restriction bounds / values in the error messages for the context of the time range use case. 

If there is a cleaner way of borrowing all implementation from OffsetRangeTracker but changing format of these error messages, I'd be all ears.

I suppose I could add a  `private final bool offestsAreInstants` member to OffsetRangeTracker and throw conditional formatting on each error message?

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
lukecwik: Yeah, you'll have to choose a reasonable value based upon your knowledge within this domain and I understand that this may not work in practice.

In the case where people send a lot of data for a narrow time range, you would have to support filtering based upon other properties such as messageType or sendFacility and could break up the space based upon prefixes such as `~a`, `~b`, `NOT ~a AND NOT ~b` but this is dependent on this being efficient in the service as well.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/OrderedTimeRange.java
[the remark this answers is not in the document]
lukecwik: Any bug in this space is an error in the implementation of the IO connector so these error messages will mostly come up during development and then if any bugs are discovered during the usage the user won't be able to do much and this will go to the Beam community/IO author to fix.

If you really want to, adding a `format` method and creating a subclass would make more sense then having effectively a copy.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/OffsetRangeTracker.java
```diff
@@ -94,6 +94,9 @@ public void checkDone() throws IllegalStateException {
     if (range.getFrom() == range.getTo()) {
       return;
     }
+    if (lastAttemptedOffset == null) {
+      throw new IllegalStateException("lastAttemptedOffset should not be null");
+    }
     checkState(
         lastAttemptedOffset >= range.getTo() - 1,
         "Last attempted offset was %s in range %s, claiming work in [%s, %s) was not attempted",
```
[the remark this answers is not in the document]
jaketf: Sure, I can move this to separate PR.

Kindly, disagree on suggestion as it will just cause a different NPE on [L102](https://github.com/apache/beam/pull/11596/files/d2a094d5a133132d015fc7ed335e5b430a19f183#diff-b07e9d49c549836ef7e2d7006fa93f3cL102) when you call `lastAttemptedOffset + 1` to try and print this error message.

I think the lastAttemptedOffset null check should be separate and throw a more specific error message before we get to this stateCheck.

This failure mode is almost definitely mis-use of OffsetRangeTracker that would even cause this, and would be difficult to say what work was or wasn't attempted.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/OrderedTimeRange.java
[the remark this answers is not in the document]
jaketf: that make a lot of sense, thanks for the suggestion.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/OffsetRangeTracker.java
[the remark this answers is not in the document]
lukecwik: Good point but we should structure message saying that we haven't claimed anything in the range which is non-empty.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
jaketf: Yeah I think the "spiky backfill" (many cases in a small sendTime) is a corner case of a hot split that would just be slow and users would have to accept that or take it up with their upstream system.

splitting on messageType / sendFacility are probably more popular logical filters and feels like a hack for a corner case that might mess with performance under the "typical" distribution of data in sendTime.

jaketf: My intuition says hourly splits seems like a more reasonable starting point than day but I'm unsure about this without testing on a few datasets.

However for super large time ranges 10 years = 87,600 splits. naturally this explodes if we think about even smaller time ranges.

For example, I assume the initial collection of splits returned in SplitRestriction has to fit in memory? 
Or even just the loop to define all these might take a significant amount of time

lukecwik: That seems like a lot.

Dataflow has an API limit of 20mbs for split descriptions when being returned which usually tops out around 10k splits for sources but even 10k is too much. Typically 20-50 splits is enough since dynamic splitting will ramp that up to 1000s if necessary.

jaketf: Interesting, thanks for those details! 

But I thought Dataflow doesn't really do dynamic splitting yet so the initial split is all that matters? Maybe I'm confusing dynamic rebalancing != dynamic splitting?
  
Perhaps daily is the appropriate happy medium for now? 
1000s not 10,000s of initial splits.

In the future when dynamic splitting is supported bump initial split default to weeks?

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/OffsetRangeTracker.java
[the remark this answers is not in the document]
jaketf: opened https://github.com/apache/beam/pull/11648

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
lukecwik: dynamic splitting is a subset of dynamic rebalancing since dynamic rebalancing also covers other aspects such as splitting a GroupByKey, scaling up/down workers, ...

dynamic splitting is supported by batch Dataflow for bounded sinks using runner v1 and splittable dofns when running a batch pipeline when using runner v2

jaketf: Ok I think that dynamic splitting is "nice to have" for this connector and can wait for v2 connector.

Just adding initial splitting w/ SDF and v1 runner is a huge improvement over the existing implementation.

jaketf:
@pabloem PTAL / retest this if you have the chance.

jaketf:
@pabloem bump ^. please retest this.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
```diff
@@ -431,25 +455,70 @@ private Message fetchMessage(HealthcareApiClient client, String msgId)
...
+     * @param filter the filter
+     * @param initialSplitDuration the initial split duration for sendTime dimension splits
+     */
+    ListHL7v2Messages(
+        ValueProvider<List<String>> hl7v2Stores,
+        ValueProvider<String> filter,
+        Duration initialSplitDuration) {
+      this.hl7v2Stores = hl7v2Stores.get();
+      this.filter = filter.get();
+      this.initialSplitDuration = initialSplitDuration;
```
pabloem, line 471: `ValueProvider` arguments are usually not known at pipeline construction time, so it is not useful to call `get` on them when you create the PCollection. The usual method is to have ValueProvider attributes for the class - when you get String arguments, you would wrap them with `StaticValueProvider` - and in the execution-time methods (e.g. processelement, finish/startbundle), you would call `get` on the valueproviders. I seem to have missed this point earlier. sorry about that.

pabloem:
retest this please

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
```diff
@@ -472,24 +547,118 @@ public void initClient() throws IOException {
...
+                  i, lastClaimedMilliSecond));
+          if (hangingClaim || tracker.tryClaim(lastClaimedMilliSecond)) {
+            // This means we have claimed an entire millisecond we need to make sure that we
+            // process all messages for this millisecond because sendTime is allegedly nano second
+            // resolution.
+            // https://cloud.google.com/healthcare/docs/reference/rest/v1beta1/projects.locations.datasets.hl7V2Stores.messages#Message
+            while (cursor.getMillis() == lastClaimedMilliSecond
+                && i < page.size()) { // loop over messages in millisecond.
+              outputReceiver.output(msg);
+              msg = page.get(i++);
```
pabloem, line 632: Would we output the first message twice?
in the first iteration, `msg = page.get(0)`, from line 616
in the second iteration, `msg = page.get(i++)` - which is still `0`, and then `i` is incremented?
You could use `++i`, but then you would have to adjust the condition in the while loop, right?

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
```diff
@@ -472,24 +547,118 @@ public void initClient() throws IOException {
...
+        int i = 0;
+        HL7v2Message msg = page.get(i);
+        while (i < page.size()) { // loop over messages in page
+          cursor = Instant.parse(msg.getSendTime());
+          lastClaimedMilliSecond = cursor.getMillis();
+          LOG.info(
+              String.format(
+                  "initial claim for page %s lastClaimedMilliSecond = %s",
+                  i, lastClaimedMilliSecond));
+          if (hangingClaim || tracker.tryClaim(lastClaimedMilliSecond)) {
```
pabloem, line 624: I don't know the API requirements, but you would be calling `tryClaim` twice on a milliSecond? Once in line 645, and once more here?

pabloem:
Run Java PreCommit

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
jaketf: tl;dr No, we don't output the element twice. This is intended behavior.
 
`page` is  `List<HL7v2Message>`
When we call `page.get` we just call [List::get](https://docs.oracle.com/javase/8/docs/api/java/util/List.html#get-int-).
This is safe to call twice and will "get" the first element of the list both times.


We only emit results in L631 where we call [OutputReceiver::output](https://beam.apache.org/releases/javadoc/2.20.0/index.html?org/apache/beam/sdk/transforms/DoFn.OutputReceiver.html)

pabloem: but won't we go right back to 631, and emit the message that we got from 632 after calling `get(0)` for the second time?

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
jaketf: tl;dr No that is not possible based on 629-640
`tryClaim` on 624 claims `lastClaimedMillisSecond`

`tryClaim` on 645 claims the millisecond of the cursor (which is advanced based on the while loop in L629-633) 

if the cusor is not advanced to a new millisecond (this happens when many messages at the end of a page came in the same millisecond) the if block in L636-640 contains `continue` which will eagerly exit this iteration of the while loop before getting to the try claim on L645 again. I called this scenario "hangingClaim" because we  cannot know if the first message(s) of the next page will also be from this millisecond.

pabloem: sgtm

pabloem:
retest this please

pabloem:
retest this please

pabloem:
Run Java PostCommit

pabloem:
Run Java PostCommit

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
```diff
@@ -472,24 +551,118 @@ public void initClient() throws IOException {
...
+            // tracker.
+            tracker.tryClaim(cursor.getMillis());
+            lastClaimedMilliSecond = cursor.getMillis();
+            LOG.info(
+                String.format(
+                    "After claiming between messages lastClaimedMilliSecond = %s",
+                    lastClaimedMilliSecond));
+          }
+        }
         messageListingLatencyMs.update(Instant.now().getMillis() - reqestTime);
```
lukecwik, line 657: It would sense to move this into wherever we do the list call.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
lukecwik: I agree with @pabloem that this logic is not simple to follow and I think you could really simplify your code if you used https://guava.dev/releases/21.0/api/docs/com/google/common/collect/FluentIterable.html#concat-java.lang.Iterable-
since it would convert `Iterable<List<HL7v2Message>>` into `Iterable<HL7v2Message>` and only accesses the elements lazily so it wouldn't prefetch everything.

This would allow you to not worry that `messages` are in a `page` and your processing multiple `pages`.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
```diff
@@ -472,24 +551,118 @@ public void initClient() throws IOException {
...
+                  "initial claim for page %s lastClaimedMilliSecond = %s",
+                  i, lastClaimedMilliSecond));
+          if (hangingClaim || tracker.tryClaim(lastClaimedMilliSecond)) {
+            // This means we have claimed an entire millisecond we need to make sure that we
+            // process all messages for this millisecond because sendTime is allegedly nano second
+            // resolution.
+            // https://cloud.google.com/healthcare/docs/reference/rest/v1beta1/projects.locations.datasets.hl7V2Stores.messages#Message
+            while (cursor.getMillis() == lastClaimedMilliSecond
+                && i < page.size()) { // loop over messages in millisecond.
+              outputReceiver.output(msg);
```
lukecwik, line 635: Should we be outputting elements with the timestamp of the message and should we be reporting a watermark?

Even though you have a bounded SDF, it could be useful to report the watermark incase it is used in a streaming pipeline or users wanted to assign windows and perform grouping per window.

The current logic will assign the input's timestamp to all outputs which won't allow users to use windowing to effectively window the elements being output without assigning timestamps themselves. If we do want to go down this path it is simple right now because this transform always starts with PBegin but what would you want to do it the timestamp of the record is before the timestamp of the input element to the SDF (since it is illegal to output messages with timestamps before the input elements timestamp)?

To add watermark tracking based on timestamp of elements output, you would need to add the implementation for `@GetInitialWatermarkEstimatorState` and `@NewWatermarkEstimator` as seen in https://github.com/apache/beam/blob/27656d74fa9fb7085e2532275c821c3601c3f4b7/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/Watch.java#L763

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
```diff
@@ -415,10 +423,29 @@ private Message fetchMessage(HealthcareApiClient client, String msgId)
...
   }
 
-  /** List HL7v2 messages in HL7v2 Stores with optional filter. */
+  /**
+   * List HL7v2 messages in HL7v2 Stores with optional filter.
+   *
+   * <p>This transform is optimized for dynamic splitting of message.list calls for large batches of
+   * historical data and assumes rather continuous stream of sendTimes. It will dynamically
+   * rebalance resources to handle "peak traffic times" but will waste resources if there are large
+   * durations (days) of the sendTime dimension without data.
```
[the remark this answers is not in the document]
jaketf: correct.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
```diff
@@ -415,10 +423,29 @@ private Message fetchMessage(HealthcareApiClient client, String msgId)
...
+   *
+   * <p>Implementation includes overhead for: 1. two api calls to determine the min/max sendTime of
+   * the HL7v2 store at invocation time. 2. initial splitting into non-overlapping time ranges
+   * (default daily) to achieve parallelization in separate messages.list calls.
```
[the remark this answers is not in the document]
jaketf: I originally included this for users who may try to benchmark this against tiny / sparse results set and be surprised why it is slow / making so many api calls.

I see your point will remove.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
```diff
@@ -415,10 +423,29 @@ private Message fetchMessage(HealthcareApiClient client, String msgId)
...
+   *
+   * <p>This will make more queries than necessary when used with very small data sets. (or very
+   * sparse data sets in the sendTime dimension).
+   *
+   * <p>If you have large but sparse data (e.g. hours between consecutive message sendTimes) and
+   * know something about the time ranges where you have no data, consider using multiple instances
+   * of this transform specifying sendTime filters to omit the ranges where there is no data.
```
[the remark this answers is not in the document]
jaketf: That's great to know! will remove this guidance as it will lead to unnecessary complexity.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
```diff
@@ -415,10 +423,29 @@ private Message fetchMessage(HealthcareApiClient client, String msgId)
     }
   }
 
-  /** List HL7v2 messages in HL7v2 Stores with optional filter. */
+  /**
+   * List HL7v2 messages in HL7v2 Stores with optional filter.
+   *
+   * <p>This transform is optimized for dynamic splitting of message.list calls for large batches of
```
jaketf:
```suggestion
   * <p>This transform is optimized for splitting of message.list calls for large batches of
```

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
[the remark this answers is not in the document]
jaketf: note to self: remove reference to "dynamically rebalance" as this is not yet supported by dataflow runner.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
jaketf: Thanks for suggestion.
In the interest of time, can I punt this to a future PR?

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
```diff
@@ -427,29 +454,75 @@ private Message fetchMessage(HealthcareApiClient client, String msgId)
...
-    private final String filter;
+    private static final Logger LOG = LoggerFactory.getLogger(ListHL7v2MessagesFn.class);
+    private ValueProvider<String> filter;
+    // These control the initial restriction split which means that the list of integer pairs
+    // must comfortably fit in memory.
+    private static final Duration DEFAULT_DESIRED_SPLIT_DURATION = Duration.standardDays(1);
+    private static final Duration DEFAULT_MIN_SPLIT_DURATION = Duration.standardHours(1);
+    private Duration initialSplitDuration;
+    private Instant from;
+    private Instant to;
```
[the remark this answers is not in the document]
jaketf: I don't think so they don't get set until we make the earliest / lastest sendTime query in @GetInitialRestriction

pabloem:
retest this please

pabloem:
Run Java PostCommit

pabloem:
Run Java PostCommit

jaketf:
@pabloem please retest this

pabloem:
Retest this please

pabloem:
Run Java postcommit

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
```diff
@@ -472,24 +523,77 @@ public void initClient() throws IOException {
...
+      long lastClaimedMilliSecond = startRestriction.getMillis() - 1;
+      for (HL7v2Message msg : FluentIterable.concat(pages)) {
+        cursor = Instant.parse(msg.getSendTime());
+        if (cursor.getMillis() > lastClaimedMilliSecond && tracker.tryClaim(cursor.getMillis())) {
+          lastClaimedMilliSecond = cursor.getMillis();
+        }
+
+        if (cursor.getMillis() == lastClaimedMilliSecond) { // loop over messages in millisecond.
+          outputReceiver.output(msg);
+        }
```
lukecwik, line 591:
```suggestion
        if (cursor.getMillis() > lastClaimedMilliSecond) {
          // Return early after the first claim failure preventing us from iterating
          // through the remaining messages.
          if (!tracker.tryClaim(cursor.getMillis())) {
            return;
          }
          lastClaimedMilliSecond = cursor.getMillis();
        }

        outputReceiver.output(msg);
```

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
lukecwik: Yes we can defer since this would change what people are getting from the existing implementation and would likely require opt in to not break any existing users.

If nobody depends on this transform yet then it would be wise to address it before adoption since a lot of users typically expect the output timestamp to match the record's source timestamp. This may not apply to this specific source and is dependent on what users expect so itis your judgement call.

If you do go with changing the output timestamp, the watermark tracking would help for streaming pipelines since it would allow them to perform better. The current implementation would still produce correct results with or without and adding it later would be very safe (it may expose problems in pipelines that were already broken for other reasons).

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
```diff
@@ -427,39 +458,59 @@ private Message fetchMessage(HealthcareApiClient client, String msgId)
...
+    // These control the initial restriction split which means that the list of integer pairs
+    // must comfortably fit in memory.
+    private static final Duration DEFAULT_DESIRED_SPLIT_DURATION = Duration.standardDays(1);
+    private static final Duration DEFAULT_MIN_SPLIT_DURATION = Duration.standardHours(1);
+
+    private static final Logger LOG = LoggerFactory.getLogger(ListHL7v2MessagesFn.class);
+    private ValueProvider<String> filter;
+    private Duration initialSplitDuration;
+    private Instant from;
+    private Instant to;
```
lukecwik, line 499: `from` and `to` seem to only be used within `@GetInitialRestriction`, can we make them local variables there?

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/healthcare/HL7v2IO.java
```diff
@@ -415,10 +424,32 @@ private Message fetchMessage(HealthcareApiClient client, String msgId)
...
   }
 
-  /** List HL7v2 messages in HL7v2 Stores with optional filter. */
+  /**
+   * List HL7v2 messages in HL7v2 Stores with optional filter.
+   *
+   * <p>This transform is optimized for splitting of message.list calls for large batches of
+   * historical data and assumes rather continuous stream of sendTimes.
+   *
+   * <p>Note on Benchmarking By default, this will make more queries than necessary when used with
```
lukecwik, line 433: `Benchmarking By`?

Awkward sentence and capitalization.

pabloem:
Retest this please

pabloem:
retest this please

pabloem:
Run Java Postcommit

pabloem:
Run Java Postcommit

pabloem:
Run Java PostCommit

PR #11600, opened 2020-05-04
Allow accessing window parameters in finishBundle

reuvenlax:
A bundle in Beam today can contain many windows. This often makes using finishBundle correctly tricky. Users are sometimes seen keeping maps in their DoFn of windows seen in a bundle so they can properly process them in finishBundle. We also don't support injecting an OutputReceiver in finishBundle, as there's no good way to associate a window to the output.

This PR allows injecting a BoundedWindow or OutputReceiver to the finishBundle function. In this case, the finishBundle will be invoked once per window seen in the bundle. If there is no such parameter, then the previous behavior of invoking finishBundle once is preserved.


pabloem:
can you please create a JIRA?

PR #11607, opened 2020-05-04
[BEAM-9430] Fixes the bounds of initial watermark set to estimators instead of raising an error

chamikaramj:
Fixes the bounds of initial watermark set to estimators instead of raising an error.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/io/Read.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/io/Read.java
@@ -508,7 +508,7 @@ public class Read {
       while (tracker.tryClaim(out)) {
         receiver.outputWithTimestamp(
             new ValueWithRecordId<>(out[0].getValue(), out[0].getId()), out[0].getTimestamp());
-        watermarkEstimator.setWatermark(out[0].getWatermark());
+        watermarkEstimator.setWatermark(ensureTimestampWithinBounds(out[0].getWatermark()));
       }
 
       // Add the checkpoint mark to be finalized if the checkpoint mark isn't trivial.
@@ -535,16 +535,19 @@ public class Read {
       return currentElementTimestamp;
     }
 
+    private Instant ensureTimestampWithinBounds(Instant timestamp) {
+      if (timestamp.isBefore(BoundedWindow.TIMESTAMP_MIN_VALUE)) {
+        timestamp = BoundedWindow.TIMESTAMP_MIN_VALUE;
+      } else if (timestamp.isAfter(BoundedWindow.TIMESTAMP_MAX_VALUE)) {
+        timestamp = BoundedWindow.TIMESTAMP_MAX_VALUE;
+      }
+      return timestamp;
+    }
+
     @NewWatermarkEstimator
     public WatermarkEstimators.Manual newWatermarkEstimator(
         @WatermarkEstimatorState Instant watermarkEstimatorState) {
-      // Making sure that the watermark is within bounds.
-      if (watermarkEstimatorState.isBefore(BoundedWindow.TIMESTAMP_MIN_VALUE)) {
-        watermarkEstimatorState = BoundedWindow.TIMESTAMP_MIN_VALUE;
-      } else if (watermarkEstimatorState.isAfter(BoundedWindow.TIMESTAMP_MAX_VALUE)) {
-        watermarkEstimatorState = BoundedWindow.TIMESTAMP_MAX_VALUE;
-      }
-      return new WatermarkEstimators.Manual(watermarkEstimatorState);
+      return new WatermarkEstimators.Manual(ensureTimestampWithinBounds(watermarkEstimatorState));
     }
 
     @GetRestrictionCoder
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/io/Read.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/io/Read.java
@@ -46,6 +46,7 @@ import org.apache.beam.sdk.transforms.splittabledofn.RestrictionTracker;
 import org.apache.beam.sdk.transforms.splittabledofn.RestrictionTracker.HasProgress;
 import org.apache.beam.sdk.transforms.splittabledofn.SplitResult;
 import org.apache.beam.sdk.transforms.splittabledofn.WatermarkEstimators;
+import org.apache.beam.sdk.transforms.windowing.BoundedWindow;
 import org.apache.beam.sdk.util.NameUtils;
 import org.apache.beam.sdk.util.SerializableUtils;
 import org.apache.beam.sdk.values.KV;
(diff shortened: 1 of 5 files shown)


ihji:
Shouldn't we also disable `IllegalArgumentException`s in the constructors?

chamikaramj:
Closing this temporarily while I look into this further.

chamikaramj:
Replaced the exception with a bound adjustment. PTAL.

chamikaramj:
Run Java PreCommit

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/WatermarkEstimators.java
```diff
@@ -37,14 +37,16 @@
...
-      this.watermark = checkNotNull(watermark, "watermark must not be null.");
-      if (watermark.isBefore(GlobalWindow.TIMESTAMP_MIN_VALUE)
-          || watermark.isAfter(GlobalWindow.TIMESTAMP_MAX_VALUE)) {
-        throw new IllegalArgumentException(
-            String.format(
-                "Provided watermark %s must be within bounds [%s, %s].",
-                watermark, GlobalWindow.TIMESTAMP_MIN_VALUE, GlobalWindow.TIMESTAMP_MAX_VALUE));
+      checkNotNull(watermark, "watermark must not be null.");
+
+      // Making sure that the watermark is within bounds.
```
lukecwik, line 42: The bounds are wrong, they should be `BoundedWindow.TIMESTAMP_MIN_VALUE` and `BoundedWindow.TIMESTAMP_MAX_VALUE`. This could have been the issue with the timestamp being out of range. If that is the case then I would rather keep the bounds as they were and make setWatermark more restrictive. We should also update the javadoc for setWatermark/currentWatermark (on the interfaces) saying that the value must always be within these bounds.

I think it makes more sense keep the restriction here and also make the restriction within setWatermark as well so the user gets an error when setting it.

chamikaramj: What's the difference between those values. Seems like GlobalWindow get's these constants from the super class BoundedWindow ?

lukecwik: `MIN` is the same but `GlobalWindow.MAX = BoundedWindow.MAX - 1 day` as can be seen here: https://github.com/apache/beam/blob/25b4ebc65953af34a0cfe83ed80feee0bc6df4ad/model/pipeline/src/main/proto/beam_runner_api.proto#L38

chamikaramj: Right, but we are using the constant from the super class here (TIMESTAMP_MAX_VALUE) not END_OF_GLOBAL_WINDOW.
https://github.com/apache/beam/blob/25b4ebc65953af34a0cfe83ed80feee0bc6df4ad/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/windowing/GlobalWindow.java#L39

Also, the error we saw was following which points to min value being off, not max value.
java.lang.IllegalArgumentException: Provided watermark -290308-12-21T19:59:05.224Z must be within bounds [-290308-12-21T19:59:05.225Z, 294247-01-10T04:00:54.775Z].

lukecwik: Your right, it would be good to migrate to use BoundedWindow as the import for the static though.

I think it makes sense to make the constructor validate the bounds and have setWatermark ensure that the value is within the range as expected. We can fix the UnboundedSource SDF wrapper to clamp the watermark value that is being reported from UnboundedReader instead.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/WatermarkEstimators.java
```diff
@@ -37,18 +37,23 @@
     private Instant lastReportedWatermark;
 
     public Manual(Instant watermark) {
+      validateWatermark(watermark);
       this.watermark = checkNotNull(watermark, "watermark must not be null.");
-      if (watermark.isBefore(GlobalWindow.TIMESTAMP_MIN_VALUE)
-          || watermark.isAfter(GlobalWindow.TIMESTAMP_MAX_VALUE)) {
+    }
+
+    private void validateWatermark(Instant watermark) {
```
ihji, line 44: How about moving to parent scope and making static so we can share this method in all three estimators?

chamikaramj: Done.

ihji:
I can confirm that this PR fixes the issue.

sdks/java/core/src/main/java/org/apache/beam/sdk/io/Read.java
```diff
@@ -537,6 +538,12 @@ public Instant getInitialWatermarkEstimatorState(@Timestamp Instant currentEleme
     @NewWatermarkEstimator
     public WatermarkEstimators.Manual newWatermarkEstimator(
         @WatermarkEstimatorState Instant watermarkEstimatorState) {
+      // Making sure that the watermark is within bounds.
+      if (watermarkEstimatorState.isBefore(BoundedWindow.TIMESTAMP_MIN_VALUE)) {
+        watermarkEstimatorState = BoundedWindow.TIMESTAMP_MIN_VALUE;
+      } else if (watermarkEstimatorState.isAfter(BoundedWindow.TIMESTAMP_MAX_VALUE)) {
+        watermarkEstimatorState = BoundedWindow.TIMESTAMP_MAX_VALUE;
+      }
```
lukecwik, line 546: I was thinking that this logic would be where we call setWatermark above on line 510

Note that the initial watermark estimate state is current element timestamp which is always between MIN and MAX timestamp values.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/WatermarkEstimator.java
```diff
@@ -44,4 +45,19 @@
    * <p>The state returned must not be mutated.
    */
   WatermarkEstimatorStateT getState();
+
+  /**
+   * Validates that a given watermark is within timestamp min and max bounds.
+   *
+   * @param watermark watermark to validate
+   */
+  static void ensureWatermarkWithinBounds(Instant watermark) {
```
lukecwik, line 54: It makes more sense to add this method to BoundedWindow as it should apply to a lot more places then just WatermarkEstimators.

chamikaramj: Done.

sdks/java/core/src/main/java/org/apache/beam/sdk/io/Read.java
chamikaramj: Added to both.

PR #11609, opened 2020-05-05
[BEAM-9887] Throw IllegalArgumentException when building Row with logical types with Invalid input

rahul8383:
`schema.logicaltypes.FixedBytes` logical type expects an argument - the length of the byte[].

When an invalid input value (with length < expectedLength) is provided while building the Row with FixedBytes logical type, `IllegalArgumentException` is expected. But, the Exception is not thrown. The below code illustrates the behaviour:

```
 Schema schema = Schema.builder().addLogicalTypeField("char", FixedBytes.of(10)).build();
 byte[] byteArray = {1, 2, 3, 4, 5};
 Row row = Row.withSchema(schema).withFieldValue("char", byteArray).build();
 System.out.println(Arrays.toString(row.getLogicalTypeValue("char", byte[].class)));
```
The above code prints "[1, 2, 3, 4, 5]" with length 5 to the console, whereas the expected length of FixedBytes, is 10.

This PR fixes the issue.
Negative and Positive Test for FixedBytes Logical type are added. 

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/values/RowUtils.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/values/RowUtils.java
@@ -519,7 +519,7 @@ class RowUtils {
       Object retValue = null;
       FieldOverride override = override(rowPosition);
       if (override != null) {
-        retValue = override.getOverrideValue();
+        retValue = logicalType.toInputType(logicalType.toBaseType(override.getOverrideValue()));
       } else if (value != null) {
         retValue = logicalType.toInputType(logicalType.toBaseType(value));
       }
--- a/sdks/java/core/src/test/java/org/apache/beam/sdk/values/RowTest.java
+++ b/sdks/java/core/src/test/java/org/apache/beam/sdk/values/RowTest.java
@@ -23,6 +23,7 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
 
 import java.math.BigDecimal;
 import java.nio.ByteBuffer;
@@ -34,12 +35,12 @@ import java.util.stream.Stream;
 import org.apache.beam.sdk.schemas.Schema;
 import org.apache.beam.sdk.schemas.Schema.FieldType;
 import org.apache.beam.sdk.schemas.logicaltypes.EnumerationType;
+import org.apache.beam.sdk.schemas.logicaltypes.FixedBytes;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.ImmutableList;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.ImmutableMap;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.Lists;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
-import org.junit.Assert;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
@@ -724,7 +725,7 @@ public class RowTest {
     Row a = Row.withSchema(schema).addValue(a0).build();
     Row b = Row.withSchema(schema).addValue(b0).build();
 
-    Assert.assertEquals(a, b);
+    assertEquals(a, b);
   }
 
   @Test
@@ -737,6 +738,28 @@ public class RowTest {
     Row a = Row.withSchema(schema).addValue(ByteBuffer.wrap(a0)).build();
     Row b = Row.withSchema(schema).addValue(ByteBuffer.wrap(b0)).build();
 
-    Assert.assertEquals(a, b);
+    assertEquals(a, b);
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void testLogicalTypeWithInvalidInputValueByFieldName() {
+    Schema schema = Schema.builder().addLogicalTypeField("char", FixedBytes.of(10)).build();
+    byte[] byteArrayWithLengthFive = {1, 2, 3, 4, 5};
+    Row row = Row.withSchema(schema).withFieldValue("char", byteArrayWithLengthFive).build();
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void testLogicalTypeWithInvalidInputValueByFieldIndex() {
+    Schema schema = Schema.builder().addLogicalTypeField("char", FixedBytes.of(10)).build();
+    byte[] byteArrayWithLengthFive = {1, 2, 3, 4, 5};
+    Row row = Row.withSchema(schema).addValues(byteArrayWithLengthFive).build();
+  }
+
+  @Test
+  public void testFixedBytes() {
+    Schema schema = Schema.builder().addLogicalTypeField("char", FixedBytes.of(10)).build();
+    byte[] byteArray = {1, 2, 3, 4, 5, 6, 7, 8, 9, 10};
+    Row row = Row.withSchema(schema).withFieldValue("char", byteArray).build();
+    assertTrue(Arrays.equals(byteArray, row.getLogicalTypeValue("char", byte[].class)));
   }
 }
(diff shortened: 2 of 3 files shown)


rahul8383:
R: @reuvenlax @TheNeuralBit

rahul8383:
I have a couple of questions regarding the behaviour of logical types:

- Will this line ever get hit? If not, can we modify `FixedBytes` as a `PassThroughLogicalType`?: https://github.com/apache/beam/blob/34c58c42f14d2534a2f72f9194ddf9cc69138eea/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/logicaltypes/FixedBytes.java#L78

- Can we consider that the input value provided is of BaseType, which we can convert to InputType and store in memory?

iemejia:
retest this please

TheNeuralBit:
Hm so there are several ways of manually building a Row instance that provide different levels of runtime type-checking. `Row#addValues` explicitly validates everything, and `Row#attachValues` explicitly does not, for performance reasons. In SQL we have an option to switch between the two: 

https://github.com/apache/beam/blob/34c58c42f14d2534a2f72f9194ddf9cc69138eea/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/impl/rel/BeamAggregationRel.java#L354-L357

So we can have runtime type-checking for debugging, but then turn it off for performance.

I'm not sure how `withFieldValue` is intended to work. I'm not sure if the missing `toInputType(toBaseType(value))` for that code path is intentional or an oversight. Can you clarify @reuvenlax?

rahul8383:
@TheNeuralBit
schemas.logicaltypes.LogicalTypesTest.testFixedBytesIllegalArgument Test will fail even if addValue() method is used instead of withFieldValue() method.

reuvenlax:
@TheNeuralBit withFieldValue should replace addValues for must users. addValues is difficult and error prone and withFieldValues allows building a row based on named fields instead of positional fields.

reuvenlax:
LGTM

rahul8383:
@reuvenlax @TheNeuralBit 
Thanks for the review.
Could you please clarify the questions that I have asked above?
I wanted to understand more about logical types for my PR #11581

TheNeuralBit:
> Will this line ever get hit?

We always convert logical types to their base type when serializing with SchemaCoder, and convert back to the input type when deserializing. Other than that I think the only time it should get called is when constructing a Row instance (unless you use attachValues).

> Can we consider that the input value provided is of BaseType, which we can convert to InputType and store in memory?

Would this just be so that we're guaranteed to call `toInputType` whenever setting a value on Row? This PR accomplishes the same thing right?

sdks/java/core/src/test/java/org/apache/beam/sdk/schemas/logicaltypes/LogicalTypesTest.java
```diff
@@ -97,4 +99,19 @@ public void testNanosDuration() {
...
+    Row row = Row.withSchema(schema).withFieldValue("char", byteArrayWithLengthFive).build();
+  }
+
+  @Test
+  public void testFixedBytes() {
+    Schema schema = Schema.builder().addLogicalTypeField("char", FixedBytes.of(10)).build();
+    byte[] byteArray = {1, 2, 3, 4, 5, 6, 7, 8, 9, 10};
+    Row row = Row.withSchema(schema).withFieldValue("char", byteArray).build();
+    assertTrue(Arrays.equals(byteArray, row.getLogicalTypeValue("char", byte[].class)));
+  }
```
TheNeuralBit, line 116: Since these tests are really checking `Row`'s verification, I think they would be better in `RowTest`. Could you move them there?

rahul8383:
> We always convert logical types to their base type when serializing with SchemaCoder, and convert back to the input type when deserializing. Other than that I think the only time it should get called is when constructing a Row instance (unless you use attachValues).

In that case, there is no need to handle this `else` case right? as we are making sure that the input has expected length while building the Row.
https://github.com/apache/beam/blob/5e1571760b61b8ce247d5375b71c8df4d69d6409/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/logicaltypes/FixedBytes.java#L77
Even if `attachValues` is used while building the Row and the provided input value is invalid(invalid length), during serialization in `SchemaCoder`, the input value cannot be converted to base type as it doesn't have expected length and an `IllegalArgumentException` will be thrown.

> Would this just be so that we're guaranteed to call `toInputType` whenever setting a value on Row? This PR accomplishes the same thing right?

Can we support this feature: depending on the type of the input value provided while building the Row, we can call `toInputType(toBaseType(inputValue))` or `toInputType(inputValue)` i.e. support for providing base value while building the Row. If both the InputType and BaseType are one and the same, we can directly call `toInputType(inputValue)`. I am thinking that this might be helpful for logical types like `FixedBytes` or `FixedLengthString`.

sdks/java/core/src/test/java/org/apache/beam/sdk/schemas/logicaltypes/LogicalTypesTest.java
```diff
@@ -97,4 +99,19 @@ public void testNanosDuration() {
     assertEquals(duration, row.getLogicalTypeValue(0, NanosDuration.class));
     assertEquals(durationAsRow, row.getBaseValue(0, Row.class));
   }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void testFixedBytesIllegalArgument() {
+    Schema schema = Schema.builder().addLogicalTypeField("char", FixedBytes.of(10)).build();
+    byte[] byteArrayWithLengthFive = {1, 2, 3, 4, 5};
+    Row row = Row.withSchema(schema).withFieldValue("char", byteArrayWithLengthFive).build();
+  }
```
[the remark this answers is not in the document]
rahul8383: Added `RowTest.testLogicalTypeWithInvalidInputValueByFieldIndex`

sdks/java/core/src/test/java/org/apache/beam/sdk/schemas/logicaltypes/LogicalTypesTest.java
rahul8383: Moved the tests to RowTest.java

case in point! 
How can I write `FixedBytes` test which tests the behaviour of appending zeros? To test this behaviour, the input value should have length < expectedLength. But, if the input value's length is less than expected length, an `IllegalArgumentException` is thrown while building the Row.

TheNeuralBit:
> In that case, there is no need to handle this else case right? as we are making sure that the input has expected length while building the Row.

`checkArgument` ensures that the condition is true, so we are rejecting byte arrays that are too short, and accepting anything longer than the fixed size. I don't think there's any capability that pads zeroes for short arrays, this logic just allows _longer_ byte arrays, and copies out the appropriate portion.

TheNeuralBit:
Whoops sorry I completely misread that. We accept short byte arrays, and zero-pad them. We do _not_ accept long byte arrays so that we don't have to truncate them.

TheNeuralBit:
Ahh ok. I'm sorry for being so dense, I see what you're saying now. In `toBaseType` we validate that the array's length == the fixed length: 
https://github.com/apache/beam/blob/5e1571760b61b8ce247d5375b71c8df4d69d6409/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/logicaltypes/FixedBytes.java#L67-L70

This is where the exception checked in your new tests is thrown.

So we never actually get into `toInputType` where we could accept a shorter byte array and zero-pad. And there doesn't seem to be any other way to set a value by base type with a shorter byte array and pad it.

This seems to be a holdover. Previously Row stored logical type values as their base type, so we probably called `toBaseType(toInputType(x))`.

TheNeuralBit:
I'd be +1 for just dropping the padding logic. I don't think it should be the responsibility of the LogicalType to coerce values like this. What do you think @reuvenlax?

rahul8383:
> This seems to be a holdover. Previously Row stored logical type values as their base type, so we probably called `toBaseType(toInputType(x))`.

Even before the code which changed `Row` to store logical type values instead of base values in memory, while building the row, the input value is expected to be of correct length.
https://github.com/apache/beam/blob/9f0cb649d39ee6236ea27f111acb4b66591a80ec/sdks/java/core/src/main/java/org/apache/beam/sdk/values/Row.java#L658-L660
I think this issue is present since the introduction of `FixedBytes`.

rahul8383:
If there are no comments, can we close this PR?

rahul8383:
> I'd be +1 for just dropping the padding logic. I don't think it should be the responsibility of the LogicalType to coerce values like this. What do you think @reuvenlax?

@TheNeuralBit @reuvenlax I have taken care of this for `FixedBytes` and other standard logical types in #11581

rahul8383:
cc: @ibzib 

@ibzib can this fix be cherry-picked for 2.21.0 release?

PR #11610, opened 2020-05-05
[BEAM-9825] | Implement Intersect,Union,Except transforms

darshanj:
Motivation:
1. Give high level general purpose set relational algebra transforms.
2. Utiltize them from BeamSQL code.

Implement new high-level transforms and use them in BeamSQL code.

- Intersect: Compute the intersection between elements of two PCollection.
- Except: Compute the difference between elements of two PCollection.
- Union: Find the elements that are either of two PCollection.

Implement IntersectAll, ExceptAll and UnionAll variants of transforms.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
@@ -51,7 +51,8 @@ public class SetFns {
   public static <T> SetImpl<T> intersect(PCollection<T> rightCollection) {
     checkNotNull(rightCollection, "rightCollection argument is null");
     SerializableBiFunction<Long, Long, Long> intersectFn =
-        (numberOfElementsinLeft, numberOfElementsinRight) -> (numberOfElementsinLeft > 0 && numberOfElementsinRight > 0) ? 1L : 0L;
+        (numberOfElementsinLeft, numberOfElementsinRight) ->
+            (numberOfElementsinLeft > 0 && numberOfElementsinRight > 0) ? 1L : 0L;
     return new SetImpl<>(rightCollection, intersectFn);
   }
 
@@ -79,7 +80,10 @@ public class SetFns {
   public static <T> SetImpl<T> intersectAll(PCollection<T> rightCollection) {
     checkNotNull(rightCollection, "rightCollection argument is null");
     SerializableBiFunction<Long, Long, Long> intersectFn =
-        (numberOfElementsinLeft, numberOfElementsinRight) -> (numberOfElementsinLeft > 0 && numberOfElementsinRight > 0) ? Math.min(numberOfElementsinLeft, numberOfElementsinRight) : 0L;
+        (numberOfElementsinLeft, numberOfElementsinRight) ->
+            (numberOfElementsinLeft > 0 && numberOfElementsinRight > 0)
+                ? Math.min(numberOfElementsinLeft, numberOfElementsinRight)
+                : 0L;
     return new SetImpl<>(rightCollection, intersectFn);
   }
 
@@ -106,7 +110,8 @@ public class SetFns {
   public static <T> SetImpl<T> except(PCollection<T> rightCollection) {
     checkNotNull(rightCollection, "rightCollection argument is null");
     SerializableBiFunction<Long, Long, Long> exceptFn =
-        (numberOfElementsinLeft, numberOfElementsinRight) -> numberOfElementsinLeft > 0 && numberOfElementsinRight == 0 ? 1L : 0L;
+        (numberOfElementsinLeft, numberOfElementsinRight) ->
+            numberOfElementsinLeft > 0 && numberOfElementsinRight == 0 ? 1L : 0L;
     return new SetImpl<>(rightCollection, exceptFn);
   }
 
@@ -167,7 +172,8 @@ public class SetFns {
    */
   public static <T> SetImpl<T> union(PCollection<T> rightCollection) {
     checkNotNull(rightCollection, "rightCollection argument is null");
-    SerializableBiFunction<Long, Long, Long> unionFn = (numberOfElementsinLeft, numberOfElementsinRight) -> 1L;
+    SerializableBiFunction<Long, Long, Long> unionFn =
+        (numberOfElementsinLeft, numberOfElementsinRight) -> 1L;
     return new SetImpl<>(rightCollection, unionFn);
   }
 
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
@@ -19,7 +19,9 @@ package org.apache.beam.sdk.transforms;
 
 import static org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Preconditions.checkNotNull;
 
+import java.util.ArrayList;
 import java.util.List;
+import org.apache.beam.sdk.Pipeline;
 import org.apache.beam.sdk.transforms.join.CoGbkResult;
 import org.apache.beam.sdk.transforms.join.CoGroupByKey;
 import org.apache.beam.sdk.transforms.join.KeyedPCollectionTuple;
@@ -50,15 +52,15 @@ public class SetFns {
    * PCollection<String> right = p.apply(Create.of("1", "3", "4","4", "6"));
    *
    * PCollection<String> results =
-   *     left.apply(SetFns.intersect(right)); // results will be PCollection<String> containing: "1","3","4"
+   *     left.apply(SetFns.intersectDistinct(right)); // results will be PCollection<String> containing: "1","3","4"
    *
    * }</pre>
    *
    * @param <T> the type of the elements in the input and output {@code PCollection<T>}s.
    */
-  public static <T> SetImpl<T> intersect(PCollection<T> rightCollection) {
+  public static <T> SetImpl<T> intersectDistinct(PCollection<T> rightCollection) {
     checkNotNull(rightCollection, "rightCollection argument is null");
-    return new SetImpl<>(rightCollection, intersect());
+    return new SetImpl<>(rightCollection, intersectDistinct());
   }
 
   /**
@@ -87,14 +89,14 @@ public class SetFns {
    * // Following example will perform (first intersect second) intersect third.
    * PCollection<String> results =
    *     PCollectionList.of(first).and(second).and(third)
-   *     .apply(SetFns.intersect()); // results will be PCollection<String> containing: "3","4"
+   *     .apply(SetFns.intersectDistinct()); // results will be PCollection<String> containing: "3","4"
    *
    * }</pre>
    *
    * @param <T> the type of the elements in the input {@code PCollectionList<T>} and output {@code
    *     PCollection<T>}s.
    */
-  public static <T> SetImplCollections<T> intersect() {
+  public static <T> SetImplCollections<T> intersectDistinct() {
     SerializableBiFunction<Long, Long, Long> intersectFn =
         (numberOfElementsinLeft, numberOfElementsinRight) ->
             (numberOfElementsinLeft > 0 && numberOfElementsinRight > 0) ? 1L : 0L;
@@ -185,14 +187,14 @@ public class SetFns {
    * PCollection<String> right = p.apply(Create.of("1","1", "3", "4","4", "6"));
    *
    * PCollection<String> results =
-   *     left.apply(SetFns.except(right)); // results will be PCollection<String> containing: "2","5"
+   *     left.apply(SetFns.exceptDistinct(right)); // results will be PCollection<String> containing: "2","5"
    * }</pre>
    *
    * @param <T> the type of the elements in the input and output {@code PCollection<T>}s.
    */
-  public static <T> SetImpl<T> except(PCollection<T> rightCollection) {
+  public static <T> SetImpl<T> exceptDistinct(PCollection<T> rightCollection) {
     checkNotNull(rightCollection, "rightCollection argument is null");
-    return new SetImpl<>(rightCollection, except());
+    return new SetImpl<>(rightCollection, exceptDistinct());
   }
 
   /**
@@ -222,14 +224,14 @@ public class SetFns {
    * // Following example will perform (first intersect second) intersect third.
    * PCollection<String> results =
    *     PCollectionList.of(first).and(second).and(third)
-   *     .apply(SetFns.except()); // results will be PCollection<String> containing: "5"
+   *     .apply(SetFns.exceptDistinct()); // results will be PCollection<String> containing: "5"
    *
    * }</pre>
    *
    * @param <T> the type of the elements in the input {@code PCollectionList<T>} and output {@code
    *     PCollection<T>}s.
    */
-  public static <T> SetImplCollections<T> except() {
+  public static <T> SetImplCollections<T> exceptDistinct() {
     SerializableBiFunction<Long, Long, Long> exceptFn =
         (numberOfElementsinLeft, numberOfElementsinRight) ->
             numberOfElementsinLeft > 0 && numberOfElementsinRight == 0 ? 1L : 0L;
@@ -325,14 +327,14 @@ public class SetFns {
    * PCollection<String> right = p.apply(Create.of("1", "3", "4","4"));
    *
    * PCollection<String> results =
-   *     left.apply(SetFns.distinctUnion(right)); // results will be PCollection<String> containing: "1","2","3","4"
+   *     left.apply(SetFns.unionDistinct(right)); // results will be PCollection<String> containing: "1","2","3","4"
    * }</pre>
    *
    * @param <T> the type of the elements in the input and output {@code PCollection<T>}s.
    */
-  public static <T> SetImpl<T> distinctUnion(PCollection<T> rightCollection) {
+  public static <T> SetImpl<T> unionDistinct(PCollection<T> rightCollection) {
     checkNotNull(rightCollection, "rightCollection argument is null");
-    return new SetImpl<>(rightCollection, distinctUnion());
+    return new SetImpl<>(rightCollection, unionDistinct());
   }
 
   /**
@@ -357,14 +359,14 @@ public class SetFns {
    * // Following example will perform (first intersect second) intersect third.
    * PCollection<String> results =
    *     PCollectionList.of(first).and(second).and(third)
-   *     .apply(SetFns.distinctUnion()); // results will be PCollection<String> containing: "1","2","3","4","5"
+   *     .apply(SetFns.unionDistinct()); // results will be PCollection<String> containing: "1","2","3","4","5"
    *
    * }</pre>
    *
    * @param <T> the type of the elements in the input {@code PCollectionList<T>} and output {@code
    *     PCollection<T>}s.
    */
-  public static <T> SetImplCollections<T> distinctUnion() {
+  public static <T> SetImplCollections<T> unionDistinct() {
     SerializableBiFunction<Long, Long, Long> unionFn =
         (numberOfElementsinLeft, numberOfElementsinRight) -> 1L;
     return new SetImplCollections<>(unionFn);
@@ -434,7 +436,7 @@ public class SetFns {
     return Flatten.pCollections();
   }
 
-  public static class SetImpl<T> extends PTransform<PCollection<T>, PCollection<T>> {
+  private static class SetImpl<T> extends PTransform<PCollection<T>, PCollection<T>> {
 
     private final transient PCollection<T> rightCollection;
     private final PTransform<PCollectionList<T>, PCollection<T>> listTransformFn;
@@ -452,7 +454,8 @@ public class SetFns {
     }
   }
 
-  public static class SetImplCollections<T> extends PTransform<PCollectionList<T>, PCollection<T>> {
+  private static class SetImplCollections<T>
+      extends PTransform<PCollectionList<T>, PCollection<T>> {
 
     private final transient SerializableBiFunction<Long, Long, Long> fn;
 
@@ -466,26 +469,7 @@ public class SetFns {
       int size = all.size();
       if (size == 1) {
         return inputs.get(0); // Handle only one PCollection in list. Coder is already specified
-      } else {
-        PCollection<T> accum = inputs.get(0);
-        List<PCollection<T>> tail = all.subList(1, size);
-
-        for (PCollection<T> second : tail) {
-          accum = performSetOperation(accum, second, fn);
-        }
-
-        return accum.setCoder(accum.getCoder());
       }
-    }
-
-    private static <T> PCollection<T> performSetOperation(
-        PCollection<T> leftCollection,
-        PCollection<T> rightCollection,
-        SerializableBiFunction<Long, Long, Long> fn) {
-
-      TupleTag<Void> leftCollectionTag = new TupleTag<>();
-      TupleTag<Void> rightCollectionTag = new TupleTag<>();
-
       MapElements<T, KV<T, Void>> elementToVoid =
           MapElements.via(
               new SimpleFunction<T, KV<T, Void>>() {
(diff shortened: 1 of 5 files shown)


darshanj:
@amaliujia

darshanj-tvlk:
Added changes for BeamSQL in this PR itself.

amaliujia:
R: @amaliujia 
cc: @jhnmora000 @MarkZeng1998 (expose more PRs to GSoC students)

amaliujia:
cc @Mark-Zeng to make sure I tagged the right person.

amaliujia:
@darshanj 

You can run `./gradlew spotlessApply` to fix checkstyle issues.
You need to run `./gradlew  ${module}:check` command to not only run tests but also run style check.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
```diff
@@ -0,0 +1,261 @@
...
+   * PCollection<String> right = p.apply(Create.of("1", "3", "4", "6"));
+   *
+   * PCollection<String> results =
+   *     left.apply(SetFns.intersectAll(right));
+   * }</pre>
+   */
+  public static <T> SetImpl<T> intersectAll(PCollection<T> rightCollection) {
+    checkNotNull(rightCollection, "rightCollection argument is null");
+    SerializableBiFunction<Long, Long, Long> intersectFn =
+        (numberOfElementsinLeft, numberOfElementsinRight) -> (numberOfElementsinLeft > 0 && numberOfElementsinRight > 0) ? Math.min(numberOfElementsinLeft, numberOfElementsinRight) : 0L;
```
robertwb, line 82: Is not `Math.min(numberOfElementsinLeft, numberOfElementsinRight)` sufficient?

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
```diff
@@ -0,0 +1,261 @@
...
+   * }</pre>
+   */
+  public static <T> SetImpl<T> exceptAll(PCollection<T> rightCollection) {
+    checkNotNull(rightCollection, "rightCollection argument is null");
+    SerializableBiFunction<Long, Long, Long> exceptFn =
+        (numberOfElementsinLeft, numberOfElementsinRight) -> {
+          if (numberOfElementsinLeft > 0 && numberOfElementsinRight == 0) {
+            return numberOfElementsinLeft;
+          } else if (numberOfElementsinLeft > 0 && numberOfElementsinRight > 0) {
+            return Math.max(numberOfElementsinLeft - numberOfElementsinRight, 0L);
```
robertwb, line 142: Isn't this expression always sufficient?

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
```diff
@@ -0,0 +1,261 @@
...
+   *
+   * <pre>{@code
+   * Pipeline p = ...;
+   *
+   * PCollection<String> left = p.apply(Create.of("1", "2", "3", "4", "5"));
+   * PCollection<String> right = p.apply(Create.of("1", "3", "4", "6"));
+   *
+   * PCollection<String> results =
+   *     left.apply(SetFns.union(right));
+   * }</pre>
```
robertwb, line 166: For all of these, it'd be useful to document what the expected output is.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
```diff
@@ -0,0 +1,261 @@
...
+   */
+  public static <T> SetImpl<T> union(PCollection<T> rightCollection) {
+    checkNotNull(rightCollection, "rightCollection argument is null");
+    SerializableBiFunction<Long, Long, Long> unionFn = (numberOfElementsinLeft, numberOfElementsinRight) -> 1L;
+    return new SetImpl<>(rightCollection, unionFn);
+  }
+
+  /**
+   * Returns a new {@code SetFns.SetImpl<T>} transform that compute the unionAll with provided
```
robertwb, line 175: Isn't this just `Flatten.pcollections`? (Also, isn't that more efficient?)

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
```diff
@@ -0,0 +1,261 @@
...
+
+import org.apache.beam.sdk.transforms.join.CoGbkResult;
+import org.apache.beam.sdk.transforms.join.CoGroupByKey;
+import org.apache.beam.sdk.transforms.join.KeyedPCollectionTuple;
+import org.apache.beam.sdk.values.KV;
+import org.apache.beam.sdk.values.PCollection;
+import org.apache.beam.sdk.values.TupleTag;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.Iterables;
+
+public class SetFns {
```
robertwb, line 30: Though not as chain-able, a `PTransform<PCollectionList<T>, PCollection<T>>` feels more natural for many of these operations.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
```diff
@@ -0,0 +1,261 @@
...
+
+    return coGbkResults.apply(
+        ParDo.of(
+            new DoFn<KV<T, CoGbkResult>, T>() {
+
+              @ProcessElement
+              public void processElement(ProcessContext c) {
+                KV<T, CoGbkResult> elementGroups = c.element();
+
+                CoGbkResult value = elementGroups.getValue();
```
robertwb, line 234: I wonder if it's worth dropping at TODO here to lift combiners through the CoGBK.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
darshanj: Thanks very valuable. I have fixed this.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
darshanj: Thanks very valuable. I have fixed this.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
darshanj: Done.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
darshanj: Thanks very valuable. I have fixed this.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
darshanj: I feel as you told it is not chain-able, keep it `PTransform<PCollection<T>, PCollection<T>>`. Most of the time which I encountered, you will have to use a different set operation followed by another operation. like union and then Intersect or so.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
darshanj: Thanks. I have put a comment. can you clarify please? You mean use `CoGbk.compose`? If possible I would implement it instead of TODO if is not much of an effort.

darshanj:
> Thanks!
> 
> In terms of naming, I found it surprising that union (for example) does deduplication. Maybe name them distinctUnion and multisetUnion or something like that?



> Thanks!
> 
> In terms of naming, I found it surprising that union (for example) does deduplication. Maybe name them distinctUnion and multisetUnion or something like that?

I feel that As it is SetFns.union, it follows SET DISTINCT semantics. What is your view?

darshanj:
retest this please

amaliujia:
retest this please

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
amaliujia: Depends on if we treat SET as binary operation or a "set" operation. If it is binary, this PR's implementation makes sense (plus it is chain-able). If we consider it is "set" operation, then it should be `PTransform<PCollectionList<T>, PCollection<T>>` by the nature of a set.

amaliujia:
> > Thanks!
> > In terms of naming, I found it surprising that union (for example) does deduplication. Maybe name them distinctUnion and multisetUnion or something like that?
> 
> > Thanks!
> > In terms of naming, I found it surprising that union (for example) does deduplication. Maybe name them distinctUnion and multisetUnion or something like that?
> 
> I don't have strong view here but I feel that as it is inside `SetFns` and will be used like `SetFns.union`, it follows SET DISTINCT semantics. It is equivalent of UNION in SQL like `A U B`.Also all other functions in `SetFns` follow SET DISTINCT semantics. What is your view, how do we make it better?

I am wondering if you can use a bool flag to indicate it is "ALL" or "DISTINCT" semantics.  By doing so there is no naming problem.

amaliujia:
You can run `./gradlew $module:check` to run all checks, include unit testing and style check.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
darshanj: Thanks. I understand the reasoning. My thinking was in lines similar like a Join implementation. As an end-user abstraction, isn't it would be more natural and straightforward to think left diff right which abstract that there is CoGrouping inside. This looks in my mind a natural usecase. 

Or will it be good idea to both variants `PCollectionList<T>` and `PCollection<T>` of API?

darshanj: Hi @amaliujia I tried to implement both APIs and it looks like following:
1. Binary version: `left.apply("strings", SetFns.intersect(right))`
2. PCollectionList version: `PCollectionList.of(set1).and(set1).and(set3).apply("stringsColls", SetFns.intersect())` will return. (set1 intersection set2) intersection set3 results.
Please give your thoughts.

darshanj:
> > > Thanks!
> > > In terms of naming, I found it surprising that union (for example) does deduplication. Maybe name them distinctUnion and multisetUnion or something like that?
> > 
> > 
> > > Thanks!
> > > In terms of naming, I found it surprising that union (for example) does deduplication. Maybe name them distinctUnion and multisetUnion or something like that?
> > 
> > 
> > I don't have strong view here but I feel that as it is inside `SetFns` and will be used like `SetFns.union`, it follows SET DISTINCT semantics. It is equivalent of UNION in SQL like `A U B`.Also all other functions in `SetFns` follow SET DISTINCT semantics. What is your view, how do we make it better?
> 
> I am wondering if you can use a bool flag to indicate it is "ALL" or "DISTINCT" semantics. By doing so there is no naming problem.

I think as java doesn't have default parameters, either we need to use builders or API may have always have sent the boolean. I lean towards to rename `union` to `distinctUnion`. That way it is simple to call and also gives clarity on intent.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
```diff
@@ -187,83 +237,112 @@
...
+      if (size == 1) {
+        return inputs.get(0); // Handle only one PCollection in list. Coder is already specified
+      } else {
+        PCollection<T> accum = inputs.get(0);
+        List<PCollection<T>> tail = all.subList(1, size);
 
-    private SetUnionAllImpl(PCollection<T> rightCollection) {
-      this.rightCollection = rightCollection;
+        for (PCollection<T> second : tail) {
+          accum = performSetOperation(accum, second, fn);
```
robertwb, line 474: It'd be better to implement the binary operations in terms of n-ary operations, so that one only need to so a single [Co]GBK. The operations are the same (e.g. union -> sum or 1, intersection = min or max(min(...), 1) for all/distinct variants respectively.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
```diff
@@ -187,83 +237,112 @@
...
+    private SetImplCollections(SerializableBiFunction<Long, Long, Long> fn) {
+      this.fn = fn;
+    }
+
+    private static <T> PCollection<T> performSetOperationCollectionList(
+            PCollectionList<T> inputs, SerializableBiFunction<Long, Long, Long> fn) {
+      List<PCollection<T>> all = inputs.getAll();
+      int size = all.size();
+      if (size == 1) {
+        return inputs.get(0); // Handle only one PCollection in list. Coder is already specified
```
robertwb, line 468: This is only true for the all variants, right?

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
darshanj: Thanks. Thats very good idea. I have done it.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
darshanj: Yes. I need to check this as i can't bindly call `setCoder` if there is only one element in the list.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
darshanj: @robertwb : Do we need this comment? I not understand what is purpose of this TODO quiet well.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
darshanj: renamed all methods with explicit Distinct and All as prefix

darshanj:
retest this please

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
darshanj: I have removed this check now.

darshanj:
@amaliujia @robertwb : Do we verify WindowFns for all pCollections to be the same with first in the list?

amaliujia:
+1 to keep both `List` and non-`list` version.




> @amaliujia @robertwb : Do we verify WindowFns for all pCollections to be the same with first in the list?

I think it is CoGBK's duty to check if all WindowFns are compatible.Set transforms here are built on CoGBK and a ParDo, and ParDo should not worry about WindowFn.

amaliujia:
retest this please

amaliujia:
retest this please

darshanj-tvlk:
:sdks:java:io:rabbitmq:test fails. But when i ran it locally it passes!. Can you please rerun the tests please.

amaliujia:
Run Java PreCommit

amaliujia:
Rerun to see if it is a flaky test.

amaliujia:
Run Java PreCommit

amaliujia:
Tests passed! Will take a final pass soon!

darshanj:
Thanks @amaliujia I have create a JIRA:
https://issues.apache.org/jira/browse/BEAM-9992 based on my understanding of what is required to be implemented
Can you add more details if anything.

amaliujia:
Thank you! The Jira looks good to me!

Will merge this PR tomorrow if there is no other comments.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
```diff
@@ -0,0 +1,528 @@
...
+import org.apache.beam.sdk.transforms.join.CoGbkResult;
+import org.apache.beam.sdk.transforms.join.CoGroupByKey;
+import org.apache.beam.sdk.transforms.join.KeyedPCollectionTuple;
+import org.apache.beam.sdk.values.KV;
+import org.apache.beam.sdk.values.PCollection;
+import org.apache.beam.sdk.values.PCollectionList;
+import org.apache.beam.sdk.values.TupleTag;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.Iterables;
+
+public class SetFns {
```
lukecwik, line 34: We should mention in the comments that we rely on the deterministic encoding of the coder similar to how we do GroupByKey.

Also, this implementation assumes that there will only be a single firing of the trigger. If there are multiple then the results are likely undefined.

darshanj: Added Comments for coders. 

Regarding triggers, I assume, multiple triggers should work if they are compatible.
It uses CGBK (internally uses flattens) , which checks if triggers compatible and windowFns are compatible. User will get results based on whatever data is triggered.

My understanding may be flawed. Please correct if you think that is not the case.

I have added a comment for compatible triggers and same windowFns.

lukecwik: Your understanding is correct that users will get results based upon whatever data is fired because of a trigger but from a cursory reading of the docs, we mention doing intersect/distinct/... over PCollections and not trigger firings which could confuse a user into thinking that intersect/distinct will be over all elements in these PCollections.

darshanj: Thanks. I was thinking to add "If you have multiple triggers configured and fired, output of
   this transform will be calculated on data which is in the respective trigger."

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
```diff
@@ -0,0 +1,618 @@
...
+   * PCollection<String> left = p.apply(Create.of("1","1","1", "2", "3", "3","3","4", "5"));
+   * PCollection<String> right = p.apply(Create.of("1", "3", "4","4", "6"));
+   *
+   * PCollection<String> results =
+   *     left.apply(SetFns.exceptAll(right)); // results will be PCollection<String> containing: "1","1","2","3","3","5"
+   * }</pre>
+   *
+   * @param <T> the type of the elements in the input and output {@code PCollection<T>}s.
+   */
+  public static <T> SetImpl<T> exceptAll(PCollection<T> rightCollection) {
```
lukecwik, line 321: Since SetImpl and SetImplCollection are private classes, users won't be able to assign them to a local variable without converting it to a PTransform which most IDEs won't do for you automatically. Was this your intent?

In other transforms we have made these kinds of classes public but they are usually specific to what they do, so in this case we would have a UnionAll/Intersect/IntersectAll/Distinct/DistinctAll/... PTransform that was public that had whatever internal implementation you want. The alternative would be to return PTransform<PCollection<T>, PCollection<T>> which seems ok.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
```diff
@@ -0,0 +1,618 @@
...
+import org.apache.beam.sdk.transforms.join.KeyedPCollectionTuple;
+import org.apache.beam.sdk.transforms.windowing.Trigger;
+import org.apache.beam.sdk.transforms.windowing.WindowFn;
+import org.apache.beam.sdk.values.KV;
+import org.apache.beam.sdk.values.PCollection;
+import org.apache.beam.sdk.values.PCollectionList;
+import org.apache.beam.sdk.values.TupleTag;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.Iterables;
+
+public class SetFns {
```
lukecwik, line 38: Note, we already have a Distinct transform with some additional capabilities over what we have here but it currently requires users to flatten/merge their PCollections into a single PCollection<T> already.

@robertwb any suggestions on what to do here? (having very similar but duplicated functionality seems like bloat)

lukecwik, line 38: A simple class comment would help people reading the javadoc as to what kinds of transforms are contained in this class.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
darshanj: Yes. It is to avoid mutation and create correct abstraction. As this Transform doesn't have any builder like pattern, does taking out a variable makes any sense?

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
darshanj: Just thought to add my original intention was to have only binary operation. We can make this only 2 to nry transform. As we added pCollectionList api, we just allowed PcollectionList with size==1. It is never meant to perform as Distinct.

darshanj: Thanks.Done

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
```diff
@@ -0,0 +1,618 @@
...
+   * Trigger#isCompatible(Trigger)}).
+   *
+   * <p>By default, the output {@code PCollection<T>} encodes its elements using the same {@code
+   * Coder} that of first in {@code PCollectionList<T>}
+   *
+   * <pre>{@code
+   * Pipeline p = ...;
+   * PCollection<String> first = p.apply(Create.of("1","1","2"));
+   * PCollection<String> second = p.apply(Create.of("1", "3", "4","4"));
+   * PCollection<String> third = p.apply(Create.of("1", "5"));
```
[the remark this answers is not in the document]
darshanj: Thanks.Done

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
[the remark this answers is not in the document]
darshanj: Thanks.Done

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
```diff
@@ -0,0 +1,618 @@
...
+   *
+   * <p>Note that this transform requires that the {@code Coder} of the all {@code PCollection<T>}
+   * to be deterministic (see {@link Coder#verifyDeterministic()}). If the collection {@code Coder}
+   * is not deterministic, an exception is thrown at pipeline construction time.
+   *
+   * <p>All inputs must have equal {@link WindowFn}s and compatible triggers (see {@link
+   * Trigger#isCompatible(Trigger)}).
+   *
+   * <p>By default, the output {@code PCollection<T>} encodes its elements using the same {@code
+   * Coder} that of first in {@code PCollectionList<T>}
```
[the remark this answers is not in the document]
darshanj: Thanks.Done

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
```diff
@@ -0,0 +1,618 @@
...
```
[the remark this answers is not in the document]
darshanj: Thanks.Done

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
```diff
@@ -0,0 +1,618 @@
...
+    return new SetImpl<>(rightCollection, intersectDistinct());
+  }
+
+  /**
+   * Returns a {@code PTransform} that takes a {@code PCollectionList<PCollection<T>>} and returns a
+   * {@code PCollection<T>} containing intersection of collections done in order for all collections
+   * in {@code PCollectionList<T>}.
+   *
+   * <p>Returns a new {@code PTransform} transform that follows SET DISTINCT semantics which takes a
+   * {@code PCollectionList<PCollection<T>>} and returns a {@code PCollection<T>} containing
```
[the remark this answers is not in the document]
darshanj: Thanks.Done

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
```diff
@@ -0,0 +1,618 @@
...
+   *
+   * <p>Note that this transform requires that the {@code Coder} of the all {@code PCollection<T>}
+   * to be deterministic (see {@link Coder#verifyDeterministic()}). If the collection {@code Coder}
+   * is not deterministic, an exception is thrown at pipeline construction time.
+   *
+   * <p>All inputs must have equal {@link WindowFn}s and compatible triggers (see {@link
+   * Trigger#isCompatible(Trigger)}).
+   *
+   * <p>By default, the output {@code PCollection<T>} encodes its elements using the same {@code
+   * Coder} that of first in {@code PCollectionList<T>}
```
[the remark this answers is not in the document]
darshanj: Thanks.Done

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
```diff
@@ -0,0 +1,618 @@
...
+   *
+   * <p>Note that this transform requires that the {@code Coder} of the all {@code PCollection<T>}
+   * to be deterministic (see {@link Coder#verifyDeterministic()}). If the collection {@code Coder}
+   * is not deterministic, an exception is thrown at pipeline construction time.
+   *
+   * <p>All inputs must have equal {@link WindowFn}s and compatible triggers (see {@link
+   * Trigger#isCompatible(Trigger)}).
+   *
+   * <p>By default, the output {@code PCollection<T>} encodes its elements using the same {@code
+   * Coder} that of {@code PCollection<T>}
```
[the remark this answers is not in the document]
darshanj: Thanks.Done

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
darshanj: Thanks.Done

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
darshanj: I have removed to support PCollectionList with only one PCollection

amaliujia:
retest this please

darshanj:
Failed
org.apache.beam.examples.WordCountIT > testE2EWordCount FAILED
Task :runners:google-cloud-dataflow-java:examples-streaming:windmillPreCommit FAILED
@amaliujia 
Looks like not related to this changes. Can you please run tests again.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/SetFns.java
darshanj: Added return value as respective PTransforms.

amaliujia:
retest this please

amaliujia:
run Java Precommit

amaliujia:
Tests retriggered. 

And it becomes a really big PR :-) Nice work!

amaliujia:
Run Java PreCommit

aaltay:
All tests passed. I do not see a LGTM, maybe I am missing. Is this ready to be merged?

amaliujia:
@aaltay I have approved this PR. But will still wait a bit to see if @lukecwik or @robertwb has more comments.

amaliujia:
From what I can tell, comments are addressed properly already. So merging this PR.

amaliujia:
This really becomes a big contribution at the end! Thanks @darshanj!

darshanj:
Thanks all for patiently reviewing this PR and providing valuable feedback.

PR #11613, opened 2020-05-05
[BEAM-9833] Fix .asf.yaml issues, sort labels and disable rebase button

iemejia:
Issues reported by yamllint and some minor fixes. Also set merge button as the only strategy because we don't want to encourage (o even make possible the other two).

R: @pabloem
CC: @ibzib

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -1,36 +1,41 @@
-#    Licensed to the Apache Software Foundation (ASF) under one or more
-#    contributor license agreements.  See the NOTICE file distributed with
-#    this work for additional information regarding copyright ownership.
-#    The ASF licenses this file to You under the Apache License, Version 2.0
-#    (the "License"); you may not use this file except in compliance with
-#    the License.  You may obtain a copy of the License at
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements. See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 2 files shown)


robertwb:
I thought consensus was that we did want to encourage squash for those PRs with a huge pile of fixup commits (and otherwise no semantically meaningful commits).

iemejia:
@robertwb Agree, a consensus that nobody respects :). In the case of this PR I set into the merge approach because we have only [three options](https://help.github.com/en/github/administering-a-repository/about-merge-methods-on-github) and the other two (squash and rebase) do not create the additional merge commit required by the [Beam committer guide](https://beam.apache.org/contribute/committer-guide/#merging-it).

pabloem:
I think squash is used by many, and it facilitates receiving contributions without the extra round trip to the contributors. I think we need to discuss more before removing it. Wdyt @iemejia?

iemejia:
I am ok with squashing if it creates extra commits, but it does not seem to be the case or does it create the additional extra merge commit? I just want to ensure we follow the rules.

Now if the goal is to change the rules maybe we should move the discussion to the ML.
Pinging @kennknowles who created our default policy (extra Merge commit) to see what he thinks.

robertwb:
OK, the rest of this change looks good. Let's get this in, disabling rebase but keeping the others, and taking squash+merge to the list as a separate discussion (and PR if need be).

iemejia:
Done the changes suggested by @robertwb let only rebase disabled. Can someone PTAL. Thanks

pabloem:
LGTM!

PR #11632, opened 2020-05-07
[BEAM-7746] Fix type errors and enable checks for apache_beam.dataframe.*

robertwb:
------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/dataframe/convert.py
+++ b/sdks/python/apache_beam/dataframe/convert.py
@@ -17,6 +17,7 @@
 from __future__ import absolute_import
 
 import inspect
+from typing import TYPE_CHECKING
 from typing import Any
 from typing import Dict
 from typing import Tuple
@@ -27,7 +28,7 @@ from apache_beam.dataframe import expressions
 from apache_beam.dataframe import frame_base
 from apache_beam.dataframe import transforms
 
-if typing.TYPE_CHECKING:
+if TYPE_CHECKING:
   # pylint: disable=ungrouped-imports
   import pandas
 
--- a/sdks/python/apache_beam/dataframe/convert.py
+++ b/sdks/python/apache_beam/dataframe/convert.py
@@ -22,12 +22,14 @@ from typing import Dict
 from typing import Tuple
 from typing import Union
 
-import pandas
-
 from apache_beam import pvalue
 from apache_beam.dataframe import expressions
 from apache_beam.dataframe import frame_base
 from apache_beam.dataframe import transforms
+
+if typing.TYPE_CHECKING:
+  # pylint: disable=ungrouped-imports
+  import pandas
 
 
 # TODO: Or should this be called as_dataframe?
--- a/sdks/python/apache_beam/dataframe/convert.py
+++ b/sdks/python/apache_beam/dataframe/convert.py
@@ -16,22 +16,19 @@
 
 from __future__ import absolute_import
 
-import typing
-
 import inspect
+import typing
+from typing import Any
+from typing import Dict
+from typing import Tuple
+from typing import Union
+
+import pandas
 
 from apache_beam import pvalue
 from apache_beam.dataframe import expressions
 from apache_beam.dataframe import frame_base
 from apache_beam.dataframe import transforms
-
-if typing.TYPE_CHECKING:
-  # pylint: disable=ungrouped-imports
-  from typing import Any
-  from typing import Dict
-  from typing import Tuple
-  from typing import Union
-  import pandas
 
 
 # TODO: Or should this be called as_dataframe?
--- a/sdks/python/apache_beam/dataframe/convert.py
+++ b/sdks/python/apache_beam/dataframe/convert.py
@@ -16,12 +16,22 @@
 
 from __future__ import absolute_import
 
+import typing
+
 import inspect
 
 from apache_beam import pvalue
 from apache_beam.dataframe import expressions
 from apache_beam.dataframe import frame_base
 from apache_beam.dataframe import transforms
+
+if typing.TYPE_CHECKING:
+  # pylint: disable=ungrouped-imports
+  from typing import Any
+  from typing import Dict
+  from typing import Tuple
+  from typing import Union
+  import pandas
 
 
 # TODO: Or should this be called as_dataframe?
@@ -49,9 +59,9 @@ def to_dataframe(
 
 # TODO: Or should this be called from_dataframe?
 def to_pcollection(
-    *dataframes,  # type: Tuple[frame_base.DeferredFrame]
+    *dataframes,  # type: frame_base.DeferredFrame
     **kwargs):
-  # type: (...) -> Union[pvalue.PCollection, Tuple[pvalue.PCollection]]
+  # type: (...) -> Union[pvalue.PCollection, Tuple[pvalue.PCollection, ...]]
 
   """Converts one or more deferred dataframe-like objects back to a PCollection.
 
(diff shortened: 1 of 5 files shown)


robertwb:
R: @chadrik

sdks/python/apache_beam/dataframe/convert.py
```diff
@@ -16,13 +16,23 @@
...
 from apache_beam.dataframe import expressions
 from apache_beam.dataframe import frame_base
 from apache_beam.dataframe import transforms
 
+if typing.TYPE_CHECKING:
+  # pylint: disable=ungrouped-imports
+  from typing import Any
+  from typing import Dict
+  from typing import Tuple
+  from typing import Union
```
chadrik, line 33: Scoping these imports is not a bad idea since it keeps the module namespace cleaner, but there are a couple of issues with this:

- once we get to python 3.x, and move from type comments to annotations, this will fail, as we'll be referencing non-existent objects in our annotations.  
- it's not consistent with how we've done this throughout the rest of the code

chadrik: I should mention that a solution to the first issue is that we can refer to these types as strings, such as `'Any'`, but that's certainly a lot more awkward, and developers are likely to forget to do so and get confused/frustrated.

sdks/python/apache_beam/dataframe/transforms.py
```diff
@@ -248,7 +261,12 @@ def _dict_union(dicts):
   return result
 
 
-def _flatten(valueish, root=()):
+def _flatten(
+    valueish,  # type: Union[T, Tuple[T, ...], Dict[Any, T]]
```
chadrik, line 265: do we want to cover the `List[T]` case for `valueish`?  

Alternately, do we wan to make this more generic:

```python
def _flatten(
    valueish,  # type: Union[T, Iterable[T], Mapping[Any, T]]
    root=(),  # type: Tuple[Any, ...]
    ):
  # type: (...) -> Dict[Tuple[Any, ...], T]
  """Given a nested structure of dicts, tuples, and lists, return a flat
  dictionary where the values are the leafs and the keys are the "paths" to
  these leaves.

  For example `{a: x, b: (y, z)}` becomes `{(a,): x, (b, 0): y, (b, 1): c}`.
  """
  if isinstance(valueish, typing.Mapping):
    return _dict_union(_flatten(v, root + (k, )) for k, v in valueish.items())
  elif isinstance(valueish, typing.Iterable):
    return _dict_union(
        _flatten(v, root + (ix, )) for ix, v in enumerate(valueish))
  else:
    return {root: valueish}
```

Another thought, is it valid / worthwhile to create a relationship between `valueish`, `root`, and the result key?:

```python
def _flatten(
    valueish,  # type: Union[T, Iterable[T], Mapping[U, T]]
    root=(),  # type: Tuple[U, ...]
    ):
  # type: (...) -> Dict[Tuple[U, ...], T]
```

chadrik: I should mention that typing provides a lot of opportunity to go down rabbit-holes, so the right answer is often "it would be more accurate, but it's not that valuable".   A common motivator behind investing in accurately typing utility functions is when it allows you to avoid adding manual / explicit types or casts elsewhere in the code.   Imagine a scenario where mypy knows key of the `valueish` map, but after it passes through `_flatten`, you need to re-type the result because it becomes `Any`.

sdks/python/apache_beam/dataframe/convert.py
robertwb: I actually prefer unconditionally importing them, but was just trying to avoid lint issues (and did see this pattern elsewhere). Changed.

sdks/python/apache_beam/dataframe/transforms.py
robertwb: Generally I'd agree that `Iterable`/`Mapping` would be preferable, but here I want to be restrictive about the kinds of values I decompose. 

The keys types of the mapping would be `Union[None, int, U]`, so you'd have to cast anyway, so I think it's simpler to leave as is. 

Still, good food for thought.

sdks/python/apache_beam/dataframe/convert.py
robertwb: So lint complains about unguarded imports, so I put them back. We'll just to a massive sweep to fix these when we change to use type annotations.

robertwb:
PTAL

sdks/python/apache_beam/dataframe/convert.py
chadrik: What's the lint error?   Is it because of the unused `typing` import?  

I'm confused because unguarded typing imports are used all over the beam codebase without any lint errors.  Check `pipeline`, `pipeline_context`, `pipeline_options`, for starters.

chadrik: I had a look at the lint errors, and they are legitimate, but scoping the imports is not the right solution.

For example, these:

```
  | apache_beam/dataframe/frame_base.py:21:0: W0611: Unused Any imported from typing (unused-import)
  | apache_beam/dataframe/frame_base.py:23:0: W0611: Unused Tuple imported from typing (unused-import) 
  | apache_beam/dataframe/frame_base.py:24:0: W0611: Unused Union imported from typing (unused-import)
```

`Any`, `Tuple`, and `Union` _aren't_ used in this module. `Dict` is the only typing import that is used.

chadrik: Note that pylint is pretty good about detecting the use of typing imports, even in comments.  I have encountered a few edge cases but I reported them and they were fixed pretty quickly.   

The only issue I know of in the version of pylint that we're using is with `typing.BinaryIO`.  You'll notice a few lines like these in the code:

```python
from typing import BinaryIO  # pylint: disable=unused-import
```

I [reported](https://github.com/PyCQA/pylint/issues/3217) this bug and it's been fixed, so we can actually get rid of these exceptions if we upgrade to the latest pylint.

robertwb:
Ping.

sdks/python/apache_beam/dataframe/convert.py
chadrik: This note hasn't been addressed:  

> I had a look at the lint errors, and they are legitimate, but scoping the imports is not the right solution.

see above.

robertwb: Oh, yes. Thanks for catching this. Fixed.

robertwb:
Run PythonLint PreCommit

sdks/python/apache_beam/dataframe/convert.py
chadrik: I think it still needs to be fixed for `dataframe.convert`.

robertwb:
Oh, yes. Fixed all uses of typing.TYPE_CHECKING now.

sdks/python/apache_beam/dataframe/convert.py
robertwb: [sigh] It still didn't like PCollection. `apache_beam/dataframe/transforms.py:30:0: W0611: Unused PCollection imported from apache_beam.pvalue (unused-import)`. But the rest are OK.

robertwb: And pandas needs to be guarded. Hopefully that should be it. PTAL.

sdks/python/apache_beam/dataframe/convert.py
```diff
@@ -17,12 +17,20 @@
 from __future__ import absolute_import
 
 import inspect
+from typing import Any
```
chadrik, line 20:
```suggestion
from typing import TYPE_CHECKING
from typing import Any
```

sdks/python/apache_beam/dataframe/convert.py
```diff
@@ -17,12 +17,20 @@
...
+from typing import Dict
+from typing import Tuple
+from typing import Union
 
 from apache_beam import pvalue
 from apache_beam.dataframe import expressions
 from apache_beam.dataframe import frame_base
 from apache_beam.dataframe import transforms
 
+if typing.TYPE_CHECKING:
```
chadrik, line 30:
```suggestion
if TYPE_CHECKING:
```

sdks/python/apache_beam/dataframe/transforms.py
```diff
@@ -16,13 +16,28 @@
...
+from typing import Union
+
 import pandas as pd
 
 import apache_beam as beam
 from apache_beam import transforms
 from apache_beam.dataframe import expressions
 from apache_beam.dataframe import frames  # pylint: disable=unused-import
 
+if typing.TYPE_CHECKING:
```
chadrik, line 35: The prevailing style for `TYPE_CHECKING` is to import it as `from typing import TYPE_CHECKING`.  I think we should stay consistent.  If we want to change that, it's fine by me, but we can do that in another PR.

robertwb: +1 for consistency. Changed.

chadrik:
LGTM!

robertwb:
Thanks!

PR #11715, opened 2020-05-15
[BEAM-9977] Implement GrowableOffsetRangeTracker

boyuanzz:

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/GrowableOffsetRangeTracker.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/GrowableOffsetRangeTracker.java
@@ -1,0 +1,143 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.beam.sdk.transforms.splittabledofn;
+
+import static org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Preconditions.checkNotNull;
+
+import java.math.BigDecimal;
+import java.math.MathContext;
+import org.apache.beam.sdk.annotations.Experimental;
+import org.apache.beam.sdk.annotations.Experimental.Kind;
+import org.apache.beam.sdk.io.range.OffsetRange;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Suppliers;
+
+/**
+ * An {@link OffsetRangeTracker} for tracking a growable offset range. {@code Long.MAX_VALUE} is
+ * used as the end of the range to indicate infinity.
+ *
+ * <p>An offset range is considered growable when the end offset could grow (or change) during
+ * execution time (e.g., Kafka topic partition offset, appended file, ...).
+ *
+ * <p>The growable range is marked as done by claiming {@code Long.MAX_VALUE}.
+ */
+@Experimental(Kind.SPLITTABLE_DO_FN)
+public class GrowableOffsetRangeTracker extends OffsetRangeTracker {
+  /**
+   * Provides the estimated end offset of the range.
+   *
+   * <p>{@link #estimate} is called to give the end offset when {@link #trySplit} or {@link
+   * #getProgress} is invoked. The end offset is exclusive for the range. The estimated end is not
+   * required to monotonically increase as it will only be taken into consideration when the
+   * estimated end offset is larger than the current position. Returning {@code Long.MAX_VALUE} as
+   * the estimate implies the largest possible position for the range is {@code Long.MAX_VALUE - 1}.
+   * Return {@code Long.MIN_VALUE} if an estimate can not be provided.
+   *
+   * <p>Providing a good estimate is important for an accurate progress signal and will impact
+   * splitting decisions by the runner.
+   *
+   * <p>If {@link #estimate} is expensive to compute, consider wrapping the implementation with
+   * {@link Suppliers#memoizeWithExpiration} or equivalent as an optimization.
+   *
+   * <p>TODO(BEAM-10032): Also consider using {@link RangeEndEstimator} when the range is not ended
+   * with {@code Long.MAX_VALUE}.
+   */
+  @FunctionalInterface
+  public interface RangeEndEstimator {
+    long estimate();
+  }
+
+  private final RangeEndEstimator rangeEndEstimator;
+
+  public GrowableOffsetRangeTracker(long start, RangeEndEstimator rangeEndEstimator) {
+    super(new OffsetRange(start, Long.MAX_VALUE));
+    this.rangeEndEstimator = checkNotNull(rangeEndEstimator);
+  }
+
+  @Override
+  public SplitResult<OffsetRange> trySplit(double fractionOfRemainder) {
+    // If current tracking range is no longer growable, split it as a normal range.
+    if (range.getTo() != Long.MAX_VALUE || range.getTo() == range.getFrom()) {
+      return super.trySplit(fractionOfRemainder);
+    }
+    // If current range has been done, there is no more space to split.
+    if (lastAttemptedOffset != null && lastAttemptedOffset == Long.MAX_VALUE) {
+      return null;
+    }
+    BigDecimal cur =
+        (lastAttemptedOffset == null)
+            ? BigDecimal.valueOf(range.getFrom()).subtract(BigDecimal.ONE, MathContext.DECIMAL128)
+            : BigDecimal.valueOf(lastAttemptedOffset);
+
+    // Fetch the estimated end offset. If the estimated end is smaller than the next offset, use
+    // the next offset as end.
+    BigDecimal estimateRangeEnd =
+        BigDecimal.valueOf(rangeEndEstimator.estimate())
+            .max(cur.add(BigDecimal.ONE, MathContext.DECIMAL128));
+
+    // Convert to BigDecimal in computation to prevent overflow, which may result in loss of
+    // precision.
+    // split = cur + max(1, (estimateRangeEnd - cur) * fractionOfRemainder)
+    BigDecimal splitPos =
+        cur.add(
+            estimateRangeEnd
+                .subtract(cur, MathContext.DECIMAL128)
+                .multiply(BigDecimal.valueOf(fractionOfRemainder), MathContext.DECIMAL128)
+                .max(BigDecimal.ONE),
+            MathContext.DECIMAL128);
+    long split = splitPos.longValue();
+    if (split > estimateRangeEnd.longValue()) {
+      return null;
+    }
+    OffsetRange res = new OffsetRange(split, range.getTo());
+    this.range = new OffsetRange(range.getFrom(), split);
+    return SplitResult.of(range, res);
+  }
+
+  @Override
+  public Progress getProgress() {
+    // If current tracking range is no longer growable, get progress as a normal range.
+    if (range.getTo() != Long.MAX_VALUE || range.getTo() == range.getFrom()) {
+      return super.getProgress();
+    }
+
+    // Convert to BigDecimal in computation to prevent overflow, which may result in lost of
+    // precision.
+    BigDecimal estimateRangeEnd = BigDecimal.valueOf(rangeEndEstimator.estimate());
+
+    if (lastAttemptedOffset == null) {
+      return Progress.from(
+          0,
+          estimateRangeEnd
+              .subtract(BigDecimal.valueOf(range.getFrom()), MathContext.DECIMAL128)
+              .max(BigDecimal.ZERO)
+              .doubleValue());
+    }
+
+    BigDecimal workRemaining =
+        estimateRangeEnd
+            .subtract(BigDecimal.valueOf(lastAttemptedOffset), MathContext.DECIMAL128)
+            .max(BigDecimal.ZERO);
+    BigDecimal totalWork =
+        estimateRangeEnd
+            .max(BigDecimal.valueOf(lastAttemptedOffset))
+            .subtract(BigDecimal.valueOf(range.getFrom()), MathContext.DECIMAL128);
+    return Progress.from(
+        totalWork.subtract(workRemaining, MathContext.DECIMAL128).doubleValue(),
+        workRemaining.doubleValue());
+  }
+}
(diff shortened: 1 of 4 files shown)


sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/GrowableOffsetRangeTracker.java
```diff
@@ -0,0 +1,103 @@
...
+@Experimental(Kind.SPLITTABLE_DO_FN)
+public class GrowableOffsetRangeTracker extends OffsetRangeTracker {
+  /**
+   * An interface that should be implemented to fetch estimated end offset of range.
+   *
+   * <p>{@code estimateRangeEnd} is called to give te end offset when {@code trySplit} or {@code
+   * getProgress} is invoked. The end offset is exclusive for the range. It's not necessary to
+   * increase monotonically but it's only taken into computation when it's larger than the current
+   * position. When returning Long.MAX_VALUE as estimate, it means the largest possible position for
+   * the range is Long.MAX_VALUE - 1. Having a good estimate is important for providing a good
```
boyuanzz, line 42: Currently I take `Long.MAX_VALUE` as a numeric end offset. But we may also need a notion  to say that the `OffsetPoller` doesn't have a good estimate and still want to keep current range as infinite.  `Long.MAX_VALUE` is not suitable because it's possible that the actual end is `Long.MAX_VALUE`. Do we want to provide a notion here? Like `null`?

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/GrowableOffsetRangeTracker.java
```diff
@@ -0,0 +1,103 @@
...
+            + Math.max(
+                1L,
+                (Double.valueOf((estimateLatestOffset - cur) * fractionOfRemainder)).longValue());
+
+    if (splitPos > estimateLatestOffset) {
+      return null;
+    }
+    OffsetRange res = new OffsetRange(splitPos, range.getTo());
+    this.range = new OffsetRange(range.getFrom(), splitPos);
+    return SplitResult.of(range, res);
```
lukecwik, line 82: Why not construct a new OffsetRangeTracker over `[cur, estimateLatestOffset)` and call trySplit on it?

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/GrowableOffsetRangeTracker.java
```diff
@@ -0,0 +1,103 @@
...
+   * increase monotonically but it's only taken into computation when it's larger than the current
+   * position. When returning Long.MAX_VALUE as estimate, it means the largest possible position for
+   * the range is Long.MAX_VALUE - 1. Having a good estimate is important for providing a good
+   * signal of progress and splitting at a proper position.
+   */
+  public interface OffsetPoller {
+    long estimateRangeEnd();
+  }
+
+  private final OffsetPoller poller;
```
lukecwik, line 49: nit: poller -> estimator

sdks/java/core/src/test/java/org/apache/beam/sdk/transforms/splittabledofn/GrowableOffsetRangeTrackerTest.java
```diff
@@ -0,0 +1,212 @@
...
+    poller.setEstimateRangeEnd(10);
+    SplitResult res = tracker.trySplit(0);
+    tracker.checkDone();
+    assertEquals(new OffsetRange(0, 0), res.getPrimary());
+    assertEquals(new OffsetRange(0, 0), tracker.currentRestriction());
+    assertEquals(new OffsetRange(0, Long.MAX_VALUE), res.getResidual());
+  }
+
+  @Test
+  public void testCheckpointJustStar() throws Exception {
```
lukecwik, line 79:
```suggestion
  public void testCheckpointJustStarted() throws Exception {
```

boyuanzz:
R: @lukecwik 
CC: @robertwb

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/GrowableOffsetRangeTracker.java
boyuanzz: There is a slight difference around comparison of `splitPos` and `end`.  `GrowableOffsetTracker` allows split when `splitPos <= end` but `OffsetRangeTracker` allows split when `splitPos < end`. The purpose of `OffsetRangeTracker` is to reduce resuming from an empty range. The purpose of `GrowableOffsetRangeTracker` is to handle split at `cur + 1`.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/GrowableOffsetRangeTracker.java
```diff
@@ -0,0 +1,103 @@
...
+
+  public GrowableOffsetRangeTracker(long start, OffsetPoller offsetPoller) {
+    super(new OffsetRange(start, Long.MAX_VALUE));
+    this.poller = checkNotNull(offsetPoller);
+  }
+
+  @Override
+  public SplitResult<OffsetRange> trySplit(double fractionOfRemainder) {
+    // If current tracking range is no longer growable, split it as a normal range.
+    if (range.getTo() != Long.MAX_VALUE || range.getTo() == range.getFrom()) {
```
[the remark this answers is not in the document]
boyuanzz: Yes `super.trySplit()` handles empty range. The special case here is to handle `range.getTo() == range.getFrom == Long.MAX_VALUE`.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/GrowableOffsetRangeTracker.java
boyuanzz: btw we will also loss `lastAttemptedOffset` by creating a new tracker.

boyuanzz:
Latest changes are for addressing comments and using double during computation. @lukecwik PTAL. Thanks for your help!

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/GrowableOffsetRangeTracker.java
```diff
@@ -0,0 +1,115 @@
...
+      return super.trySplit(fractionOfRemainder);
+    }
+    // If current range has been done, there is no more space to split.
+    if (lastAttemptedOffset != null && lastAttemptedOffset == Long.MAX_VALUE) {
+      return null;
+    }
+    double cur =
```
lukecwik, line 76: We should only use doubles if there would be an issue with overflow. For really large long values we can have the math be incorrect which can lead to the splitPos being before from or after to even though with higher precision we would get something that works. For example:
```
    long a = 123456789012345677L;
    long b = 123456789012345679L;
    long c = 123456789012345680L;
    double ad = a;
    double bd = b;
    double cd = c;
    System.out.println((long)ad + " " + (long)bd + " " + (long)cd);
```
prints
```
123456789012345680 123456789012345680 123456789012345680
```
Worthwhile to have a test which exercises a case where `from = 123456789012345677L` and `to = 123456789012345679L` and we can have a split in between. Also worthwhile to test `from = 123456789012345681L` and `to = 123456789012345683L`. All these values (and the values in between) when converted to double and back to long round to `123456789012345680L`.

We can solve this by having two ways to calculate this or using BigDecimal. The two ways to calculate method would use longs when (estimateRangeEnd - cur) won't overflow and doubles otherwise. Using BigDecimal with a MathContext with enough precision wouldn't run into this either and would only require one implemention (I believe DECIMAL128 should have enough precision to not run into this issue).

lukecwik:
Should we be using the RangeEndEstimator when providing progress/splitting for ranges not ending at `Long.MAX_VALUE`?

Lets say the range estimate is bad and is `MAX_VALUE - 3` but the real end is `5000`, then after a split we end up with `[0, (MAX_VALUE - 3) * 0.5)` and `[(MAX_VALUE - 3) * 0.5, MAX_VALUE)`. We may quickly learn that the residual is empty and then lose all effective progress on the primary.

boyuanzz:
> Should we be using the RangeEndEstimator when providing progress/splitting for ranges not ending at `Long.MAX_VALUE`?
> 
> Lets say the range estimate is bad and is `MAX_VALUE - 3` but the real end is `5000`, then after a split we end up with `[0, (MAX_VALUE - 3) * 0.5)` and `[(MAX_VALUE - 3) * 0.5, MAX_VALUE)`. We may quickly learn that the residual is empty and then lose all effective progress on the primary.

I can see the benefit of using `RangeEndEstimator` for the finite range here.  But as long as we don't modify the range end to estimate end or use estimate ed in `tryClaim`, we still cannot say the residual is empty.

lukecwik:
> > Should we be using the RangeEndEstimator when providing progress/splitting for ranges not ending at `Long.MAX_VALUE`?
> > Lets say the range estimate is bad and is `MAX_VALUE - 3` but the real end is `5000`, then after a split we end up with `[0, (MAX_VALUE - 3) * 0.5)` and `[(MAX_VALUE - 3) * 0.5, MAX_VALUE)`. We may quickly learn that the residual is empty and then lose all effective progress on the primary.
> 
> I can see the benefit of using `RangeEndEstimator` for the finite range here. But as long as we don't modify the range end to estimate end or use estimate ed in `tryClaim`, we still cannot say the residual is empty.

That is true but I was thinking it would make better splitting decisions instead of creating a bunch of empty splits trimming the range down. The advantage of not using the estimator is that we don't have to invoke since it could be expensive for the user and in many situations will produce a value greater than `to`.

boyuanzz:
> > > Should we be using the RangeEndEstimator when providing progress/splitting for ranges not ending at `Long.MAX_VALUE`?
> > > Lets say the range estimate is bad and is `MAX_VALUE - 3` but the real end is `5000`, then after a split we end up with `[0, (MAX_VALUE - 3) * 0.5)` and `[(MAX_VALUE - 3) * 0.5, MAX_VALUE)`. We may quickly learn that the residual is empty and then lose all effective progress on the primary.
> > 
> > 
> > I can see the benefit of using `RangeEndEstimator` for the finite range here. But as long as we don't modify the range end to estimate end or use estimate ed in `tryClaim`, we still cannot say the residual is empty.
> 
> That is true but I was thinking it would make better splitting decisions instead of creating a bunch of empty splits trimming the range down. The advantage of not using the estimator is that we don't have to invoke since it could be expensive for the user and in many situations will produce a value greater than `to`.
> 
> We can leave it out for now unless some compelling use case comes up.

Do we want to have a TODO here to track this?

lukecwik:
> > > > Should we be using the RangeEndEstimator when providing progress/splitting for ranges not ending at `Long.MAX_VALUE`?
> > > > Lets say the range estimate is bad and is `MAX_VALUE - 3` but the real end is `5000`, then after a split we end up with `[0, (MAX_VALUE - 3) * 0.5)` and `[(MAX_VALUE - 3) * 0.5, MAX_VALUE)`. We may quickly learn that the residual is empty and then lose all effective progress on the primary.
> > > 
> > > 
> > > I can see the benefit of using `RangeEndEstimator` for the finite range here. But as long as we don't modify the range end to estimate end or use estimate ed in `tryClaim`, we still cannot say the residual is empty.
> > 
> > 
> > That is true but I was thinking it would make better splitting decisions instead of creating a bunch of empty splits trimming the range down. The advantage of not using the estimator is that we don't have to invoke since it could be expensive for the user and in many situations will produce a value greater than `to`.
> > We can leave it out for now unless some compelling use case comes up.
> 
> Do we want to have a TODO here to track this?

Sure.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/GrowableOffsetRangeTracker.java
boyuanzz: Using `BigDecimal` in the latest revision. Thanks for your help!

boyuanzz:
Run Java PreCommit

lukecwik:
Run Java PreCommit

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/GrowableOffsetRangeTracker.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    BigDecimal workRemaining =
+        estimateRangeEnd
+            .subtract(BigDecimal.valueOf(lastAttemptedOffset), MathContext.DECIMAL128)
+            .max(BigDecimal.ZERO);
+    BigDecimal totalWork =
+        estimateRangeEnd
+            .max(BigDecimal.valueOf(lastAttemptedOffset))
+            .subtract(BigDecimal.valueOf(range.getFrom()));
```
lukecwik, line 138: Doesn't this subtract need the DECIMAL128 math context?

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/OffsetRangeTracker.java
```diff
@@ -120,13 +136,27 @@ public String toString() {
...
 
     // Compute the amount of work remaining from where we are to where we are attempting to get to
     // with a minimum of zero in case we have claimed beyond the end of the range.
-    long workRemaining = Math.max(range.getTo() - lastAttemptedOffset, 0);
-    return Progress.from(range.getTo() - range.getFrom() - workRemaining, workRemaining);
+    BigDecimal workRemaining =
+        BigDecimal.valueOf(range.getTo())
+            .subtract(BigDecimal.valueOf(lastAttemptedOffset), MathContext.DECIMAL128)
+            .max(BigDecimal.ZERO);
+    BigDecimal wholeWork =
```
lukecwik, line 155: nit: `wholeWork` -> `totalWork`

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/GrowableOffsetRangeTracker.java
boyuanzz: It's needed as well.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/OffsetRangeTracker.java
boyuanzz: Done. Thanks!

boyuanzz:
Thanks for your help!

iemejia:
Now that this is merged. Can I ask a question. What is the intrinsic limitation that did not allow old `OffsetRangeTracker` to be refactored for this use case? or why we want to have both?

Does this mean also that we might need `GrowableBytekeyRangeTracker` and basically 'dynamic' versions for every `RestrictionTracker` ?

boyuanzz:
> What is the intrinsic limitation that did not allow old `OffsetRangeTracker` to be refactored for this use case? or why we want to have both?
> 
`GrowableOffsetRangeTracker` and `OffsetRangeTracker` should be applied to different kind of `OffsetRange`. `GrowableOffsetRangeTracker` is for the `OffsetRange` that the end could be changed during execution time, which mostly happens in streaming case. `OffsetRangeTracker` is for the range that we know what the exact end is, which is perfect for batch.  The reason that I didn't make them into one is because I don't want to introduce additional complexity to `OffsetRangeTracker`.  It's doable to have the dynamic one as general case where the range with fixed end is a special case. But I want to make them specifically with less confusion.

> Does this mean also that we might need `GrowableBytekeyRangeTracker` and basically 'dynamic' versions for every `RestrictionTracker` ?

I think it will depend on the actual usage. If we have an application scenario that requires for a dynamic version, then I would say yes.

iemejia:
Thanks for answering and for the clear explaination @boyuanzz 

I would have tended towards having the additional complexity in `OffsetRangeTracker` just because it is the de-facto reference, but I understand the different preference.

Nice to see this an the Kafka SDF happening, congrats!

PR #11725, opened 2020-05-15
[BEAM-10015] Fix output timestamp on dataflow runner

reuvenlax:
This issue is fixed by encoding the output timestamp in the timer tag so that we can recover it when firing the timer.

We also remove the old TimerData.of builders, as their accidental use caused the original bug.


reuvenlax:
run dataflow validatesrunner

reuvenlax:
R: @rehmanmuradali

reuvenlax:
run dataflow validatesrunner

rehmanmuradali:
LGTM. Thanks for taking care of it.

PR #11744, opened 2020-05-18
[BEAM-10007] Handle ValueProvider pipeline options in PortableRunner

TheNeuralBit:
R: @aaltay 

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/runners/portability/portable_runner.py
+++ b/sdks/python/apache_beam/runners/portability/portable_runner.py
@@ -40,6 +40,7 @@ from apache_beam.options.pipeline_options import DebugOptions
 from apache_beam.options.pipeline_options import PortableOptions
 from apache_beam.options.pipeline_options import SetupOptions
 from apache_beam.options.pipeline_options import StandardOptions
+from apache_beam.options.value_provider import ValueProvider
 from apache_beam.portability import common_urns
 from apache_beam.portability.api import beam_artifact_api_pb2_grpc
 from apache_beam.portability.api import beam_job_api_pb2
@@ -163,10 +164,19 @@ class JobServiceHandle(object):
     all_options = self.options.get_all_options(
         add_extra_args_fn=add_runner_options,
         retain_unknown_options=self._retain_unknown_options)
+
+    def convert_pipeline_option_value(v):
+      # convert int values: BEAM-5509
+      if type(v) == int:
+        return str(v)
+      elif isinstance(v, ValueProvider):
+        return convert_pipeline_option_value(
+            v.get()) if v.is_accessible() else None
+      return v
+
     # TODO: Define URNs for options.
-    # convert int values: https://issues.apache.org/jira/browse/BEAM-5509
     p_options = {
-        'beam:option:' + k + ':v1': (str(v) if type(v) == int else v)
+        'beam:option:' + k + ':v1': convert_pipeline_option_value(v)
         for k,
         v in all_options.items() if v is not None
     }
--- a/sdks/python/apache_beam/runners/portability/fn_api_runner/fn_runner_test.py
+++ b/sdks/python/apache_beam/runners/portability/fn_api_runner/fn_runner_test.py
@@ -54,6 +54,7 @@ from apache_beam.metrics.execution import MetricKey
 from apache_beam.metrics.metricbase import MetricName
 from apache_beam.options.pipeline_options import DebugOptions
 from apache_beam.options.pipeline_options import PipelineOptions
+from apache_beam.options.value_provider import RuntimeValueProvider
 from apache_beam.runners.portability import fn_api_runner
 from apache_beam.runners.portability.fn_api_runner import fn_runner
 from apache_beam.runners.sdf_utils import RestrictionTrackerView
@@ -830,6 +831,21 @@ class FnApiRunnerTest(unittest.TestCase):
           | beam.ParDo(SyntheticSDFAsSource())
           | beam.combiners.Count.Globally())
       assert_that(res, equal_to([total_num_records]))
+
+  def test_create_value_provider_pipeline_option(self):
+    # Verify that the runner can execute a pipeline when there are value
+    # provider pipeline options
+    # pylint: disable=unused-variable
+    class FooOptions(PipelineOptions):
+      @classmethod
+      def _add_argparse_args(cls, parser):
+        parser.add_value_provider_argument(
+            "--foo", help='a value provider argument', default="bar")
+
+    RuntimeValueProvider.set_runtime_options({})
+
+    with self.create_pipeline() as p:
+      assert_that(p | beam.Create(['a', 'b']), equal_to(['a', 'b']))
 
 
 # These tests are kept in a separate group so that they are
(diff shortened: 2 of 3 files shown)


sdks/python/apache_beam/runners/portability/portable_runner.py
```diff
@@ -164,10 +165,19 @@ def add_runner_options(parser):
     all_options = self.options.get_all_options(
         add_extra_args_fn=add_runner_options,
         retain_unknown_options=self._retain_unknown_options)
+
     # TODO: Define URNs for options.
     # convert int values: https://issues.apache.org/jira/browse/BEAM-5509
+    def convert_pipeline_option_value(v):
+      if type(v) == int:
```
aaltay, line 172: why there is a special case for int? Can we do str(v) for everything except for valueproviders?

TheNeuralBit: Good question. I was just maintaining the status quo and adding a special case for `ValueProvider`. Let me see if I can dig up why the str(v) is there for ints

TheNeuralBit: Oh the relevant jira is linked right there. It seems that this is really a workaround for a bug in json_format where ints and floats aren't treated differently. I can look into fixing that upstream so we can get rid of the special case here.

I don't think it would be good to call str(v) on everything

aaltay: Interesting. For this PR, could you move that comment closer to the if type(v) == int line.

> I can look into fixing that upstream so we can get rid of the special case here.
If you have time, maybe file a bug. I do not think this is a very high priority for us.

> I don't think it would be good to call str(v) on everything
I agree. I was trying to understand why are we doing this.

TheNeuralBit:
Run Python PreCommit

TheNeuralBit:
Somehow this seems to introduce an additional flake in `SdkWorkerMainTest`. I can't replicate it locally... I suspect it only occurs when pipeline options get polluted from other tests.

aaltay:
> Somehow this seems to introduce an additional flake in `SdkWorkerMainTest`. I can't replicate it locally... I suspect it only occurs when pipeline options get polluted from other tests.

What is the error?

TheNeuralBit:
```
self = <apache_beam.runners.worker.sdk_worker_main_test.SdkWorkerMainTest testMethod=test_parse_pipeline_options>

    def test_parse_pipeline_options(self):
      expected_options = PipelineOptions([])
      expected_options.view_as(
          SdkWorkerMainTest.MockOptions).m_m_option = ['beam_fn_api']
      expected_options.view_as(
          SdkWorkerMainTest.MockOptions).m_option = '/tmp/requirements.txt'
      self.assertEqual(
          expected_options.get_all_options(),
          sdk_worker_main._parse_pipeline_options(
              '{"options": {' + '"m_option": "/tmp/requirements.txt", ' +
>             '"m_m_option":["beam_fn_api"]' + '}}').get_all_options())
E     AssertionError: {'inf[2214 chars]7fd7c06bc278>, 'environment_config': None, 'in[616 chars]None} != {'inf[2214 chars]7fd7c10a35c0>, 'environment_config': None, 'in[616 chars]None}
E     Diff is 3496 characters long. Set self.maxDiff to None to see it.

apache_beam/runners/worker/sdk_worker_main_test.py:70: AssertionError
```

I haven't looked into the test any further, but it looks like a memory address is getting pulled into the options and not matching.

aaltay:
> test_parse_pipeline_options

This test is a bit flawed. `expected_options.get_all_options()` could return all registered options. Test is assuming that no other option related tests has run.

The main flaw is still in our subclass based option discovery.

TheNeuralBit:
Ah yeah I could change the test to just verify the options that are being set there (either by dropping default values in the get_all_options call, or by adding explicit asserts)

TheNeuralBit:
Run Python PreCommit

aaltay:
Run Python PreCommit

aaltay:
Run Portable_Python PreCommit

TheNeuralBit:
Looks like the CI failure is actually another flake due to BEAM-10006. It can be replicated reliably locally with

```
$ pytest apache_beam/runners/portability/fn_api_runner/fn_runner_test.py::FnApiRunnerTest::test_create_value_provider_pipeline_option apache_beam/runners/worker/sdk_worker_main_test.py
```

TheNeuralBit:
Oh I see now. I tried to fix this issue by adding `drop_defaults=true` but we explicitly don't drop the default value for ValueProvider arguments: https://github.com/apache/beam/blob/1fbc55e7819187018c2cb3cb295eb84e6f348d8a/sdks/python/apache_beam/options/pipeline_options.py#L319-L320

I'm not really sure why that is but it's been there for three years so I'm a little hesitant to remove it... I just pushed 982e087 that does this, let's see if it breaks anything in CI or Google tests.

I'll check back in the morning, if it breaks anything (or if anyone tells me this is a bad idea), I can re-write the test to only assert on specific options.

aaltay:
> Oh I see now. I tried to fix this issue by adding `drop_defaults=true` but we explicitly don't drop the default value for ValueProvider arguments:
> 
> https://github.com/apache/beam/blob/1fbc55e7819187018c2cb3cb295eb84e6f348d8a/sdks/python/apache_beam/options/pipeline_options.py#L319-L320
> 
> I'm not really sure why that is but it's been there for three years so I'm a little hesitant to remove it... I just pushed [982e087](https://github.com/apache/beam/commit/982e087fd09509277057b76991598cad3e6acf8e) that does this, let's see if it breaks anything in CI or Google tests.
> 
> I'll check back in the morning, if it breaks anything (or if anyone tells me this is a bad idea), I can re-write the test to only assert on specific options.

I do not remember the reason for this. I am looking.

@tvalentyn - do you remember?

aaltay:
> > Oh I see now. I tried to fix this issue by adding `drop_defaults=true` but we explicitly don't drop the default value for ValueProvider arguments:
> > https://github.com/apache/beam/blob/1fbc55e7819187018c2cb3cb295eb84e6f348d8a/sdks/python/apache_beam/options/pipeline_options.py#L319-L320
> > 
> > I'm not really sure why that is but it's been there for three years so I'm a little hesitant to remove it... I just pushed [982e087](https://github.com/apache/beam/commit/982e087fd09509277057b76991598cad3e6acf8e) that does this, let's see if it breaks anything in CI or Google tests.
> > I'll check back in the morning, if it breaks anything (or if anyone tells me this is a bad idea), I can re-write the test to only assert on specific options.
> 
> I do not remember the reason for this. I am looking.
> 
> @tvalentyn - do you remember?

I _believe_ default value of RuntimeValueProviders are actually user provided non-default values. (If I am reading this correctly: https://github.com/apache/beam/blob/1fbc55e7819187018c2cb3cb295eb84e6f348d8a/sdks/python/apache_beam/options/pipeline_options.py#L115)

TheNeuralBit:
Removing the ValueProvider line didn't actually break any Google tests, but I lost my resolve to remove it. It seems likely it would break some untested behavior, e.g. display_data uses drop_defaults=True:
https://github.com/apache/beam/blob/f65a18760a07a48c11fe4aff2a48a845df1f522d/sdks/python/apache_beam/options/pipeline_options.py#L328-L329

Instead I pushed d2d4ecb which makes the assertions in sdk_worker_main_test more specific. I think this is good to go assuming CI passses.

PR #11754, opened 2020-05-20
[BEAM-10037] BeamSqlExample.java fails to build 

omarismail94:
R: @TheNeuralBit 

In the `BeamSqlExample.java` class, the instructions state that to run the example, use: 

`./gradlew :sdks:java:extensions:sql:runBasicExample`. 

I tried this and the build failed due to `java.lang.IllegalStateException: Unable to return a default Coder`

I fixed this by setting the Coder for both anon transforms. 

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/example/BeamSqlExample.java
+++ b/sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/example/BeamSqlExample.java
@@ -18,6 +18,7 @@
 package org.apache.beam.sdk.extensions.sql.example;
 
 import org.apache.beam.sdk.Pipeline;
+import org.apache.beam.sdk.coders.RowCoder;
 import org.apache.beam.sdk.extensions.sql.SqlTransform;
 import org.apache.beam.sdk.options.PipelineOptions;
 import org.apache.beam.sdk.options.PipelineOptionsFactory;
@@ -45,6 +46,7 @@ import org.apache.beam.sdk.values.TupleTag;
  * Beam documentation on how to run pipelines.
  */
 class BeamSqlExample {
+
   public static void main(String[] args) {
     PipelineOptions options = PipelineOptionsFactory.fromArgs(args).create();
     Pipeline p = Pipeline.create(options);
@@ -66,19 +68,21 @@ class BeamSqlExample {
         inputTable.apply(SqlTransform.query("select c1, c2, c3 from PCOLLECTION where c1 > 1"));
 
     // print the output record of case 1;
-    outputStream.apply(
-        "log_result",
-        MapElements.via(
-            new SimpleFunction<Row, Row>() {
-              @Override
-              public Row apply(Row input) {
-                // expect output:
-                //  PCOLLECTION: [3, row, 3.0]
-                //  PCOLLECTION: [2, row, 2.0]
-                System.out.println("PCOLLECTION: " + input.getValues());
-                return input;
-              }
-            }));
+    outputStream
+        .apply(
+            "log_result",
+            MapElements.via(
+                new SimpleFunction<Row, Row>() {
+                  @Override
+                  public Row apply(Row input) {
+                    // expect output:
+                    //  PCOLLECTION: [3, row, 3.0]
+                    //  PCOLLECTION: [2, row, 2.0]
+                    System.out.println("PCOLLECTION: " + input.getValues());
+                    return input;
+                  }
+                }))
+        .setCoder(RowCoder.of(type));
 
     // Case 2. run the query with SqlTransform.query over result PCollection of case 1.
     PCollection<Row> outputStream2 =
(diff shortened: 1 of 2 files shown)


sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/example/BeamSqlExample.java
```diff
@@ -66,38 +68,47 @@ public static void main(String[] args) {
...
+                  @Override
+                  public Row apply(Row input) {
+                    // expect output:
+                    //  PCOLLECTION: [3, row, 3.0]
+                    //  PCOLLECTION: [2, row, 2.0]
+                    System.out.println("PCOLLECTION: " + input.getValues());
+                    return input;
+                  }
+                }))
+        .setCoder(RowCoder.of(type));
```
TheNeuralBit, line 85: could you change this to `withRowSchema(type)`? It does the same thing, but it's less verbose

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/example/BeamSqlExample.java
```diff
@@ -66,38 +68,47 @@ public static void main(String[] args) {
...
+                    System.out.println("CASE1_RESULT: " + input.getValues());
+                    return input;
+                  }
+                }))
+        .setCoder(
+            RowCoder.of(
+                Schema.builder()
+                    .addStringField("stringField")
+                    .addDoubleField("doubleField")
+                    .build()));
```
TheNeuralBit, line 111: Here as well

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/example/BeamSqlExample.java
omarismail94: I can do that. I did `setRowSchema(type)` and it worked!

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/example/BeamSqlExample.java
omarismail94: I tried `setRowSchema(type) and it failed with `: java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Integer`.

I think it is inferring the schema as 3 fields, but the result only returns fields, and that's why it throws the error

omarismail94:
> Thank you @omarismail94!
> 
> We should probably be running this continuously to make sure we don't break it again. Would you mind adding the gradle task for this to the SQL preCommit [here](https://github.com/apache/beam/blob/master/build.gradle#L154)? That way it will run before we merge any PR that affects SQL.

I will add both this and runPojoExample to SQL preCommit

sdks/java/extensions/sql/src/main/java/org/apache/beam/sdk/extensions/sql/example/BeamSqlExample.java
omarismail94: This is part of the Stack trace that makes me think that
```
Caused by: java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Integer
        at org.apache.beam.sdk.coders.VarIntCoder.encode(VarIntCoder.java:33)
        at org.apache.beam.sdk.coders.RowCoderGenerator$EncodeInstruction.encodeDelegate(RowCoderGenerator.java:270)
        at org.apache.beam.sdk.coders.Coder$ByteBuddy$E99UrF3W.encode(Unknown Source)
        at org.apache.beam.sdk.coders.Coder$ByteBuddy$E99UrF3W.encode(Unknown Source)
        at org.apache.beam.sdk.schemas.SchemaCoder.encode(SchemaCoder.java:115)
```

TheNeuralBit: oh for this call you will need to use
```
Schema.builder()
￼                    .addStringField("stringField")
￼                    .addDoubleField("doubleField")
￼                    .build()
```
like you had in the setCoder call

omarismail94: Actually, it is not due to the reduction in the number of field, but the order in which the fields are printed. Here is the order it expects

* Int, String, Double

and the fields that represent those types are: c1, c2, c3

If your results print out of order, it fails due to the `ClassCastException`. I tried doing this query and it failed:
`select  c2, sum(c1), sum(c3) from CASE1_RESULT group by c2`,

but if I do 
`select  sum(c1),c2, sum(c3) from CASE1_RESULT group by c2` 

it works! You can see that in the one that failed, c1 and c2s positions have switched, so the encoder trips out. What's cool is that you can see the results correctly calculated in:
` System.out.println("CASE1_RESULT: " + input.getValues());`

but it seems that when the result is encoded, the program throws an error due to the results being out of order

TheNeuralBit:
retest this please

TheNeuralBit:
Run Java PreCommit

TheNeuralBit:
Run Java PreCommit

TheNeuralBit:
I'm going to go ahead and merge since CI failure is unrelated

PR #11792, opened 2020-05-22
[BEAM-10435] Add ValidatesRunner task for local_job_service and Java SDK harness

kennknowles:
This adds a ValidatesRunner suite for the Java SDK against the local Python ULR.

------------------------

 - [n/a] Update `CHANGES.md` with noteworthy changes.

--- | --- | --- | --- | ---

### DIFF
--- a/runners/portability/java/build.gradle
+++ b/runners/portability/java/build.gradle
@@ -1,3 +1,5 @@
+import groovy.json.JsonOutput
+
 /*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
@@ -18,10 +20,15 @@
 
 plugins { id 'org.apache.beam.module' }
 applyJavaNature(automaticModuleName: 'org.apache.beam.runners.portability', enableChecker:false)
+applyPythonNature()
 
 description = "Apache Beam :: Runners :: Portability :: Java"
 ext.summary = """A Java implementation of the Beam Model which utilizes the portability
 framework to execute user-definied functions."""
+
+configurations {
+  validatesRunner
+}
 
 dependencies {
   compile library.java.vendored_guava_26_0_jre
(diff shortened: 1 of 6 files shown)


runners/portability/java/build.gradle
```diff
@@ -31,9 +45,123 @@ dependencies {
...
+
+void execInVirtualenv(String... args) {
+  String shellCommand = ". ${virtualenvDir}/bin/activate && " + args.collect { arg -> "'" + arg.replaceAll("'", "\\'") + "'" }.join(" ")
+  exec {
+    workingDir pythonSdkDir
+    commandLine "sh", "-c", shellCommand
+  }
 }
+
+void execBackgroundInVirtualenv(String... args) {
```
kennknowles, line 79: This verbosity is because Gradle exec tasks do not terminate when there are open subprocesses.

runners/portability/java/build.gradle
```diff
@@ -31,9 +45,123 @@ dependencies {
...
+  String shellCommand = ". ${virtualenvDir}/bin/activate && " + args.collect { arg -> "'" + arg.replaceAll("'", "\\'") + "'" }.join(" ")
+  println "execBackgroundInVirtualEnv: ${shellCommand}"
+  ProcessBuilder pb = new ProcessBuilder().redirectErrorStream(true).directory(new File(pythonSdkDir)).command(["sh", "-c", shellCommand])
+  Process proc = pb.start();
+
+  // redirectIO does not work for connecting to groovy/gradle stdout
+  BufferedReader reader = new BufferedReader(new InputStreamReader(proc.getInputStream()));
```
kennknowles, line 86: This verbosity is because ProcessBuilder.inheritIO does not work in this context (don't know if it is Gradle or Groovy breaking it)

runners/portability/java/build.gradle
```diff
@@ -31,9 +45,123 @@ dependencies {
...
+  String line
+  while ((line = reader.readLine()) != null) {
+    println line
+  }
+  proc.waitFor();
+}
+
+task virtualenv {
```
kennknowles, line 94: I tried registering `outputs.dir virtualenvDir` and `inputs.dir pythonSdkDir` but it did not result in incremental build.

runners/portability/java/build.gradle
```diff
@@ -1,3 +1,13 @@
+import groovy.json.JsonOutput
+
+import java.nio.file.FileSystems
```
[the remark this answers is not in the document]
kennknowles: Great, I'll remove that. I think in an early draft I made the gradle task not wait for the process to terminate (because the daemonized process was causing the hang, but that turned out to be a different configuration)

runners/portability/java/build.gradle
kennknowles: I didn't use them because I was attempting to do more proper idempotent/cached gradle tasks.

kennknowles: I hey now I recall - `applyPythonNature` causes conflicting configurations. Really would like to move away from these nature things.

kennknowles:
Once I get the suite running I can do some refactors to try to share with `setupVirtualEnv` and clear out the extraneous sleepwaiting and whatnot.

runners/portability/java/build.gradle
lukecwik: I wouldn't want Beam to move to a build setup where each gradle file does its own thing because the fragmentation will hurt debugging build issues and slow down rolling out build changes that impact more then one project.

One example where we decided to split a common setup was between releasing java projects and releasing vendored projects which lead to fixes that weren't done in both places leading to bugs that lasted for months.

kennknowles: I'm not suggesting moving each `build.gradle` to do their own thing. Beam Java modules and vendored libraries can share as much code as they like. They shouldn't share an entrypoint because they are different things.

In this case, the code suggested to be shared is coupled with creating a Beam Python module, which this is not. "Do Python things" is not an adequate or meaningful abstraction. Applying the vague blanket logic is a liability, even if it worked here, which it does not. It is likely that I can do some tweaks to `applyPythonNature` to "make it work", but that would be bad engineering.

 - Adding near-duplicate code when maybe there is an abstraction: tech debt
 - Adding a dependency on something that isn't ready/meant for it: tech debt

I interpret Robert's comment as an invitation to improve our build code into some kind of meaningful abstraction that can be shared without incurring yet more tech debt.

kennknowles: (I will explore this invitation once the tests are running properly)

kennknowles: Adding a dependency on something that _is_ ready and _is_ meant for it, but is not polished and named to make clear that it is a _logical necessity_: tech debt)

kennknowles:
Found a lot of exclusions before I started getting my disks filled by the build & local runner.

@robertwb you may be interested in the failures where it seems empty side inputs don't work

kennknowles:
PTAL. I have a final set of exclusions, categorized, with all other tests passing locally. The local run took about an hour. I have adjusted `applyPythonNature` to be compatible with `applyJavaNature` and reused the code despite my misgivings. I did not refactor the build support code.

kennknowles:
I have curated the commits into orthogonal changes, if you wish to focus your review.

kennknowles:
https://ci-beam.apache.org/job/beam_PreCommit_Python_Commit/13653/ appears to be a failure in unrelated code. It does not look like something that would flake. Perhaps `master` got broken.

kennknowles:
Run Python PreCommit

runners/portability/java/build.gradle
```diff
@@ -31,9 +38,161 @@ dependencies {
...
+  dependsOn setupVirtualenv
+
+  doLast {
+    execBackgroundInVirtualenv "python",
+        "-m", "apache_beam.runners.portability.local_job_service_main",
+        "--background",
+        "--stdout_file=${localJobServiceStdoutFile}",
+        "--pid_file=${localJobServicePidFile}",
+        "--port_file=${localJobServicePortFile}"
+//
```
robertwb, line 95: These lines can be removed now, right?

runners/portability/java/build.gradle
```diff
@@ -31,9 +38,161 @@ dependencies {
...
+startLocalJobService.finalizedBy stopLocalJobService
+
+/**
+ * Runs Java ValidatesRunner tests against the Universal Local Runner (ULR) aka local_job_service_main
+ * with subprocess SDK harness environments.
+ */
+task ulrValidatesRunnerTests(type: Test) {
+  dependsOn ":sdks:java:container:docker"
+
+  if (!project.hasProperty("localJobServicePortFile")) {
```
robertwb, line 126: Very convenient.

runners/portability/java/build.gradle
```diff
@@ -31,9 +38,161 @@ dependencies {
...
+  }
+
+  group = "Verification"
+  description "PortableRunner Java subprocess ValidatesRunner suite"
+  classpath = configurations.validatesRunner
+  systemProperty "beamTestPipelineOptions", JsonOutput.toJson([
+      "--runner=TestUniversalRunner",
+      "--experiments=beam_fn_api",
+      "--localJobServicePortFile=${localJobServicePortFile}"
+  ])
```
robertwb, line 137: Consider passing `--defaultEnvironmentType=LOOPBACK`. You can then remove the docker dependency as well. (Maybe we could run one test with docker, but all of them seems overkill and expensive.)

runners/portability/java/build.gradle
```diff
@@ -31,9 +38,161 @@ dependencies {
...
+    includeCategories 'org.apache.beam.sdk.testing.ValidatesRunner'
+    excludeCategories 'org.apache.beam.sdk.testing.UsesGaugeMetrics'
+    excludeCategories 'org.apache.beam.sdk.testing.UsesOnWindowExpiration'
+    excludeCategories 'org.apache.beam.sdk.testing.UsesMapState'
+    excludeCategories 'org.apache.beam.sdk.testing.UsesSetState'
+    excludeCategories 'org.apache.beam.sdk.testing.UsesTestStreamWithMultipleStages'
+    excludeCategories 'org.apache.beam.sdk.testing.UsesTimersInParDo'
+  }
+  filter {
+    // There is not currently a category for excluding these _only_ in committed mode
```
robertwb, line 149: Could you create JIRAs for these (at whatever granularity seems appropriate) and add them here?

runners/portability/java/src/main/java/org/apache/beam/runners/portability/testing/TestUniversalRunner.java
```diff
@@ -0,0 +1,119 @@
...
+      portableOptions.setJobEndpoint(endpoint);
+      portableOptions.setRunner(PortableRunner.class);
+      PortableRunner runner = PortableRunner.fromOptions(portableOptions);
+      PipelineResult result = runner.run(pipeline);
+      assertThat(
+          "Pipeline did not succeed.",
+          result.waitUntilFinish(),
+          Matchers.is(PipelineResult.State.DONE));
+      return result;
+    } catch (IOException e) {
```
robertwb, line 84: Nit: I'd make this more local (it's thrown only at `Files.readAllBytes`).

runners/portability/java/src/main/java/org/apache/beam/runners/portability/testing/TestUniversalRunner.java
```diff
@@ -0,0 +1,119 @@
...
+      throw new RuntimeException(e);
+    }
+  }
+
+  public interface Options extends TestPipelineOptions, PortablePipelineOptions {
+    /**
+     * A file containing the job service port, since Gradle needs to know this filename statically
+     * to provide it in Beam testing options.
+     */
+    @Description("File containing local job service port.")
```
robertwb, line 94: Logically, it would make sense to let this be optional (e.g. one could instead provide jobEndpoint directly). A point could be made that the testing infrastructure should be the one reading the file and setting jobEndpoint, rather than passing the file path as an option (but I don't know how much messier that'd make things).

sdks/python/apache_beam/runners/portability/local_job_service_main.py
```diff
@@ -99,11 +104,23 @@ def run(argv):
       options.port_file = os.path.splitext(options.pid_file)[0] + '.port'
       argv.append('--port_file')
       argv.append(options.port_file)
+
+    if not options.stdout_file:
+      raise RuntimeError('--stdout_file must be specified with --background')
+    stdout_dest = open(options.stdout_file, mode='w')
```
robertwb, line 110: I think I mentioned this before, but is it an issue that these file descriptors might get closed on completion of this process?

runners/portability/java/build.gradle
kennknowles: Done. Also added a new Jira component `runner-universal` since I did not find one, in case there's a need to search for these.

runners/portability/java/src/main/java/org/apache/beam/runners/portability/testing/TestUniversalRunner.java
kennknowles: Done

runners/portability/java/src/main/java/org/apache/beam/runners/portability/testing/TestUniversalRunner.java
kennknowles: Yea, this sort of thing is what took 90% of the time for this PR actually. Scraping around Gradle's docs and the internet for ways to insert that little bit of logic, because realizing it was sort of against the grain. Pipeline options are passed as a Java system property, and those are set up in the Gradle graph construction phase. More generally, there's not a Gradle graph execution-time slot for free-form code that also re-uses the Test task type. Perhaps they expect you to use inheritance and make a new Task type. Which I would rather not do ;_;

It would be fine to have two pipeline options, so that simple use could be simple.

sdks/python/apache_beam/runners/portability/local_job_service_main.py
kennknowles: Yea I think my rebasing clobbered that thread. It is not an issue. Parent file descriptors are not closed. You can find some links I think on the PR front page it will still have the prior conversation.

(I won't rebase from here on out, until review is done)

runners/portability/java/build.gradle
robertwb: Ack. I don't think it makes sense to run every test to verify the container/harness setup works (this is reminiscent of other threads) but definitely agree these choices can be postponed while we get this PR in.

runners/portability/java/src/main/java/org/apache/beam/runners/portability/testing/TestUniversalRunner.java
robertwb: We already have the plain `jobEndpoint` option, just make (re)setting it conditional on `LocalJobServicePortFile` being set.

sdks/python/apache_beam/runners/portability/local_job_service_main.py
robertwb: No problem. My comment was from something like a mont ago. Thanks for the references.

kennknowles:
Flakes in this run:

 - https://issues.apache.org/jira/browse/BEAM-10474 failure in py DisplayDataTest

And then I just introduced an actual compile error in my fixups.

runners/portability/java/src/main/java/org/apache/beam/runners/portability/testing/TestUniversalRunner.java
kennknowles: Done

kennknowles:
On the portable precommit, 4 gradle scans succeeded and then the Gradle daemon crashed.

kennknowles:
Run Portable_Python PreCommit

kennknowles:
(it could have been interrupted, I didn't dig too deep)

kennknowles:
Flakes this time in Java precommit:

 - https://issues.apache.org/jira/browse/BEAM-9968 RemoteExecutionTest
 - https://issues.apache.org/jira/browse/BEAM-10485 ElasticSearchIOTest (just filed from seeing it flake on another PR)
 - https://issues.apache.org/jira/browse/BEAM-8025 CassandraIOTest
 - https://issues.apache.org/jira/browse/BEAM-10460 SparkPortableExecutionTest (now resolved, I think)

kennknowles:
Run Java PreCommit

kennknowles:
OK this is known flake https://issues.apache.org/jira/browse/BEAM-10470 in `JdbcDriverTest`, totally unrelated to this and shouldn't even be running. At this point I feel comfortable merging this so things can move on despite flakes.

PR #11864, opened 2020-05-29
[BEAM-10051] Scope instruction data. 

lostluck:
Make DataChannels instruction scope aware so that whenever data arrives, it will either be used, or safely dropped.

In particular, the start of an instruction isn't coupled to when the data will arrive. So there are a few cases that needed particular handling.

* Happy Path: Instruction starts processing, creating a reader, and starts to read data before any data arrives.
   * This case was handled very well already.
* Initial data arrives before instruction start, and reader creation.
  * This case was also handled well, data would be cached, and eventually cause push back until the instruction requests a reader, at which point it's the same as the happy path.
* Initial data arrives and instruction never creates a reader.
  * This can happen if there are issues initializing a DoFn in Setup or StartBundle. This can cause the worker to be hung for other bundles.
* Initial data arrives after instruction ends.
  * Similarly to the never creates a reader case, this can cause a hung worker. The data channel needs to be aware at least temporarily that the instruction is over and shouldn't create readers for it.

When the ScopedDataManager is closed, all channels are notified to remove the instruction and close the readers and writers. 
At the point in the stack where ScopedDataMana

ger is closed, readers needed to be handled carefully, but writers do not. Writers are only created by the DataSink node, and managed by the framework. In particular, if the writer was already closed, it would have removed themselves from the writer map on the Close call, preventing the ScopedDataManager from a double close. Otherwise, the writer would be in the map (such as due to a panic preventing the ordinary close), and need closing.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/go/pkg/beam/core/runtime/harness/datamgr.go
+++ b/sdks/go/pkg/beam/core/runtime/harness/datamgr.go
@@ -38,8 +38,6 @@ type ScopedDataManager struct {
 	mgr    *DataChannelManager
 	instID instructionID
 
-	// TODO(herohde) 7/20/2018: capture and force close open reads/writes. However,
-	// we would need the underlying Close to be idempotent or a separate method.
 	closed bool
 	mu     sync.Mutex
 }
@@ -82,9 +80,10 @@ func (s *ScopedDataManager) open(ctx context.Context, port exec.Port) (*DataChan
 // Close prevents new IO for this instruction.
 func (s *ScopedDataManager) Close() error {
 	s.mu.Lock()
+	defer s.mu.Unlock()
 	s.closed = true
+	s.mgr.closeInstruction(s.instID)
 	s.mgr = nil
-	s.mu.Unlock()
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


lostluck:
Run Go PostCommit

lostluck:
R: @youngoli

lukecwik:
`Initial data arrives and instruction never creates a reader` and `Initial data arrives after instruction ends` should eventually allow progress via a timeout on how long we expect this data to sit there without an active bundle. Anything on the order of 10s of seconds seems like plenty of time since we expect the ProcessBundle instruction to be in the control stream already (albeit possibly stuck in a network buffer or so). The SDK just needs to record the id in memory and if it ever sees a bundle for such an id, it would fail it immediately since it had dropped the data already.

lostluck:
This PR does have the assumption baked in that we'll always get a data message and a control message for an instruction.

It doesn't handle the less well behaved cases of "only receive instructions, and never any data" or the "only data, never instructions" cases, which, as you say, probably require a time component to handle them properly. 

Only instruction, is a bit of a waste of CPU, and ends up using little to no CPU while it's waiting for a channel send. On the other hand, it will then never signal the runner half or self terminate. So if the runner is waiting on that, it's it's own fault, and not well behaved. This may cause problems for a stage as a whole if the runner doesn't decide to disregard this bundle.

Only data is a bigger risk for an individual worker, since it will block the worker eventually with what I call the Boulder problem, since the channel buffer may fill, and will start "pushback" on the data channel preventing data messages from reaching the other processing threads. This is overall desirable behavior, up until it isn't and the instruction for that data never comes.

PR #1925, opened 2017-02-06
Add git .mailmap file

kennknowles:
Be sure to do all of the following to help us incorporate your contribution
quickly and easily:

 - [x] Make sure the PR title is formatted like:
   `[BEAM-<Jira issue #>] Description of pull request`
 - [x] Make sure tests pass via `mvn clean verify`. (Even better, enable
       Travis-CI on your fork and ensure the whole test matrix passes).
 - [x] Replace `<Jira issue #>` in the title with the actual Jira issue
       number, if there is one.
 - [x] If this contribution is large, please file an Apache
       [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.txt).

---

No JIRA, just a little fun `git` hack. Makes the names in the log nice. I only used information you could find in the log itself - the classic pattern is that someone contributes for a while before setting up their `git` configuration to have good metadata. Or in this case, also there was a tool that doesn't preserve good metadata.

PR #11886, opened 2020-06-02
[BEAM-8647] delete .mailmap

rmannibucau:
side note: if you really need this file  (like for the website) maybe use a private asf repo and ensure it is not in a public deliverable


lukecwik:
R: @kennknowles

lukecwik:
R: @iemejia

iemejia:
I think we are not using this file any more so ok to remove can you confirm @kennknowles

iemejia:
For extra info the discussion on `privacy-discuss@` on how to handle these cases (conclusion nothing is clear and we do not have real ASF recommendations yet)

https://lists.apache.org/thread.html/ra3dc372d3bcbcfa675e1e68867fceebc00aec0562d4c787a36f42562%40%3Cprivacy-discuss.apache.org%3E

rmannibucau:
@iemejia yes and no, 1. it is quite clear in europe and affects US based entities + 2 ASF can't justify to need this file (side note: it can trivially be replaced by GH API nowdays).

Lastly the overall PR was "by fun" and rather -1 (https://github.com/apache/beam/pull/1925) than +1 so I think it confirms it is just not used and it does not hurt to drop so no real need to discuss for year on that. If a better mapping is needed I suspect gh API can be used, a small script can be added to beam and .mailmap can be generated locally and added to .gitignore instead of explicitly exposing it (to scrappers/crawlers).

aaltay:
I do not believe we are using this file. I could not find a reference to it in any process documents. I agree that, it does not make much sense to keep this information in the repository. Also noting that removing this does not add much to the level of privacy, the same information is already publicly available in git log, or github apis.

ajamato:
@rmannibucau Do you need more help to move this forward? Or did you find an alternative way, if so can this PR be closed?

rmannibucau:
@ajamato I just need somebody hitting merge button ;). The side note to @aaltay comment is that it is easier to bulk clone that to bulk GET (api) due to the rate limits, so often scrappers use clone rather than the API.

kennknowles:
This file is very useful for producing the release notes and integrating with various git stats tools. It is just a reformatting of information available in the git log. I will read the linked thread.

kennknowles:
FWIW I am not sure everyone on this PR is familiar with the file. It is a config file to help git pretty print the log.

See the "List Of Contributors" on https://beam.apache.org/blog/beam-2.22.0/. Without this file, the list will have duplicate entries like `Kenneth Knowles <kenn@apache.org>` and `Kenn Knowles <klk@google.com>` and possibly some that are just `klk@google.com` etc. The file is a hint to git to de-dupe these when printing the log.

If anything, using GitHub APIs is worse, because it associates actual registered accounts that are connected to other information. All this file does is associate strings that already exist in the log.

kennknowles:
I see now that the thread link that Ismaël provided is the root discussion that we should follow to resolve this. I want to emphasize that also individual requests to remove information from this file I would of course support. Anyone removing their information should be warned that it does not protect the information or make it less public.

rmannibucau:
@kennknowles the argument that it is public for github so it can be made public in the git repo is wrong, when you have a github account you ack or not that and you can request to drop it due to GPDR so this file should be regenerated each time github changes anything (I assume once a day would work). Also, project must request authorization to use these data there and track ack+enable to change it to be legal. AFAIK none of both is done (right?) so this workflow does not work and can trivially be replaced by a github request - with a falling back on "unknown" or anything you like when the email is no more available- companies can request github to  become GPDR compliant which means data can be missing. You can also reverse the pattern and just list duplicates instead and handle the name in a case insensitive manner maybe? This reduces the number of entries a lot AFAIK.

kennknowles:
I am not arguing that "it is public for GitHub so it can be made public in the git repo". This file is not based on any GitHub data. It is based entirely on the git history itself.

rmannibucau:
Can it be done reopen and fixed please?

rmannibucau:
up

PR #12016, opened 2020-06-15
[BEAM-10341] Support drain in python and java SDF

boyuanzz:

------------------------

Lang | SDK | Apex | Dataflow | Flink | Samza | Spark
Go | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go/lastCompletedBuild/) | --- | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_VR_Flink/lastCompletedBuild/) | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_VR_Spark/lastCompletedBuild/)
Java | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Apex/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Dataflow/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Dataflow_Java11/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Flink/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Flink_Java11/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Flink_Batch/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Flink_Streaming/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Samza/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Spark/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Spark_Batch/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_SparkStructuredStreaming/lastCompletedBuild/)
Python | [Build Status](https://builds.apache.org/job/beam_PostCommit_Python2/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Python35/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Python36/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Python37/lastCompletedBuild/) | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Py_VR_Dataflow/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Py_VR_Dataflow_V2/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Py_ValCont/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PreCommit_Python2_PVR_Flink_Cron/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Python35_VR_Flink/lastCompletedBuild/) | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Python_VR_Spark/lastCompletedBuild/)

--- | --- | --- | --- | ---

### DIFF
--- a/model/pipeline/src/main/proto/beam_runner_api.proto
+++ b/model/pipeline/src/main/proto/beam_runner_api.proto
@@ -384,8 +384,10 @@ message StandardPTransforms {
     // Input: KV(KV(element, restriction), size); output: DoFn's output.
     PROCESS_SIZED_ELEMENTS_AND_RESTRICTIONS = 2 [(beam_urn) = "beam:transform:sdf_process_sized_element_and_restrictions:v1"];
 
-    // Truncates the restriction of each element/restriction pair and returns the
-    // finite restriction which is going to be processed when drain starts.
+    // Truncates the restriction of each element/restriction pair and returns
+    // the finite restriction which will be processed when a pipeline is
+    // drained: https://docs.google.com/document/d/1NExwHlj-2q2WUGhSO4jTu8XGhDPmm3cllSN8IMmWci8/edit#.
+    //
     // Input: KV(KV(element, restriction), size);
     // Output: KV(KV(element, restriction), size).
     TRUNCATE_SIZED_RESTRICTION = 3 [(beam_urn) = "beam:transform:sdf_truncate_sized_restrictions:v1"];
--- a/model/pipeline/src/main/proto/beam_runner_api.proto
+++ b/model/pipeline/src/main/proto/beam_runner_api.proto
@@ -383,6 +383,12 @@ message StandardPTransforms {
     //
     // Input: KV(KV(element, restriction), size); output: DoFn's output.
     PROCESS_SIZED_ELEMENTS_AND_RESTRICTIONS = 2 [(beam_urn) = "beam:transform:sdf_process_sized_element_and_restrictions:v1"];
+
+    // Truncates the restriction of each element/restriction pair and returns the
+    // finite restriction which is going to be processed when drain starts.
+    // Input: KV(KV(element, restriction), size);
+    // Output: KV(KV(element, restriction), size).
+    TRUNCATE_SIZED_RESTRICTION = 3 [(beam_urn) = "beam:transform:sdf_truncate_sized_restrictions:v1"];
   }
 }
 
--- a/model/pipeline/src/main/proto/beam_runner_api.proto
+++ b/model/pipeline/src/main/proto/beam_runner_api.proto
@@ -386,7 +386,9 @@ message StandardPTransforms {
 
     // Truncates the restriction of each element/restriction pair and returns
     // the finite restriction which will be processed when a pipeline is
-    // drained: https://docs.google.com/document/d/1NExwHlj-2q2WUGhSO4jTu8XGhDPmm3cllSN8IMmWci8/edit#.
+    // drained. See
+    // https://docs.google.com/document/d/1NExwHlj-2q2WUGhSO4jTu8XGhDPmm3cllSN8IMmWci8/edit#.
+    // for additional details about drain.
     //
     // Input: KV(KV(element, restriction), size);
     // Output: KV(KV(element, restriction), size).
--- a/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/PTransformTranslation.java
+++ b/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/PTransformTranslation.java
@@ -113,6 +113,8 @@ public class PTransformTranslation {
   // SplittableParDoComponents
   public static final String SPLITTABLE_PAIR_WITH_RESTRICTION_URN =
       "beam:transform:sdf_pair_with_restriction:v1";
+  public static final String SPLITTABLE_TRUNCATE_SIZED_RESTRICTION_URN =
+      "beam:transform:sdf_truncate_sized_restrictions:v1";
   /**
    * @deprecated runners should move away from using `SplittableProcessKeyedElements` and prefer to
    *     internalize any necessary SplittableDoFn expansion.
--- a/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/PTransformTranslation.java
+++ b/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/PTransformTranslation.java
@@ -195,6 +195,9 @@ public class PTransformTranslation {
     checkState(
         SPLITTABLE_PROCESS_SIZED_ELEMENTS_AND_RESTRICTIONS_URN.equals(
             getUrn(SplittableParDoComponents.PROCESS_SIZED_ELEMENTS_AND_RESTRICTIONS)));
+    checkState(
+        SPLITTABLE_TRUNCATE_SIZED_RESTRICTION_URN.equals(
+            getUrn(SplittableParDoComponents.TRUNCATE_SIZED_RESTRICTION)));
   }
 
   private static final Collection<TransformTranslator<?>> KNOWN_TRANSLATORS =
--- a/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/graph/SplittableParDoExpander.java
+++ b/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/graph/SplittableParDoExpander.java
@@ -85,7 +85,7 @@ public class SplittableParDoExpander {
    * sideInputA ---------\---------------------\----------------------\--------------------------\
    * sideInputB ---------V---------------------V----------------------V--------------------------V
    * mainInput ---> PairWithRestriction --> SplitAndSize --> TruncateAndSize --> ProcessSizedElementsAndRestriction --> outputA
-   *                                                                                                               \-> outputB
+   *                                                                                                                \-> outputB
    * }</pre>
    */
   public static TransformReplacement createTruncateReplacement() {
--- a/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/graph/SplittableParDoExpander.java
+++ b/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/graph/SplittableParDoExpander.java
@@ -297,8 +297,8 @@ public class SplittableParDoExpander {
     return prefix + i;
   }
 
-  /** See {@link #createTruncateReplacement()} ()} for details. */
-  private static class TruncateReplacement extends SizedReplacement {
+  /** See {@link #createTruncateReplacement} for details. */
+  private static class TruncateReplacement implements TransformReplacement {
     private static final TruncateReplacement INSTANCE = new TruncateReplacement();
 
     @Override
--- a/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/graph/SplittableParDoExpander.java
+++ b/runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/graph/SplittableParDoExpander.java
@@ -69,6 +69,29 @@ public class SplittableParDoExpander {
     return SizedReplacement.INSTANCE;
   }
 
+  /**
+   * Returns a transform replacement in drain mode which expands a splittable ParDo from:
+   *
+   * <pre>{@code
+   * sideInputA ---------\
+   * sideInputB ---------V
+   * mainInput ---> SplittableParDo --> outputA
+   *                                \-> outputB
+   * }</pre>
+   *
+   * into:
+   *
+   * <pre>{@code
+   * sideInputA ---------\---------------------\----------------------\--------------------------\
+   * sideInputB ---------V---------------------V----------------------V--------------------------V
+   * mainInput ---> PairWithRestriction --> SplitAndSize --> TruncateAndSize --> ProcessSizedElementsAndRestriction --> outputA
+   *                                                                                                               \-> outputB
+   * }</pre>
+   */
+  public static TransformReplacement createTruncateReplacement() {
+    return TruncateReplacement.INSTANCE;
+  }
+
   /** See {@link #createSizedReplacement()} for details. */
   private static class SizedReplacement implements TransformReplacement {
 
@@ -272,5 +295,212 @@ public class SplittableParDoExpander {
       i += 1;
     }
     return prefix + i;
+  }
+
+  /** See {@link #createTruncateReplacement()} ()} for details. */
+  private static class TruncateReplacement extends SizedReplacement {
+    private static final TruncateReplacement INSTANCE = new TruncateReplacement();
+
+    @Override
+    public MessageWithComponents getReplacement(
+        String transformId, ComponentsOrBuilder existingComponents) {
+      try {
+        MessageWithComponents.Builder rval = MessageWithComponents.newBuilder();
+
+        PTransform splittableParDo = existingComponents.getTransformsOrThrow(transformId);
+        ParDoPayload payload = ParDoPayload.parseFrom(splittableParDo.getSpec().getPayload());
+        // Only perform the expansion if this is a splittable DoFn.
+        if (payload.getRestrictionCoderId() == null || payload.getRestrictionCoderId().isEmpty()) {
+          return null;
+        }
+
+        String mainInputName = ParDoTranslation.getMainInputName(splittableParDo);
+        String mainInputPCollectionId = splittableParDo.getInputsOrThrow(mainInputName);
+        PCollection mainInputPCollection =
+            existingComponents.getPcollectionsOrThrow(mainInputPCollectionId);
+        Map<String, String> sideInputs =
+            Maps.filterKeys(
+                splittableParDo.getInputsMap(), input -> payload.containsSideInputs(input));
+
+        String pairWithRestrictionOutCoderId =
+            generateUniqueId(
+                mainInputPCollection.getCoderId() + "/PairWithRestriction",
+                existingComponents::containsCoders);
+        rval.getComponentsBuilder()
+            .putCoders(
+                pairWithRestrictionOutCoderId,
+                ModelCoders.kvCoder(
+                    mainInputPCollection.getCoderId(), payload.getRestrictionCoderId()));
+
+        String pairWithRestrictionOutId =
+            generateUniqueId(
+                mainInputPCollectionId + "/PairWithRestriction",
+                existingComponents::containsPcollections);
+        rval.getComponentsBuilder()
+            .putPcollections(
+                pairWithRestrictionOutId,
+                PCollection.newBuilder()
+                    .setCoderId(pairWithRestrictionOutCoderId)
+                    .setIsBounded(mainInputPCollection.getIsBounded())
+                    .setWindowingStrategyId(mainInputPCollection.getWindowingStrategyId())
+                    .setUniqueName(
+                        generateUniquePCollectonName(
+                            mainInputPCollection.getUniqueName() + "/PairWithRestriction",
+                            existingComponents))
+                    .build());
+
+        String splitAndSizeOutCoderId =
+            generateUniqueId(
+                mainInputPCollection.getCoderId() + "/SplitAndSize",
+                existingComponents::containsCoders);
+        rval.getComponentsBuilder()
+            .putCoders(
+                splitAndSizeOutCoderId,
+                ModelCoders.kvCoder(
+                    pairWithRestrictionOutCoderId, getOrAddDoubleCoder(existingComponents, rval)));
+
+        String splitAndSizeOutId =
+            generateUniqueId(
+                mainInputPCollectionId + "/SplitAndSize", existingComponents::containsPcollections);
+        rval.getComponentsBuilder()
+            .putPcollections(
+                splitAndSizeOutId,
+                PCollection.newBuilder()
+                    .setCoderId(splitAndSizeOutCoderId)
+                    .setIsBounded(mainInputPCollection.getIsBounded())
+                    .setWindowingStrategyId(mainInputPCollection.getWindowingStrategyId())
+                    .setUniqueName(
+                        generateUniquePCollectonName(
+                            mainInputPCollection.getUniqueName() + "/SplitAndSize",
+                            existingComponents))
+                    .build());
+
+        String truncateAndSizeCoderId =
+            generateUniqueId(
+                mainInputPCollection.getCoderId() + "/TruncateAndSize",
+                existingComponents::containsCoders);
+        rval.getComponentsBuilder()
+            .putCoders(
+                truncateAndSizeCoderId,
+                ModelCoders.kvCoder(
+                    splitAndSizeOutCoderId, getOrAddDoubleCoder(existingComponents, rval)));
+        String truncateAndSizeOutId =
+            generateUniqueId(
+                mainInputPCollectionId + "/TruncateAndSize",
+                existingComponents::containsPcollections);
+
+        rval.getComponentsBuilder()
+            .putPcollections(
+                truncateAndSizeOutId,
+                PCollection.newBuilder()
+                    .setCoderId(truncateAndSizeCoderId)
+                    .setIsBounded(mainInputPCollection.getIsBounded())
+                    .setWindowingStrategyId(mainInputPCollection.getWindowingStrategyId())
+                    .setUniqueName(
+                        generateUniquePCollectonName(
+                            mainInputPCollection.getUniqueName() + "/TruncateAndSize",
+                            existingComponents))
+                    .build());
+
+        String pairWithRestrictionId =
+            generateUniqueId(
+                transformId + "/PairWithRestriction", existingComponents::containsTransforms);
+        {
+          PTransform.Builder pairWithRestriction = PTransform.newBuilder();
+          pairWithRestriction.putAllInputs(splittableParDo.getInputsMap());
+          pairWithRestriction.putOutputs("out", pairWithRestrictionOutId);
+          pairWithRestriction.setUniqueName(
+              generateUniquePCollectonName(
+                  splittableParDo.getUniqueName() + "/PairWithRestriction", existingComponents));
+          pairWithRestriction.setSpec(
+              FunctionSpec.newBuilder()
+                  .setUrn(PTransformTranslation.SPLITTABLE_PAIR_WITH_RESTRICTION_URN)
+                  .setPayload(splittableParDo.getSpec().getPayload()));
+          pairWithRestriction.setEnvironmentId(splittableParDo.getEnvironmentId());
+          rval.getComponentsBuilder()
+              .putTransforms(pairWithRestrictionId, pairWithRestriction.build());
+        }
+
+        String splitAndSizeId =
+            generateUniqueId(transformId + "/SplitAndSize", existingComponents::containsTransforms);
+        {
+          PTransform.Builder splitAndSize = PTransform.newBuilder();
+          splitAndSize.putInputs(mainInputName, pairWithRestrictionOutId);
+          splitAndSize.putAllInputs(sideInputs);
+          splitAndSize.putOutputs("out", splitAndSizeOutId);
+          splitAndSize.setUniqueName(
+              generateUniquePCollectonName(
+                  splittableParDo.getUniqueName() + "/SplitAndSize", existingComponents));
+          splitAndSize.setSpec(
+              FunctionSpec.newBuilder()
+                  .setUrn(PTransformTranslation.SPLITTABLE_SPLIT_AND_SIZE_RESTRICTIONS_URN)
+                  .setPayload(splittableParDo.getSpec().getPayload()));
+          splitAndSize.setEnvironmentId(splittableParDo.getEnvironmentId());
+          rval.getComponentsBuilder().putTransforms(splitAndSizeId, splitAndSize.build());
+        }
+
+        String truncateAndSizeId =
+            generateUniqueId(
+                transformId + "/TruncateAndSize", existingComponents::containsTransforms);
+        {
+          PTransform.Builder truncateAndSize = PTransform.newBuilder();
+          truncateAndSize.putInputs(mainInputName, splitAndSizeOutId);
+          truncateAndSize.putAllInputs(sideInputs);
+          truncateAndSize.putOutputs("out", truncateAndSizeOutId);
+          truncateAndSize.setUniqueName(
+              generateUniquePCollectonName(
+                  splittableParDo.getUniqueName() + "/TruncateAndSize", existingComponents));
+          truncateAndSize.setSpec(
+              FunctionSpec.newBuilder()
+                  .setUrn(PTransformTranslation.SPLITTABLE_TRUNCATE_SIZED_RESTRICTION_URN)
+                  .setPayload(splittableParDo.getSpec().getPayload()));
+          truncateAndSize.setEnvironmentId(splittableParDo.getEnvironmentId());
+          rval.getComponentsBuilder().putTransforms(truncateAndSizeId, truncateAndSize.build());
+        }
+
+        String processSizedElementsAndRestrictionsId =
+            generateUniqueId(
+                transformId + "/ProcessSizedElementsAndRestrictions",
+                existingComponents::containsTransforms);
+        {
+          PTransform.Builder processSizedElementsAndRestrictions = PTransform.newBuilder();
+          processSizedElementsAndRestrictions.putInputs(mainInputName, truncateAndSizeOutId);
+          processSizedElementsAndRestrictions.putAllInputs(sideInputs);
+          processSizedElementsAndRestrictions.putAllOutputs(splittableParDo.getOutputsMap());
+          processSizedElementsAndRestrictions.setUniqueName(
+              generateUniquePCollectonName(
+                  splittableParDo.getUniqueName() + "/ProcessSizedElementsAndRestrictions",
+                  existingComponents));
+          processSizedElementsAndRestrictions.setSpec(
+              FunctionSpec.newBuilder()
+                  .setUrn(
+                      PTransformTranslation.SPLITTABLE_PROCESS_SIZED_ELEMENTS_AND_RESTRICTIONS_URN)
+                  .setPayload(splittableParDo.getSpec().getPayload()));
+          processSizedElementsAndRestrictions.setEnvironmentId(splittableParDo.getEnvironmentId());
+          rval.getComponentsBuilder()
+              .putTransforms(
+                  processSizedElementsAndRestrictionsId,
+                  processSizedElementsAndRestrictions.build());
+        }
+
+        PTransform.Builder newCompositeRoot =
+            splittableParDo
+                .toBuilder()
+                // Clear the original splittable ParDo spec and add all the new transforms as
+                // children.
+                .clearSpec()
+                .addAllSubtransforms(
+                    Arrays.asList(
+                        pairWithRestrictionId,
+                        splitAndSizeId,
+                        truncateAndSizeId,
+                        processSizedElementsAndRestrictionsId));
+        rval.setPtransform(newCompositeRoot);
+
+        return rval.build();
+      } catch (IOException e) {
+        throw new RuntimeException("Unable to perform expansion for transform " + transformId, e);
+      }
+    }
   }
 }
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/DoFn.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/DoFn.java
@@ -1071,9 +1071,6 @@ public abstract class DoFn<InputT, OutputT> implements Serializable, HasDisplayD
    * <p>This method must satisfy the following constraints:
    *
    * <ul>
-   *   <li>If one of the arguments is of type {@link OutputReceiver}, then it will be passed an
-   *       output receiver for outputting the truncated restrictions. All truncated restrictions
-   *       must be output through this parameter.
    *   <li>If one of its arguments is tagged with the {@link Element} annotation, then it will be
    *       passed the current element being processed; the argument must be of type {@code InputT}.
    *       Note that automatic conversion of {@link Row}s and {@link FieldAccess} parameters are
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/DoFn.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/DoFn.java
@@ -1094,6 +1094,8 @@ public abstract class DoFn<InputT, OutputT> implements Serializable, HasDisplayD
    *       options for the current pipeline.
    * </ul>
    *
+   * <p>Returns a truncated restriction representing a bounded amount of work that must be processed before the pipeline can be drained or {@code null} if no work is necessary.
+   *
    * <p>The default behavior when a pipeline is being drained is that {@link
    * org.apache.beam.sdk.transforms.splittabledofn.RestrictionTracker.IsBounded#BOUNDED}
    * restrictions process entirely while {@link
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/DoFn.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/DoFn.java
@@ -1066,7 +1066,7 @@ public abstract class DoFn<InputT, OutputT> implements Serializable, HasDisplayD
    * <p>This method is used to perform truncation of the restriction while it is not actively being
    * processed.
    *
-   * <p>Signature: {@code TruncateResult<RestrictionT> truncateRestriction(<arguments>);}
+   * <p>Signature: {@code @Nullable TruncateResult<RestrictionT> truncateRestriction(<arguments>);}
    *
    * <p>This method must satisfy the following constraints:
    *
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/DoFn.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/DoFn.java
@@ -1066,7 +1066,7 @@ public abstract class DoFn<InputT, OutputT> implements Serializable, HasDisplayD
    * <p>This method is used to perform truncation of the restriction while it is not actively being
    * processed.
    *
-   * <p>Signature: {@code Optional<RestrictionT> truncateRestriction(<arguments>);}
+   * <p>Signature: {@code TruncateResult<RestrictionT> truncateRestriction(<arguments>);}
    *
    * <p>This method must satisfy the following constraints:
    *
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/DoFn.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/DoFn.java
@@ -1064,7 +1064,7 @@ public abstract class DoFn<InputT, OutputT> implements Serializable, HasDisplayD
    * <p>This method is used to perform truncation of the restriction while it is not actively being
    * processed.
    *
-   * <p>Signature: {@code void truncateRestriction(<arguments>);}
+   * <p>Signature: {@code Optional<RestrictionT> truncateRestriction(<arguments>);}
    *
    * <p>This method must satisfy the following constraints:
    *
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/DoFn.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/DoFn.java
@@ -1034,6 +1034,9 @@ public abstract class DoFn<InputT, OutputT> implements Serializable, HasDisplayD
    *   <li>If one of its arguments is tagged with the {@link Timestamp} annotation, then it will be
    *       passed the timestamp of the current element being processed; the argument must be of type
    *       {@link Instant}.
+   *   <li>If one of its arguments is a {@link RestrictionTracker}, then it will be passed a tracker
+   *       that is initialized for the current {@link Restriction}. The argument must be of the
+   *       exact type {@code RestrictionTracker<RestrictionT, PositionT>}.
    *   <li>If one of its arguments is a subtype of {@link BoundedWindow}, then it will be passed the
    *       window of the current element. When applied by {@link ParDo} the subtype of {@link
    *       BoundedWindow} must match the type of windows on the input {@link PCollection}. If the
@@ -1057,7 +1060,7 @@ public abstract class DoFn<InputT, OutputT> implements Serializable, HasDisplayD
    *
    * <p>This method is used to perform truncating the restriction before actual processing.
    *
-   * <p>Signature: {@code void truncateSizedRestriction(<arguments>);}
+   * <p>Signature: {@code void truncateRestriction(<arguments>);}
    *
    * <p>This method must satisfy the following constraints:
    *
@@ -1092,7 +1095,7 @@ public abstract class DoFn<InputT, OutputT> implements Serializable, HasDisplayD
   @Retention(RetentionPolicy.RUNTIME)
   @Target(ElementType.METHOD)
   @Experimental(Kind.SPLITTABLE_DO_FN)
-  public @interface TruncateSizedRestriction {}
+  public @interface TruncateRestriction {}
 
   /**
    * Annotation for the method that creates a new {@link RestrictionTracker} for the restriction of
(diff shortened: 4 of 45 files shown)


sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/GrowableOffsetRangeTracker.java
```diff
@@ -140,4 +140,9 @@ public Progress getProgress() {
         totalWork.subtract(workRemaining, MathContext.DECIMAL128).doubleValue(),
         workRemaining.doubleValue());
   }
+
+  @Override
+  public boolean isBounded() {
+    return false;
```
lukecwik, line 146: Shouldn't this be true if `range.getTo() != Long.MAX_VALUE`?

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/RestrictionTracker.java
```diff
@@ -99,6 +100,10 @@
    */
   public abstract void checkDone() throws IllegalStateException;
 
+  public boolean isBounded() throws NotImplementedException {
+    throw new NotImplementedException("isBounded is not implemented.");
```
lukecwik, line 104: If we want to make this optional, we should follow the `HasProgress` style and use an interface like `HasBoundedness`

We could also make this a required method and update the changelog stating that this is a breaking change. This would remove the "default" inference on the `@UnboundedPerElement`/`@BoundedPerElement` annotations on the DoFn.

Depending on the default ordering we want, we'll want to update update the restriction trackers in PeriodicSequence and Watch as well (note that Watch will have to be updated regardless since it has a tracker that converts unbounded restrictions to bounded ones).

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/RestrictionTracker.java
```diff
@@ -99,6 +100,10 @@
    */
   public abstract void checkDone() throws IllegalStateException;
 
+  public boolean isBounded() throws NotImplementedException {
```
lukecwik, line 103: We should use the return type `IsBounded`.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/RestrictionTracker.java
lukecwik, line 104: We should add a comment stating the unbounded restriction trackers can only be used with `@UnboundedPerElement` SDFs. We should add a comment on each restriction tracker whether it could be unbounded.

sdks/java/harness/src/main/java/org/apache/beam/fn/harness/FnApiDoFnRunner.java
```diff
@@ -574,6 +578,99 @@ public Instant timestamp(DoFn<InputT, OutputT> doFn) {
...
+          this.processContext =
+              new WindowObservingProcessBundleContext() {
+                @Override
+                public void outputWithTimestamp(OutputT output, Instant timestamp) {
+                  double size =
+                      doFnInvoker.invokeGetSize(
+                          new DelegatingArgumentProvider<InputT, OutputT>(
+                              this,
+                              PTransformTranslation.SPLITTABLE_TRUNCATE_SIZED_RESTRICTION_URN
+                                  + "/GetSize") {
```
lukecwik, line 598:
```suggestion
                                  + "/TruncateRestriction") {
```

sdks/java/harness/src/main/java/org/apache/beam/fn/harness/FnApiDoFnRunner.java
```diff
@@ -574,6 +578,99 @@ public Instant timestamp(DoFn<InputT, OutputT> doFn) {
...
+            || (doFnSignature.getSize() != null && doFnSignature.getSize().observesWindow())
+            || !sideInputMapping.isEmpty()) {
+          mainInputConsumer = this::processElementForWindowObservingTruncateSizedRestriction;
+          // OutputT == RestrictionT
+          this.processContext =
+              new WindowObservingProcessBundleContext() {
+                @Override
+                public void outputWithTimestamp(OutputT output, Instant timestamp) {
+                  double size =
+                      doFnInvoker.invokeGetSize(
```
lukecwik, line 594: invokeGetSize?

sdks/java/harness/src/main/java/org/apache/beam/fn/harness/FnApiDoFnRunner.java
```diff
@@ -574,6 +578,99 @@ public Instant timestamp(DoFn<InputT, OutputT> doFn) {
...
         break;
+      case PTransformTranslation.SPLITTABLE_TRUNCATE_SIZED_RESTRICTION_URN:
+        if ((doFnSignature.truncateSizedRestriction() != null
+                && doFnSignature.truncateSizedRestriction().observesWindow())
+            || (doFnSignature.newTracker() != null && doFnSignature.newTracker().observesWindow())
+            || (doFnSignature.getSize() != null && doFnSignature.getSize().observesWindow())
+            || !sideInputMapping.isEmpty()) {
+          mainInputConsumer = this::processElementForWindowObservingTruncateSizedRestriction;
+          // OutputT == RestrictionT
+          this.processContext =
```
lukecwik, line 589: Since this shares the same ProcessBundleContext as GetSize (except for the error context string), can we make this a class called SizedRestrictionWindowObservingProcessBundleContext that extends WindowObservingProcessBundleContext and is used in both places.

Ditto for the non window observing one as well.

sdks/java/harness/src/main/java/org/apache/beam/fn/harness/FnApiDoFnRunner.java
```diff
@@ -608,17 +705,6 @@ public void accept(WindowedValue input) throws Exception {
     pCollectionConsumerRegistry.register(
         pTransform.getInputsOrThrow(mainInput), pTransformId, (FnDataReceiver) mainInputConsumer);
 
-    switch (pTransform.getSpec().getUrn()) {
```
lukecwik, line 611: Thanks

sdks/java/harness/src/main/java/org/apache/beam/fn/harness/FnApiDoFnRunner.java
```diff
@@ -829,6 +917,68 @@ private void processElementForWindowObservingSplitRestriction(
     this.stateAccessor.finalizeState();
   }
 
+  private void processElementForTruncateSizedRestriction(
+      WindowedValue<KV<KV<InputT, KV<RestrictionT, WatermarkEstimatorStateT>>, Double>> elem) {
+    currentElement = elem.withValue(elem.getValue().getKey().getKey());
+    currentRestriction = elem.getValue().getKey().getValue().getKey();
+    currentWatermarkEstimatorState = elem.getValue().getKey().getValue().getValue();
+    currentTracker =
```
lukecwik, line 925: It looks like currentTracker isn't being set within both `processElementForSplitRestriction` methods. Could you fix that in this PR?

sdks/java/harness/src/main/java/org/apache/beam/fn/harness/FnApiDoFnRunner.java
```diff
@@ -858,6 +1008,10 @@ public static WindowedSplitResult forRoots(
   private void processElementForWindowObservingSizedElementAndRestriction(
       WindowedValue<KV<KV<InputT, KV<RestrictionT, WatermarkEstimatorStateT>>, Double>> elem) {
     currentElement = elem.withValue(elem.getValue().getKey().getKey());
+    // No need to process if current restriction is null.
+    if (elem.getValue().getKey().getValue().getKey() == null) {
```
lukecwik, line 1012: I don't think this is the right way to go. The output received should produce 0 outputs instead of null.

sdks/python/apache_beam/runners/common.py
```diff
@@ -968,6 +973,9 @@ def process_with_sized_restriction(self, windowed_value):
     (element, (restriction, estimator_state)), _ = windowed_value.value
     watermark_estimator = (
         self.do_fn_invoker.invoke_create_watermark_estimator(estimator_state))
+    if restriction is None:
```
lukecwik, line 976: Same issue as in java, we should allow None restrictions and instead produce 0 outputs.

sdks/python/apache_beam/runners/worker/bundle_processor.py
```diff
@@ -1428,6 +1428,36 @@ def process(self, element_restriction, *args, **kwargs):
...
+def create_truncate_sized_restriction(*args):
+  class TruncateAndSizeRestriction(beam.DoFn):
+    def __init__(self, fn, restriction_provider, watermark_estimator_provider):
+      self.restriction_provider = restriction_provider
+      self.dofn_signature = common.DoFnSignature(fn)
+
+    def process(self, element_restrictin, *args, **kwargs):
+      ((element, (restriction, estimator_state)), _) = element_restrictin
+      try:
+        truncated_restriciton = self.restriction_provider.truncate(
```
lukecwik, line 1443: truncated_restriciton -> truncated_restriction

sdks/python/apache_beam/runners/worker/bundle_processor.py
```diff
@@ -1428,6 +1428,36 @@ def process(self, element_restriction, *args, **kwargs):
...
+@BeamTransformFactory.register_urn(
+    common_urns.sdf_components.TRUNCATE_SIZED_RESTRICTION.urn,
+    beam_runner_api_pb2.ParDoPayload)
+def create_truncate_sized_restriction(*args):
+  class TruncateAndSizeRestriction(beam.DoFn):
+    def __init__(self, fn, restriction_provider, watermark_estimator_provider):
+      self.restriction_provider = restriction_provider
+      self.dofn_signature = common.DoFnSignature(fn)
+
+    def process(self, element_restrictin, *args, **kwargs):
```
lukecwik, line 1440: element_restrictin -> element_restriction

sdks/python/apache_beam/runners/worker/bundle_processor.py
lukecwik, line 1443: We should be returning the same as split_and_size allowing for 0 restrictions to be returned.

sdks/python/apache_beam/transforms/core.py
```diff
@@ -320,6 +320,22 @@ def split_and_size(self, element, restriction):
...
+    restriction. If the restriction is unbounded, it will return None.
+
+    The method throws NotImplementError when RestrictionTracker.is_bounded() is
+    not implemented.
+
+    It's recommended to implement this API if more granularity is required.
+    """
+    restriction_tracker = self.create_tracker(restriction)
+    if restriction_tracker.is_bounded():
+      return restriction
```
lukecwik, line 337:
```suggestion
      yield restriction
```

sdks/python/apache_beam/transforms/core.py
```diff
@@ -320,6 +320,22 @@ def split_and_size(self, element, restriction):
     for part in self.split(element, restriction):
       yield part, self.restriction_size(element, part)
 
+  def truncate(self, element, restriction):
```
lukecwik, line 323: Should we also provide truncate_and_size?

sdks/python/apache_beam/transforms/core.py
```diff
@@ -320,6 +320,22 @@ def split_and_size(self, element, restriction):
...
+    """Truncate the given restriction into finite amount of work when the
+    pipeline starts to drain.
+
+    By default, if the restriction is bounded, it will return the entire current
+    restriction. If the restriction is unbounded, it will return None.
```
lukecwik, line 328:
```suggestion
    By default, if the restriction is bounded, it will return the entire 
    restriction. If the restriction is unbounded, it will not return anything.
```

sdks/python/apache_beam/io/iobase.py
```diff
@@ -1243,6 +1243,17 @@ def try_claim(self, position):
...
+  def is_bounded(self):
+    """Identify whether the output produced by the current restriction is
+    bounded.
+
+    The value is important for the default behavior of truncate when the
+    pipeline starts to drain in streaming. If the current restriction is
+    bounded, it will be processed completely by default. If the restriction is
+    unbounded, it will be truncated into null and finish processing immediately.
+    """
+    raise NotImplementedError
```
lukecwik, line 1255: ThreadsafeRestrictionTracker should call the delegate so that users of it get the same behavior as if they were calling the real restriction tracker

We might want to update the RestrictionTrackerView if we want to expose the boundedness property during execution.

model/pipeline/src/main/proto/beam_runner_api.proto
```diff
@@ -383,6 +383,12 @@ message StandardPTransforms {
     //
     // Input: KV(KV(element, restriction), size); output: DoFn's output.
     PROCESS_SIZED_ELEMENTS_AND_RESTRICTIONS = 2 [(beam_urn) = "beam:transform:sdf_process_sized_element_and_restrictions:v1"];
+
+    // Truncates the restriction of each element/restriction pair and returns the
+    // finite restriction which is going to be processed when drain starts.
+    // Input: KV(KV(element, restriction), size);
+    // Output: KV(KV(element, restriction), size).
```
lukecwik, line 390: Add a link to the drain design doc that Reuven shared with the Beam community.

```suggestion
    // Truncates the restriction of each element/restriction pair and returns the
    // finite restriction which will be processed when a pipeline is drained <link>.
    //
    // Input: KV(KV(element, restriction), size);
    // Output: KV(KV(element, restriction), size).
```

sdks/python/apache_beam/transforms/core.py
```diff
@@ -622,6 +638,16 @@ class DoFn(WithTypeHints, HasDisplayData, urns.RunnerApiFn):
   def from_callable(fn):
     return CallableWrapperDoFn(fn)
 
+  @staticmethod
+  def unbounded_per_element():
```
lukecwik, line 642: This should also control whether the PCollection is unbounded or not during expand.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/GrowableOffsetRangeTracker.java
boyuanzz: Yeah you are right.

sdks/java/harness/src/main/java/org/apache/beam/fn/harness/FnApiDoFnRunner.java
boyuanzz: The input of `processElementForSizedRestriction` is `((element, (restriction, watermark_state)), size)`. Thus the output of `truncate` should also be sized.

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/PTransformTranslation.java
```diff
@@ -113,6 +113,8 @@
   // SplittableParDoComponents
   public static final String SPLITTABLE_PAIR_WITH_RESTRICTION_URN =
       "beam:transform:sdf_pair_with_restriction:v1";
+  public static final String SPLITTABLE_TRUNCATE_SIZED_RESTRICTION_URN =
+      "beam:transform:sdf_truncate_sized_restrictions:v1";
```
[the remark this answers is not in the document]
boyuanzz: Should we use `getUrn` directly instead of hardcoding the const string and the checkState?

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/DoFn.java
```diff
@@ -1051,7 +1092,7 @@ public Duration getAllowedTimestampSkew() {
   @Retention(RetentionPolicy.RUNTIME)
   @Target(ElementType.METHOD)
   @Experimental(Kind.SPLITTABLE_DO_FN)
-  public @interface SplitRestriction {}
+  public @interface TruncateSizedRestriction {}
```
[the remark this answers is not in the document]
boyuanzz: That makes sense. Just want to mention that the python naming `TruncateAndSizeRestriction` will stay the same since we expose the size there.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/RestrictionTracker.java
boyuanzz: Are you suggesting that using `PCollection.IsBounded` as return type? Any reason that a boolean doesn't work here?

sdks/java/harness/src/main/java/org/apache/beam/fn/harness/FnApiDoFnRunner.java
boyuanzz: Yeah I'm going to update it.

sdks/python/apache_beam/transforms/core.py
boyuanzz: I created a JIRA to track this: https://issues.apache.org/jira/browse/BEAM-10380. I want to have a separate PR to do this since it is out of the scope of this PR.

sdks/python/apache_beam/transforms/core.py
boyuanzz: `restriction_size` is a required API, thus `truncate` is enough here since the caller should be the system.

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/PTransformTranslation.java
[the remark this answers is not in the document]
lukecwik: No. Only compile time constants can be used in switch case statements which is why it is like this.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/RestrictionTracker.java
lukecwik: Your right in using PCollection.IsBounded would be weird.

The purpose of not using a boolean on PCollection was to make it clear whether something was bounded or unbounded. Boolean values also require the person to remember what the contract of the method is.

We can define an enum in RestrictionTracker similar to what we do in PCollection.

sdks/java/harness/src/main/java/org/apache/beam/fn/harness/FnApiDoFnRunner.java
lukecwik: Your right, my comment didn't make sense.

model/pipeline/src/main/proto/beam_runner_api.proto
lukecwik: please address

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/PTransformTranslation.java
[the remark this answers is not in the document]
lukecwik: please address

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/DoFn.java
```diff
@@ -1034,6 +1034,47 @@ public Duration getAllowedTimestampSkew() {
...
+   * </ul>
+   */
+  @Documented
+  @Retention(RetentionPolicy.RUNTIME)
+  @Target(ElementType.METHOD)
+  @Experimental(Kind.SPLITTABLE_DO_FN)
+  public @interface SplitRestriction {}
+
+  /**
+   * Annotation for the method that truncates restriction of a <a
```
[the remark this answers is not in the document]
lukecwik: please address

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/reflect/ByteBuddyDoFnInvokerFactory.java
```diff
@@ -306,6 +309,30 @@ public static void invokeSplitRestriction(DoFnInvoker.ArgumentProvider argumentP
...
+    DefaultTruncateRestriction(DoFnSignature doFnSignature) {
+      this.isBoundedPerElement = doFnSignature.isBoundedPerElement().equals(IsBounded.BOUNDED);
+    }
+
+    /** Return the current restriction if it's bounded.Otherwise, return null. */
+    @SuppressWarnings("unused")
+    public void invokeTruncateRestriction(DoFnInvoker.ArgumentProvider argumentProvider) {
+      boolean isBounded = this.isBoundedPerElement;
+      try {
+        isBounded = argumentProvider.restrictionTracker().isBounded();
```
lukecwik, line 326: As discussed offline, this will be required to be implemented.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/reflect/DoFnSignatures.java
```diff
@@ -1794,6 +1812,60 @@ private static boolean hasAnnotation(Class<?> annotation, List<Annotation> annot
...
+            "Uses restriction type %s, but @%s method uses restriction type %s",
+            format(((RestrictionParameter) extraParam).restrictionT()),
+            format(DoFn.GetInitialRestriction.class),
+            format(restrictionT));
+      }
+      methodContext.addParameter(extraParam);
+    }
+
+    for (Parameter parameter : methodContext.getExtraParameters()) {
+      checkParameterOneOf(errors, parameter, ALLOWED_SPLIT_RESTRICTION_PARAMETERS);
```
lukecwik, line 1862:
```suggestion
      checkParameterOneOf(errors, parameter, ALLOWED_TRUNCATE_RESTRICTION_PARAMETERS);
```

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/RestrictionTracker.java
lukecwik: As discussed offline, we'll make `RestrictionTracker#isBounded` a required method and we will not use `@UnboundedPerElement`/`@BoundedPerElement` as part of the default inference.

lukecwik: Not relevant anymore.

sdks/java/harness/src/main/java/org/apache/beam/fn/harness/FnApiDoFnRunner.java
```diff
@@ -1636,6 +1657,96 @@ public void outputWithTimestamp(OutputT output, Instant timestamp) {
     }
   }
 
+  private class SizedRestrictionWindowObservingProcessBundleContext
```
lukecwik, line 1660: Class comment that this context outputs `KV<KV<Element, Restriction>, KV<Size, WatermarkEsimtatorStateT>>`

sdks/java/harness/src/main/java/org/apache/beam/fn/harness/FnApiDoFnRunner.java
```diff
@@ -1636,6 +1657,96 @@ public void outputWithTimestamp(OutputT output, Instant timestamp) {
...
+                      KV.of(
+                          currentElement.getValue(), KV.of(output, currentWatermarkEstimatorState)),
+                      size),
+                  timestamp,
+                  currentWindow,
+                  currentElement.getPane()));
+    }
+  }
+
+  private class SizedRestrictionNonWindowObservingProcessBundleContext
```
lukecwik, line 1705: Class comment that this context outputs `KV<KV<Element, Restriction>, KV<Size, WatermarkEsimtatorStateT>>`

sdks/python/apache_beam/transforms/core.py
lukecwik: sg

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/RestrictionTracker.java
boyuanzz: I would prefer having Enum in RestrictionTracker. `isBounded` should mean current restriction is bounded or not, kind of different from what `PCollection.IsBounded` means.

lukecwik: sgtm

sdks/python/apache_beam/transforms/core.py
boyuanzz: I think we should return here since we expect `truncate` to produce only one truncated restriction per input restriction.

sdks/java/core/src/test/java/org/apache/beam/sdk/transforms/reflect/DoFnInvokersTest.java
```diff
@@ -742,10 +785,202 @@ public void outputWithTimestamp(String output, Instant instant) {
...
+          }
+
+          @Override
+          public OutputReceiver<String> outputReceiver(DoFn<String, String> doFn) {
+            return new DoFn.OutputReceiver<String>() {
+              private boolean invoked;
+
+              @Override
+              public void output(String output) {
+                assertFalse(invoked);
```
lukecwik, line 874: This validation only happens within this method so if it is never invoked, no validation occurs. We have to perform the validation outside of this method or validate that the method was invoked somehow.

I know that other tests did this incorrectly as well. We can fix now or as a follow-up PR.

sdks/java/core/src/test/java/org/apache/beam/sdk/transforms/reflect/DoFnInvokersTest.java
```diff
@@ -742,10 +785,202 @@ public void outputWithTimestamp(String output, Instant instant) {
...
+              @Override
+              public void output(String output) {
+                assertTrue(shouldInvoked);
+              }
+
+              // This should not be invoked.
+              @Override
+              public void outputWithTimestamp(String output, Instant instant) {
+                assertTrue(shouldInvoked);
+              }
```
lukecwik, line 979:
```suggestion

              // This should not be invoked.
              @Override
              public void output(String output) {
                fail("Never expected truncated restriction to be output: " + output);
              }

              // This should not be invoked.
              @Override
              public void outputWithTimestamp(String output, Instant instant) {
                fail("Never expected truncated restriction to be output: " + output + " timestamp: " + instant);
              }
```

sdks/java/core/src/test/java/org/apache/beam/sdk/transforms/reflect/DoFnSignaturesSplittableDoFnTest.java
```diff
@@ -956,6 +983,96 @@ public void splitRestriction(
     DoFnSignatures.getSignature(BadFn.class);
   }
 
+  @Test
+  public void testTruncateRestrictionReturnsWrongType() throws Exception {
```
lukecwik, line 987: nit: rename this test and the other ones like this since we aren't returning but having the wrong output receiver type
```suggestion
  public void testTruncateRestrictionWithOutputReceiverWithWrongGenericType() throws Exception {
```

sdks/java/harness/src/main/java/org/apache/beam/fn/harness/FnApiDoFnRunner.java
```diff
@@ -1636,6 +1657,98 @@ public void outputWithTimestamp(OutputT output, Instant timestamp) {
     }
   }
 
+  /** This context outputs KV<KV<Element, <Restriction, WatemarkEstimatorState>>, size> */
```
lukecwik, line 1660:
```suggestion
  /** This context outputs KV<KV<Element, KV<Restriction, WatemarkEstimatorState>>, Size> */
```

sdks/java/harness/src/main/java/org/apache/beam/fn/harness/FnApiDoFnRunner.java
```diff
@@ -1636,6 +1657,98 @@ public void outputWithTimestamp(OutputT output, Instant timestamp) {
...
+                      KV.of(
+                          currentElement.getValue(), KV.of(output, currentWatermarkEstimatorState)),
+                      size),
+                  timestamp,
+                  currentWindow,
+                  currentElement.getPane()));
+    }
+  }
+
+  /** This context outputs KV<KV<Element, <Restriction, WatemarkEstimatorState>>, size> */
```
lukecwik, line 1706:
```suggestion
  /** This context outputs KV<KV<Element, KV<Restriction, WatemarkEstimatorState>>, Size> */
```

sdks/python/apache_beam/transforms/core.py
lukecwik: We expect 0 or 1 which is why the yield is important. It also wouldn't matter if they returned more then 1. It might even help the drain complete sooner.

sdks/java/harness/src/main/java/org/apache/beam/fn/harness/FnApiDoFnRunner.java
```diff
@@ -488,90 +492,32 @@
...
-                              timestamp,
-                              currentElement.getWindows(),
-                              currentElement.getPane()));
-                }
-              };
+              new SizedRestrictionNonWindowObservingProcessBundleContext(
+                  PTransformTranslation.SPLITTABLE_SPLIT_AND_SIZE_RESTRICTIONS_URN);
+        }
+        break;
+      case PTransformTranslation.SPLITTABLE_TRUNCATE_SIZED_RESTRICTION_URN:
```
lukecwik, line 506: Must propagate progress/split calls to the downstream operator otherwise checkpointing will not work causing pipelines to produce to much output.

sdks/python/apache_beam/runners/worker/operations.py
```diff
@@ -743,11 +743,36 @@ def pcollection_count_monitoring_infos(self, tag_to_pcollection_id):
...
+    # type: () -> Optional[iobase.RestrictionProgress]
+    return self.sdf_process_op.current_element_progress()
+
+  def try_split(self, fraction_of_remainder):  # type: (...) -> Optional[Any]
+    result = self.sdf_process_op.try_split(fraction_of_remainder)
+    if result is not None:
+      return result
+    return None
+
+  def add_receiver(self, operation, output_index=0):
```
lukecwik, line 761: Why did we go down this path instead of replicating what DataInputOperation was doing where it accessed receivers[0] and invoked progress/split on it directly (which invokes progress/split on the ConsumerSet which only gets forwarded if it is a SingletonConsumerSet)?

This transform always has a single output so receivers[0] should always be populated.

sdks/python/apache_beam/runners/portability/fn_api_runner/execution.py
```diff
@@ -681,7 +681,21 @@ def input_for(self, transform_id, input_id):
     input_pcoll = self.process_bundle_descriptor.transforms[
         transform_id].inputs[input_id]
     for read_id, proto in self.process_bundle_descriptor.transforms.items():
+      # The GrpcRead is followed by the SDF/Process.
       if (proto.spec.urn == bundle_processor.DATA_INPUT_URN and
           input_pcoll in proto.outputs.values()):
         return read_id
+      # The GrpcRead is follwoed by the SDF/Truncate -> SDF/Process.
```
lukecwik, line 688: Why did this need to change?

lukecwik, line 688:
```suggestion
      # The GrpcRead is followed by the SDF/Truncate -> SDF/Process.
```

sdks/python/apache_beam/runners/worker/bundle_processor.py
```diff
@@ -1079,6 +1079,13 @@ def construct_bundle_application(self,
                                   ):
     # type: (...) -> beam_fn_api_pb2.BundleApplication
     transform_id, main_input_tag, main_input_coder, outputs = op.input_info
+    # The main_input_coder should be the main_input_coder of
+    # SdfTruncateSizedRestrictions if SdfProcessSizedElements is following
+    # SdfTruncateSizedRestrictions.
+    if (isinstance(op, operations.SdfProcessSizedElements) and
```
lukecwik, line 1085: I don't think this is necessary.

The bundle application should have splits/residuals for the SdfProcessSizedElements and should use its input coder and not the input coder of the sdf_truncate_op. The runner can choose to feed the split back through the truncate operation or it can choose to feed the split directly to the SdfProcessSizedElements based upon how it chooses to construct the graph.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/reflect/DoFnSignatures.java
```diff
@@ -1830,13 +1828,13 @@ private static boolean hasAnnotation(Class<?> annotation, List<Annotation> annot
...
-      TypeDescriptor<?> outputT,
       TypeDescriptor<?> restrictionT,
       FnAnalysisContext fnContext) {
     // Method is of the form:
     // @TruncateRestriction
-    // void truncateRestriction(... parameters ...);
-    errors.checkArgument(void.class.equals(m.getReturnType()), "Must return void");
+    // Optional<RestrictionT> truncateRestriction(... parameters ...);
+    errors.checkArgument(
+        Optional.class.equals(m.getReturnType()), "Must return Optional<Restriction>");
```
lukecwik, line 1837: We should be able to validate that the return type is Optional<RestrictionT>

sdks/java/core/src/test/java/org/apache/beam/sdk/transforms/reflect/DoFnInvokersTest.java
```diff
@@ -847,44 +848,26 @@ public Object restriction() {
...
-          @Override
-          public Object restriction() {
-            return "foo";
-          }
-
-          @Override
-          public OutputReceiver<String> outputReceiver(DoFn<String, String> doFn) {
-            return new DoFn.OutputReceiver<String>() {
-              private boolean invoked;
+    Optional result =
```
lukecwik, line 851:
```suggestion
    Optional<?> result =
```

sdks/java/core/src/test/java/org/apache/beam/sdk/transforms/reflect/DoFnInvokersTest.java
```diff
@@ -944,42 +927,25 @@ public Object restriction() {
...
-          @Override
-          public Object restriction() {
-            return "foo";
-          }
-
-          @Override
-          public OutputReceiver<String> outputReceiver(DoFn<String, String> doFn) {
-            return new DoFn.OutputReceiver<String>() {
-              private final boolean shouldInvoked = false;
+    Optional result =
```
lukecwik, line 930:
```suggestion
    Optional<?> result =
```

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/reflect/DoFnInvoker.java
```diff
@@ -95,7 +96,8 @@ void invokeOnTimer(
   void invokeSplitRestriction(ArgumentProvider<InputT, OutputT> arguments);
 
   /** Invoke the {@link TruncateRestriction} method on the bound {@link DoFn}. */
-  void invokeTruncateRestriction(ArgumentProvider<InputT, OutputT> arguments);
+  <RestrictionT> Optional<RestrictionT> invokeTruncateRestriction(
```
lukecwik, line 99: The reason why I though it would make sense to return Optional was that I was hoping that it would support the case when RestrictionT being null is a valid use case.

This doesn't turn out to work with Optional since it doesn't support null values. Would you rather have this API return null like Python returning None for consistency? (null meaning that there is nothing meaningful to process)

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/DoFn.java
```diff
@@ -1053,6 +1056,51 @@ public Duration getAllowedTimestampSkew() {
...
 
+  /**
+   * Annotation for the method that truncates the restriction of a <a
+   * href="https://s.apache.org/splittable-do-fn">splittable</a> {@link DoFn} into a bounded one to
+   * be processed when pipeline starts to drain.
+   *
+   * <p>This method is used to perform truncation of the restriction while it is not actively being
+   * processed.
+   *
+   * <p>Signature: {@code Optional<RestrictionT> truncateRestriction(<arguments>);}
```
lukecwik, line 1067: We should state what Optional.empty means as part of the contract.

lukecwik:
Note to self, make sure that capabilities are added for SDKs that support drain.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/reflect/DoFnInvoker.java
boyuanzz: My intent was to use `Optional.empty()` to represent there is no more work. So we could have 2 options:
1. Use `Optional.empty()` as a notion of no more remaining work and use `Optional.of(value)` as a  notion of bounded remaining work.
2. Change the function signature to `@Nullable RestrictionT truncateRestriction()`, where `null` means no more work.

I don't have a strong preference over these 2 options. But 2 is more consistent with python.

sdks/python/apache_beam/runners/worker/bundle_processor.py
boyuanzz: The problem here is the restriction coder. At least for `FnApiRunner`, when a `PCollection` crosses the wire, it seems like any non-standard coder will be length-prefixed. Here, the input coder of `SdfTruncateSizedRestrictions ` is length-prefixed, but `SdfProcessSizedElements ` is not. On the `FnApiRunner` side, it decodes the element-restriction pair with the length-prefixed coder.

sdks/python/apache_beam/runners/portability/fn_api_runner/execution.py
boyuanzz: It's the implementation details of how `FnApiRunner` re-schedules the delayed work. The original assumption here is that any operation that can produce delayed work(`timer`, `DelayedBundleApplication`) must consume inputs from `GrpcRead`. So when the `FnApiRunner` gets the `DelayedBundleApplication` from `SDF/Process`, it must find a corresponding `GrpcRead` transform id, which should be the input node of `SDF/Truncate` when there is a truncate.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/GrowableOffsetRangeTracker.java
```diff
@@ -140,4 +140,15 @@ public Progress getProgress() {
         totalWork.subtract(workRemaining, MathContext.DECIMAL128).doubleValue(),
         workRemaining.doubleValue());
   }
+
+  @Override
+  public RestrictionBoundness isBounded() {
+    // If current range has been done, the range should be bounded.
+    if (lastAttemptedOffset != null && lastAttemptedOffset == Long.MAX_VALUE) {
```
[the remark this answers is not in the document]
boyuanzz: This is for correctness purpose. Image `isBounded` is called after `tryClaim(Long.MAX_VALUE)`.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/RestrictionTracker.java
```diff
@@ -99,6 +99,24 @@
    */
   public abstract void checkDone() throws IllegalStateException;
 
+  public enum RestrictionBoundness {
```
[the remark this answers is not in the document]
boyuanzz: Thanks. Let's go with `IsBounded`.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/RestrictionTracker.java
```diff
@@ -99,6 +99,24 @@
...
+   * Return the boundedness of current tracking restriction. If the current restriction produces
+   * finite output, it should return {@link RestrictionBoundness#IS_BOUNDED}. Otherwise, it should
+   * return {@link RestrictionBoundness#IS_UNBOUNDED}.
+   *
+   * <p>The API is called by the system when the pipeline starts to drain and there is no
+   * implementation of {@link DoFn.TruncateRestriction}. Based on the boundness of the restriction,
+   * the system will give the default behavior of truncating restrictions.
+   *
+   * <p>The API is required to be implemented.
+   */
```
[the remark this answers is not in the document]
boyuanzz: Created JIRA https://issues.apache.org/jira/browse/BEAM-10480 for tracking this cleanup.

sdks/java/core/src/test/java/org/apache/beam/sdk/transforms/reflect/DoFnInvokersTest.java
boyuanzz: Given that we decided to make `@TruncateRestriction` returns truncated restrictions, changes here are no longer applied. Do you want me to create a JIRA for other tests?

sdks/java/core/src/test/java/org/apache/beam/sdk/transforms/reflect/DoFnInvokersTest.java
boyuanzz: It's no longer applied.

sdks/python/apache_beam/runners/worker/operations.py
boyuanzz: I did this for performance concerns. The `receivers` in python is a map of <output_index, list of receivers>. It's not efficient to get only one element from a map of <output_index, list of receivers> every time when `try_split` and `progress` are called. We can cache the receiver either when the first time `try_split` or `progress` is called, or when receivers are added.

boyuanzz:
> Note to self, make sure that capabilities are added for SDKs that support drain.

Could you please point me to the place of adding capabilities?

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/GrowableOffsetRangeTracker.java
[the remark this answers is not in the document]
lukecwik: I see.

sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/splittabledofn/RestrictionTracker.java
```diff
@@ -99,6 +99,20 @@
...
+    /** Indicates that a {@code Restriction} represents a bounded amount of work. */
+    BOUNDED,
+    /** Indicates that a {@code Restriction} represents an unbounded amount of work. */
+    UNBOUNDED
+  }
+
+  /**
+   * Return the boundedness of the current restriction. If the current restriction represents a
+   * finite amount of work, it should return {@link IsBounded#BOUNDED}. Otherwise, it should return
+   * {@link IsBounded#UNBOUNDED}.
```
lukecwik, line 112:
```suggestion
   * {@link IsBounded#UNBOUNDED}.
   *
   * <p>It is valid to return {@link IsBounded#BOUNDED} after returning {@link IsBounded#UNBOUNDED} once
   * the end of a restriction is discovered. It is not valid to return {@link IsBounded#UNBOUNDED} after returning {@link IsBounded#BOUNDED}.
```

sdks/java/harness/src/main/java/org/apache/beam/fn/harness/FnApiDoFnRunner.java
```diff
@@ -488,90 +493,70 @@
...
+        if ((doFnSignature.truncateRestriction() != null
+                && doFnSignature.truncateRestriction().observesWindow())
+            || (doFnSignature.newTracker() != null && doFnSignature.newTracker().observesWindow())
+            || (doFnSignature.getSize() != null && doFnSignature.getSize().observesWindow())
+            || !sideInputMapping.isEmpty()) {
+          mainInputConsumer =
+              new SplittableFnDataReceiver() {
+                private final HandlesSplits splitDelegate =
```
lukecwik, line 515: This logic assumes that there will always be one output and it must support splitting. It is valid for a runner to insert gRPC ports between truncate and process sized restriction (and some may choose to do so).

It would be best if we didn't assume the downstream consumer is splittable and that there is only one of them and instead check that these conditions are met within trySplit/getProgress. This is what BeamFnDataReadRunner does.

sdks/java/harness/src/main/java/org/apache/beam/fn/harness/FnApiDoFnRunner.java
```diff
@@ -488,90 +493,70 @@
...
-          // OutputT == RestrictionT
           this.processContext =
-              new NonWindowObservingProcessBundleContext() {
+              new SizedRestrictionWindowObservingProcessBundleContext(
+                  PTransformTranslation.SPLITTABLE_TRUNCATE_SIZED_RESTRICTION_URN);
+        } else {
+          mainInputConsumer =
+              new SplittableFnDataReceiver() {
+                private final HandlesSplits splitDelegate =
+                    (HandlesSplits) Iterables.get(mainOutputConsumers, 0);
```
lukecwik, line 540: Ditto for not assuming for the same reasons mentioned above.

sdks/java/harness/src/main/java/org/apache/beam/fn/harness/FnApiDoFnRunner.java
```diff
@@ -488,90 +493,70 @@
...
-                              KV.of(
-                                  KV.of(
-                                      currentElement.getValue(),
-                                      KV.of(output, currentWatermarkEstimatorState)),
-                                  size),
-                              timestamp,
-                              currentElement.getWindows(),
-                              currentElement.getPane()));
+                public SplitResult trySplit(double fractionOfRemainder) {
+                  return splitDelegate.trySplit(fractionOfRemainder);
```
lukecwik, line 549: We should return null here and leave a todo BEAM-10303 for window observing optimization.

Because of window exploding, we'll only split on the current window which will return the wrong primary/residual since it won't take into account these other windows that have been fully processed or yet to be started.

sdks/java/harness/src/main/java/org/apache/beam/fn/harness/FnApiDoFnRunner.java
```diff
@@ -488,90 +493,70 @@
...
+                }
+
+                @Override
+                public double getProgress() {
+                  return splitDelegate.getProgress();
```
lukecwik, line 554: return null here as well

sdks/java/harness/src/test/java/org/apache/beam/fn/harness/FnApiDoFnRunnerTest.java
```diff
@@ -2892,4 +2899,492 @@ public void testProcessElementForWindowedSplitAndSizeRestriction() throws Except
...
+    }
+
+    @Override
+    public void accept(WindowedValue input) throws Exception {
+      mainOutputValues.add(input);
+    }
+  }
+
+  @Test
+  public void testProcessElementForTruncateAndSizeRestrictionForwardSplitAndProgress()
```
lukecwik, line 2948: Please make this test the non window observing variant and add a window observing one which returns null instead of progress/split.

runners/core-construction-java/src/main/java/org/apache/beam/runners/core/construction/graph/SplittableParDoExpander.java
```diff
@@ -273,4 +296,211 @@ private static String generateUniqueId(String prefix, Predicate<String> isExisti
     }
     return prefix + i;
   }
+
+  /** See {@link #createTruncateReplacement()} ()} for details. */
+  private static class TruncateReplacement extends SizedReplacement {
```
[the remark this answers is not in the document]
boyuanzz: Nope, I thought some util functions were from `SizedReplacement ` but they are not.

boyuanzz:
'm going to split this PR into java changes and python changes since it has grown too large.

Here is the java one: https://github.com/apache/beam/pull/12287

boyuanzz:
Opened https://github.com/apache/beam/pull/12289 one. I'm going to close this PR. Thanks for all comments and discussions.

PR #12022, opened 2020-06-16
[BEAM-10135] Add Jdbc Write external transform

piotr-szuberski:
Add External transform for Jdbc write.

------------------------

Thank you for your contribution! Follow this checklist to help us incorporate your contribution quickly and easily:

 - [ ] [**Choose reviewer(s)**](https://beam.apache.org/contribute/#make-your-change) and mention them in a comment (`R: @username`).
 - [ ] Format the pull request title like `[BEAM-XXX] Fixes bug in ApproximateQuantiles`, where you replace `BEAM-XXX` with the appropriate JIRA issue, if applicable. This will automatically link the pull request to the issue.
 - [ ] Update `CHANGES.md` with noteworthy changes.
 - [ ] If this contribution is large, please file an Apache [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

See the [Contributor Guide](https://beam.apache.org/contribute) for more tips on [how to make review process smoother](https://beam.apache.org/contribute/#make-reviewers-job-easier).

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

Lang | SDK | Apex | Dataflow | Flink | Samza | Spark
--- | --- | --- | --- | --- | --- | ---
Go | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go/lastCompletedBuild/) | --- | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_VR_Flink/lastCompletedBuild/) | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_VR_Spark/lastCompletedBuild/)
[...]

PR #12023, opened 2020-06-16
[BEAM-10135] Add Python wrapper for Jdbc Write external transform

piotr-szuberski:
Add python wrapper for jdbc write transform

Note: This is a very limited transform - it can't insert timestamp, boolean, bytes as the actual row_coder is the limitation here.

This PR is dependent on #12022

------------------------

Lang | SDK | Apex | Dataflow | Flink | Samza | Spark
Go | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go/lastCompletedBuild/) | --- | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_VR_Flink/lastCompletedBuild/) | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go_VR_Spark/lastCompletedBuild/)
Java | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Apex/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Dataflow/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Dataflow_Java11/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Flink/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Flink_Java11/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Flink_Batch/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Flink_Streaming/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Samza/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Spark/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Spark_Batch/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_SparkStructuredStreaming/lastCompletedBuild/)
Python | [Build Status](https://builds.apache.org/job/beam_PostCommit_Python2/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Python35/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Python36/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Python37/lastCompletedBuild/) | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Py_VR_Dataflow/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Py_VR_Dataflow_V2/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Py_ValCont/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PreCommit_Python2_PVR_Flink_Cron/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Python35_VR_Flink/lastCompletedBuild/) | --- | [Build Status](https://builds.apache.org/job/beam_PostCommit_Python_VR_Spark/lastCompletedBuild/)

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/io/external/jdbc.py
+++ b/sdks/python/apache_beam/io/external/jdbc.py
@@ -61,9 +61,9 @@ WriteToJdbcSchema = typing.NamedTuple(
         ('jdbc_url', unicode),
         ('username', unicode),
         ('password', unicode),
-        ('statement', unicode),
-        ('connection_properties', unicode),
+        ('connection_properties', typing.Optional[unicode]),
         ('connection_init_sqls', typing.Optional[typing.List[unicode]]),
+        ('statement', unicode),
     ],
 )
 
@@ -101,10 +101,7 @@ class WriteToJdbc(ExternalTransform):
       username,
       password,
       statement,
-      connection_properties='',
-      connection_init_sqls=None,
-      expansion_service=None,
-  ):
+      **kwargs):
     """
     Initializes a write operation to Jdbc.
 
@@ -117,6 +114,10 @@ class WriteToJdbc(ExternalTransform):
     :param connection_init_sqls: required only for MySql and MariaDB
     :param expansion_service: The address (host:port) of the ExpansionService.
     """
+    connection_properties = kwargs.get('connection_properties')
+    connection_init_sqls = kwargs.get('connection_init_sqls')
+    expansion_service = kwargs.get('expansion_service')
+
     super(WriteToJdbc, self).__init__(
         self.URN,
         NamedTupleBasedPayloadBuilder(
--- a/sdks/python/apache_beam/io/external/jdbc.py
+++ b/sdks/python/apache_beam/io/external/jdbc.py
@@ -1,0 +1,134 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""
+  PTransforms for supporting Jdbc in Python pipelines. These transforms do not
+  run a Jdbc client in Python. Instead, they expand to ExternalTransforms
+  which the Expansion Service resolves to the Java SDK's JdbcIO.
+
+  Note: To use these transforms, you need to start a Java Expansion Service.
+  Please refer to the portability documentation on how to do that. Flink Users
+  can use the built-in Expansion Service of the Flink Runner's Job Server. The
+  expansion service address has to be provided when instantiating the
+  transforms.
+
+  If you start Flink's Job Server, the expansion service will be started on
+  port 8097. This is also the configured default for this transform. For a
+  different address, please set the expansion_service parameter.
+
+  For more information see:
+  - https://beam.apache.org/documentation/runners/flink/
+  - https://beam.apache.org/roadmap/portability/
+"""
+
+# pytype: skip-file
+
+from __future__ import absolute_import
+
+import typing
+
+from past.builtins import unicode
+
+from apache_beam.transforms.external import BeamJarExpansionService
+from apache_beam.transforms.external import ExternalTransform
+from apache_beam.transforms.external import NamedTupleBasedPayloadBuilder
+
+__all__ = ['WriteToJdbc']
+
+
+def default_io_expansion_service():
+  return BeamJarExpansionService('sdks:java:io:expansion-service:shadowJar')
+
+
+WriteToJdbcSchema = typing.NamedTuple(
+    'WriteToJdbcSchema',
+    [
+        ('driver_class_name', unicode),
+        ('jdbc_url', unicode),
+        ('username', unicode),
+        ('password', unicode),
+        ('statement', unicode),
+        ('connection_properties', unicode),
+        ('connection_init_sqls', typing.Optional[typing.List[unicode]]),
+    ],
+)
+
+
+class WriteToJdbc(ExternalTransform):
+  """
+  An external PTransform which writes Rows to the specified database.
+
+  This transform receives Rows defined as NamedTuple type and registered in
+  the coders registry to use RowCoder, e.g.
+
+    import typing
+    from apache_beam import coders
+
+    ExampleRow = typing.NamedTuple(
+      "ExampleRow",
+      [
+        ("id", int),
+        ("name", unicode),
+        ("budget", float),
+      ],
+    )
+    coders.registry.register_coder(ExampleRow, coders.RowCoder)
+
+  Experimental; no backwards compatibility guarantees.  It requires special
+  preparation of the Java SDK.  See BEAM-7870.
+  """
+
+  URN = 'beam:external:java:jdbc:write:v1'
+
+  def __init__(
+      self,
+      driver_class_name,
+      jdbc_url,
+      username,
+      password,
+      statement,
+      connection_properties='',
+      connection_init_sqls=None,
+      expansion_service=None,
+  ):
+    """
+    Initializes a write operation to Jdbc.
+
+    :param driver_class_name: name of the jdbc driver class
+    :param jdbc_url: full jdbc url to the database.
+    :param username: database username
+    :param password: database password
+    :param statement: sql statement to be executed
+    :param connection_properties: properties of the jdbc connection
+    :param connection_init_sqls: required only for MySql and MariaDB
+    :param expansion_service: The address (host:port) of the ExpansionService.
+    """
+    super(WriteToJdbc, self).__init__(
+        self.URN,
+        NamedTupleBasedPayloadBuilder(
+            WriteToJdbcSchema(
+                driver_class_name=driver_class_name,
+                jdbc_url=jdbc_url,
+                username=username,
+                password=password,
+                statement=statement,
+                connection_properties=connection_properties,
+                connection_init_sqls=connection_init_sqls,
+            ),
+        ),
+        expansion_service or default_io_expansion_service(),
+    )
--- a/sdks/python/apache_beam/io/external/xlang_jdbcio_it_test.py
+++ b/sdks/python/apache_beam/io/external/xlang_jdbcio_it_test.py
@@ -78,13 +78,6 @@ class JdbcExternalTransformTest(unittest.TestCase):
   """
   ROW_COUNT = 10
 
-  pipeline_options = {
-      'runner': 'FlinkRunner',
-      'flink_version': '1.10',
-      'experiment': 'beam_fn_api',
-      'environment_type': 'LOOPBACK'
-  }
-
   def test_xlang_jdbc_write(self):
     table_name = 'jdbc_external_test_write'
     self.engine.execute(
--- a/sdks/python/apache_beam/io/external/xlang_jdbcio_it_test.py
+++ b/sdks/python/apache_beam/io/external/xlang_jdbcio_it_test.py
@@ -23,6 +23,7 @@ import logging
 import typing
 import unittest
 
+from nose.plugins.attrib import attr
 from past.builtins import unicode
 
 import apache_beam as beam
@@ -55,7 +56,7 @@ JdbcTestRow = typing.NamedTuple(
 
 coders.registry.register_coder(JdbcTestRow, coders.RowCoder)
 
-
+@attr('UsesCrossLanguageTransforms')
 @unittest.skipIf(sqlalchemy is None, "sql alchemy package is not installed.")
 @unittest.skipIf(
     PostgresContainer is None, "testcontainers package is not installed")
--- a/sdks/python/apache_beam/io/external/xlang_jdbcio_it_test.py
+++ b/sdks/python/apache_beam/io/external/xlang_jdbcio_it_test.py
@@ -29,6 +29,7 @@ from past.builtins import unicode
 import apache_beam as beam
 from apache_beam import coders
 from apache_beam.io.external.jdbc import WriteToJdbc
+from apache_beam.options.pipeline_options import PipelineOptions
 from apache_beam.testing.test_pipeline import TestPipeline
 
 # pylint: disable=wrong-import-order, wrong-import-position, ungrouped-imports
@@ -45,16 +46,18 @@ except ImportError:
   PostgresContainer = None
 # pylint: enable=wrong-import-order, wrong-import-position, ungrouped-imports
 
-JdbcTestRow = typing.NamedTuple(
-    "JdbcTestRow",
+ROW_COUNT = 10
+
+JdbcWriteTestRow = typing.NamedTuple(
+    "JdbcWriteTestRow",
     [
         ("f_id", int),
         ("f_real", float),
         ("f_string", unicode),
     ],
 )
+coders.registry.register_coder(JdbcWriteTestRow, coders.RowCoder)
 
-coders.registry.register_coder(JdbcTestRow, coders.RowCoder)
 
 @attr('UsesCrossLanguageTransforms')
 @unittest.skipIf(sqlalchemy is None, "sql alchemy package is not installed.")
@@ -84,13 +87,14 @@ class JdbcExternalTransformTest(unittest.TestCase):
         "CREATE TABLE {}(f_id INTEGER, f_real REAL, f_string VARCHAR)".format(
             table_name))
     inserted_rows = [
-        JdbcTestRow(i, float(i + 0.1), str(i)) for i in range(self.ROW_COUNT)
+        JdbcWriteTestRow(i, i + 0.1, 'Test{}'.format(i))
+        for i in range(ROW_COUNT)
     ]
     with TestPipeline() as p:
       p.not_use_test_runner_api = True
       _ = (
           p
-          | beam.Create(inserted_rows).with_output_types(JdbcTestRow)
+          | beam.Create(inserted_rows).with_output_types(JdbcWriteTestRow)
           | 'Write to jdbc' >> WriteToJdbc(
               driver_class_name=self.driver_class_name,
               jdbc_url=self.jdbc_url,
@@ -101,7 +105,7 @@ class JdbcExternalTransformTest(unittest.TestCase):
           ))
     fetched_data = self.engine.execute("select * from {}".format(table_name))
     fetched_rows = [
-        JdbcTestRow(int(row[0]), float(row[1]), str(row[2]))
+        JdbcWriteTestRow(int(row[0]), float(row[1]), str(row[2]))
         for row in fetched_data
     ]
 
--- a/sdks/python/apache_beam/io/external/xlang_jdbcio_it_test.py
+++ b/sdks/python/apache_beam/io/external/xlang_jdbcio_it_test.py
@@ -1,0 +1,138 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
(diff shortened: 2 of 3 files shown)


sdks/python/apache_beam/io/external/jdbc.py
```diff
@@ -0,0 +1,116 @@
...
+  which the Expansion Service resolves to the Java SDK's JdbcIO.
+
+  Note: To use these transforms, you need to start a Java Expansion Service.
+  Please refer to the portability documentation on how to do that. Flink Users
+  can use the built-in Expansion Service of the Flink Runner's Job Server. The
+  expansion service address has to be provided when instantiating the
+  transforms.
+
+  If you start Flink's Job Server, the expansion service will be started on
+  port 8097. This is also the configured default for this transform. For a
```
TheNeuralBit, line 30: It looks to me like the default is to start an expansion service with `sdks:java:io:expansion-service:shadowJar`, not to use port 8097, is this out of date?

sdks/python/apache_beam/io/external/jdbc.py
```diff
@@ -0,0 +1,116 @@
...
+"""
+  PTransforms for supporting Jdbc in Python pipelines. These transforms do not
+  run a Jdbc client in Python. Instead, they expand to ExternalTransforms
+  which the Expansion Service resolves to the Java SDK's JdbcIO.
+
+  Note: To use these transforms, you need to start a Java Expansion Service.
+  Please refer to the portability documentation on how to do that. Flink Users
+  can use the built-in Expansion Service of the Flink Runner's Job Server. The
+  expansion service address has to be provided when instantiating the
+  transforms.
```
TheNeuralBit, line 27: I don't think we should indicate this is an ExternalTransform in the module docstring, it should be transparent to the user (I received similar guidance for the xlang SqlTransform: https://github.com/apache/beam/blob/master/sdks/python/apache_beam/transforms/sql.py).

This is useful information though, maybe it could be moved to the docstring for WriteToJdbc.__init__ to document the `expansion_service` param?

sdks/python/apache_beam/io/external/xlang_jdbcio_it_test.py
```diff
@@ -0,0 +1,165 @@
...
+  To run with the local expansion service and flink job server you need to build it,
+  e.g. via command:
+  ./gradlew :sdks:java:io:expansion-service:shadowJar
+  ./gradlew :runners:flink:1.10:job-server:shadowJar
+  and have flink1.10 master cluster running
+
+  If on development branch, it may be necessary to build java sdk docker image and tag
+  it with the latest released version.
+  ./gradlew :sdks:java:container:docker
+  docker tag apache/beam_java_sdk:x.xx.x.dev apache/beam_java_sdk:y.yy.y.dev
```
TheNeuralBit, line 67: I don't think this `docker tag` step should be necessary. At least I didn't need to do it when testing SqlTransform

sdks/python/apache_beam/io/external/xlang_jdbcio_it_test.py
```diff
@@ -0,0 +1,165 @@
...
+    #  Required for table initialization and drop
+    self.host = pipeline.get_option('host')
+    self.port = pipeline.get_option('port')
+    self.database = pipeline.get_option('database')
+
+    #  The actual pipeline options
+    self.pipeline_options = {
+        'runner': 'FlinkRunner',
+        'flink_version': '1.10',
+        'flink_master': 'localhost:8081',
```
TheNeuralBit, line 99: If you specify FlinkRunner without specifying a flink_master a Flink instance should be started for you.

sdks/python/apache_beam/io/external/xlang_jdbcio_it_test.py
```diff
@@ -0,0 +1,165 @@
...
+  If on development branch, it may be necessary to build java sdk docker image and tag
+  it with the latest released version.
+  ./gradlew :sdks:java:container:docker
+  docker tag apache/beam_java_sdk:x.xx.x.dev apache/beam_java_sdk:y.yy.y.dev
+
+  python setup.py nosetests \
+  --tests=apache_beam.io.external.jdbc_test:JdbcExternalTransformTest.test_read_and_write \
+  --test-pipeline-options="
+    --driver_class_name=org.postgresql.Driver
+    --jdbc_url=jdbc:postgresql://localhost:5432/postgres
```
TheNeuralBit, line 73: This would require the user to setup an instance of postgres with these parameters right? I think you should have this test create the postgres instance it needs, I bet there are python libraries that can help with that, e.g. https://pypi.org/project/pytest-postgresql/

piotr-szuberski:
@TheNeuralBit Thanks for the review! I'll take care of it on monday, unfortunately i can't do it earlier.

sdks/python/apache_beam/io/external/jdbc.py
piotr-szuberski: There are 2 approaches.
1. Start flink job server using `sdks:java:io:expansion-service:runShadow` which runs expansion server at localhost:8097
2. Create shadow-jars for expansion service and flink job server via gradle `sdks:java:io:expansion-service:shadowJar`, `runners:flink:1.10:job-server:shadowJar` and leave expansion_service as None, then those jars will be used with random ports.
I din't dig into it but in 2) at first it tries to run expansion service and when it fails it runs the flink job server.

My main example for the code I wrote was kafka.py and generate_sequence external transforms

sdks/python/apache_beam/io/external/xlang_jdbcio_it_test.py
piotr-szuberski: Cool! I missed that info when reading portability docs.

sdks/python/apache_beam/io/external/xlang_jdbcio_it_test.py
piotr-szuberski: I didn't think of it as I thought that there is some postgres instance running where this test could be executed. I will take a look how to easily instantiate the postgres image and try to make this test click and run.

sdks/python/apache_beam/io/external/xlang_jdbcio_it_test.py
piotr-szuberski: Without it I get ```Caused by: org.apache.beam.vendor.guava.v26_0_jre.com.google.common.util.concurrent.UncheckedExecutionException: java.io.IOException: Received exit code 125 for command 'docker run -d --mount type=bind,src=/Users/piotr/.config/gcloud,dst=/root/.config/gcloud --network=host --env=DOCKER_MAC_CONTAINER=null apache/beam_java_sdk:2.21.0.dev --id=2-10 --provision_endpoint=host.docker.internal:51317'. stderr: Unable to find image 'apache/beam_java_sdk:2.21.0.dev' locallydocker: Error response from daemon: manifest for apache/beam_java_sdk:2.21.0.dev not found: manifest unknown: manifest unknown.See 'docker run --help'.```

sdks/python/apache_beam/io/external/jdbc.py
piotr-szuberski: You can run the flink job server with ```runners:flink:1.10:job-server:runShadow``` - the expansion server will be available at port 8097.

On the other hand when you leave expansion_service as None, then ```sdks:java:io:expansion-service:shadowJar``` will be taken with a random port. But I found that flink job server's jar is still needed for it to work.

sdks/python/apache_beam/io/external/jdbc.py
piotr-szuberski: I found it was done similarly in KafkaIO. If I move it to Write transform, then it should also be in Read transform which is the next step of this task.

sdks/python/apache_beam/io/external/xlang_jdbcio_it_test.py
piotr-szuberski: I used https://github.com/testcontainers/testcontainers-python and now postgres instance is instantiated automatically. Now the test can be run without any params

sdks/python/apache_beam/io/external/xlang_jdbcio_it_test.py
piotr-szuberski: Ok, I rebased and now it uses 2.23.0 version. I think this info indeed is not that necessary

piotr-szuberski:
@TheNeuralBit I've made the requested changes. If you have further suggestions - go ahead :)

piotr-szuberski:
Run Java PreCommit

piotr-szuberski:
Run Python PreCommit

piotr-szuberski:
> Run Java PreCommit

I was just curious if I can run it without a commiter privileges. Hooray new Jenkins!

piotr-szuberski:
Run Python PreCommit

piotr-szuberski:
Run Python PreCommit

piotr-szuberski:
Run Python PreCommit

piotr-szuberski:
Run Python PreCommit

piotr-szuberski:
Run Python PreCommit

piotr-szuberski:
Run Python PreCommit

piotr-szuberski:
@TheNeuralBit ping

sdks/python/apache_beam/io/external/jdbc.py
```diff
@@ -0,0 +1,134 @@
...
+      self,
+      driver_class_name,
+      jdbc_url,
+      username,
+      password,
+      statement,
+      connection_properties='',
+      connection_init_sqls=None,
+      expansion_service=None,
+  ):
```
TheNeuralBit, line 107: Could you make these parameters into keyword args? It's easy for a user to get lost in a long list of function parameters, but the keywords args will make it explicit. The only cost is you'll need explicit checks the the required parameters aren't `None`.

sdks/python/apache_beam/io/external/xlang_jdbcio_it_test.py
```diff
@@ -0,0 +1,138 @@
...
+        ("f_string", unicode),
+    ],
+)
+
+coders.registry.register_coder(JdbcTestRow, coders.RowCoder)
+
+
+@unittest.skipIf(sqlalchemy is None, "sql alchemy package is not installed.")
+@unittest.skipIf(
+    PostgresContainer is None, "testcontainers package is not installed")
```
TheNeuralBit, line 61: I think we should add these packages to the test extra: https://github.com/apache/beam/blob/92170e85aa6284ad3908656ecc2d6f31505d0310/sdks/python/setup.py#L181-L198

cc: @aaltay in case there are any concerns with adding these dependencies. For licenses they're both under Apache 2.0 (and it's also just a test dependency).

sdks/python/apache_beam/io/external/jdbc.py
TheNeuralBit: But where does the code for (1) live on the python side?

As far as I can tell the only two options for this transform are either overriding the expansion service with the `expansion_service=` parameter, or the default - running shadowJar with `BeamJarExpansionService` (what you have in (2)). I can't find any logic that looks for an expansion service on port 8097.  Am I missing something?

sdks/python/apache_beam/io/external/xlang_jdbcio_it_test.py
aaltay: Thank you. I agree. It is fine to add as test only dependencies.

sdks/python/apache_beam/io/external/jdbc.py
piotr-szuberski: On the python side you can add a parameter expansion_service='localhost:8097' and it will use this address instead of the one from BeamJarExpansionService.
In most cases a user of the transform will start a container with flink job server which exposes localhost:8097 and use this address.

sdks/python/apache_beam/io/external/xlang_jdbcio_it_test.py
piotr-szuberski: Is it ok if the tests will run for python >=3.5 only? testcontainers don't support python2, as well as pytest-postgresql.

sdks/python/apache_beam/io/external/jdbc.py
piotr-szuberski: Aren't kwargs more confusing to the user and forces him to look into documentation? How about leaving the required params as they are and use kwargs for the optional ones?
For me it's always useful that IDE tells me what params are yet to pass.
I'm more a Java than Python guy so I may be wrong.

sdks/python/apache_beam/io/external/xlang_jdbcio_it_test.py
TheNeuralBit: I think that's fine, you can just add a skipIf based on the python version

piotr-szuberski:
@TheNeuralBit  I've made the requested changes - I've made only optional params to be kwargs - if you think that all params should be kwargs then I'll change it.
I wonder whether putting this change as separate PRs was a good idea.
JdbcRead is ready - I'll make a PR rebased on those 2 PRs and mark them as WIP

sdks/python/apache_beam/io/external/jdbc.py
TheNeuralBit: Hm so I was under the impression that in Python positional arguments had to be passed as positional args, I didn't realize that callers are allowed to name them as you've done in the test:
```python
          | 'Write to jdbc' >> WriteToJdbc(
              driver_class_name=self.driver_class_name,
              jdbc_url=self.jdbc_url,
              username=self.username,
              password=self.password,
              statement='INSERT INTO {} VALUES(?, ?, ?)'.format(table_name),
          ))
```
So thanks for teaching me something new! :)

My concern was that users would end up passing a long parameter list of all primitive types, which can be hard to keep track of (e.g. in Java you'd use a Builder rather than having a long parameter list like that). But since users can just name the args as you've done that alleviates my concern.

sdks/python/apache_beam/io/external/jdbc.py
TheNeuralBit: Ok, but my concern is that this docstring indicates using the expansion service at localhost:8097 is the "configured default for this transform," which doesn't seem to be the case. It looks to me like the default is to use `BeamJarExpansionService('sdks:java:io:expansion-service:shadowJar')`, which will start up an expansion service in a subprocess. 

I think that's a great default and should be used in the vast majority of cases, I don't think there's any reason to build JdbcIO into the flink expansion service and encourage users to reference that. My understanding is the flink expansion service was just a temporary solution before we had the ability to run expansion services with `BeamJarExpansionService`, so there's no need to use it anymore (@chamikaramj, @robertwb to confirm or correct me).

I think you should:
- [ ] Update the docstring to indicate the default is to use `sdks:java:io:expansion-service:shadowJar`, or even not reference it at all since it should happen transparently for the user.
- [ ] Drop mention of the Flink job server and it's expansion service.

TheNeuralBit:
> I've made the requested changes - I've made only optional params to be kwargs - if you think that all params should be kwargs then I'll change it.

That sounds great. I'm fine with the required params being positional args, given what I discussed in https://github.com/apache/beam/pull/12023#discussion_r447799680. Sorry for the misunderstanding

> JdbcRead is ready - I'll make a PR rebased on those 2 PRs and mark them as WIP

Yeah it was helpful to have the separation of concerns when reviewing, but I agree I think it will be better to have a single PR that we can test together.

sdks/python/apache_beam/io/external/xlang_jdbcio_it_test.py
TheNeuralBit: Thanks!

sdks/python/apache_beam/io/external/jdbc.py
piotr-szuberski: Oh ok, I thought that you meant using **kwargs. It confused me but since I'm not familiar with python tricks I thought that there is an advantage of using it. But if you only meant passing named parameters then I'll switch back to the default values as then a user has IDE suggestions.

sdks/python/apache_beam/io/external/xlang_jdbcio_it_test.py
piotr-szuberski: I left those skipIfs and added testcontainers only for python3 ```testcontainers<3.0.3; python_version >= "3.5"'```

sdks/python/apache_beam/io/external/jdbc.py
piotr-szuberski: I can see now where the misunderstanding came from - I was certain that the default is to run ```docker run --net=host apache/beam_flink1.10_job_server:latest``` as is in flink runner docs for portable setup. But it limited the user to FlinkRunner which in the future will be not sufficient.

Really cool that now ```BeamJarExpansionService``` is the default.

piotr-szuberski:
@TheNeuralBit  I've created a separate PR with write and read  #12145

sdks/python/apache_beam/io/external/jdbc.py
piotr-szuberski: I've removed most of those information as when the default is to use BeamJarExpansionService(sdks:java:io:expansion-service:shadowJar) they are redundant.

chamikaramj: We updated Kafka to build or download the IO expansion jar by default. Probably you can follow the same model here. I updated documentation for Kafka recently: https://github.com/apache/beam/pull/11928/files

BTW one downside with a single IO expansion jar is that the size of the jar increases linearly with the number of cross-language wrappers that we have. This will increase the pipeline setup time for users since a larger jar has to be staged. What is the size increase of IO expansion jar due to this wrapper ? 

cc: @robertwb

sdks/python/apache_beam/io/external/jdbc.py
```diff
@@ -0,0 +1,135 @@
...
+
+import typing
+
+from past.builtins import unicode
+
+from apache_beam.transforms.external import BeamJarExpansionService
+from apache_beam.transforms.external import ExternalTransform
+from apache_beam.transforms.external import NamedTupleBasedPayloadBuilder
+
+__all__ = ['WriteToJdbc']
```
chamikaramj, line 50: Just curious. Were you able to successfully test this with portable Flink/Spark and Dataflow ?

sdks/python/apache_beam/io/external/jdbc.py
```diff
@@ -0,0 +1,135 @@
...
+      driver_class_name,
+      jdbc_url,
+      username,
+      password,
+      statement,
+      **kwargs):
+    """
+    Initializes a write operation to Jdbc.
+
+    :param driver_class_name: name of the jdbc driver class
```
chamikaramj, line 108: Should these parameters be specified as strings ? Probably good if we can clarify that here.

piotr-szuberski: I added those changes in the PR #12145

sdks/python/apache_beam/io/external/jdbc.py
piotr-szuberski: The jar is about 1Mb larger.

piotr-szuberski: Yes, with FlinkRunner, SparkRunner and DirectRunner. On Dataflow with sdk_location and dataflow_worker_jar set I wasn't successful, got `RuntimeError: java.lang.UnsupportedOperationException: Unknown urn: beam:external:java:jdbc:write:v1`

sdks/python/apache_beam/io/external/jdbc.py
piotr-szuberski: The Jar is about 1Mb larger

robertwb: Staging time is a valid concern; thanks for calling this out. I think 1MB is a small price to pay for the convenience here--we could have dozens of wrappers here of that size without meaningfully affecting staging time. If it were 10s of MB, or we were getting into a long tail of obscure IOs, that starts to be a different story.

chamikaramj: Yeah, 1 MB is fine. I think we should continue to monitor the size of this dependency in future Python wrappers that add to it though to make sure that the size does not get increased significantly.

PR #12050, opened 2020-06-22
[BEAM-10292] DefaultFilenamePolicy.ParamsCoder is not able to decode directory on the local file system

davidak09:
`DefaultFilenamePolicy.ParamsCoder` used `StringUtf8Coder` for encoding/decoding `baseFilename` and therefore information whether `baseFilename resourceID` is file or directory was lost. Using  `ResourceIdCoder` instead fixes this issue.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/io/DefaultFilenamePolicy.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/io/DefaultFilenamePolicy.java
@@ -171,7 +171,7 @@ public final class DefaultFilenamePolicy extends FilenamePolicy {
   /** A Coder for {@link Params}. */
   public static class ParamsCoder extends AtomicCoder<Params> {
     private static final ParamsCoder INSTANCE = new ParamsCoder();
-    private Coder<String> stringCoder = StringUtf8Coder.of();
+    private static final Coder<String> STRING_CODER = StringUtf8Coder.of();
 
     public static ParamsCoder of() {
       return INSTANCE;
@@ -182,17 +182,17 @@ public final class DefaultFilenamePolicy extends FilenamePolicy {
       if (value == null) {
         throw new CoderException("cannot encode a null value");
       }
-      stringCoder.encode(value.baseFilename.get().toString(), outStream);
-      stringCoder.encode(value.shardTemplate, outStream);
-      stringCoder.encode(value.suffix, outStream);
+      STRING_CODER.encode(value.baseFilename.get().toString(), outStream);
+      STRING_CODER.encode(value.shardTemplate, outStream);
+      STRING_CODER.encode(value.suffix, outStream);
     }
 
     @Override
     public Params decode(InputStream inStream) throws IOException {
       ResourceId prefix =
-          FileBasedSink.convertToFileResourceIfPossible(stringCoder.decode(inStream));
-      String shardTemplate = stringCoder.decode(inStream);
-      String suffix = stringCoder.decode(inStream);
+          FileBasedSink.convertToFileResourceIfPossible(STRING_CODER.decode(inStream));
+      String shardTemplate = STRING_CODER.decode(inStream);
+      String suffix = STRING_CODER.decode(inStream);
       return new Params()
           .withBaseFilename(prefix)
           .withShardTemplate(shardTemplate)
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/io/LocalFileSystem.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/io/LocalFileSystem.java
@@ -202,9 +202,15 @@ class LocalFileSystem extends FileSystem<LocalResourceId> {
 
   @Override
   protected LocalResourceId matchNewResource(String singleResourceSpec, boolean isDirectory) {
-    if (singleResourceSpec.endsWith(File.separator) && !isDirectory) {
-      throw new IllegalArgumentException(
-          String.format("Expected file path but received directory path %s", singleResourceSpec));
+    if (isDirectory) {
+      if (!singleResourceSpec.endsWith(File.separator)) {
+        singleResourceSpec += File.separator;
+      }
+    } else {
+      checkArgument(
+          !singleResourceSpec.endsWith(File.separator),
+          "Expected file path but received directory path [%s].",
+          singleResourceSpec);
     }
     Path path = Paths.get(singleResourceSpec);
     return LocalResourceId.fromPath(path, isDirectory);
--- a/sdks/java/core/src/main/java/org/apache/beam/sdk/io/LocalFileSystem.java
+++ b/sdks/java/core/src/main/java/org/apache/beam/sdk/io/LocalFileSystem.java
@@ -202,6 +202,10 @@ class LocalFileSystem extends FileSystem<LocalResourceId> {
 
   @Override
   protected LocalResourceId matchNewResource(String singleResourceSpec, boolean isDirectory) {
+    if (singleResourceSpec.endsWith(File.separator) && !isDirectory) {
+      throw new IllegalArgumentException(
+          String.format("Expected file path but received directory path %s", singleResourceSpec));
+    }
     Path path = Paths.get(singleResourceSpec);
     return LocalResourceId.fromPath(path, isDirectory);
   }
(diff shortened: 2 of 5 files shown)


davidak09:
R: @dmvk

davidak09:
Thanks for the review @dmvk !

You're right, I forgot to ensure backward compatibility. I tried to implement the requested changes, will you take a look at it please?

dmvk:
Run Java PreCommit

sdks/java/core/src/main/java/org/apache/beam/sdk/io/DefaultFilenamePolicy.java
```diff
@@ -182,19 +184,26 @@ public void encode(Params value, OutputStream outStream) throws IOException {
...
-      String suffix = stringCoder.decode(inStream);
+      String prefix = STRING_CODER.decode(inStream);
+      String shardTemplate = STRING_CODER.decode(inStream);
+      String suffix = STRING_CODER.decode(inStream);
+      ResourceId baseFilename;
+      if (inStream.available() > 0) {
+        baseFilename = FileSystems.matchNewResource(prefix, BOOLEAN_CODER.decode(inStream));
+      } else {
+        // fallback for ensure backward compatibility
+        baseFilename = FileBasedSink.convertToFileResourceIfPossible(prefix);
```
dmvk, line 203: Would it be possible to add test case for bw compatible code path?

davidak09: ok, I added the requested test for bw compatibility

btw you can see the buggy behavior there (lost information whether baseFilename is file/directory)

lukecwik: convertToFileResourceIfPossible attempts to match a file and if that fails attempts to match a directory.

Is the issue that the underlying filesystem erroneously says something is a file when really it is a directory?

davidak09: Exactly, that's the issue, for example `LocalFileSystem` matches path `/tmp/dirA/` as file..

How about something like this?
```java
public static class ParamsCoder extends AtomicCoder<Params> {
    private static final ParamsCoder INSTANCE = new ParamsCoder();
    private static final Coder<String> STRING_CODER = StringUtf8Coder.of();

    public static ParamsCoder of() {
      return INSTANCE;
    }

    @Override
    public void encode(Params value, OutputStream outStream) throws IOException {
      if (value == null) {
        throw new CoderException("cannot encode a null value");
      }
      ResourceId baseFilename = value.getBaseFilename().get();
      String baseFilenameString =
          (baseFilename.isDirectory() && !baseFilename.toString().endsWith("/"))
              ? baseFilename.toString() + "/"
              : baseFilename.toString();
      STRING_CODER.encode(baseFilenameString, outStream);
      STRING_CODER.encode(value.getShardTemplate(), outStream);
      STRING_CODER.encode(value.getSuffix(), outStream);
    }

    @Override
    public Params decode(InputStream inStream) throws IOException {
      String baseFilenameString = STRING_CODER.decode(inStream);
      ResourceId prefix =
          FileSystems.matchNewResource(baseFilenameString, baseFilenameString.endsWith("/"));
      String shardTemplate = STRING_CODER.decode(inStream);
      String suffix = STRING_CODER.decode(inStream);
      return new Params()
          .withBaseFilename(prefix)
          .withShardTemplate(shardTemplate)
          .withSuffix(suffix);
    }
  }
```
But I'm not sure how to handle the suffix (`/`) to ensure compatibility among (file) systems

lukecwik: I think that there are a couple of options here:
1) Take the breaking change to the coder because it is a fix, make sure it is documented in the release notes
2) Try to fix the underlying filesystem to do a better job of file/dir matching
3) Deprecate this filename policy, create a new one (DefaultFilenamePolicy2) and tell people to use it in new code.

I'm for #1 but would ask for consensus on the mailing list.

davidak09: OK, I agree with the first solution - taking the breaking change. I will ask on the mailing list.

BTW I did workaround in my project where I use only files (not directories) as `Params` base filenames and then everything works fine. That means I no longer needs this fix but it still seems to me like a bug which is worth fixing.

(P.S.: The suffix wouldn't have to be the path separator (`/`) but e.g. `sometotallyrandomstring123`  - but it's very ugly...)

davidak09: https://lists.apache.org/thread.html/r1d844830648b88948d38d0c28d106dd7fe2ed92af4a170e2679f8c2f%40%3Cdev.beam.apache.org%3E

davidak09: Hi @lukecwik, finally after the discussion in the mailing list I chose the second proposed option and fixed the underlying filesystem. Turns out that only `LocalFileSystem` is broken, on the other hand e.g. HDFS and S3 look fine and already have this kind of check.

I still think that the proper solution should be to use the `ResourceIdCoder` instead of `StringUtf8Coder` but I understand the consequences.

sdks/java/core/src/test/java/org/apache/beam/sdk/io/DefaultFilenamePolicyTest.java
```diff
@@ -48,6 +51,26 @@ private static String constructName(
     return constructed.toString();
   }
 
+  private static DefaultFilenamePolicy.Params encodeDecodeParams(
```
lukecwik, line 54: Use [CoderUtils.clone](https://github.com/apache/beam/blob/f3ac4822191ca63d74a8fd71f81da976b3d2fbd1/sdks/java/core/src/main/java/org/apache/beam/sdk/util/CoderUtils.java#L140)

sdks/java/core/src/main/java/org/apache/beam/sdk/io/LocalFileSystem.java
```diff
@@ -202,6 +202,10 @@ protected void delete(Collection<LocalResourceId> resourceIds) throws IOExceptio
 
   @Override
   protected LocalResourceId matchNewResource(String singleResourceSpec, boolean isDirectory) {
+    if (singleResourceSpec.endsWith(File.separator) && !isDirectory) {
```
lukecwik, line 205: Do you want to add the file separator if it doesn't exist if `isDirectory == true`. This will help make it less error prone for users.

lukecwik, line 205: What if the character is escaped and being used in the name component of the path string?

sdks/java/core/src/test/java/org/apache/beam/sdk/io/DefaultFilenamePolicyTest.java
davidak09: thanks for the advice :+1:

sdks/java/core/src/main/java/org/apache/beam/sdk/io/LocalFileSystem.java
davidak09: sounds good, I added it

davidak09: Is it even possible? For example according to https://stackoverflow.com/questions/9847288/is-it-possible-to-use-in-a-filename it's not. Or do you mean something different?

davidak09:
> We should update LocalResourceId and drop the isDirectory field

I removed the `isDirectory` field, nevertheless I left `isDirectory` parameter in the constructor and the factory method untouched. WDYT?

sdks/java/core/src/main/java/org/apache/beam/sdk/io/LocalFileSystem.java
lukecwik: Your right, it seems as though Linux doesn't allow `/`, Mac OS X doesn't allow `:`, Windows doesn't allow `\`.

lukecwik:
> > We should update LocalResourceId and drop the isDirectory field
> 
> I removed the `isDirectory` field, nevertheless I left `isDirectory` parameter in the constructor and the factory method untouched. WDYT?

Sg

lukecwik:
@dmvk The backwards compat issues have been addressed. Will merge, feel free to add additional comments if there was something that was not addressed.

PR #12059, opened 2020-06-23
Adding a helper function for making custom coders.

youngoli:
------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/go/pkg/beam/coder.go
+++ b/sdks/go/pkg/beam/coder.go
@@ -144,20 +144,20 @@ func inferCoder(t FullType) (*coder.Coder, error) {
 			if err != nil {
 				return nil, err
 			}
-			return &coder.Coder{Kind: coder.Custom, T: t, Custom: c}, nil
+			return coder.CoderFrom(c), nil
 		case reflectx.Uint, reflectx.Uint8, reflectx.Uint16, reflectx.Uint32, reflectx.Uint64:
 			c, err := coderx.NewVarUintZ(t.Type())
 			if err != nil {
 				return nil, err
 			}
-			return &coder.Coder{Kind: coder.Custom, T: t, Custom: c}, nil
+			return coder.CoderFrom(c), nil
 
 		case reflectx.Float32:
 			c, err := coderx.NewFloat(t.Type())
 			if err != nil {
 				return nil, err
 			}
-			return &coder.Coder{Kind: coder.Custom, T: t, Custom: c}, nil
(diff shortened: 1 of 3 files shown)


youngoli:
R: @lostluck 

As an aside, and not in this PR, what do you think about refactoring the locations of all the coder code in the SDK right now. I noticed while trying to figure out how to make some coders that its very unintuitive trying to figure out how to make coders for certain types (for example ints). Right now it seems like built-in coders are all in graph/coder and custom coders for Go types are in runtime/coderx. I was thinking of moving those into the same directory and splitting them into a file for builtin coders, a file for custom coders (for code generation I'm guessing), and a file for helpers for users to create custom coders for various types.

None of that is urgent at all, it's all just stuff that I think makes the SDK a bit more intuitively organized.

lostluck:
1. Not while I'm in the middle of modifying all the coder code for Schemas :D
2. As you've described, there's a reason for the two packages, one is for Beam Model coders, and the other is for *not* beam model coders. That signal is valuable though I agree it's a bit subtle, and not called out very well. It's the kind of thing that I'd have called out in a document of a walkthrough of the Go SDK, but didn't in my talk.  
The value is that it's harder to accidentally use a non-standard coder when one must use a standard one.
By being separate, they also serve as an example of how to write some of these.

I do agree that the runtime folder is probably not the right place for the coderx package though. It's also a bit moot since it's not a package users will interact with.

If we were to get rid of the coderx package, I'd rather the coders move to the main beam package or similar, since they're invoked in inferCoders IIRC?

These are good things to think about, and the whole SDK could use a clear overhaul, but we should have a total plan for that so we can apply them evenly to all facets of the SDK at once, rather than one at a time.

Note, that this would probably need to wait until a beam v3 or similar, as disrupting so many packages would break folks who are importing them for some reason out of the SDK.

youngoli:
> If we were to get rid of the coderx package, I'd rather the coders move to the main beam package or similar, since they're invoked in inferCoders IIRC?

My only objection is that we'd likely want to use those coders in packages further down in the SDK, but importing `beam` from those is a common cause of import cycles from my experience, so I'd prefer putting it in a subpackage instead (and we could always forward it to `beam` if we really want). I think having something like `graph/coder/custom` would be better. It's consolidated in the same directory tree as the builtin coders, but clearly separate since its in a subdirectory, and the package name is still clear (ex. `custom.NewIntCoder()`).

Agreed with everything else, including waiting until a major version update before going through with this.

PR #12063, opened 2020-06-23
[BEAM-10294] using SparkMetricsContainerStepMap for readable metrics presentation in Spark history server UI

davidak09:
Beam metrics in Spark history server are not readable because now they're rendered as protobuffer formatted as JSON where metric's value is stored in `bytes` field.

Using  already existing `SparkMetricsContainerStepMap` which overrides `toString()` method by calling `SparkBeamMetric.renderAll()` method.

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/runners/spark/src/main/java/org/apache/beam/runners/spark/metrics/MetricsAccumulator.java
+++ b/runners/spark/src/main/java/org/apache/beam/runners/spark/metrics/MetricsAccumulator.java
@@ -18,6 +18,7 @@
 package org.apache.beam.runners.spark.metrics;
 
 import java.io.IOException;
+import org.apache.beam.runners.core.metrics.MetricsContainerStepMap;
 import org.apache.beam.runners.spark.SparkPipelineOptions;
 import org.apache.beam.runners.spark.translation.streaming.Checkpoint;
 import org.apache.beam.runners.spark.translation.streaming.Checkpoint.CheckpointDir;
@@ -57,13 +58,13 @@ public class MetricsAccumulator {
               opts.isStreaming()
                   ? Optional.of(new CheckpointDir(opts.getCheckpointDir()))
                   : Optional.absent();
-          SparkMetricsContainerStepMap metricsContainerStepMap = new SparkMetricsContainerStepMap();
+          MetricsContainerStepMap metricsContainerStepMap = new SparkMetricsContainerStepMap();
           MetricsContainerStepMapAccumulator accumulator =
               new MetricsContainerStepMapAccumulator(metricsContainerStepMap);
           jsc.sc().register(accumulator, ACCUMULATOR_NAME);
 
           if (maybeCheckpointDir.isPresent()) {
-            Optional<SparkMetricsContainerStepMap> maybeRecoveredValue =
+            Optional<MetricsContainerStepMap> maybeRecoveredValue =
                 recoverValueFromCheckpoint(jsc, maybeCheckpointDir.get());
             if (maybeRecoveredValue.isPresent()) {
               accumulator = new MetricsContainerStepMapAccumulator(maybeRecoveredValue.get());
@@ -86,7 +87,7 @@ public class MetricsAccumulator {
     }
   }
 
-  private static Optional<SparkMetricsContainerStepMap> recoverValueFromCheckpoint(
+  private static Optional<MetricsContainerStepMap> recoverValueFromCheckpoint(
       JavaSparkContext jsc, CheckpointDir checkpointDir) {
     try {
       Path beamCheckpointPath = checkpointDir.getBeamCheckpointDir();
@@ -120,8 +121,7 @@ public class MetricsAccumulator {
   }
 
   /**
-   * Spark Listener which checkpoints {@link SparkMetricsContainerStepMap} values for
-   * fault-tolerance.
+   * Spark Listener which checkpoints {@link MetricsContainerStepMap} values for fault-tolerance.
    */
   public static class AccumulatorCheckpointingSparkListener extends JavaStreamingListener {
     @Override
--- a/runners/spark/src/main/java/org/apache/beam/runners/spark/metrics/MetricsAccumulator.java
+++ b/runners/spark/src/main/java/org/apache/beam/runners/spark/metrics/MetricsAccumulator.java
@@ -18,7 +18,6 @@
 package org.apache.beam.runners.spark.metrics;
 
 import java.io.IOException;
-import org.apache.beam.runners.core.metrics.MetricsContainerStepMap;
 import org.apache.beam.runners.spark.SparkPipelineOptions;
 import org.apache.beam.runners.spark.translation.streaming.Checkpoint;
 import org.apache.beam.runners.spark.translation.streaming.Checkpoint.CheckpointDir;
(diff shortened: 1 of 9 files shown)


davidak09:
R: @dmvk

runners/spark/src/main/java/org/apache/beam/runners/spark/metrics/SparkBeamMetric.java
```diff
@@ -51,11 +51,12 @@
...
-      metrics.put(renderName(metricResult) + ".sum", result.getSum());
-      metrics.put(renderName(metricResult) + ".min", result.getMin());
-      metrics.put(renderName(metricResult) + ".max", result.getMax());
-      metrics.put(renderName(metricResult) + ".mean", result.getMean());
+      String name = renderName(metricResult);
+      metrics.put(name + ".count", result.getCount());
+      metrics.put(name + ".sum", result.getSum());
+      metrics.put(name + ".min", result.getMin());
+      metrics.put(name + ".max", result.getMax());
+      metrics.put(name + ".mean", result.getMean());
```
davidak09, line 59: I'd personally prefer single metric, possibly with `.distribution` suffix, which could include all 5 stats (count, sum, min, max, mean), it would definitely be more readable in Spark UI

aromanenko-dev:
Run Java PreCommit

aromanenko-dev:
Run Spark ValidatesRunner

runners/spark/src/main/java/org/apache/beam/runners/spark/metrics/MetricsAccumulator.java
```diff
@@ -58,13 +57,13 @@ public static void init(SparkPipelineOptions opts, JavaSparkContext jsc) {
               opts.isStreaming()
                   ? Optional.of(new CheckpointDir(opts.getCheckpointDir()))
                   : Optional.absent();
-          MetricsContainerStepMap metricsContainerStepMap = new MetricsContainerStepMap();
+          SparkMetricsContainerStepMap metricsContainerStepMap = new SparkMetricsContainerStepMap();
```
aromanenko-dev, line 60: Won't it be enough to just create new instance of `SparkMetricsContainerStepMap` here but keep the same interface of `MetricsContainerStepMap` everywhere as it was before get broken? Why do we need to change it everywhere?

davidak09: You're right, it's enough to change it just in initialization. I fixed it.

I will try to add a regression test.

runners/spark/src/main/java/org/apache/beam/runners/spark/metrics/SparkMetricsContainerStepMap.java
```diff
@@ -27,7 +27,7 @@
 
   @Override
   public String toString() {
-    return new SparkBeamMetric().renderAll().toString();
+    return asAttemptedOnlyMetricResults(this).toString();
```
davidak09, line 30: We ran our application built with Beam `2.18.0` and the metrics are displayed correctly, with this fix Spark displays the same results.

The change that broke things up for Spark was part of BEAM-9600 - #11369 .

aromanenko-dev: @ibzib fyi

aromanenko-dev:
Run Spark ValidatesRunner

aromanenko-dev:
Run Spark Runner Nexmark Tests

aromanenko-dev:

runners/spark/src/main/java/org/apache/beam/runners/spark/metrics/SparkMetricsContainerStepMap.java
ibzib: Thanks for the heads up. My assumption in #11369 was that the result was meant to be consumed by code, not humans. After this change, metrics would remain not too readable for Flink. Ideally we could find a way to make the Flink metrics protos available to the Python client without `toString`, since `toString` should generally should be human-readable. The problem is that without a job server, there's nowhere to access job information like metrics outside of what Flink provides natively.

aromanenko-dev:
Run Spark Runner Nexmark Tests

runners/spark/src/main/java/org/apache/beam/runners/spark/metrics/SparkMetricsContainerStepMap.java
aromanenko-dev: @ibzib Thanks for details. I agree that we should not rely on `toString()` when it's going to be used by other program. 
I'm sorry but I'm not sure that I fully understand your last sentence - does it seem that it's only specific Flink-related issue?
Regarding readable Flink metrics - would it be better to fix it on FlinkRunner side, like we do with `SparkMetricsContainerStepMap` for Spark Runner?

ibzib: There are two ways of running Beam Python Flink/Spark. One way involves starting a Java job server, the other (newer) way does not require a Java job server and instead uses only Python. When using a Java job server, it's easy to get metrics, since the job server has access to the Flink/Spark context object. But in Python, we rely on the Flink REST API to get metrics, so the formatting of the results displayed there is important.

The reason metrics formatting isn't a problem for Spark is that spark_uber_jar_job_server.py just doesn't implement get_metrics yet.

The problem is that if we want to make `MetricsContainerStepMap::toString` human-readable, we'll need to make the protobuf-formatted metrics accessible somewhere else.

Perhaps there is a compromise though if we can somehow format metrics so they are readable both by the proto parser and humans.

aromanenko-dev: @ibzib Thanks for more details. I see the picture bit still not convinced why `MetricsContainerStepMap::toString ` is used for REST API (if I get it right) - dedicated way  to make the protobuf-formatted metrics accessible, as you said, should be better imho. 

Are you ok to merge this fix for Spark metrics?

aromanenko-dev:
Run Spark Runner Nexmark Tests

runners/spark/src/main/java/org/apache/beam/runners/spark/metrics/SparkMetricsContainerStepMap.java
ibzib: > @ibzib Thanks for more details. I see the picture bit still not convinced why MetricsContainerStepMap::toString is used for REST API (if I get it right) - dedicated way to make the protobuf-formatted metrics accessible, as you said, should be better imho.

The reason I used `toString` is because the Flink REST API displays accumulator values using `toString`. I simply wasn't able to come up with a dedicated way to make the protobuf-formatted metrics accessible.

> Are you ok to merge this fix for Spark Runner metrics?

Yes, this is fine for now. I might need to give Flink runner metrics another look to make sure we don't have a similar problem there.

aromanenko-dev:
Run Spark Runner Nexmark Tests

aromanenko-dev:
Spark Runner Nexmark Tests

aromanenko-dev:
Run Spark Runner Nexmark Tests

PR #12067, opened 2020-06-23
[BEAM-10308] Make component ID assignments consistent across PipelineContext instances

TheNeuralBit:
This PR creates a global cache for component ID assignments made in PipelineContext instances. See BEAM-10308 and BEAM-10143 for motivation.

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/pipeline.py
+++ b/sdks/python/apache_beam/pipeline.py
@@ -58,6 +58,7 @@ import sys
 import tempfile
 from builtins import object
 from builtins import zip
+from collections import defaultdict
 from typing import TYPE_CHECKING
 from typing import Dict
 from typing import FrozenSet
@@ -221,12 +222,11 @@ class Pipeline(object):
     # then the transform will have to be cloned with a new label.
     self.applied_labels = set()  # type: Set[str]
 
-    # Create a context for assigning IDs to components. Ensures that any
+    # Create a ComponentIdMap for assigning IDs to components. Ensures that any
     # components that receive an ID during pipeline construction (for example in
     # ExternalTransform), will receive the same component ID when generating the
     # full pipeline proto.
-    from apache_beam.runners import pipeline_context
-    self._component_id_context = pipeline_context.ComponentIdContext()
+    self.component_id_map = ComponentIdMap()
 
 
   @property  # type: ignore[misc]  # decorated property not supported
@@ -819,7 +819,7 @@ class Pipeline(object):
     if context is None:
       context = pipeline_context.PipelineContext(
           use_fake_coders=use_fake_coders,
-          component_id_context=self._component_id_context,
+          component_id_map=self.component_id_map,
           default_environment=default_environment)
     elif default_environment is not None:
       raise ValueError(
@@ -1347,6 +1347,32 @@ class PTransformOverride(with_metaclass(abc.ABCMeta,
     """
     # Returns a PTransformReplacement
     raise NotImplementedError
+
+
+class ComponentIdMap(object):
+  """A utility for assigning unique component ids to Beam components.
+
+  Component ID assignments are only guaranteed to be unique and consistent
+  within the scope of a ComponentIdMap instance.
+  """
+  def __init__(self, namespace="ref"):
+    self.namespace = namespace
+    self._counters = defaultdict(lambda: 0)  # type: Dict[type, int]
+    self._obj_to_id = {}  # type: Dict[Any, str]
+
+  def get_or_assign(self, obj=None, obj_type=None, label=None):
+    if obj not in self._obj_to_id:
+      self._obj_to_id[obj] = self._unique_ref(obj, obj_type, label)
+
+    return self._obj_to_id[obj]
+
+  def _unique_ref(self, obj=None, obj_type=None, label=None):
+    self._counters[obj_type] += 1
+    return "%s_%s_%s_%d" % (
+        self.namespace,
+        obj_type.__name__,
+        label or type(obj).__name__,
+        self._counters[obj_type])
 
 
 if sys.version_info >= (3, ):
--- a/sdks/python/apache_beam/pipeline.py
+++ b/sdks/python/apache_beam/pipeline.py
@@ -221,6 +221,14 @@ class Pipeline(object):
     # then the transform will have to be cloned with a new label.
     self.applied_labels = set()  # type: Set[str]
 
+    # Create a context for assigning IDs to components. Ensures that any
+    # components that receive an ID during pipeline construction (for example in
+    # ExternalTransform), will receive the same component ID when generating the
+    # full pipeline proto.
+    from apache_beam.runners import pipeline_context
+    self._component_id_context = pipeline_context.ComponentIdContext()
+
+
   @property  # type: ignore[misc]  # decorated property not supported
   @deprecated(
       since='First stable release',
@@ -811,6 +819,7 @@ class Pipeline(object):
     if context is None:
       context = pipeline_context.PipelineContext(
           use_fake_coders=use_fake_coders,
+          component_id_context=self._component_id_context,
           default_environment=default_environment)
     elif default_environment is not None:
       raise ValueError(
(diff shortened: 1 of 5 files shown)


TheNeuralBit:
One shortcoming of this approach is that component ids are stored in a _global_ cache. I think it would be better if the cache could be scoped to a pipeline, but that would be more difficult to implement. I don't _think_ the global cache would cause any issues except for an extreme case where thousands of pipelines are being launched from a single python process and we consume a huge amount of memory.

TheNeuralBit:
R: @chamikaramj

TheNeuralBit:
CC: @lukecwik

TheNeuralBit:
I suppose another place where global state is ill-advised is when running tests, since we run all the tests in the same process and many of them create pipelines. 

Python precommits are failing because of this. It looks like the failing tests are making assertions about specific values of component keys which seems brittle, for example:

```
self = <apache_beam.runners.interactive.pipeline_instrument_test.PipelineInstrumentTest testMethod=test_cacheable_key_with_version_map>

    def test_cacheable_key_with_version_map(self):
      p = beam.Pipeline(interactive_runner.InteractiveRunner())
      # pylint: disable=range-builtin-not-iterating
      init_pcoll = p | 'Init Create' >> beam.Create(range(10))
    
      # It's normal that when executing, the pipeline object is a different
      # but equivalent instance from what user has built. The pipeline instrument
      # should be able to identify if the original instance has changed in an
      # interactive env while mutating the other instance for execution. The
      # version map can be used to figure out what the PCollection instances are
      # in the original instance and if the evaluation has changed since last
      # execution.
      p2 = beam.Pipeline(interactive_runner.InteractiveRunner())
      # pylint: disable=range-builtin-not-iterating
      init_pcoll_2 = p2 | 'Init Create' >> beam.Create(range(10))
      _, ctx = p2.to_runner_api(use_fake_coders=True, return_context=True)
    
      # The cacheable_key should use id(init_pcoll) as prefix even when
      # init_pcoll_2 is supplied as long as the version map is given.
      self.assertEqual(
          instr.cacheable_key(
              init_pcoll_2,
              instr.pcolls_to_pcoll_id(p2, ctx),
              {'ref_PCollection_PCollection_8': str(id(init_pcoll))}),
>         str(id(init_pcoll)) + '_ref_PCollection_PCollection_8')
E     AssertionError: '140176476148624_ref_PCollection_PCollection_4539' != '140176476499024_ref_PCollection_PCollection_8'
E     - 140176476148624_ref_PCollection_PCollection_4539
E     ?          - ^^                               ^^^^
E     + 140176476499024_ref_PCollection_PCollection_8
E     ?           ^^^      
```

Its probably easier to do the work to make sure cached component ids are scoped to an individual pipeline rather than fixing all of these tests (and a global cache shared across tests will be problematic anyway).

TheNeuralBit:
I have a POC working that solves the problem by fixing all of the component IDs in the external transforms in `to_runner_api()`. It feels pretty error-prone, but it doesn't maintain any global state, and keeps all of the special logic within ExternalTransform. I'll clean it up and push it here tomorrow morning.

chamikaramj:
Do you think we can add another unit test that shows how current state will result in an invalid proto during cross-language expansion ? That will help us understand the scope of the issue.

TheNeuralBit:
I'll see if I can add a simple unit test tomorrow.

sdks/python/apache_beam/runners/pipeline_context.py
```diff
@@ -49,6 +50,52 @@
...
+    return "%s_%s_%d" % (self._base, label or type(obj).__name__, self._counter)
+
+  @classmethod
+  def get_instance(cls, pipeline, base):
+    # type: (Optional[Pipeline], str) -> _UniqueRefAssigner
+
+    """Return the _UniqueRefAssigner with the given base string.
+
+    Creates a new instance if one doesn't already exist for this base string."""
+    key = (id(pipeline), base)
```
TheNeuralBit, line 91: Partitioning the cached ID assignments by `id(pipeline)` is a sub-par and dangerous solution because `id` is only guaranteed to be unique among objects that have non-overlapping lifetimes. If we go with something like this approach, we need to at least tie these instances' lifetimes to the lifetime of the pipeline.

TheNeuralBit:
Run XVR_Flink PostCommit

TheNeuralBit:
I think something like this will fix the ExternalTransform bug, without breaking all the tests that rely on specific component id assignments.

It will still grow without bound with each pipeline that's created in a process though.

chamikaramj:
I think that's fine. We should not be creating too many pipeline objects in parallel.

TheNeuralBit:
Run XVR_Flink PostCommit

TheNeuralBit:
Run XVR_Spark PostCommit

TheNeuralBit:
Ok I think I'm finally done churning on this. Sorry for all the changes. What I have now stores the component ids in a `ComponentIdContext` object stored on the Pipeline. The same context is used in both `ExternalTransform.expand` and in `Pipeline.to_runner_api`, so component ids will be consistent across both of those functions, but everything we store in memory will be cleaned up with the Pipeline. PTAL

TheNeuralBit:
retest this please

TheNeuralBit:
Run XVR_Flink PostCommit

TheNeuralBit:
Run XVR_Spark PostCommit

TheNeuralBit:
The only failure in Python precommit seems to be due to BEAM-10007. I think this is safe to merge after an LGTM

TheNeuralBit:
@lukecwik do you happen to have time to review this? Looks like cham is OOO

I'd like to get this in before the release cut tomorrow since it fixes a significant issue with ExternalTransform.

sdks/python/apache_beam/pipeline.py
```diff
@@ -221,6 +221,14 @@ def __init__(self, runner=None, options=None, argv=None):
     # then the transform will have to be cloned with a new label.
     self.applied_labels = set()  # type: Set[str]
 
+    # Create a context for assigning IDs to components. Ensures that any
+    # components that receive an ID during pipeline construction (for example in
+    # ExternalTransform), will receive the same component ID when generating the
+    # full pipeline proto.
+    from apache_beam.runners import pipeline_context
```
kennknowles, line 228: There's not actually a circular dependency of classes. You could move this class in here and that would be fine.

sdks/python/apache_beam/runners/pipeline_context.py
```diff
@@ -49,6 +50,32 @@
   from apache_beam.coders.coder_impl import IterableStateWriter
 
 
+class ComponentIdContext(object):
```
kennknowles, line 53: There's a lot of overlap between the purpose of this and the purpose of the `PipelineContext` but I can see how they are different. I see that the pipeline only takes the context at `to_runner_api` time, and the context has tweaks like `use_fake_coders` and `default_environment`. So when using xlang expansion you can keep just the ids and throw away the contents that will be generated according to these tweaks later.

sdks/python/apache_beam/transforms/external.py
```diff
@@ -286,7 +286,8 @@ def expand(self, pvalueish):
     pipeline = (
         next(iter(self._inputs.values())).pipeline
         if self._inputs else pvalueish.pipeline)
-    context = pipeline_context.PipelineContext()
+    context = pipeline_context.PipelineContext(
+        component_id_context=pipeline._component_id_context)
```
kennknowles, line 290: Based on this use, maybe `_component_id_context` isn't private at all? And maybe it is a `component_id_map` or some such. It isn't "context" in the sense that it isn't "the stuff surrounding the thing you are interested in".

TheNeuralBit:
Run XVR_Flink PostCommit

TheNeuralBit:

sdks/python/apache_beam/transforms/external.py
TheNeuralBit: Good point, renamed it to `component_id_map` everywhere

sdks/python/apache_beam/pipeline.py
TheNeuralBit: Done, thanks!

TheNeuralBit:
@kennknowles are you ok with merging this now?

chamikaramj:
Took a look and confirmed that x-lang KafkaIO works with this change. LGTM from me.

PR #7676, opened 2019-01-30
[BEAM-6431] Implement Execution Time metrics start,process,finish in the Java SDK

ajamato:
[BEAM-6431] Implement Execution Time metrics start,process,finish in the Java SDK

**Please** add a meaningful description for your change here

------------------------

Follow this checklist to help us incorporate your contribution quickly and easily:

 - [ ] Format the pull request title like `[BEAM-XXX] Fixes bug in ApproximateQuantiles`, where you replace `BEAM-XXX` with the appropriate JIRA issue, if applicable. This will automatically link the pull request to the issue.
 - [ ] If this contribution is large, please file an Apache [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

It will help us expedite review of your Pull Request if you tag someone (e.g. `@username`) to look at it.

Post-Commit Tests Status (on master branch)
------------------------------------------------------------------------------------------------

Lang | SDK | Apex | Dataflow | Flink | Gearpump | Samza | Spark
--- | --- | --- | --- | --- | --- | --- | ---
Go | [Build Status](https://builds.apache.org/job/beam_PostCommit_Go/lastCompletedBuild/) | --- | --- | --- | --- | --- | ---
Java | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Apex/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Dataflow/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Flink/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Flink_Batch/lastCompletedBuild/)<br>[Build Status](https://builds.apache.org/job/beam_PostCommit_Java_PVR_Flink_Streaming/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Gearpump/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Samza/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Spark/lastCompletedBuild/)
[...]

PR #12070, opened 2020-06-24
Add some metrics for BigQueryStorageStreamReader

kmjung:
Add basic latency metrics for BigQuery storage stream source

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryStorageStreamSource.java
+++ b/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryStorageStreamSource.java
@@ -48,6 +48,7 @@ import org.apache.beam.sdk.io.BoundedSource;
 import org.apache.beam.sdk.io.gcp.bigquery.BigQueryServices.BigQueryServerStream;
 import org.apache.beam.sdk.io.gcp.bigquery.BigQueryServices.StorageClient;
 import org.apache.beam.sdk.metrics.Counter;
+import org.apache.beam.sdk.metrics.Distribution;
 import org.apache.beam.sdk.metrics.Metrics;
 import org.apache.beam.sdk.options.PipelineOptions;
 import org.apache.beam.sdk.transforms.SerializableFunction;
@@ -157,15 +158,13 @@ public class BigQueryStorageStreamSource<T> extends BoundedSource<T> {
   public static class BigQueryStorageStreamReader<T> extends BoundedSource.BoundedReader<T> {
 
     private final Counter streamingResponseCount =
-        Metrics.counter(
-            BigQueryStorageStreamReader.class, "bq_storage_read_streaming_response_count");
-    private final Counter streamingResponseTotalLatencyMs =
-        Metrics.counter(
-            BigQueryStorageStreamReader.class, "bq_storage_read_streaming_response_latency_ms");
-    private final Counter decoderTotalLatencyMs =
-        Metrics.counter(BigQueryStorageStreamReader.class, "bq_storage_read_decoder_latency_ms");
-    private final Counter parseFnTotalLatencyMs =
-        Metrics.counter(BigQueryStorageStreamReader.class, "bq_storage_read_parse_latency_ms");
+        Metrics.counter(BigQueryStorageStreamReader.class, "streaming_response_count");
+    private final Distribution streamingResponseLatenciesMs =
+        Metrics.distribution(BigQueryStorageStreamReader.class, "streaming_response_latencies_ms");
+    private final Distribution decodeLatenciesMs =
+        Metrics.distribution(BigQueryStorageStreamReader.class, "decode_latencies_ms");
+    private final Distribution parseLatenciesMs =
+        Metrics.distribution(BigQueryStorageStreamReader.class, "parse_latencies_ms");
 
     private final DatumReader<GenericRecord> datumReader;
     private final SerializableFunction<SchemaAndRecord, T> parseFn;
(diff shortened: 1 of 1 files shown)


sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryStorageStreamSource.java
```diff
@@ -247,8 +267,19 @@ private synchronized boolean readNextRecord() throws IOException {
...
-      record = datumReader.read(record, decoder);
-      current = parseFn.apply(new SchemaAndRecord(record, tableSchema));
+      Stopwatch stopwatch = Stopwatch.createStarted();
+      try {
+        record = datumReader.read(record, decoder);
+      } finally {
+        decoderTotalLatencyMs.inc(stopwatch.elapsed(TimeUnit.MILLISECONDS));
+      }
+
+      stopwatch.reset();
```
emkornfield, line 277: I think you might need to resetart the stopwatch here: https://guava.dev/releases/19.0/api/docs/com/google/common/base/Stopwatch.html#reset()

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryStorageStreamSource.java
```diff
@@ -153,6 +156,17 @@ public String toString() {
   @Experimental(Kind.SOURCE_SINK)
   public static class BigQueryStorageStreamReader<T> extends BoundedSource.BoundedReader<T> {
 
+    private final Counter streamingResponseCount =
+        Metrics.counter(
+            BigQueryStorageStreamReader.class, "bq_storage_read_streaming_response_count");
+    private final Counter streamingResponseTotalLatencyMs =
+        Metrics.counter(
```
emkornfield, line 163: maybe [distribution](https://beam.apache.org/releases/javadoc/2.0.0/org/apache/beam/sdk/metrics/Distribution.html) for time base metrics?

kmjung: Done.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryStorageStreamSource.java
kmjung: Good catch.

kmjung:
Please review, @chamikaramj and @ajamato.

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryStorageStreamSource.java
```diff
@@ -219,7 +232,15 @@ private synchronized boolean readNextRecord() throws IOException {
         }
 
         fractionConsumedFromPreviousResponse = fractionConsumedFromCurrentResponse;
-        ReadRowsResponse currentResponse = responseIterator.next();
+        ReadRowsResponse currentResponse;
+        Stopwatch stopwatch = Stopwatch.createStarted();
```
chamikaramj, line 236: Do you think these operations can have a performance impact ? If so will it make sense to add an option to disable/enable these ?

chamikaramj:
Please add a JIRA

sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryStorageStreamSource.java
ajamato, line 236: +1 to adding an options to allow disabling this.

In the past we have avoided using tools which grab the system clock frequently as this system call can be quite slow and add performance impact. Especially if its done frequently. In this case however, it looks like the system clock may only be called infrequently. I.e. whenever an RPC is made (doing it or every element would be an issue) In which case its fine to do so.

Instead we have used the state sampler classes to do this. This uses a separate thread to count which periodically wakes up and checks what "state" other threads are executing under. Which would . 

However, there is no user metric or sdk level API to time blocks of code using the state sampler.

We only used it for the execution time metrics. To calculate the time spend in the process, startBundle and finishBundle methods of each ParDos. Here it is for reference.
https://github.com/apache/beam/pull/7676

kmjung: I can certainly add options to disable the collection of these metrics if we establish that this is the right thing to do.

The ReadRows API is a server streaming API which, on the server side, tries to send ReadRowsResponse messages as fast as the client can accept them. Each response message contains up to 100 MiB of encoded data -- in the extreme case, this is one row per response, but in the common case this is a batch of rows, usually 1024. The timer is read once -- or, I guess, twice -- per response message.

What I'd like to establish with these metrics is whether the existing client-side read-ahead is sufficient in the common case to prevent (a large amount of) blocking while reading from a stream, or whether we're spending time waiting for I/O on this thread in addition to doing Avro decoding, etc.

ajamato: It's probably fine to not use an option, since it occurs infrequently. Though it may be a good idea to add an options as a precaution. Making it opt in. We can turn it on by default later.

It may be a lot of effort to do this with the state sampler. So the approach you have is probably best to start with.

Other approaches would be to run the job through some sort of profiler. Though I am not sure how to do that today with any of our runners. Possibly it could be setup with direct runner. I think that would help you to get plenty of useful info on the problem as well. May be worth asking on the dev list about that.

ajamato:
R: @ihji

PR #12088, opened 2020-06-25
Add output typehints to GroupIntoBatches

aaltay:

------------------------

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/transforms/util.py
+++ b/sdks/python/apache_beam/transforms/util.py
@@ -741,6 +741,7 @@ def WithKeys(pcoll, k):
 
 @experimental()
 @typehints.with_input_types(Tuple[K, V])
+@typehints.with_output_types(Tuple[K, List[V]])
 class GroupIntoBatches(PTransform):
   """PTransform that batches the input into desired batch size. Elements are
   buffered until they are equal to batch size provided in the argument at which


sdks/python/apache_beam/transforms/util.py
```diff
@@ -741,6 +741,7 @@ def WithKeys(pcoll, k):
 
 @experimental()
 @typehints.with_input_types(Tuple[K, V])
+@typehints.with_output_types(Tuple[K, List[V]])
```
udim, line 744: I was thinking it might be Iterable rather than List, so I looked at the code and it doesn't seem to group by key at all.

I modified a test from `GroupIntoBatchesTest` to print to elements:
```py
  def test_in_global_window(self):
    with TestPipeline() as pipeline:
      def print_elements(e):
        print(e)
        return e

      collection = pipeline \
                   | beam.Create(GroupIntoBatchesTest._create_test_data()) \
                   | util.GroupIntoBatches(GroupIntoBatchesTest.BATCH_SIZE) \
                   | beam.Map(print_elements)
```

Printout:
```py
[('key', 'Einstein'), ('key', 'Darwin'), ('key', 'Copernicus'), ('key', 'Pasteur'), ('key', 'Curie')]
[('key', 'Faraday'), ('key', 'Newton'), ('key', 'Bohr'), ('key', 'Galilei'), ('key', 'Maxwell')]
```

udim: cc: @lukecwik

aaltay: I wonder why the `test_in_global_window` tests passes. The assert following that code is seemingly checking for the right batching.

The implementation matches the Java one (https://github.com/apache/beam/blob/master/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/GroupIntoBatches.java#L70). I _think_ grouping by key is supposed to happen as part of the state api, because element_state will be keyed by key and window.

kennknowles: Correct. It is up to the runner how to implement, but per key and window must be executed serially.

kennknowles: A very real situation is that it might already be partitioned however the runner wants, via analyzing the pipeline.

aaltay: Thank you. This might be a bug with whatever runner Udi used (one of the direct runners). I am still puzzled by what would be a good implementation of this case in batch. Would a naive implementation be always insert a shuffle, and use an memory backed state store for the batch case?

udim: The test passes because 2 batches/elements are output. And I was running the test via pytest as usual.

I don't follow the state api reasoning, and this is not documented in GroupIntoBatches. If you get a different result in DirectRunner vs DataflowRunner, it's a bug.

udim: The Java implementation splits (key, value) pairs.
https://github.com/apache/beam/blob/eaa41cc4cbcc4f94d0ec1a36ff2b0f3fcee962f9/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/GroupIntoBatches.java#L173-L174

I don't see that in Python - is the runner supposed to do that?

lukecwik: > Thank you. This might be a bug with whatever runner Udi used (one of the direct runners). I am still puzzled by what would be a good implementation of this case in batch. Would a naive implementation be always insert a shuffle, and use an memory backed state store for the batch case?

Using an in memory representation is exactly what is done for all Beam runners today and can only be done because these runners execute the whole stage for a single key within a single bundle.

lukecwik: > The Java implementation splits (key, value) pairs.
> https://github.com/apache/beam/blob/eaa41cc4cbcc4f94d0ec1a36ff2b0f3fcee962f9/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/GroupIntoBatches.java#L173-L174
> 
> I don't see that in Python - is the runner supposed to do that?

This is a space saving optimization since the key is always going to be the same. Instead of having a bag of (<K, V1>, <K, V2>, <K, V3>, ...) we use two state cells, one storing K and the other storing (V1, V2, V3, ...) and combine them to recreate the output. It would be worthwhile to replicate this optimization in Python as well but it isn't necessary from a correctness point of view.

udim: So is grouping by key being done at all by this transform? If so, how?
And why isn't the test getting a grouped result?

aaltay: I think I understand the difference between Python and Java.

In Java, this works by emitting (Key, Iterable of Values) (here: https://github.com/apache/beam/blob/51bd3a4588984c9680f61a3268513fe5bde3ba96/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/GroupIntoBatches.java#L212)

In Python, it stores (K, V) pairs elements (here: https://github.com/apache/beam/blob/51bd3a4588984c9680f61a3268513fe5bde3ba96/sdks/python/apache_beam/transforms/util.py#L784) and emits them as a list (here: https://github.com/apache/beam/blob/51bd3a4588984c9680f61a3268513fe5bde3ba96/sdks/python/apache_beam/transforms/util.py#L800) result in a List[Tuple[Key, Value]]

If my understanding is correct, I can change the Python behavior to be similar to the Java one. And I can also add the space sharing optimization.

aaltay: I am thinking something like this: https://github.com/apache/beam/pull/12129 -- Have not really tested this yet beyond a unit test. I ran the the print test Udi wrote above and the output is batched by key.

aaltay:
Closing this PR. https://github.com/apache/beam/pull/12129 supersedes this.

PR #12133, opened 2020-06-30
[BEAM-10385] Integrate SQL expansion into Flink job server 

mxm:
The expansion server is started with the job server to allow for an expansion of external transforms during pipeline submission. At the moment, an extra expansion server has to be started to expand SqlTransform. It would be desirable to include the SQL expansion into the expansion server that is started with the job server.

Lang | SDK | Dataflow | Flink | Samza | Spark | Twister2
Go | [Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Go/lastCompletedBuild/) | --- | [Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Go_VR_Flink/lastCompletedBuild/) | --- | [Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Go_VR_Spark/lastCompletedBuild/) | ---
Java | [Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Java/lastCompletedBuild/) | [Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Dataflow/lastCompletedBuild/)<br>[Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Dataflow_Java11/lastCompletedBuild/) | [Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Flink/lastCompletedBuild/)<br>[Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Flink_Java11/lastCompletedBuild/)<br>[Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Java_PVR_Flink_Batch/lastCompletedBuild/)<br>[Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Java_PVR_Flink_Streaming/lastCompletedBuild/) | [Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Samza/lastCompletedBuild/) | [Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Spark/lastCompletedBuild/)<br>[Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Java_PVR_Spark_Batch/lastCompletedBuild/)<br>[Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Java_ValidatesRunner_SparkStructuredStreaming/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Twister2/lastCompletedBuild/)
Python | [Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Python2/lastCompletedBuild/)<br>[Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Python35/lastCompletedBuild/)<br>[Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Python36/lastCompletedBuild/)<br>[Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Python37/lastCompletedBuild/) | [Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Py_VR_Dataflow/lastCompletedBuild/)<br>[Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Py_VR_Dataflow_V2/lastCompletedBuild/)<br>[Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Py_ValCont/lastCompletedBuild/) | [Build Status](https://ci-beam.apache.org/job/beam_PreCommit_Python2_PVR_Flink_Cron/lastCompletedBuild/)<br>[Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Python35_VR_Flink/lastCompletedBuild/) | --- | [Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Python_VR_Spark/lastCompletedBuild/) | ---
XLang | --- | --- | [Build Status](https://ci-beam.apache.org/job/beam_PostCommit_XVR_Flink/lastCompletedBuild/) | --- | --- | [Build Status](https://ci-beam.apache.org/job/beam_PostCommit_XVR_Spark/lastCompletedBuild/) | ---

--- | --- | --- | --- | ---

### DIFF
--- a/sdks/python/apache_beam/transforms/sql.py
+++ b/sdks/python/apache_beam/transforms/sql.py
@@ -67,10 +67,18 @@ class SqlTransform(ExternalTransform):
   """
   URN = 'beam:external:java:sql:v1'
 
-  def __init__(self, query, dialect=None):
+  def __init__(self, query, dialect=None, expansion_service=None):
+    """
+    Creates a SqlTransform which will be expanded to Java's SqlTransform.
+    (See class docs).
+    :param query: The SQL query.
+    :param dialect: (optional) The dialect, e.g. use 'zetasql' for ZetaSQL.
+    :param expansion_service: (optional) The URL of the expansion service to use
+    """
+    expansion_service = expansion_service or BeamJarExpansionService(
+        ':sdks:java:extensions:sql:expansion-service:shadowJar')
     super(SqlTransform, self).__init__(
         self.URN,
         NamedTupleBasedPayloadBuilder(
             SqlTransformSchema(query=query, dialect=dialect)),
-        BeamJarExpansionService(
-            ':sdks:java:extensions:sql:expansion-service:shadowJar'))
+        expansion_service=expansion_service)
--- a/runners/direct-java/build.gradle
+++ b/runners/direct-java/build.gradle
@@ -26,7 +26,6 @@ def dependOnProjects = [":runners:core-construction-java",
                         ":runners:core-java",
                         ":runners:local-java",
                         ":runners:java-fn-execution",
-                        ":runners:java-job-service",
                         ":sdks:java:fn-execution"]
 
 applyJavaNature(
--- a/runners/flink/job-server/flink_job_server.gradle
+++ b/runners/flink/job-server/flink_job_server.gradle
@@ -86,10 +86,14 @@ dependencies {
   runtime project(":sdks:java:extensions:google-cloud-platform-core")
   runtime library.java.slf4j_simple
 //  TODO: Enable AWS and HDFS file system.
-  // For resolving external transform requests
-  runtime project(":sdks:java:io:kafka")
-  runtime library.java.kafka_clients
-  runtime project(":sdks:java:io:google-cloud-platform")
+  // External transform expansion
+  // Kafka
+  runtimeOnly project(":sdks:java:io:kafka")
+  runtimeOnly library.java.kafka_clients
+  // PubSub
+  runtimeOnly project(":sdks:java:io:google-cloud-platform")
+  // SqlTransform
+  runtimeOnly project(":sdks:java:extensions:sql:expansion-service")
 }
 
 // NOTE: runShadow must be used in order to run the job server. The standard run
--- a/sdks/java/expansion-service/src/main/java/org/apache/beam/sdk/expansion/service/ExpansionService.java
+++ b/sdks/java/expansion-service/src/main/java/org/apache/beam/sdk/expansion/service/ExpansionService.java
@@ -303,12 +303,16 @@ public class ExpansionService extends ExpansionServiceGrpc.ExpansionServiceImplB
   private Map<String, TransformProvider> registeredTransforms = loadRegisteredTransforms();
 
   private Map<String, TransformProvider> loadRegisteredTransforms() {
-    ImmutableMap.Builder<String, TransformProvider> registeredTransforms = ImmutableMap.builder();
+    ImmutableMap.Builder<String, TransformProvider> registeredTransformsBuilder =
+        ImmutableMap.builder();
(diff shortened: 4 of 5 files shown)


mxm:

mxm:

sdks/python/apache_beam/transforms/sql.py
```diff
@@ -67,10 +66,9 @@ class SqlTransform(ExternalTransform):
...
 
-  def __init__(self, query, dialect=None):
+  def __init__(self, query, dialect=None, expansion_service=None):
     super(SqlTransform, self).__init__(
         self.URN,
         NamedTupleBasedPayloadBuilder(
             SqlTransformSchema(query=query, dialect=dialect)),
-        BeamJarExpansionService(
-            ':sdks:java:extensions:sql:expansion-service:shadowJar'))
+        expansion_service=expansion_service)
```
robertwb, line 74: We definitely want to accept an expansion service as an argument, but the default should remain `BeamJarExpansionService(':sdks:java:extensions:sql:expansion-service:shadowJar')`.

robertwb: Note that you can use `JavaJarServer.beam_services()` (automatically populated by the `--beam_services` flag) to specify an alternative jar or live expansion service for any beam target.

mxm: Thanks for the pointer. The workaround via overriding the Gradle target may be acceptable for developers but exposing it via an argument to users does not seem right. Users shouldn't have to know Gradle to use Beam Python. 

Also, the corresponding value can either be a URL or a `host:port` string which is confusing to users and probably will break for Ipv6 addresses.

robertwb: Yes, the `beam_services` flag is primarily for developers/testing. The default should just work after running "pip install apache_beam" which is what `BeamJarExpansionService` aims to provide. Externally-started job servers running universal expansion services at a fixed port was useful for bootstrapping, but is a bit of a special configuration. Exactly how to best configure this (which, in large part, boils down to being able to name expansion (or other) services), is a good open question.

tweise: If the goal is to make things easier/flexible for developers, then convenience built into Gradle tasks may be a good option. Users would probably find it rather problematic if by default there is an attempt to pull down artifacts behind the scenes and execute them. That is probably also going to fail within isolated production environments.

mxm: For a good out-of-the-box experience it may be acceptable to bundle the expansion service with the job server, which is already the case for the Flink Runner. That way we don't have to start a separate service. I was thinking to move the dependencies for expansion (SQL, Kafka, etc.) to the base job-service project (https://issues.apache.org/jira/browse/BEAM-10386), but decided to postpone it for now because it causes a larger refactoring. IMHO we should get rid of the Runner-specific job service projects entirely, each Runner's artifact should carry all the necessary dependencies through the central job-service project. 

The current approach here is very modular but downloading jars is unexpected because most users would expect everything to be self-contained. The downside also is that it won't easily work for other languages. In the future we may want to either (a) include a self-contained expansion server in the final artifact (b) publish container images for all expansion servers.

I'll keep the current default because this warrants a mailing list discussion.

mxm:

sdks/python/apache_beam/transforms/sql.py
robertwb: +1, a wider discussion about expansion service discovery and bundling would be very good to have on the mailing list.

mxm:
Run Python PreCommit

PR #12145, opened 2020-07-01
[BEAM-10136] [BEAM-10135] Add JdbcIO for cross-language with python wrapper

piotr-szuberski:
Encountered problems:
1. For now the python transforms are very limited as RowCoder is yet to be upgraded:
- Write can accept only primitive types without boolean, bytes
- Read can accept even more limited types because LogicalTypes are not implemented in python
2. Write transform uses JdbcIO.write() - it would be better to use writeRows() with automated Row coder when it's implemented.

In general coders need to be upgraded before these transforms are usable.

R: @TheNeuralBit 

------------------------

Lang | SDK | Dataflow | Flink | Samza | Spark | Twister2
Go | [Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Go/lastCompletedBuild/) | --- | [Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Go_VR_Flink/lastCompletedBuild/) | --- | [Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Go_VR_Spark/lastCompletedBuild/) | ---
Java | [Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Java/lastCompletedBuild/) | [Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Dataflow/lastCompletedBuild/)<br>[Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Dataflow_Java11/lastCompletedBuild/) | [Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Flink/lastCompletedBuild/)<br>[Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Flink_Java11/lastCompletedBuild/)<br>[Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Java_PVR_Flink_Batch/lastCompletedBuild/)<br>[Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Java_PVR_Flink_Streaming/lastCompletedBuild/) | [Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Samza/lastCompletedBuild/) | [Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Spark/lastCompletedBuild/)<br>[Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Java_PVR_Spark_Batch/lastCompletedBuild/)<br>[Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Java_ValidatesRunner_SparkStructuredStreaming/lastCompletedBuild/) | [Build Status](https://builds.apache.org/job/beam_PostCommit_Java_ValidatesRunner_Twister2/lastCompletedBuild/)
Python | [Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Python2/lastCompletedBuild/)<br>[Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Python35/lastCompletedBuild/)<br>[Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Python36/lastCompletedBuild/)<br>[Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Python37/lastCompletedBuild/) | [Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Py_VR_Dataflow/lastCompletedBuild/)<br>[Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Py_VR_Dataflow_V2/lastCompletedBuild/)<br>[Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Py_ValCont/lastCompletedBuild/) | [Build Status](https://ci-beam.apache.org/job/beam_PreCommit_Python2_PVR_Flink_Cron/lastCompletedBuild/)<br>[Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Python35_VR_Flink/lastCompletedBuild/) | --- | [Build Status](https://ci-beam.apache.org/job/beam_PostCommit_Python_VR_Spark/lastCompletedBuild/) | ---
XLang | --- | --- | [Build Status](https://ci-beam.apache.org/job/beam_PostCommit_XVR_Flink/lastCompletedBuild/) | --- | --- | [Build Status](https://ci-beam.apache.org/job/beam_PostCommit_XVR_Spark/lastCompletedBuild/) | ---

--- | --- | --- | --- | ---

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -58,6 +58,7 @@
 
 * New overloads for BigtableIO.Read.withKeyRange() and BigtableIO.Read.withRowFilter()
   methods that take ValueProvider as a parameter (Java) ([BEAM-10283](https://issues.apache.org/jira/browse/BEAM-10283)).
+* Add cross-language support to Java's JdbcIO, now available in the Python module `apache_beam.io.external.jdbc` ([BEAM-10135](https://issues.apache.org/jira/browse/BEAM-10135), [BEAM-10136](https://issues.apache.org/jira/browse/BEAM-10136)).
 * Support for X source added (Java/Python) ([BEAM-X](https://issues.apache.org/jira/browse/BEAM-X)).
 
 ## New Features / Improvements
--- a/sdks/java/io/expansion-service/build.gradle
+++ b/sdks/java/io/expansion-service/build.gradle
@@ -33,6 +33,9 @@ ext.summary = "Expansion service serving several Java IOs"
 dependencies {
   compile project(":sdks:java:expansion-service")
   compile project(":sdks:java:io:kafka")
+  compile project(":sdks:java:io:jdbc")
+  // Include postgres so it can be used with external JDBC
+  compile library.java.postgres
   runtime library.java.kafka_clients
   runtime library.java.slf4j_jdk14
 }
--- a/sdks/python/apache_beam/io/jdbc.py
+++ b/sdks/python/apache_beam/io/jdbc.py
@@ -1,0 +1,276 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""PTransforms for supporting Jdbc in Python pipelines.
+
+  These transforms are currently supported by Beam portable
+  Flink and Spark runners.
+
+  **Setup**
+
+  Transforms provided in this module are cross-language transforms
+  implemented in the Beam Java SDK. During the pipeline construction, Python SDK
+  will connect to a Java expansion service to expand these transforms.
+  To facilitate this, a small amount of setup is needed before using these
+  transforms in a Beam Python pipeline.
+
+  There are several ways to setup cross-language Jdbc transforms.
+
+  * Option 1: use the default expansion service
+  * Option 2: specify a custom expansion service
+
+  See below for details regarding each of these options.
+
+  *Option 1: Use the default expansion service*
+
+  This is the recommended and easiest setup option for using Python Jdbc
+  transforms. This option is only available for Beam 2.24.0 and later.
+
+  This option requires following pre-requisites before running the Beam
+  pipeline.
+
+  * Install Java runtime in the computer from where the pipeline is constructed
+    and make sure that 'java' command is available.
+
+  In this option, Python SDK will either download (for released Beam version) or
+  build (when running from a Beam Git clone) a expansion service jar and use
+  that to expand transforms. Currently Jdbc transforms use the
+  'beam-sdks-java-io-expansion-service' jar for this purpose.
+
+  *Option 2: specify a custom expansion service*
+
+  In this option, you startup your own expansion service and provide that as
+  a parameter when using the transforms provided in this module.
+
+  This option requires following pre-requisites before running the Beam
+  pipeline.
+
+  * Startup your own expansion service.
+  * Update your pipeline to provide the expansion service address when
+    initiating Jdbc transforms provided in this module.
+
+  Flink Users can use the built-in Expansion Service of the Flink Runner's
+  Job Server. If you start Flink's Job Server, the expansion service will be
+  started on port 8097. For a different address, please set the
+  expansion_service parameter.
+
+  **More information**
+
+  For more information regarding cross-language transforms see:
+  - https://beam.apache.org/roadmap/portability/
+
+  For more information specific to Flink runner see:
+  - https://beam.apache.org/documentation/runners/flink/
+"""
+
+# pytype: skip-file
+
+from __future__ import absolute_import
+
+import typing
+
+from past.builtins import unicode
+
+from apache_beam.transforms.external import BeamJarExpansionService
+from apache_beam.transforms.external import ExternalTransform
+from apache_beam.transforms.external import NamedTupleBasedPayloadBuilder
+
+__all__ = [
+    'WriteToJdbc',
+    'ReadFromJdbc',
+]
+
+
+def default_io_expansion_service():
+  return BeamJarExpansionService('sdks:java:io:expansion-service:shadowJar')
+
+
+WriteToJdbcSchema = typing.NamedTuple(
+    'WriteToJdbcSchema',
+    [
+        ('driver_class_name', unicode),
+        ('jdbc_url', unicode),
+        ('username', unicode),
+        ('password', unicode),
+        ('connection_properties', typing.Optional[unicode]),
+        ('connection_init_sqls', typing.Optional[typing.List[unicode]]),
+        ('statement', unicode),
+    ],
+)
+
+
+class WriteToJdbc(ExternalTransform):
+  """A PTransform which writes Rows to the specified database via JDBC.
+
+  This transform receives Rows defined as NamedTuple type and registered in
+  the coders registry, e.g.::
+
+    ExampleRow = typing.NamedTuple('ExampleRow',
+                                   [('id', int), ('name', unicode)])
+    coders.registry.register_coder(ExampleRow, coders.RowCoder)
+
+    with TestPipeline() as p:
+      _ = (
+          p
+          | beam.Create([ExampleRow(1, 'abc')])
+              .with_output_types(ExampleRow)
+          | 'Write to jdbc' >> WriteToJdbc(
+              driver_class_name='org.postgresql.Driver',
+              jdbc_url='jdbc:postgresql://localhost:5432/example',
+              username='postgres',
+              password='postgres',
+              statement='INSERT INTO example_table VALUES(?, ?)',
+          ))
+
+  Experimental; no backwards compatibility guarantees.
+  """
+
+  URN = 'beam:external:java:jdbc:write:v1'
+
+  def __init__(
+      self,
+      driver_class_name,
+      jdbc_url,
+      username,
+      password,
+      statement,
+      connection_properties=None,
+      connection_init_sqls=None,
+      expansion_service=None,
+  ):
+    """
+    Initializes a write operation to Jdbc.
+
+    :param driver_class_name: name of the jdbc driver class
+    :param jdbc_url: full jdbc url to the database.
+    :param username: database username
+    :param password: database password
+    :param statement: sql statement to be executed
+    :param connection_properties: properties of the jdbc connection
+                                  passed as string with format
+                                  [propertyName=property;]*
+    :param connection_init_sqls: required only for MySql and MariaDB.
+                                 passed as list of strings
+    :param expansion_service: The address (host:port) of the ExpansionService.
+    """
+
+    super(WriteToJdbc, self).__init__(
+        self.URN,
+        NamedTupleBasedPayloadBuilder(
+            WriteToJdbcSchema(
+                driver_class_name=driver_class_name,
+                jdbc_url=jdbc_url,
+                username=username,
+                password=password,
+                statement=statement,
+                connection_properties=connection_properties,
+                connection_init_sqls=connection_init_sqls,
+            ),
+        ),
+        expansion_service or default_io_expansion_service(),
+    )
+
+
+ReadFromJdbcSchema = typing.NamedTuple(
+    'ReadFromJdbcSchema',
+    [
+        ('driver_class_name', unicode),
+        ('jdbc_url', unicode),
+        ('username', unicode),
+        ('password', unicode),
+        ('connection_properties', typing.Optional[unicode]),
+        ('connection_init_sqls', typing.Optional[typing.List[unicode]]),
+        ('query', unicode),
+        ('fetch_size', typing.Optional[int]),
+        ('output_parallelization', typing.Optional[bool]),
+    ],
+)
+
+
+class ReadFromJdbc(ExternalTransform):
+  """A PTransform which reads Rows from the specified database via JDBC.
+
+  This transform delivers Rows defined as NamedTuple registered in
+  the coders registry, e.g.::
+
+    ExampleRow = typing.NamedTuple('ExampleRow',
+                                   [('id', int), ('name', unicode)])
+    coders.registry.register_coder(ExampleRow, coders.RowCoder)
+
+    with TestPipeline() as p:
+      result = (
+          p
+          | 'Read from jdbc' >> ReadFromJdbc(
+              driver_class_name='org.postgresql.Driver',
+              jdbc_url='jdbc:postgresql://localhost:5432/example',
+              username='postgres',
+              password='postgres',
+              query='SELECT * FROM example_table',
+          ))
+
+  Experimental; no backwards compatibility guarantees.
+  """
+
+  URN = 'beam:external:java:jdbc:read_rows:v1'
+
+  def __init__(
+      self,
+      driver_class_name,
+      jdbc_url,
+      username,
+      password,
+      query,
+      output_parallelization=None,
+      fetch_size=None,
+      connection_properties=None,
+      connection_init_sqls=None,
+      expansion_service=None,
+  ):
+    """
+    Initializes a read operation from Jdbc.
+
+    :param driver_class_name: name of the jdbc driver class
+    :param jdbc_url: full jdbc url to the database.
+    :param username: database username
+    :param password: database password
+    :param query: sql query to be executed
+    :param output_parallelization: is output parallelization on
+    :param fetch_size: how many rows to fetch
+    :param connection_properties: properties of the jdbc connection
+                                  passed as string with format
+                                  [propertyName=property;]*
+    :param connection_init_sqls: required only for MySql and MariaDB.
+                                 passed as list of strings
+    :param expansion_service: The address (host:port) of the ExpansionService.
+    """
+    super(ReadFromJdbc, self).__init__(
+        self.URN,
+        NamedTupleBasedPayloadBuilder(
+            ReadFromJdbcSchema(
+                driver_class_name=driver_class_name,
+                jdbc_url=jdbc_url,
+                username=username,
+                password=password,
+                query=query,
+                output_parallelization=output_parallelization,
+                fetch_size=fetch_size,
+                connection_properties=connection_properties,
+                connection_init_sqls=connection_init_sqls,
+            ),
+        ),
+        expansion_service or default_io_expansion_service(),
+    )
--- a/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/CrossLanguageConfiguration.java
+++ b/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/CrossLanguageConfiguration.java
@@ -1,0 +1,76 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.beam.sdk.io.jdbc;
+
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+/** Abstract Parameters class to expose the Jdbc transforms to an external SDK. */
+public abstract class CrossLanguageConfiguration {
+  String driverClassName;
+  String jdbcUrl;
+  String username;
+  String password;
+  String connectionProperties;
+  Iterable<String> connectionInitSqls;
+
+  public void setDriverClassName(String driverClassName) {
+    this.driverClassName = driverClassName;
+  }
+
+  public void setJdbcUrl(String jdbcUrl) {
+    this.jdbcUrl = jdbcUrl;
+  }
+
+  public void setUsername(String username) {
+    this.username = username;
+  }
+
+  public void setPassword(String password) {
+    this.password = password;
+  }
+
+  public void setConnectionProperties(String connectionProperties) {
+    this.connectionProperties = connectionProperties;
+  }
+
+  public void setConnectionInitSqls(Iterable<String> connectionInitSqls) {
+    this.connectionInitSqls = connectionInitSqls;
+  }
+
+  protected JdbcIO.DataSourceConfiguration getDataSourceConfiguration() {
+    JdbcIO.DataSourceConfiguration dataSourceConfiguration =
+        JdbcIO.DataSourceConfiguration.create(driverClassName, jdbcUrl)
+            .withUsername(username)
+            .withPassword(password);
+
+    if (connectionProperties != null) {
+      dataSourceConfiguration =
+          dataSourceConfiguration.withConnectionProperties(connectionProperties);
+    }
+
+    if (connectionInitSqls != null) {
+      List<String> initSqls =
+          StreamSupport.stream(connectionInitSqls.spliterator(), false)
+              .collect(Collectors.toList());
+      dataSourceConfiguration = dataSourceConfiguration.withConnectionInitSqls(initSqls);
+    }
+    return dataSourceConfiguration;
+  }
+}
--- a/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcReadRowsRegistrar.java
+++ b/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcReadRowsRegistrar.java
@@ -1,0 +1,84 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.beam.sdk.io.jdbc;
+
+import com.google.auto.service.AutoService;
+import java.util.Map;
+import org.apache.beam.sdk.annotations.Experimental;
+import org.apache.beam.sdk.annotations.Experimental.Kind;
+import org.apache.beam.sdk.expansion.ExternalTransformRegistrar;
+import org.apache.beam.sdk.io.jdbc.JdbcIO.DataSourceConfiguration;
+import org.apache.beam.sdk.transforms.ExternalTransformBuilder;
+import org.apache.beam.sdk.transforms.PTransform;
+import org.apache.beam.sdk.values.PBegin;
+import org.apache.beam.sdk.values.PCollection;
+import org.apache.beam.sdk.values.Row;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.ImmutableMap;
+
+/** Exposes {@link JdbcIO.ReadRows} as an external transform for cross-language usage. */
+@Experimental(Kind.PORTABILITY)
+@AutoService(ExternalTransformRegistrar.class)
+public class JdbcReadRowsRegistrar implements ExternalTransformRegistrar {
+
+  public static final String URN = "beam:external:java:jdbc:read_rows:v1";
+
+  @Override
+  public Map<String, Class<? extends ExternalTransformBuilder>> knownBuilders() {
+    return ImmutableMap.of(URN, JdbcReadRowsRegistrar.Builder.class);
+  }
+
+  /** Parameters class to expose the Read transform to an external SDK. */
+  public static class ReadConfiguration extends CrossLanguageConfiguration {
+    private String query;
+    private Integer fetchSize;
+    private Boolean outputParallelization;
+
+    public void setOutputParallelization(Boolean outputParallelization) {
+      this.outputParallelization = outputParallelization;
+    }
+
+    public void setFetchSize(Integer fetchSize) {
+      this.fetchSize = fetchSize;
+    }
+
+    public void setQuery(String query) {
+      this.query = query;
+    }
+  }
+
+  public static class Builder
+      implements ExternalTransformBuilder<ReadConfiguration, PBegin, PCollection<Row>> {
+    @Override
+    public PTransform<PBegin, PCollection<Row>> buildExternal(ReadConfiguration configuration) {
+      DataSourceConfiguration dataSourceConfiguration = configuration.getDataSourceConfiguration();
+
+      JdbcIO.ReadRows readRows =
+          JdbcIO.readRows()
+              .withDataSourceConfiguration(dataSourceConfiguration)
+              .withQuery(configuration.query);
+
+      if (configuration.fetchSize != null) {
+        readRows = readRows.withFetchSize(configuration.fetchSize);
+      }
+      if (configuration.outputParallelization != null) {
+        readRows = readRows.withOutputParallelization(configuration.outputParallelization);
+      }
+      return readRows;
+    }
+  }
+}
--- a/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcUtil.java
+++ b/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcUtil.java
@@ -19,6 +19,8 @@ package org.apache.beam.sdk.io.jdbc;
 
 import java.sql.Date;
 import java.sql.JDBCType;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
 import java.sql.Time;
 import java.sql.Timestamp;
 import java.util.Calendar;
@@ -29,6 +31,7 @@ import java.util.TimeZone;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import org.apache.beam.sdk.schemas.Schema;
+import org.apache.beam.sdk.values.Row;
 import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.ImmutableMap;
 import org.joda.time.DateTime;
 
@@ -155,6 +158,28 @@ class JdbcUtil {
                     + " in schema is not supported while writing. Please provide statement and preparedStatementSetter");
           }
         }
+    }
+  }
+
+  static class BeamRowPreparedStatementSetter implements JdbcIO.PreparedStatementSetter<Row> {
+    @Override
+    public void setParameters(Row row, PreparedStatement statement) {
+      Schema schema = row.getSchema();
+      List<Schema.Field> fieldTypes = schema.getFields();
+      IntStream.range(0, fieldTypes.size())
+          .forEachOrdered(
+              i -> {
+                Schema.FieldType type = fieldTypes.get(i).getType();
+                try {
+                  JdbcUtil.getPreparedStatementSetCaller(type)
+                      .set(row, statement, i, SchemaUtil.FieldWithIndex.of(schema.getField(i), i));
+                } catch (SQLException throwables) {
+                  throwables.printStackTrace();
+                  throw new RuntimeException(
+                      String.format("Unable to create prepared statement for type: %s", type),
+                      throwables);
+                }
+              });
     }
   }
 
--- a/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcWriteRegistrar.java
+++ b/sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/JdbcWriteRegistrar.java
@@ -1,0 +1,67 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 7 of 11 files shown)


piotr-szuberski:
Unfortunately the python test is taken by tox precommit suites - I suppose it shouldn't?
Btw in which suites should this test be running? postCommitPortable?

Other xlang tests have some skipIfs, e.g.:
```
@unittest.skipUnless(
    os.environ.get('EXPANSION_PORT'))
```
But they are really using those env variables and my test doesn't need any setup. How would you suggest to solve it?

TheNeuralBit:
Run XVR_Flink PostCommit

TheNeuralBit:
> Unfortunately the python test is taken by tox precommit suites - I suppose it shouldn't?

I think that's actually ok. The precommit will uses the fn api runner, which supports cross-language transforms. I haven't been able to run SqlTransform with the fn api runner yet though (BEAM-10010), which is why I still have it disabled in sql_test. 

If JdbcIO is also having trouble running on the fn api runner you could disable it like I did in sql_test, as long as we're still testing continuously with XVR_Flink and XVR_Spark, but we should file a jira for JdbcIO on fn api runner and track down the cause.

piotr-szuberski:
Run Python PreCommit

piotr-szuberski:
Run Python PreCommit

piotr-szuberski:
Run Java PreCommit

piotr-szuberski:
Run SQL_Java11 PreCommit

piotr-szuberski:
Run Java PreCommit

piotr-szuberski:
Run XVR_Flink PostCommit

piotr-szuberski:
Run XVR_Flink PostCommit

piotr-szuberski:
Run XVR_Flink PostCommit

piotr-szuberski:
Run XVR_Flink PostCommit

piotr-szuberski:
Run XVR_Flink PostCommit

piotr-szuberski:
Run XVR_Flink PostCommit

piotr-szuberski:
@TheNeuralBit For some reason "Flink CrossLanguageValidatesRunner Tests" timeouts on xlang_jdbc test and on the other hand the same test passes on Python precommit suite.

At first it didn't see the URN as it wasn't added to the test expansion service. But now I have no idea what's going on, everything seems to be like in generate_sequence but for some reason my test gets hanged forever.

I changed the test to use EXPANSION_PORT when available because Postcommit suites run test expansion service and run tests with it.

Do you have any idea what could be the reason for the test to hang in this way?

TheNeuralBit:
From jenkins it looks like the test is timing out:
```
04:38:09 INFO:apache_beam.runners.portability.portable_runner:Job state changed to RUNNING
05:41:59 Build timed out (after 100 minutes). Marking the build as aborted.
```

I noticed a similar behavior when I try to run the test locally with
```
python setup.py nosetests \
        --tests apache_beam.io.external.xlang_jdbcio_it_test \
        --test-pipeline-options="--runner=FlinkRunner"
```
Eventually the job server logs that the pipeline is finished but then the test runner just hangs. I'm not sure where it's getting stuck

piotr-szuberski:
Run XVR_Flink PostCommit

piotr-szuberski:
Run XVR_Flink PostCommit

piotr-szuberski:
Run XVR_Flink PostCommit

piotr-szuberski:
Run XVR_Flink PostCommit

piotr-szuberski:
Yeah, I noticed that it times out.

Thanks that you tried to run the tests on your setup. Locally on my computer `:runners:flink:1.10:job-server:validatesCrossLanguageRunner` passes without a hassle, I don't get what is the reason the `XVR_Flink PostCommit` fails.

Especially that `Python PreCommit` also runs those tests and they pass.

Tomorrow I'll try to run them on Ubuntu, maybe OSX is the problem.. but I doubt it.

piotr-szuberski:
Run XVR_Spark PostCommit

piotr-szuberski:
Run XVR_Spark PostCommit

piotr-szuberski:
Run XVR_Spark PostCommitRun XVR_Spark PostCommit

piotr-szuberski:
Run XVR_Spark PostCommit

piotr-szuberski:
Run Python 3.7 PostCommit

piotr-szuberski:
Run Python 3.7 PostCommit

piotr-szuberski:
Run XVR_Flink PostCommit

piotr-szuberski:
Run Python 3.7 PostCommit

piotr-szuberski:
Run XVR_Flink PostCommit

piotr-szuberski:
Run Python 3.7 PostCommit

piotr-szuberski:
Run Python 3.8 PostCommit

piotr-szuberski:
Run Python 3.8 PostCommit

piotr-szuberski:
Run Python 3.7 PostCommit

piotr-szuberski:
Run Java PreCommit

piotr-szuberski:
Run Python 3.7 PostCommit

piotr-szuberski:
Run Python 3.8 PostCommit

piotr-szuberski:
Run Java PreCommit

piotr-szuberski:
Run Python 3.7 PostCommit

piotr-szuberski:
Run Python 3.8 PostCommit

piotr-szuberski:
Run Python 3.8 PostCommit

piotr-szuberski:
@TheNeuralBit I've moved the tests execution to python postcommit suite, like KafkaIO xlang test. Now the Jdbc tests pass, but I continously have `apache_beam.io.gcp.bigquery_write_it_test.BigQueryWriteIntegrationTests.test_big_query_write_new_types` failing. I checked the postcommit cron executions and it fails quite often. I'll trigger it from time to time to confirm whether it's a flake or not.

If you have further suggestions for code improvement then go ahead.

piotr-szuberski:
Run Python 3.7 PostCommit

piotr-szuberski:
Run Python 3.8 PostCommit

piotr-szuberski:
Run XVR_Flink PostCommit

TheNeuralBit:
Interesting. Looks like it is running fine on Python PostCommit, and I'm able to run it locally with the fn api runner as well:
```
python setup.py nosetests --tests apache_beam.io.external.xlang_jdbcio_it_test
```
(cc: @sclukas77 - it looks like this command should work for you now if you have all the dependencies built)

Running on Python PostCommit is fine with me, I just wanted to make sure we have it running continuously _somewhere_.

Can you file a jira so we don't forget to investigate the issue on Flink/Spark?

I'd like to take another look through the code when I'm fresh tomorrow, sorry this is taking so long :grimacing:

piotr-szuberski:
@TheNeuralBit Not a problem. Sorry for lots of changes, I still have little experience in submitting PRs to Beam.

Those tests run on Flink and Spark in Python Postcommit suite and pass, so I think there is no need for the ticket.

The problem was in validatesCrossLanguageRunner task. It has more complicated setup (run flink/spark job-service, then run test expansion service, use PortableRunner). But even though it passed locally on my setup so I wasn't able to reconstruct this issue. I added the issue:
https://issues.apache.org/jira/browse/BEAM-10429

piotr-szuberski:
Run Python 3.8 PostCommit

piotr-szuberski:
Run Python PreCommit

TheNeuralBit:
> Those tests run on Flink and Spark in Python Postcommit suite.

Oh really? I didn't realize that. How do you know? I'm having a hard time looking through the logs to verify since they're so large.

build.gradle
```diff
@@ -273,6 +273,7 @@ task python37PostCommit() {
 
 task python38PostCommit() {
   dependsOn ":sdks:python:test-suites:portable:py38:crossLanguagePythonJavaKafkaIOFlink"
+  dependsOn ":sdks:python:test-suites:portable:py38:crossLanguagePythonJavaJdbcIO"
```
TheNeuralBit, line 276: Ah sorry, I see now that this is how you know it's running on Flink and Spark :)

buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy
```diff
@@ -1982,17 +1982,25 @@ class BeamModulePlugin implements Plugin<Project> {
         return argList.join(' ')
       }
 
-      project.ext.toxTask = { name, tox_env ->
+      project.ext.toxTask = { name, tox_env, needsExpansionServiceJar = true ->
```
TheNeuralBit, line 1985: wouldn't it be preferable to make false the default? I'd think most tox tasks do not need this.

CHANGES.md
```diff
@@ -78,6 +78,8 @@
   is experimental. It reads data from BigQuery by exporting data to Avro files, and reading those files. It also supports
   reading data by exporting to JSON files. This has small differences in behavior for Time and Date-related fields. See
   Pydoc for more information.
+* Add cross-language support to JdbcIO.ReadRows([BEAM-10135](https://issues.apache.org/jira/browse/BEAM-10135)).
+* Add cross-language support to JdbcIO.Write([BEAM-10136](https://issues.apache.org/jira/browse/BEAM-10136)).
```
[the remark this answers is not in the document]
piotr-szuberski: Yeah, the time has passed :)

sdks/java/container/boot.go
```diff
@@ -122,6 +122,7 @@ func main() {
 		filepath.Join(jarsDir, "beam-sdks-java-harness.jar"),
 		filepath.Join(jarsDir, "beam-sdks-java-io-kafka.jar"),
 		filepath.Join(jarsDir, "kafka-clients.jar"),
+		filepath.Join(jarsDir, "beam-sdks-java-io-jdbc.jar"),
```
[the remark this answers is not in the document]
piotr-szuberski: You're right, that's a good news! :)

sdks/java/io/jdbc/build.gradle
```diff
@@ -25,6 +25,7 @@ description = "Apache Beam :: SDKs :: Java :: IO :: JDBC"
 ext.summary = "IO to read and write on JDBC datasource."
 
 dependencies {
+  compile library.java.postgres
```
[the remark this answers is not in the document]
piotr-szuberski: Good idea!

buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy
piotr-szuberski: Yeah, definitely. I made it true as default without thinking at the moment of writing it.

piotr-szuberski:
Run Python 3.8 PostCommit

piotr-szuberski:
Run Python 3.8 PostCommit

piotr-szuberski:
Run Python 3.8 PostCommit

piotr-szuberski:
Run Python PreCommit

piotr-szuberski:

piotr-szuberski:
Run Python 3.8 PostCommit

piotr-szuberski:
@TheNeuralBit I've done the requested changes :)

sdks/java/io/expansion-service/build.gradle
```diff
@@ -33,6 +33,8 @@ ext.summary = "Expansion service serving several Java IOs"
 dependencies {
   compile project(":sdks:java:expansion-service")
   compile project(":sdks:java:io:kafka")
+  runtime project(":sdks:java:io:jdbc")
   runtime library.java.kafka_clients
+  runtime library.java.postgres
```
TheNeuralBit, line 38: nit: It might not be clear why this is here
```suggestion
  // Include postgres so it can be used with external JDBC
  runtime library.java.postgres
```

CHANGES.md
```diff
@@ -55,6 +55,8 @@
 
 * New overloads for BigtableIO.Read.withKeyRange() and BigtableIO.Read.withRowFilter()
   methods that take ValueProvider as a parameter (Java) ([BEAM-10283](https://issues.apache.org/jira/browse/BEAM-10283)).
+* Add cross-language support to JdbcIO.ReadRows([BEAM-10135](https://issues.apache.org/jira/browse/BEAM-10135)).
+* Add cross-language support to JdbcIO.Write([BEAM-10136](https://issues.apache.org/jira/browse/BEAM-10136)).
```
TheNeuralBit, line 59:
```suggestion
* Add cross-language support to Java's JdbcIO, now available in the Python module `apache_beam.io.external.jdbc` ([BEAM-10135](https://issues.apache.org/jira/browse/BEAM-10135), [BEAM-10136](https://issues.apache.org/jira/browse/BEAM-10136)).
```

chamikaramj:
Thanks. Taking a look as well.

piotr-szuberski:
@TheNeuralBit @chamikaramj BTW the BigQuery tests (BigQueryWriteIntegrationTests.test_big_query_write_new_types, BigQueryQueryToTableIT.test_big_query_new_types_native and few others) in Python Postcommit are flaky enough to cause this job fail more often than pass. Is there a jira ticket for it?

Other thing that disturbs me is that those JdbcIO python tests run ~15min on Python Precommit. I think this is ok now but when other xlang transforms appear then it will be a serious problem - they'll probably need a separate precommit or something.

sdks/java/io/expansion-service/build.gradle
piotr-szuberski: Done

CHANGES.md
piotr-szuberski: Done

TheNeuralBit:
Why are they taking ~15min? When I run them locally they're quite fast

piotr-szuberski:
Each of them takes ~40-70s and they are run on cython, python-clouds on python3.5, 3.6, 3.7, 3.8

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/external/JdbcExternalRead.java
```diff
@@ -0,0 +1,85 @@
...
+import org.apache.beam.sdk.transforms.PTransform;
+import org.apache.beam.sdk.values.PBegin;
+import org.apache.beam.sdk.values.PCollection;
+import org.apache.beam.sdk.values.Row;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.ImmutableMap;
+
+/** Exposes {@link JdbcIO.ReadRows} as an external transform for cross-language usage. */
+@Experimental(Kind.PORTABILITY)
+@AutoService(ExternalTransformRegistrar.class)
+public class JdbcExternalRead implements ExternalTransformRegistrar {
```
chamikaramj, line 37: How about renaming this to "JdbcReadRowsRegistrar". Current name make this sounds like a transform while in reality this is just a utility to register the existing transform with the expansion service.

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/external/JdbcExternalRead.java
```diff
@@ -0,0 +1,85 @@
...
+
+  public static final String URN = "beam:external:java:jdbc:read:v1";
```
chamikaramj, line 39: Probably "beam:external:java:jdbc:read_rows:v1" is more unique to this transform

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/external/JdbcExternalWrite.java
```diff
@@ -0,0 +1,121 @@
...
+import org.apache.beam.sdk.transforms.PTransform;
+import org.apache.beam.sdk.values.PCollection;
+import org.apache.beam.sdk.values.PDone;
+import org.apache.beam.sdk.values.Row;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.collect.ImmutableMap;
+
+/** Exposes {@link JdbcIO.Write} as an external transform for cross-language usage. */
+@Experimental(Kind.PORTABILITY)
+@AutoService(ExternalTransformRegistrar.class)
+public class JdbcExternalWrite implements ExternalTransformRegistrar {
```
chamikaramj, line 42: JdbcWriteRegistrar

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/external/JdbcExternalWrite.java
```diff
@@ -0,0 +1,121 @@
...
+      DataSourceConfiguration dataSourceConfiguration = configuration.getDataSourceConfiguration();
+
+      // TODO: BEAM-10396 use writeRows() when it's available
+      return JdbcIO.<Row>write()
+          .withDataSourceConfiguration(dataSourceConfiguration)
+          .withStatement(configuration.statement)
+          .withPreparedStatementSetter(new XlangPreparedStatementSetter());
+    }
+
+    private static class XlangPreparedStatementSetter
```
chamikaramj, line 73: How about calling this "BeamRowPreparedStatementSetter" and moving this out of external directory ? This sounds like a more generic utility.

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/external/JdbcExternalWrite.java
```diff
@@ -0,0 +1,121 @@
...
+        implements JdbcIO.PreparedStatementSetter<Row> {
+      @Override
+      public void setParameters(Row row, PreparedStatement statement) throws SQLException {
+        List<Schema.Field> fieldTypes = row.getSchema().getFields();
+        for (int i = 0; i < fieldTypes.size(); ++i) {
+          Schema.TypeName typeName = fieldTypes.get(i).getType().getTypeName();
+          switch (typeName) {
+            case DATETIME:
```
chamikaramj, line 81: Please add unit tests for these conversions.

sdks/python/apache_beam/io/external/jdbc.py
```diff
@@ -0,0 +1,254 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
```
chamikaramj, line 5: Please move this to io module directly. "apache_beam/io/jdbc.py"

sdks/python/apache_beam/io/external/jdbc.py
```diff
@@ -0,0 +1,254 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""PTransforms for supporting Jdbc in Python pipelines.
+
+  These transforms are currently supported by Beam portable runners (for
+  example, portable Flink and Spark) as well as Dataflow runner.
```
chamikaramj, line 21: Is support for Dataflow runner confirmed and tested ?

sdks/python/apache_beam/io/external/jdbc.py
```diff
@@ -0,0 +1,254 @@
...
+  There are several ways to setup cross-language Jdbc transforms.
+
+  * Option 1: use the default expansion service
+  * Option 2: specify a custom expansion service
+
+  See below for details regarding each of these options.
+
+  *Option 1: Use the default expansion service*
+
+  This is the recommended and easiest setup option for using Python Jdbc
```
chamikaramj, line 40: This should be Beam 2.24.0 for Jdbc.

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/external/JdbcExternalRead.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+  }
+
+  public static class Builder
+      implements ExternalTransformBuilder<ReadConfiguration, PBegin, PCollection<Row>> {
+    @Override
+    public PTransform<PBegin, PCollection<Row>> buildExternal(ReadConfiguration configuration) {
+      DataSourceConfiguration dataSourceConfiguration = configuration.getDataSourceConfiguration();
+
+      JdbcIO.ReadRows readRows =
```
chamikaramj, line 71: How about adding unit tests for Builder classes for read and write ?

sdks/python/apache_beam/io/external/jdbc.py
```diff
@@ -0,0 +1,254 @@
...
+  """A PTransform which writes Rows to the specified database via JDBC.
+
+  This transform receives Rows defined as NamedTuple type and registered in
+  the coders registry, e.g.::
+
+    ExampleRow = typing.NamedTuple('ExampleRow',
+                                   [('id', int), ('name', unicode)])
+    coders.registry.register_coder(ExampleRow, coders.RowCoder)
+
+  An example can be found in
```
chamikaramj, line 126: Is this example being added in a separate PR ?

sdks/python/apache_beam/io/external/jdbc.py
```diff
@@ -0,0 +1,254 @@
...
+
+class WriteToJdbc(ExternalTransform):
+  """A PTransform which writes Rows to the specified database via JDBC.
+
+  This transform receives Rows defined as NamedTuple type and registered in
+  the coders registry, e.g.::
+
+    ExampleRow = typing.NamedTuple('ExampleRow',
+                                   [('id', int), ('name', unicode)])
+    coders.registry.register_coder(ExampleRow, coders.RowCoder)
```
chamikaramj, line 124: Let's add a simple code example here to showcase the usage of this transform.

sdks/python/apache_beam/io/external/jdbc.py
```diff
@@ -0,0 +1,254 @@
...
+
+class ReadFromJdbc(ExternalTransform):
+  """A PTransform which reads Rows from the specified database via JDBC.
+
+  This transform delivers Rows defined as NamedTuple registered in
+  the coders registry, e.g.::
+
+    ExampleRow = typing.NamedTuple('ExampleRow',
+                                   [('id', int), ('name', unicode)])
+    coders.registry.register_coder(ExampleRow, coders.RowCoder)
```
chamikaramj, line 200: Ditto regarding adding a simple code example here.

chamikaramj:
10 mins for pre-commit is a lot. Probably we should limit this to some of test combinations or Python versions. @tvalentyn for recommendations on this.

piotr-szuberski:
> 10 mins for pre-commit is a lot. Probably we should limit this to some of test combinations or Python versions. @tvalentyn for recommendations on this.

I'll limit the execution of these tests to python 3.8 with
```
@unittest.skipIf(
    sys.version_info <= (3, 8), 'Run the test on py38 or higher only')
```

sdks/python/apache_beam/io/external/jdbc.py
piotr-szuberski: No, the example leads to the integration test which is a quite straightforward example of using this transform.

buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy
```diff
@@ -1982,17 +1982,25 @@ class BeamModulePlugin implements Plugin<Project> {
...
 
           doLast {
             // Python source directory is also tox execution workspace, We want
             // to isolate them per tox suite to avoid conflict when running
             // multiple tox suites in parallel.
             project.copy { from project.pythonSdkDeps; into copiedSrcRoot }
-
+            if (needsExpansionServiceJar) {
+              def expansionServiceJar =  project.project(':sdks:java:io:expansion-service').shadowJar.archivePath
+              def expansionServiceDestinationDir = "${copiedSrcRoot}/sdks/java/io/expansion-service/build/libs"
```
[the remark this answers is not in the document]
piotr-szuberski: Yeah, these are python tox suites that include pythonLint, pythonFormatter, pythonDocs that do not require expansion service to be built for their executions. Only test suites require them.

sdks/java/io/expansion-service/build.gradle
```diff
@@ -33,6 +33,9 @@ ext.summary = "Expansion service serving several Java IOs"
 dependencies {
   compile project(":sdks:java:expansion-service")
   compile project(":sdks:java:io:kafka")
+  runtime project(":sdks:java:io:jdbc")
   runtime library.java.kafka_clients
+  // Include postgres so it can be used with external JDBC
+  runtime library.java.postgres
```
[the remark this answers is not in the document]
piotr-szuberski: To be honest I don't really know these dependencies work under the hood.

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/external/JdbcExternalRead.java
piotr-szuberski: Good point, I'll change it.

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/external/JdbcExternalRead.java
piotr-szuberski: Sure! Done.

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/external/JdbcExternalWrite.java
piotr-szuberski: Done

sdks/java/io/expansion-service/build.gradle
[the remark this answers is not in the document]
chamikaramj: Let's test/confirm before adding it :)
For example this should work for a distributed Flink deployment as well not just a local instance running in a single JVM.

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/external/JdbcExternalWrite.java
piotr-szuberski: Good idea. Done.

sdks/python/apache_beam/io/external/jdbc.py
piotr-szuberski: Ok, done

sdks/python/apache_beam/io/external/jdbc.py
piotr-szuberski: Sorry, no. I didn't manage to run in on Dataflow. I'll remove this information.

sdks/python/apache_beam/io/external/jdbc.py
piotr-szuberski: Right, done

sdks/python/apache_beam/io/external/jdbc.py
piotr-szuberski: Ok, done

sdks/python/apache_beam/io/external/jdbc.py
piotr-szuberski: Done

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/external/JdbcExternalRead.java
piotr-szuberski: Ok, I'll do that on Monday. Could it be on separate PR? This one is already a big one.

chamikaramj: A separate PR for this is fine. Thanks.

sdks/java/io/expansion-service/build.gradle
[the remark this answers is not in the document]
TheNeuralBit: They must be in the shadowJar though right? Otherwise the request to expand jdbc:read would fail immediately

sdks/java/io/jdbc/src/main/java/org/apache/beam/sdk/io/jdbc/external/JdbcExternalWrite.java
piotr-szuberski: @chamikaramj  I'll do a separate PR for this on Monday. I don't get how should I deal with such PRs that depend on each other? I'm used to Gerrit where every commit is a separate PR rebased on the previous commit. But here it would be a duplicate PR with one more commit. Is that ok or there is a smarter way to publish a PR based on the changes that are not merged yet and reviewed in another PR?

sdks/java/io/expansion-service/build.gradle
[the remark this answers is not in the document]
piotr-szuberski: I changed it to compile dependencies just to be on the safe side. Strange thing is that the size of the io-expansion-service has the same growth in size as with the runtime dependencies - from 51,4Mb to 52,7Mb - maybe it was why it worked.

I managed to start flink job server on docker and the io-expansion-server on non-default port and run the tests using the io-expansion-server. I'm not sure though whether it confirms it works or not.

buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy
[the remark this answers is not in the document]
piotr-szuberski: Sorry, I didn't read carefully. Tox suites use separate environment and BeamJarExpansionService('sdks:java:io:expansion-service:shadowJar') cannot find the jar if it is not copied into the tox's root path

piotr-szuberski:
@chamikaramj @TheNeuralBit  I've made the requested changes. Also, I found an utility in JdbcUtil that is able to call proper set method on PreparedStatement, so I think there is no need to test it separately anymore (there already is a test of this utility). I put BeamRowPreparedStatementSetter to JdbcUtil. I had to make the util class public instead of package-private because I wanted to have external classes in separate package.

I'll publish the external builder test after this PR is merged.

piotr-szuberski:
Run Python PreCommit

piotr-szuberski:
Run Python 3.8 PostCommit

piotr-szuberski:
Run Python PreCommit

piotr-szuberski:
Run Python PreCommit

piotr-szuberski:

piotr-szuberski:
Run Python PreCommit

piotr-szuberski:
Run Python PreCommit

piotr-szuberski:
Run Portable_Python PreCommit

piotr-szuberski:

piotr-szuberski:

piotr-szuberski:
Run Python PreCommit

piotr-szuberski:

piotr-szuberski:

piotr-szuberski:
Run Python PreCommit

piotr-szuberski:

piotr-szuberski:
Run PythonDocker PreCommit

piotr-szuberski:
Run Python PreCommit

piotr-szuberski:
Run Python PreCommit

piotr-szuberski:
Run Python 3.8 PostCommit

piotr-szuberski:
Run Python PreCommit

piotr-szuberski:
@chamikaramj ping

tvalentyn:
> 10 mins for pre-commit is a lot. Probably we should limit this to some of test combinations or Python versions. @tvalentyn for recommendations on this.

All precommit unit test suites (3.5, 3.6, 3.7, 3.8) run in parallel on the same machine, so if we add the tests to 3.8 only, total precommit time will increase to a similar extent as if we ran the test in all precommits.

buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy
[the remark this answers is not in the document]
tvalentyn: Why to we need a jar in tox (unit test) suite?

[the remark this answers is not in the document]
tvalentyn: I understand that we may need it for postcommit (integration tests), but why precommits-on-tox?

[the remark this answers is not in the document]
piotr-szuberski: Because I understood that we want these tests to run in precommit with fn_api runner. If we don't then I'll leave them running in postcommit suites (which is good enough as I assume) and disable them in precommit.

piotr-szuberski:
> versions

Ok, then I'll leave execution of these tests in postcommit only :)

build.gradle
chamikaramj, line 276: Probably add to other post-commits as well ?

sdks/python/apache_beam/io/jdbc.py
```diff
@@ -0,0 +1,272 @@
...
+
+  *Option 2: specify a custom expansion service*
+
+  In this option, you startup your own expansion service and provide that as
+  a parameter when using the transforms provided in this module.
+
+  This option requires following pre-requisites before running the Beam
+  pipeline.
+
+  * Startup your own expansion service.
```
chamikaramj, line 62: Let's mark new transforms here as experimental.

piotr-szuberski:
Run Python 3.8 PostCommit

build.gradle
piotr-szuberski: I'm ok with that. The reason why I chose py38 only was that KafkaIO cross-language tests are run on py38 only.

sdks/python/apache_beam/io/jdbc.py
piotr-szuberski: I added `Experimental; no backwards compatibility guarantees.` at the end of the docstring.

piotr-szuberski:
Run Python 3.8 PostCommit

piotr-szuberski:
Run Python 3.5 PostCommit

piotr-szuberski:
Run Python 3.6 PostCommit

piotr-szuberski:
Run Python 3.7 PostCommit

piotr-szuberski:
Run Python 3.8 PostCommit

piotr-szuberski:
> Thanks. LGTM other than a couple of comments.
> 
> Did the post-commit 3.8 tests trigger with new tests ? Could not find it in the list.

Yes, I could find the crossLanguagePythonJavaJdbcIO tasks running both for Flink and Spark. It runs in a separate task, so it's easy to find them in jenkins logs.

piotr-szuberski:
Run Python PreCommit

piotr-szuberski:
Run Python 3.6 PostCommit

piotr-szuberski:
Run Python 3.5 PostCommit

piotr-szuberski:

chamikaramj:
Hmm, any idea why it's not listed here ? https://ci-beam.apache.org/job/beam_PostCommit_Python37_PR/206/testReport/

build.gradle
tvalentyn: Some tests will run only in high-priority python versions and that's ok. @lazylynx  is working on making the workflow easy to configure in https://github.com/apache/beam/pull/12239.

piotr-szuberski:
> Hmm, any idea why it's not listed here ? https://ci-beam.apache.org/job/beam_PostCommit_Python37_PR/206/testReport/

It's for the same reason why Kafka's cross-language tests are not listed in python 3.8 postcommit: https://ci-beam.apache.org/job/beam_PostCommit_Python38_PR/21/testReport/

They run from a separate task, not from the postCommitIT which runs. It could probably be run via that task, could I add it as a task in Jira and look after it in a different PR?

build.gradle
piotr-szuberski: @tvalentyn should I remove this from other python versions then?

chamikaramj:
Can you please add it to regular post-commit please ? There's no point in adding an IT that we cannot easily track. I don't think following the path of Kafka IT is good since it's not run anywhere and not tracked.

piotr-szuberski:
> Can you please add it to regular post-commit please ? There's no point in adding an IT that we cannot easily track. I don't think following the path of Kafka IT is good since it's not run anywhere and not tracked.

It is run in python 3.8 postcommit. But I see the point, I'll try to run it in regular post-commit

chamikaramj:
Ah "python 3.8 postcommit" is good. And based on Valentyn's comment seems like this is by choice. So let's just wait for that to finish :))

tvalentyn:
You should be able to see task execution in console logs. To see meaningful signals in Jenkins test report,  logs need to be exported into xml test reports, like: https://github.com/apache/beam/blob/c8d21625235bd3f4d685383ebc4a04f6f5ace787/sdks/python/scripts/run_integration_test.sh#L276

piotr-szuberski:
Oh sorry, I pushed before I could see your response. But maybe that's ok if the job passes :)

piotr-szuberski:
Run Python 3.8 PostCommit

chamikaramj:
Sorry about not being clear. When I said regular PostCommit I did not mean Python 2. We should definitely run this in Python 3 post-commits. Let's add it back to Python 3 and check console log to make sure that the tasks run as Valentyn mentioned. A separate task will be to export logs to xml test reports for easy visibility.

piotr-szuberski:
> Sorry about not being clear. When I said regular PostCommit I did not mean Python 2. We should definitely run this in Python 3 post-commits. Let's add it back to Python 3 and check console log to make sure that the tasks run as Valentyn mentioned. A separate task will be to export logs to xml test reports for easy visibility.

They won't execute on python 2 because there are skips when there are no packages installed. I was able to remove the gradle task and run the test in `:sdks:python:test-suites:portable:py38:postCommitPy38`. It should publish the test results by default, so maybe it's worth to run it there? The only disadvantage is that `:sdks:java:container:docker` will run on all postcommit python jobs.

chamikaramj:
Sounds good. Thanks.

piotr-szuberski:
Run Python 3.8 PostCommit

piotr-szuberski:

piotr-szuberski:

chamikaramj:
1. leave the separate task and try to publish the test results
2. leave it in the postcommitIT

Either is fine by me. We just have to make sure that the IT you add here is successfully running in at least one active post-commit test suite and we can detect/track failures. Based on what you said, seems like it'll be running in "Python 3.8 PostCommit".

piotr-szuberski:
Run Java PreCommit

piotr-szuberski:
>     1. leave the separate task and try to publish the test results
> 
>     2. leave it in the postcommitIT
> 
> 
> Either is fine by me. We just have to make sure that the IT you add here is successfully running in at least one active post-commit test suite and we can detect/track failures. Based on what you said, seems like it'll be running in "Python 3.8 PostCommit".

Ok, it's already visible in postcommit tests results

piotr-szuberski:
> Nice! Thanks.

Ok, I'll run other python postcommits just to be on the safe side

piotr-szuberski:
Run Python 3.7 PostCommit

piotr-szuberski:
Run Python 2 PostCommit

piotr-szuberski:
Run Python 3.6 PostCommit

piotr-szuberski:
Run Python 3.5 PostCommit

piotr-szuberski:
> Nice! Thanks.

Should I rerun everything until everything is green or can I omit obvious flakes like the one in precommit?

chamikaramj:
We need tests to be green before merging.

chamikaramj:
Run Python PreCommit

piotr-szuberski:
Run Java PreCommit

piotr-szuberski:
Run Python 2 PostCommit

piotr-szuberski:
Run Python 3.5 PostCommit

piotr-szuberski:
Run Python 3.7 PostCommit

piotr-szuberski:
Run Python 3.7 PostCommit

piotr-szuberski:
Run Python 3.6 PostCommit

piotr-szuberski:
Run Python 3.6 PostCommit

piotr-szuberski:
Run Java PreCommit

piotr-szuberski:
Run Python PreCommit

piotr-szuberski:
Run Java PreCommit

piotr-szuberski:
Run Python 3.7 PostCommit

piotr-szuberski:
Run Python 3.7 PostCommit

piotr-szuberski:
@chamikaramj Ok, tests are green :)

egeucak:
hey, sorry for resurrecting the PR, but I don't really feel like Jira is being monitored that closely. I couldn't get "statement", and "query" arguments for python API to work. Any idea why that could be happening? @piotr-szuberski

